Issue #47, opened 2016-03-23
The FIDO Metadata Service needs a non-FIDO analogous mechanism


jcjones:
Right now the WebAuthn specification utilizes the FIDO Metadata Service to determine facts about authenticators, such as characteristics and certificates. We should consider developing a similar capability that does not depend on one organization.

rlin1:
There are specification for
a) Metadata Statements, i.e. a way how to describe authenticator characteristics and
b) Metadata Service, i.e. how to provide a list of authenticators with pointers to their respective authenticator characteristics.

These specs are not dependent on any specific service provider - they just describe the format.

nadalin:
These specifications are out of scope for this WG but if these can be referenced then we should do so.

vijaybh:
The spec just says FIDO metadata service is an example of the kind of place you can get data to make a trust decision. An RP may choose to use this, another service or none at all. Clarify text to say this.

Issue #46, opened 2016-03-21
Genericize nomenclature, where possible


jcjones:
Where possible, replace references to the "FIDO" trademark in the specification.

(From https://www.w3.org/2016/03/16-webauthn-minutes.html#action02)

vijaybh:
Closing as this was merged to master earlier today.

PR #48, opened 2016-03-23
Issue #46 - Update Nomenclature using non-trademarked wordings where possible

jcjones:
Generally, the following substitutions were made:
- Extensions were renamed from "fido." to "webauth."
- CredentialType "FIDO" was renamed to "ScopedUserCredential"
- "FIDO Authenticators" are now "WebAuth Authenticators"
- "FIDO Credential" and similar are now "Scoped Credential"
- "FIDO method" and similar are now "WebAuth method"
- "FIDO Relying Party" and similar are now "WebAuthn Relying Party"
- The WebIDL DOM interface is now type "WebAuthentication" and named "webauth"

I did not attempt to change the OIDs, references to the ECDAA specification, or the FIDO Metadata Service (see Issue #47).

**NOTE to reviewers**: As a byproduct of editing, some of the extraneous end-of-line whitespace got cleaned up as well, but not all of it. You may want to append `?w=1` to your URL to instruct Github to ignore whitespace changes.

### DIFF
--- a/index.src.html
+++ b/index.src.html
@@ -15,7 +15,7 @@ Editor: Michael B. Jones, Microsoft, mbj@microsoft.com
 Editor: Rolf Lindemann, Nok Nok Labs, rolf@noknok.com
 group: webauthn
 Ignored Vars: op, current.algorithm, alg
-Abstract: This specification defines an API that enables web pages to access WebAuth compliant strong cryptographic credentials through browser script. Conceptually, credentials are stored on a WebAuth authenticator, and each credential is bound to a single Relying Party. Authenticators are responsible for ensuring that no operation is performed without the user's consent. The user agent mediates access to credentials in order to preserve user privacy. Web Authentication uses the concept of attestation to provide a cryptographic proof of the authenticator model to the relying party.  The relying party can derive the authenticator security characteristics from this proof.
+Abstract: This specification defines an API that enables web pages to access WebAuthn compliant strong cryptographic credentials through browser script. Conceptually, credentials are stored on an authenticator, and each credential is bound to a single Relying Party. Authenticators are responsible for ensuring that no operation is performed without the user's consent. The user agent mediates access to credentials in order to preserve user privacy. Web Authentication uses the concept of attestation to provide a cryptographic proof of the authenticator model to the relying party.  The relying party can derive the authenticator security characteristics from this proof.
 Boilerplate: omit conformance, omit feedback-header
 Markup Shorthands: css off, markdown on
 </pre>
@@ -24,7 +24,7 @@ Markup Shorthands: css off, markdown on
 
   <em>This section is not normative.</em>
 
-  <p>This specification defines an API for web pages to access scoped credentials through JavaScript, for the purpose of strongly authenticating a user. Scoped credentials are always scoped to a single Relying Party, and the API respects this requirement. Credentials created by a Relying Party can only be accessed by web origins belonging to that Relying Party. Additionally, privacy across Relying Parties must be maintained; scripts must not be able to detect any properties, or even the existence, of credentials belonging to other Relying Parties.</p>
+  <p>This specification defines an API for web pages to access scoped credentials through JavaScript, for the purpose of strongly authenticating a user. Scoped credentials are always scoped to a single WebAuthn Relying Party, and the API respects this requirement. Credentials created by a Relying Party can only be accessed by web origins belonging to that Relying Party. Additionally, privacy across Relying Parties must be maintained; scripts must not be able to detect any properties, or even the existence, of credentials belonging to other Relying Parties.</p>
 
   <p>Scoped credentials are located on authenticators, which can use them to perform operations subject to user consent. Broadly, authenticators are of two types:</p>
 
@@ -42,7 +42,7 @@ Markup Shorthands: css off, markdown on
       <li>
         On the phone:
         <ul>
-          <li>User goes to example.com in the browser, and signs in using whatever method they have been using (possibly a pre-WebAuth method such as a password).</li>
+          <li>User goes to example.com in the browser, and signs in using whatever method they have been using (possibly a legacy method such as a password).</li>
           <li>The phone prompts, &quot;Do you want to register this device with example.com?&quot;</li>
           <li>User agrees.</li>
           <li>The phone prompts the user for a previously configured authorization gesture (PIN, biometric, etc.); the user provides this.</li>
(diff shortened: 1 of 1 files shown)


leshi:
I wonder whether bikeshed supports global variables... It would be nice to just have a $CREDENTIAL_NAME variable that we can re-use everywhere.  That way if next week, we decide to add a letter to the end, you won't have to go through this again :)  Just a thought.  Maybe it's not worth it.

index.src.html
```diff
@@ -15,7 +15,7 @@
...
+Abstract: This specification defines an API that enables web pages to access WebAuth compliant strong cryptographic credentials through browser script. Conceptually, credentials are stored on a WebAuth authenticator, and each credential is bound to a single Relying Party. Authenticators are responsible for ensuring that no operation is performed without the user's consent. The user agent mediates access to credentials in order to preserve user privacy. Web Authentication uses the concept of attestation to provide a cryptographic proof of the authenticator model to the relying party.  The relying party can derive the authenticator security characteristics from this proof.
```
vijaybh: Can we just always say "an authenticator" instead of "a WebAuth authenticator" since WebAuth itself writes no standards or conformance tests for authenticators? It may have made sense in the FIDO context but not here.

jcjones:
Comments from @equalsJeffH:

    Generally, the following substitutions were made:

        Extensions were renamed from "fido." to "webauth."

given the ambiguity of "auth" wrt authentication or authorization, I suggest we use "webauthn" here and in the items below that contain the (sub)string "webauth". 

        CredentialType "FIDO" was renamed to "ScopedUserCredential"

given the below "FIDO Credential" => "Scoped Credential", I'd make the above string "ScopedCredential" (or even "ScopedCred")

        "FIDO Authenticators" are now "WebAuth Authenticators"
        "FIDO Credential" and similar are now "Scoped Credential"
        "FIDO method" and similar are now "WebAuth method"
        "FIDO Relying Party" and similar are now just "Relying Party"

arguably the greatest point of misunderstanding between webauthn (nee FIDO) and the web federation / Web SSO worlds is using the unadorned term "Relying Party"  -- thus I suggest we do further qualify the term and try to faithfully use it. How about "WebAuthn Relying Party" ?

I've made a similar suggestion to the federation community that they(we) use the term "Federation Relying Party" in the fed context, or at least when one is discussing both webauthn and federation at the same time, eg where they interstect. 

[ a webauthn RP is typically the IDP in a fed setting, while a Fed RP is likely not, but may be, a webauthn RP ]

        The WebIDL DOM interface is now type "WebAuthentication" and named "webauth"

index.src.html
vijaybh: Maybe also make Authenticator a glossary term so it would show up bolded or whatever.

index.src.html
```diff
@@ -42,7 +42,7 @@
       <li>
         On the phone:
         <ul>
-          <li>User goes to example.com in the browser, and signs in using whatever method they have been using (possibly a pre-FIDO method such as a password).</li>
+          <li>User goes to example.com in the browser, and signs in using whatever method they have been using (possibly a pre-WebAuth method such as a password).</li>
```
vijaybh: Can we just say "legacy" instead of "pre-Webauth"?

index.src.html
```diff
@@ -89,13 +89,13 @@
...
 
 
 # Conformance # {#conformance}
 
-  <p>This specification defines criteria for a <dfn>conforming user agent</dfn>. A user agent MUST behave as described in this specification in order to be considered conformant. User agents MAY implement algorithms given in this specification in any way desired, so long as the end result is indistinguishable from the result that would be obtained by the specification's algorithms. A conforming FIDO Credential API user agent MUST also be a conforming implementation of the IDL fragments of this specification, as described in the “Web IDL” specification. [[!WebIDL-1]]</p>
+  <p>This specification defines criteria for a <dfn>conforming user agent</dfn>. A user agent MUST behave as described in this specification in order to be considered conformant. User agents MAY implement algorithms given in this specification in any way desired, so long as the end result is indistinguishable from the result that would be obtained by the specification's algorithms. A conforming Scoped Credential API user agent MUST also be a conforming implementation of the IDL fragments of this specification, as described in the “Web IDL” specification. [[!WebIDL-1]]</p>
```
vijaybh: It seems like our terminology is that we have a Web Authentication API which uses Scoped Credentials. So when we refer to the API it should be as the "Web Authentication API" not as the "Scoped Credential API".

index.src.html
```diff
@@ -174,79 +174,77 @@
...
+      DOMString          imageUri;
+  };
+
+  dictionary ScopedCredentialParameters {
+      required CredentialType        type;
+      required AlgorithmIdentifier   algorithm;
+  };
+
+  enum CredentialType {
+      "ScopedUserCredential"
```
vijaybh: Could we just say "Scoped" or "ScopedUser"? It is a CredentialType already.

index.src.html
```diff
@@ -174,79 +174,77 @@
...
+      Promise < ScopedCredentialInfo > makeCredential (
+          Account                                 accountInformation,
+          sequence < ScopedCredentialParameters > cryptoParameters,
+          DOMString                               attestationChallenge,
+          optional unsigned long                  credentialTimeoutSeconds,
+          optional sequence < Credential >        blacklist,
+          optional WebAuthExtensions              credentialExtensions
+      );
+
+      Promise < WebAuthAssertion > getAssertion (
```
vijaybh: It seems inconsistent to say WebAuthAssertion but not WebAuthAttestationStatement for example. OTOH Assertion seems too generic. Your thoughts?

index.src.html
```diff
@@ -761,7 +759,7 @@
 
 ### Attestation Models ### {#attestation-models}
       
-	<p>FIDO 2.0 specifies multiple attestation models:</p>
+	<p>WebAuth specifies multiple attestation models:</p>
```
vijaybh: Pending resolution of the other discussion on abstracting away attestation details, can we say something like "WebAuth supports..." instead of "WebAuth specifies..."?

index.src.html
```diff
@@ -995,7 +993,7 @@
 	<dl dfn-for="AttestationHeader">
 	  <dt>readonly attribute DOMString <dfn>claimedAAGUID</dfn></dt>
 	  <dd>The claimed Authenticator Attestation GUID (a version 4 GUID, see [[RFC4122]]).  
-	    This value is used by the FIDO Server to 
+	    This value is used by the server to
```
vijaybh: Relying Party?

index.src.html
jcjones: For the sake of right now, going with @equalsJeffH 's `ScopedCred`, but I'm also a fan of `ScopedUser`.

index.src.html
jcjones: Added to the list in #50.

index.src.html
jcjones: I think there's room for improvement there; I've opened #62 to discuss.

nadalin:
I'm uncomfortable renaming FIDO Relying Party to just Relying Party as they are not the same in any context, I suggest that we use Authenticator Relying Party or something that designates this different from what most people call a relying party

jcjones:
@nadalin: Per Jeff's comments, all places that were `FIDO Relying Party` are now `WebAuthn Relying Party` in the current PR.

vijaybh:
LGTM. Given the size of this change I think it is best to merge quickly and we can always fix up more things later if we find any.

It looks like the PR needs to be rebased, @jcjones can you do that and let me know so I can merge?

@equalsJeffH any objections to merging this in its current form?

jcjones:
@vijaybh: Rebased.

PR #54, opened 2016-03-29
TravisCI Integration

jcjones:
This PR introduces `lib/` containing [jcjones/bikeshed-template](https://github.com/jcjones/bikeshed-template), a minimalist fork of mt's [martinthomson/i-d-template](https://github.com/martinthomson/i-d-template) but intended for bikeshed documents. When I get time I may try to unify it back to mt's original, but I've made sure that won't cause any significant disruption here (except we'd want to rename `index.src.html` to `webauthn-something-something.src.html`.)

This changeset is integrated with Travis, [here's a run from the fork](https://travis-ci.org/jcjones/webauthn).

The change also supports auto-pushing to `gh-pages` should we want to do that, by setting a token into Travis's environment variables ([example](https://jcjones.github.io/webauthn/)). That's open to further discussion as said token would need to come from a robot account, and I believe W3C may have other mechanisms for building...?


vijaybh:
Looks good to me.

jcjones:
Rebased for the license changes

Issue #50, opened 2016-03-23
Add a Terminology Section


jcjones:
Some terms like `Relying Party` are known sources of confusion. We should add a terminology section to the specification that explains particular terms.

If you have a known confusable term, please add a comment to this issue so I can pick those up, too.

leshi:
Some quick thoughts: `Webapp Authentication Backend`, `Webapp Backend`, `Authentication Backend`, `Authentication Server`, `Webapp Server`

nicola:
OpenID uses `Relying Party` maybe it is a good idea to keep it and explain the term.

jcjones:
Also, "Authenticator"

flpm:
I took a first pass and made an initial list of terms. There are some generic terms we may not want to include, but since it's easier to cut I kept them anyway:

```
account
assertion
assertion challenge
assertion extension
attestation
attestation certificate
attestation challenge
attestation core
attestation header
attestation key
attestation statement
authentication
authenticator
authenticator attestation GUID (AAGUID)
authenticator data
authenticator operation
authenticator session
blacklist
browser
certificate authority (CA)
client
client data
client data hash
client extension
client platform
contextual binding
credential
credential extension
credential identifier
credential type
cryptographic algorithm
cryptographic key
cryptographic proof
cryptographic signature
DAA root key
device
Direct Anonymous Attestation (DAA)
effective top-level domain plus one (eTLD+1)
embedded authenticator
encoded client data
endorsement key
extension identifier
external authenticator
friendly fraud
global correlation handle
hash algorithm
intermediate CA
local storage
origin
privacy CA
private key
promise
public key
public suffix plus one (PS+1)
raw data
relying party
relying party identifier (RPID)
request
RogueList
root CA
script
secure context
server
standard extension
test of user presence
transaction authorization
user
user agent
user verification
user verification index (UVI)
web application
whitelist
```

jcjones:
Sorry for the delay on this; it's taking a bit more time than I anticipated, so it ran into today.

jcjones:
Closed by e3dc893a2325693e8ae627c0cd6ed0b293bd8a42; follow-ons to continue as needed.

PR #68, opened 2016-04-08
Issue #50: Add a Terminology Section

jcjones:
This is a first pass, placing the **Terminology** section just before **Acknowledgements**.

I made a pass through, attempting to ensure the first use of a term within a given section was always linked back to the definition.

I also made an effort to always spell out "WebAuthn Relying Party" for the first use in each section, and remove the use of "RP" (excepting the use of "RP ID").

There are some more whitespace cleanups as well, though I tried to minimize them. Again, I recommend reviewing using the `?w=1` URL such as:
https://github.com/w3c/webauthn/pull/68/files?expand=1&w=1


vijaybh:
LGTM as a starting point.

PR #69, opened 2016-04-13
Rename imageUri to imageURL

### LINKED ISSUE #59: TAG review feedback: imageUri
w3ctag/spec-reviews#97

imageUri should be renamed to imageURL, and could be aligned with manifest syntax for multiple sizes.

vijaybh:
Also removed the level from the metadata since it was appending a 1 to
the document title. Fixes #59. I decided not to mandate manifest
formatting of the icon to avoid the added complexity. Authenticators
would have to implement it, and we'd have to explore all the things a
bad person could do (e.g. have very different looking avatars at
different resolutions).

### DIFF
--- a/index.src.html
+++ b/index.src.html
@@ -4,7 +4,7 @@ Status: ED
 TR: http://www.w3.org/TR/webauthn/
 ED: http://w3c.github.io/webauthn/
 Shortname: webauthnapi
-Level: 1
+Level: 
 Editor: Vijay Bharadwaj, Microsoft, vijay.bharadwaj@microsoft.com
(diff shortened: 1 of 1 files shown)


index.src.html
```diff
@@ -4,7 +4,7 @@
 TR: http://www.w3.org/TR/webauthn/
 ED: http://w3c.github.io/webauthn/
 Shortname: webauthnapi
-Level: 1
+Level:
```
leshi: what's this ```Level``` change?

leshi:
LGTM

PR #73, opened 2016-04-19
Proposed refactor to separate web API and autehnticator model

vijaybh:
As discussed in the previous call, this is my proposal to separate the web API parts (which matter to UA implementers) from the authenticator parts (which matter to authenticator makers). Please take a look and comment on the direction. This will of course need whitespace cleanups and some proofreading later, but I am hoping that those things would not be critical to understanding the general direction.


equalsJeffH:
LGTM

jcjones:
Also LGTM. Build passes; merging to master.

PR #75, opened 2016-04-22
Markdown cleanup

vijaybh:
No changes to substance here. Goal of this change is to make the spec
easier to work on, not to fix the content.

- Wrap all lines at 128 chars (wide enough for modern screens, small
enough to fit in most commonly-sized windows)
- Remove extraneous HTML tags
- Convert HTML tags to Markdown equivalents
- Define macros for a couple of commonly used phrases so it's easier to
type and to keep consistency throughout the spec


equalsJeffH:
Looks REALLY Good To Me !

vijaybh:
Merging based on email feedback

PR #76, opened 2016-04-29
Simplify attestation structure by moving details into authnr model

### LINKED ISSUE #1: Abstraction layer for attestations
We are looking for the correct way to define an abstraction for attestations so that we don't have to go to operating-system particular (i.e. Android) discussions in the W3C spec.

vijaybh:
Fixes #1 

Moved most of the attestation details into the authenticator model
section, and flattened the structures for simplicity. Updated spec
language to match.

It may be possible to further simplify the attestation details in the
authenticator model section, but I would prefer that an attestation
expert weigh in on that.

I renamed the top-level attestation object to WebAuthnAttestation to match how the top-level signature object is called WebAuthnAssertion.

Also a few other small formatting fixups.


index.src.html
```diff
@@ -806,22 +757,59 @@
...
+    [[RFC4122]]). This value is used by the <a>[RP]</a> to look up the trust anchor for verifying the following signature. If
+    the verification succeeds, the AAGUID related to the trust anchor is trusted. This field MUST be present, if either no
+    attestation certificates are used (e.g., DAA) or if the attestation certificate doesn't contain the AAGUID value in an
+    extension.
+
+    The <dfn>x5c</dfn> attribute contains the attestation certificate and its certificate chain as described in [[!RFC7515]]
+    section 4.1.6.
+
+    The <dfn>alg</dfn> element contains the name of the algorithm used to generate the attestation signature according to
+    [[!RFC7518]]. See [[#packed-attestation-signature]] for the signature algorithms to be implemented by servers.
```
equalsJeffH: this reference is to section 3 of RFC7518, rather than section 4, yes?

vijaybh: I believe that is correct (I did not add that line, just moved it). However I can certainly add that detail. 

Also, I also notice that the reference in the next link (also merely moved) seems out of place. Maybe that sentence should just be removed, and we can let the individual attestation types specify what must be implemented by servers. WDYT?

vijaybh: Made these changes in following commit.

index.src.html
```diff
@@ -202,33 +202,20 @@
...
     };
 
-    interface AttestationStatement {
-        readonly    attribute AttestationHeader header;
-        readonly    attribute AttestationCore   core;
-        readonly    attribute DOMString         signature;
-    };
-
-    interface AttestationCore {
+    interface WebAuthnAttestation {
```
equalsJeffH: the prior AttestationStatement design mimicked that of JWT/JWS -- we may have a use case for that particular arrangement -- Rolf?  

The header is essentially the "trust" info and the core is the "data" portion.  This proposed re-structuring eliminates those distinctions.

index.src.html
```diff
@@ -202,33 +202,20 @@
...
-        readonly    attribute unsigned long version;
-        readonly    attribute DOMString     rawData;
         readonly    attribute DOMString     clientData;
-    };
-
-    interface AttestationHeader {
-        readonly    attribute DOMString   claimedAAGUID;
-        readonly    attribute DOMString[] x5c;
-        readonly    attribute DOMString   alg;
+        readonly    attribute any           statement;
```
equalsJeffH: why is this an `any` rather than of type AttestationStatement ?

index.src.html
```diff
@@ -504,64 +491,28 @@
...
-<div dfn-for="AttestationHeader">
-    The <dfn>claimedAAGUID</dfn> element contains the claimed Authenticator Attestation GUID (a version 4 GUID, see
-    [[RFC4122]]). This value is used by the <a>[RP]</a> to look up the trust anchor for verifying the following
-    {{AttestationCore}} object. If the verification succeeds, the AAGUID related to the trust anchor is trusted. This field MUST
-    be present, if either no attestation certificates are used (e.g., DAA) or if the attestation certificate doesn't contain the
-    AAGUID value in an extension.
-
-    The <dfn>x5c</dfn> attribute contains the attestation certificate and its certificate chain as described in [[!RFC7515]]
-    section 4.1.6.
+    The <dfn>clientData</dfn> member contains the base64url encoding of the parameters sent to the authenticator by the client.
```
equalsJeffH: s/client/client and [RP]/

index.src.html
vijaybh: See below for reasoning.

index.src.html
vijaybh: As I mentioned in the commit message, I think that for Android attestation it would be preferable for this just to become "DOMString safetyNetResponse" and thereby eliminate a bunch of unnecessary manipulation of bits on the client side. That would allow us to delete an entire (sub)section of the spec with no loss in expressiveness.

So keeping this as any allows us to accommodate other attestation structures as needed.

index.src.html
vijaybh: It's sent to the authenticator by the client. The client uses the RP-supplied parameters as input to its decision, sure, but what the client decides to send is what gets signed over, and the RP only has an indirect influence over that.

equalsJeffH: fine, but that needs to be explained.

index.src.html
```diff
@@ -504,64 +491,28 @@
...
-    of this specification.
-
-    The <dfn>version</dfn> member specifies the version number of the rawData object.
-
-    The <dfn>rawData</dfn> object contains the attested public key and the |clientDataHash|.
-
-    The <dfn>clientData</dfn> member contains the base64url encoding of <a>clientDataJSON</a> (see [[#signature-format]]). The
-    exact encoding must be preserved as the hash (clientDataHash) has been computed over it.
-</div>
+This attestation statement is delivered to the <a>[RP]</a> by the script, using methods outside the scope of this specification.
```
equalsJeffH: s/ <a>[RP]</a> by the script / <a>[RP]</a> server by the RP's script /

see also issue #80.

index.src.html
```diff
@@ -806,22 +757,59 @@
...
+        readonly    attribute DOMString[]   x5c;
+        readonly    attribute DOMString     alg;
+        readonly    attribute DOMString     rawData;
+        readonly    attribute DOMString     signature;
+    };
+</pre>
 
-### Overview ### {#attestation-overview}
+<div dfn-for="AttestationStatement">
+    The <dfn>version</dfn> member specifies the version number of the rawData object.
```
equalsJeffH: `version` is relative to the rawData type, yes?  if so, should mention that.

index.src.html
equalsJeffH: OK

index.src.html
equalsJeffH: OK by me.

equalsJeffH:
overall, LGTM.

leshi:
LGTM

equalsJeffH:
Looks fine to me.

index.src.html
vijaybh: Done

index.src.html
vijaybh: Done

jcjones:
The change looks good to me, too. Thanks @vijaybh! Let's get it merged, and then we can continue fixing the wording from #80.

rlin1:
Looks good to me. Let's do the merge.

vijaybh:
Thanks for all the reviews. Merging.

PR #77, opened 2016-04-30
Represent binary data as ArrayBuffers instead of base64-encoded DOMStrings

### LINKED ISSUE #61: TAG review feedback: Should we be using array types instead of base64-encoded DOMStrings?
I believe Alex Russell mentioned this in person at the San Francisco F2F, elaborating on comments in w3ctag/spec-reviews#97

It may seem more natural to use BufferSource, ArrayBuffer, or ArrayBufferView in various places where we have base64-encoded DOMStrings currently. This may also make it easier for script authors to manipulate these locally if needed (e.g. WebCrypto APIs take array types) and it avoids prescribing the use of base64.

vijaybh:
Fixes #61.

I switched the main API completely from base64-encoded DOMStrings to
Buffersource (for input parameters) and ArrayBuffer (for output
parameters). The actual signatures are still computed over the same data
as before, so signatures computed after this change will be compatible
with those computed before, except for being represented differently.

I moved the ClientData section into the Authenticator model section
since it is not directly used by script authors. This structure still
does base64 encoding of the challenge, for two reasons. First, this
maintains backward compatibility. Second, it is more natural to
represent a binary challenge in JSON as base64 rather than the clunky
array notation.

### DIFF
--- a/index.src.html
+++ b/index.src.html
@@ -161,14 +161,14 @@ The API is defined by the following Web IDL fragment.
         Promise < ScopedCredentialInfo > makeCredential (
             Account                                 accountInformation,
             sequence < ScopedCredentialParameters > cryptoParameters,
-            DOMString                               attestationChallenge,
+            BufferSource                            attestationChallenge,
             optional unsigned long                  credentialTimeoutSeconds,
             optional sequence < Credential >        blacklist,
             optional WebAuthnExtensions             credentialExtensions
         );
 
         Promise < WebAuthnAssertion > getAssertion (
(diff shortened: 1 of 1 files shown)


jcjones:
Your decision to maintain the Base64-encoding in the `ClientData` structure makes sense to me. The other revisions, including the section move and the type changes, I also endorse.

LGTM. r=me.

equalsJeffH:
LGTM.

vijaybh:
Merging this in now that we have two reviews.

PR #83, opened 2016-05-07
Lots of FPWD issues

vijaybh:
This branch fixes 12 editorial issues, including all the ones currently marked for FPWD. The API and formats are not changed, only the wording around them.

### DIFF
--- a/index.src.html
+++ b/index.src.html
@@ -17,7 +17,7 @@ group: webauthn
 Text Macro: RP WebAuthn Relying Party
 Text Macro: RPS WebAuthn Relying Parties
 Text Macro: INFORMATIVE <em>This section is not normative.</em>
-Ignored Vars: op, alg
+Ignored Vars: op, alg, type, algorithm
 Abstract: This specification defines an API that enables web pages to access WebAuthn compliant strong cryptographic credentials
     through browser script. Conceptually, one or more credentials are stored on an authenticator, and each credential is scoped
     to a single Relying Party. Authenticators are responsible for ensuring that no operation is performed without the user's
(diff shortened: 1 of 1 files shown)


index.src.html
```diff
@@ -61,8 +61,8 @@
 ## Registration (<a>embedded authenticator</a> mode) ## {#registration-embedded}
 
 - On the phone:
-    * User goes to example.com in the browser, and signs in using whatever method they have been using (possibly a legacy
-        method such as a password).
+    * User goes to example.com in the browser and signs in to an existing account using whatever method they have been using
```
leshi: I know my English is sometimes a bit Russian, but I think we want `into` here
http://blog.oxforddictionaries.com/2015/08/in-to-versus-into/

vijaybh: Second opinion?

IMO "signs in" is the verb here, "to" is the preposition. That would advocate for leaving this as is, unless you're saying that "signs" is the verb and "into" the direction.

equalsJeffH: I nominally agree with @leshi   -- further, i suggest choosing between these two possible resolutions: 
a) use "sign-in"/"signs-in" (or "sign-on"/"signs-on") consistently in spec rather than "sign in"/"signs in", and then this sentence has "signs-in to"
b) use "signs into" here

fwiw, I prefer (a).

equalsJeffH:
I am nominally fine with merging these commits to master given that this is a working draft and nothing is cast in stone as yet.  

I am somewhat concerned that we are diluting the description/definition of the RP ID approach in terms of how it also applies to native platform APIs and is a cohesive approach which does not obviate the Web Same Origin Policy and does not supplant federated identity approaches.  I argue that we (where "we" is some superset containing interested W3C WebAuthn WG participants) need to determine where this overall "RP ID and facet denotation" approach is defined/published (it seems there are nominally valid arguments that the W3C is not the applicable place), and do so, at least as a explanatory document.

index.src.html
vijaybh: I still believe "sign in to XXX" (no hyphen) is the correct usage.

Examples:

- "[Sign in to your account](https://login.microsoftonline.com/)" (see page title)
- "[Sign in to Gmail](https://support.google.com/mail/answer/8494?hl=en)"
- "[Sign in to iCloud.com](https://support.apple.com/kb/PH2610?locale=en_US)"

(Paypal seems to use "Log in" everywhere but they don't hyphenate either.)

vijaybh:
Merging since this is all minor editorial issues, and two reviewers seem okay with this overall. Let's create new issues for the additional concerns.

Jeff, let's address your RP ID concerns in #8.

Thanks!

equalsJeffH:
ok w/ addressing RP ID concerns in #8.

PR #93, opened 2016-05-12
Update reference to Secure-Contexts (from Powerful-Features)

jcjones:


equalsJeffH:
Do we actually need to add the secure-contexts ref to biblio.js ?  Should not just 's/[[powerful-features]]/[[secure-contexts]]/' within index.src.html properly include the ref via specref.org ?

jcjones:
It threw fatals locally, even after an update.
>   FATAL ERROR: Couldn't find 'secure-contexts' in bibliography data. Did you mean:  [....]

jcjones:
Here's a revision; let's see what happens at CI. Could be my local bikeshed, I suppose.

jcjones:
CI took it, and this is purely editorial; merging.

Issue #66, opened 2016-04-07
Terminology consistency


flpm:
There are several places on the document what looks like the same concept is addressed by different names. In some cases it seems to be inherited from the generic terminology used in the FIDO version of the document. In some cases the same term is capitalized differently or abbreviated. Some examples:

- browser vs user agent
- platform vs client platform vs OS vs OS platform
- RP vs Relying Party vs relying party

There might be cases where the meaning is not exactly the same but we should be careful to explicitly notice those nuances.

equalsJeffH:
see also #80

Issue #80, opened 2016-05-03
adopt consistent terms for RP server-side and client-side components


equalsJeffH:
see also issue #358

We use the terms `script`, `a script`, `the script`, `scripts`, `browser script`, `javascript`, `WebAuthn Relying Party script`, `WebAuthn Relying Party client-side app`, `the client-side app`, `RP’s client-side app`, and `the WebAuthn Relying Party client-side application` to refer to the same **RP client-side component**.  

We use the terms `the server`, `server`, `servers`, and `WebAuthn Relying Party’s server` to refer to the same **RP server-side component**. 

We should decide on one consistent series of terms (and acronyms, as appropriate) to cover all the necessary semantic bases. 

some ideas: 

[ Note: " {....}" denotes optionality and possibly use of acronyms such as "RP", "RP's", "RPs". ]

client-side:   {  [[RP]] | [[RPS]]  }  script    // plus create glossary entry and clearly equate to "WebAuthn Relying Party's client-side webapp component" (or unambiguous equivalent).  

server-side:  {  [[RP]] | [[RPS]]  }  server   // plus create glossary entry and clearly equate to "WebAuthn Relying Party's server-side webapp component" (or unambiguous equivalent).

nadalin:
@selfissued can you please verify that the document is consistent

AngeloKai:
From what I am reading here, the issue shouldn't change the name of the API. I am removing the subtype:renaming tag.

equalsJeffH:
see also #358

nadalin:
@equalsJeffH will clean up dups

equalsJeffH:
I checked the current editors' draft and most of the cruft I'd noted in the orig post has been cleaned up, so am closing this.

PR #109, opened 2016-05-21
Editorial polishing of abstract, intro material, terminology, etc.

### LINKED ISSUE #92: move biblio.js into in-line <pre class=biblio> block
we should move biblio.js into an in-line `<pre class=biblio>` block within index.src.html per bikeshed developer Tab Atkins, see..
https://lists.w3.org/Archives/Public/spec-prod/2016AprJun/0021.html

equalsJeffH:
The commits represented in this PR comprise the following edits:

* Polish abstract.

* Convert "Use Cases" section into an "Introduction" containing a "use cases" subsection.

* Add explanatory material to Introduction.

* Make modest clarifying alterations to the use cases proper.

* Polish/clarify "Conformance" section.

* Add RFC2119 statement and reference. 

* Move "Terminology" section to be after "Conformance".

* Polish/Expand "Terminology" section. 

* Move biblio.json into index.src.html and rename biblio.json to biblio.json.bak

* Added some folks to #acknowledgements.

Note 1: this PR includes an experiment regarding section titles that refer to WebAuthn API WebIDL objects -- altered the WebAuthn Assertion section to contain a link directly into the IDL rather than be self-referential -- seems more useful to me, WDYT ?

Note 2: this PR creates a `diffs` subdirectory, containing a diff of rendered plain-text of this index.src.html  from the latest in master. See https://github.com/w3c/webauthn/tree/vijaybh-lots-of-fpwd-issues-jeffh-build/diffs

Fixes #92.

makes some progress on #66 et al.


biblio.json.bak
```diff
@@ -1,4 +1,12 @@
...
+    "title": "Ceremony Design and Analysis",
+    "href": "https://eprint.iacr.org/2007/399.pdf",
+    "authors": ["Carl Ellison"],
+    "date": "2007"
+  },
+
+	
   "GeoJSON": {
     "title": "The GeoJSON Format Specification",
     "href": "http://geojson.org/geojson-spec.html"
```
vijaybh: This should be deleted too?

vijaybh: Also, the PDF file below should also be removed from the commit. (I can't comment on the file itself since it's binary.)

index.src.html
```diff
@@ -17,96 +17,99 @@
...
 Text Macro: INFORMATIVE <em>This section is not normative.</em>
+Text Macro: WAC WebAuthn Client
 Ignored Vars: op, alg, type, algorithm
-Abstract: This specification defines an API that enables web pages to access WebAuthn compliant strong cryptographic credentials
-    through browser script. Conceptually, one or more credentials are stored on an authenticator, and each credential is scoped
-    to a single Relying Party. Authenticators are responsible for ensuring that no operation is performed without the user's
-    consent. The user agent mediates access to credentials in order to preserve user privacy. Authenticators use attestation
-    to provide cryptographic proof of their properties to the relying party. This specification also describes a functional
-    model of a WebAuthn compliant authenticator, including its signature and attestation functionality.
+Abstract: This specification defines a Web API enabling the creation and use of  strong, attested, cryptographic, scoped credentials
```
vijaybh: Should there be a comma after "cryptographic"? I thought this and the preceding words were adjectives for the noun phrase "scoped credential"?

index.src.html
```diff
@@ -17,96 +17,99 @@
...
+    by web applications, for the purpose of strongly authenticating users. Conceptually, one or more credentials, scoped
```
vijaybh: A lot of these lines are over 128 characters...

index.src.html
```diff
@@ -17,96 +17,99 @@
...
-User Agent implementing the Web Authentication API and the authenticator that holds the credential, by constraining the
-availability and usage of credentials. Scoped credentials created by a [RP] can only be accessed by web origins belonging to that
-[RP]. Additionally, privacy across [RPS] must be maintained; scripts must not be able to detect any properties, or even the
-existence, of scoped credentials belonging to other [RPS].
+This specification defines a Web API enabling the creation and use of strong, attested, cryptographic, <em><a>scoped credentials</a></em> by web applications, 
+for the purpose of strongly authenticating users. 
+A <a>scoped credential</a>, created and stored, subject to <em><a>user verification</a></em>, by an <em><a>authenticator</a></em> at the behest of a <em><a>[RP]</a></em>, can only be accessed by web origins belonging to that [RP]. 
+This scoping is enforced jointly by <em><a>conforming User Agents</a></em> and <em><a>authenticators</a></em>. 
+Additionally, privacy across <a>[RPS]</a> is maintained; [RPS]' browser scripts are not able to detect any properties, or even the existence, of credentials scoped to other [RPS].
```
vijaybh: Can we avoid using the term "browser script"? It is inconsistent with the use of "web application" for the same concept in the abstract above. Also, not all web clients are browsers, so a more neutral term is better.

index.src.html
```diff
@@ -17,96 +17,99 @@
...
 
-This specification defines an API for web pages to access scoped credentials through JavaScript, for the purpose of strongly
-authenticating a user. Scoped credentials are always scoped to a single <a>[RP]</a>. This scoping is enforced jointly by the
-User Agent implementing the Web Authentication API and the authenticator that holds the credential, by constraining the
-availability and usage of credentials. Scoped credentials created by a [RP] can only be accessed by web origins belonging to that
-[RP]. Additionally, privacy across [RPS] must be maintained; scripts must not be able to detect any properties, or even the
-existence, of scoped credentials belonging to other [RPS].
+This specification defines a Web API enabling the creation and use of strong, attested, cryptographic, <em><a>scoped credentials</a></em> by web applications, 
+for the purpose of strongly authenticating users. 
+A <a>scoped credential</a>, created and stored, subject to <em><a>user verification</a></em>, by an <em><a>authenticator</a></em> at the behest of a <em><a>[RP]</a></em>, can only be accessed by web origins belonging to that [RP].
```
vijaybh: Too many parentheticals in this sentence. May be better to split it up into two or more sentences if needed to avoid all the commas.

index.src.html
vijaybh: "Web API for use by web applications" seems redundantly redundant. Suggest dropping the first web".

index.src.html
```diff
@@ -17,96 +17,99 @@
...
 Ignored Vars: op, alg, type, algorithm
-Abstract: This specification defines an API that enables web pages to access WebAuthn compliant strong cryptographic credentials
-    through browser script. Conceptually, one or more credentials are stored on an authenticator, and each credential is scoped
-    to a single Relying Party. Authenticators are responsible for ensuring that no operation is performed without the user's
-    consent. The user agent mediates access to credentials in order to preserve user privacy. Authenticators use attestation
-    to provide cryptographic proof of their properties to the relying party. This specification also describes a functional
-    model of a WebAuthn compliant authenticator, including its signature and attestation functionality.
+Abstract: This specification defines a Web API enabling the creation and use of  strong, attested, cryptographic, scoped credentials
+    by web applications, for the purpose of strongly authenticating users. Conceptually, one or more credentials, scoped
+    to a given Relying Party, are created and stored on an authenticator. Authenticators are responsible for ensuring that no operation is performed without user
```
vijaybh: Ambiguous. Could be read to imply that all the credentials are scoped to a given relying party. Should say "each scoped to..." instead.

index.src.html
```diff
@@ -17,96 +17,99 @@
...
-    to a single Relying Party. Authenticators are responsible for ensuring that no operation is performed without the user's
-    consent. The user agent mediates access to credentials in order to preserve user privacy. Authenticators use attestation
-    to provide cryptographic proof of their properties to the relying party. This specification also describes a functional
-    model of a WebAuthn compliant authenticator, including its signature and attestation functionality.
+Abstract: This specification defines a Web API enabling the creation and use of  strong, attested, cryptographic, scoped credentials
+    by web applications, for the purpose of strongly authenticating users. Conceptually, one or more credentials, scoped
+    to a given Relying Party, are created and stored on an authenticator. Authenticators are responsible for ensuring that no operation is performed without user 
+    consent. The user agent mediates access to credentials in order to preserve user privacy. Authenticators provide cryptographic proof of their properties to relying parties via attestation. This specification also describes the functional
```
vijaybh: Suggest moving the user agent sentence up, to a point before you start describing the role of authenticators. That would allow the text to flow naturally into defining the model for authenticators. Otherwise you are ping-ponging from authenticators to user agents and back to authenticators.

index.src.html
```diff
@@ -17,96 +17,99 @@
...
+Additionally, privacy across <a>[RPS]</a> is maintained; [RPS]' browser scripts are not able to detect any properties, or even the existence, of credentials scoped to other [RPS].
 
-Scoped credentials are located on <a>authenticators</a>, which can use them to perform operations subject to user consent.
-Broadly, authenticators are of two types:
+[RPS] employ the <a>Web Authentication API</a> during two distinct, but related, <a>ceremonies</a> involving a user. The first is <a>Registration</a>, where a <a>scoped credential</a> is created on an <a>authenticator</a>, and associated by the <a>[RP]</a> with the present user's account (the account may already exist or may be created at this time). The second is <a>Authentication</a>, where a <a>[RP]</a> is presented a <em><a>WebAuthn Assertion</a></em> proving the present user is the same as the user who registered the <a>scoped credential</a> associated with the user account. Functionally, the <a>Web Authentication API</a> comprises two methods (along with associated data structures): <a>makeCredential()</a> and <a>getAssertion()</a>. The former is used during <a>Registration</a> and the latter during <a>Authentication</a>.
```
vijaybh: I really like this paragraph but the sentence about Authentication is really overcomplicated. Can't we just say "[RP] is presented with a WebAuthn Assertion signed by the credential to signify the user's consent to the requested operation" or something like that?

index.src.html
```diff
@@ -17,96 +17,99 @@
...
+
+Broadly, <a>authenticators</a> are of two types:
 
 1. <dfn>Embedded authenticators</dfn> have their operation managed by the same computing device (e.g., smart phone, tablet,
     desktop PC) as the user agent is running on. For instance, such an authenticator might consist of a Trusted Platform Module
-    (TPM) or Secure Element (SE) integrated into the computing device, along with appropriate platform software to mediate
-    access to this device's functionality.
+    (TPM) or Secure Element (SE) integrated into the computing device in conjunction with some biometric sensor (e.g., fingerprint), along with appropriate platform software to mediate
```
vijaybh: Why does it have to be a biometric sensor? It's just some method for the user to authenticate to the device and indicate consent - it could be a PIN or just a button tap for example.

index.src.html
```diff
@@ -17,96 +17,99 @@
...
 module containing the credential is functioning as an embedded authenticator. However, the credential may also be accessed over
 BLE by a user agent on a nearby laptop. In this latter case, the phone is functioning as an external authenticator. These modes
-may even be used in a single end-to-end user scenario. One such scenario is described in the remainder of this section.
+may even be used in a single end-to-end user scenario. One such scenario is described among the use cases below (in [[#authentication-external]]).
+
 
+## Use Cases ## {#use-cases}
 
-## Registration (<a>embedded authenticator</a> mode) ## {#registration-embedded}
+The below use case scenarios illustrate both <a>embedded authenticator</a> and <a>external authenticator</a> modes, as well as outline further scenarios.
```
vijaybh: Isn't this basically content-free? Is there a need to add this extra level of sectioning?

index.src.html
```diff
@@ -17,96 +17,99 @@
...
     * User sees a discreet prompt or notification, "Sign in to example.com."
     * User selects this prompt / notification.
     * User is shown a list of their example.com identities, e.g., "Sign in as Alice / Sign in as Bob."
-    * User picks an identity, is prompted for an authorization gesture (PIN, biometric, etc.) and provides this.
+    * User picks an identity, is prompted for an <a>authorization gesture</a> (PIN, biometric, etc.) and provides this.
 
-- Now, on the laptop:
-    * Web page shows that the selected user is signed in, and navigates to the signed-in page.
+- Now, back on the laptop:
+    * Web page shows that the selected user is signed-in, and navigates to the signed-in page.
```
vijaybh: Should be no hyphen in the first instance. "Signed-in" is only hyphenated when used as an adjective, as in the second instance. When used as a verb, it is just "signed in", similar to how one writes "he was ushered in" and not "he was ushered-in".

index.src.html
```diff
@@ -17,96 +17,99 @@
...
+### Other use cases and configurations ### {#other-configurations}
 
 A variety of additional use cases and configurations are also possible, including (but not limited to):
 
-- User goes to example.com on their laptop, is guided through a flow to create and register a credential on their phone.
+- User navigates to example.com on their laptop, is guided through a flow to create and register a credential on their phone.
 
-- User employs a scoped credential as described above to authorize a single transaction, such as a payment or other financial
-    transaction.
+- A [RP] can cause a signed-in user to be prompted for their <a>authorization gesture</a> in order to authorize a single transaction, such as a payment or other financial transaction.
```
vijaybh: Unnecessarily wordy. Also the user need not be signed in for the RP to authorize a transaction. Can we reduce to simply "A [RP] prompts the user for their authorization gesture..."? That would also match the voice used in the other bullet points.

index.src.html
```diff
@@ -17,96 +17,99 @@
...
+specification in order to be considered conformant. <a>Conforming User Agents</a> MAY implement algorithms given in this specification in any way
 desired, so long as the end result is indistinguishable from the result that would be obtained by the specification's
-algorithms. A conforming Web Authentication API User Agent MUST also be a conforming implementation of the IDL fragments of this
+algorithms. A conforming User Agent MUST also be a conforming implementation of the IDL fragments of this
 specification, as described in the “Web IDL” specification. [[!WebIDL-1]]
 
-This specification also defines a model of a Web Authentication compliant authenticator. This is a set of functional and
-security requirements for an authenticator to be usable by a User Agent that implements the Web Authentication API. As described
-in [[#use-cases]], the authenticator itself may be implemented in the operating system underlying the User Agent, or in external
+This specification also defines a model of a conformant Web Authentication authenticator (see [[#authenticator-model]]). This is a set of functional and
```
vijaybh: Can we just say "Web Authenticator"? "Web Authentication authenticator" seems inelegant.

index.src.html
```diff
@@ -137,8 +140,72 @@
...
+"OPTIONAL" in this document are to be interpreted as described in
+[[!RFC2119]].
+
+
+# Terminology # {#terminology}
+
+: <dfn>Assertion</dfn>
+:: See <a>WebAuthn Assertion</a>.
+
+: <dfn>Attestation Certificate</dfn>
```
vijaybh: If we simplify attestation as Alexei has suggested, perhaps this would go away?

index.src.html
```diff
@@ -137,8 +140,72 @@
...
+
+: <dfn>Registration</dfn>
+:: The <a>ceremony</a> where a user, a <a>[RP]</a>, and the user's computing device(s) (containing at least one <a>authenticator</a>) work in concert to create a <a>scoped credential</a> and associate it with the user's <a>[RP]</a> account.  Note that this includes employing <a>user verification</a>.
+
+: <dfn>Relying Party Identifier</dfn>
+: <dfn>RP ID</dfn>
+:: A Relying Party Identifier is derived from a <a>[RP]</a>'s web origin's hostname by computing the hostname's <a>eTLD+1</a>. 
+    
+: <dfn>Scoped Credential</dfn>
+:: Generically, a credential is data one entity presents to another in order to authenticate the former's identity [[RFC4949]]. A <em><a>scoped credential</a></em>, as defined here, is one that is cryptographically bound to a <a>[RP]</a>'s <a>RP ID</a>. Per specification, it is conveyed--i.e., registered--to only the <a>[RP]</a> instigating the credential's creation (see: <a>Registration</a>). Subsequently, only that <a>[RP]</a>, as identified by it's <a>RP ID</a> is able to employ the <a>scoped credential</a> in <a>authentication</a> ceremonies.
```
vijaybh: Re: "a credential is data one entity presents to another" this is clearly not true for credentials as described in this spec since the credential is a key pair and that is never presented to anyone. Perhaps more accurate to say "a credential is data that can be used by one entity..."?

index.src.html
```diff
@@ -137,8 +140,72 @@
...
+
+: <dfn>Relying Party Identifier</dfn>
+: <dfn>RP ID</dfn>
+:: A Relying Party Identifier is derived from a <a>[RP]</a>'s web origin's hostname by computing the hostname's <a>eTLD+1</a>. 
+    
+: <dfn>Scoped Credential</dfn>
+:: Generically, a credential is data one entity presents to another in order to authenticate the former's identity [[RFC4949]]. A <em><a>scoped credential</a></em>, as defined here, is one that is cryptographically bound to a <a>[RP]</a>'s <a>RP ID</a>. Per specification, it is conveyed--i.e., registered--to only the <a>[RP]</a> instigating the credential's creation (see: <a>Registration</a>). Subsequently, only that <a>[RP]</a>, as identified by it's <a>RP ID</a> is able to employ the <a>scoped credential</a> in <a>authentication</a> ceremonies. 
+
+: <dfn>User Consent</dfn>
+:: In this specification, the term <em>user consent</em> is equivalent to <a>user verification</a>.
```
vijaybh: I disagree. Consent means the user agrees with what they are being asked. User verification is the technical means by which this is conveyed to the authenticator. So for example user consent should encompass reading and understanding the prompt for transaction confirmation, while user verification is just the gesture that the user employs to indicate this.

index.src.html
equalsJeffH: sorry, but I don't have a tool to conveniently enforce a particular line length, especially long ones.  perhaps we can set up some filter that fixes up such things at merge-to-master time?

index.src.html
equalsJeffH: by "this", are you referring to  <dfn>Attestation Certificate</dfn> ?

index.src.html
equalsJeffH: see "Tutorial" section of the "$ credential" entry in https://tools.ietf.org/html/rfc4949, especially last parag...
```
   $ credential
    [...]
      Tutorial: In general English, "credentials" are evidence or
      testimonials that (a) support a claim of identity or authorization
      and (b) usually are intended to be used more than once (i.e., a
      credential's life is long compared to the time needed for one
      use). [...]

      The problem with adopting this term in Internet security is that
      an automated process for authentication or access control usually
      requires multiple steps using multiple data objects, and it might
      not be immediately obvious which of those objects should get the
      name "credential".

      For example, if the verification step in a user authentication
      process employs public-key technology, then the process involves
      at least three data items: (a) the user's private key, (b) a
      signed value -- signed with that private key and passed to the
      system, perhaps in response to a challenge from the system -- and
      (c) the user's public-key certificate, which is validated by the
      system and provides the public key needed to verify the signature.
      -  Private key: The private key is *not* a credential, because it
         is never transferred or presented. Instead, the private key is
         "authentication information", which is associated with the
         user's identifier for a specified period of time and can be
         used in multiple authentications during that time.
      -  Signed value: The signed value is *not* a credential; the
         signed value is only ephemeral, not long lasting. The OSIRM
         definition could be interpreted to call the signed value a
         credential, but that would conflict with general English.
      -  Certificate: The user's certificate *is* a credential. It can
         be "transferred" or "presented" to any person or process that
         needs it at any time. A public-key certificate may be used as
         an "identity credential", and an attribute certificate may be
         used as an "authorization credential".
```

biblio.json.bak
equalsJeffH: i"m not sure what you're referring to by "This should be deleted too" ?

WRT the PDF diff file -- yeah, that is an experiment for including a textual diff of the changes in the PR.  if it doesn't work for folks then we can remove it and the /diffs subdir.

index.src.html
equalsJeffH: OK to both. thx.

index.src.html
equalsJeffH: OK.  suggestion for term?  see also issue #80

biblio.json.bak
equalsJeffH: oh, you mean delete the file biblio.json.bak ?   sure, if we're fine with moving the biblio to be in-line in the spec source.

index.src.html
equalsJeffH: whoops, actually, 'browser script' is not held over, so this should be addressed.

index.src.html
equalsJeffH: I am assuming you're referring to the sentence in line 38.  Ok.

index.src.html
equalsJeffH: ok. though, one could have multiple different credentials mapped to a given RP, each for a different RP account.  in any case I added "each" as suggested. thx.

index.src.html
equalsJeffH: ok.

index.src.html
equalsJeffH: Hm, though I like being explicit about tying it up to the user level which is what the sentence is attempting to do, without diving into crypto details. 

how about simplifying to:  the [RP] is presented with a WebAuthn Assertion proving the present user is the the one who registered the scoped credential.    ?

index.src.html
equalsJeffH: hm, it's just an example and says "might consist of", tho have edited to presently be "or a Secure Element (SE) integrated into the computing device in conjunction with some means for user verification, ..."

index.src.html
equalsJeffH: sort of -- some readers may find it useful that the Use Cases have a ToC entry.  Also, I'll add to that intro text that there are further use case + code examples in a later spec section (in order to address one of AdamP's comments), ok?

index.src.html
equalsJeffH: oh gosh, I don't really care that much, but it seems to me that in the present cited sentence (L90), the first instance of "signed in"/"signed-in" is not a verb but a noun as it is being used to summarize/name the user's present state of being signed-in:

  `Web page shows that the selected user is signed-in, ...`

..rather than saying something like..

 `The user was signed in by the web app...`
 `The user signs in to the web app..."

index.src.html
equalsJeffH: ok. good catch.

index.src.html
equalsJeffH: ok.

index.src.html
equalsJeffH: Ok, good catch.  how about..

User consent means the user agrees with what they are being asked, i.e., it encompasses reading and understanding prompts. User verification encompasses the means employed by the user to indicate consent.

index.src.html
vijaybh: Still reads backwards to me. How about:  the [RP] is presented with a WebAuthn Assertion proving the presence and consent of the user who registered the scoped credential?

I think your text seems to imply that the presence and consent of a user is given, and the RP is trying to figure out who it is. But it seems to me that presence and consent is also being proved.

index.src.html
vijaybh: Yes, agree.

index.src.html
vijaybh: OK

index.src.html
vijaybh: So then if the credential is the public key, then presenting it does not authenticate identity. Maybe in that case we should say it is a piece of data that identifies a user and which can be used in a user authentication process to prove identity?

index.src.html
vijaybh: works for me.

index.src.html
vijaybh: ok

biblio.json.bak
vijaybh: yes, I mean the .bak and the .pdf seem to have become part of the commit and they should not be.

index.src.html
equalsJeffH: Ok, did s/[RPS]' browser scripts/[RPS]/

biblio.json.bak
equalsJeffH: ok. removed (once i push)

index.src.html
equalsJeffH: I will manually wrap at 128 this time around, but its a pain. also, for viewing the github-provided diffs, and my screen real-estate, I would prefer a line length on the order of 80 chars...

index.src.html
equalsJeffH: ah the key portion being "the presence and consent of the user" -- incorporated. thx.

index.src.html
```diff
@@ -17,97 +17,114 @@
...
 
+## Use Cases ## {#use-cases}
 
-## Registration (<a>embedded authenticator</a> mode) ## {#registration-embedded}
+The below use case scenarios illustrate both <a>embedded authenticator</a> and <a>external authenticator</a> modes, as well as
+outline further scenarios. Additional scenarios, including sample code, are given later in [[#sample-scenarios]].
 
-- On the phone:
-    * User goes to example.com in the browser and signs in to an existing account using whatever method they have been using
+### Registration (<a>embedded authenticator</a> mode) ### {#registration-embedded}
```
vijaybh: Small thing - could we change this title from "Registration (embedded authenticator mode)" to "Registration (with an embedded authenticator)"? Would make it more friendly and less jargony-sounding especially for readers who are just reading the introduction.

index.src.html
```diff
@@ -17,97 +17,114 @@
...
     * User agrees.
-    * The phone prompts the user for a previously configured authorization gesture (PIN, biometric, etc.); the user provides
-        this.
+    * The phone prompts the user for a previously configured <a>authorization gesture</a> (PIN, biometric, etc.); the user
+        provides this.        
     * Website shows message, "Registration complete."
 
 
-## Authentication (<a>external authenticator</a> mode) ## {#authentication-external}
+### Authentication (<a>external authenticator</a> mode) ### {#authentication-external}
```
vijaybh: See above suggestion on making these titles more friendly.

vijaybh:
Thanks for doing this Jeff! The document reads a lot better now.

I made one more small suggestion regarding section titles. Other than that, Travis CI is complaining that the document isn't valid markdown due to whitespace issues. Would you please take a look?

Other than that, LGTM.

index.src.html
equalsJeffH: sounds fine to me. done.

index.src.html
equalsJeffH: ok, done.

equalsJeffH:
whitespace issues fixed along with other minor tweaks.   am still working on credential definition.

index.src.html
equalsJeffH: Here's an updated candidate Scoped Credential definition:
<blockquote>
Scoped Credential: 
Generically, a credential is data one entity presents to another in order to authenticate the former’s identity [RFC4949]. A WebAuthn scoped credential is an { identifier, type } pair identifying authentication information established by the authenticator and the [RP], together, at registration time.  The authentication information consists of an asymmetric key pair, where the public key portion is returned to the [RP]. The private key is mapped to the [RP]'s RP ID and held by the authenticator.  Subsequently, only that [RP], as identified by its RP ID, is able to employ the scoped credential in authentication ceremonies, via the getAssertion() method. The [RP] uses the stored public key to verify the generated WebAuthn Assertion.
</blockquote>

Thoughts?

equalsJeffH: I went ahead and polished the above a bit, committed, and pushed it.  see commit ecee518

equalsJeffH:
latest commit includes refined scoped credential definition.

vijaybh:
Looks great, thanks. Merging since nobody else has expressed a desire to review this.

PR #110, opened 2016-05-25
[5 second review] Fix spec shortname to prep for FPWD

vijaybh:
webauthnapi -> webauthn

This change is already made in the FPWD copy, the PR just syncs up the repo with it.


vijaybh:
Merging this in since this has already been discussed and achieved consensus on the list.

PR #130, opened 2016-06-15
Clean up extensions section

### LINKED ISSUE #51: AAGUID extension underspecified
The description of AAGUID extension in attestation certificate in FIDO 2 is little under-specified. See https://www.w3.org/Submission/2015/SUBM-fido-key-attestation-20151120/#attestation-statement-certificate-requirements

Look at he extension value for Extension OID 1.3.6.1.4.1.45724.1.1.4 (id-fido-gen-ce-aaguid).

### LINKED ISSUE #97: Clarify when an extension may be ignored by user agent
There does not seem to be any reason for a web authn. client to block an inbound or outbound extension that is supported by the authenticator and does not require any client processing.  Last paragraph in 5 does not make this clear.

Recommend following text:

"All WebAuthn extensions are optional for both clients and authenticators. Thus, any extensions requested by a WebAuthn Relying Party may be ignored by the client browser or OS (if client processing is required for the extension) and not passed to the authenticator at all, or they may be ignored by the authenticator. However, an extension that is supported by an authenticator that does not require client processing should be passed to the authenticator.  Ignoring an extension is never considered a failure in the WebAuthn API, so when WebAuthn Relying Parties include extensions with any API calls, they must be prepared to handle cases where some or all of those extensions are ignored."

vijaybh:
- Resolve inconsistent wording in different parts of the section: fix
places that incorrectly suggested extensions cannot be used with
makeCredential, and resolve conflict between wording that said client
arguments are required vs. not.
- Treat implementations that choose to pass through extensions as a
first-class citizen by requiring that extensions be defined to allow for
this. Fixes #97.
- Fix all the predefined extensions to conform to the updated
guidelines.
- Minor issues: AAGUID should be binary not string (see #61, fixes #51)
and encode true as Boolean in CBOR instead of forcing it to integer
value 1.

### DIFF
--- a/index.src.html
+++ b/index.src.html
@@ -1422,9 +1422,9 @@ calls, they must be prepared to handle cases where some or all of those extensio
 Clients wishing to support the widest possible range of extensions may choose to pass through any extensions that they do not
 recognize to authenticators, generating the authenticator argument by simply encoding the client argument in CBOR. All
 WebAuthn extensions MUST be defined in such a way that this implementation choice does not endanger the user's security or
-privacy. However, if an extension requires client processing, such a naïve pass-through may result in a semantically invalid
-authenticator argument, which in turn may result in the extension being ignored by the authenticator. Since all extensions are
-optional, this will not cause a functional failure in the API operation.
+privacy. For instance, if an extension requires client processing, it could be defined in a manner that ensures such a naïve
+pass-through will produce a semantically invalid authenticator argument, resulting in the extension being ignored by the
+authenticator. Since all extensions are optional, this will not cause a functional failure in the API operation.
 
 
 ## Extension identifiers ## {#extension-id}
--- a/index.src.html
+++ b/index.src.html
@@ -1401,24 +1401,21 @@ signature extension. Extensions can define additions to the following steps and 
 
 - Client processing, and the {{ClientData}} structure, for registration extensions and signature extensions.
 
-- Authenticator processing, and the <a>authenticatorData</a> structure, for signature extensions.
+- Authenticator processing, and the <a>authenticatorData</a> structure, for registration extensions and signature extensions.
 
 When requesting an assertion for a scoped credential, a [RP] can list a set of extensions to be used, if they are supported by
-the client and/or the authenticator. It sends the request parameters for each extension in the {{getAssertion()}} call (for
+the client and/or the authenticator. It sends the client arguments for each extension in the {{getAssertion()}} call (for
 signature extensions) or {{makeCredential()}} call (for registration extensions) to the client platform. The client platform
 performs additional processing for each extension that it supports, and augments {{ClientData}} as required by the extension.
-For extensions that the client platform does not support, it passes the request parameters on to the authenticator when possible
-(criteria defined below). This allows one to define extensions that affect the authenticator only.
-
-Similarly, the authenticator performs additional processing for the extensions that it supports, and augments
-|authenticatorData| as specified by the extension.
-
-Extensions that are not supported are ignored.
+In addition, the client collects the authenticator arguments for the above extensions, and passes them to the authenticator in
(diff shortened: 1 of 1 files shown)


index.src.html
```diff
@@ -1436,23 +1433,21 @@
...
-An extension defines two request arguments. The <dfn>client argument</dfn> is passed from the <a>[RP]</a> to the client in the
-{{getAssertion()}} or {{makeCredential()}} call, while the <dfn>authenticator argument</dfn> is passed from the client to the
-authenticator during the processing of these calls.
-
-Extension definitions MUST specify the valid values for their client argument. Clients are free to ignore extensions with an
-invalid client argument. Specifying an authenticator argument is optional, since some extensions may only affect client
-processing.
+An extension defines up to two request arguments. The <dfn>client argument</dfn> is passed from the <a>[RP]</a> to the client
+in the {{getAssertion()}} or {{makeCredential()}} call, while the <dfn>authenticator argument</dfn> is passed from the client
+to the authenticator during the processing of these calls.
```
equalsJeffH: this parag has always been ambiguous as to whether it is possible for an RP to pass an `authenticator argument` that the client then simply passes on to the authnr.  If that was an original intention of the definition of extensions, AND we intend to continue that, then lines 1453-1455 ought not be deleted.

index.src.html
```diff
@@ -1401,24 +1401,21 @@
 
 - Client processing, and the {{ClientData}} structure, for registration extensions and signature extensions.
 
-- Authenticator processing, and the <a>authenticatorData</a> structure, for signature extensions.
+- Authenticator processing, and the <a>authenticatorData</a> structure, for registration extensions and signature extensions.
```
equalsJeffH: we need to add "registration extension" and "signature extension" (whcih should perhaps be "assertion extn" or "authn extn") to the terminology section.

index.src.html
```diff
@@ -1436,23 +1433,21 @@
...
+`authenticatorData` (in the case of signature extensions), or both. Finally, if the extension requires any authenticator 
+processing, it must also specify an authenticator argument to be sent via the <a>authenticatorGetAssertion</a> or 
+<a>authenticatorMakeCredential</a> call.
 
-Note: An extension that does not define additions to {{ClientData}} nor `authenticatorData` is possible, but should be avoided.
-    In such cases, the [RP] would have no indication whether the extension was supported or processed by the client and/or
-    authenticator.
+Note: Any extensions that do not define additions to {{ClientData}} nor `authenticatorData` must be defined in a way that allows
+    the [RP] to unambiguously infer, from the returned signature or attestation statement, whether the intent of the extension
+    was followed by the client and/or authenticator. See the [[#extension-authenticator-selection]] for an example.
```
equalsJeffH: sam suggested on the 15-jun webauthn call: Extensions should add to client data in some way to indicate to the RP that the extension was honored/respected/processed.

index.src.html
vijaybh: Lines 1453-1455 are not deleted, they are just moved down (see 1462-1464 in the green, below). Lines 1474+ explicitly say that what you ask is true, except for transcoding (the RP passes a client argument in JSON, it gets sent as an authenticator argument in the corresponding CBOR). Perhaps we can do a round of editorial improvements later?

index.src.html
vijaybh: Changed "signature extension" to "authentication extension" since we define registration and authentication to be our two ceremonies. I will open a new issue to add the terms to the glossary, since I am having trouble coming up with polished definitions right now.

index.src.html
vijaybh: Done.

index.src.html
equalsJeffH: thanks for clarification, sounds fine.

leshi:
LGTM

PR #134, opened 2016-06-30
Rename index.src.html to index.bs

jcjones:
The Bikeshed documentation indicates that the preferred filename for the input
file into Bikeshed tool should use the suffix ".bs" [1]. We chose to start with
".src.html" instead because the templates we were working from used that form,
but it's now deprecated.

This change renames the input file to match Bikeshed's recommendations.

1) https://github.com/tabatkins/bikeshed#bikeshed-file-extension


jcjones:
Several days of no objections on the mailing list, so merging.

PR #135, opened 2016-06-30
Text aesthetic updates

jcjones:
Here are a number of aesthetic update commits as a result of reviewing the document again from the perspective of a Web API consumer.

First, I have removed the distinctions between "external" and "embedded" authenticators. This distinction caused confusion for the web developers I solicited feedback from, as one of the goals of the API is to abstract that sort of hardware awareness away.

From a functional point of view, the document's scope is limited to User Agent <-- API --> Relying Party, and by agreement in the spring it shouldn't deal with hardware, so I argue that talk of external and embedded authenticators, while interesting, distracts from the nuts-and-bolts of actually using the standard.

Thus, this first change removes those distinctions, while keeping discussion that authenticators come in many types, and the user agent will abstract their differences away from the API.

Second, I changed the macro for "WebAuthn Relying Party" to the less-clumsy "Relying Party." Per the concerns in April about being a confusable for the same term in OAuth or X.509, I added a green "note" block.

Third, I moved the whole attestation type definitions into their own top-level section. They're a pretty top-level concept anyway, and as it was they were buried down in headings like "5.3.2.3.1."

Last, I added myself as an editor, per the first F2F in San Francisco.

Each of these things is its own standalone commit, should we need to pull bits back apart.


index.bs
```diff
@@ -52,43 +52,37 @@ consent of the user who registered the <a>scoped credential</a>. Functionally, t
...
-2. <dfn>External authenticators</dfn> operate autonomously from the computing device running the user agent, and are accessed
-    over a transport such as Universal Serial Bus (USB), Bluetooth Low Energy (BLE) or Near Field Communications (NFC). 
-
-Note that an external authenticator may itself contain an embedded authenticator. For example, consider a smart phone that
-contains a scoped credential. The credential may be accessed by a web browser running on the phone itself. In this case the
-module containing the credential is functioning as an embedded authenticator. However, the credential may also be accessed over
-BLE by a user agent on a nearby laptop. In this latter case, the phone is functioning as an external authenticator. These modes
-may even be used in a single end-to-end user scenario. One such scenario is described among the use cases below (in
-[[#authentication-external]]).
+Broadly, <a>authenticators</a> operate in different ways, but compliant
```
vijaybh: Suggest deleting the words "<a>authenticators</a> operate in different ways, but" since they add little information content to the sentence.

index.bs
```diff
@@ -14,8 +14,8 @@ Editor: Jeff Hodges, PayPal, Jeff.Hodges@paypal.com
 Editor: Michael B. Jones, Microsoft, mbj@microsoft.com
 Editor: Rolf Lindemann, Nok Nok Labs, rolf@noknok.com
 group: webauthn
-Text Macro: RP WebAuthn Relying Party
-Text Macro: RPS WebAuthn Relying Parties
+Text Macro: RP Relying Party
+Text Macro: RPS Relying Parties
```
vijaybh: I'm glad we stared using macros for these things :)

vijaybh:
Thanks, LGTM.

PR #136, opened 2016-06-30
Use W3C-standard bikeshed/CI scripts

jcjones:
Copied from https://github.com/w3c/html

This change updates the CI scripts to match the brand-new, much simpler setup from `w3c/html`. It also freezes `.spec-data` and uses a static copy of `bikeshed` to reduce the chances of us getting transient build problems due to updates upstream.


jcjones:
As this is a utility-only change, as soon as I verify the CI build for this PR, I will merge it, and verify the CI build for master.

PR #137, opened 2016-07-03
moved AAGUID into packed attestation core. Removed AAGUID extension.

rlin1:
Please review.


apowers313:
How would non-packed attestations have a GUID associated with a record in a metadata service?

equalsJeffH:
overall LGTM, tho @apowers313 asks a pertinent question...

Issue #60, opened 2016-03-30
TAG review feedback: Align Credential interface with Credential Management?


vijaybh:
w3ctag/spec-reviews#97

vijaybh:
Possible conflict with #61 - the Credential Management spec uses USVString for credential IDs. Should we push for ArrayBuffer?

equalsJeffH:
for reference:
http://heycam.github.io/webidl/#idl-buffer-source-types
http://heycam.github.io/webidl/#idl-USVString

vijaybh:
My understanding of the current consensus on this one is that we believe this alignment no longer makes sense. So I propose to:
- Move this interface from the global window namespace to the navigator namespace to reduce pollution of the global namespace
- Rename interface Credential to ScopedCredential to fix the name collision with Credential Management.
- Possibly rename CredentialType to version to make it clearer what this field signifies (since ScopedCredential is itself arguably a credential type and ScopedCredentialType is unwieldy)

Feedback welcome.

jeisinger:
I'd be interested in learning about why alignment no longer makes sense?

RByers:
Yes more details would be useful.  FWIW a feature may not pass blink's "intent to ship" process if there is a concern from TAG that hasn't been at least adequately evaluated in detail in a public forum.  So you should expect that, despite being closed, this issue is still a potential blocker for shipping in Chrome.

vijaybh:
Reopening to track the action item in http://www.w3.org/2017/02/22-webauthn-minutes.html for moving the namespace to navigator.credentials.crypto.

selfissued:
We should attempt to arrive at final resolutions for all the renaming issues for the Implementer's Draft.

selfissued:
This can be closed after merging PR #384

PR #142, opened 2016-07-11
API aesthetic updates

jcjones:
These changes are also aesthetic, similar to PR #135, but in this case they are all related to API aesthetics. They are still, for the most part, editorial in nature. It'd be good to get some care in the review of the `makeCredential()` updates per the removal of `ScopedCredentialParameters`, though.

First, per Vijay's comment [on issue 60](https://github.com/w3c/webauthn/issues/60#issuecomment-230693738), moving the interface from the global window namespace to the navigator namespace to reduce pollution of the global namespace. This also makes semantic sense, as the HTML5 `window` object represents an open window in the browser, whereas the `navigator` object represents the browser itself. Since authenticators belong to the whole system, not one window, the object model makes more sense to be a property attached to `navigator`.

Second, I've renamed the factory object from `navigator.webauthn` to `navigator.authentication`. In general, web APIs shouldn't use the word "web" in their name. It's implicit. (I considered "authn" but I feel the verbosity here is justified -- most of the users will assign the object to a local variable for ease of use anyway, like our example code does.)

Third, I changed the type of `ScopedCredentialInfo.publicKey` from `any` to `CryptoKey`, referencing the WebCryptoAPI for the type information. This feels like a fairly substantial change, but it really isn't: javascript gets flexibility to ask for the pubkey in a variety of formats (including JsonWebKey), and we get a tighter (and more expected) type.

Fourth, I've removed the `ScopedCredentialParameters` tuple by unpairing the `type` and `algorithm` fields. `ScopedCredentialParameters` is used in the procedure for normalizing algorithms and types, but said procedure doesn't actually use the types! This changes the makeCredential() method to instead have a list of supported types and a list of supported algorithms, and perform its procedure that way.

Keeping these together in an object gives a kind of flexibility, but after consideration, I don't see much actual purpose in said flexibility. A RP can either handle an algorithm or not, and is either willing to handle an authenticator type or not -- particular combinations seem an unnecessary burden.

Fifth, in response to [Vijay's note about extension names](https://lists.w3.org/Archives/Public /public- webauthn/2016Jul/0046.html), I've renamed the example extensions (correcting a CBOR mismatch in the process). I also noted in section 7.1, Extension identifiers, that  "Use of dot-separated notation here does not imply an object hierarchy.", and provided a counterexample of a differently- separated, versioned id: `mycompany-myextension_v01`. None of the pre-defined extensions were modified.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -126,7 +126,7 @@ or a combination of both.
 This specification relies on several other underlying specifications.
 
 : HTML5
-:: The concept of <dfn>origin</dfn> and the <dfn>Window</dfn> interface are defined in [[!HTML5]].
+:: The concept of <dfn>origin</dfn> and the <dfn>Navigator/dfn> interface are defined in [[!HTML5]].
 
 : Web IDL
 :: Many of the interface definitions and all of the IDL in this specification depend on [[!WebIDL-1]]. This updated version of
@@ -272,7 +272,7 @@ operation. Since this is an integral part of the WebAuthn security model, user a
 The API is defined by the following Web IDL fragment.
 
 <pre class="idl">
-    partial interface Window {
+    partial interface Navigator {
         readonly attribute WebAuthentication webauthn;
     };
 
@@ -1849,7 +1849,7 @@ This is the first time flow, when a new credential is created and registered wit
 The sample code for generating and registering a new key follows:
 
 <pre class="example highlight">
-    var webauthnAPI = window.webauthn;
+    var webauthnAPI = navigator.webauthn;
 
     if (!webauthnAPI) { /* Platform not capable. Handle error. */ }
 
@@ -1929,7 +1929,7 @@ If the [RP] script does not have any hints available (e.g., from locally stored 
 then the sample code for performing such an authentication might look like this:
 
 <pre class="example highlight">
-    var webauthnAPI = window.webauthn;
+    var webauthnAPI = navigator.webauthn;
 
     if (!webauthnAPI) { /* Platform not capable. Handle error. */ }
 
@@ -1950,7 +1950,7 @@ performing such an authentication might look like the following. Note that this 
 extension for transaction authorization.
 
 <pre class="example highlight">
-    var webauthnAPI = window.webauthn;
+    var webauthnAPI = navigator.webauthn;
 
     if (!webauthnAPI) { /* Platform not capable. Handle error. */ }
 
--- a/index.bs
+++ b/index.bs
@@ -1451,14 +1451,12 @@ authenticator. Since all extensions are optional, this will not cause a function
 ## Extension identifiers ## {#extension-id}
 
 Extensions are identified by a string, chosen by the extension author. Extension identifiers should aim to be globally unique,
-e.g., by using reverse domain-name of the defining entity such as `com.example.webauthn.myextension`.
-
-Note: Use of dot-separated notation here does not imply an object hierarchy.
+e.g., by including the defining entity such as `myCompanyExtension`.
 
 Extensions that may exist in multiple versions should take care to include a version in their identifier. In effect, different
-versions are thus treated as different extensions, e.g., `mycompany-myextension_v01`
+versions are thus treated as different extensions, e.g., `myCompanyExtension01`
 
-Extensions defined in this specification use a fixed prefix of `webauthn.` for the extension identifiers. This prefix should not
+Extensions defined in this specification use a fixed prefix of `webauthn` for the extension identifiers. This prefix should not
 be used for extensions not defined by the W3C.
 
 
@@ -1487,7 +1485,7 @@ A [RP] simultaneously requests the use of an extension and sets its client argum
 
 <pre class="example highlight">
     var assertionPromise = credentials.getAssertion(..., /* extensions */ {
-        "webauthn-example.foobar": 42
+        "webauthnExampleFoobar": 42
     });
 </pre>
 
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -1410,10 +1410,12 @@ authenticator. Since all extensions are optional, this will not cause a function
 Extensions are identified by a string, chosen by the extension author. Extension identifiers should aim to be globally unique,
 e.g., by using reverse domain-name of the defining entity such as `com.example.webauthn.myextension`.
 
+Note: Use of dot-separated notation here does not imply an object hierarchy.
```
vijaybh: Could we go further and just strip the dots out of all the examples and predefined extensions? We could advise camel-casing, e.g. WebAuthnGeo or webauthnTxauthSimple. That seems better than straddling two boats as we'd be doing here.

index.bs
```diff
@@ -296,7 +298,7 @@ The API is defined by the following Web IDL fragment.
 
     interface ScopedCredentialInfo {
         readonly attribute Credential           credential;
-        readonly attribute any                  publicKey;
+        readonly attribute CryptoKey            publicKey;
```
vijaybh: This seems bad from an API usability perspective though. It makes the IDL more elegant but usage harder.
- Return JsonWebKey (represented in WebIDL as any for technical reasons): Caller can simply call JSON.serialize on publicKey and send it to server.
- Return CryptoKey: Caller gets CryptoKey, then must know to call WebCrypto's exportKey method to export the key for sending to the server.
Considering that the most common (perhaps only) usage of the publicKey here is to upload to the server for registration, why would we force the programmer to learn about WebCrypto for this one thing?

index.bs
```diff
@@ -281,7 +281,8 @@ The API is defined by the following Web IDL fragment.
     interface WebAuthentication {
         Promise < ScopedCredentialInfo > makeCredential (
             Account                                 accountInformation,
-            sequence < ScopedCredentialParameters > cryptoParameters,
+            sequence < CredentialType >             supportedTypes,
+            sequence < AlgorithmIdentifier >        supportedAlgorithms,
```
vijaybh: Does this now mean that the RP must be prepared to support all possible combinations of (type, algorithm) from these two lists? It did not have to do that before.

So e.g. if a new type gets defined next year, and a new algorithm the year after that, the RP would have to backport the new algorithm to old types as well?

index.bs
```diff
@@ -394,15 +390,12 @@ When this method is invoked, the user agent MUST execute the following algorithm
 
 4. Initialize |issuedRequests| to an empty list.
 
-5. Process each element of <a>cryptoParameters</a> using the following steps, to produce a new sequence `normalizedParameters`:
-    - Let |current| be the currently selected element of <a>cryptoParameters</a>.
-    - If `current.type` does not contain a {{CredentialType}} supported by this implementation, then stop processing |current|
-        and move on to the next element in <a>cryptoParameters</a>.
+5. Process each element of <a>supportedAlgorithms</a> using the following steps, to produce a new sequence `normalizedAlgorithms`:
```
vijaybh: Should there be a step where the client checks whether it supports the requested type?

index.bs
```diff
@@ -126,7 +126,7 @@ or a combination of both.
 This specification relies on several other underlying specifications.
 
 : HTML5
-:: The concept of <dfn>origin</dfn> and the <dfn>Window</dfn> interface are defined in [[!HTML5]].
+:: The concept of <dfn>origin</dfn> and the <dfn>Navigator/dfn> interface are defined in [[!HTML5]].
```
vijaybh: +1 on this change

index.bs
```diff
@@ -273,7 +273,7 @@ The API is defined by the following Web IDL fragment.
 
 <pre class="idl">
     partial interface Navigator {
-        readonly attribute WebAuthentication webauthn;
+        readonly attribute WebAuthentication authentication;
```
vijaybh: +1 for this one as well.

index.bs
jcjones: Sure, added in https://github.com/w3c/webauthn/pull/142/commits/67bf014529ff6d28a52fb32dd373115650dff301

index.bs
jcjones: You're assuming the RP would prefer a JsonWebKey, and that JWK will be the preference for the lifetime of the standard. Using the `CryptoKey` type gives the implementer flexibility to, today, choose to use DER-encoding of the SPKI field for that pubkey, and future-proofing that type.

I'm not convinced that it's much burden for an implementer writing authentication software to search the web for how to serialize a `CryptoKey`-type. To be honest, my reservations with this have to do with it returning a `Promise`, adding another `.then()` onto the chain of a `makeCredential()`. But IMO the future-proofing is worth that one.

index.bs
jcjones: My conceptual model here is that there's no relationship between `CredentialType` and `AlgorithmIdentifier`. A relying party would either support an algorithm, or not, and support a credential type, or not. I haven't built an RP. Is this not the case?

vijaybh: My experience from hanging out with RP types is that they would prefer to minimize their attack surface by carefully choosing a handful of combinations to support, while leaving older mechanisms around for compatibility (until they can be safely removed). I would not be surprised to see new algorithms to only be added to new credential types, for example. So a few years from now you'd very likely end up with something like [(originalType, RSA), (originalType, ECDSA), (newType, postQuantumAlg)].

index.bs
```diff
@@ -1408,14 +1408,12 @@ authenticator. Since all extensions are optional, this will not cause a function
 ## Extension identifiers ## {#extension-id}
 
 Extensions are identified by a string, chosen by the extension author. Extension identifiers should aim to be globally unique,
-e.g., by using reverse domain-name of the defining entity such as `com.example.webauthn.myextension`.
-
-Note: Use of dot-separated notation here does not imply an object hierarchy.
+e.g., by including the defining entity such as `myCompanyExtension`.
```
vijaybh: Thanks, I like this approach much better.

index.bs
jcjones: OK. I see your point.
@equalsJeffH suggested taking this to the list, so I will do so, and move this commit into a separate pull request.

jcjones: This is now PR #143.

index.bs
vijaybh: OK, I can live with that.

vijaybh:
Merging this now as discussed earlier since there have been no more comments.

PR #144, opened 2016-07-13
Extend the signature format to also include the RP ID

### LINKED ISSUE #107: Security: Signature format doesn't cover whole context
The basic principle we should follow is that the Authenticator should sign over the entire context that is presented to it; this just a best practice. This allows the RP to verify exactly what the authenticator saw when it created a registration or assertion intended for the RP.

Currently, the signature produced by getAssertion does not sign over the RP ID seen by the authenticator.  **We should modify the signature format so that it also covers the RP ID.**

These concerns are not purely theoretical.  For example, here is one possible consequence of not following best practices.  This is just an example, there may be other attacks that we haven't though of.

**Example**:
The TL;DR the attacker achieves in this attack is that a user with a compromised client device and a properly functioning authenticator is put in a situation where that authenticator will only work with the compromised client device (i.e., attack persistence). The external authenticator won't work with any normally functioning client (i.e, a denial of service on any un-compromised client). So the user will have an incentive to keep using the compromised client device, because their external authenticator "doesn't work with anything else".

Details: The user registers their good external authenticator on Bank.com using this compromised client device. During the registration (and in all subsequent sign-ins) the attacker changes the RPID being sent to the external authenticator from Bank.com to Changed-Bank.com. The registration happens correctly as far as the user and Bank.com are concerned. After registering, the user is able to log-in to Bank.com on this compromised client device, since the change of RPID is consistent.

However, if the user now takes that authenticator to a normally functioning client device, he/she cannot login to Bank.com. The Credential ID sent by bank.com does not match the RPID for Bank.com and so the authenticator silently refuses to sign. The user sees it as "My token doesn't work with this device"

If the authenticator also signed over the RPID that it saw, the above attack would not work. At registration time and with assertions, Bank.com would realize that the external authenticator saw some RPID other than Bank.com.

**There are likely more attack examples**

vijaybh:
Fixes #107


index.bs
```diff
@@ -388,7 +388,8 @@ When this method is invoked, the user agent MUST execute the following algorithm
 
 3. Set |callerOrigin| to the <a>origin</a> of the caller. Derive the RP ID from |callerOrigin| by computing the
     "public suffix + 1" or "PS+1" (which is also referred to as the "Effective Top-Level Domain plus One" or "<a>eTLD+1</a>")
-    part of |callerOrigin| [[PSL]]. Set |rpId| to the RP ID.
+    part of |callerOrigin| [[PSL]]. Set |rpIdHash| to the SHA-256 hash of the UTF-16 encoding of the lowercase form of this RP
```
bifurcation: UTF-8, please.  There's no UTF-16 anywhere else in the spec.

balfanz:
LGTM

index.bs
vijaybh: OK. I picked UTF-16 because that's what DOMString is, and to hopefully make i18n easier, but I don't have an issue with UTF-8 either.

jcjones:
LGTM, as well.

vijaybh:
Merging since this has received two positive and no negative reviews in a week.

PR #145, opened 2016-07-16
IDL cleanup: replace optional parameters with a single optional dictionary

vijaybh:


vijaybh:
Merging this in for now since we've discussed this before with broad consensus, the solution as described in the issue itself has received multiple +1s, and this PR has been out for a week with no objections. If there are any suggestions for further change, we can treat them separately under a different issue.

PR #147, opened 2016-07-22
Various minor technical fixes

vijaybh:
All of these are small and contained fixes, and should be non-controversial. The most impactful one is moving secure contexts from SHOULD to MUST. Please review and provide feedback. Thanks!

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -915,7 +915,7 @@ Note: A simple, undelimited concatenation is safe to use here because the |authe
 
 The authenticator MUST return both the <a>authenticatorData</a> and the raw signature back to the client. The client, in turn,
 MUST return <a>clientDataJSON</a>, <a>authenticatorData</a> and the signature to the RP. The <a>clientDataJSON</a> is returned
-in the `clientData` member of the {{WebAuthnAssertion}} and {{AttestationStatement}} structures.
+in the `clientData` member of the {{WebAuthnAssertion}} and {{WebAuthnAttestation}} structures.
 
 
 ## Credential Attestation Statements ## {#attestation}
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -1142,8 +1152,9 @@ The attestation data encodes contextual bindings made by the <a>authenticator</a
 data has a compact but extensible encoding. This is desired since authenticators can be devices with limited capabilities and
 low power requirements, with much simpler software stacks than the client platform components.
 
-The field `rawData` for this type is a byte array of 45 bytes + length of public key + length of KeyHandle + potentially more 
-extensions. The first bytes before the extensions have a fixed layout as follows: 
+The field `rawData` for this type is a byte array of 17 bytes + length of AAGUID + length of public key + length of Credential
+ID + length of clientDataHash + potentially more extensions. The first bytes before the extensions have a fixed layout as
+follows:
```
equalsJeffH: I think I (finally) figured out how we keep getting various different values for the fixed length of the rawData before the public key itself -- it is because some of us (i.e. myself) take the `17 bytes` (or the old `45 bytes`) to mean the length of rawData before the *first* variable length field, rather than the *sum of all of the fixed-length rawData fields*, which it seems is (now more clearly) stated above. 

Thus I think the above paragraph is needlessly misleading and we should update it.  Here is a proposal..
> The packed attestation `rawData` field is a variable-length byte array, consisting of various fixed-length and variable-length sub-fields, arranged as given in the below table:

equalsJeffH: Overall, other than my prior comment/suggestion, this PR looks good to me  -- thanks for reviewing and implementing those various outstanding issues.

vijaybh: Thanks, that is an excellent suggestion and it makes the text a lot less clunky. Will update the PR shortly.

jcjones:
This all looks OK to me, too. I'm really glad you caught the last Key Handle / Credential ID changes - I must have looked at that text a dozen times and never recognized it was incongruous. Thank you.

vijaybh:
Thanks! Merging, now that this has two reviews.

Issue #152, opened 2016-07-27
Problem with authn selection extension


vijaybh:
The current text says:
>If an authenticator was selected from AuthenticatorSelectionList, its AAGUID MUST be added by the client to the ClientData as the client data value for this extension.

However this creates a chicken-and-egg problem. The client may send out a request to a number of authenticators. It needs to include the AAGUID of the authenticator that finally responds in its clientData. But by the time the client knows the AAGUID, the clientData has already been sent to the authenticator and signed over.

One way (the simplest way?) to address this is to simply remove this requirement for adding anything to the clientData, since the AAGUID of the selected authenticator will be in the attestation anyways.

equalsJeffH:
It seems to me that removing the above-quoted requirement to add AAGUID to to ClientData is ok, given that claimedAAGUID is returned in the attestation Though I suggest we have Rolf sign off on it also -- he may have had a specific reason for stipulating AAGUID's inclusion in ClientData.

PR #154, opened 2016-07-28
Cleaned up exposition of processing rules

vijaybh:
This makes the processing rules consistent with the authenticator model by clarifying the text related to computing clientDataHash and passing it to the authenticator.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -832,7 +832,7 @@ string-valued keys. Values may be any type that has a valid encoding in JSON. It
 <pre class="idl">
     dictionary ClientData {
         required DOMString           challenge;
-        required DOMString           facet;
+        required DOMString           origin;
         required DOMString           rpId;
         required AlgorithmIdentifier hashAlg;
         JsonWebKey                   tokenBinding;
@@ -843,7 +843,7 @@ string-valued keys. Values may be any type that has a valid encoding in JSON. It
 <div dfn-for="ClientData">
     The <dfn>challenge</dfn> member contains the base64url encoding of the challenge provided by the RP.
 
-    The <dfn>facet</dfn> member contains the fully qualified web origin of the requester, as provided to the authenticator by
+    The <dfn>origin</dfn> member contains the fully qualified web origin of the requester, as provided to the authenticator by
     the client, in the syntax defined by [[RFC6454]].
 
     The <dfn>rpId</dfn> member contains the RP ID of the requester, as computed by the client.
@@ -1465,7 +1465,7 @@ A [RP] shall verify the clientData contextual bindings (see step 4 in [[#verifyi
 - Check that `AndroidAttestationClientData.challenge` equals the attestationChallenge that was passed into the
     {{makeCredential()}} call.
 
-- Check that the `facet` and `tokenBinding` parameters in the `AndroidAttestationClientData` match the [RP] App.
+- Check that the `origin` and `tokenBinding` parameters in the `AndroidAttestationClientData` match the [RP] App.
 
 - Check that `AndroidAttestationClientData.publicKey` is the same key as the one returned in the `ScopedCredentialInfo` by the
     `makeCredential` call.
--- a/index.bs
+++ b/index.bs
@@ -408,9 +408,9 @@ When this method is invoked, the user agent MUST execute the following algorithm
 6. If {{CredentialOptions/extensions}} was specified, process any extensions supported by this client platform, to produce the
     extension data that needs to be sent to the authenticator. Call this data |clientExtensions|.
 
-7. Use {{attestationChallenge}}, |callerOrigin| and |rpId|, along with the token binding key associated with |callerOrigin|, to
-    create a {{ClientData}} structure representing this request. Choose a hash algorithm for {{ClientData/hashAlg}} and compute
-    the <a>clientDataJSON</a> and <a>clientDataHash</a>.
+7. Use {{attestationChallenge}}, |callerOrigin| and |rpId|, along with the token binding key associated with |callerOrigin| (if
+    any), to create a {{ClientData}} structure representing this request. Choose a hash algorithm for {{ClientData/hashAlg}} and
+    compute the <a>clientDataJSON</a> and <a>clientDataHash</a>.
     
 8. Initialize |issuedRequests| to an empty list.
 
@@ -472,18 +472,19 @@ When this method is invoked, the user agent MUST execute the following algorithm
 4. If {{AssertionOptions/extensions}} was specified, process any extensions supported by this client platform, to produce the
     extension data that needs to be sent to the authenticator. Call this data |clientExtensions|.
 
-5. Use {{assertionChallenge}}, |callerOrigin| and |rpId|, along with the token binding key associated with |callerOrigin|, to
-    create a {{ClientData}} structure representing this request. Choose a hash algorithm for {{ClientData/hashAlg}} and compute
-    the <a>clientDataJSON</a> and <a>clientDataHash</a>.
+5. Use {{assertionChallenge}}, |callerOrigin| and |rpId|, along with the token binding key associated with |callerOrigin| (if
+    any), to create a {{ClientData}} structure representing this request. Choose a hash algorithm for {{ClientData/hashAlg}} and
+    compute the <a>clientDataJSON</a> and <a>clientDataHash</a>.
     
 6. Initialize |issuedRequests| to an empty list.
 
 7. For each authenticator currently available on this platform, perform the following steps:
     - If <a>allowList</a> is undefined or empty, let |credentialList| be an empty list. Otherwise, execute a platform-specific
-        procedure to determine which of these credentials might be present on this authenticator, and set |credentialList| to
-        this filtered list. If no such filtering is possible, set |credentialList| to an empty list.
+        procedure to determine which, if any, credentials listed in <a>allowList</a> might be present on this authenticator, and
+        set |credentialList| to this filtered list. If no such filtering is possible, set |credentialList| to an empty list.
     - If the above filtering process concludes that none of the credentials on <a>allowList</a> can possibly be on this
-        authenticator, ignore this authenticator and do not perform any of the following per-authenticator steps.
+        authenticator, do not perform any of the following steps for this authenticator, and proceed to the next authenticator
+        (if any).
     - Asynchronously invoke the <a>authenticatorGetAssertion</a> operation on this authenticator with |rpIdHash|,
         <a>clientDataHash</a>, |credentialList|, and |clientExtensions| as parameters.
     - Add an entry to |issuedRequests|, corresponding to this request.
@@ -918,7 +919,7 @@ credential does not change between operations but instead remains the same for t
 validated by the authenticator during the <a>authenticatorGetAssertion</a> operation, by making sure that the RP ID hash
 associated with the requested credential exactly matches the RP ID hash supplied by the client. These differences also explain
 why the RP ID hash is always a SHA-256 hash instead of being crypto-agile like the <a>clientDataHash</a>; for a given RP ID, we
-need the hash to be computed the same way by all clients for all operations so that authenticators can move between clients
+need the hash to be computed the same way by all clients for all operations so that authenticators can roam among clients
 without losing interoperability.
 
 The `TUP` flag SHALL be set if and only if the authenticator detected a user through an authenticator specific gesture. The
@@ -1048,7 +1049,7 @@ Upon receiving an attestation statement, the [RP] shall:
     - Verify `rawData` syntax and that it doesn't contradict the signing algorithm specified in `alg`.
     - The DAA root key is dedicated to a single Authenticator model.
 
-4. Verify the contextual bindings (e.g., channel binding) from the `clientData` (see [[#authenticator-signature]]).
+4. Verify the contextual bindings (e.g., token binding) from the `clientData` (see [[#authenticator-signature]]).
 
 5. Verify that user verification method and other authenticator characteristics related to this authenticator model, match the
     [RP] policy. The FIDO Metadata Service [[FIDOMetadataService]] provides an easy way to access the authenticator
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -127,7 +127,7 @@ or a combination of both.
 This specification relies on several other underlying specifications.
 
 : HTML5
-:: The concept of <dfn>origin</dfn> and the <dfn>Navigator/dfn> interface are defined in [[!HTML5]].
+:: The concept of <dfn for='web'>origin</dfn> and the <dfn>Navigator/dfn> interface are defined in [[!HTML5]].
```
equalsJeffH: this definition is not strictly true. unfortunately whatwg (and thus W3C) have plowed ahead on re-defining origin without coordinating with IETF wrt RFC6464.  E.g., see https://lists.w3.org/Archives/Public/ietf-http-wg/2016JulSep/0225.html

equalsJeffH:
this  has quite a few detail-level changes across a somewhat wide swath of the draft spec, and unfortunately I will not have time to review this in detail for a week or two, and would like to do so prior to its being merged.  thanks.

index.bs
vijaybh: OK. Note I did not change this text at all, only the markup - this was required to disambiguate "origin" used in this context from the origin field in ClientData for Bikeshed.

If we need a better definition for origin, can we handle that in a separate issue?

jcjones: Opened #160

equalsJeffH: thanks for opening #160

index.bs
```diff
@@ -381,22 +381,19 @@ This method takes the following parameters:
...
 2. Let |promise| be a new <a data-lt="Promises">Promise</a>. Return |promise| and start a timer for |adjustedTimeout| seconds.
     Then asynchronously continue executing the following steps.
 
-3. Set |callerOrigin| to the <a>origin</a> of the caller. Derive the RP ID from |callerOrigin| by computing the
+3. Set |callerOrigin| to the <a link-for='web'>origin</a> of the caller. Derive the RP ID from |callerOrigin| by computing the
     "public suffix + 1" or "PS+1" (which is also referred to as the "Effective Top-Level Domain plus One" or "<a>eTLD+1</a>")
-    part of |callerOrigin| [[PSL]]. Set |rpIdHash| to the SHA-256 hash of the UTF-8 encoding of the lowercase form of this RP
-    ID.
+    part of |callerOrigin| [[PSL]]. Let |rpId| be the lowercase form of this RP ID. Set |rpIdHash| to the SHA-256 hash of the
+    UTF-8 encoding of |rpId|.
```
equalsJeffH: see #167

index.bs
```diff
@@ -406,17 +403,22 @@ When this method is invoked, the user agent MUST execute the following algorithm
...
 
-6. If <a>excludeList</a> is undefined, set it to the empty list.
+5. If <a>excludeList</a> is undefined, set it to the empty list.
+
+6. If {{CredentialOptions/extensions}} was specified, process any extensions supported by this client platform, to produce the
+    extension data that needs to be sent to the authenticator. Call this data |clientExtensions|.
 
-7. If <a for="CredentialOptions">extensions</a> was specified, process any extensions supported by this client platform, to
-    produce the extension data that needs to be sent to the authenticator. Call this data |clientExtensions|.
+7. Use {{attestationChallenge}}, |callerOrigin| and |rpId|, along with the token binding key associated with |callerOrigin|, to
```
equalsJeffH: suggest: `along with the token binding key associated with |callerOrigin| (if any),`    see also #168

index.bs
```diff
@@ -455,35 +457,38 @@ This method takes the following parameters:
...
-    - If |credentialList| is empty, ignore this authenticator and do not perform any of the following per-authenticator steps.
-    - Asynchronously invoke the <a>authenticatorGetAssertion</a> operation on this authenticator with |callerOrigin|,
-        |rpIdHash|, <a>assertionChallenge</a>, |credentialList|, and |clientExtensions| as parameters.
+    part of |callerOrigin| [[PSL]]. Let |rpId| be the lowercase form of this RP ID. Set |rpIdHash| to the SHA-256 hash of the
+    UTF-8 encoding of |rpId|.
+
+4. If {{AssertionOptions/extensions}} was specified, process any extensions supported by this client platform, to produce the
+    extension data that needs to be sent to the authenticator. Call this data |clientExtensions|.
+
+5. Use {{assertionChallenge}}, |callerOrigin| and |rpId|, along with the token binding key associated with |callerOrigin|, to
```
equalsJeffH: suggest: `along with the token binding key associated with |callerOrigin| (if any),`    see also #168

index.bs
```diff
@@ -455,35 +457,38 @@ This method takes the following parameters:
...
+    create a {{ClientData}} structure representing this request. Choose a hash algorithm for {{ClientData/hashAlg}} and compute
+    the <a>clientDataJSON</a> and <a>clientDataHash</a>.
+    
+6. Initialize |issuedRequests| to an empty list.
+
+7. For each authenticator currently available on this platform, perform the following steps:
+    - If <a>allowList</a> is undefined or empty, let |credentialList| be an empty list. Otherwise, execute a platform-specific
+        procedure to determine which of these credentials might be present on this authenticator, and set |credentialList| to
+        this filtered list. If no such filtering is possible, set |credentialList| to an empty list.
```
equalsJeffH: suggest:  Otherwise, execute a platform-specific procedure to determine which, if any, credentials contained in the <a>allowList</a> are present on this authenticator, and set |credentialList| to this filtered list. If no such filtering is possible, set |credentialList| to an empty list.

index.bs
```diff
@@ -455,35 +457,38 @@ This method takes the following parameters:
...
+    - If the above filtering process concludes that none of the credentials on <a>allowList</a> can possibly be on this
+        authenticator, ignore this authenticator and do not perform any of the following per-authenticator steps.
```
equalsJeffH: suggest for just this sentence: If the above filtering process concludes that none of the credentials on <a>allowList</a> are on this authenticator, ignore this authenticator and go on to the next authenticator, if any.

index.bs
```diff
@@ -653,6 +658,51 @@ well as to decode and validate the bindings of both the client and authenticator
...
+
+    The <dfn>rpId</dfn> member contains the RP ID of the requester, as computed by the client.
+
+    The <dfn>hashAlg</dfn> member specifies the hash algorithm used to compute <a>clientDataHash</a> (see
+    [[#authenticator-signature]]). Use "S256" for SHA-256, "S384" for SHA384, "S512" for SHA512, and "SM3" for SM3 (see
+    [[#iana-considerations]]). This algorithm is chosen by the client at its sole discretion.
+
+    The <dfn>tokenBinding</dfn> member contains a JsonWebKey object as defined by [[WebCryptoAPI#JsonWebKey-dictionary]]
+    describing the public key that this client uses for the Token Binding protocol when communicating with the [RP]. This can be
+    omitted if no Token Binding has been negotiated between the client and the [RP].
```
equalsJeffH: see also #168

index.bs
```diff
@@ -907,9 +917,9 @@ generated. However, it differs from other client data in some important ways. Fi
 credential does not change between operations but instead remains the same for the lifetime of that credential. Secondly, it is
 validated by the authenticator during the <a>authenticatorGetAssertion</a> operation, by making sure that the RP ID hash
 associated with the requested credential exactly matches the RP ID hash supplied by the client. These differences also explain
-why the RP ID hash is always a SHA-256 hash instead of being crypto-agile like the |clientDataHash|; for a given RP ID, we need
-the hash to be computed the same way by all clients for all operations so that authenticators can move between clients without
-losing interoperability.
+why the RP ID hash is always a SHA-256 hash instead of being crypto-agile like the <a>clientDataHash</a>; for a given RP ID, we
+need the hash to be computed the same way by all clients for all operations so that authenticators can move between clients
```
equalsJeffH: s/move between/roam among/

index.bs
```diff
@@ -1052,7 +1048,7 @@ Upon receiving an attestation statement, the [RP] shall:
     - Verify `rawData` syntax and that it doesn't contradict the signing algorithm specified in `alg`.
     - The DAA root key is dedicated to a single Authenticator model.
 
-4. Verify the contextual bindings (e.g., channel binding) from the clientData (see [[#authenticator-signature]]).
+4. Verify the contextual bindings (e.g., channel binding) from the `clientData` (see [[#authenticator-signature]]).
```
equalsJeffH: s/channel/token/
see also #168

index.bs
```diff
@@ -1453,24 +1449,25 @@ For this attestation format, the ClientData dictionary is extended in the follow
         this key is authorized to be used after the user is successfully authenticated.
 </div>
 
-In order to generate an attestation statement, the client MUST create clientDataJSON by UTF8-encoding a structure of type
-AndroidAttestationClientData, and compute clientDataHash as the hash of clientDataJSON. It must then provide clientDataHash as
-the Nonce value when requesting the SafetyNet attestation.
+In order to generate an attestation statement, the client MUST create <a>clientDataJSON</a> by UTF8-encoding a structure of type
+AndroidAttestationClientData, and compute <a>clientDataHash</a> as the hash of <a>clientDataJSON</a>. It must then provide
```
equalsJeffH: s/AndroidAttestationClientData/`AndroidAttestationClientData`/ ?

index.bs
```diff
@@ -1741,8 +1738,7 @@ credential. It is intended primarily for [RPS] that wish to tightly control the
 :: This extension can only be used during {{makeCredential()}}. If the client supports the Authenticator Selection Extension, it
     MUST use the first available authenticator whose AAGUID is present in the <dfn>AuthenticatorSelectionList</dfn>. If none of
     the available authenticators match a provided AAGUID, the client MUST select an authenticator from among the available
-    authenticators to generate the credential. If an authenticator was selected from {{AuthenticatorSelectionList}}, its
-    AAGUID MUST be added by the client to the ClientData as the client data value for this extension.
+    authenticators to generate the credential.
```
equalsJeffH: why is L1745 ` AAGUID MUST be added by the client to the ClientData as the client data value for this extension.` deleted?

equalsJeffH:
overall this PR looks ok, modulo specific line notes above, thanks.

index.bs
vijaybh: See #152

index.bs
vijaybh: Thanks, reworded this slightly differently but tried to keep the spirit of your suggestion.

equalsJeffH:
thanks for addressing comments, looks good to me (while realizing that there's open issues that read on the resultant spec)

jcjones:
This LGTM as of 7 days ago.

index.bs
```diff
@@ -406,17 +403,22 @@ When this method is invoked, the user agent MUST execute the following algorithm
...
+    compute the <a>clientDataJSON</a> and <a>clientDataHash</a>.
+    
+8. Initialize |issuedRequests| to an empty list.
 
-8. For each authenticator currently available on this platform: asynchronously invoke the
-    <a>authenticatorMakeCredential</a> operation on that authenticator with |callerOrigin|, |rpIdHash|,
-    <a>accountInformation</a>, `normalizedParameters`, <a>excludeList</a>, <a>attestationChallenge</a> and |clientExtensions| as
-    parameters. Add a corresponding entry to |issuedRequests|.
+9. For each authenticator currently available on this platform: asynchronously invoke the <a>authenticatorMakeCredential</a>
+    operation on that authenticator with |rpIdHash|, <a>clientDataHash</a>, <a>accountInformation</a>, `normalizedParameters`,
```
leshi: currently, `authenticatorMakeCredential()` in CTAP accepts unhashed rpId: https://github.com/fido-alliance/fido-2-specs/blob/master/fido-client-to-authenticator-protocol.html#L204

Adding this quickly for us to ponder.  Will comment with more insightful thoughts a bit later.

index.bs
equalsJeffH: I think the proposed resolution in #152 is OK though Rolf should also weigh-in on it.

index.bs
```diff
@@ -653,6 +659,51 @@ well as to decode and validate the bindings of both the client and authenticator
...
+### Client data used in WebAuthn signatures (dictionary <dfn dictionary>ClientData</dfn>) ### {#sec-client-data}
+
+The client data represents the contextual bindings of both the [RP] and the client platform. It is a key-value mapping with
+string-valued keys. Values may be any type that has a valid encoding in JSON. Its structure is defined by the following Web IDL.
+
+<pre class="idl">
+    dictionary ClientData {
+        required DOMString           challenge;
+        required DOMString           origin;
+        required DOMString           rpId;
```
leshi: I think the rpId is also signed over outside of the clientData (see bytes 0-31 in the Authenticator Data part of the signature format).  I think we should consider removing rpId from the client data -- what do you think?

index.bs
```diff
@@ -455,35 +457,39 @@ This method takes the following parameters:
...
+6. Initialize |issuedRequests| to an empty list.
+
+7. For each authenticator currently available on this platform, perform the following steps:
+    - If <a>allowList</a> is undefined or empty, let |credentialList| be an empty list. Otherwise, execute a platform-specific
+        procedure to determine which, if any, credentials listed in <a>allowList</a> might be present on this authenticator, and
+        set |credentialList| to this filtered list. If no such filtering is possible, set |credentialList| to an empty list.
+    - If the above filtering process concludes that none of the credentials on <a>allowList</a> can possibly be on this
+        authenticator, do not perform any of the following steps for this authenticator, and proceed to the next authenticator
+        (if any).
+    - Asynchronously invoke the <a>authenticatorGetAssertion</a> operation on this authenticator with |rpIdHash|,
```
leshi: same comment for rpIdHash vs rpId as above

index.bs
```diff
@@ -714,15 +765,13 @@ The following operations can be invoked by the client in an authenticator sessio
 This operation must be invoked in an authenticator session which has no other operations in progress. It takes the following
 input parameters:
 
-- The web origin of the script on whose behalf the operation is being initiated, as determined by the user agent and the client.
-- The SHA-256 hash of the RP ID corresponding to the above web origin, as determined by the user agent and the client.
+- The SHA-256 hash of the caller's RP ID, as determined by the user agent and the client.
```
leshi: Aha, here is where we change from `origin` to `rpId hash`

leshi:
Sorry for being late on the review :(.  I think there may be a couple things to discuss wrt rpID:

1. Should `authenticatorMakeCredential` and `authenticatorGetAssertion` as defined in this specification send the hashed or unhashed rpId?
2. Should `rpId` be removed from the clientData give that it's explicitly part of the signature format now?

I think maybe we should create issues for these two things and proceed with merging this PR since it does improve general consistency and clarity 👍

equalsJeffH:
@leshi makes good suggestions.  I submitted #188 and #189.  I am OK with merging this PR since we have captured the latter issues, along with others, as noted above.

PR #157, opened 2016-08-04
Update index.bs with proposed location extension

gmandyam:
As discussed in the Webauthn weekly telco on Aug-03-16.

Not that familiar with bikeshed.  Hope this works.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1839,14 +1839,14 @@ credential. It is intended primarily for [RPS] that wish to tightly control the 
 
 : Authenticator processing
 :: If the <a>authenticator</a> does not support the extension, then the authenticator MUST ignore the extension request.  
-If the <a>authenticator</a> accepts the extension, then the authenticator SHOULD only add this extension data to a packed 
-attestation or assertion.
+    If the <a>authenticator</a> accepts the extension, then the authenticator SHOULD only add this extension data to a packed 
+    attestation or assertion.
 
 : Authenticator data
 :: If the <a>authenticator</a> accepts the extension request, then authenticator data SHOULD provide location data in the form of a 
-CBOR-encoded map, with the first value being the extension identifier and the second being an array of returned values.  The array 
-elements SHOULD be derived from (key,value) pairings for each location attribute that the <a>authenticator</a> supports. The following is 
-an example of authenticator data where the returned array is comprised of a {longitude, latitude, altitude} triplet.
+    CBOR-encoded map, with the first value being the extension identifier and the second being an array of returned values.  The array 
+    elements SHOULD be derived from (key,value) pairings for each location attribute that the <a>authenticator</a> supports. The following is 
+    an example of authenticator data where the returned array is comprised of a {longitude, latitude, altitude} triplet.
 
     <pre class="highlight">
         F1 D0                                       -- This is a WebAuthn packed rawData object
--- a/index.bs
+++ b/index.bs
@@ -1846,14 +1846,16 @@ credential. It is intended primarily for [RPS] that wish to tightly control the 
 :: If the <a>authenticator</a> accepts the extension request, then authenticator data SHOULD provide location data in the form of a 
     CBOR-encoded map, with the first value being the extension identifier and the second being an array of returned values.  The array 
     elements SHOULD be derived from (key,value) pairings for each location attribute that the <a>authenticator</a> supports. The following is 
-    an example of authenticator data where the returned array is comprised of a {longitude, latitude, altitude} triplet.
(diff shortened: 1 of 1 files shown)


gmandyam:
Closing PR - travis build failed

jcjones:
@gmandyam: You don't need to close the PR; you can just update your branch and the PR will move forward and retry. Feel free to re-open!

gmandyam:
@jcjones 
Thanks, will do.

index.bs
```diff
@@ -1823,7 +1823,51 @@ credential. It is intended primarily for [RPS] that wish to tightly control the
...
+    CBOR-encoded map, with the first value being the extension identifier and the second being an array of returned values.  The array 
+    elements SHOULD be derived from (key,value) pairings for each location attribute that the <a>authenticator</a> supports. The following is 
+    an example of authenticator data where the returned array is comprised of a {longitude, latitude, altitude} triplet.
+
+    <pre class="highlight">
+        F1 D0                                       -- This is a WebAuthn packed rawData object
+        81                                          -- TUP and ED set
+        00 00 00 01                                 -- (initial) signature counter
+        ...                                         -- all public key alg etc.
+        A2                                          -- extension: CBOR map of two elements
```
jcjones: I believe this should be a CBOR map of 1 element, as there's one key/value pair.

index.bs
```diff
@@ -1823,7 +1823,51 @@ credential. It is intended primarily for [RPS] that wish to tightly control the
...
+                68				    -- Element 1:  CBOR text string of 8 bytes
+                   6C 61 74 69 74 75 64 65          -- “latitude” UTF-8 string
+                FB ...				    -- Element 2:  Latitude as CBOR encoded double-precision float
+                69				    -- Element 3:  CBOR text string of 9 bytes
+                   6C 6F 6E 67 69 74 75 64 65       -- “longitude” UTF-8 string
+                FB ...				    -- Element 4:  Longitude as CBOR encoded double-precision float
+                68				    -- Element 5:  CBOR text string of 8 bytes
+                  61 6C 74 69 74 75 64 65           -- “altitude” UTF-8 string
+                FB ...				    -- Element 6:  Altitude as CBOR encoded double-precision float
+    </pre>
```
jcjones: Do we need a horizontal (and vertical?) position error?

index.bs
```diff
@@ -1823,7 +1823,51 @@ credential. It is intended primarily for [RPS] that wish to tightly control the
...
+: Authenticator processing
+:: If the <a>authenticator</a> does not support the extension, then the authenticator MUST ignore the extension request.  
+    If the <a>authenticator</a> accepts the extension, then the authenticator SHOULD only add this extension data to a packed 
+    attestation or assertion.
+
+: Authenticator data
+:: If the <a>authenticator</a> accepts the extension request, then authenticator data SHOULD provide location data in the form of a 
+    CBOR-encoded map, with the first value being the extension identifier and the second being an array of returned values.  The array 
+    elements SHOULD be derived from (key,value) pairings for each location attribute that the <a>authenticator</a> supports. The following is 
+    an example of authenticator data where the returned array is comprised of a {longitude, latitude, altitude} triplet.
```
jcjones: We need to define whether altitude means "meters above mean sea level", "feet above MSL", "meters above WGS84 ellipsoid", "feet above WGS84 ellipsoid" or something else.

jcjones: It's unclear whether the `lat` and `lon` are decimal degrees, or - like NMEA strings - a numerical encoding of `dddmm.ss`

gmandyam: Good point.  Was deriving from https://dev.w3.org/geo/api/spec-source.html#coordinates_interface - one possibility is to add a reference to this section of the Geoloc API to clarify.

index.bs
gmandyam: Can be added, but it seems that there is considerable variability in how platforms support.  Android provides getAccuracy() method (https://developer.android.com/reference/android/location/Location.html#getAccuracy%28%29) which gives accuracy at 68% confidence level.  Windows provides a similar value (https://msdn.microsoft.com/en-us/library/windows/desktop/windows.devices.geolocation.geocoordinate.accuracy.aspx), but it is not clear what confidence level corresponds to this accuracy reading.

index.bs
gmandyam: Yes.  Will correct.

gmandyam:
@jcjones 
Modified PR to address some of your concerns as best as I could.  Have not added pos.'n accuracy for the reason I stated previously.

jcjones:
Thanks, LGTM. r+

PR #158, opened 2016-08-04
Eliminate Credential object

vijaybh:
Sending this out as PR in the hopes it will get more review this way. This is the experiment to remove the Credential type that I sent out a couple weeks ago for comments.

Note that I have removed the two "future possibility" pieces - basically, we can suport versioning in the future either by adding a WebAuthn version to the Algorithm object, or by adding an extension to support version selection.


index.bs
```diff
@@ -2035,18 +1953,12 @@ extension for transaction authorization.
...
-        id: "ISEhISEhIWhpIHRoZXJlISEhISEhIQo="
-    };
-    var acceptableCredential2 = {
-        type: "ScopedCred",
-        id: "cm9zZXMgYXJlIHJlZCwgdmlvbGV0cyBhcmUgYmx1ZQo="
-    };
-    
+    var timeoutSeconds = 300;  // 5 minutes
+    var cred1 = base64Decode("ISEhISEhIWhpIHRoZXJlISEhISEhIQo="); // Returns ArrayBuffer
+    var cred2 = base64Decode("cm9zZXMgYXJlIHJlZCwgdmlvbGV0cyBhcmUgYmx1ZQo=");
```
jcjones: Nit: So much opportunity here for easter eggs, and yet not.

PR #159, opened 2016-08-04
IDL cleanup: don't require callers to create Credential objects

vijaybh:
This is the minimal version of the change required to fix the IDL issue
that @jcjones brought up regarding Credential objects. Unlike PR #158,
it retains the credential type and leaves the actual API surface and
samples unchanged.

If we pick this approach, PR #158 can be closed out and will no longer
be necessary.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -314,9 +314,9 @@ The API is defined by the following Web IDL fragment.
     };
 
     dictionary CredentialOptions {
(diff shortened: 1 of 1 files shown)


jcjones:
LGTM.

equalsJeffH:
this looks ok.

PR #161, opened 2016-08-10
Make attestation more modular

vijaybh:


rlin1:
Does this essentially making packed attestation binary format similar to the signature format?

gmandyam:
General comment on USVString format.  I think we can provide rules in the specification itself for vendor-defined attestation formats much like CSS did (https://www.w3.org/TR/CSS21/syndata.html#vendor-keywords).  For instance, a proprietary CBOR attestation could be unicode(vendor_prefix + "_cbor").

Related question:  why unicode?

rlin1:
Some comments:

a) section 5.2.1 AuthenticatorData: This is defined as data controlled by the authenticator.   This is not always the case.
a.1) When using a KeyStore based authenticator (e.g. keymaster impl in TEE providing generic keyStore interface to the key), then this key can be used by the calling App to sign anything (key is not restricted to sign valid WebAuthn signature assertions only).
a.2) When using an Android "N" KeyStore based authenticator, the "N" HW attestation allows the calling App to include an arbitrary Nonce (or challenge) value. In the current draft, this structure is also called authenticatorData (same structure as in 5.2.1).  However, in this case this authenticatorData is also not fully controlled by the Authenticator (i.e. keymaster), but by the calling App.
As a consequence, I propose renaming authenticatorData to level2Data.  Level2Data might be fully controlled by the Authenticator or by the App, the RP has to check Metaddata and/or the attestation certificate in order to understand whether it is controlled by the authenticator (or the App).

b) the field "statement" in WebAuthnAttestation might be controlled by a dedicated secure element being part of an authenticator or by the authenticator in general.  In order to make it more consistent, I propose to rename this field to level1Data.

See branch rolf-modular-attestation-changes

rlin1:
It is unclear whether extensions could be added to the authenticatorData or to the statement (or both)

rlin1:
The term attestation statement is currently used for two different things:
a) WebAuthnAttestation structure (section 4.9) and
b) the field statement inside the WebAuthnAttestation structure.

This is (IMHO) another reson to rename field statement to level1Data (see above)

rlin1:
Regarding Step 6 in Verifying an Attestation Statement:
If the authenticatorData in the attestation statement is controlled by the calling App, it doesn't make sense (from a security perspective) to verify the rpId hash include in authenticatorData to the one in clientData.
In this case, the rpId would have to be included in the statement (i.e. level1Data).

rlin1:
The name "Credential Attestation Statement" of section 4.9 is a bit surprising.
We are talking about attestation statements in general and the WebAuthnAttestation interface in particular.  Could we rename section 4.9 to either Attestation Statement or Web Authentication Attestation Interface (in order to avoid yet another term for it)?

rlin1:
New naming proposal (Vijay): 
level1Data --> attestedData
level2Data --> authenticatorData

rlin1:
We still have to specify whether extensions go into attestedData or authenticatorData or both.
Either way: it has to be specified in the doc.

vijaybh:
Please see updated PR.

@gmandyam - Re: why unicode - did you mean why not unicode? 6.1 says format identifiers must be printable ASCII.

@rlin1:
- I did some more renaming to resolve the overloading of "attestation statement"
- Re: step 6 of the verification process - to me this is checking up on the client. The authenticatorData is created by the authenticator (even if it is the "rest of the authenticator") based on what the client tells it, and it would be good to verify that the client is not playing games by telling us one thing and the authenticator another.
- Extensions are in authenticatorData now. Whenever there is an opportunity to verify such an extension (or anything else really) against the contents of the signature, this should be done. This way authenticators and formats can vary in what they put into signature, and that only changes the trust level of the relevant items in authenticatorData.

PR #162, opened 2016-08-10
Allow caller to pick between strict and eTLD+1 matching

vijaybh:
Per suggestion from @bifurcation on this morning's WG call.


jcjones:
This LGTM; it does require a rebase/merge.

balfanz:
How about we do something akin to domain lowering:

```
dictionary CredentialOptions {
  unsigned long                       timeoutSeconds;
  DOMString                           rpId;
  sequence < CredentialDescription >  excludeList;
  WebAuthnExtensions                  extensions;
};
```

rpId would be optional, and if left unspecified, the caller's origin shall be used. If it *is* specified, then valid values are postfixes of the caller's origin, up to - but no further than - eTLD+1.

We'd have to define what happens to the scheme and port of the caller's origin, but I assume domain lowering already deals with this in some way (i.e., both http://foo.example.com:1234 and https://foo.example.com:9876 are allowed to lower to "example.com").

vijaybh:
I do think @balfanz proposes a more elegant solution than the Boolean. Let me write it up in a new PR and we can pick one of the two.

equalsJeffH:
Ok, I can live with either this "boolean switch" approach, or Dirk's as-yet-not-in-a-PR-it-seems approach. There are detail-level issues in this "boolean switch" approach, tho I'll wait to see which approach we go with before working on detail polishing. thanks for writing this up.

index.bs
```diff
@@ -343,7 +344,8 @@ When this method is invoked, the user agent MUST execute the following algorithm
 2. Let |promise| be a new <a data-lt="Promises">Promise</a>. Return |promise| and start a timer for |adjustedTimeout| seconds.
     Then asynchronously continue executing the following steps.
 
-3. Set |callerOrigin| to the <a link-for='web'>origin</a> of the caller. Derive the RP ID from |callerOrigin| by computing the
+3. Set |callerOrigin| to the <a link-for='web'>origin</a> of the caller. If {{CredentialOptions/rpStrict}} was set to true or
+    not specified, then set the RP ID to |callerOrigin|. Otherwise, erive the RP ID from |callerOrigin| by computing the
```
yaronf: typo: erive

yaronf:
For either of these solutions, please note that RFC 7719 basically says that "eTLD+1" is not well defined. Quoting: "Note that the term "public suffix" is controversial in the DNS community for many reasons, and may be significantly changed in the future." Can we clarify that the Mozilla-maintained list of public suffixes is authoritative on this term?

equalsJeffH:
Let's go with #198 rather than this one.

vijaybh:
Closing this out since the consensus on the previous call was similar, that this is inferior to #198.

PR #164, opened 2016-08-15
Token Binding ID: JWK -> opaque

### LINKED ISSUE #163: tokenBinding in ClientData should be a binary type (perhaps base64) and not JsonWebKey
The current Token Binding drafts at https://datatracker.ietf.org/doc/draft-ietf-tokbind-protocol/ define a binary token binding ID - currently this is just a serialization of the public key but the idea is to create an abstraction where it might include other things in future. My understanding is that users of this protocol would be encouraged to treat the TB ID as an opaque binary blob and do bitwise comparisons.

Accordingly, it seems better to change the type of tokenBinding from JsonWebKey to a binary type. Since the ClientData is serialized and hashed into clientDataHash, it seems reasonable to do base64 rather than an array type, just to make for a less messy JSON serialization.

@balfanz is our resident Token Binding expert - any thoughts on this?

vijaybh:
Fixes #163


jcjones:
LGTM.

equalsJeffH:
am nominally ok with this, however see also: #168

vijaybh:
Merging after email signoff from @balfanz

equalsJeffH:
is it just me or @balfanz 's email didn't seem to be cc'd to public-webauthn@ ?

PR #169, opened 2016-08-17
Adding the User Verification Mode extension as per discussions in the…

rdghosh:
… WG weekly metings. After a lot of thought I retained the sub-array instead of converting the sub-items in the extension to a map. This was primarily to avoid having to define string based keys that increases the size of the extension. Given that the array is well defined I think it keeps it easier to process in software and extendable for future additions as well.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -127,7 +127,7 @@ or a combination of both.
 This specification relies on several other underlying specifications.
 
 : HTML5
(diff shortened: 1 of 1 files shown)


equalsJeffH:
thx for the PR, tho it would be helpful if you would add your actual name to your github user profile....

Issue #113, opened 2016-06-01
clarify content of algorithm member of ScopedCredentialParameters


equalsJeffH:
in https://w3c.github.io/webauthn/#credential-params, the statement..
```
  The algorithm member specifies the cryptographic algorithm with 
  which the newly generated credential will be used.
```
ought to be something more akin to..
```
  The algorithm member specifies the cryptographic key generation 
  algorithm parameters with which the credential key pair MUST be 
  generated.
```
..and further refinement beyond that may be needed: 

I.e., a question is just how much of gnarly key gneration params ought a RP webapp be supplying, and how much does the underlying client platform supply?
See https://www.w3.org/TR/WebCryptoAPI/#examples-section for an example of the full unabridged key gen params used to generate an RSA key pair.  Perhaps the RP webapp **could** pass in only as much as it cares/needs to specify..
```
  { "name": "RSASSA-PKCS1-v1_5" }
```
..on the makeCredential() method, and the underlying platform takes care of the other details (i.e., modulusLength, publicExponent, and hash), or it could pass in..
```
  { "name": "RSASSA-PKCS1-v1_5",
     "hash":"SHA-256" }
```
..or the complete algorithmIdentifier as shown in the webcrypto spec example..
```
var algorithmKeyGen = {
  name: "RSASSA-PKCS1-v1_5",
  // RsaHashedKeyGenParams
  modulusLength: 2048,
  publicExponent: new Uint8Array([0x01, 0x00, 0x01]),  // Equivalent to 65537
  hash: {
    name: "SHA-256"
  }
};
```

apowers313:
The generation of keys based on the requested crypto parameters is currently specified as a "best effort" activity. From [Section 4.1.1 makeCredential](http://w3c.github.io/webauthn/#makeCredential):

> The <b>cryptoParameters</b> parameter supplies information about the desired properties of the credential to be created. The sequence is ordered from most preferred to least preferred. The platform makes a best effort to create the most preferred credential that it can.

equalsJeffH:
I believe you are misunderstanding the meaning of the prose you quote above.  that prose in {#makeCredential} section is regarding the *sequence* of cryptoParameters passed into the makeCredential() promise.  It is regarding (a) how the RP denotes the priority of the cryptoparameters it desires, and (b) how the platform selects which entry of the  cryptoParameters sequence it honors.  This is orthogonal to the theme of the original issue above.

apowers313:
[...]

Issue #100, opened 2016-05-13
Section 4.3: add reference to privacy


rlin1:
We already mention that in section 4.3.4.1. but it is important to already be mentioned in the into

equalsJeffH:
at the time @rlin1 submitted this issue, S 4.3.4.1 mapped to this spot in the spec source (index.src.html): `{#sec-attestation-privacy}`.   we need to note such spec source anchors in issues because section #s will change over time as we move sections around.  

What i interpret this issue to be about is:   there are various privacy considerations specific to attestation, and the various manners in which it (attestation) may be manifested. Thus there should be some, perhaps modest, notation of such in the overall spec's introduction since we mention attestation there, and even if the detailed specification for  the values of WebAuthnAttestation.statement is moved to a different spec (see #108), we will still have linkage to said attestation spec  and the notion that WebAuthn Scoped Credentials are "attested".

equalsJeffH:
hm. given the level of abstraction the spec's Introduction is written at, it seems to me that a forward link to `{#sec-attestation-privacy}` is most appropriate in the section "Credential Attestation Statements ## {#attestation}"

PR #185, opened 2016-08-30
editorial cleanups

### LINKED ISSUE #47: The FIDO Metadata Service needs a non-FIDO analogous mechanism

equalsJeffH:
These are editorial cleanups I took care of while going thru the issues list. it fixes #47, #100, #113.


index.bs
```diff
@@ -558,8 +564,8 @@ authorizing an authenticator with which to complete the operation.
 
     The <dfn>type</dfn> member specifies the type of credential to be created.
 
-    The <dfn>algorithm</dfn> member specifies the cryptographic algorithm with which the newly generated credential will be
-    used.
+    The <dfn>algorithm</dfn> member specifies the cryptographic key generation algorithm parameters with which the credential 
+    key pair MUST be generated.
```
vijaybh: This changes the meaning of the sentence, and in a potentially inaccurate way. Two problems here - changing "cryptographic algorithm" to "key generation algorithm" and changing "algorithm" to "algorithm parameters".

The algorithm ID is the signature algorithm ID. It is not the key generation algorithm ID, which can be different (e.g. there are multiple algorithms to generate an RSA key, but we would not be specifying which one of those to use, we'd just be saying RSA).

Also, the algorithm may or may not have parameters (e.g. with RSA one might specify a modulus length but maybe not with other algorithms). In this case I think you mean "parameters" as in parameters to the WebAuthn API but due to the overloaded nature of the term this would not be the obvious reading I think.

index.bs
```diff
@@ -1063,26 +1072,27 @@ Upon receiving an attestation statement, the [RP] shall:
...
 
 - Reject the request, such as a registration request, associated with the attestation statement.
 
 - Accept the request associated with the attestation statement but treat the attested Scoped Credential as one with surrogate
-    basic attestation (see [[#attestation-types]]), if policy allows it. If doing so, there is no cryptographic proof that the
-    Scoped Credential has been generated by a particular Authenticator model. See [[FIDOSecRef]] and [[UAFProtocol]] for more
-    details on the relevance on attestation.
+    basic attestation (see [[#sctn-attestation-types]]), if policy allows it. If doing so, there is no cryptographic proof 
+    that the Scoped Credential has been generated by a particular Authenticator model. See [[FIDOSecRef]] and [[UAFProtocol]]
+    for more details on the relevance on attestation.
```
vijaybh: Not your fault, but probably should be "relevance to attestation" at the end there.

equalsJeffH: agreed, thx.

index.bs
equalsJeffH: > The algorithm ID is the signature algorithm ID. It is not the key generation algorithm ID, which can be different (e.g. there are multiple algorithms to generate an RSA key, but we would not be specifying which one of those to use, we'd just be saying RSA).

ok, then we need to make that distinction clear. How about..
```
The algorithm member specifies the cryptographic signature algorithm with which the newly generated credential will be used, and thus also the type of asymmetric key pair to be generated, e.g., RSA or Elliptic Curve. 
```
..?

Issue #117, opened 2016-06-02
challenge in ClientData should be ArrayBuffer or BufferSource ?


equalsJeffH:
https://w3c.github.io/webauthn/#sec-client-data
```dictionary ClientData {
    required DOMString           challenge;    <<==  should be ArrayBuffer or BufferSource ?
    [...]
};
```
I *think* ArrayBuffer is more appropriate in ClientData than BufferSource, but I'm not sure, and I'm not sure it actually matters since ClientData is used internal-to-the-authnr in order to marshal info before hashing it.

vijaybh:
No, this is left as DOMString on purpose.

ClientData is, as you note, an internal structure that needs to be serialized and hashed. So we need to define what that serialization looks like. We have two choices:

1. Define these types as DOMString, and say they will be produced by encoding the appropriate quantity as base64.
2. Define these types as ArrayBuffer, in which case they will be serialized in the array literal notation, i.e. something like "{ challenge: [31, 41, 59, 26, 53, 58, 97, 93, 238, ...], ... }"

The first seems more readable and is more compact, so I went with that when fixing #61.

PR #186, opened 2016-08-30
breakup IDL

### LINKED ISSUE #112: consider moving IDL from monolithic block to individual spec sections
now that we are using bikeshed which automagically produces a consolidated webidl at the end of the spec <https://w3c.github.io/webauthn/#idl-index> we could perhaps move  individual IDL fragements to appropriate spots in the spec...

equalsJeffH:
This simply moves the various IDL pieces from one single monolithic block to the individual spec sections where they are discussed.  Bikeshed automagically provides a consolidated IDL appendix in any case. fixes #112 and #117.


jcjones:
LGTM.

PR #193, opened 2016-09-07
Update draft-hodges-webauthn-registries.xml

gmandyam:
Primarily for attestation registry

### DIFF
--- a/draft-hodges-webauthn-registries.xml
+++ b/draft-hodges-webauthn-registries.xml
@@ -28,9 +28,9 @@
 <rfc category="info" ipr="trust200902" docName="draft-hodges-webauthn-registries-00c">
   <front>
     <title>Registries for Web Authentication (WebAuthn)</title>
-    
+
     <author initials="J." surname="Hodges" fullname="Jeff Hodges">
-          <organization>PayPal</organization>
+      <organization>PayPal</organization>
       <address>
         <postal>
           <street>2211 North First Street</street>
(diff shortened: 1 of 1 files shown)


draft-hodges-webauthn-registries.xml
```diff
@@ -42,164 +42,207 @@
...
-        target="WebAuthn"/> attestation formats and extension identifiers, and supplies initial 
+        target="WebAuthn"/> attestation formats and extension identifiers, and supplies initial
         entries within each registry.
       </t>
+      <t>
+        Candidate entries must be announced on the public Web Authentication mailing list,
+        public-webauthn@w3.org, so they can be discussed and evaluated for compliance 
+        to the requirements presented in this document before being added to its associated
+        registry.
+      </t>
```
equalsJeffH: I do not agree with adding the above paragraph.  This is the sort of information that is addressed below where we say..

"The Expert(s) will establish procedures for requesting registrations, and make them available from the registry page. and make them available from the registry page."

..Mark Nottingham (mnot) indicated that their experience with RFC5988 and its rigid stipulation of registry rules was that it was too bureaucratic and inflexible (one has to create a new RFC to alter them (eg fine grained details like which mailing list, detail-level criteria, etc).  

Thus we should only say, in this spec, that the "experts" will make and publish the rules on the registry page.

gmandyam: There is clearly precedent for this in other W3C registries (see requirement 5 of https://www.w3.org/2013/12/byte-stream-format-registry/).  I disagree and will not make the change.

gmandyam: Apologies to @equalsJeffH.  After group discussion on Sep. 7 2016 call, it is clear that expert review involving posting to the public webauthn mailing list will be specified.  It just won't be specified in this document.  I will adjust the PR accordingly.

draft-hodges-webauthn-registries.xml
```diff
@@ -28,9 +28,9 @@
 <rfc category="info" ipr="trust200902" docName="draft-hodges-webauthn-registries-00c">
   <front>
     <title>Registries for Web Authentication (WebAuthn)</title>
-    
+
     <author initials="J." surname="Hodges" fullname="Jeff Hodges">
-          <organization>PayPal</organization>
+      <organization>PayPal</organization>
```
vijaybh: I thought we were using faux entities not real names and companies?

draft-hodges-webauthn-registries.xml
```diff
@@ -235,146 +276,209 @@
...
+              Specification Document: <xref target="WebAuthn"/>
+            </t>
+          </list>
+        </t>
+        <t>
+          <vspace blankLines="1" />
+        </t>
+        <t>
+          <list style="empty">
+            <t>WebAuthn Attestation Formats: goog-android</t>
```
equalsJeffH: I do not understand the need for such a prefix ?

gmandyam: See b) in https://lists.w3.org/Archives/Public/public-webauthn/2016Aug/0136.html.

draft-hodges-webauthn-registries.xml
```diff
@@ -42,164 +42,201 @@
...
+        <postal>
+          <street>5775 Morehouse Drive</street>
+          <!-- Reorder these if your country does things differently -->
+          <city>San Diego</city>
+          <region>California</region>
+          <code>92121</code>
+          <country>USA</country>
+        </postal>
+        <phone>+1 858 651 7200</phone>
+        <email>mandyam@qti.qualcomm.com.com</email>
```
gmandyam: @jcjones Thanks.  Have addressed.

draft-hodges-webauthn-registries.xml
equalsJeffH: (b) says you suggest a vendor prefix, but doesn't supply justification -- what might that be?

equalsJeffH:
this change looks fine, thanks -- https://github.com/w3c/webauthn/pull/193/commits/aaf184cb2f4b0c09d8e0c96bea2acf736e50c902

PR #194, opened 2016-09-08
Adds optional transport hints to CredentialDescription

### LINKED ISSUE #6: Where should transport hints go?
<a href="https://github.com/balfanz"></img></a> **Originally submitted by: [balfanz](https://github.com/balfanz)**, on: _Wednesday Aug 19, 2015 at 22:20 GMT_

----

We currently don't have a way to hint to the platform which transports an Authenticator supports. The RP shouldn't have to be consciously aware of this (hence we don't have a parameter for this in the API). Instead, the transport hints need to be encoded somewhere else - perhaps as part of the Credential IDs. We need to figure out how we're going to do this.

leshi:
Resolves #6


vijaybh:
Should there be some addition to the processing rules for getAssertion mentioning how to potentially use this?

leshi:
Right.  I did think a bit about it and looked at the processing rules as they are right now.  I didn't find a spot that really resonated with me.  Also, since the transports are just hints -- I thought maybe it was okay to omit mention of them from the processing rules.

I'll check again to see if putting them somewhere makes sense.

One question I had was whether we should have an "internal" transport.  I purposely labeled the transports "external" to avoid any confusion.  WDYT?

vijaybh:
I wonder what adding "internal" as a transport would do that omitting the transport hint doesn't. I would assume the client already knows its internal authenticators and does not need hinting for them.

About the processing rules, it seems to me that the transport hints don't really make sense in the excludeList (or can you think of a way to use them there?) so I wonder if we could say something about using the transport hints to inform the filtering process in step 7 of section 4.1.2 (i.e. the getAssertion section).

index.bs
```diff
@@ -788,6 +789,32 @@ the {{makeCredential()}} or {{getAssertion()}} method. It mirrors the fields of
...
 
 
+### Credential Transport enumeration (enum <dfn enum>ExternalTransport</dfn>) ### {#transport}
+
+<pre class="idl">
+    enum ExternalTransport {
+        "USB",
+        "NFC",
+        "BLE"
+    };
```
equalsJeffH: in https://fidoalliance.org/specs/fido-u2f-v1.0-nfc-bt-amendment-20150514/fido-u2f-javascript-api.html the Transport enum is..
```
enum Transport {
    "bt",
    "ble",
    "nfc",
    "usb"
};
```
..any reason for webauthn's ExternalTransport to differ? ie, why not also have "bt", and perhaps it'll be better for deployers to also use lowercase for enum members since that's what U2F is using?

equalsJeffH:
@vijaybh wrote:
> Should there be some addition to the processing rules for `getAssertion` mentioning how to potentially use this?
@leshi replied: 
>  I did think a bit about it and looked at the processing rules as they are right now. I didn't find a spot that really resonated with me. Also, since the transports are just hints -- I thought maybe it was okay to omit mention of them from the processing rules.
@vijaybh replied:
> About the processing rules, it seems to me that the transport hints don't really make sense in the excludeList (or can you think of a way to use them there?) so I wonder if we could say something about using the transport hints to inform the filtering process in step 7 of section 4.1.2 (i.e. the getAssertion section).

I tend to agree with this suggestion regarding the filtering process in `getAssertion` processing rules. I also note U2F does not say anything specific regarding the use of the `transports` member of `RegisteredKey` during a register operation.  

For a `u2f.sign()` request (equivalent of `getAssertion()`), this is what is noted in the processing rules...
> When the RP provides the transports value for any RegisteredKey, the client MAY treat that value has a hint about which transports to prefer for the key handle. The client MAY also use the transports as a hint about user interface, if the client presents any. Irrespective of whether the RP sets any transports value for any RegisteredKey, the client SHOULD send each key handle over all transports supported by the client. 

Perhaps WebAuthn should say at least the equivalent, if the "Transports" hints notion is essentially the same as in U2F?


@vijaybh wrote:
> I wonder what adding "internal" as a transport would do that omitting the transport hint doesn't. I would assume the client already knows its internal authenticators and does not need hinting for them.

I tend to agree with this.  Also, we removed almost all occurrences of the term "external" from the spec, so I wonder whether the proposed ExternalTransport enum should just be termed "Transport" as in U2F  https://fidoalliance.org/specs/fido-u2f-v1.0-nfc-bt-amendment-20150514/fido-u2f-javascript-api.html.  And if the RP omits the Transport hint, then it is presumed the underlying "platform" can figure things out. 

I also note that U2F allows for a authenticator to be reachable over more than a single transport...
```
enum Transport {
    "bt",
    "ble",
    "nfc",
    "usb"
};

typedef sequence<Transport> Transports;
```
..e.g., a smartphone-based authnr might work via either BT or NFC say.  Might we desire the same in WebAuthn?

index.bs
leshi: When we defined BT in the U2F JavaScript API, we believed that folks would build Authenticators using BR/EDR (or what we call "bluetooth classic").  However, that never came to be and we removed that from the Bluetooth specification.  Said another way "bt" is meaningless in U2F.  I'm trying to avoid making that mistake in webauthn :)

I'll make everything lowercase.

leshi:
Addressing comments:

**Processing Rules and exclude list**
The hints are needed because of how the exclude list is used in practice.  Specifically, for Authenticators that don't have storage, the Client will actually use the exclude list to construct dummy challenges to those Authenticators.  If a non-error reply is received, then the Client knows to exclude said Authenticator.  Hence, the Client gets benefit from the transport hints in such a case.

**About 'internal' authenticators**
Consider the case where a Client receives a credential without any hints.  The Client checks to see if it's from a local Authenticator and finds out 'no'.  Now, this credential could either be an internal credential from another device, so it doesn't even make sense for this Client to ask the user to plug anything in, pair anything, tap anything.  Or it could be a credential from an external device, so some user action is required.  What is a Client to do in such cases?

**Multiple transports**
As written multiple transports are supported.  Did I miss something?

**How the server gets transports in the first place**
During the call, we proposed that this comes from metadata service.  Do people think that I should go ahead and add such details to the spec?

**Processing Rules**
Will add after above issues are resolved.

vijaybh:
OK, so sounds like you're saying the list of transports should include 'internal' and should be used also in the makeCredential? I haven't given this enough thought to intelligently contradict that, so I'm okay with trying it out.

index.bs
```diff
@@ -374,6 +374,8 @@ When this method is invoked, the user agent MUST execute the following algorithm
 9. For each authenticator currently available on this platform: asynchronously invoke the <a>authenticatorMakeCredential</a>
     operation on that authenticator with |rpIdHash|, <a>clientDataHash</a>, <a>accountInformation</a>, `normalizedParameters`,
     <a>excludeList</a> and |clientExtensions| as parameters. Add a corresponding entry to |issuedRequests|.
+    - For each credential C in <a>excludeList</a> that has a non-empty |transports| list, use only the specified transports
+       to test for the existence of C.
```
vijaybh: Needs more indent to be valid Markdown.

Also, this makes it mandatory for the client to use the hint, maybe we should soften that a bit?

index.bs
```diff
@@ -439,6 +441,8 @@ When this method is invoked, the user agent MUST execute the following algorithm
     - If <a>allowList</a> is undefined or empty, let |credentialList| be an empty list. Otherwise, execute a platform-specific
         procedure to determine which, if any, credentials listed in <a>allowList</a> might be present on this authenticator, and
         set |credentialList| to this filtered list. If no such filtering is possible, set |credentialList| to an empty list.
+    - For each credential C within the |credentialList| that has a non-empty |transports| list, use only the specified transports
+        to get assertions using credential C.
```
vijaybh: Also needs more indent (see Travis-CI error).

Also might want to soften this so the hint is not mandatory for the client.

leshi:
BT was removed from most of the U2F's specs -- we figured out that nobody wanted to implement it (every one wanted to do BLE).

index.bs
leshi: This one is fine I think.

vijaybh:
LGTM. I suspect we will want to polish the wording as part of a later editorial pass, but that is better done as one pass over the entire document.

Issue #189, opened 2016-08-31
Should rpId be removed from clientData?


equalsJeffH:
see: https://github.com/w3c/webauthn/pull/154#issuecomment-241952439 where @leshi writes: 
> Should `rpId` be removed from the clientData give[n] that it's explicitly part of the signature format now?

Issue #178, opened 2016-08-18
lowercase form of this RP ID


annevk:
You need to actually define what this means.

However, given that origins are already normalized, perhaps the algorithm you use (which you also don't seem to define) to derive the RP ID is already fine and "lowercase form" is just adding confusion.

annevk:
Yeah, I meant normalized with respect to the dictionary meaning, not as a standards term.

equalsJeffH:
OK, thx, thought so.   Though, is the answer to my question (2nd parag) "yes", i.e., origin.host and origin.domain are asciiDomains, by default, if non-null ?

annevk:
They are. Note that the domain field is only there for legacy `document.domain` purposes. You don't want to touch it here.

equalsJeffH:
this issue is closely related to #171 and will be addressed as part of addressing the latter.

equalsJeffH:
this is addressed by PR #202 -- closing manually.

annevk:
You still haven't defined lowercase as far as I can tell...

annevk:
You closed a bunch of my issues, but I'm not really convinced the fixes are it.

equalsJeffH:
sorry, this is going to be fixed by PR #198, which I thought was going to get merged right after we merged PR #202, but that didn't occur, reopening...

Issue #172, opened 2016-08-17
What if callerOrigin is an opaque origin


annevk:
You can't get a public suffix out of that.

annevk:
In particular, you can only determine that if the origin has a host and the host is not an IP address of some kind.

annevk:
(See the WHATWG HTML Standard for the actual terms you need to use here.)

equalsJeffH:
i believe @annevk is referring to: https://html.spec.whatwg.org/multipage/browsers.html#origin

balfanz:
@annevk - can you explain what the problem would be? Is there something in the spec that wouldn't work for opaque origins?

annevk:
Yes, see OP.

equalsJeffH:
this issue is closely related to #171 and will be addressed as part of addressing the latter.

vijaybh:
I would propose that we do not handle such cases for now. For v1 I would focus on the core use case of an interactive user in an active browsing context.

annevk:
You need to handle it somehow. You can't just leave things undefined.

rlin1:
see https://www.w3.org/TR/html51/browsers.html#opaque-origin for definition of opaque origin.

annevk:
You really want to be using https://html.spec.whatwg.org/multipage/browsers.html#concept-origin-opaque as a reference at all times. The copy-and-pasters of that work don't always do a good job.

rlin1:
In what situations would such opaque origins appear?

annevk:
Data URLs and sandboxing, at least.

rlin1:
I see two potential approaches to deal with such situation:

Approach 1:
Some JavaScript without a clear association of a creator browsing context could be seen as using the Browser as an "App".

Remember: outside the web use case, Apps can also create such scoped authentication credentials.  But instead of scoping the credential to some web server the App might be talking to, we scope the credential to the App itself.

A web browser apparently can act on behalf of some RP identified by a non-opaque origin, or it could act as an App, i.e. within the scope of a browser vendor.

Approach 2:
Such opaque origin will never be equal to any other opaque-origin.  Consequently, it seems impossible at any later point in time to access things which have been scoped to such Opaque Origin.
Persistently stored credentials scoped to such Opaque Origin could never be re-used at a later point (i.e. after this opaque-origin is gone).
Why would someone want to do that?  So we might argue there is no point in supporting that and makeCredential would just fail.

Opinions?

annevk:
[...]

Issue #171, opened 2016-08-17
callerOrigin


annevk:
What do you mean by origin of the caller?

equalsJeffH:
you're apparently referring to..
>Set callerOrigin to the origin of the caller.

..in both..
https://w3c.github.io/webauthn/#makeCredential
https://w3c.github.io/webauthn/#getAssertion

"origin of the caller" refers to the origin of the document whose javascript is invoking either makeCredential() or getAssertion(). 

perhaps you could suggest more appropriate terminology?

annevk:
So these APIs are not available in workers?

Note that "document whose JavaScript" is woefully imprecise since the JavaScript can come from anywhere.

What you likely mean is something like "current settings object's origin" (see WHATWG HTML Standard).

equalsJeffH:
@annevk asked:
> So these APIs are not available in workers?

it is not entirely clear at this time (seems to me) -- see also: #199

PR #198, opened 2016-09-14
Allow callers to explicitly specify RP ID

vijaybh:
Alternative approach to #162 suggested by @balfanz. This change also gets rid of references to PSL and eTLD+1, replacing them with references to the HTML spec.


index.bs
```diff
@@ -130,8 +130,8 @@ or a combination of both.
 
 This specification relies on several other underlying specifications.
 
-: HTML5
-:: The concept of <dfn for='web'>origin</dfn> and the <dfn>Navigator</dfn> interface are defined in [[!HTML5]].
+: HTML 5.1
```
equalsJeffH: I would s/HTML 5.1/HTML/ here since the citation in the next line is clearly to HTML51.

index.bs
```diff
@@ -222,7 +218,8 @@ NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED",  "MAY", and
 
 : <dfn>Relying Party Identifier</dfn>
 : <dfn>RP ID</dfn>
-:: A Relying Party Identifier is derived from a <a>[RP]</a>'s web origin's hostname by computing the hostname's <a>eTLD+1</a>. 
+:: A Relying Party Identifier defines the scope of a given credential, i.e. the set of web origins that the client will permit
```
equalsJeffH: s/web origin/origin/

index.bs
```diff
@@ -345,10 +342,13 @@ When this method is invoked, the user agent MUST execute the following algorithm
 2. Let |promise| be a new <a data-lt="Promises">Promise</a>. Return |promise| and start a timer for |adjustedTimeout| seconds.
     Then asynchronously continue executing the following steps.
 
-3. Set |callerOrigin| to the <a link-for='web'>origin</a> of the caller. Derive the RP ID from |callerOrigin| by computing the
-    "public suffix + 1" or "PS+1" (which is also referred to as the "Effective Top-Level Domain plus One" or "<a>eTLD+1</a>")
-    part of |callerOrigin| [[PSL]]. Let |rpId| be the lowercase form of this RP ID. Set |rpIdHash| to the SHA-256 hash of the
-    UTF-8 encoding of |rpId|.
+3. Set |callerOrigin| to the <a link-for='web'>origin</a> of the caller. If {{CredentialOptions/rpId}} is not specified, then
```
equalsJeffH: fyi, the fix I'm working on for #171 & #172 alters the first sentence, but not the remainder of this step, i.e., it ought to mesh ok (I hope).

index.bs
```diff
@@ -345,10 +342,13 @@ When this method is invoked, the user agent MUST execute the following algorithm
...
+    set |rpId| to the Unicode serialization of |callerOrigin| as specified in [[!HTML51-20160621]] section 6.4. If
```
equalsJeffH: I do not think we want/need to invoke the unicode serialization here.  The |rpId| will already be in asciiDomain format and we shouldn't alter.  Anne confirmed in the comments on #178 that origin.host and origin.domain are asciiDomains, by default, if non-null.

index.bs
```diff
@@ -345,10 +342,13 @@ When this method is invoked, the user agent MUST execute the following algorithm
...
+    {{CredentialOptions/rpId}} is specified, then check if its value would be an acceptable setting for the current document's
+    `domain` by running the algorithm in [[!HTML51-20160621]] section 6.4.1 (but do not change the current document's `domain`).
+    If it is not permissible, reject |promise| with a <a>DOMException</a> whose name is "SecurityError", and terminate this
+    algorithm. If it is permissible, then set |rpId| to the Unicode serialization of {{CredentialOptions/rpId}}. Set |rpIdHash|
+    to the SHA-256 hash of the UTF-8 encoding of |rpId|.
```
equalsJeffH: I think we want to write the above set of lines more like so:
>If {{CredentialOptions/rpId}} is specified, then invoke the <a link-for='web'>relaxing the same-origin restriction</a> "setting" algorithm using {{CredentialOptions/rpId}} as the `given value` without setting <a link-for='web'>origin</a>'s |domain|. If no errors are thrown, then set |rpId| to the value of |host| as yielded by the algorithm. Set |rpIdHash| to the SHA-256 hash of |rpId|. 

given that the internal default representation of `origin.host` and `origin.domain` is `asciiDomain` ( U-labels have been converted to A-labels (see https://tools.ietf.org/html/rfc5890#section-2.3.2 and foregoing figure 1 on pg 10 ) ) format, we should not be stipulating UTF-8 encoding of them, AFAICT. 

The jeffh-origin-et-al branch will soon have the `<pre class="anchors">` anchors set up to properly ref these things in HTML51 and so once both branches are eventually merged we should be good.

index.bs
```diff
@@ -421,10 +421,13 @@ When this method is invoked, the user agent MUST execute the following algorithm
...
```
equalsJeffH: similar comments to the above chunk of test as I had on the foregoing.

index.bs
```diff
@@ -564,6 +568,9 @@ authorizing an authenticator with which to complete the operation.
 
     - The <dfn>timeoutSeconds</dfn> parameter specifies a time, in seconds, that the caller is willing to wait for the call to
         complete. This is treated as a hint, and may be overridden by the platform.
+    
+    - The <dfn>rpId</dfn> parameter explicitly specifies the RP ID that the credential should be associated with. If it is
+        omitted, the RP ID will be set to the caller's origin.
```
equalsJeffH: i suspect that the term "caller's origin" will not pass muster here -- will try to figure out more precise term for where we say "caller" or "caller's".

index.bs
```diff
@@ -620,6 +628,9 @@ user consent to a specific transaction. The structure of these signatures is def
 
     - The optional <dfn>timeoutSeconds</dfn> parameter specifies a time, in seconds, that the caller is willing to wait for the
         call to complete. This is treated as a hint, and may be overridden by the platform.
+        
+    - The optional <dfn>rpId</dfn> parameter specifies the rpId claimed by the caller. If it is omitted, it will be assumed to
+        be equal to the caller's origin.
```
equalsJeffH: see comment above wrt "caller" et al.

index.bs
equalsJeffH: similar comments to the above chunk of text as I had on the foregoing.

index.bs
```diff
@@ -222,7 +218,8 @@ NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED",  "MAY", and
 
 : <dfn>Relying Party Identifier</dfn>
 : <dfn>RP ID</dfn>
-:: A Relying Party Identifier is derived from a <a>[RP]</a>'s web origin's hostname by computing the hostname's <a>eTLD+1</a>. 
+:: A Relying Party Identifier defines the scope of a given credential, i.e. the set of web origins that the client will permit
+    to access that credential. It is derived from a <a>[RP]</a>'s web origin's hostname or directly specified by the [RP].
```
equalsJeffH: s/access/wield/

suggest revise last sentence to be:
> It a <a>[RP]</a>'s origin's |host| component, or is explicitly specified by the [RP] to be either its origin's |host| component or the |host| component with subdomains removed such that it continues to pass the checks in <a>relaxing the same-origin restriction</a> algorithm.  See step 3 of both {{#makeCredential}} and {{#getAssertion}}.

index.bs
```diff
@@ -345,10 +342,13 @@ When this method is invoked, the user agent MUST execute the following algorithm
...
     Then asynchronously continue executing the following steps.
 
-3. Set |callerOrigin| to the <a link-for='web'>origin</a> of the caller. Derive the RP ID from |callerOrigin| by computing the
-    "public suffix + 1" or "PS+1" (which is also referred to as the "Effective Top-Level Domain plus One" or "<a>eTLD+1</a>")
-    part of |callerOrigin| [[PSL]]. Let |rpId| be the lowercase form of this RP ID. Set |rpIdHash| to the SHA-256 hash of the
-    UTF-8 encoding of |rpId|.
+3. Set |callerOrigin| to the <a link-for='web'>origin</a> of the caller. If {{CredentialOptions/rpId}} is not specified, then
+    set |rpId| to the Unicode serialization of |callerOrigin| as specified in [[!HTML51-20160621]] section 6.4. If
+    {{CredentialOptions/rpId}} is specified, then check if its value would be an acceptable setting for the current document's
+    `domain` by running the algorithm in [[!HTML51-20160621]] section 6.4.1 (but do not change the current document's `domain`).
```
equalsJeffH: fyi/fwiw, the linking style used in PR #202 allows direct linking to things like that algorithm such that we do not need to mention section #s.  also, i suggest citing HTML51 as simply [[!HTML51]].

index.bs
```diff
@@ -553,6 +554,7 @@ authorizing an authenticator with which to complete the operation.
 <pre class="idl">
     dictionary CredentialOptions {
         unsigned long                       timeoutSeconds;
+        USVString                           rpId;
```
balfanz: Why USVString here? See http://heycam.github.io/webidl/#idl-USVString

index.bs
```diff
@@ -610,6 +615,7 @@ user consent to a specific transaction. The structure of these signatures is def
 <pre class="idl">
     dictionary AssertionOptions {
         unsigned long                      timeoutSeconds;
+        USVString                          rpId;
```
balfanz: Why USVString here? See http://heycam.github.io/webidl/#idl-USVString

index.bs
```diff
@@ -691,7 +700,6 @@ string-valued keys. Values may be any type that has a valid encoding in JSON. It
     dictionary ClientData {
         required DOMString           challenge;
         required DOMString           origin;
-        required DOMString           rpId;
```
balfanz: How come you're removing it from the ClientData?

vijaybh: See #189. The rpId is being passed in the options so it's not clear why it should also be in the client data. Also the RP ID now gets hashed into the signature as well so it seems redundant to have this here.

index.bs
vijaybh: This RP ID will be sent to the authenticator "over the wire" when authenticatorMakeCredential is invoked. That seems to fit the USVString use case in the above guidance, though I would welcome expert guidance here.

index.bs
vijaybh: See above comment regarding same issue in CredentialOptions.

index.bs
vijaybh: I rewrote the definition - let me know if this is better.

index.bs
vijaybh: Fixed.

index.bs
vijaybh: Also fixed.

index.bs
vijaybh: Tweaked the text a little bit to hopefully make it better.

equalsJeffH:
I've reviewed portions of the latest commits and see some issues that ought to be cleaned up. I may be able to comment more tonight, and if not, I will tomorrow morning.

index.bs
equalsJeffH: fyi/fwiw, i did nose around in the HTML spec(s) and they also use the term "caller" (in terms of the caller of an algorithm/operation), so I think/hope we're ok with using it in a similar fashion.

index.bs
```diff
@@ -1287,16 +1288,15 @@ Upon receiving an attestation statement in the form of a {{WebAuthnAttestation}}
 
 2. Verify that the {{ClientData/challenge}} in the {{ClientData}} matches the challenge that was sent to the authenticator.
 
-3. Verify that the {{ClientData/origin}} and {{ClientData/rpId}} in the {{ClientData}} match the origin and RP ID used by the
-    RP.
+3. Verify that the {{ClientData/origin}} in the {{ClientData}} matches the origin used by the RP.
```
equalsJeffH: s/origin used by the RP/[RP]'s origin/  ?

equalsJeffH:
I did "approve changes" ahead of time so @vijaybh is otherwise unimpeded in finishing this up and merging.

equalsJeffH:
Also, the current state of these changes (updated per my latest comments) may address @annevk's concerns expressed in #218 "relaxing the same-origin restriction", though we have not answered his query "why are you using this?"

index.bs
```diff
@@ -383,14 +379,14 @@ When this method is invoked, the user agent MUST execute the following algorithm
...
-    "public suffix + 1" or "PS+1" (which is also referred to as the "Effective Top-Level Domain plus One" or "<a>eTLD+1</a>")
-    part of |callerOrigin| [[PSL]]. Let |rpId| be the lowercase form of this RP ID. Set |rpIdHash| to the SHA-256 hash of the
+3. Set |callerOrigin| to the <a>current settings object</a>'s <a link-for='web'>origin</a>. If |callerOrigin| is an <a>opaque
+    origin</a>, reject |promise| with a <a>DOMException</a> whose name is "NotAllowedError", and terminate this algorithm.
+    Otherwise, if {{ScopedCredentialOptions/rpId}} is not specified, then set |rpId| to |callerOrigin|. If
+    {{ScopedCredentialOptions/rpId}} is specified, then invoke the procedure used for
+    <a>relaxing the same-origin restriction</a> by setting the `document.domain` attribute, using
+    {{ScopedCredentialOptions/rpId}} as the given value but without changing the current document's `domain`. If any errors are
+    thrown, reject |promise| with a <a>DOMException</a> whose name is "SecurityError", and terminate this algorithm. If no
+    errors are thrown, set |rpId| to the value of `host` as computed by the algorithm. Set |rpIdHash| to the SHA-256 hash of the
```
vijaybh: Thanks. I also rearranged the sentences so it reads a bit better (avoid switching back and forth between "this algorithm" and "that procedure".

index.bs
```diff
@@ -383,14 +379,14 @@ When this method is invoked, the user agent MUST execute the following algorithm
...
     UTF-8 encoding of |rpId|.
```
vijaybh: You're right, but I wonder if we should be Unicode serializing the RP ID here since it will go to the authenticator which may want to display it. However we can discuss that separately.

PR #202, opened 2016-09-16
Proper origin terminology, references, and setting of callerOrigin 

equalsJeffH:
This resolves several issues regarding "origin" and references into the HTML (html51) spec.  It needs to be composed with #198 to fully make sense.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -37,7 +37,8 @@ Markup Shorthands: css off, markdown on
 
 <pre class="anchors">
 
-spec: HTML; urlPrefix: https://html.spec.whatwg.org/multipage/
+<!-- spec: HTML; urlPrefix: https://html.spec.whatwg.org/multipage/ -->
+spec: HTML51; urlPrefix: http://www.w3.org/TR/html51/
     type: dfn
         urlPrefix: browsers.html; for: web
             text: origin
@@ -147,7 +148,7 @@ This specification relies on several other underlying specifications.
 <!-- , <a>opaque origin</a>, -->
 
 : HTML
-:: The concepts of <a for='web'>origin</a> and the <a>Navigator</a> interface are defined in [[!HTML]].
+:: The concepts of <a for='web'>origin</a> and the <a>Navigator</a> interface are defined in [[!HTML51]].
 
 : Web IDL
 :: Many of the interface definitions and all of the IDL in this specification depend on [[!WebIDL-1]]. This updated version of
--- a/index.bs
+++ b/index.bs
@@ -38,15 +38,16 @@ Markup Shorthands: css off, markdown on
 <pre class="anchors">
 
 <!-- spec: HTML; urlPrefix: https://html.spec.whatwg.org/multipage/ -->
-spec: HTML51; urlPrefix: http://www.w3.org/TR/html51/
+spec: HTML51; urlPrefix: http://www.w3.org/TR/html51/; for: web
     type: dfn
-        urlPrefix: browsers.html; for: web
-            text: origin
+        urlPrefix: browsers.html
+            text: origin; url: concept-cross-origin
             text: opaque origin
             text: tuple origin
-            text: current settings object
+            text: relaxing the same-origin restriction
     type: dfn
         urlPrefix: webappapis.html; for: interface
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -130,8 +147,12 @@ or a combination of both.
 
 This specification relies on several other underlying specifications.
 
-: HTML5
-:: The concept of <dfn for='web'>origin</dfn> and the <dfn>Navigator</dfn> interface are defined in [[!HTML5]].
+<!-- , <a>opaque origin</a>, -->
+
+: HTML
+:: The concepts of <a link-for='web'>current settings object</a>, |given value|, <a link-for='web'>opaque origin</a>,
```
vijaybh: You can remove the for='web' everywhere now. It was only required for disambiguation before because there were two definitions of "origin" (one here, one as a member of ClientData). So the link-for helped bikeshed figure out which one we were talking about.

index.bs
```diff
@@ -345,7 +366,12 @@ When this method is invoked, the user agent MUST execute the following algorithm
 2. Let |promise| be a new <a data-lt="Promises">Promise</a>. Return |promise| and start a timer for |adjustedTimeout| seconds.
     Then asynchronously continue executing the following steps.
 
-3. Set |callerOrigin| to the <a link-for='web'>origin</a> of the caller. Derive the RP ID from |callerOrigin| by computing the
+3. Set |callerOrigin| to the <a link-for='web'>current settings object</a>'s <a link-for='web'>origin</a>. If |callerOrigin| is
+    an <a link-for='web'>opaque origin</a>, reject |promise| with a <a>DOMException</a> whose name is "NotAllowedError", and
+    terminate this algorithm. 
+    Otherwise, apply the <a link-for='web'>relaxing the same-origin restriction</a> "setting" algorithm using |callerOrigin| as
+    the |given value|. 
+    If no errors are thrown, then derive the RP ID from |callerOrigin| by computing the
```
vijaybh: If no errors are thrown, we should take the value of `host` computed by the algorithm, right? This sounds redundant, as if we need to compute eTLD again independently.

index.bs
```diff
@@ -421,7 +447,12 @@ When this method is invoked, the user agent MUST execute the following algorithm
 2. Let |promise| be a new <a data-lt="Promises">Promise</a>. Return |promise| and start a timer for |adjustedTimeout| seconds.
     Then asynchronously continue executing the following steps.
 
-3. Set |callerOrigin| to the <a link-for='web'>origin</a> of the caller. Derive the RP ID from |callerOrigin| by computing the
+3. Set |callerOrigin| to the <a link-for='web'>current settings object</a>'s <a link-for='web'>origin</a>. If |callerOrigin|
+    is an <a link-for='web'>opaque origin</a>, reject |promise| with a <a>DOMException</a> whose name is "NotAllowedError", and
+    terminate this algorithm.
+    Otherwise, apply the <a link-for='web'>relaxing the same-origin restriction</a> "setting" algorithm using |callerOrigin| as
+    the |given value|. 
+    If no errors are thrown, then derive the RP ID from |callerOrigin| by computing the
```
vijaybh: See above.

index.bs
equalsJeffH: hm, well, given this branch's current state, it needs the link-for='web' to compile.  perhaps we can clean that up after merging all these PRs to master?

index.bs
equalsJeffH: oh, so I didn't touch the text beyond the "derive the RP ID..." because PR #198 is altering it.  should we try to clean up before merging these PRs, or merge one to master, the merge master to the un-merged branch, clean-up, and then merge the latter?

index.bs
equalsJeffH: ditto :)

index.bs
vijaybh: WFM

index.bs
vijaybh: I see. Yeah, I would do (2).

index.bs
vijaybh: ditto :)

index.bs
equalsJeffH: ok, which PR merge first?

PR #203, opened 2016-09-16
add [SecureContext] extended attr to WebAuthentication member of Navi…

### LINKED ISSUE #146: Secure context should be MUST not SHOULD
WebAuthn depends on origin isolation to provide privacy protection for the user. Also, if origins can be spoofed, then MITM attacks become possible. Currently the spec says secure contexts SHOULD be required. We SHOULD really upgrade this to a MUST...

equalsJeffH:
…gator interface. fixes #146.


index.bs
```diff
@@ -288,6 +288,7 @@ The Web Authentication API is defined by the union of the Web IDL fragments pres
 
 <pre class="idl">
     partial interface Navigator {
+        [SecureContext]
```
vijaybh: Should this be on the definition of interface WebAuthentication also (or instead)? Also @mikewest seemed to suggest adding this to other interfaces as well, such as ScopedCredentialInfo and so on. I'm not sure what that would do, but it's a thought.

equalsJeffH: > Should this be on the definition of interface WebAuthentication also (or instead)? 

yes, as also noted by YaronS over in issue #146.  this was a haste-makes-waste moment, sigh.   Ok, so we instead do this..
```
    [SecureContext]
    interface WebAuthentication {
        Promise < ScopedCredentialInfo > makeCredential (
            Account                                 accountInformation,
            sequence < ScopedCredentialParameters > cryptoParameters,
            BufferSource                            attestationChallenge,
            optional CredentialOptions              options
        );

        Promise < WebAuthnAssertion > getAssertion (
            BufferSource               assertionChallenge,
            optional AssertionOptions  options
        );
    };
```
And in #146 @mikewest notes..
> I'd also suggest using the [SecureContext] IDL attribute on entry points to the API like navigator.authentication, as well as visible interfaces like ScopedCredentialInfo, WebAuthentication, etc.

..so I'll add that also.

equalsJeffH: see latest commit.

PR #220, opened 2016-09-23
Move accountInformation into options.

### LINKED ISSUE #219: Add explanation of why the account argument is useful
Richard Barnes wrote:

> I can't remember if we talked about this before.  Would it make sense to move the `account` argument to `makeCredential` into the `options` dictionary?  It seems like there are at least some credential types that don't require it (e.g., U2F credentials), and it makes the interface a bit simpler.

I agree.

leshi:
Closes #219


equalsJeffH:
see comments on #219 for discussion whether this change makes overall sense or not.

Issue #216, opened 2016-09-20
Moving callers from U2F to WebAuthn


leshi:
Add an optional extension that helps callers move from u2f to webauthn

We need an extension that specifies legacy appId

rlin1:
Please rename extension to webauthn-appid as it applies to other FIDO spec families as well.

vijaybh:
Closed by #229

PR #229, opened 2016-10-03
Adding fido appid extension

leshi:
Closes Issue #216


rlin1:
Please rename extension to webauthn_appid as it is used by other FIDO spec families as well.

index.bs
```diff
@@ -1923,6 +1923,34 @@ This section defines an initial set of extensions. These are recommended for imp
...
+|appId| to overwrite the otherwise computed |rpId|.
+
+: Extension identifier
+:: `webauthn_u2f-appid`
+
+: Client argument
+:: A single UTF-8 encoded string specifying a FIDO U2F |appId|.
+
+: Client processing
+:: AppId validation (as defined by the FIDO U2F specification).  If valid, the
```
vijaybh: Need a real reference for this. Please contribute to http://www.specref.org/ or add to the biblio block at the end of the Bikeshed source (it's also in SpecRef format).

leshi:
@rlin1, thanks for the suggestion!  Done.

index.bs
tobie: Merge: http://www.specref.org/?q=FIDO-APPID

leshi:
@vijaybh, any idea how to update travic cl's bikeshed to pull in the new biblio data?

plinss:
You should run 'bikeshed update' after install to get the latest spec/biblio data. Alternatively you can use the online bikeshed at https://api.csswg.org/bikeshed (which is always up to date)

jcjones:
@leshi: Can you do an update of the bikeshed data in the repo (in `.spec-data`) as another commit on this PR? That should fix it.

It'd be something like:
```
git clone --depth=100 --branch=master https://github.com/tabatkins/bikeshed.git ./bikeshed
pip install --editable ./bikeshed
bikeshed update
cp ./bikeshed/bikeshed/spec-data/* .spec-data/
```

and then test, and commit the changeset.

leshi:
Thanks, @jcjones, that's what I was looking for!

Thanks @tobie and @plinss!

index.bs
```diff
@@ -1923,6 +1923,34 @@ This section defines an initial set of extensions. These are recommended for imp
...
+: Extension identifier
+:: `webauthn_appid`
+
+: Client argument
+:: A single UTF-8 encoded string specifying a FIDO |appId|.
+
+: Client processing
+:: The Client must perform AppId validation (as defined by [[FIDO-APPID]]).  If
+    valid, the value of |rpId| for all Client processing should be replaced by
+    the specified |appId|.
```
vijaybh: What if rpId is also specified in the options? Should this win, or should the call just error out? (FWIW I'm in favor of erroring out by throwing an exception - it is a programming error and the code should never be written that way.)

index.bs
```diff
@@ -1923,6 +1923,34 @@ This section defines an initial set of extensions. These are recommended for imp
...
+Authenticators.  Specifically, this extension allows [RP]s to specify an
+|appId| [[FIDO-APPID]] to overwrite the otherwise computed |rpId|.
+
+: Extension identifier
+:: `webauthn_appid`
+
+: Client argument
+:: A single UTF-8 encoded string specifying a FIDO |appId|.
+
+: Client processing
```
vijaybh: Stylistically, I wonder if this would be better written as "Replace step 3 in section foo with the following: ..." so this would be very clear. Would also help with defining error behavior as below.

index.bs
```diff
@@ -1923,6 +1923,34 @@ This section defines an initial set of extensions. These are recommended for imp
...
 
+## FIDO AppId ## {#extension-appid}
+
+This authentication extension allows [RP]s who have previously registered an
+Authentication using the legacy FIDO JavaScript APIs to communicate with
+Authenticators.  Specifically, this extension allows [RP]s to specify an
+|appId| [[FIDO-APPID]] to overwrite the otherwise computed |rpId|.
+
+: Extension identifier
+:: `webauthn_appid`
```
equalsJeffH: that would be fine with me.

equalsJeffH: also: s/[RP]s/[RPS]/
"Authenticator" and "Client", in prose text should be lower case (except at beginning of sentence).

index.bs
equalsJeffH: ok, i can agree with the approach Vijay suggests -- though the spec ought to explicitly indicate that rpId/appId is an "either/or", both in this extension section, and in the section on `dictionary AssertionOptions`.

index.bs
equalsJeffH: Agreed, it would be good to be very explicit. See also comment below.

index.bs
leshi: Done on renaming `webauthn_appid` to `fido_appid`

Done on correcting capitalization of `Authenticator` and `Client`

index.bs
leshi: Added language about it being an error to explicitly specify the rpId if using this extension.

Also added language to say that it is an error to use this extension outside the scope of `getAssertion()`

index.bs
leshi: I have made this change as requested.

index.bs
```diff
@@ -1923,6 +1923,39 @@ This section defines an initial set of extensions. These are recommended for imp
 interoperability.
 
 
+## FIDO AppId ## {#extension-appid}
+
+This authentication extension allows [RP]s who have previously registered an
+Authentication using the legacy FIDO JavaScript APIs to communicate with
```
vijaybh: "registered an Authentication" sounds weird. Maybe "registered a credential"?

index.bs
```diff
@@ -1923,6 +1923,39 @@ This section defines an initial set of extensions. These are recommended for imp
...
+
+: Client argument
+:: A single UTF-8 encoded string specifying a FIDO |appId|.
+
+: Client processing
+:: Replace the calculation of |rpId| in Step 3 of [[#getAssertion]] with the
+    following procedure:  The client uses the value of |fido_appid| to perform
+    the AppId validation procedure (as defined by [[FIDO-APPID]]).  If valid,
+    the value of |rpId| for all client processing should be replaced by the
+    value of |fido_appid|.  It is an application error to both use this
```
vijaybh: Might be better to make the error explicit and move it to the beginning of the procedure. Something like "If {{AssertionOptions/rpId}} is present, reject promise with a DOMException whose name is "NotAllowedError", and terminate this algorithm. Otherwise, ..."

vijaybh:
Added two editorial suggestions. Other than that this LGTM.

index.bs
leshi: Thanks for the suggestion.  Wording improved.

index.bs
leshi: Done

equalsJeffH:
Looks fine to me.

PR #235, opened 2016-10-18
remove ScopedCredentialInfo.publicKey; use "credential public key" term consistently

### LINKED ISSUE #94: clarify conveyance of attested public key
the makeCredential() promise returns a ScopedCredentialInfo..

```
    readonly attribute Credential           credential;
    readonly attribute any                  publicKey;
    readonly attribute WebAuthnAttestation  attestation;
};
```

..wherein the publicKey attribute is intended to convey the "attested public key", which is the user's newly-generated public key to be registered with the webauthn relying party (WRP).

However, in the "Packed Attestation (type="packed")" section, there's this this text..

```
The rawData object contains the attested public key and the clientDataHash. 
See §4.3.2.1.1 Attestation rawData for details.
```

..which is clearly saying the attested public key is being conveyed in the rawData object. 

Perhaps a hash of the attested public key ought to be conveyed in the rawData object, and it be made clear that the plaintext attested public key be conveyed in ScopedCredentials.publicKey.

also note that our terminology for the so-called attested public key needs to be normailzed, see #79 .

equalsJeffH:
Fixes #94


index.bs
```diff
@@ -531,7 +537,6 @@ authorizing an authenticator with which to complete the operation.
     [SecureContext]
     interface ScopedCredentialInfo {
         readonly attribute ScopedCredential     credential;
-        readonly attribute CryptoKey            publicKey;
```
vijaybh: Should also delete lines 165-168 since there are no longer any references to CryptoKey or JsonWebKey left in this spec.

equalsJeffH: doh. thx :)   See new commit.

equalsJeffH:
I've made some updates to this PR #235 and have also stashed a textual diff from the current state of the master branch here..

http://kingsmountain.com/doc/diff/diff-webauthn-index-master-1eebeed-from-index-jeffh-pubkey-94-4b3d958.pdf

please review.

index.bs
```diff
@@ -1670,8 +1676,8 @@ with the fields of the attestation certificate's extension data.
     - Verify that {{AndroidKeyAttestation/signature}} is a valid certificate chain, consisting of a time-valid X.509 certificate
         chaining up to a trusted attestation root key.
 
-    - Verify that the public key in the attestation certificate matches the credential public key in the attestation data field
-        of the given <a>authenticatorData</a>.
+    - Verify that the public key in the attestation certificate matches the <a>credential public key</a> in the attestation
```
vijaybh: Suggestion based on earlier email thread: "attestation certificate" could be replaced with something like "leaf certificate in the chain represented by {{AndroidKeyAttestation/signature}}".

vijaybh:
One editorial suggestion which captures something we also discussed before. Other than that this LGTM.

bifurcation:
This PR removes important functionality.  Right now, it is possible for a RP to use this API without any knowledge of attestation formats (as long as it's willing to believe the browser that the key is good enough).  With this PR, the RP can now only interoperate with an authenticator if it understand the attestation format produced by that authenticator, and it has to have code to process all the attestation formats on its critical path, before it can even look at the public key.  Those seem like pretty serious constraints on interoperability and usability, so I'm inclined to close this PR.

equalsJeffH:
Hi Rolf, thx for the review. 

@rlin1 wrote:
>  With this PR, the RP can now only interoperate with an authenticator if it understand the attestation format produced by that authenticator

actually, that is no longer the case since PR #161 "make attestation more modular" which added this section..
```
{#generating-an-attestation-statement}

This section specifies the algorithm for generating an attestation statement, independent
of <a>attestation format</a>.
```
..and which specifies (via the table therein) a common format for conveying AAGUID and CredID and pubkey alg & encoding and attested public key, across all attstn formats. 

Thus  an RP may still, if it accepts the risk, simply pluck the attested public key from the attestation statement without understanding the various attstn formats (which remains necessary if the RP wishes to verify the attstn signature).

Additionally, this PR normalizes terminology, which we would want/need to do in any case.

index.bs
equalsJeffH: ok, thx, will address this.

equalsJeffH:
Ok, have addressed @vijaybh 's comment above. Would be good to have @rlin1 check the new language in {#android-key-attestation}.   

The generated index.html likely has some dangling hyperlinks, at least it does when processed  via the 'cloud (i.e., latest) bikeshed' -- we can fix these later i trust (they also exist in other branches I'm working on).

equalsJeffH:
I have fixed the dangling hyperlinks I've experienced in other branches I'm working on.

equalsJeffH:
webauthn 9-Nov meeting minutes https://www.w3.org/2016/11/09-webauthn-minutes.html has this regarding this PR:
<blockquote>
We are now discussing PR #235

It'd be good to have Rolf double-check on the language for PR #235
Vijay is ok with the change
There are some contentious points about #235. Alexei will respond to that.
There is no clear documentation for CBOR mapping. This issue will addressed in a separate conversation.
Alexei will take a look at PR #235 before we merge it
Is any issue on WD-03 that is not addressed by a open pull request?
</blockquote>

Alexei has been busy and not gotten to this, and we wish to get WD-03 published, and @vijaybh is ok with merging this, so am doing so. If there's issues with it, please file new issues, and we can address in another WD update.

PR #236, opened 2016-10-18
Jeffh editorial cleanups 2

equalsJeffH:
Various WD-03 editorial cleanups, see commit list for issues addressed. I may add more commits to this.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -106,7 +106,7 @@ scenarios. Additional scenarios, including sample code, are given later in [[#sa
 
 ### Authentication ### {#usecase-authentication}
 
-- On a laptop:
+- On a laptop or desktop:
     * User navigates to example.com in a browser, sees an option to "Sign in with your phone."
     * User chooses this option and gets a message from the browser, "Please complete this action on your phone."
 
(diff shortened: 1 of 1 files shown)


leshi:
LGTM

equalsJeffH:
just added another commit to this.

index.bs
```diff
@@ -124,7 +124,9 @@ scenarios. Additional scenarios, including sample code, are given later in [[#sa
 
 A variety of additional use cases and configurations are also possible, including (but not limited to):
 
-- User navigates to example.com on their laptop, is guided through a flow to create and register a credential on their phone.
+- A user navigates to example.com on their laptop, is guided through a flow to create and register a credential on their phone.
+
+- A user obtains an discreet, cross-platform authenticator, such as a "fob" with USB or USB+NFC/BLE connectivity options, loads example.com in their browser on a laptop or phone, and is guided though a flow to create and register a credential on the fob.
```
vijaybh: I assume you mean discrete (i.e. separate) rather than discreet (i.e. cautious or understated) here.

vijaybh: Regarding #11, I would not on that issue that we are not doing the part about being able to select credentials on the client, as that would require the client to be able to enumerate credentials from the device, leading to a privacy leak.

equalsJeffH: @vijaybh wrote:
> I assume you mean discrete (i.e. separate) rather than discreet (i.e. cautious or understated) here.

yes, the former, will fix, thanks :)     [http://grammarist.com/usage/discrete-discreet/]   

> Regarding #11, I would note on that issue that we are not doing the part about being able to select credentials on the client, as that would require the client to be able to enumerate credentials from the device, leading to a privacy leak.

perhaps you can make the above comment in #11 ?   

Are you otherwise ok with L129:
>A user obtains an discrete, cross-platform authenticator, such as a "fob" with USB or USB+NFC/BLE connectivity options, loads example.com in their browser on a laptop or phone, and is guided though a flow to create and register a credential on the fob. 

..?

equalsJeffH:
@leshi - thx for review

vijaybh:
LGTM as well.

Issue #213, opened 2016-09-20
Add notion of internal/external to options


leshi:
Add notion of internal/external to options so that Caller can indicate that it only wants to talk to internal authenticators.

Example:

```
dictionary CredentialOptions {
    unsigned long                       timeoutSeconds;
    sequence < CredentialDescription >  excludeList;
    WebAuthnExtensions                  extensions;
    Attachment                          attachment;
};

enum Attachment {
    "INTERNAL",
    "EXTERNAL"
}
```

jcjones:
Rather than this enumeration, perhaps we should look for something a bit richer, such as what [Media Capture and Streams API does for getUserMedia](https://w3c.github.io/mediacapture-main/#dom-mediadevices-getusermedia). (Moz docs:  https://developer.mozilla.org/en-US/docs/Web/API/MediaDevices/getUserMedia)

rlin1:
Note: internal/external and second-factor (only)/first-factor authentictors are different things.

yaronf:
And then maybe rename ``Attachment`` to ``AuthenticatorAttachment`` or even ``AuthenticatorFilter``.

equalsJeffH:
to build on @rlin1 's comment, there are perhaps more authnr attributes that may be of interest than just "internal / external", and thus a single 1-dimensional enum may prove insufficiently expressive.  However, we may not need the expressability of the rich "[constrainable patten](https://w3c.github.io/mediacapture-main/#constrainable-interface)" from "Media Capture and Streams", that @jcjones points to.

vijaybh:
I agree with @equalsJeffH - I wonder if we could just adopt this suggestion as-is and add more members to the Options dictionary in future if strong use cases come up. It's not clear to me that simply adding another layer of nesting (e.g. a Constraints structure with all these dimensions inside it) would be superior enough to justify the overhead.

leshi:
I like this simple structure.  Though I think the suggested naming was `PLATFORM` instead of 'internal' and `CROSS_PLATFORM` instead of 'external'.

leshi:
Took an initial stab at this in #243

apowers313:
Section 5.1.1 says:

> Prompt the user for consent to create a new credential. The prompt for obtaining this consent is shown by the authenticator if it has its own output capability, or by the user agent otherwise. 

As long as we are doing the modification above, do we need a "hasPrompt" option too? Or do you guys have another way in mind of figuring out if an authnr can prompt users?

[...]

PR #243, opened 2016-11-02
Adds authenticator attachement to credential options

leshi:
Closes issue #213

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -664,7 +664,7 @@ authorizing an authenticator with which to complete the operation.
 
 <div dfn-for="Attachment">
     Clients may communicate with authenticators using a variety of mechanisms.
-    For example, a client may use a platform-specific API to communicate with
+    For example, a client may use a platform-specific mechanism to communicate with
     an authenticator which is physically bound do a platform.  On the other
     hand, a client may use a variety of transports such as Bluetooth
     (see [[#transport]] to discover and communicate with authenticators.
@@ -672,12 +672,14 @@ authorizing an authenticator with which to complete the operation.
     a |platform| attachment, while those that are reachable via standardized
     cross-platform protocols are defined as having |cross_platform| attachment.
     This distinction is important because some [RPS] are known to have
-    use-cases where only a |platform| attached authenticator is acceptable 
-    (or use-cases where only |cross_platform| attachment may be used).  As
-    concrete example for the former, [RPS] that offer accounts whereby a
-    webauthn authenticator may be used as the only valid second factor will not
-    want to allow users to use platform-attached authenticators -- as this will
-    lead to user lockout.
+    use-cases where only a |platform| attached authenticator is acceptable (or
+    use-cases where only |cross_platform| attachment may be used).  As a
+    concrete example of the former, a credential on a platform authenticator
+    may be used by [RPS] to quickly and conveniently reauthentication the user
+    with a minimum of friction.  As a concrete example of the latter, when the
+    user is accessing the [RP] from a given client for the first time, they may
+    be required to use a cross-platform authenticator which was originally set
+    up on a different client.
 
     <ul>
         <li><dfn>platform</dfn> - the respective authenticator is attached
--- a/index.bs
+++ b/index.bs
@@ -427,13 +427,16 @@ When this method is invoked, the user agent MUST execute the following algorithm
 
 9. Initialize |issuedRequests| to an empty list.
 
-10. For each authenticator currently available on this platform: asynchronously invoke the <a>authenticatorMakeCredential</a>
+10. If {{ScopedCredentialOptions/attachment}} is specified, only consider authenticators which meet the attachment criteria for
+    executing further steps.
+
+11. For each available authenticator: asynchronously invoke the <a>authenticatorMakeCredential</a>
     operation on that authenticator with |rpIdHash|, <a>clientDataHash</a>, {{accountInformation}}, |normalizedParameters|,
     {{ScopedCredentialOptions/excludeList}} and |clientExtensions| as parameters. Add a corresponding entry to |issuedRequests|.
     - For each credential C in {{ScopedCredentialOptions/excludeList}} that has a non-empty |transports| list, optionally use
         only the specified transports to test for the existence of C.
 
-11. While |issuedRequests| is not empty, perform the following actions depending upon the |adjustedTimeout| timer and responses
+12. While |issuedRequests| is not empty, perform the following actions depending upon the |adjustedTimeout| timer and responses
     from the authenticators:
     - If the |adjustedTimeout| timer expires, then for each entry in |issuedRequests| invoke the <a>authenticatorCancel</a>
         operation on that authenticator and remove its entry from the list.
@@ -620,8 +623,9 @@ authorizing an authenticator with which to complete the operation.
     dictionary ScopedCredentialOptions {
         unsigned long                             timeoutSeconds;
         USVString                                 rpId;
-        sequence < ScopedCredentialDescriptor >  excludeList;
+        sequence < ScopedCredentialDescriptor >   excludeList;
         WebAuthnExtensions                        extensions;
+        Attachment                                attachment;
     };
 </pre>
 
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -637,8 +641,47 @@ authorizing an authenticator with which to complete the operation.
...
+<pre class="idl">
+    enum Attachment {
+        "platform",
+        "cross_platform"
+    };
+</pre>
+
+<div dfn-for="Attachment">
+    Clients may communicate with authenticators using a variety of mechanisms.
+    For example, a client may use a platform-specific API to communicate with
```
vijaybh: Nit: "platform-specific mechanism" may be better than "platform-specific API" here.

index.bs
```diff
@@ -637,8 +641,47 @@ authorizing an authenticator with which to complete the operation.
...
+    We define authenticators that are part of the client's platform as having
+    a |platform| attachment, while those that are reachable via standardized
+    cross-platform protocols are defined as having |cross_platform| attachment.
+    This distinction is important because some [RPS] are known to have
+    use-cases where only a |platform| attached authenticator is acceptable 
+    (or use-cases where only |cross_platform| attachment may be used).  As
+    concrete example for the former, [RPS] that offer accounts whereby a
+    webauthn authenticator may be used as the only valid second factor will not
+    want to allow users to use platform-attached authenticators -- as this will
+    lead to user lockout.
```
vijaybh: Two problems here. First, the example purports to present a use case for the "former" (i.e. requiring platform authenticator only) and ends up giving a use case for the opposite. Second, it seems like we could find a better example - maybe something like "As a concrete example of the former, a credential on a platform authenticator may be used by [RPS] to quickly and conveniently reauthentication the user with a minimum of friction." Alternatively, "As a concrete example of the latter, when the user is accessing the [RP] from a given client for the first time, they may be required to use a cross-platform authenticator which was originally set up on a different client."

index.bs
leshi: done

index.bs
```diff
@@ -637,8 +641,49 @@ authorizing an authenticator with which to complete the operation.
...
+    enum Attachment {
+        "platform",
+        "cross_platform"
+    };
+</pre>
+
+<div dfn-for="Attachment">
+    Clients may communicate with authenticators using a variety of mechanisms.
+    For example, a client may use a platform-specific mechanism to communicate with
+    an authenticator which is physically bound do a platform.  On the other
```
equalsJeffH: s/do/to/

index.bs
```diff
@@ -637,8 +641,49 @@ authorizing an authenticator with which to complete the operation.
...
+    hand, a client may use a variety of transports such as Bluetooth
```
equalsJeffH: s/transports/cross-platform transport protocols/

index.bs
```diff
@@ -637,8 +641,49 @@ authorizing an authenticator with which to complete the operation.
...
+    (see [[#transport]] to discover and communicate with authenticators.
+    We define authenticators that are part of the client's platform as having
+    a |platform| attachment, while those that are reachable via standardized
+    cross-platform protocols are defined as having |cross_platform| attachment.
```
equalsJeffH: s/protocols/transport protocols/

index.bs
```diff
@@ -637,8 +641,49 @@ authorizing an authenticator with which to complete the operation.
...
+    This distinction is important because some [RPS] are known to have
+    use-cases where only a |platform| attached authenticator is acceptable (or
+    use-cases where only |cross_platform| attachment may be used).  As a
```
equalsJeffH: wordsmith of L667-L669: 

This distinction is important because there are use-cases where only  |platform| attached authenticators are acceptable to a [RP], and conversely ones where only |cross_platform| attached authenticators are employed.

index.bs
```diff
@@ -637,8 +641,49 @@ authorizing an authenticator with which to complete the operation.
...
+    concrete example of the former, a credential on a platform authenticator
+    may be used by [RPS] to quickly and conveniently reauthentication the user
```
equalsJeffH: s/reauthentication/reauthenticate/

equalsJeffH:
I made some editorial line-comments.

wrt terminology, `platform` and `cross_platform` refer to  authenticator (authnr) "attachment" modalities which us technologists understand and need when reasoning about under-the-hood-details.  

I can see where the term "platform attached authnr" or "platform authnr" might work amongst users, but have concerns with "cross-platform authnr" being confusing. 

Perhaps we can define a "roaming authenticator" as one being one with a "cross-platform attachment" modality. The latter term may be more meaningful to users, while the definition wrt attachment modality is meaningful to readers of the spec.

index.bs
leshi: Done

index.bs
leshi: Done

index.bs
```diff
@@ -644,8 +648,50 @@ authorizing an authenticator with which to complete the operation.
         the credential, or that additional information be returned in the attestation statement. Alternatively, the caller may
         specify an additional message that they would like the authenticator to display to the user. Extensions are defined in
         [[#extensions]].
+
+    - The <dfn>attachment</dfn> parameter contains additional constraints on how authenticators may be attached to the
+        client platform.  See [[#attachment]] for a description of the valid attachment values and their meanings.
```
equalsJeffH: suggest delete "valid"

index.bs
```diff
@@ -427,13 +427,16 @@ When this method is invoked, the user agent MUST execute the following algorithm
 
 9. Initialize |issuedRequests| to an empty list.
 
-10. For each authenticator currently available on this platform: asynchronously invoke the <a>authenticatorMakeCredential</a>
+10. If {{ScopedCredentialOptions/attachment}} is specified, only consider authenticators which meet the attachment criteria for
+    executing further steps.
```
leshi: Done. Clarified.

index.bs
```diff
@@ -649,8 +652,10 @@ authorizing an authenticator with which to complete the operation.
         specify an additional message that they would like the authenticator to display to the user. Extensions are defined in
         [[#extensions]].
 
-    - The <dfn>attachment</dfn> parameter contains additional constraints on how authenticators may be attached to the
-        client platform.  See [[#attachment]] for a description of the valid attachment values and their meanings.
+    - The <dfn>attachment</dfn> parameter contains authenticator attachment descriptions, which is used as an additional
```
equalsJeffH: s/is used/are used/

equalsJeffH:
thx @rlin1 for fixing the closing div :)

index.bs
```diff
@@ -427,8 +427,11 @@ When this method is invoked, the user agent MUST execute the following algorithm
 
 9. Initialize |issuedRequests| to an empty list.
 
-10. If {{ScopedCredentialOptions/attachment}} is specified, only consider authenticators which meet the attachment criteria for
-    executing further steps.
+10. If {{ScopedCredentialOptions/attachment}} is specified, let |attachment| be
```
leshi: sg

leshi:
Okay, I think all comments are addressed.  Waiting on final LGTM, then will merge.

Can I get a final LGTM from someone?

vijaybh:
LGTM

index.bs
```diff
@@ -644,8 +651,52 @@ authorizing an authenticator with which to complete the operation.
...
+</pre>
+
+<div dfn-for="Attachment">
+    Clients may communicate with authenticators using a variety of mechanisms.
+    For example, a client may use a platform-specific API to communicate
+    with an authenticator which is physically bound to a platform.  On the
+    other hand, a client may use a variety of cross-platform transport
+    protocols such as Bluetooth (see [[#transport]] to discover and communicate
+    with authenticators.  We define authenticators that are part of the
+    client's platform as having a |platform| attachment, while those that are
```
equalsJeffH: whoops, making this suggestion on the incorrect sentence :(

jcjones: I agree this is easier to understand.

index.bs
```diff
@@ -644,8 +651,52 @@ authorizing an authenticator with which to complete the operation.
...
+    reachable via standardized cross-platform transport protocols are defined
+    as having |cross_platform| attachment.  This distinction is important
```
jcjones: I'll leave it to others to bikeshed about `roaming` vs other forms, but I think having such a term is more natural to my mind than "cross-platform transport", which kind of implies changing OS rather than changing device.

index.bs
```diff
@@ -644,8 +651,52 @@ authorizing an authenticator with which to complete the operation.
...
+    because there are use-cases where only |platform| attached authenticators
```
jcjones: I'd keep the definition as |platform| authenticators here, but 👍

index.bs
```diff
@@ -644,8 +651,52 @@ authorizing an authenticator with which to complete the operation.
...
+    are acceptable to a [RP], and conversely ones where only |cross_platform|
+    attached authenticators are employed.  As a concrete example of the former,
```
jcjones: If we take the above, then yes, but since it's a defined term then perhaps it should be `|roaming authenticators|`?

index.bs
```diff
@@ -644,8 +651,52 @@ authorizing an authenticator with which to complete the operation.
...
+    a credential on a platform authenticator may be used by [RPS] to quickly
```
jcjones: We need a `<dfn>` for `platform authenticator` too, in this case, and this should link to it.

index.bs
```diff
@@ -644,8 +651,52 @@ authorizing an authenticator with which to complete the operation.
...
+    and conveniently reauthenticate the user with a minimum of friction.  As
+    a concrete example of the latter, when the user is accessing the [RP] from
+    a given client for the first time, they may be required to use a
+    cross-platform authenticator which was originally set up on a different
+    client.
+
+    <ul>
```
jcjones: I've no opinion about pulling this or not.

Issue #102, opened 2016-05-13
Add section describing verification of a WebAuthnAssertion


rlin1:
We have described a recommended verification for attestation and we should add such recommendation for the signature assertion.

rlin1:
see branch rolf-verify-assertion

PR #246, opened 2016-11-04
section verification of WebAuthnAssertion added. See #102

rlin1:
Please cross-check and merge.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1180,7 +1180,8 @@ Upon receiving a signature assertion in the form of a {{WebAuthnAssertion}} stru
 
 3. Verify that the {{ClientData/origin}} in the {{ClientData}} matches the [RP]'s origin.
 
-4. Verify that the {{ClientData/tokenBinding}} in the {{ClientData}} matches the token binding public key for the TLS connection
+4. Verify that the {{ClientData/tokenBinding}} (if present) in the {{ClientData}} matches the token 
+    binding public key for the TLS connection
     over which the attestation was obtained.
 
 5. Verify that the {{ClientData/extensions}} in the {{ClientData}} is a proper subset of the extensions requested by the RP.
@@ -1190,11 +1191,11 @@ Upon receiving a signature assertion in the form of a {{WebAuthnAssertion}} stru
 
 7. Compute the <a>clientDataHash</a>, i.e. hash of {{WebAuthnAssertion/clientData}}.
 
-8. Look up the public key that was returned by {{makeCredential()}} and is related to {{WebAuthnAssertion/credential}} and
+8. Look up the previously registered public key associated with the credential (see {{makeCredential()}}) and
     verify the signature in {{WebAuthnAssertion/signature}} computed over the
     binary concatenation of {{WebAuthnAssertion/authenticatorData}} and <a>clientDataHash</a>.
 
-The Relying Party shall reject the request if the verification fails.
+If all the above steps succeed, then the signature is valid, otherwise it is invalid.
 
 ## Credential Attestation Statements ## {#cred-attestation-stmts}
 
--- a/index.bs
+++ b/index.bs
@@ -891,7 +891,7 @@ these methods.
         <li><dfn>ble</dfn> - the respective Authenticator may be contacted over
             Bluetooth Smart (Bluetooth Low Energy / BLE).
     </ul>
-<div>
+</div>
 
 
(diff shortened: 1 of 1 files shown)


rlin1:
as discussed in our last call.

index.bs
```diff
@@ -1168,6 +1168,33 @@ The authenticator MUST return both the <a>authenticatorData</a> and the raw sign
...
+6. Verify that the RP ID hash in the {{WebAuthnAssertion/authenticatorData}} is indeed the SHA-256 hash of the RP ID expected
+    by the RP.
+
+7. Compute the <a>clientDataHash</a>, i.e. hash of {{WebAuthnAssertion/clientData}}.
+
+8. Look up the public key that was returned by {{makeCredential()}} and is related to {{WebAuthnAssertion/credential}} and
+    verify the signature in {{WebAuthnAssertion/signature}} computed over the
+    binary concatenation of {{WebAuthnAssertion/authenticatorData}} and <a>clientDataHash</a>.
+
+The Relying Party shall reject the request if the verification fails.
```
vijaybh: I would stick to the verification aspects here. The RP will do what it will do.

So I'd suggest replacing this with something like "if all the above steps succeed, then the signature is valid, otherwise not".

index.bs
```diff
@@ -1168,6 +1168,33 @@ The authenticator MUST return both the <a>authenticatorData</a> and the raw sign
...
+
+Upon receiving a signature assertion in the form of a {{WebAuthnAssertion}} structure, the [RP] shall:
+
+1. Perform JSON decoding to extract the {{ClientData}} used for the assertion from the {{WebAuthnAssertion/clientData}}.
+
+2. Verify that the {{ClientData/challenge}} in the {{ClientData}} matches the challenge that was sent to the authenticator.
+
+3. Verify that the {{ClientData/origin}} in the {{ClientData}} matches the [RP]'s origin.
+
+4. Verify that the {{ClientData/tokenBinding}} in the {{ClientData}} matches the token binding public key for the TLS connection
```
vijaybh: Add "if present" here since token binding is optional.

vijaybh:
Thanks for writing this up. I think this is important enough we should try and get something into the spec, and we can polish it as we go. I added some small comments to clarify the wording.

index.bs
```diff
@@ -1168,6 +1168,33 @@ The authenticator MUST return both the <a>authenticatorData</a> and the raw sign
 MUST return <a>clientDataJSON</a>, <a>authenticatorData</a> and the signature to the RP. The first two are returned in the
 `clientData` and `authenticatorData` members respectively of the {{WebAuthnAssertion}} and {{WebAuthnAttestation}} structures.
 
+### Verifying a signature ### {#authenticator-signature-verification}
+
+This section specifies the algorithm for verifying a signature assertion.
+
+Upon receiving a signature assertion in the form of a {{WebAuthnAssertion}} structure, the [RP] shall:
+
+1. Perform JSON decoding to extract the {{ClientData}} used for the assertion from the {{WebAuthnAssertion/clientData}}.
```
equalsJeffH: hm, if WebAuthnAssertion.clientData contains JSON-serialized data, perhaps we should rename it serializedClientData

index.bs
rlin1: OK. See my latest commit.

index.bs
```diff
@@ -1168,6 +1168,33 @@ The authenticator MUST return both the <a>authenticatorData</a> and the raw sign
...
+    over which the attestation was obtained.
+
+5. Verify that the {{ClientData/extensions}} in the {{ClientData}} is a proper subset of the extensions requested by the RP.
+
+6. Verify that the RP ID hash in the {{WebAuthnAssertion/authenticatorData}} is indeed the SHA-256 hash of the RP ID expected
+    by the RP.
+
+7. Compute the <a>clientDataHash</a>, i.e. hash of {{WebAuthnAssertion/clientData}}.
+
+8. Look up the public key that was returned by {{makeCredential()}} and is related to {{WebAuthnAssertion/credential}} and
```
rlin1: OK, see my latest commit

index.bs
rlin1: see my latest commit

index.bs
equalsJeffH: yes we could treat it that way.  otherwise LGTM

PR #307, opened 2016-12-10
Specify encoding of rpid to bytes before hashing. 

### LINKED ISSUE #244: Simplifying attestation, take two
Some colleagues and I have spent a fair bit of time with attestation recently, thinking through implementation issues. It became apparent to us that a browser implementation has to do significant parsing of the attestation data returned from the authenticator even though the client really should not care about the contents of this.

Strawman suggestion that we came up with:
- Reduce WebAuthnAttestation to { ArrayBuffer clientData; ArrayBuffer attestation; }
- Have attestation be a CBOR map containing
    - Format
    - authenticatorData
    - format-specific content (different for packed vs. TPM and so on)
- Perhaps this will become obsolete if we implement the above, but any place we insert the algorithm of the attestation signature, we could remove that field for the case of self attestation

The above CBOR map would not be parsed at all on the client, only at the server. This might also simplify the IDL a little bit.

### LINKED ISSUE #258: Need to clearly define what it means to SHA-256 hash the "host" output of the document.domain setter
The output is a Unicode string.  SHA-256 takes bytes as input.  The exact conversion of Unicode to bytes here needs to be defined.

This can be sort of guessed at, because I _think_ the actual string involved will always be in the ASCII range.  But I'm not 100% sure, and it would be best to eliminate the guessing.

kpaulh:
Specify the encoding of rpid to be a UTF-8 before doing any hashing.
Fixes #258 .


equalsJeffH:
kpaulh -- thx for submitting this -- some thoughts/notes on this PR:

1.  the kpaulh github account (Kimberly Paulhamus?) is perhaps not anointed as a "member" of the WebAuthn WG ?  this might be causing the issue with the IPR check.

2.  the actual substantive changes in this PR are in lines 403 and 499.  All the other changes appear to be elimination of line-end blank chars. 

3.  the commit message (incorrectly) says it closes issue #244 rather than #258..

4.  ..also, while the content of the changes to lines 403 and 499 seem nominally correct given how the algorithm steps are presently written, they will need to be re-written when we address various of the issues submitted by @bzbarsky.  as a part of that, we will likely invoke https://encoding.spec.whatwg.org/#utf-8-encode as suggested by @annevk https://github.com/w3c/webauthn/issues/258#issuecomment-258775438 -- so perhaps we shouldn't close #258 with this PR as it is presently written.

kpaulh:
IPR check - Yeah, that's what we finally figured out. I've been nominated but still waiting for confirmation.
Oops, will update the issue reference. Don't know how that happened.
@equalsJeffH, @jcjones Thanks!

annevk:
You should probably reference https://encoding.spec.whatwg.org/#utf-8-encode to make sure it's actually defined.

jcjones:
Handling @annevk's comment in a follow-on commit, https://github.com/jcjones/webauthn/commit/af03460eccd5c913d887ecc10c94076ccde88bef

Issue #308, opened 2016-12-13
authenticator attachment terminology


equalsJeffH:
wrt terminology, platform and cross_platform refer to authenticator (authnr) "attachment" modalities which us technologists understand and need when reasoning about under-the-hood-details.

I can see where the term "platform attached authenticator" or "platform authenticator" might work amongst users, but have concerns with the term "cross-platform authenticator" being confusing to users.

Perhaps we can define a "roaming authenticator" as being one with a "cross-platform attachment" modality. The latter term may be more meaningful to users, while the definition wrt attachment modality is meaningful to readers of the spec.

(this originally appeared as a comment on #243 here https://github.com/w3c/webauthn/pull/243#issuecomment-259476527)

equalsJeffH:
This was fixed by merging of https://github.com/w3c/webauthn/pull/309, closing manually.

PR #306, opened 2016-12-07
Jeffh editorial cleanups4

equalsJeffH:
Please do not merge as yet -- will continue to add to this as we work on WD-04

PR #309, opened 2016-12-13
authenticator attachment terminology refinement

equalsJeffH:
I noted the following in this PR comment https://github.com/w3c/webauthn/pull/243#issuecomment-259476527 back on 9-Nov-2016, but it did not garner discussion..
<blockquote>
wrt terminology, platform and cross_platform refer to authenticator (authnr) "attachment" modalities which us technologists understand and need when reasoning about under-the-hood-details.

I can see where the term "platform attached authnr" or "platform authnr" might work amongst users, but have concerns with the term "cross-platform authnr" being confusing to users.

Perhaps we can define a "roaming authenticator" as being one with a "cross-platform attachment" modality. The latter term may be more meaningful to users, while the definition wrt attachment modality is meaningful to readers of the spec.
</blockquote>

..prior to PR #243 being merged last Fri. I had thought I'd followed up prior to the merge but had messed up (the `tl;dr` is below). In any case, this PR embodies those enhancements and is further polished. 

This PR constitutes the approach suggested by @jcjones in https://lists.w3.org/Archives/Public/public-webauthn/2016Dec/0074.html, and is in reply to issue #308

`tl;dr`:

I _had_ crafted [specific line comments](https://github.com/w3c/webauthn/pull/243#pullrequestreview-11243597) (last Wed 7-Dec-2016) that build on PR #246 and addressed my above comments. HOWEVER, it turns out no one had yet seen the latter comments on PR #243 because I had messed up -- I had made those detailed comments as "review comments" (rather than "line comments") and had not yet submitted the review :(

[ also note that at this time "reviews", "review-comments", "review closed" events are not forwarded to the mailing list :(  ]


jcjones:
This change in text matches what I was thinking, and I think it's a great clarification as to what these attachment options really mean.

It does require a rebase, but this will probably land after #306, which will probably require a rebase there, too.

equalsJeffH:
thanks @rlin1 for doing the merge-from-master :)

index.bs
```diff
@@ -673,38 +675,35 @@ authorizing an authenticator with which to complete the operation.
...
-    a credential on a platform authenticator may be used by [RPS] to quickly
-    and conveniently reauthenticate the user with a minimum of friction.  As
-    a concrete example of the latter, when the user is accessing the [RP] from
-    a given client for the first time, they may be required to use a
-    cross-platform authenticator which was originally set up on a different
-    client.
+    Clients may communicate with authenticators using a variety of mechanisms. For example, a client may use a platform-specific
+    API to communicate with an authenticator which is physically bound to a platform. On the other hand, a client may use a
+    variety of standardized cross-platform transport protocols such as Bluetooth (see [[#transport]]) to discover and
+    communicate with <a>cross-platform attached</a> authenticators. We define authenticators that are part of the client's
```
leshi: I don't know what a "cross-platform attached" authenticator is.  I don't understand why "cross-platform" was not clear here.

leshi:
We were waiting for me to take a look at this.  I've taken a look and it looks fine.  Merging it in.

equalsJeffH:
thanks @leshi !

Issue #256, opened 2016-11-04
Please coordinate with the HTML spec to extract the relevant bits of the document.domain setter so you can call them


bzbarsky:
Currently https://w3c.github.io/webauthn/#dom-webauthentication-makecredential step 3 says to do https://www.w3.org/TR/html51/browsers.html#relaxing-the-same-origin-restriction but not quite.  Except it doesn't make it clear _which_ document, and it's not clear whether you actually want the interaction with sandboxing that the document.domain setter has, whether you actually want the behavior to be affected by previous document.domain sets (as the actual document.domain setter is) and so forth.

All this stuff should be clearly defined, ideally by:

1. Defining clearly which document you're working with.
2. Working together with the HTML spec to extract the relevant algorithm for your use.  If this proves impossible, because you actually want somewhat different behavior from the document.domain setter, then the algorithm needs to just be clearly defined.

bzbarsky:
This applies to https://w3c.github.io/webauthn/#getAssertion step 3 too.

equalsJeffH:
Yes, the present spec language for this is a stab-in-the-dark-with-the-hope-of-sparking-helpful-feedback, and it seems to have worked :)

In terms of (2), i.e. the relevant algorithm, it seems to me it is _nominally_ steps 4 - 7 (inclusive) of the domain attribute setter defined hereabouts: https://html.spec.whatwg.org/#dom-document-domain

So, by "extract the algorithm" you mean to say "revise the HTML spec such that the those steps are defined as a distinct algorithm which we an then invoke from the webauthn spec (and which the document.domain setter also utilizes) ?

bzbarsky:
> nominally steps 4 - 7 (inclusive) of the domain attribute setter defined hereabouts

OK.  So if a page is loaded from "https://foo.bar.com" and then sets document.domain to "bar.com", should it be able to use "foo.bar.com" as its rpId?  Sounds to me like you're saying it shouldn't be able to....

> So, by "extract the algorithm" you mean to say

Yes.  Again, assuming the desired behavior is similar enough that this makes sense.

equalsJeffH:
>> nominally steps 4 - 7 (inclusive) of the domain attribute setter defined hereabouts

> OK.  So if a page is loaded from "https://foo.bar.com" and then sets document.domain to "bar.com", should it be able to use "foo.bar.com" as its rpId?  Sounds to me like you're saying it shouldn't be able to....

[...]

Issue #271, opened 2016-11-04
callerOrigin isn't actually the origin of the caller; it's the origin of the callee


bzbarsky:
callerOrigin in https://w3c.github.io/webauthn/#dom-webauthentication-makecredential step 3 is defined as the "current settings origin".  That means it's the origin of the callee function, which may or may not match the origin of the caller.  And either one of those two might or might not match the origin of the WebAuthentication object itself, of course.

At the very least, this should be named more clearly.

bzbarsky:
This applies to https://w3c.github.io/webauthn/#getAssertion as well.

AngeloKai:
When you mention the origins of the callee function and the caller function may not match, are you talking about the cases, such as the situation where a site calls into its iframe's makeCredential function?

bzbarsky:
That would be a simple example, yes.

equalsJeffH:
> callerOrigin in https://w3c.github.io/webauthn/#dom-webauthentication-makecredential step 3 is defined as the "current settings origin". That means it's the origin of the callee function, which may or may not match the origin of the caller. And either one of those two might or might not match the origin of the WebAuthentication object itself, of course.

OK, I have been looking into this and have a question for @bzbarsky and @domenic:  in our particular case of crafting the `makeCredential()` and `getAssertion()` algorithms, is this the (or "a") correct way to obtain the "caller's origin":
* let |callerOrigin| be the [incumbent settings object](https://html.spec.whatwg.org/#incumbent-settings-object)'s origin.

..?

The overall apparently applicable section of [HTML](https://html.spec.whatwg.org/) is:  https://html.spec.whatwg.org/#realms-settings-objects-global-objects  -- yes?

domenic:
In general as noted in that section of the spec, the incumbent concept should not be used by new specifications going forward.

Can you explain what you are actually trying to accomplish here? In general any behavior that depends on the callstack is surprising and not something we want to include in the platform.

equalsJeffH:
FWIW, I note that HTML has a goal of "[Minimize the usage of the incumbent concept](https://github.com/whatwg/html/issues/1430)". However, I have not otherwise been able to discern another straightforward way to obtain one's caller's origin.

domenic:
Yes; the idea is to minimize the number of places one ever checks the caller's anything (especially origin). Why does this spec want to introduce a new way of doing so? It's generally a very confusing programming model for developers with some security drawbacks as well.

equalsJeffH:
[...]

Issue #257, opened 2016-11-04
The W3C HTML spec is broken, and probably shouldn't be referenced


bzbarsky:
I was pretty confused by https://www.w3.org/TR/html51/browsers.html#relaxing-the-same-origin-restriction step 7.2, as referenced from https://w3c.github.io/webauthn/#dom-webauthentication-makecredential step 3, because it's completely nonsensical: it causes the algorithm to _always_ throw.

Then I realized that this is the W3C version of the HTML spec, which has all sorts of known issues like this.  For comparison, https://html.spec.whatwg.org/multipage/browsers.html#relaxing-the-same-origin-restriction step 7.2 is correct...

Please reference the things that actually have the behavior you want.  I guess if you have to reference W3C HTML, that makes it your problem to check that it has the behavior you want and push on its editors to fix it if it does not.

bzbarsky:
Oh, and in particular, browser implementors are, as far as I know, implementing from WHATWG HTML, not W3C HTML.  As a result, if you reference the latter there is a good chance that your spec won't match what actually gets implemented, if they instead substitute the behavior of the former...  Worse yet would be if different implementors handled this differently and ended up with incompatible implementations.

equalsJeffH:
see also:
RE: referencing W3C or WhatWG specs (was: [webauthn] new commits pushed by equalsJeffH
https://lists.w3.org/Archives/Public/public-webauthn/2016Sep/0313.html

bzbarsky:
It's up to you (the WG working on webauthn) if you want to reference W3C HTML, as I said.  But then you will have to do due diligence on whether it actually says what you want it to say, because it has a history of all sorts of bugs that get introduced during the forking process.

In the end, of course, either the tests that will presumably be written to test webauthn will catch issues, or the tests won't reflect what the spec actually says (and UAs that try to implement what the spec does say will discover the hard way that it doesn't mean what it says).  It's going to be a giant waste of everyone's time either way, of course...

AngeloKai:
I filed a [issue](https://github.com/w3c/html/issues/730) on the W3C HTML spec repo. @bzbarsky thank you for catching it!

bzbarsky:
Fwiw, I had done that too: https://github.com/w3c/html/issues/705

The real problem is that they keep reintroducing these sorts of issues...

equalsJeffH:
[...]

Issue #255, opened 2016-11-04
Setting _rpId_ to _callerOrigin_ doesn't make sense


bzbarsky:
_rpId_ is set to a string by the "rpId is specified" branch.  But an origin is a tuple (since we already returned an error for opaque origins).  I really doubt the following parts of the algorithm can handle a non-string _rpId_, and in fact the very next bit that says "the SHA-256 hash of _rpId_" already doesn't handle it.

Presumably this is meant to assign a serialization of the origin to _rpId_.  The spec needs to be clear whether this means assigning https://www.w3.org/TR/html51/browsers.html#unicode-serialization or https://www.w3.org/TR/html51/browsers.html#ascii-serialization-of-an-origin

Once you decide that, you need to specify what it means to take an SHA-256 hash of _rpId_.  The serialization of an origin (either one, really, but this is especially bad for the not-guaranteed-ASCII one) is a Unicode string, whereas SHA-256 works on a byte buffer.  There is no canonical encoding of Unicode into bytes, so the exact encoding used here needs to be specified.

bzbarsky:
This applies to both makeCredential and getAssertion.

selfissued:
We should specify UTF-8 as the encoding of Unicode into bytes used in this (and other cases) in the specification.

selfissued:
@equalsJeffH plans to fix this.

PR #314, opened 2016-12-20
Inline and clarify the "Relaxing the Same-Origin Restriction" algorithm

jcjones:
Issue #256 notes that it is "not clear whether [we] actually want the interaction with sandboxing that the document.domain setter has", nor "whether [we] actually want the behavior to be affected by previous `document.domain` sets".

This patch offers a way to fix that, by:

1) Extracting the procedure from HTML51 into a forked algorithm
2) Adjusting said algorithm to operate on a Document's "original Domain" so as to be independent of previous `document.domain` set operations.
3) Keeping the sandboxing interactions, though I'm not entirely versed in  whether there are problematic corner cases here.

This spec necessarily adds normative reference to the PSL (which was transitively referenced via normative reference from HTML51 before), and also to the URL specification (also previously transitive from HTML51).

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -48,23 +48,24 @@ spec: HTML51; urlPrefix: http://www.w3.org/TR/html51/; for: web
             text: tuple origin
             text: relaxing the same-origin restriction
             text: domain; url: origin-domain
+            text: active sandboxing flag set
     type: dfn
         urlPrefix: webappapis.html;
             text: current settings object; for:web; url:current-settings-object
             text: Navigator; for: interface; url:the-navigator-object
 
-spec: WebCryptoAPI; urlPrefix: https://www.w3.org/TR/WebCryptoAPI/; for: web 
+spec: WebCryptoAPI; urlPrefix: https://www.w3.org/TR/WebCryptoAPI/; for: web
     type: dfn
         text: normalizing an algorithm; url: dfn-normalize-an-algorithm
 
-<!-- spec: HTML; urlPrefix: https://url.spec.whatwg.org/ -->
 spec: URL; urlPrefix: https://url.spec.whatwg.org/; for: url
     type: dfn
-        text: host parser; url: concept-host-parser
+        text: host parser
         text: domain; url: concept-domain
-        text: host; url: origin-host
-        text: IPv4 address; url: concept-ipv4
-        text: IPv6 address; url: concept-ipv6
+
+spec: WebIDL; urlPrefix: https://heycam.github.io/webidl/; for: webidl
+    type: dfn
+        text: securityerror
 </pre> <!-- class=anchors -->
 
 # Introduction # {#intro}
@@ -417,7 +418,7 @@ When this method is invoked, the user agent MUST execute the following algorithm
     - If the {{ScopedCredentialOptions/rpId}} member of {{options}} is <a>present</a>, then invoke the [[#algo-relax-same-origin-restriction]],
         using {{ScopedCredentialOptions/rpId}}. If no errors are thrown, set |rpId| to the value of `host` as
         computed by this algorithm, and |rpIdHash| to the SHA-256 hash of the UTF-8 encoding of |rpId|. Otherwise, reject
-        |promise| with a <a>DOMException</a> whose name is "SecurityError", and terminate this algorithm.
+        |promise| with a <a>DOMException</a> whose name is "<a>SecurityError</a>", and terminate this algorithm.
 
 4. Process each element of {{cryptoParameters}} using the following steps, to produce a new sequence |normalizedParameters|.
     - Let |current| be the currently selected element of {{cryptoParameters}}.
@@ -511,7 +512,7 @@ When this method is invoked, the user agent MUST execute the following algorithm
     - If the {{AssertionOptions/rpId}} member of {{options}} is <a>present</a>, then invoke the [[#algo-relax-same-origin-restriction]], using
         {{AssertionOptions/rpId}} as the given value. If no errors are thrown, set |rpId| to the value of `host` as computed
         by this algorithm, and |rpIdHash| to the SHA-256 hash of the UTF-8 encoding of |rpId|. Otherwise, reject |promise| with
-        a <a>DOMException</a> whose name is "SecurityError", and terminate this algorithm.
+        a <a>DOMException</a> whose name is "<a>SecurityError</a>", and terminate this algorithm.
 
 4. If the {{AssertionOptions/extensions}} member of {{options}} is <a>present</a>, process any extensions supported by this 
     client platform, to produce the extension data that needs to be sent to the authenticator. If an error is encountered while
@@ -1523,36 +1524,35 @@ should be specified in the attestation certificate itself, so that it can be ver
 
 ## Algorithm to Relax the Same-Origin Restriction ## {#algo-relax-same-origin-restriction}
 
-The same-origin policy prevents web applications from interacting, unless they both have the same origin. HTML5, via
-the `document.domain` attribute, permits "relaxing" this same-origin restriction under certain circumstances. This
-algorithm implements a similar procedure that of setting the `document.domain` attribute from [[!HTML51]], but without
-modifying the <a>Document</a> object:
+In general, the same-origin policy prevents web applications from interacting
+within user agents, unless they have the same origin. However, there are many
+specific facets and exceptions to this overall policy.
 
-1. If this <a>Document</a> object has no browsing context, throw a "SecurityError" <a>DOMException</a>.
-2. If this <a>Document</a> object’s active sandboxing flag set has its <a>sandboxed `document.domain` browsing context
-    flag</a> set, then throw a "SecurityError" <a>DOMException</a>.
-3. If the given value is the empty string, then throw a "SecurityError" <a>DOMException</a>.
-4. Let |host| be the result of parsing the given value.
-5. If |host| is failure, then throw a "SecurityError" <a>DOMException</a>.
-6. Let |originalDomain| be this <a>Document</a> object’s <a link-for='web'>origin</a>’s original <a link-for='web'>`domain`</a>,
-    assigned when the <a>Document</a> is created.
+For example, cookies [[RFC6265]] have same-origin restrictions similar to those
+of the HTML `document.domain` attribute. Both permit <a>relaxing the same-origin
+restriction</a> in a particular fashion, under certain circumstances.
 
-    Note: The |originalDomain| should not be affected by prior calls to the <a>Document</a> object’s
-     <a link-for='web'>`domain`</a> setter.
+The below algorithm implements a same-origin relaxation in the same vein as
+cookies and `document.domain`, but without modifying the <a>Document</a> object.
 
-7. If |host| is not <a link-for='url'>equal</a> to |originalDomain|, then run these substeps:
-    1. If |host| or |originalDomain| is not a <a link-for='url'>`domain`</a>, then throw a "SecurityError" <a>DOMException</a>.
+1. If the given value is the empty string, then throw a "<a>SecurityError</a>" <a>DOMException</a>.
+2. Let |host| be the result of <a>parsing</a> the given value.
+3. If |host| is failure, then throw a "<a>SecurityError</a>" <a>DOMException</a>.
+4. If this <a>Document</a> object’s <a link-for='web'>origin</a> is an <a link-for='web'>opaque origin</a>, then throw a "<a>SecurityError</a>" <a>DOMException</a>.
+5. Let |originalHost| be this <a>Document</a> object’s <a link-for='web'>origin</a>’s <a link-for='url'>`host`</a>.
+6. If |host| is not <a link-for='url'>equal</a> to |originalHost|, then run these substeps:
+    1. If |host| or |originalHost| is not a <a link-for='url'>domain</a>, then throw a "<a>SecurityError</a>" <a>DOMException</a>.
 
-    Note: This is meant to exclude <a link-for='url'>hosts</a> that are an <a link-for='url'>IPv4 address</a> or an
-     <a link-for='url'>IPv6 address</a>.
+    Note: This is meant to exclude <a link-for='url'>hosts</a> that are an <a>IPv4 address</a> or an
+     <a>IPv6 address</a>.
 
-    2. If |host|, prefixed by a U+002E FULL STOP (.), does not exactly match the end of |originalDomain|, then throw a
-        "SecurityError" <a>DOMException</a>.
+    2. If |host|, prefixed by a U+002E FULL STOP (.), does not exactly match the end of |originalHost|, then throw a
+        "<a>SecurityError</a>" <a>DOMException</a>.
     3. If |host| matches a suffix in the Public Suffix List, or, if |host|, prefixed by a U+002E FULL STOP (.), matches the
-        end of a suffix in the Public Suffix List, then throw a "SecurityError" <a>DOMException</a>. [[!PSL]]
+        end of a suffix in the Public Suffix List, then throw a "<a>SecurityError</a>" <a>DOMException</a>. [[!PSL]]
 
         Suffixes must be compared after applying the <a link-for='url'>host parser</a> algorithm.
-8. Return |host|.
+7. Return |host|.
 
 # Defined Attestation Formats # {#defined-attestation-formats}
 
--- a/index.bs
+++ b/index.bs
@@ -412,10 +412,10 @@ When this method is invoked, the user agent MUST execute the following algorithm
     terminate this algorithm. Otherwise,
     - If the {{ScopedCredentialOptions/rpId}} member of {{options}} is not <a>present</a>, then set |rpId| to |callerOrigin|, 
         and |rpIdHash| to the SHA-256 hash of the UTF-8 encoding of |rpId|.
-    - If the {{ScopedCredentialOptions/rpId}} member of {{options}} is <a>present</a>, then invoke the [[#algo-relax-same-origin-restriction]],
-        using {{ScopedCredentialOptions/rpId}}. If no errors are thrown, set |rpId| to the value of `host` as
+    - If the {{ScopedCredentialOptions/rpId}} member of {{options}} is <a>present</a>, then invoke the [[#algo-relax-same-origin-restriction]]
+        using these arugments: let |relaxingRpId| be {{ScopedCredentialOptions/rpId}}, and let |currentDocument| be the current <a>Document</a>. If no errors are thrown, set |rpId| to the value of `host` as
         computed by this algorithm, and |rpIdHash| to the SHA-256 hash of the UTF-8 encoding of |rpId|. Otherwise, reject
-        |promise| with a <a>DOMException</a> whose name is "{{SecurityError}}", and terminate this algorithm.
+        |promise| with the thrown error, and terminate this algorithm.
 
 4. Process each element of {{cryptoParameters}} using the following steps, to produce a new sequence |normalizedParameters|.
     - Let |current| be the currently selected element of {{cryptoParameters}}.
@@ -506,10 +506,10 @@ When this method is invoked, the user agent MUST execute the following algorithm
     terminate this algorithm. Otherwise,
     - If the {{AssertionOptions/rpId}} member of {{options}} is not <a>present</a>, then set |rpId| to |callerOrigin|, and 
         |rpIdHash| to the SHA-256 hash of the UTF-8 encoding of |rpId|.
-    - If the {{AssertionOptions/rpId}} member of {{options}} is <a>present</a>, then invoke the [[#algo-relax-same-origin-restriction]], using
-        {{AssertionOptions/rpId}} as the given value. If no errors are thrown, set |rpId| to the value of `host` as computed
-        by this algorithm, and |rpIdHash| to the SHA-256 hash of the UTF-8 encoding of |rpId|. Otherwise, reject |promise| with
-        a <a>DOMException</a> whose name is "{{SecurityError}}", and terminate this algorithm.
+    - If the {{AssertionOptions/rpId}} member of {{options}} is <a>present</a>, then invoke the [[#algo-relax-same-origin-restriction]],
+        using these arugments: let |relaxingRpId| be {{ScopedCredentialOptions/rpId}}, and let |currentDocument| be the current <a>Document</a>. If no errors are thrown, set |rpId| to the value of `host` as computed
+        by this algorithm, and |rpIdHash| to the SHA-256 hash of the UTF-8 encoding of |rpId|. Otherwise, reject
+        |promise| with the thrown error, and terminate this algorithm.
 
 4. If the {{AssertionOptions/extensions}} member of {{options}} is <a>present</a>, process any extensions supported by this 
     client platform, to produce the extension data that needs to be sent to the authenticator. If an error is encountered while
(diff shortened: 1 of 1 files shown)


equalsJeffH:
One aside to explicitly note here: referencing whatwg's URL spec intersects with, and raises the spector of, our spec-referencing issue #257.

index.bs
```diff
@@ -1508,7 +1516,38 @@ used to help facilitate isolating problems with a specific version of a device.
 If the attestation root certificate is not dedicated to a single WebAuthn Authenticator device line (i.e., AAGUID), the AAGUID
 should be specified in the attestation certificate itself, so that it can be verified against the <a>authenticatorData</a>.
 
+## Algorithm to Relax the Same-Origin Restriction ## {#algo-relax-same-origin-restriction}
+
+The same-origin policy prevents web applications from interacting, unless they both have the same origin. HTML5, via
+the `document.domain` attribute, permits "relaxing" this same-origin restriction under certain circumstances. This
+algorithm implements a similar procedure that of setting the `document.domain` attribute from [[!HTML51]], but without
+modifying the <a>Document</a> object:
+
```
equalsJeffH: suggested build on the above 4 lines 1521-1524:

###
In general, the same-origin policy prevents web applications from interacting within user agents, unless they have the same origin. However, there are many specific facets and exceptions to this overall policy. 

For example, cookies [[RFC6265]] have same-origin restrictions similar to those of the HTML `document.domain` attribute. Both permit relaxing the same-origin restriction in a particular fashion, under certain circumstances. 

The below algorithm implements a same-origin relaxation in the same vein as cookies and `document.domain`, but without modifying the <a>Document</a> object.
###

[ for reference, see: http://identitymeme.org/http-cookie-processing-algorithm-etlds/ ]

index.bs
```diff
@@ -1508,7 +1516,38 @@ used to help facilitate isolating problems with a specific version of a device.
...
+1. If this <a>Document</a> object has no browsing context, throw a "SecurityError" <a>DOMException</a>.
+2. If this <a>Document</a> object’s active sandboxing flag set has its <a>sandboxed `document.domain` browsing context
```
equalsJeffH: xref this:  `<a>active sandboxing flag set</a>`

index.bs
```diff
@@ -1508,7 +1516,38 @@ used to help facilitate isolating problems with a specific version of a device.
...
+    flag</a> set, then throw a "SecurityError" <a>DOMException</a>.
+3. If the given value is the empty string, then throw a "SecurityError" <a>DOMException</a>.
+4. Let |host| be the result of parsing the given value.
```
equalsJeffH: xref this: `a>parsing</a>`

index.bs
```diff
@@ -1508,7 +1516,38 @@ used to help facilitate isolating problems with a specific version of a device.
...
+5. If |host| is failure, then throw a "SecurityError" <a>DOMException</a>.
+6. Let |originalDomain| be this <a>Document</a> object’s <a link-for='web'>origin</a>’s original <a link-for='web'>`domain`</a>,
+    assigned when the <a>Document</a> is created.
```
equalsJeffH: I seem to recall there being an actual means in the whatwg html spec to get to a document's "original origin", will have to look for it...

index.bs
```diff
@@ -1508,7 +1516,38 @@ used to help facilitate isolating problems with a specific version of a device.
...
+
+    Note: The |originalDomain| should not be affected by prior calls to the <a>Document</a> object’s
+     <a link-for='web'>`domain`</a> setter.
```
equalsJeffH: s/setter/attribute's setter/

index.bs
```diff
@@ -1508,7 +1516,38 @@ used to help facilitate isolating problems with a specific version of a device.
...
+the `document.domain` attribute, permits "relaxing" this same-origin restriction under certain circumstances. This
+algorithm implements a similar procedure that of setting the `document.domain` attribute from [[!HTML51]], but without
+modifying the <a>Document</a> object:
+
+1. If this <a>Document</a> object has no browsing context, throw a "SecurityError" <a>DOMException</a>.
+2. If this <a>Document</a> object’s active sandboxing flag set has its <a>sandboxed `document.domain` browsing context
+    flag</a> set, then throw a "SecurityError" <a>DOMException</a>.
+3. If the given value is the empty string, then throw a "SecurityError" <a>DOMException</a>.
+4. Let |host| be the result of parsing the given value.
+5. If |host| is failure, then throw a "SecurityError" <a>DOMException</a>.
```
equalsJeffH: I think bikeshed will xref to webidl if we do `"<a>SecurityError</a>"` for the dom exceptions...

index.bs
```diff
@@ -1508,7 +1516,38 @@ used to help facilitate isolating problems with a specific version of a device.
 If the attestation root certificate is not dedicated to a single WebAuthn Authenticator device line (i.e., AAGUID), the AAGUID
 should be specified in the attestation certificate itself, so that it can be verified against the <a>authenticatorData</a>.
 
+## Algorithm to Relax the Same-Origin Restriction ## {#algo-relax-same-origin-restriction}
```
equalsJeffH: I suggest moving this, from here in (the present) section 5 {#authenticator-model}, to (the bottom of?) section 4 {#api}.  it is not properly part of the authnr model AIUI.

index.bs
```diff
@@ -1508,7 +1516,38 @@ used to help facilitate isolating problems with a specific version of a device.
...
+
+The same-origin policy prevents web applications from interacting, unless they both have the same origin. HTML5, via
+the `document.domain` attribute, permits "relaxing" this same-origin restriction under certain circumstances. This
+algorithm implements a similar procedure that of setting the `document.domain` attribute from [[!HTML51]], but without
+modifying the <a>Document</a> object:
+
+1. If this <a>Document</a> object has no browsing context, throw a "SecurityError" <a>DOMException</a>.
```
equalsJeffH: xref this: `<a>browsing context</a>`  ?

index.bs
equalsJeffH: Also, regarding steps 1 and 2 -- I'm not sure we need them for this algorithm since this does not directly involve setting document.domain. Perhaps the algorithm can start with the present step 3 ?

index.bs
domenic: So we have the creation URL for environment settings objects (<-> Windows), but not necessarily for documents... which is a bit tricky since they're not always 1:1 (e.g. `document.implement.creationDocument()`, or the initial transition from `about:blank` to a new `Document` within the same `Window`).

I think what you want is to simply get the origin's **host**. That will not change ever. See https://html.spec.whatwg.org/#concept-origin-tuple and some of the stuff following. In particular step 3 of the effective domain algorithm is what gets returned if `document.domain` is never messed with.

index.bs
```diff
@@ -1508,7 +1516,38 @@ used to help facilitate isolating problems with a specific version of a device.
...
+4. Let |host| be the result of parsing the given value.
+5. If |host| is failure, then throw a "SecurityError" <a>DOMException</a>.
+6. Let |originalDomain| be this <a>Document</a> object’s <a link-for='web'>origin</a>’s original <a link-for='web'>`domain`</a>,
+    assigned when the <a>Document</a> is created.
+
+    Note: The |originalDomain| should not be affected by prior calls to the <a>Document</a> object’s
+     <a link-for='web'>`domain`</a> setter.
 
+7. If |host| is not <a link-for='url'>equal</a> to |originalDomain|, then run these substeps:
+    1. If |host| or |originalDomain| is not a <a link-for='url'>`domain`</a>, then throw a "SecurityError" <a>DOMException</a>.
```
domenic: "domain" should not be backticked

index.bs
domenic: This needs to be `<a for="Document">browsing context</a>`, not just `<a>browsing context</a>

index.bs
```diff
@@ -47,10 +47,20 @@ spec: HTML51; urlPrefix: http://www.w3.org/TR/html51/; for: web
...
             text: tuple origin
             text: relaxing the same-origin restriction
+            text: domain; url: origin-domain
     type: dfn
         urlPrefix: webappapis.html;
             text: current settings object; for:web; url:current-settings-object
             text: Navigator; for: interface; url:the-navigator-object
+
+<!-- spec: HTML; urlPrefix: https://url.spec.whatwg.org/ -->
+spec: URL; urlPrefix: https://url.spec.whatwg.org/; for: url
```
domenic: Most of these anchors should not be necessary. (Both here and for HTML, at least if you reference HTML instead of HTML51.)

In general it's best to minimize the number of anchors you have; it implies either the original spec is not correctly exporting its terms, or you're not using enough metadata to tell Bikeshed what's what (e.g. for="" attributes).

index.bs
jcjones: It didn't, but I added it into the anchors so it'd xref.

index.bs
jcjones: Thanks, I gave this a shot in eb69af86dd54a88e59356c8b489cfa64e9ed346a; see what you think!

index.bs
jcjones: I think you're right, so I pulled them out. I added a step to explicitly check for opaque origins, though, so the numbering is only off by 1.

index.bs
jcjones: Done in 72ba450a674c57f48f584513eca69c51d74a21ff

index.bs
jcjones: I gave it a shot to remove several that we didn't really need (e.g., host, ipv4/ipv6 address...), but I may need to come back to it in another commit.

index.bs
```diff
@@ -47,17 +47,26 @@ spec: HTML51; urlPrefix: http://www.w3.org/TR/html51/; for: web
...
     type: dfn
         text: normalizing an algorithm; url: dfn-normalize-an-algorithm
 
-</pre> <!-- class=anchors -->
+spec: URL; urlPrefix: https://url.spec.whatwg.org/; for: url
+    type: dfn
+        text: host parser
+        text: domain; url: concept-domain
 
+spec: WebIDL; urlPrefix: https://heycam.github.io/webidl/; for: webidl
```
domenic: This can be removed if you use `{{SecurityError}}` instead of `<a>SecurityError</a>`

index.bs
```diff
@@ -967,6 +973,39 @@ A string or dictionary identifying a cryptographic algorithm and optionally a se
...
+restriction</a> in a particular fashion, under certain circumstances.
+
+The below algorithm implements a same-origin relaxation in the same vein as
+cookies and `document.domain`, but without modifying the <a>Document</a> object.
+
+1. If the given value is the empty string, then throw a "<a>SecurityError</a>" <a>DOMException</a>.
+2. Let |host| be the result of <a>parsing</a> the given value.
+3. If |host| is failure, then throw a "<a>SecurityError</a>" <a>DOMException</a>.
+4. If this <a>Document</a> object’s <a link-for='web'>origin</a> is an <a link-for='web'>opaque origin</a>, then throw a "<a>SecurityError</a>" <a>DOMException</a>.
+5. Let |originalHost| be this <a>Document</a> object’s <a link-for='web'>origin</a>’s <a link-for='url'>`host`</a>.
```
domenic: No backticks around host

index.bs
```diff
@@ -967,6 +968,39 @@ A string or dictionary identifying a cryptographic algorithm and optionally a se
...
+restriction</a> in a particular fashion, under certain circumstances.
+
+The below algorithm implements a same-origin relaxation in the same vein as
+cookies and `document.domain`, but without modifying the <a>Document</a> object.
+
+1. If the given value is the empty string, then throw a "{{SecurityError}}" <a>DOMException</a>.
+2. Let |host| be the result of <a>parsing</a> the given value.
+3. If |host| is failure, then throw a "{{SecurityError}}" <a>DOMException</a>.
+4. If this <a>Document</a> object’s <a link-for='web'>origin</a> is an <a link-for='web'>opaque origin</a>, then throw a "{{SecurityError}}" <a>DOMException</a>.
+5. Let |originalHost| be this <a>Document</a> object’s <a link-for='web'>origin</a>’s <a link-for='url'>host</a>.
```
equalsJeffH: Step 5 is crucial, would be good to have review from @bzbarsky and @domenic.

equalsJeffH: Also, reviewers should note that this PR does not address issues #253 #254 #255 #271 etc.

index.bs
```diff
@@ -405,13 +411,11 @@ When this method is invoked, the user agent MUST execute the following algorithm
...
-        <a>relaxing the same-origin
-        restriction</a> by setting the `document.domain` attribute, using {{ScopedCredentialOptions/rpId}} as the given value
-        but without changing the current document's `domain`. If no errors are thrown, set |rpId| to the value of `host` as
-        computed by this procedure, and |rpIdHash| to the SHA-256 hash of the UTF-8 encoding of |rpId|. Otherwise, reject |promise| with a
-        <a>DOMException</a> whose name is "SecurityError", and terminate this algorithm.
+        and |rpIdHash| to the SHA-256 hash of the UTF-8 encoding of |rpId|.
+    - If the {{ScopedCredentialOptions/rpId}} member of {{options}} is <a>present</a>, then invoke the [[#algo-relax-same-origin-restriction]],
+        using {{ScopedCredentialOptions/rpId}}. If no errors are thrown, set |rpId| to the value of `host` as
+        computed by this algorithm, and |rpIdHash| to the SHA-256 hash of the UTF-8 encoding of |rpId|. Otherwise, reject
+        |promise| with a <a>DOMException</a> whose name is "{{SecurityError}}", and terminate this algorithm.
```
bzbarsky: So this does _not_ propagate out the errors from the relax-same-origin-restriction algorithm?

In that case, why have that throw at all instead of just returning a fail/success boolean?  Or if the idea is to just propagate out the errors from the relax-... algorithm, why not do that?

index.bs
```diff
@@ -967,6 +970,39 @@ A string or dictionary identifying a cryptographic algorithm and optionally a se
 defined in [[!WebCryptoAPI]].
 
 
+## Algorithm to Relax the Same-Origin Restriction ## {#algo-relax-same-origin-restriction}
+
+In general, the same-origin policy prevents web applications from interacting
+within user agents, unless they have the same origin. However, there are many
+specific facets and exceptions to this overall policy.
```
bzbarsky: This entire paragraph is completely pointless, in my opinion.

index.bs
```diff
@@ -967,6 +970,39 @@ A string or dictionary identifying a cryptographic algorithm and optionally a se
...
+
+For example, cookies [[RFC6265]] have same-origin restrictions similar to those
+of the HTML `document.domain` attribute. Both permit <a>relaxing the same-origin
+restriction</a> in a particular fashion, under certain circumstances.
```
bzbarsky: And this one should probably say something more like:

> We want to allow an rpid to represent a parent domain, subject to constraints similar to those of cookies or `document.domain`.

or something like that.

index.bs
```diff
@@ -967,6 +970,39 @@ A string or dictionary identifying a cryptographic algorithm and optionally a se
...
+
+The below algorithm implements a same-origin relaxation in the same vein as
+cookies and `document.domain`, but without modifying the <a>Document</a> object.
+
+1. If the given value is the empty string, then throw a "{{SecurityError}}" <a>DOMException</a>.
+2. Let |host| be the result of <a link-for='url'>parsing</a> the given value.
+3. If |host| is failure, then throw a "{{SecurityError}}" <a>DOMException</a>.
+4. If this <a>Document</a> object’s <a link-for='web'>origin</a> is an <a link-for='web'>opaque origin</a>, then throw a "{{SecurityError}}" <a>DOMException</a>.
```
bzbarsky: Which "this Document"?  There is no Document passed to this algorithm, and the caller doesn't make it clear which Document is the one involved (current, incumbent, relevant, entry, something else?).

I suggest having this algorithm explicitly take two arguments: a Document and a string.  And then having callers explicitly say what they pass to it.

index.bs
```diff
@@ -967,6 +970,39 @@ A string or dictionary identifying a cryptographic algorithm and optionally a se
...
+
+For example, cookies [[RFC6265]] have same-origin restrictions similar to those
+of the HTML `document.domain` attribute. Both permit <a>relaxing the same-origin
+restriction</a> in a particular fashion, under certain circumstances.
+
+The below algorithm implements a same-origin relaxation in the same vein as
+cookies and `document.domain`, but without modifying the <a>Document</a> object.
+
+1. If the given value is the empty string, then throw a "{{SecurityError}}" <a>DOMException</a>.
+2. Let |host| be the result of <a link-for='url'>parsing</a> the given value.
```
bzbarsky: Hmm.  This ends up linking to https://url.spec.whatwg.org/#concept-host-parser because of the bits added at the top, I assume?

index.bs
equalsJeffH: > So this does not propagate out the errors from the relax-same-origin-restriction algorithm?

A learning-curve oversight, thanks for catching it. 

How about this as a rewrite of the last sentence from above:
> Otherwise, reject |promise| with any <a>DOMException</a> thrown by [[#algo-relax-same-origin-restriction]] and terminate these steps.

..?

index.bs
equalsJeffH: personally, I think this spec will have readers that are not so clued-in to web security and the same origin policy, thus I'm inclined to provide a modicum of context (tho the above could certainly be polished).

index.bs
equalsJeffH: Ok, thx for suggestion.

index.bs
equalsJeffH: yes, is that not correct? it is the same #concept-host-parser as referenced from ["the domain attribute's setter" alg](https://html.spec.whatwg.org/#dom-document-domain).

index.bs
bzbarsky: Well... is this part happening before or after the algorithm goes async?  Presumably it should happen before (existing issues on that), and then you can just rethrow any exceptions and they will get turned into rejected promises for you, right?

index.bs
bzbarsky: I think if you want to keep this, it's best to put it this way:

> In general, the same-origin policy prevents pages from different origins from interacting with each other.  However, for payments we want to allow subdomains to create payment requests that look like they come from the main domain.

or so.  That is, explain what problem you're really trying to solve.  The general "there are some exceptions" verbiage mostly confuses people (including me) as to where we're heading with this.

index.bs
bzbarsky: That's correct, yes.  I just wanted to make sure the end result links to the right place, not to some other part of the URL spec.  All is good here.

index.bs
equalsJeffH: > the caller doesn't make it clear which Document is the one involved (current, incumbent, relevant, entry, something else?).

doh.

> I suggest having this algorithm explicitly take two arguments: a Document and a string. And then having callers explicitly say what they pass to it.

sounds good to me.

index.bs
equalsJeffH: ok, though, all the algs I see in [HTML](https://html.spec.whatwg.org/) that implement a promise say "...then reject the promise with an "<someError>" DOMException, and abort these steps."  And it seems all the occurrences of "Rethrow any exceptions" are in method algs. 

in any case, if it is legit to do so, you're suggesting that that last sentence could be simply..
> Otherwise, rethrow any exceptions. 

..yes?   (does that imply "..and terminate these steps" ?)

index.bs
equalsJeffH: Ok, thx.

index.bs
bzbarsky: The "reject the promise and abort" is the right thing to do in the async part of an algorithm.  Aborting in the _sync_ part is nonsense, because you have to either throw or return something.

> does that imply "..and terminate these steps" 

Throwing (from the sync part of an algorithm) does mean not running any more steps.  Throwing from an async part is nonsense.

index.bs
jcjones: I've given this a shot in f796e9e8d228f99c5377dc317a9a72491242858c but I'm not very confident in my result. Suggestions are welcome!

index.bs
jcjones: I tried the simplest option here: 

https://github.com/w3c/webauthn/pull/314/commits/f796e9e8d228f99c5377dc317a9a72491242858c#diff-ec9cfa5f3f35ec1f84feb2e59686c34dR418

index.bs
```diff
@@ -972,22 +972,22 @@ defined in [[!WebCryptoAPI]].
 
 ## Algorithm to Relax the Same-Origin Restriction ## {#algo-relax-same-origin-restriction}
 
-In general, the same-origin policy prevents web applications from interacting
-within user agents, unless they have the same origin. However, there are many
-specific facets and exceptions to this overall policy.
+In general, the same-origin policy prevents pages from different origins from
+interacting with each other. However, for payments we want to allow subdomains
+to create payment requests that look like they come from the main domain.
```
equalsJeffH: I don't know where the "payments" term came from and do not think we should use that term here.  I would delete that sentence since the sentence below essentially covers that base.

index.bs
```diff
@@ -972,22 +972,22 @@ defined in [[!WebCryptoAPI]].
...
 
-For example, cookies [[RFC6265]] have same-origin restrictions similar to those
-of the HTML `document.domain` attribute. Both permit <a>relaxing the same-origin
-restriction</a> in a particular fashion, under certain circumstances.
+We want to allow an |rpId| to represent a parent domain, subject to constraints
+similar to those of cookies or `document.domain`.
```
equalsJeffH: suggest: s/want/need/

index.bs
```diff
@@ -412,10 +412,10 @@ When this method is invoked, the user agent MUST execute the following algorithm
     terminate this algorithm. Otherwise,
     - If the {{ScopedCredentialOptions/rpId}} member of {{options}} is not <a>present</a>, then set |rpId| to |callerOrigin|, 
         and |rpIdHash| to the SHA-256 hash of the UTF-8 encoding of |rpId|.
-    - If the {{ScopedCredentialOptions/rpId}} member of {{options}} is <a>present</a>, then invoke the [[#algo-relax-same-origin-restriction]],
-        using {{ScopedCredentialOptions/rpId}}. If no errors are thrown, set |rpId| to the value of `host` as
+    - If the {{ScopedCredentialOptions/rpId}} member of {{options}} is <a>present</a>, then invoke the [[#algo-relax-same-origin-restriction]]
+        using these arugments: let |relaxingRpId| be {{ScopedCredentialOptions/rpId}}, and let |currentDocument| be the current <a>Document</a>. If no errors are thrown, set |rpId| to the value of `host` as
```
equalsJeffH: is this the style the html et al spec uses to invoke an alg with multiple args?  If not, I'd just copy their extant style (sorry, don't have time now to investigate myself)

jcjones: Not exactly, but my wordsmithing juice was low. I was following https://html.spec.whatwg.org/#the-application-cache-selection-algorithm
>When the application cache selection algorithm algorithm is invoked with a Document document and optionally a manifest URL manifest URL, the user agent must run the first applicable set of steps from the following list:
 
and https://html.spec.whatwg.org/#navigate:
>Certain actions cause the browsing context to navigate to a new resource. A user agent may provide various ways for the user to explicitly cause a browsing context to navigate, in addition to those defined in this specification.
but couldn't quite make the wording make sense as it was.



Suggestions welcome.

jcjones:
Note to all reviewers: Given criticism on the call Wednesday that maybe we should avoid inlining thie algorithm, I've also opened an alternative PR #319 which, rather than *inlining* the algorithm, clarifies how to "call it." (Please see that PR for further caveats). That should be considered if we don't choose to proceed with this PR.

index.bs
```diff
@@ -405,13 +411,11 @@ When this method is invoked, the user agent MUST execute the following algorithm
...
-        and |rpIdHash| to the SHA-256 hash of |rpId|.
-    - If the {{ScopedCredentialOptions/rpId}} member of {{options}} is <a>present</a>, then invoke the procedure used for 
-        <a>relaxing the same-origin
-        restriction</a> by setting the `document.domain` attribute, using {{ScopedCredentialOptions/rpId}} as the given value
-        but without changing the current document's `domain`. If no errors are thrown, set |rpId| to the value of `host` as
-        computed by this procedure, and |rpIdHash| to the SHA-256 hash of the UTF-8 encoding of |rpId|. Otherwise, reject |promise| with a
-        <a>DOMException</a> whose name is "SecurityError", and terminate this algorithm.
+        and |rpIdHash| to the SHA-256 hash of the UTF-8 encoding of |rpId|.
+    - If the {{ScopedCredentialOptions/rpId}} member of {{options}} is <a>present</a>, then invoke the [[#algo-relax-same-origin-restriction]]
+        using these arugments: let |relaxingRpId| be {{ScopedCredentialOptions/rpId}}, and let |currentDocument| be the current <a>Document</a>. If no errors are thrown, set |rpId| to the value of `host` as
```
AngeloKai: Current document is not a well-defined term neither the W3C nor the WHATWG spec. One of the reason why issue [256](https://github.com/w3c/webauthn/issues/256) was raised was because the HTML5 spec definition doesn't specify which document object to interact with. 

[Active document](https://html.spec.whatwg.org/multipage/browsers.html#windows) is a possible replacement for "current document". However, to define specifically which document the origin is supposed to derive, you will need to also specify which browsing context you want. That may inevitably require us to rely on the caller vs. callee programming model ([271](https://github.com/w3c/webauthn/issues/271)) which is something we try to avoid.

One way to avoid this hassle would be to restrict the API to be only available on top level browsing context.

domenic: Is there a clear object whose method is being called when this algorithm runs? If so, the object's relevant settings object's responsible document is the clear choice. (Might have gotten the exact spec terms wrong; on mobile at the moment. But basically the object's global's document.)

I don't think there's any caller vs. callee debate to speak of. Callee is correct in all cases.

index.bs
```diff
@@ -47,18 +47,24 @@ spec: HTML51; urlPrefix: http://www.w3.org/TR/html51/; for: web
...
+spec: WebCryptoAPI; urlPrefix: https://www.w3.org/TR/WebCryptoAPI/; for: web
     type: dfn
         text: normalizing an algorithm; url: dfn-normalize-an-algorithm
 
+spec: URL; urlPrefix: https://url.spec.whatwg.org/; for: url
+    type: dfn
+        url: concept-host-parser
+            text: parsing
+            text: host parser
+        text: domain; url: concept-domain
```
annevk: You shouldn't need to do this. All these terms are exported.

index.bs
```diff
@@ -967,6 +970,39 @@ A string or dictionary identifying a cryptographic algorithm and optionally a se
 defined in [[!WebCryptoAPI]].
 
 
+## Algorithm to Relax the Same-Origin Restriction ## {#algo-relax-same-origin-restriction}
```
annevk: You're not relaxing it. You're bypassing it as far as I can tell.

index.bs
```diff
@@ -967,6 +970,39 @@ A string or dictionary identifying a cryptographic algorithm and optionally a se
...
+We want to allow an |rpId| to represent a parent domain, subject to constraints
+similar to those of cookies or `document.domain`.
+
+The below algorithm should be invoked with a <a>Document</a> |currentDocument| and a
+string |relaxingRpId|. It implements a same-origin relaxation in the same vein as
+cookies and `document.domain`, but without modifying the |currentDocument| argument.
+
+1. If |relaxingRpId| is the empty string, then throw a "{{SecurityError}}" <a>DOMException</a>.
+2. Let |host| be the result of <a link-for='url'>parsing</a> the |relaxingRpId|.
+3. If |host| is failure, then throw a "{{SecurityError}}" <a>DOMException</a>.
```
annevk: Why not check for host being a domain here? Saves you checks later on.

jcjones:
Closing in favor of PR #319, which references the refactored algorithm at whatwg/html.

PR #315, opened 2016-12-20
Reference the Encoding spec's UTF-8 Encode definition

jcjones:
Per https://github.com/w3c/webauthn/pull/307#issuecomment-267266190

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -56,6 +56,10 @@ spec: WebCryptoAPI; urlPrefix: https://www.w3.org/TR/WebCryptoAPI/; for: web
     type: dfn
         text: normalizing an algorithm; url: dfn-normalize-an-algorithm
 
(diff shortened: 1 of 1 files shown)


equalsJeffH:
One aside to explicitly note here:  referencing whatwg's encoding spec intersects with, and raises the spector of, our spec-referencing issue #257.

equalsJeffH:
> One aside to explicitly note here: referencing whatwg's encoding spec intersects with, and raises the spector of, our spec-referencing issue #257.

Ah, there is now a W3C Encoding spec at CR maturity level -- https://www.w3.org/TR/encoding/ -- that we should reference per https://github.com/w3c/webauthn/issues/257#issuecomment-268655973 and https://github.com/w3c/webauthn/issues/257#issuecomment-268862182

jcjones:
Rebased, and moved to the CR-maturity Encoding spec.

equalsJeffH:
LGTM

Issue #311, opened 2016-12-16
Consider removing the "n" in "WebAuthnAttestation", etc


kpaulh:
Reason being that it's easy for developers to overlook and could be a general annoyance. The distinction between (authn) and (authz) is important, but wouldn't referencing the spec provide that clarification?

equalsJeffH:
I disagree -- people/developers/archtects/everyone confuse "authentication" and "authorization", and regard "auth" as either or both, often regardless of context.  Note also that the official short name of both the working group and the spec is "WebAuthn".   To me, the main issue is whether we spell out "authentication" in the webauthn api names or not, and to me that is mostly a matter of taste.

equalsJeffH:
see #288

nadalin:
I agree with Jeff

arshadnoor:
And, what about the confusion that's going to result from the fact that the "Authentication" protocol enables "transaction confirmation" aka "transaction authorization"?  How does the specification anticipate addressing that?  You do realize that attorneys are going to want clarification on whether "transaction confirmation" === "authorization".

nadalin:
This API is strictly about Web Authentication the implementations may or may not implement transaction confirmation, so that is out of scope

zcorpan:
I would suggest resolving this by removing "Web" from all API names and spelling out "Authentication" everywhere. 🙂

arshadnoor:
> This API is strictly about Web Authentication the implementations may or may not implement transaction confirmation, so that is out of scope

It seems to me that this is starting go down the well-trodden path of PKI.  

The authentication part of the spec is "mandatory", but even though [transaction authorization](https://www.w3.org/TR/2016/WD-webauthn-20160531/#extension-txauth) is part of the same spec, it is optional?  So, between Authenticators that can choose to implement what they want, servers that can choose to implement what they want and platforms/browsers in-between that can choose to implement what they want, WebAuthn Relying Parties are going to be in the same pickle that PKI RPs are in - different parts of the ecosystem that will **not** interoperate unless you buy/choose everything from the same or an extremely limited set of manufacturers.  So, what would be the point of the standard?  We had RFC 5280 as a standard for two decades; do you see many people using TLS ClientAuth - just _ClientAuth_ and not any of the extensions - to solve authentication problems?  (Arguing that hundreds of millions of Client digital certificates have been issued worldwide is not the right answer).

[...]

Issue #310, opened 2016-12-16
Can we get rid of the "Web" prefix for WebAuthnAttestation, etc?


kpaulh:
With respect to "WebAuthentication", "WebAuthnAttestation", "WebAuthnAssertion", etc. (not the overall name WebAuth). It seems to be unusual for Web APIs in general.

equalsJeffH:
related to #288

zcorpan:
How is this editorial? Renaming interfaces is a normative change.

equalsJeffH:
Well, we are considering it "editorial" at this time, amongst the spec editors, because the spec is only a working draft, and such a change is presently trivial to accomplish for us (e.g., it can presently be accomplished with `sed`).   We understand that from other perspectives, such as implementors, issues that we've labeled "editorial" may not necessarily be viewed as such.

equalsJeffH:
fixed by pr #317 https://github.com/w3c/webauthn/commit/6267cc1e81ae9999f11d75f37b4f9a3b558ce6dc

Issue #288, opened 2016-11-04
Is there a reason "WebAuthnAssertion" is not "WebAuthenticationAssertion"?


bzbarsky:
Is there a particular reason to try to save a few characters here?

bzbarsky:
Same for WebAuthnAttestation.

equalsJeffH:
fixed by pr #317 https://github.com/w3c/webauthn/commit/6267cc1e81ae9999f11d75f37b4f9a3b558ce6dc

PR #317, opened 2016-12-29
Renaming WebAuthn -> Authentication

equalsJeffH:
In terms of these issues -- #288 #310 #311 -- regarding our having prefixed some interfaces and dicts with "WebAuthn", @zcorpan suggests [in this comment](https://github.com/w3c/webauthn/issues/311#issuecomment-269247084) to..
> ..[resolve] this by removing "Web" from all API names and spelling out "Authentication" everywhere. 

So, I have crafted up an example of doing that and that is what is in this PR.  

Additionally, a rendered text diff of the current state of this branch, from master, is available here in pdf..

http://kingsmountain.com/doc/diff/diff-webauthn-jeffh-renaming-intfs-dicts-288-2016-12-27-fbec0ac--from--master-491ae9a.pdf

Thoughts?

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -74,10 +74,10 @@ the existence, of credentials scoped to other [RPS].
 [RPS] employ the <a>Web Authentication API</a> during two distinct, but related, <a>ceremonies</a> involving a user. The first
 is <a>Registration</a>, where a <a>scoped credential</a> is created on an <a>authenticator</a>, and associated by a <a>[RP]</a>
 with the present user's account (the account may already exist or may be created at this time). The second is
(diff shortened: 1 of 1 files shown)


leshi:
I support this color of the shed.

jcjones:
Please provide comments before 18 January, we'll resolve this one way or another then.

wseltzer:
No preference between WebAuthn and Authentication. (Authentication is better than removing the N)

equalsJeffH:
@wseltzer wrote:
> No preference between WebAuthn and Authentication. (Authentication is better than removing the N)

+1

AngeloKai:
The name change looks reasonable. I am ok with the change.

PR #319, opened 2017-01-06
Bug #256 - Clarify call of the "Relaxing the Same-Origin Restriction" algorithm

jcjones:
This is an **alternative** to PR #314 which inlines the "Relaxing the Same-Origin Restriction" algorithm within our document. Instead, this clarifies:

1) That the setter should be called on the current Document
2) That the active sandboxing flag set should be empty when the algorithm runs
3) Thrown errors should be used in the promise, not occluded

Note: This version **permits** earlier set operations to `document.domain` to affect the outcome of the algorithm, in contrast to PR #314 (at this time). This is mostly because the language is more natural here, but either PR could be changed to honor earlier set operations -- or not.


bzbarsky:
> That the active sandboxing flag set should be empty when the algorithm runs

That (changing sandboxing flags for a document on the fly) is a new primitive that I would really rather we didn't introduce, because it has pretty undesirable security properties.

jcjones:
That's fair, but are the sandboxing flags going to be sensible in this context? I've only added that line from earlier commentary that the removed them from PR #314 (https://github.com/w3c/webauthn/pull/314#discussion_r93720311) but if you're questioning, I'm only too happy to add that back in, @bzbarsky !

bzbarsky:
> That's fair, but are the sandboxing flags going to be sensible in this context?

Imo, no.  Which is why I think you should talk to the HTML spec editor about factoring out the parts of the algorithm that you need so you can call them properly.  Presumably with "sandboxing flags" as an argument, not ambient state.

jcjones:
I've updated this PR with commit 45063f6d2c68c0762b865aabf0daff361a398fab which moves to the nomenclature which currently looks like it will land in https://github.com/whatwg/html/pull/2365 .

index.bs
```diff
@@ -178,8 +177,8 @@ This specification relies on several other underlying specifications.
 
 : HTML
 :: The concepts of <a link-for='web'>current settings object</a>, <a link-for='web'>origin</a>,
-    <a link-for='web'>opaque origin</a>, <a>relaxing the same-origin restriction</a>, and the <a>Navigator</a> interface are
-    defined in [[!HTML51]].
+    <a link-for='web'>opaque origin</a>, <a link-for='web'>is a registrable domain suffix of or is equal
```
equalsJeffH: Also, the new `is a registrable domain suffix of or is equal to` is in whatwg/html, not HTML51.

btw, pls, what's the mistake in the `anchors` block?

equalsJeffH: fyi/fwiw, wrt "mistake in anchors block" mentioned by @jyasskin (which may have been mine): see #336

jcjones:
OK, this should be ready for re-review. I made a style call and un-nested "If the {{ScopedCredentialOptions/rpId}} member of {{options}} is not <a>present</a>, then ..." to be at the top-level of the algorithm, since it was getting wordy in that one bullet. I also, instead of renumbering, moved to the `1.` markdown auto-lists here. I can undo that if there's pushback.

jcjones:
I'll merge this at noon unless I hear otherwise. Special thanks to bzbarsky, jyasskin, annevk, and domenic for all the help in fixing this up.

index.bs
```diff
@@ -406,22 +405,26 @@ When this method is invoked, the user agent MUST execute the following algorithm
...
-        {{ScopedCredentialOptions/rpId}} as the given value but without changing the current document's `domain`. If no errors
-        are thrown, set |rpId| to the value of `host` as computed by this procedure. Otherwise, reject |promise| with a
-        {{DOMException}} whose name is "{{SecurityError}}", and terminate this algorithm.
-
-4. Process each element of {{cryptoParameters}} using the following steps, to produce a new sequence |normalizedParameters|.
+    terminate this algorithm.
+
+1. If the {{ScopedCredentialOptions/rpId}} member of {{options}} is not <a>present</a>, then set |rpId| to
+    |callerOrigin|. Otherwise,
+    - Let |effectiveDomain| be the |callerOrigin|'s <a>effective domain</a>.
```
jyasskin: I'd suggest using a numbered list for this second level too, since these steps also have to be executed in order.

index.bs
```diff
@@ -504,33 +507,37 @@ When this method is invoked, the user agent MUST execute the following algorithm
...
-    - If the {{AssertionOptions/rpId}} member of {{options}} is not <a>present</a>, then set |rpId| to |callerOrigin|.
-    - If the {{AssertionOptions/rpId}} member of {{options}} is <a>present</a>, then invoke the procedure used for <a>relaxing
-        the same-origin restriction</a> by setting the `document.domain` attribute, using {{AssertionOptions/rpId}} as the given
-        value but without changing the current document's `domain`. If no errors are thrown, set |rpId| to the value of `host`
-        as computed by this procedure. Otherwise, reject |promise| with a {{DOMException}} whose name is "{{SecurityError}}", and
+    terminate this algorithm.
+
+1. If the {{ScopedCredentialOptions/rpId}} member of {{options}} is not <a>present</a>, then set |rpId| to
+    |callerOrigin|. Otherwise,
+    - Let |effectiveDomain| be the |callerOrigin|'s <a>effective domain</a>.
```
jyasskin: Ditto here.

jcjones: Good thought. I'll fix both of those in another commit before I push.

PR #321, opened 2017-01-11
Clean up attestation, abstract it from UA, fix TPM format, add U2F format

### LINKED ISSUE #123: Clarify names for crypto algs, e.g. RS256
JSON Web Key defines RS256 which is equivalent to the Web Crypto defined "RSASSA-PKCS1-v1_5" (see http://www.w3c.org/TR/WebCryptoAPI/)

In Section 3.8.3 in web authentication spec, we reference the WebCrypto terminology, but in example two (and other places) we use RS256 - the JWK term.

### LINKED ISSUE #238: fix signature alg names and RFC3447 cite in {#generating-an-attestation-statement}
[ see also #123 ]

wrt S 5.3.3 {#generating-an-attestation-statement} and the notation therein of "RSA PKCS1" and "RSASSA-PSS": the former is incorrect/imprecise in terms of both WebCrypto [1] and RFC3447 [2]. The latter, "RSASSA-PSS", is used only in RFC3447.

Also, the phrase they are used within -- "Raw encoded RSA PKCS1 or RSASSA-PSS public key" -- is incorrect in that the former is an identifier for the RSA PKCS #1 spec series itself, the latter is the identifier for the RSASSA-PSS signature scheme, and neither are identifiers for ostensible subtypes of RSA keys. See Section 3 "Key Types" of [2].

Plus, RFC3447 is nigh to being obsoleted by [draft-moriarty-pkcs1](uhttps://tools.ietf.org/html/draft-moriarty-pkcs1rl)  which is approved and in RFC-editor's queue for pub as RFC -- the RFC3447 cite & biblio need updating.

[2] https://tools.ietf.org/html/rfc3447

vijaybh:
In addition to the bugs named in the commit messages, this also fixes #123 and fixes #238 by cleaning up the algorithm naming.


vijaybh:
@rlin1 - Looking at the latest changes, I wonder if we could put the CBOR field name back to just "attestation". This would be the only field with such a name, and it would save bytes on a structure that may be produced on an authenticator with limited capability. WDYT?

equalsJeffH:
@vijaybh wrote:
> Looking at the latest changes, I wonder if we could put the CBOR field name back to just "attestation". This would be the only field with such a name, and it would save bytes on a structure that may be produced on an authenticator with limited capability. WDYT?

@rlin1 replied in [email here](https://lists.w3.org/Archives/Public/public-webauthn/2017Jan/0078.html):
>First I think the spec is much better to read after those changes.
We call it attestation statement and the field containing it is also called
attestation statement.
Same with attestation object.
If the number of bytes is so important, we could abbreviate
attestationObject to "attstnObj" - that is even fewer bytes than the
original.

equalsJeffH:
@vijaybh wrote:
> Looking at the latest changes, I wonder if we could put the CBOR field name back to just "attestation". This would be the only field with such a name, and it would save bytes on a structure that may be produced on an authenticator with limited capability. WDYT?

@rlin1 replied in [email here](https://lists.w3.org/Archives/Public/public-webauthn/2017Jan/0078.html):
>If the number of bytes is so important, we could abbreviate
attestationObject to "attstnObj"

I agree with Rolf @rlin1 and was about to myself suggest something akin to `attstnObj` for the same reasons.

equalsJeffH:
a rendered textual diff of the spec in branch vgb-u2f-attestation at commit https://github.com/w3c/webauthn/pull/321/commits/0d0fcea44eb3e2a6a809e95bd892a0ff2e377a9b from master at commit 6267cc1 is here: http://kingsmountain.com/doc/diff/diff-webauthn-vgb-u2f-attestation-0d0fcea--from--master-6267cc1.pdf

equalsJeffH:
regarding these three prior comments -- https://github.com/w3c/webauthn/pull/321#issuecomment-271991391 and https://github.com/w3c/webauthn/pull/321#issuecomment-272210301 -- we ( @rlin1 and I ) and @vijaybh were referring to different things.  @vijaybh was referring to the now-named `attestation` member of the CBOR-encoded attestation object value, see https://github.com/w3c/webauthn/pull/321/commits/0d0fcea44eb3e2a6a809e95bd892a0ff2e377a9b.  I.e., nevermind regarding that particular comment of mine.

vijaybh:
I fiddled with the naming of fields inside the attestation structures to make them sort-of-consistent across formats. Do let me know if you have suggestions for improvement.

equalsJeffH:
New diffs:

just the diff of the 3 commits added today (from the original set of commits in this PR) 
http://kingsmountain.com/doc/diff/diff-webauthn-vgb-u2f-attestation-dc90eab--from--vgb-u2f-attestation-0d0fcea.pdf

diff of all commits presently in this PR from current state of master branch
http://kingsmountain.com/doc/diff/diff-webauthn-vgb-u2f-attestation-dc90eab--from--master-6267cc1.pdf

index.bs
```diff
@@ -1520,94 +1459,189 @@ If the attestation root certificate is not dedicated to a single WebAuthn Authen
...
-        3. "PS256" [[!RFC7518]]
-        4. "ED256" [[!FIDOEcdaaAlgorithm]]
-        5. "ED512" [[!FIDOEcdaaAlgorithm]]
+    : sig
+    :: A byte string containing the attestation signature.
 
-        The <dfn>signature</dfn> element contains the attestation signature.
-    </div>
+    : x5c
+    :: A definite-length array of byte strings. The elements of the array contain the attestation certificate and its
```
equalsJeffH: why does this need to be a definite-length?  a CBOR decoder ought to transparently be able to handle both definite- and indefinite-length arrays at run time, yes?  Note that CDDL does not make any prescription as to whether arrays or maps use the definite length or indefinite length encoding.  see: https://tools.ietf.org/html/draft-greevenbosch-appsawg-cbor-cddl#section-3.2

index.bs
```diff
@@ -1520,94 +1459,189 @@ If the attestation root certificate is not dedicated to a single WebAuthn Authen
...
+        specified in |alg|.
+    - Verify that |x5c| meets the requirements in [[#packed-attestation-cert-requirements]].
+    - If |x5c| contains an extension with OID `1 3 6 1 4 1 45724 1 1 4` (id-fido-gen-ce-aaguid) verify that
+        the value of this extension matches the AAGUID in the claimed <a>authenticatorData</a>.
+    - If successful, return attestation type Basic and trust path |x5c|.
+
+    If |daaKey| is present, then the attestation type is DAA. In this case:
+    - Verify that |alg| is "ED256" or "ED512".
+    - Follow the procedure in [[#authenticator-signature-verification]] to verify that |sig| is a valid signature over the given
+        <a>authenticatorData</a> and <a>clientDataHash</a>  using DAA-Verify with |daaKey| (see [[!FIDOEcdaaAlgorithm]]).
```
equalsJeffH: in this and all the other `verification procedure` subsections below, need to `s/clientDataHash/clientDataJSON/` because that is what is input to {{#authenticator-signature-verification}}

index.bs
```diff
@@ -177,13 +177,10 @@ This specification relies on several other underlying specifications.
...
     [[WebCryptoAPI#algorithm-dictionary]].
 
-
-
 : Base64url encoding
 :: The term <dfn>Base64url Encoding</dfn> refers to the base64 encoding using the URL- and filename-safe character set defined
     in Section 5 of [[!RFC4648]], with all trailing '=' characters omitted (as permitted by Section 3.2) and without the
-    inclusion of any line breaks, whitespace, or other additional characters. This is the same encoding as used by JSON Web
-    Signature (JWS) [[RFC7515]].
+    inclusion of any line breaks, whitespace, or other additional characters.
```
vijaybh: It just seems random. The only places we mention base64url are related to encoding a challenge and a token binding ID in ClientData. Neither of those is in any way related to RFC 7515 at first sight. So why did we pick this RFC to call out as opposed to any other standard that uses base64url?

index.bs
```diff
@@ -556,15 +556,16 @@ When this method is invoked, the user agent MUST execute the following algorithm
...
 
 
 ## Information about Scoped Credential (interface <dfn interface>ScopedCredentialInfo</dfn>) ## {#iface-credentialInfo}
 
 <pre class="idl">
     [SecureContext]
     interface ScopedCredentialInfo {
-        readonly attribute ScopedCredential           credential;
-        readonly attribute AuthenticationAttestation  attestation;
+        readonly    attribute ArrayBuffer   clientData;
```
vijaybh: Tried this out. Let me know if this reads better.

index.bs
```diff
@@ -573,12 +574,16 @@ authorizing an authenticator with which to complete the operation.
     to locate it later for use, and also contains metadata that can be used by the <a>[RP]</a> to assess the strength of the
     credential during registration.
 
-    The <dfn>credential</dfn> attribute contains an object whose attributes state the type of, and identifier for, the 
-    credential represented by {{ScopedCredentialInfo}}.
+    The <dfn>clientData</dfn> member contains the <a>clientDataJSON</a> (see [[#signature-format]]) passed to the authenticator
```
vijaybh: Replaced globally

index.bs
```diff
@@ -573,12 +574,16 @@ authorizing an authenticator with which to complete the operation.
...
+    by the client in order to generate this credential. The exact JSON encoding must be preserved as a cryptographic hash
+    (<a>clientDataHash</a>) has been computed over it.
 
-    The <dfn>attestation</dfn> attribute contains an attestation statement returned by the authenticator. This provides
-    information about the credential and the authenticator it is held in, such as the <a>credential public key</a> and the level
-    of security assurance provided by the authenticator.
+    The <dfn>attestationObject</dfn> element contains the <a>attestation object</a>. The contents of this object are determined
+    by the <a>attestation statement format</a> used by the authenticator. This object is opaque to, and cryptographically
+    protected against tampering by, the client. It contains the unique identifier of the credential, the <a>credential public
```
vijaybh: AAGUID is the authenticator's identifier, not the credential's. I tweaked the wording along these lines though.

index.bs
```diff
@@ -1128,19 +1089,19 @@ The goals of this design can be summarized as follows.
 
 - The design aims to reuse as much as possible of existing encoding formats in order to aid adoption and implementation.
 
-The contextual bindings are divided in two: Those added by the RP or the client platform, referred to as client data; and those
-added by the authenticator, referred to as the authenticator data. The client data must be signed over, but an authenticator is
-otherwise not interested in its contents. To save bandwidth and processing requirements on the authenticator, the client
-platform hashes the {{ClientData}} and sends only the result to the authenticator. The authenticator signs over the combination
-of this <a>clientDataHash</a>, and its own authenticator data.
+The contextual bindings are divided in two: Those added by the RP or the client, referred to as client data; and those added by
+the authenticator, referred to as the authenticator data. The client data must be signed over, but an authenticator is otherwise
```
vijaybh: Also changed to active voice.

index.bs
```diff
@@ -1190,14 +1151,11 @@ The encoding of authenticator data is a byte array of 37 bytes or more, as follo
...
-The RP ID hash is originally received from the client when the credential is created, and again when an assertion is generated.
+The RP ID is originally received from the client when the credential is created, and again when an assertion is generated.
 However, it differs from other client data in some important ways. First, unlike the client data, the RP ID of a credential does
 not change between operations but instead remains the same for the lifetime of that credential. Secondly, it is validated by the
-authenticator during the <a>authenticatorGetAssertion</a> operation, by making sure that the RP ID hash associated with the
-requested credential exactly matches the RP ID hash supplied by the client. These differences also explain why the RP ID hash is
-always a SHA-256 hash instead of being crypto-agile like the <a>clientDataHash</a>; for a given RP ID, we need the hash to be
-computed the same way by all clients for all operations so that authenticators can roam among clients without losing
-interoperability.
+authenticator during the <a>authenticatorGetAssertion</a> operation, by making sure that the RP ID associated with the
```
vijaybh: "Verifying" seemed better. WDYT?

index.bs
```diff
@@ -1220,11 +1178,65 @@ The attestation data (which is only present if the AT flag is set) describes its
...
+
 ### Generating a signature ### {#authenticator-signature}
 
-A raw cryptographic signature must assert the integrity of both the client data and the authenticator data. Thus, an
-<a>authenticator</a> SHALL compute a signature over the concatenation of the <a>authenticatorData</a> and the
-<a>clientDataHash</a>.
+Authenticators produce cryptographic signatures for two distinct purposes:
+1. An <dfn>attestation signature</dfn> is produced when a new credential is created, and provides cryptographic proof of certain
+    properties of the credential and the authenticator. For instance, an attestation signature asserts the type of authenticator
+    (as denoted by its AAGUID) and the public key of the credential. The attestation signature is signed by an authority key,
```
vijaybh: Changed to "attestation key" which matches the term we're using elsewhere

index.bs
```diff
@@ -1220,11 +1178,65 @@ The attestation data (which is only present if the AT flag is set) describes its
...
+    which is chosen depending on the type of attestation desired.
+2. An <dfn>assertion signature</dfn> is produced when the <a>authenticatorGetAssertion</a> method is invoked. It asserts that
+    the user has consented to a specific transaction. Thus, an assertion signature asserts the identity of the requester, and
```
vijaybh: Attempted to add such nuance.

index.bs
```diff
@@ -1234,95 +1246,103 @@ A raw cryptographic signature must assert the integrity of both the client data
...
-`clientData` and `authenticatorData` members respectively of the {{AuthenticationAssertion}} and {{AuthenticationAttestation}} structures.
 
 ### Verifying a signature ### {#authenticator-signature-verification}
 
-This section specifies the algorithm for verifying a signature assertion.
+This section specifies the algorithm for verifying any signature generated using the method in [[#authenticator-signature]].
 
-Upon receiving a signature assertion in the form of a {{AuthenticationAssertion}} structure, the [RP] shall:
+To verify a signature |S| given a claimed <a>clientDataJSON</a> |C| and a claimed <a>authenticatorData</a> |data| using a given
+public key, the [RP] shall:
```
vijaybh: Done

index.bs
```diff
@@ -1234,95 +1246,103 @@ A raw cryptographic signature must assert the integrity of both the client data
...
-    by the RP.
+6. Verify that the RP ID hash in |data| is indeed the SHA-256 hash of the RP ID expected by the RP.
 
-7. Compute the <a>clientDataHash</a>, i.e. hash of {{AuthenticationAssertion/clientData}}.
+7. Compute the <a>clientDataHash</a> over |C| using the {{ClientData/hashAlg}} algorithm found in the {{ClientData}} structure.
 
-8. Look up the previously registered public key associated with the credential (see {{makeCredential()}}) and
-    verify the signature in {{AuthenticationAssertion/signature}} computed over the
-    binary concatenation of {{AuthenticationAssertion/authenticatorData}} and <a>clientDataHash</a>.
+8. Use the given public key to verify that |S| is a valid signature over the binary concatenation of |data| and the
```
vijaybh: Done

index.bs
```diff
@@ -1520,94 +1459,189 @@ If the attestation root certificate is not dedicated to a single WebAuthn Authen
 should be specified in the attestation certificate itself, so that it can be verified against the <a>authenticatorData</a>.
 
 
+# [RP] Operations # {#rp-operations}
+
+Upon successful execution of a {{makeCredential()}} or {{getAssertion()}} call, the [RP]'s script receives a
+{{ScopedCredentialInfo}} or {{AuthenticationAssertion}} structure respectively from the client. It must then deliver the
+contents of this structure to the <a>[RP]</a>, using methods outside the scope of this specification. This section describes the
+operations that the [RP] must perform upon receipt of these structures.
+
```
vijaybh: Isn't that too limiting? It's fine to break up the structure into pieces if desired, each of the elements stands on its own.

index.bs
```diff
@@ -1520,94 +1459,189 @@ If the attestation root certificate is not dedicated to a single WebAuthn Authen
...
+
+## Registering a new credential ## {#registering-a-new-credential}
+
+When requested to register a new credential with a {{ScopedCredentialInfo}} structure, a [RP] must proceed as follows:
```
vijaybh: Revised along these lines

index.bs
vijaybh: OK

index.bs
```diff
@@ -1520,94 +1459,189 @@ If the attestation root certificate is not dedicated to a single WebAuthn Authen
...
+
+15. If the attestation statement could not be correctly verified, the [RP] MUST reject the registration operation.
+
+Verification of <a>attestation objects</a> requires that the [RP] has a trusted method of determining acceptable trust anchors
+in Step 11 above. Also, if certificates are being used, the [RP] must have access to certificate status information for the
+intermediate CA certificates. The [RP] must also be able to build the attestation certificate chain if the client did not
+provide this chain in the attestation information.
+
+To avoid ambiguity during authentication, the [RP] SHOULD check that each credential is registered to no more than one user. If
+registration is requested for a credential that is already registered to a different user, the [RP] MAY reject this operation,
```
vijaybh: Adopted this wording.

index.bs
```diff
@@ -1520,94 +1459,189 @@ If the attestation root certificate is not dedicated to a single WebAuthn Authen
...
+or it MAY decide to accept the registration while deleting the older registration.
+
+
+## Verifying an authentication assertion ## {#verifying-assertion}
+
+When requested to perform authentication with an {{AuthenticationAssertion}} structure, the [RP] MUST proceed as follows:
```
vijaybh: Updated wording

index.bs
```diff
@@ -204,9 +201,9 @@ NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED",  "MAY", and
 : <dfn>Attestation</dfn>
 :: Generally, a statement that serves to bear witness, confirm, or authenticate.
     In the WebAuthn context, attestation is employed to attest to the provenance of an authenticator and the data it emits;
-    including, for example: credential IDs, credential key pairs, signature counters, etc. <dfn>Attestation information</dfn> is 
-    conveyed in <a>attestation statements</a>. 
-    See also <a>attestation format</a>, and <a>attestation type</a>.
+    including, for example: credential IDs, credential key pairs, signature counters, etc. <dfn>Attestation information</dfn> is
+    conveyed in <dfn>attestation objects</dfn>.
+    See also <a>attestation statement format</a>, and <a>attestation type</a>.
```
vijaybh: OK

index.bs
```diff
@@ -1220,11 +1178,65 @@ The attestation data (which is only present if the AT flag is set) describes its
...
+        <td>Byte length L of Credential ID</td>
+    </tr>
+    <tr>
+        <td>L</td>
+        <td>Credential ID</td>
+    </tr>
+    <tr>
+        <td>variable</td>
+        <td>
+            Credential public key encoded in CBOR format. This is a CBOR map comprising the following fields:
```
vijaybh: Will do this in a future commit.

index.bs
```diff
@@ -1220,11 +1178,65 @@ The attestation data (which is only present if the AT flag is set) describes its
...
+
+            : alg
+            :: This is the name of the signature algorithm associated with the credential private key, expressed as defined in
+                [[!RFC7518]] section 3.1. Specifically, the following values are supported: "ES256", "ES384", "ES512", "RS256",
+                "RS384", "RS512", "PS256", "PS384" and "PS512".
+
+            : (public key fields)
```
vijaybh: See above. Will do in next commit.

index.bs
```diff
@@ -1234,95 +1246,103 @@ A raw cryptographic signature must assert the integrity of both the client data
...
-- The procedure for verifying an attestation statement, takes the following inputs:
-    - The <a>authenticator data</a> claimed to have been used for the attestation,
-    - The <a>clientDataHash</a> of the client's contextual bindings,
-    - A trust anchor (a root certificate, a DAA root key, or the <a>credential public key</a> itself),
+- The procedure for computing an attestation statement in this format given the credential to be attested, the
+    <a>authenticatorData</a> for the attestation, and a <a>clientDataHash</a>.
 
-    and returns a Boolean value indicating whether the attestation is cryptographically valid, and if so the attestation type.
+- The procedure for verifying an attestation statement, which takes as inputs the <a>authenticatorData</a> claimed to have been
+    used for the attestation and the <a>clientDataHash</a> of the client's contextual bindings, and returns either:
```
vijaybh: Reading this again, I think we can maybe reduce text and simplify. Now that we don't have that much in common between the attestation and assertion signatures, we can just delete the signature format section and absorb the signing and verification subsections into the attestation and assertion sections. Will try this out in the next commit.

index.bs
```diff
@@ -1359,107 +1378,27 @@ WebAuthn supports multiple attestation types:
...
-The authenticator MUST then concatenate this <a>authenticatorData</a> and the client-supplied <a>clientDataHash</a> as specified
-in [[#authenticator-signature]] to form <dfn>attToBeSigned</dfn>. It must then run the signing procedure for the desired
-attestation format, with <a>attToBeSigned</a> as input.
-
-
-### Verifying an Attestation Statement ### {#verifying-an-attestation-statement}
-
-This section specifies the algorithm for verifying an attestation statement, independent of attestation format.
+: authData
+:: The authenticator data used to generate the attestation statement.
```
vijaybh: Good point, but will do the naming once we've finished refactoring the text.

index.bs
```diff
@@ -1359,107 +1373,27 @@ WebAuthn supports multiple attestation types:
...
-    attestation (see [[#sctn-attestation-types]]). If doing so, the [RP] is asserting there is no cryptographic proof that the
-    Scoped Credential has been generated by a particular Authenticator model. See [[FIDOSecRef]] and [[UAFProtocol]] for a more
-    detailed discussion.
-
-Verification of attestation statements requires that the [RP] has a trusted method of determining the trust anchor in Step 8
-above. Also, if certificates are being used, the [RP] must have access to certificate status information for the intermediate
-CA certificates. The [RP] must also be able to build the attestation certificate chain if the client didn't provide
-this chain in the attestation information.
+: attStmt
+:: The attestation statement constructed above.
```
vijaybh: See above. Will do renaming once we finish with the text.

index.bs
```diff
@@ -1520,94 +1459,189 @@ If the attestation root certificate is not dedicated to a single WebAuthn Authen
...
+    the {{makeCredential()}} call.
+
+3. Verify that the {{ClientData/origin}} in the {{ClientData}} matches the [RP]'s origin.
+
+4. Verify that the {{ClientData/tokenBinding}} in the {{ClientData}} matches the token binding ID for the TLS connection over
+    which the attestation was obtained.
+
+5. Verify that the {{ClientData/extensions}} in the {{ClientData}} is a proper subset of the extensions requested by the RP.
+
+6. Compute the <a>clientDataHash</a> over {{ScopedCredentialInfo/clientData}} using the {{ClientData/hashAlg}} algorithm found
```
vijaybh: See above. Will experiment with removing the signature format section in next commit.

index.bs
vijaybh: See above. Will refactor signature format section to make this consistent.

index.bs
equalsJeffH: yeah, ok.

equalsJeffH:
New diffs:

just the diff of the "editorial revisions" commits added since the last set of diffs (above): i.e., most recent commit  7642dd6 diff'd from dc90eab:

http://kingsmountain.com/doc/diff/diff-webauthn-vgb-u2f-attestation-7642dd6--from--index-vgb-u2f-attestation-dc90eab.pdf

diff of most recent commit from current state of master branch:

http://kingsmountain.com/doc/diff/diff-webauthn-vgb-u2f-attestation-7642dd6--from--index-master-4751dac.pdf

index.bs
equalsJeffH: LGTM

index.bs
equalsJeffH: doh, mea culpa wrt AAGUID.
LGTM.

index.bs
equalsJeffH: well, the issue was that we were using "member" when we were talking about interface attributes, but now after the global change we're using "attribute" also when discussing dictionary members...     ( interfaces have attributes and dictionaries have members, go figure...)

index.bs
```diff
@@ -663,9 +668,9 @@ authorizing an authenticator with which to complete the operation.
 
     - The <dfn>extensions</dfn> parameter contains additional parameters requesting additional processing by the client and
         authenticator. For example, the caller may request that only authenticators with certain capabilities be used to create
-        the credential, or that additional information be returned in the attestation statement. Alternatively, the caller may
-        specify an additional message that they would like the authenticator to display to the user. Extensions are defined in
-        [[#extensions]].
+        the credential, or that additional information be returned in the <a>attestation object</a>. Alternatively, the caller
```
equalsJeffH: LGTM, thx.

index.bs
```diff
@@ -863,10 +829,11 @@ string-valued keys. Values may be any type that has a valid encoding in JSON. It
 This structure is used by the client to compute the following quantities:
 
 : <dfn>clientDataJSON</dfn>
-:: This is the <a>UTF-8 encoded</a> JSON serialization [[RFC7159]] of a {{ClientData}} dictionary.
+:: This is the <a>UTF-8 encoded</a> JSON serialization [[RFC7159]] of a {{ClientData}} dictionary. Any valid JSON serialization
+    may be used by the client. This specification imposes no canonicalization requirements.
```
equalsJeffH: LGTM

index.bs
equalsJeffH: touche'  ;-)    LGTM

index.bs
equalsJeffH: LGTM, thx.

index.bs
equalsJeffH: Looks nice. thx.

index.bs
equalsJeffH: LGTM

index.bs
equalsJeffH: Ok, thx.

index.bs
equalsJeffH: hm, yeah, as long at one doesnt mangle elements and break signature verification....  that's what I was getting at mostly.

index.bs
equalsJeffH: LGTM

index.bs
equalsJeffH: Hm.  ok, we can try it, but since the signature verification is largely the same between attestation and assertion then I am partial towards having a "stand alone" "signature verification" section that both attestation and assertion reference as appropriate. I guess I'm concerned about burying assertion verification in attestation because everyone is going to implement sig verification but not necessarily attestation verification (or so I've heard)

index.bs
equalsJeffH: pls see new comment on above: https://github.com/w3c/webauthn/pull/321#discussion_r100202940

index.bs
vijaybh: Shame on me for getting lazy. Fixed this more carefully.

index.bs
equalsJeffH: ok.

index.bs
equalsJeffH: ok.

equalsJeffH:
new diff of the changes Vijay pushed last night Sun 12-Feb 1900-ish PT..

http://kingsmountain.com/doc/diff-webauthn-vgb-u2f-attestation-2ebdcfe--from--index-vgb-u2f-attestation-7642dd6.pdf

vijaybh:
Merging based on in-person discussion at San Francisco f2f meeting.

PR #336, opened 2017-02-02
Fix up several linking mistakes.

jyasskin:


index.bs
```diff
@@ -39,28 +39,22 @@ Markup Shorthands: css off, markdown on
 
 <pre class="anchors">
 
-<!-- spec: HTML; urlPrefix: https://html.spec.whatwg.org/multipage/ -->
-spec: HTML51; urlPrefix: http://www.w3.org/TR/html51/; for: web
+spec: HTML; urlPrefix: https://html.spec.whatwg.org/multipage/
```
jyasskin: I could also switch this to HTML52 and https://w3c.github.io/html/. IMO, it's not quite as good as the WHATWG's version, but it fixes many mistakes in HTML5.1, which will save folks time reading your crosslinks.

index.bs
```diff
@@ -690,7 +694,7 @@ authorizing an authenticator with which to complete the operation.
     };
 </pre>
 
-<div dfn-for="Attachment">
+<div>
```
equalsJeffH: am curious why this is an unadorned `<div>` ?

jyasskin: No good reason; I just missed that I should remove it once it didn't have a `dfn-for`. I've pushed a new commit that removes it. Thanks!

PR #343, opened 2017-02-14
Make BufferSource parameters actually BufferSources.

### LINKED ISSUE #341: Examples show a DOMString passed to challenge, but it's a BufferSource
https://w3c.github.io/webauthn/#sample-registration shows:

```js
var challenge = "Y2xpbWIgYSBtb3VudGFpbg";
var options = { timeoutSeconds: 300,  // 5 minutes
                excludeList: [],      // No excludeList
                extensions: {"webauthn.location": true}  // Include location information
                                               // in attestation
};

// Note: The following call will cause the authenticator to display UI.
webauthnAPI.makeCredential(userAccountInformation, cryptoParams, challenge, options)
```

But the third parameter to [`makeCredential()`](https://w3c.github.io/webauthn/#iface-credential) is a [`BufferSource`](https://heycam.github.io/webidl/#BufferSource), which has to be one of the ArrayBuffer-related types. Which did you intend?

I think it'd be possible to use `(DOMString or BufferSource)`, and say that a DOMString is base64-decoded before it's used.

jyasskin:
Fixes #341.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2630,7 +2630,7 @@ The sample code for generating and registering a new key follows:
             algorithm: "RS256"
         }
     ];
-    var challenge = "Y2xpbWIgYSBtb3VudGFpbg";
+    var challenge = new TextEncoder().encode("climb a mountain");
     var options = { timeoutSeconds: 300,  // 5 minutes
                     excludeList: [],      // No excludeList
                     extensions: {"webauthn.location": true}  // Include location information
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -2630,7 +2630,7 @@ The sample code for generating and registering a new key follows:
             algorithm: "RS256"
         }
     ];
-    var challenge = "Y2xpbWIgYSBtb3VudGFpbg";
+    var challenge = Uint8Array.from("climb a mountain", c=>c.charCodeAt(0));
```
jyasskin: These match the old values, but it might make more sense to show random 128-bit strings for the challenges (https://w3c.github.io/webcrypto/Overview.html#Crypto-method-getRandomValues) and more likely ids for the acceptableCredentials?

Issue #334, opened 2017-01-28
Add clearer definition of API use cases to the spec 


AngeloKai:
see also issue #1389
potentially addressed by PR #1300
@christiaanbrand's https://github.com/w3c/webauthn/issues/334#issuecomment-388989355 is a build on the original post below:

The current use case section of the API has two main sections: registration and authentication. However, in reality, the API can be used for a number of means and uncertainty has led to confusion among developers we talked to. A clear definition would help us in guiding our work and give developers clear idea of what the API can do. I think the list below seems like a good starting point: 

- **Two factor - first login (bootstrap)**
  - Description:
      - The API can be used when a user logs in to a new machine for the first time. Typically a site requires the user to provide extra info to verify and then the user can use password. USB devices can be used to provide the extra info to verify the user’s identity.
  - Attributes of the authenticator devices used in this case: 
      - The device must be portable*.
      - The device doesn’t need to be positively identifiable**.

- **Two factor - always**
  - Description: 
      - The API can be used when a user logs in to a site that always requires password and second factor.
      - For example, after a user opts in to use U2F devices at Github.com, s/he is always required to type in password as well as plugs in a U2F device whenever s/he logs in.
   - Attributes of the authenticator devices used in this case:
      - The device must be positivity identifiable.
      - The device can be portable or built-in. 
	
- **Password-less Re-auth**
  - Description: 
      - User created a account along with a password saved on the server. However, the user doesn’t like having to type in password every time and opts to use an authenticator. 
  - Attributes of the authenticator devices used in this case: 
      - The device must be able to positively identify the user. 
      - The device can be either portable or built-in, though built-in authenticator likely most useful here, such as fingerprint reader, facial recognition. 

- **Password-free accounts, first login (bootstrap)**
  - Description: 
      - Unlike the case above, the user never has a password saved on the server in this case. When a user signs up, s/he used an authenticator and never used password to log in before. 
  - Attributes of the authenticator devices used in this case: 
      - The deice must be able to positively identify.
      - The device must be portable 

- **Sign in without user name**
  - Description:
      - User never created even a username in this case. The user is given a device at some point and is told to always use this device to log in. 
      - For example, a IT or retail employee uses NFC to login to site to punches in and out, record sick days, etc. 
      - The above example assumes the device is only used within a organization. However, this is not to say the case can only exist within enterprise scenarios. It is possible user ends up always use an authenticator to logs in. Then this leads to the problem of whether we want capabilities to the API in case a device’s memory is full.
[...]

PR #347, opened 2017-02-15
Make makeCredential() more precise.

jyasskin:
I've linked a lot more terms, reordered explanations to be clearer, and
specified some missing behavior.

This fixes #263 and #273, and improves #254 and #270.

PR #348, opened 2017-02-15
Add gesture verification parameter to option in both makeC and getA

AngeloKai:
As described in issue [334](https://github.com/w3c/webauthn/issues/334), certain use cases require authenticators that can verify the user while other use cases does not. This PR adds a gesture verification parameter to the option dictionary. 

If the parameter is set to true, the client will query all authenticators and see which one is capable of verification (and/or capable of cross-platform ability). Only authenticators capable of verification (and/or cross-platform ability) will be called. The authenticators will called the verification bit set to true. When authenticators see the bit set to true, they will perform user verification gestures during operations. 

The behavior of the parameter can be that it only requires authenticators capable of verifying users to be called without also sending down the bit during the actual request. However, certain authenticators can be engineered to create credentials when only simple gesture, (such as clicking a button) is performed or verification gesture is performed. For example, a phone that will sign the users in once the user clicks a button or presents fingerprint. If a developer sets this parameter, she will likely want to see the verification to actually happen.


index.bs
```diff
@@ -444,13 +444,16 @@ When this method is invoked, the user agent MUST execute the following algorithm
 8. Initialize |issuedRequests| and |currentlyAvailableAuthenticators| to empty lists.
 
 9. For each authenticator currently available on this platform, add the authenticator to |currentlyAvailableAuthenticators|
-    unless the {{ScopedCredentialOptions/attachment}} member of {{options}} is <a>present</a>. In that case, let |attachment|
-    be {{ScopedCredentialOptions/attachment}}, and add the authenticator to |currentlyAvailableAuthenticators| if its attachment
-    modality matches |attachment|.
+    unless the {{ScopedCredentialOptions/attachment}} member of {{options}} is present or the
```
jyasskin: Try not to use this "do x unless y. If y, z." form. Instead, it's clearer to write out "If y, z. Otherwise, x".

You'll probably need to indent the content of the "[for each](https://infra.spec.whatwg.org/#list-iterate)" and maybe name |authenticator|, but that's not really a problem.

index.bs
```diff
@@ -444,13 +444,16 @@ When this method is invoked, the user agent MUST execute the following algorithm
...
 
 9. For each authenticator currently available on this platform, add the authenticator to |currentlyAvailableAuthenticators|
-    unless the {{ScopedCredentialOptions/attachment}} member of {{options}} is <a>present</a>. In that case, let |attachment|
-    be {{ScopedCredentialOptions/attachment}}, and add the authenticator to |currentlyAvailableAuthenticators| if its attachment
-    modality matches |attachment|.
+    unless the {{ScopedCredentialOptions/attachment}} member of {{options}} is present or the 
+    {{ScopedCredentialOptions/verificationGesture}} member of {{options}} is set to true. In that case, let |attachment|
+    be {{ScopedCredentialOptions/attachment}} and |verificationGesture| be {{ScopedCredentialOptions/verificationGesture}}. 
+    Add the authenticator to |currentlyAvailableAuthenticators| if its attachment modality matches |attachment| and its 
+    user verification modality is set to true.
```
jyasskin: This is incorrect in the case that `attachment` is present but `verificationGesture` isn't set to true.

index.bs
```diff
@@ -444,13 +444,16 @@ When this method is invoked, the user agent MUST execute the following algorithm
 8. Initialize |issuedRequests| and |currentlyAvailableAuthenticators| to empty lists.
 
 9. For each authenticator currently available on this platform, add the authenticator to |currentlyAvailableAuthenticators|
-    unless the {{ScopedCredentialOptions/attachment}} member of {{options}} is <a>present</a>. In that case, let |attachment|
-    be {{ScopedCredentialOptions/attachment}}, and add the authenticator to |currentlyAvailableAuthenticators| if its attachment
-    modality matches |attachment|.
+    unless the {{ScopedCredentialOptions/attachment}} member of {{options}} is present or the 
+    {{ScopedCredentialOptions/verificationGesture}} member of {{options}} is set to true. In that case, let |attachment|
+    be {{ScopedCredentialOptions/attachment}} and |verificationGesture| be {{ScopedCredentialOptions/verificationGesture}}.
```
jyasskin: You should write out `<code>{{options}}.{{ScopedCredentialOptions/attachment}}</code>` to be clear which `attachment` attribute is being used.

After initializing *attachment*, you can check it for `undefined` to distinguish present from not-present.

index.bs
```diff
@@ -1121,6 +1143,9 @@ When this operation is invoked, the authenticator must perform the following pro
 - Prompt the user for consent to create a new credential. The prompt for obtaining this consent is shown by the authenticator
     if it has its own output capability, or by the user agent otherwise. If the user denies consent, return an error code
     equivalent to NotAllowedError and terminate the operation.
+- If the {{ScopedCredentialOptions/verificationGesture}} parameter is set to true, prompt the user for performing an 
+    <a>authorization gesture</a> to verify the user. If this prompt is shown with sufficient explanation that the success of 
+    this verification gesture will generate a new credential, the previous prompt for user consent may not be needed.
```
jyasskin: Try to avoid use of [COMEFROM](https://en.wikipedia.org/wiki/COMEFROM) in specs. ;) If `verificationGesture` can affect the previous prompt, you need to mention it there.

index.bs
```diff
@@ -523,9 +526,14 @@ When this method is invoked, the user agent MUST execute the following algorithm
     any), to create a {{ClientData}} structure representing this request. Choose a hash algorithm for {{ClientData/hashAlg}} and
     compute the {{AuthenticationAssertion/clientDataJSON}} and <a>clientDataHash</a>.
 
-6. Initialize |issuedRequests| to an empty list.
+6. Initialize |issuedRequests| to an empty list and |currentlyAvailableAuthenticators| to empty lists.
```
jyasskin: Remove the first "to an empty list"?

#347 expects that these are both [sets](https://infra.spec.whatwg.org/#sets) rather than lists.

index.bs
```diff
@@ -669,16 +678,21 @@ authorizing an authenticator with which to complete the operation.
         credentials for the same account on a single authenticator. The platform is requested to return an error if the new
         credential would be created on an authenticator that also contains one of the credentials enumerated in this parameter.
 
+    - The <dfn>attachment</dfn> parameter contains authenticator attachment descriptions, which are used as an additional
+        constraint on which authenticators are eligible to participate in a [[#makeCredential]] or [[#getAssertion]] operation.
+        See [[#attachment]] for a description of the attachment values and their meanings.
+
+    - The <dfn>verificationGesture</dfn> parameter is intended for use by <a>[RPS]</a> that want authenticators to perform the 
+        <a>user verification</a> process during the [[#makeCredential]] operation. A gesture that can be 
+        performed by anyone in posession of an authenticator is not considered a gesture that verifies the user, such as
```
jyasskin: This caveat should be in the definition of "user verification" (both https://w3c.github.io/webauthn/#user-verification and the behavioral definition in https://w3c.github.io/webauthn/#authenticatormakecredential) rather than just at some call sites.

index.bs
```diff
@@ -1121,6 +1143,9 @@ When this operation is invoked, the authenticator must perform the following pro
 - Prompt the user for consent to create a new credential. The prompt for obtaining this consent is shown by the authenticator
     if it has its own output capability, or by the user agent otherwise. If the user denies consent, return an error code
     equivalent to NotAllowedError and terminate the operation.
+- If the {{ScopedCredentialOptions/verificationGesture}} parameter is set to true, prompt the user for performing an 
+    <a>authorization gesture</a> to verify the user. If this prompt is shown with sufficient explanation that the success of
```
jyasskin: ["Authorization gesture"](https://w3c.github.io/webauthn/#authorization-gesture) is defined as "Essentially the same as user verification." Please only use one term for a given concept.

index.bs
```diff
@@ -1121,6 +1143,9 @@ When this operation is invoked, the authenticator must perform the following pro
 - Prompt the user for consent to create a new credential. The prompt for obtaining this consent is shown by the authenticator
     if it has its own output capability, or by the user agent otherwise. If the user denies consent, return an error code
     equivalent to NotAllowedError and terminate the operation.
+- If the {{ScopedCredentialOptions/verificationGesture}} parameter is set to true, prompt the user for performing an
```
jyasskin: The caveat that pressing a button doesn't count belongs here too, since this is the point implementers will actually be looking at when they decide whether they should allow pressing a button to count.

index.bs
```diff
@@ -651,6 +659,7 @@ authorizing an authenticator with which to complete the operation.
         USVString                            rpId;
         sequence<ScopedCredentialDescriptor> excludeList = [];
         Attachment                           attachment;
+        boolean                              verificationGesture;
```
balfanz: naming suggestion: how about "requireUserVerification" instead of "verificationGesture". A "verificationGesture" doesn't seem to be a thing that's a boolean, but (whether or not we) "require" feels more like a boolean.

index.bs
```diff
@@ -754,22 +768,28 @@ user consent to a specific transaction. The structure of these signatures is def
         unsigned long                        timeout;
         USVString                            rpId;
         sequence<ScopedCredentialDescriptor> allowList = [];
+        boolean                              verificationGesture;
```
balfanz: Why do we need this in the getA call? It's just needed in makeC, right?

index.bs
jyasskin: I've filed #358 to track this more generally. No need from my side to fix that in this PR.

index.bs
```diff
@@ -1106,9 +1126,11 @@ input parameters:
 - The <a>clientDataHash</a>, which is the hash of the serialized {{ClientData}} and is provided by the client.
 - The {{Account}} information provided by the [RP].
 - The {{ScopedCredentialType}} and cryptographic parameters requested by the [RP], with the cryptographic algorithms normalized
-    as per the procedure in [[WebCryptoAPI#algorithm-normalization-normalize-an-algorithm]].
+    as per the procedure in [[WebCryptoAPI#algorithm-normalization-normalize-an-algorithm]]
```
equalsJeffH: deleting the period was a mistake?

index.bs
```diff
@@ -1106,9 +1126,11 @@ input parameters:
...
 - A list of {{ScopedCredential}} objects provided by the [RP] with the intention that, if any of these are known to the
-    authenticator, it should not create a new credential.
+    authenticator, it should not create a new credential. 
+- A boolean called {{ScopedCredentialOptions/verificationGesture}} provided by the [RP] with the intention that, if the parameter 
+    is set to true, the authenticator SHOULD perform a user verification process during the operation.
```
equalsJeffH: s/user verification/[=user verification=]/

i wonder whether this would be a SHOULD or a MUST

AngeloKai:
@equalsJeffH To clarify, this PR doesn't attempt to address the silent auth situation at all. The difference between an authenticator that does user verification and one that doesn't is whether the user needs to show a fingerprint/face/pin to authenticate. An authenticator that doesn't do user verification is one that authenticates user based on only tapping or touching of the button. But the user in this case still made a move. Maybe instead of a binary object, I will change it to a enum to clarify this.

index.bs
vijaybh: I think I may have written this language but it's been on my list of editorial things to clean up.

index.bs
vijaybh: I think the intent is something like: for each authenticator present, the platform detects its |authnAttachment| and |authnVerificationCapability|. Then if |attachment| is not undefined, it weeds out all authenticators for which |authnAttachment| does not match |attachment|. Similarly, if |verificationGesture| is not undefined, it weeds out all authenticators for which |authnVerificationCapability| does not match |verificationGesture|. Perhaps we could reword things this way?

index.bs
equalsJeffH: Please see PR #365 for refinements of the definition of  "Authorization gesture"  and other related terms.

index.bs
equalsJeffH: the caveat..
> A gesture that can be performed by anyone in posession of an authenticator is not considered a gesture that verifies the user...

..is addressed in PR #365

index.bs
equalsJeffH: PR #365 refines Authorization gesture's definition.

index.bs
equalsJeffH: @balfanz suggested:
> how about "requireUserVerification" instead of "verificationGesture".

agreed.

index.bs
equalsJeffH: ok @vijaybh, you implicitly answered my questions, thx. yes, your reformulation makes nominal sense. Though see https://github.com/w3c/webauthn/pull/348/files#r103575859.

index.bs
equalsJeffH: @balfanz: 
> Why do we need this in the getA call? It's just needed in makeC, right?

Correct, I think, in the sense of having the user's credential created using an authenticator with the RP's desired properties, at makeCredential time, and then the RP can stipulate that authenticator's usage at getAssertion time by putting it first in the allowList sequence.  Yes?

equalsJeffH:
@AngeloKai said
> To clarify, this PR doesn't attempt to address the silent auth situation at all

I guess I was/am confused by having the proposed boolean also passed on the getAssertion call. please see https://github.com/w3c/webauthn/pull/348#discussion_r103577164

I've also added some other comments/replies in threads above (which don't go to the mailing list AFAICT).

AngeloKai:
@equalsJeffH @vijaybh @rlin1 @gmandyam please review changes. The text is based on the version with makeCredential and getAssertion both cleaned up.

index.bs
```diff
@@ -801,27 +809,32 @@ authorizing an authenticator with which to complete the operation.
...
-        the credential, or that particular information be returned in the [=attestation object=]. The caller may also specify
-        an additional message that they would like the authenticator to display to the user. Extensions are defined in
-        [[#extensions]].
-
-    - The <dfn>attachment</dfn> parameter contains authenticator attachment descriptions, which are used as an additional
+    - The optional <dfn>attachment</dfn> parameter contains authenticator attachment descriptions, which are used as an additional
         constraint on which authenticators are eligible to participate in a [[#makeCredential]] or [[#getAssertion]] operation.
         See [[#attachment]] for a description of the attachment values and their meanings.
 
+    - The optional <dfn>requireUserVerification</dfn> parameter can be used as an additional authenticator selection criteria.
```
vijaybh: Nit: s/criteria/criterion/

index.bs
```diff
@@ -1248,7 +1264,7 @@ input parameters:
 - The {{ScopedCredentialType}} and cryptographic parameters requested by the [RP], with the cryptographic algorithms normalized
     as per the procedure in [[WebCryptoAPI#algorithm-normalization-normalize-an-algorithm]].
 - A list of {{ScopedCredential}} objects provided by the [RP] with the intention that, if any of these are known to the
-    authenticator, it should not create a new credential.
+    authenticator, it should not create a new credential.
```
vijaybh: Unintentional change?

index.bs
```diff
@@ -311,6 +311,11 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
     that this does not constitute [=user verification=] because [=TUP=], by definition, is not capable of [=biometric
     recognition=], nor does it involve the presentation of a shared secret such as a password or PIN.
 
+: <dfn>Test of User Identity</dfn> 
+: <dfn>TUI</dfn> 
+:: A [=test of user identity=] is a technical process where a user's identity is verified by the authenticator through 
+    performing the <a>user verification</a> process.
```
vijaybh: We should use consistent markup - i.e. always use the [=term=] form instead of `<a>term</a>` unless there is a good reason.

index.bs
```diff
@@ -1207,9 +1221,11 @@ does not change between operations but instead remains the same for the lifetime
 by the authenticator during the [=authenticatorGetAssertion=] operation, by verifying that the RP ID associated with the
 requested credential exactly matches the RP ID supplied by the client.
 
-The `TUP` flag SHALL be set if and only if the authenticator detected a user through an authenticator specific gesture. The
+The `TUP` flag SHALL be set if and only if the authenticator detected a user through an authenticator specific gesture. The 
+`TUI` flag SHALL be set if and only if the <a>user verification</a> process has been performed by the authenticator. The
```
vijaybh: Let's use the [=term=] form of markup here as well.

index.bs
```diff
@@ -801,27 +809,32 @@ authorizing an authenticator with which to complete the operation.
...
-
-    - The <dfn>attachment</dfn> parameter contains authenticator attachment descriptions, which are used as an additional
+    - The optional <dfn>attachment</dfn> parameter contains authenticator attachment descriptions, which are used as an additional
         constraint on which authenticators are eligible to participate in a [[#makeCredential]] or [[#getAssertion]] operation.
         See [[#attachment]] for a description of the attachment values and their meanings.
 
+    - The optional <dfn>requireUserVerification</dfn> parameter can be used as an additional authenticator selection criteria. 
+        If this parameter is set to <i>true</i>, only  authenticators capable of user verification would participate in the 
+        [[#makeCredential]] operation, thereby ensuring the authenticator will perform the user verification process during 
+        the [[#makeCredential]] operation and future [[#getAssertion]] operation when it is requested to verify the credential.
```
vijaybh: You need to use this value someplace in the algorithm for makeCredential. I'm thinking somewhere around line 508.

gmandyam:
Current definition of User Verification does not cover Silent Authentication in my reading.  I think we should go forward with that assumption for <code>requireUserVerification</code> as well.  Based on this assumption, I am OK with the proposed PR.

selfissued:
This PR would change the API.  Therefore, we should thoughtfully consider it before the Implementer's Draft.

Issue #302, opened 2016-11-18
Throw "NotFoundError" when internal authenticator is not available or not found


AngeloKai:
In the current spec, four DOMException errors could be thrown in the execution process of makeCredential and getAssertion:
* "NotSupportedError": the error would be throw if a crypto algorithm developer wishes to use is not supported by the API 
* "SecurityError": the error would be throw if the procedure for relaxing the same-origin restriction went wrong and that procedure may go wrong because the developer doesn't follow security procedure
* "UnknownError": the error would be throw if fatal error event which we don't anticipate happening, such as JS exception
* "NotAllowedError": the error would be throw if  the origin is an opaque origin or the number of issued requests to authenticators is 0 

If any of the previous errors were thrown, the developer would most likely default the users back to their original flow. 

However, there is one situation that is not currently accounted for: when an internal authenticator is requested but no internal authenticator is found or not available. When this situation comes up, the developer could show some text to explain to the user that s/he could go register to start using the internal authenticator, such as enroll their fingerprint. 

Therefore, it would make sense to throw a DOMException error unlike the ones listed above. Among all the existing [DOMException errors](http://www.w3.org/TR/dom/#error-names) listed in the current HTML spec, the "NotFoundError" seems most fitting to the current situation. 

The existing "NotAllowedError" could work but the error is also thrown when the origin is an opaque origin. The "opaque origin" situation is rather different when the situation when an authenticator is not available; developer would also most likely want to do different things in the two different situations.

equalsJeffH:
This seems to be a good catch. "NotFoundError" works for me. 

note that DOMException errors are now found here: https://www.w3.org/TR/WebIDL-1/#h-idl-domexception-error-names

See "warning" here: https://www.w3.org/TR/dom/#error-names-0

vijaybh:
If we did #345 would this still be needed?

equalsJeffH:
see also #305

selfissued:
This affects interop, so should be addressed for the Implementer's Draft.

selfissued:
People agreed to do this on the call.

Issue #322, opened 2017-01-20
overall security considerations section or document


equalsJeffH:
Presently, the webauthn spec only has security (and privacy) considerations (sec cons) for Attestation. There are actually much more broad sec cons to document, especially given this is both a Web API and a cryptographic protocol spec.  For an example of comprehensive sec cons that are largely applicable to WebAuthn, see: 

https://fidoalliance.org/specs/fido-uaf-v1.1-rd-20161005/fido-security-ref-v1.1-rd-20161005.html

equalsJeffH:
see also: https://www.w3.org/TR/security-privacy-questionnaire/

nadalin:
I believe if we wanted to point to the FIDO document there would be no IPR issues but a logistical issue, so to use text from the FIDO document we would have to get the document as a contribution from FIDO much like the original specifications.

equalsJeffH:
Also, the spec ought to have a separate "privacy considerations" section.

equalsJeffH:
an [overall security considerations section](https://w3c.github.io/webauthn/#security-considerations) was recently added to the spec via PR #681.  however, IMO, to close this issue it ought to be enhanced to explain that the present web authn sec cons are incomplete, update the ref to [[FidoSecRef]]  to a newer version, and to please refer to [[FidoSecRef]] for further info.  we could/should then open another issue regarding having thorough webauthn-specific security considerations.

WDYT?

equalsJeffH:
on 29-Nov-2017 webauthn call: https://github.com/w3c/webauthn/issues/322#issuecomment-347936349 sounds like nominally OK plan, for attempting to attain CR maturity level.

PR #350, opened 2017-02-15
Throw NotFoundError when no authenticator is available

AngeloKai:
See [302](https://github.com/w3c/webauthn/issues/302) for references.


index.bs
```diff
@@ -447,6 +447,8 @@ When this method is invoked, the user agent MUST execute the following algorithm
     unless the {{ScopedCredentialOptions/attachment}} member of {{options}} is <a>present</a>. In that case, let |attachment|
     be {{ScopedCredentialOptions/attachment}}, and add the authenticator to |currentlyAvailableAuthenticators| if its attachment
     modality matches |attachment|.
+    - If no authenticator is available, reject |promise| with a <a>DOMException</a> whose name is "NotFoundError", and
```
jyasskin: I'd make this a separate step after this one:
> If |currentlyAvailableAuthenticators| is empty, [=reject=] |promise| with a "{{NotFoundError}}" {{DOMException}} and terminate this algorithm.

(See the example in https://heycam.github.io/webidl/#idl-exceptions for that exception creation wording.)

index.bs
```diff
@@ -525,7 +527,10 @@ When this method is invoked, the user agent MUST execute the following algorithm
 
 6. Initialize |issuedRequests| to an empty list.
 
-7. For each authenticator currently available on this platform, perform the following steps:
+7. If no authenticator is available, reject |promise| with a <a>DOMException</a> whose name is "NotFoundError", and 
+    terminate this algorithm.
+
+8. For each authenticator currently available on this platform, perform the following steps:
```
jyasskin: Note that Markdown is just as happy starting every numbered list item with "1.", so that new list items don't cause a diff for every subsequent item.

vijaybh: @jyasskin do you know any useful Bikeshed tricks for referring back to previous steps in an algorithm? Something like "use the quantity computed in step 7" for example is a pain because you have to manually maintain the correctness of the number over time.

jyasskin: This is https://github.com/tabatkins/bikeshed/issues/796, which we haven't yet gotten around to implementing. :-/

vijaybh: Thanks, I'd like that 🐴 too...

index.bs
vijaybh: Implementer question: what constitutes "authenticator not available" for this purpose? For NFC authenticators in particular, would the RP have to call the function in a loop until the user taps the authenticator on the reader? I understand this would be implementation specific for the platform, but some thought about sane implementation choices would be useful.

rlin1:
It is important to restrict such exception to no authenticator can be used - not no authenticator matching the list of preferred authenticators (the AAGUIDs) is present.

jcjones:
I'm nominally OK with this change, though it's not entirely clear to me how to implement it on the RP-side. I figure if early implementing RPs find this confusing, we could pull it back out before WGLC.

AngeloKai:
I will wait until https://github.com/w3c/webauthn/pull/371 is pushed in before I merged this change.

index.bs
```diff
@@ -741,6 +756,13 @@ authorizing an authenticator with which to complete the operation.
     - The <dfn>rpId</dfn> parameter explicitly specifies the RP ID that the credential should be associated with. If it is
         omitted, the RP ID will be set to the [=current settings object=]'s [=origin=].
 
+    - The <dfn>keystorage</dfn> parameter contains description of where the credential private key will be stored. The
+        private key . See [[#Key Storage]] for a description of the
```
jyasskin: There's a typo here: "The private key . See [[#Key Storage]]" ?

index.bs
```diff
@@ -727,6 +741,7 @@ authorizing an authenticator with which to complete the operation.
         USVString                            rpId;
         sequence<ScopedCredentialDescriptor> excludeList = [];
         Attachment                           attachment;
+        KeyStorage                           keystorage;
```
vijaybh: I have many comments on this topic, see my review of #378. Let's leave it out of this PR.

index.bs
```diff
@@ -269,6 +269,17 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
...
+:: The private key portion of an [=[RP]=]-specific <dfn>credential key pair</dfn>, generated by an [=authenticator=]. 
+    The exact location where the private key is stored may differ. In some cases, the private key may be always stored by
+    the [=authenticators=] and is never exported. A number of [=platform attached=] [=authenticators=] behaves in this manner. In 
+    other cases, the private key is never stored by the [=authenticator=] and is exported to the relying party as 
+    credential ID. Many [=cross-platform attached=] [=authenticators=] behave this way. In particular, all [=authenticators=] 
+    exercising the U2F protocol exports the private keys as credential ID. The public key portion of the
+    [=credential key pair=] is known as the <dfn>credential public key</dfn>. Note that in the case of [=self
+    attestation=], the [=credential key pair=] is also used as the [=attestation key pair=], see [=self attestation=]
+    for details.
+
```
vijaybh: IMO this definition is both unnecessary and inaccurate.

As mentioned in my review of #378, I disagree with using keyStorage to specify the storage of the credential private key (i.e., which speaks to the authenticator's implementation) instead of its availability (which speaks to the authenticator's behavior).

Also, no authenticator exports the private key as the credential ID. Some may do a cryptographic wrap of the key material to produce a credential ID, but even that is an implementation choice that should not be enshrined in a specification.

Finally, the last two sentences just seem out of place here.

selfissued:
It seems like this won't be ready in time for the Implementer's Draft.

selfissued:
I just reviewed these changes, as amended.  They seem straightforward and local in impact.  I believe that they are  ready to be merged, as-is.

equalsJeffH:
the github.com/AngeloKai repo needs to be updated with the latest bits from the github.com/w3c/webauthn repo, and then the AngeloKai:errors branch needs master merged into it, and any conflicts resolved (there may not be any). thanks.

equalsJeffH:
@rlin1: per your comment: https://github.com/w3c/webauthn/pull/350#issuecomment-283430835, it looks like the latest changes on lines 556-558 nominally address your concern since the error would be thrown only if there are "no currently available authnrs" -- yes?

equalsJeffH:
I do not see an answer to, or consideration of, @vijaybh's question: https://github.com/w3c/webauthn/pull/350#discussion_r103007560.

index.bs
```diff
@@ -553,7 +553,9 @@ When this method is invoked, the user agent MUST execute the following algorithm
 
 1. For each |authenticator| currently available on this platform, if
     <code>|options|.{{MakeCredentialOptions/attachment}}</code> is [=present|not present=] or its value matches
-    |authenticator|'s attachment modality, [=set/append=] |authenticator| to |currentlyAvailableAuthenticators|.
+    |authenticator|'s attachment modality, [=set/append=] |authenticator| to |currentlyAvailableAuthenticators|. If 
+    |currentlyAvailableAuthenticators| is a empty set, return a {{DOMException}} whose name is "{{NotFoundError}}", and
```
equalsJeffH: This sentence..
> If |currentlyAvailableAuthenticators| is a empty set, return a {{DOMException}} whose name is "{{NotFoundError}}", and terminate this algorithm.

..needs to be a completely separate step, I'm thinking. Otherwise a developer might evaluate it for each authenticator, rather than after all "currently available authnrs", and if the first authenticator is not added to |currentlyAvailableAuthenticators|, then the algorithm is terminated. 

also, it seems to me that..
> If |currentlyAvailableAuthenticators| is a empty set...

..ought to be..
> If |currentlyAvailableAuthenticators| is [=list is empty|empty=]...

..?

index.bs
```diff
@@ -690,6 +692,9 @@ When this method is invoked, the user agent MUST execute the following algorithm
 
 1. Let |issuedRequests| be a new [=ordered set=]. 
 
+1. If no |authenticators| is available on this platform, return a {{DOMException}} whose name is "{{NotFoundError}}",
```
equalsJeffH: s/If no |authenticators| is available/If there are no [=authenticators=] currently available/

vijaybh:
I agree with @equalsJeffH's comments but otherwise this LGTM. We're intentionally sacrificing some privacy for a better UX, and maybe at some point we should note this in a security considerations section as suggested in #322.

index.bs
```diff
@@ -553,7 +553,9 @@ When this method is invoked, the user agent MUST execute the following algorithm
 
 1. For each |authenticator| currently available on this platform, if
     <code>|options|.{{MakeCredentialOptions/attachment}}</code> is [=present|not present=] or its value matches
-    |authenticator|'s attachment modality, [=set/append=] |authenticator| to |currentlyAvailableAuthenticators|.
+    |authenticator|'s attachment modality, [=set/append=] |authenticator| to |currentlyAvailableAuthenticators|. If
```
leshi: What's an "attachment modality"?  I tried to find another reference to "modality" in the spec, but failed.  Is it here somewhere and I just don't see it?

leshi:
Can we explicitly document exactly the privacy tradeoff we're making in the comments of this PR (or the issue)?  Just good to have on record and so we can add that to the security considerations section as suggested by @vijaybh and @equalsJeffH

index.bs
equalsJeffH: we should define the term.  to me it means what [Attachment](file:///Users/jehodges/documents/work/standards/W3C/WebAuthn/index.html#enumdef-attachment) describes.

equalsJeffH:
@vijaybh: are you agreeing with my technical comments?

vijaybh:
@equalsJeffH yes, I think those should be addressed.

equalsJeffH:
IMHO this PR #350 is ready to merge as of https://github.com/w3c/webauthn/pull/350/commits/6fff645b8aae058a5743d321e4afa372a25d0afb. 

As extra special bonuses, it defines "attachment modality" ( @leshi pls reveiw ), defines "user verified", and fixes an incorrect ref to `displayName`.

selfissued:
Given all the approvals, I think it's time to merge this one.

equalsJeffH:
done -- thanks everyone :)

Issue #273, opened 2016-11-04
I don't understand how to create a ClientData in makeCredential


bzbarsky:
https://w3c.github.io/webauthn/#dom-webauthentication-makecredential step 8 says:

> Use attestationChallenge, callerOrigin and rpId, along with the token binding key associated with callerOrigin (if any), to create a ClientData structure representing this request. Choose a hash algorithm for hashAlg and compute the clientDataJSON and clientDataHash.

ClientData is defined to be a dictionary like so:

    dictionary ClientData {
      required DOMString           challenge;
      required DOMString           origin;
      required AlgorithmIdentifier hashAlg;
      DOMString                    tokenBinding;
      WebAuthnExtensions           extensions;
    };

Alright.  So how do I go about creating one?  

1. I guess I should fill in "challenge" from "attestationChallenge"?  But "attestationChallenge" is a BufferSource, while "challenge" is a DOMString.  How is the DOMString generated from the BufferSource?
2. I guess I should fill in "origin" from "callerOrigin", but the former is a DOMString while the latter is an origin; in this case a tuple.  How is the DOMString generated?
3. Where in here is rpId used?
4. This is the first mention of a "token binding key" in this algorithm.  What is that, and how does one find out whether there is one associated with `callerOrigin`?
5. It's not clear to me what "Choose a hash algorithm for hashAlg" means in practice.  hashAlg can be either a string or an ES object.  Which one is it expected to be in this case?  If it's expected to be an ES object, then we have a problem because then we need to define exactly what "JSON serialization" means for it...
6.  Is "extensions" supposed to be set to anything?  At first glance, no, since it's not in the set of data the spec says to use to create the ClientData.  But I expect that's a spec bug?

bzbarsky:
https://w3c.github.io/webauthn/#getAssertion step 5 has the same issue.

jcjones:
To echo bzbarsky, as `AlgorithmIdentifier` is typedef'd to `(object or DOMString)`, it's not default serializable. 

From the UA's perspective, it appears that I can get away with just declaring all instances of `AlgorithmIdentifier` in the webidl to be `DOMString`, as `hashAlg` is always chosen by the UA.

annevk:
(Note that where OP mentions step 8 it is now step 7.)

jyasskin:
[...]

PR #352, opened 2017-02-16
Make hashAlg a "recognized algorithm name".

jyasskin:
Fixes the hashAlg part of #273.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -56,6 +56,7 @@ spec: HTML52; urlPrefix: https://w3c.github.io/html/
 spec: WebCryptoAPI; urlPrefix: https://www.w3.org/TR/WebCryptoAPI/
     type: dfn
         text: normalizing an algorithm; url: dfn-normalize-an-algorithm
+        text: recognized algorithm name
     type: dictionary
         text: AlgorithmIdentifier; url: dfn-AlgorithmIdentifier
 </pre> <!-- class=anchors -->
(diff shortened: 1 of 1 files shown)


jyasskin:
@vijaybh @selfissued I don't have any problem with using shorter strings for these algorithms, but we do need to be clear what set of strings are valid.

I believe it was always incorrect to add SHA-256, etc. to the IANA JWS registry because S256, etc. had a different set of inputs than HS256, etc., and there was nothing in the registry to distinguish the two kinds of names. If you still want to use shorter names, we'll need to make a new registry.

vijaybh:
LGTM

I will note there is currently no "recognized algorithm name" corresponding to SM3 but maybe that should be fixed elsewhere by someone defining a new WebCrypto algorithm.

jyasskin:
Yeah, if you need SM3 for this spec, it'd be reasonable to define it here, in the same style as the other WebCrypto algorithms. We should accompany doing that with a WebCrypto issue so they know to upstream it when they have a chance.

(I'm assuming this spec has a good reason for adding a new hash algorithm.)

equalsJeffH:
wrt SM3, fyi/fwiw, it is there to support deployment in China (e.g., we added refs to the latter alg as well as a couple of other Chinese-standardized crypto algs to [the FIDO registry](https://fidoalliance.org/specs/fido-uaf-v1.1-id-20170202/fido-registry-v1.1-id-20170202.html) when defining a bit flag name for the Chinese SM2 elliptic curve based signature algorithm: ALG_SIGN_SM2_SM3).

though, it seems to me that defining "recognized alg names" for Chinese-approved crypto algs is a W3C policy/political decision beyond this working group's remit, and so we probably should not attempt to define it in this spec.

selfissued:
I am OK with this pull request.

Issue #358, opened 2017-02-22
Eliminate duplicate terminology


jyasskin:
I see several sets of equivalent terms in the spec. We should pick a canonical one from each set and eliminate the others (some of these also lack definitions, see also #462):

- [x] attestation key pair
  - [x] attestation key(s)
  - [x] public attestation keys
  - [x] key material

- [x] [Authentication Assertion](https://www.w3.org/TR/webauthn/#authentication-assertion)
  - [x] [Assertion](https://www.w3.org/TR/webauthn/#assertion)

- [ ] authenticator (these terms are the union of ones appearing in the spec and also used in various discussions):
  - [x] [platform authenticator](https://www.w3.org/TR/webauthn/#platform-authenticators)
  - [x] bound authenticator
  - [x] platform-provided authenticator
  - [x] built-in authenticator
  - [x] internal authenticator
  - [ ] others?
  - [x] [roaming authenticator](https://www.w3.org/TR/webauthn/#roaming-authenticators)
  - [x] cross-platform authenticator
  - [x] xplat authenticator   
  - [x] external authenticator
  - [ ] others?

- [x] [Client-side-resident Credential Private Key](https://w3c.github.io/webauthn/#client-side-resident-credential-private-key) (see also issue #936)
  - [x] replace [Client-side-resident Credential Private Key](https://w3c.github.io/webauthn/#client-side-resident-credential-private-key) with a "Client-side-resident [Credential Public Key Source](https://w3c.github.io/webauthn/#public-key-credential-source)" term with "resident credential" as the abbreviated form ?

- [x] create a "client-side-resident credential source" term rather than the "... private key" one?

- [x] also "client-side" is synonymous with "local", "server-side" with "remote" -- use both equivalently or coalesce to one or the other?

- [x] Collected Client Data i.e. [`CollectedClientData`](https://w3c.github.io/webauthn/#dictdef-collectedclientdata)
  - [x] [Client Data](https://www.w3.org/TR/webauthn/#sec-client-data)

- [x] [Conforming User Agent](https://www.w3.org/TR/webauthn/#conforming-user-agent)
  - [x] [Client](https://www.w3.org/TR/webauthn/#client)
  - [x] [WebAuthn Client](https://www.w3.org/TR/webauthn/#webauthn-client)

- [x] Credential ID
  - [x] credential identifier
  - [x] identifier of the credential

- [x] [Human-palatable](https://w3c.github.io/webauthn/#human-palatability) // should coverge on this
[...]

PR #365, opened 2017-02-24
refine user verification and authz gesture definitions, add Test of User Presence

### LINKED ISSUE #357: User Verification definition needs to be refined
As per WD-04:  https://www.w3.org/TR/webauthn/#terminology, the definition of User Verification does not address whether the verification takes place within the authenticator boundary.  Recommend adding the following sentence immeidately after 'Note that invocation of said operations implies use of key material managed by the authenticator':  
"Therefore the user verification operation must take place within the restricted operating environment of the authenticator itself."

equalsJeffH:
fixes #357, improves #358 -- see below for more details

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2949,7 +2949,7 @@ Brad Hill, Jing Jin, Anne van Kesteren, Giridhar Mandyam, Axel Nennker, Yaron Sh
     "href": "https://tools.ietf.org/html/draft-greevenbosch-appsawg-cbor-cddl",
     "status": "Internet Draft (work in progress)",
     "date": "21 September 2016"
-  }
+  },
   
   "ISOBiometricVocabulary": {
     "authors": ["ISO/IEC JTC1/SC37"],
--- a/index.bs
+++ b/index.bs
@@ -313,7 +313,7 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
 
 : <dfn>User Consent</dfn>
 :: User consent means the user agrees with what they are being asked, i.e., it encompasses reading and understanding prompts.
-    An [=authorization gesture=] is the [=ceremony=] component employed by a user to indicate consent.
+    An [=authorization gesture=] is a [=ceremony=] component often employed to indicate [=user consent=].
 
 : <dfn>User Verification</dfn>
 :: The technical process by which an <a>authenticator</a> <em>locally authorizes</em> the invocation of the
--- a/index.bs
+++ b/index.bs
@@ -230,13 +230,17 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
 : <dfn>Authenticator</dfn>
 :: A cryptographic device used by a <a>[WAC]</a> to (i) generate a <a>scoped credential</a> and register it with a <a>[RP]</a>,
     and (ii) subsequently used to cryptographically sign and return, in the form of an <a>Authentication Assertion</a>, a
-    challenge and other data presented by a <a>[RP]</a> (in concert with the <a>[WAC]</a>) in order to effect authentication.
+    challenge and other data presented by a <a>[RP]</a> (in concert with the <a>[WAC]</a>) in order to effect
+    [=authentication=].
 
 : <dfn>Authorization Gesture</dfn>
 :: An [=authorization gesture=] is a physical interaction performed by a user with an authenticator as part of a [=ceremony=], 
     such as [=registration=] or [=authentication=]. By making such an [=authorization gesture=], a user [=user consent|provides
-    consent=] for (i.e., <em>authorizes</em>) the [=ceremony=] to proceed. This may involve [=user verification=] if the
+    consent=] for (i.e., <em>authorizes</em>) a [=ceremony=] to proceed. This may involve [=user verification=] if the
     employed [=authenticator=] is capable, or it may involve a simple [=test of user presence=].
+
+: <dfn>Biometric Recognition</dfn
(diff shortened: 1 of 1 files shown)


equalsJeffH:
this PR refines/clarifies the definitions of:

authorization-gesture
- Ceremony
- User Consent
- user verification

adds definitions for:
- Biometric Recognition
- Test of User Presence

as well as adding some autolinks where appropriate.

index.bs
```diff
@@ -230,17 +230,25 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
...
-    that carry data. What is out-of-band to a protocol is in-band to a ceremony. In this specification, <a>Registration</a>,
-    <a>Authentication</a>, and <a>user verification</a> are ceremonies.
+: <dfn>Biometric Recognition</dfn
+:: The automated recognition of individuals based on their biological and behavioral characteristics [[ISOBiometricVocabulary]]. 
 
+: <dfn>Ceremony</dfn>
+:: The concept of a [=ceremony=] [[Ceremony]] is an extension of the concept of a network protocol, with human nodes alongside
+    computer nodes and with communication links that include user interface(s), human-to-human communication, and transfers of
+    physical objects that carry data. What is out-of-band to a protocol is in-band to a ceremony. In this specification,
+    <a>Registration</a> and <a>Authentication</a> are ceremonies, and an <a>authorization gesture</a> is often a component of
```
jcjones: Nit: The gesture is always a component of those ceremonies, right? The definition of user consent says the gesture  is _the_ ceremony component that defines consent.

index.bs
```diff
@@ -284,19 +292,29 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
...
-    <a>authenticatorMakeCredential</a> and <a>authenticatorGetAssertion</a> operations, for example through a touch plus pin
-    code, a password, a gesture (e.g., presenting a fingerprint), or other modality. Note that invocation of said operations
-    implies use of key material managed by the authenticator.
+:: The technical process by which an <a>authenticator</a> <em>locally authorizes</em> the invocation of the
+    <a>authenticatorMakeCredential</a> and <a>authenticatorGetAssertion</a> operations. [=User verification=] may be instigated
+    through various [=authorization gesture=] modalities; for example, through a touch plus pin code, password entry, or
+    [=biometric recognition=] (e.g., presenting a fingerprint) [[ISOBiometricVocabulary]]. Note that invocation of the
+    <a>authenticatorMakeCredential</a> and <a>authenticatorGetAssertion</a> operations implies use of key material managed by
+    the authenticator. Note that for security, [=user verification=] and use of [=credential private keys=] must occur within a
+    single logical security boundary defining the [=authenticator=].
```
vijaybh: Should we also add a sentence saying that a "fresh enough" user verification can also implicitly provide a TUP?

jcjones: Hmm... would we need to provide normative parameters for how to measure the freshness of the user check?

equalsJeffH: yeah, I did not as yet go into the so-called "touchless/silent" authentication aka getAssertion topic -- add that in a separate PR?

index.bs
equalsJeffH: @jcjones thx for your review.

actually, it says..
> ...an authorization gesture is _**often a component of**_
    those [=ceremonies=].

Does that address your concern?

index.bs
```diff
@@ -284,19 +292,29 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
...
 
 : <dfn>User Verification</dfn>
-:: The process by which an <a>authenticator</a> <em>locally authorizes</em> the invocation of the
-    <a>authenticatorMakeCredential</a> and <a>authenticatorGetAssertion</a> operations, for example through a touch plus pin
-    code, a password, a gesture (e.g., presenting a fingerprint), or other modality. Note that invocation of said operations
-    implies use of key material managed by the authenticator.
+:: The technical process by which an <a>authenticator</a> <em>locally authorizes</em> the invocation of the
+    <a>authenticatorMakeCredential</a> and <a>authenticatorGetAssertion</a> operations. [=User verification=] may be instigated
+    through various [=authorization gesture=] modalities; for example, through a touch plus pin code, password entry, or
+    [=biometric recognition=] (e.g., presenting a fingerprint) [[ISOBiometricVocabulary]]. Note that invocation of the
```
vijaybh: Based on discussion on today's call, maybe we should add a sentence saying that the intent is to be able to tell different human users apart by some means.

equalsJeffH: done in https://github.com/w3c/webauthn/pull/365/commits/d8b6c899a0db6d92530f044071a5e257f3ad3adf

equalsJeffH:
I believe @vijaybh and @jcjones comments are now addressed -- re-review?  thx.

index.bs
jcjones: No, but it's a nit.

Basically, if/when we permit the user consent to be obtained at a time before conducting a ceremony (a "recent consent" or "cached consent" perhaps), we'd need to update the definition of either "ceremony" or "user consent" to account for things happening not at the same time.

Right now, the statement 
>    ...an authorization gesture is often a component of those [=ceremonies=].

is flexible whereas the other definitions are not, so it seems to me it _could_ be rewritten as:
>     ...an authorization gesture is a component of those [=ceremonies=].

But I approve this PR as is, there's no need to an update on this account.

equalsJeffH: Ok, have also update the dfn for user consent:
> An [=authorization gesture=] is a [=ceremony=] component often employed to indicate [=user consent=].

how's that?

jcjones:
Thanks, 7b85ee10823e6b3453194a438ba908258f03ae2f looks perfect.

vijaybh:
LGTM, merging.

PR #369, opened 2017-03-02
Refine makeCredential description

vijaybh:
As per my comments on #347

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -461,7 +461,7 @@ When this method is invoked, the user agent MUST execute the following algorithm
     :: The [=unicode serialization of an origin|unicode serialization=] of |rpId|
     : {{hashAlg}}
     :: The [=recognized algorithm name=] of the hash algorithm selected by the client for generating the
-        [=hash of the client data=]
+        [=hash of the serialized client data=]
     : {{tokenBinding}}
     :: The [=Token Binding ID=] associated with |callerOrigin| (if any)
     : {{ClientData/extensions}}
@@ -469,7 +469,7 @@ When this method is invoked, the user agent MUST execute the following algorithm
 
 1. Let |clientDataJSON| be the [=JSON-serialized client data=] constructed from |clientData|.
 
-1. Let |clientDataHash| be the [=hash of the client data=] represented by |clientData|.
+1. Let |clientDataHash| be the [=hash of the serialized client data=] represented by |clientDataJSON|.
 
 1. Let |issuedRequests| and |currentlyAvailableAuthenticators| be new [=ordered sets=].
 
@@ -581,7 +581,8 @@ When this method is invoked, the user agent MUST execute the following algorithm
 
 1. Use {{assertionChallenge}}, |callerOrigin| and |rpId|, along with the token binding key associated with |callerOrigin| (if
     any), to create a {{ClientData}} structure representing this request. Choose a hash algorithm for {{ClientData/hashAlg}} and
-    compute the [=JSON-serialized client data=] |clientDataJSON| and the [=hash of the client data=] |clientDataHash|.
+    compute the [=JSON-serialized client data=] |clientDataJSON| and the [=hash of the serialized client data=]
+    |clientDataHash|.
 
 1. Initialize |issuedRequests| to an empty list.
 
@@ -639,7 +640,7 @@ authorizing an authenticator with which to complete the operation.
 
     The <dfn>clientDataJSON</dfn> attribute contains the [=JSON-serialized client data=] (see [[#cred-attestation]]) passed to
     the authenticator by the client in order to generate this credential. The exact JSON serialization must be preserved as the
-    [=hash of the client data=] has been computed over it.
+    [=hash of the serialized client data=] has been computed over it.
 
     The <dfn>attestationObject</dfn> attribute contains an <a>attestation object</a>. The contents of this object are
     determined by the <a>attestation statement format</a> used by the authenticator. This object is opaque to, and
@@ -875,7 +876,8 @@ string-valued keys. Values may be any type that has a valid encoding in JSON. It
     the client, in the syntax defined by [[RFC6454]].
 
     The <dfn>hashAlg</dfn> member is a [=recognized algorithm name=] that supports the `"digest"` operation, which specifies the
-    algorithm used to compute the [=hash of the client data=]. This algorithm is chosen by the client at its sole discretion.
+    algorithm used to compute the [=hash of the serialized client data=]. This algorithm is chosen by the client at its sole
+    discretion.
 
     The <dfn>tokenBinding</dfn> member contains the base64url encoding of the Token Binding ID that this client uses for the
     Token Binding protocol when communicating with the [RP]. This can be omitted if no Token Binding has been negotiated between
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -940,14 +886,13 @@ string-valued keys. Values may be any type that has a valid encoding in JSON. It
...
-        to generate them.
+    : <dfn dfn>JSON-serialized client data</dfn>
+    :: This is the [=UTF-8 encoding=] of the result of calling the initial value of {{JSON/stringify|JSON.stringify}} on a
+        {{ClientData}} dictionary. To avoid ambiguity, the {{ScopedCredentialInfo}} and {{AuthenticationAssertion}} structures
+        contain the actual serializations used by the client to generate them.
 
-    : <dfn dfn>clientDataHash</dfn>
-    :: This is the hash (computed using {{hashAlg}}) of <a for="ClientData">clientDataJSON</a>, as constructed by the client.
+    : <dfn dfn>Hash of the client data</dfn>
+    :: This is the hash (computed using {{hashAlg}}) of the [=JSON-serialized client data=], as constructed by the client.
```
equalsJeffH: I also suggest we make s/client data/<dfn>client data</dfn>/ at this line: https://github.com/w3c/webauthn/pull/369/files#diff-ec9cfa5f3f35ec1f84feb2e59686c34dR858 (it wouldnt let me insert a line comment there, so doing it here) -- and then use `[=client data=]" everywhere we are referring to just (unserialized, unhashed) "client data".

vijaybh: @equalsJeffH, did you mean s/ClientData/client data/ above? Your before and after strings look the same to me, I'm guessing due to autocorrect?

vijaybh: Or did you mean insert dfn tags there?

index.bs
```diff
@@ -1085,8 +1030,8 @@ authenticator. In verifying a signature, the server checks these bindings agains
 are divided in two: Those added by the RP or the client, referred to as client data; and those added by the authenticator,
 referred to as the authenticator data. The authenticator signs over the client data, but is otherwise not interested in its
```
vijaybh: Did you mean to add dfn tags here as well?

index.bs
```diff
@@ -413,65 +413,45 @@ This method takes the following parameters:
...
-    1. Otherwise, let |result| be the result of running |extension|'s [=client
-        processing=] algorithm on |argument|. If the algorithm returned an
-        error, [=continue=].
-
-        Issue(w3c/webauthn#363): Ensure all extensions define a client
-        processing algorithm.
+1. If the {{ScopedCredentialOptions/extensions}} member of {{options}} is <a>present</a>, then [=map/for each=]
+    |extension| → |argument| of <code>{{options}}.{{ScopedCredentialOptions/extensions}}</code>:
+    1. If |extension| is not supported by this client platform, then [=continue=].
+
```
vijaybh: I think this depends on the definition of what it takes to "support" an extension. If for certain extensions supporting them only requires doing some CBOR encoding, then maybe a browser that blindly does that should also be regarded as supporting the extension.

index.bs
equalsJeffH: yes, they are there but I hadn't escaped them :-/

equalsJeffH:
we also should rebase this branch onto the current state of master, yes?

equalsJeffH:
[Incorporate feedback from JeffH, part 1](https://github.com/w3c/webauthn/pull/369/commits/64455164f7881d198d8a9d14534a3107e6feca54)  - LGTM

index.bs
equalsJeffH: if you look at the source of my comment you can see the dfn tags. I've escapted them now, apologies for confusion....

equalsJeffH:
thx @vijaybh, this looks LGTM overall. 

A few nits:

1. Line 1115: (`[=TUP=]`) is not rendering correctly, guess we should drop the   "[=...=]"  ?

2. Line 1098:
The encoding of authenticator data is a byte array of 37 bytes or more, as follows.

suggest:
The encoding of the [=authenticator data=] structure is as a byte array of 37 bytes or more, as follows:

3. there's various instances of "authenticator data" and "client data" that I'd give the "[=...=]" treatment, e.g. around line 1053,  and then for just "authenticator data" in lines 1086-1100, and then on down thru the spec.  Similarly for "client data" on down thru the spec.   [realizing that there's some instances of either where such a change is not appropriate]

do we want to address them now or in a future PR ?

equalsJeffH:
part 3 LGTM, thx!   shall we merge this to master?

vijaybh:
Thanks for the review, merging.

index.bs
```diff
@@ -1900,7 +1859,10 @@ engine.
     :: The TPMT_PUBLIC structure (see [[TPMv2-Part2]] section 12.2.4) used by the TPM to represent the credential public key.
 
 : Signing procedure
-:: Concatenate the given <a>authenticatorData</a> and <a>clientDataHash</a> to form |attToBeSigned|.
+:: Let |authenticatorData| denote the [=authenticator data for the attestation=], and let |clientDataHash| denote the
```
jyasskin: This makes it look like the authenticator needs to compute the authenticator data and the hash of the serialized client data, when actually they're just arguments to this function. I think it'd be clearer to write the signing procedure as a function with parameters.

index.bs
```diff
@@ -499,104 +479,70 @@ When this method is invoked, the user agent MUST execute the following algorithm
...
 
-1. Let |clientDataJSON| be the [=UTF-8 encoding=] of the result of calling the
-    initial value of {{JSON/stringify|JSON.stringify}} on |clientData|.
+1. Let |clientDataJSON| be the [=JSON-serialized client data=] constructed from |clientData|.
 
-    Issue: Some extensions contain ArrayBuffers, which don't stringify well.
-    What's the intent here?
-1. Let |clientDataHash| be the hash of |clientDataJSON| using
-    <code>|clientData|.{{hashAlg}}</code>.
+1. Let |clientDataHash| be the [=hash of the serialized client data=] represented by |clientDataJSON|.
```
vijaybh: Yes, that is what it is. The intention of this change was to centralize this definition so all the places in the spec that refer to it can be consistent. I'm fine with updating this to be less ambiguous, I just want to make sure we do the same with all such instances in the spec.

index.bs
```diff
@@ -959,14 +908,13 @@ string-valued keys. Values may be any type that has a valid encoding in JSON. It
 
     This structure is used by the client to compute the following quantities:
 
-    : <dfn dfn>clientDataJSON</dfn>
-    :: This is the <a>UTF-8 encoded</a> JSON serialization [[RFC7159]] of a {{ClientData}} dictionary. Any valid JSON
-        serialization may be used by the client. This specification imposes no canonicalization requirements. Instead, the
-        {{ScopedCredentialInfo}} and {{AuthenticationAssertion}} structures contain the actual serializations used by the client
-        to generate them.
+    : <dfn dfn>JSON-serialized client data</dfn>
```
vijaybh: I see. Rewriting this as an algorithm is a great idea, and it would also help with the previous comment. Thanks for pointing this out.

index.bs
vijaybh: Yes, that is something I have wanted to do for a while as well. We should define what a signing procedure is in [Attestation Statement Formats](https://w3c.github.io/webauthn/#attestation-formats) and update the various formats to match. I started doing this, then realized that I'd also have to define data types and encoding for each argument, and so put it aside for a bit. Similar arguments can be made about the authenticator methods as well.

PR #377, opened 2017-03-14
alloc ArrayBuffers in correct global

### LINKED ISSUE #293: Please ensure that all ArrayBuffer creations are done explicitly, in the correct global
Right now it looks like the authenticator is expected to create ArrayBuffer when creating attestations, but it simply doesn't have the information to do so in the right global, as far as I can tell...

equalsJeffH:
..fixes #293, see https://github.com/w3c/webauthn/issues/293#issuecomment-286564318

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -43,6 +43,8 @@ Markup Shorthands: css off, markdown on
 spec: ECMAScript; urlPrefix: https://tc39.github.io/ecma262/#
     type: method
         for: JSON; text: stringify; url: sec-json.stringify
+    type: dfn
+        text: %ArrayBuffer%; url: sec-arraybuffer-constructor
 
 <!-- spec: HTML; urlPrefix: https://html.spec.whatwg.org/multipage/ -->
 spec: HTML52; urlPrefix: https://w3c.github.io/html/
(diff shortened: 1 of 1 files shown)


vijaybh:
This looks right to me, but I'm way out of my depth here. jyasskin or bzbarsky would be much more appropriate reviewers, though Github won't let me @ mention them or add them to the review here. Perhaps @jcjones can do a more informed review.

equalsJeffH:
> jyasskin or bzbarsky would be much more appropriate reviewers, though Github won't let me @ mention them or add them to the review here.

Indeed, I also attempted to add them to the reviewers list. @jyasskin should be able to be so added now that he is a formal member of the working group, but it may rely on having "write perms" to the repo (?)...

equalsJeffH:
also, I asked for @bzbarsky review over in issue #293

equalsJeffH:
Given this comment https://github.com/w3c/webauthn/issues/293#issuecomment-286893174 over on issue #293  -- might we get @jyasskin and/or @bzbarsky review of this PR and if it's OK then we merge it?  thx :)

equalsJeffH:
Thanks for the review @jyasskin 

> ...we should wait to insist that new specs follow this style until core specs like Fetch and HTML update, but I don't object to the improvement.

I'm not sure I understand -- did you mean to say "...that we should wait on insisting that new specs follow this style because updates to core specs like Fetch and HTML may obviate it..." ?

jyasskin:
@equalsJeffH No, I'm complaining at @bzbarsky. If the core web specs haven't updated even partially to a new style, we shouldn't be asking more peripheral specs like this one to update yet. There's no *problem* with this one updating; it's just work you didn't need to do yet.

bzbarsky:
I guess from my point of view it's easier to make improvements in stuff that's being created than it is to refactor existing things.  But yes, I agree we need to fix the existing things...

Issue #367, opened 2017-02-28
Add keyStorage enum to ScopedCredentialOptions


AngeloKai:
Authenticators with cross-platform roaming capability may or may not store key materials within the authenticator boundary. If an authenticator doesn't store key material, the RP needs to do additional actions of storing keys. Additionally, in our current spec, getAssertion call can be initiated by only providing RP ID and that means the call isn't initiated with any key material at all. 

Therefore, I propose we add a new enum called keyStorage to the options dictionary, with two values: "client" and "server". If the former value is supplied, the authenticator must store key material. If the latter value is supplied, the authenticator won't store any value at all and the server will take care of storing the key materials. If no value is supplied, that means the RP doesn't care about storage and is ok with this cornercase.

gmandyam:
@AngeloKai 

Will there be provisions in the PR for non-roaming authenticators to ignore a keyStorage option where "server" is specified?  Or would the makeCredential() fail if the authenticator cannot export key meaterial to the RP?

Also, "client" as an option is misleading.  "Client" is Webauthn is defined as "Conforming User Agent" - not the same as authenticator (in fact it is clearly distinct).

rlin1:
I don't think that authenticator typically would want to implement both options.
I one authenticator can store the key material locally, then there is no need and no benefit (IMHO) to store the key wrapped at the RP.

So rather than introducing yet another concept of dynamically controlling the authenticator behavior, we should assume a static key storage location, i.e. one AAGUID will always store keys in the same way (either locally or remotely in wrapped form).
If we do so, the already have the ability for the RP to select authenticator based on AAGUIDs - and hence no need to further complicate the API.

equalsJeffH:
@AngeloKai wrote:
> Authenticators with cross-platform roaming capability may or may not store key materials within the authenticator boundary

what you mean by "key materials" is actually the [credential private key](https://w3c.github.io/webauthn/#credential-private-key), yes?

> If an authenticator doesn't store key material, the RP needs to do additional actions of storing keys.

The implicit assumption in the spec has been that such authenticators (authnrs) would simply use their wrapped private keys as the credential ID they return to the RP.  This is how most (all?) U2F authenticators work today, yes? 

from the U2F Raw Messages spec:
> A key handle ... This a handle that allows the U2F token to identify the generated key pair. U2F tokens MAY wrap the generated private key and the application id it was generated for, and output that as the key handle. 

[...]

Issue #245, opened 2016-11-02
remove "required" on ScopedCredentialDescriptor.id


AngeloKai:
Although our current spec only has one credential type: “ScopedCred”, we may add more types to the spec in the future. In case more types were added in the future and developer hasn’t done works to adopt their server to accept newer types, they would want to only accept credentials of a certain type. However, our current design doesn’t provide a way for them to just search by type. Developer has to provide both credential type and credential id in allowList and excludeList to specify what credentials they find acceptable. 

To provide more flexibility to the developer to search, I propose we remove the “required” keyword from the id parameter in ScopedCredentialDescriptor and revise the language in the spec about allowList and excludeList being lists of acceptable credentials. Instead, let’s change the language to a “list of search criteria to find credentials acceptable to the callers.”

jcjones:
Making `id` optional for future expandability seems reasonable.

I think the language updates around usage are already handled. For example, using a credential right now says (emphasis mine):

> This optional member contains a list of PublicKeyCredentialDescriptor object representing public key credentials **acceptable to the caller, in decending order of the caller’s preference** (the first item in the list is the most preferred credential, and so on down the line).

(ExcludeList looks fine as it's written now, too)

So I'd support a PR that marks `PublicKeyCredentialDescriptor.id` optional. I could go either way about adding the note to PublicKeyCredentialType to the effect that `PublicKeyCredentialDescriptor.id` is mandatory for `"public-key"` types, because I feel like it's self-evident when you have a key handle.

equalsJeffH:
@jcjones wrote:
> I could go either way about adding the note to PublicKeyCredentialType to the effect that PublicKeyCredentialDescriptor.id is mandatory for "public-key" types

I'm confused -- which note? By "the note", perhaps you are referring to something said in conversation (eg on the webauthn call last week?) ?

Where you say "[PublicKeyCredentialType](https://w3c.github.io/webauthn/#credentialType)" above, did you actually mean "[PublicKeyCredentialDescriptor](https://w3c.github.io/webauthn/#credential-dictionary)" ?

> is mandatory for "public-key" types, because I feel like it's self-evident when you have a key handle. 

So what you are suggesting here is that if the RP does have a credential ID for a cred of type "public-key", it must supply it in the PublicKeyCredentialDescriptor.id, **_regardless of authnr capabilities_** ?

Is this because some authnrs supporting "public-key" creds must be presented with the credential ID to do anything (eg U2F authnrs) ?    

[...]

PR #378, opened 2017-03-14
1st approach: add requireResidentKey without adding authenticator selection dictionary

AngeloKai:
As per https://github.com/w3c/webauthn/issues/367, submitting a PR to add key storage enum to ScopedCredentialOptions and add resident key flag to the authenticator data. 

A couple of things about this commit:
- Whether the authenticator exports the private key as the credential ID is entirely up to the authenticator. In fact, many TPMs are designed so that the private key never leaves the device. 
- Instead of "onServer", the value is now "onRelyingParty". The change is made to give flexibility back to the RP as to where it wants to store the key information. While storing on the server is more secure, the RP can store the key in local storage.


index.bs
```diff
@@ -306,6 +311,11 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
     that this does not constitute [=user verification=] because [=TUP=], by definition, is not capable of [=biometric
     recognition=], nor does it involve the presentation of a shared secret such as a password or PIN.
 
+: <dfn>Resident Key</dfn>
+: <dfn>RK</dfn>
+:: A [=resident key=] is a flag representing whether the authenticator is capable of storing <a>credential private key</a> to 
+    generate the [RP]-specific <a>credential key pair</a>.
```
vijaybh: This definition does not make sense to me. A more natural way to define Resident Key would be as a key that can be looked up and used by the authenticator purely on the basis of its associated RP ID. We should seek to define behavior, not platform implementation details.

The important thing is that the Resident Key is a key, not a flag. If you want to define a flag that indicates an authenticator is capable of creating Resident Keys, then call it the Resident Key Capability Flag or something like that.

index.bs
```diff
@@ -494,17 +504,19 @@ When this method is invoked, the user agent MUST execute the following algorithm
...
 1. [=set/For each=] |authenticator| in |currentlyAvailableAuthenticators|:
     1. Let |excludeList| be a new [=list=].
     1. [=list/For each=] credential |C| in <code>{{options}}.{{ScopedCredentialOptions/excludeList}}</code>:
         1. If <code>|C|.{{transports}}</code> [=list/is not empty=], and |authenticator| is connected over a transport not
             mentioned in <code>|C|.{{transports}}</code>, the client MAY [=continue=].
         1. Otherwise, [=list/Append=] |C| to |excludeList|.
-    1. [=In parallel=], invoke the [=authenticatorMakeCredential=] operation on |authenticator| with |rpId|,
-        |clientDataHash|, {{accountInformation}}, |normalizedParameters|, |excludeList| and |clientExtensions| as parameters.
+    1. [=In parallel=], invoke the <a>authenticatorMakeCredential</a> operation on |authenticator| with |rpId|,
+        |clientDataHash|, {{accountInformation}}, |normalizedParameters|, |excludeList|, |keyStorage| and |clientExtensions|
```
vijaybh: You haven't defined where |keyStorage| comes from. Did you mean <code>{{options}}.{{ScopedCredentialOptions/keystorage}}</code> ?

index.bs
```diff
@@ -800,6 +813,14 @@ authorizing an authenticator with which to complete the operation.
         omitted, the RP ID will be set to the [=origin=] specified by the {{WebAuthentication}} object's [=relevant settings
         object=].
 
+    - The <dfn>attachment</dfn> parameter contains authenticator attachment descriptions, which are used as an additional
+        constraint on which authenticators are eligible to participate in a [[#makeCredential]] operation.
+        See [[#attachment]] for a description of the attachment values and their meanings.
+
+    - The <dfn>keyStorage</dfn> parameter contains description of where the <a>credential private key</a> should be stored.
```
vijaybh: Could we say something like "The <dfn>keyStorage</dfn> parameter indicates the [RP]'s requirements regarding availability of the credential."?

index.bs
```diff
@@ -850,6 +867,27 @@ example of the latter, when the user is accessing the [RP] from a given client f
 use a [=roaming authenticator=] which was originally registered with the [RP] using a different client.
 
 
+### Credential Private Key Storage Option enumeration (enum <dfn enum>KeyStorage</dfn>) ### {#keyStorage}
+
+<pre class="idl">
+    enum KeyStorage {
+        "onAuthenticator",
```
vijaybh: IMO this text is confusing and very likely to be misunderstood. Much better to have choices like say "resident" and "nonresident", and to use text similar to my suggestion on line 317. Otherwise a naive reader of this text might think it is okay to export a private key in the clear and store it on a server.

index.bs
```diff
@@ -1168,7 +1206,9 @@ The [=authenticator data=] structure is a byte array of 37 bytes or more, as fol
         <td>
             Flags (bit 0 is the least significant bit):
             - Bit 0: [=Test of User Presence=] (`TUP`) result.
-            - Bits 1-5: Reserved for future use (`RFU`).
+            - Bit 1: .
```
vijaybh: What does this mean?

index.bs
```diff
@@ -1168,7 +1206,9 @@ The [=authenticator data=] structure is a byte array of 37 bytes or more, as fol
         <td>
             Flags (bit 0 is the least significant bit):
             - Bit 0: [=Test of User Presence=] (`TUP`) result.
-            - Bits 1-5: Reserved for future use (`RFU`).
+            - Bit 1: . 
+            - Bit 2: [=Resident Key=] (`RK`) result.
```
vijaybh: What procedure is this the result of?

index.bs
```diff
@@ -1202,6 +1242,9 @@ requested credential exactly matches the RP ID supplied by the client.
 The `TUP` flag SHALL be set if and only if the authenticator detected a user through an authenticator specific gesture. The
 `RFU` bits in the flags byte SHALL be set to zero.
 
+The `RK` flag SHALL be set if and only if the authenticator is capable of storing <a>credential private key</a> to 
+    generate the [RP]-specific <a>credential key pair</a> on demand.
```
vijaybh: I'm having difficulty parsing this. Again, why not focus on this specific credential and say that it can be looked up and used in the future by this authenticator based only on RP ID?

index.bs
```diff
@@ -1241,6 +1284,7 @@ input parameters:
     as per the procedure in [[WebCryptoAPI#algorithm-normalization-normalize-an-algorithm]].
 - A list of {{ScopedCredential}} objects provided by the [RP] with the intention that, if any of these are known to the
     authenticator, it should not create a new credential.
+- The {{KeyStorage}} request provided by the [RP]
```
vijaybh: Instead of "request provided by the RP", maybe "option requested by the RP"?

index.bs
```diff
@@ -1257,7 +1301,8 @@ When this operation is invoked, the authenticator must perform the following pro
     - Generate a set of cryptographic keys using the most preferred combination of {{ScopedCredentialType}} and cryptographic
         parameters supported by this authenticator.
     - Generate an identifier for this credential, such that this identifier is globally unique with high probability across all
-        credentials with the same type across all authenticators.
+        credentials with the same type across all authenticators. If the {{onRelyingParty}} value is provided by the [RP], 
+        use the <a>credential private key</a> as the identifier for this credential.
```
vijaybh: No.

index.bs
```diff
@@ -1607,58 +1652,61 @@ ceremony, a [RP] MUST proceed as follows:
...
 
-7. Perform CBOR decoding on the {{ScopedCredentialInfo/attestationObject}} field of the {{ScopedCredentialInfo}} structure to
-    obtain the attestation statement format |fmt|, the [=authenticator data=] |authData|, and the attestation statement
-    |attStmt|.
+1. Perform CBOR decoding on the {{ScopedCredentialInfo/attestationObject}} field of the {{ScopedCredentialInfo}} structure to
+    obtain the Credential ID, the attestation statement format |fmt|, the [=authenticator data=] |authData|, and the 
+    attestation statement |attStmt|.
+
+1. Store the Credential ID if {{ScopedCredentialOptions/keyStorage/onRelyingParty}} value is supplied in the 
+    <code>{{options}}.{{ScopedCredentialOptions/keyStorage}}</code> parameter.
```
vijaybh: Please delete this. It is not well-defined. It is not clear what "store" means here, and it is not necessary.

The credential ID must *always* be stored on a successful registration, as specified already in step 13. It does not matter if the credential is resident or not.

AngeloKai:
@vijaybh Since last review, I have changed the "key storage" parameter as a Boolean called "requireResidentKey" and defined resident key in the terminology section. In addition, I removed the mention of the parameter in the authenticatorMakeCredential section based on the assumption that the authenticator doesn't need an extra parameter to tell it to store the key. If this isn't true, please let know and I will update that section.

jyasskin:
https://github.com/w3c/webauthn/issues/367#issuecomment-283439079 says this PR isn't necessary because the non-resident key is just encrypted into the credential ID that the RP needs to store regardless. Why's @equalsJeffH wrong there?

vijaybh:
@jyasskin - AIUI this is seeking to define the difference between different levels of functionality from the authenticator:
- For some authenticators, such as the ones you allude to, the private key material is in the Credential ID. So without a Credential ID, the authenticator is helpless - it cannot provide any assertions. From an RP perspective, this means that they must figure out who the user is first (to some level of confidence) so that they can supply that user's Credential IDs in the allowList to getAssertion.
- For richer authenticators, there is onboard storage. So given only an RP ID, the authenticator can pick a credential (perhaps even with help from the user, if it has UI) and generate an assertion. From the RP's perspective, they can just have the same getAssertion call with an empty allowList for all users, and figure out the username (or more generally, account ID) from the assertion they receive.
The idea is that RPs who only want to build the second type of experience should be able to indicate that.

jyasskin:
@vijaybh SG. Maybe that gives a hint as to the right naming for the parameter. It's less about where the key material is stored, and more about whether the Credential ID is a required argument to `getAssertion()`, right? That said, I don't see the Credential ID argument to `getAssertion()` at all, so I'm not sure how the non-storing option ever worked. What am I missing?

AngeloKai:
@jyasskin to pass in credential ID, the RP would add a new ScopedCredentialDescriptor with that credential ID in the allowList portion of the AssertionOptions.

jyasskin:
@AngeloKai Aha, thanks! In that case, this change should probably make [`ScopedCredentialDescriptor.id`](https://api.csswg.org/bikeshed/?url=https://raw.githubusercontent.com/w3c/webauthn/angelo-keystorage/index.bs#credential-dictionary) non-required so that RPs can avoid passing it when they know the authenticator can find the credential from just their RP ID? (This is #245 but for a different reason.)

AngeloKai:
@jyasskin I thought about making cred ID optional before. However, the challenge is that if ID is not required, the allowList as of today is basically useless. I think when Vijay mentioned empty allowList, he meant that the RP doesn't add the allowList parameter in option. If an RP trusts the authenticator to find credentials just based on RP ID, the RP would just not pass in allowList.

index.bs
```diff
@@ -237,7 +237,10 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
 :: A cryptographic device used by a [=[WAC]=] to (i) generate a [=scoped credential=] and register it with a [=[RP]=],
     and (ii) subsequently used to cryptographically sign and return, in the form of an [=Authentication Assertion=], a
     challenge and other data presented by a [=[RP]=] (in concert with the [=[WAC]=]) in order to effect
-    [=authentication=].
+    [=authentication=]. All authenticators refered to this document MUST be capable of the basic operations listed above. 
+    However, certain additive capabilites are only available among certain authenticators, such as the capability to 
+    store <a>credential private keys</a>. When an authenticator is said to be "capable of" performing a task, it means that
+    the authenticator can perform the task when requested.
```
vijaybh: Editorial nit: I am not sure that last sentence adds anything. Seems like it could be deleted.

index.bs
```diff
@@ -493,18 +501,18 @@ When this method is invoked, the user agent MUST execute the following algorithm
 
 1. Let |issuedRequests| and |currentlyAvailableAuthenticators| be new [=ordered sets=].
 
-1. For each |authenticator| currently available on this platform, if
-    <code>{{options}}.{{ScopedCredentialOptions/attachment}}</code> is [=present|not present=] or its value matches
-    |authenticator|'s attachment modality, [=set/append=] |authenticator| to |currentlyAvailableAuthenticators|.
+1. For each |authenticator| currently available on this platform:
+    1. if <code>{{options}}.{{ScopedCredentialOptions/attachment}}</code> is not [=present=] or its value matches |authenticator|'s attachment modality, continue to next step.
```
vijaybh: Else do what?

index.bs
```diff
@@ -493,18 +501,18 @@ When this method is invoked, the user agent MUST execute the following algorithm
...
 
 1. [=set/For each=] |authenticator| in |currentlyAvailableAuthenticators|:
     1. Let |excludeList| be a new [=list=].
     1. [=list/For each=] credential |C| in <code>{{options}}.{{ScopedCredentialOptions/excludeList}}</code>:
         1. If <code>|C|.{{transports}}</code> [=list/is not empty=], and |authenticator| is connected over a transport not
             mentioned in <code>|C|.{{transports}}</code>, the client MAY [=continue=].
         1. Otherwise, [=list/Append=] |C| to |excludeList|.
-    1. [=In parallel=], invoke the [=authenticatorMakeCredential=] operation on |authenticator| with |rpId|,
-        |clientDataHash|, {{accountInformation}}, |normalizedParameters|, |excludeList| and |clientExtensions| as parameters.
+    1. [=In parallel=], invoke the <a>authenticatorMakeCredential</a> operation on |authenticator| with |rpId|,
```
vijaybh: Why change this markdown?

index.bs
```diff
@@ -800,6 +809,14 @@ authorizing an authenticator with which to complete the operation.
         omitted, the RP ID will be set to the [=origin=] specified by the {{WebAuthentication}} object's [=relevant settings
         object=].
 
+    - The <dfn>attachment</dfn> parameter contains authenticator attachment descriptions, which are used as an additional
+        constraint on which authenticators are eligible to participate in a [[#makeCredential]] or [[#getAssertion]] operation.
+        See [[#attachment]] for a description of the attachment values and their meanings.
+
+    - The <dfn>requireResidentKey</dfn> parameter describes the [=[RPS]=]' requirements regarding availability of the 
+        <a>credential private key</a>. If the parameter is set to <i>true</i>, the authenticator SHOULD create a <a>resident
+        key</a> when creating a <a>scoped credential</a>. Similiar to {{attachment}}, the {{requireResidentKey}} parameter is an additional constraint on which authenticators are eligible for a [[#makeCredential]] or [[#getAssertion]] operation.
```
vijaybh: Please use the [=term=] style of markup instead of &lt;a>term&lt;/a> to be consistent with the rest of the spec.

vijaybh: It's only a constraint on which authenticators are eligible for makeCredential, right? For getAuthentication, authenticators will either respond or not, there is no reason to set this option.

index.bs
```diff
@@ -1607,58 +1623,58 @@ ceremony, a [RP] MUST proceed as follows:
...
 
-6. Compute the hash of {{ScopedCredentialInfo/clientDataJSON}} using the algorithm identified by
+1. Compute the hash of {{ScopedCredentialInfo/clientDataJSON}} using the algorithm identified by
     <code>|C|.{{ClientData/hashAlg}}</code>.
 
-7. Perform CBOR decoding on the {{ScopedCredentialInfo/attestationObject}} field of the {{ScopedCredentialInfo}} structure to
-    obtain the attestation statement format |fmt|, the [=authenticator data=] |authData|, and the attestation statement
-    |attStmt|.
+1. Perform CBOR decoding on the {{ScopedCredentialInfo/attestationObject}} field of the {{ScopedCredentialInfo}} structure to
+    obtain the Credential ID, the attestation statement format |fmt|, the [=authenticator data=] |authData|, and the
```
vijaybh: The credential ID is part of authenticator data, why single it out here?

selfissued:
It doesn't seem like this is mature enough for the Implementer's Draft.  But since it affects the API, we should probably give the API changes a thumbs-up or thumbs-down before then.

index.bs
```diff
@@ -324,6 +324,11 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
     that this does not constitute [=user verification=] because [=TUP=], by definition, is not capable of [=biometric
     recognition=], nor does it involve the presentation of a shared secret such as a password or PIN.
 
+: <dfn>Resident Key</dfn>
+: <dfn>RK</dfn>
+:: A [=resident key=] is a [=credential private keys=] that can be found and used by the authenticator based on only the
```
jyasskin: "credential private key"

index.bs
```diff
@@ -551,9 +556,12 @@ When this method is invoked, the user agent MUST execute the following algorithm
 
 1. Let |issuedRequests| and |currentlyAvailableAuthenticators| be new [=ordered sets=].
 
-1. For each |authenticator| currently available on this platform, if
-    <code>|options|.{{MakeCredentialOptions/attachment}}</code> is [=present|not present=] or its value matches
-    |authenticator|'s attachment modality, [=set/append=] |authenticator| to |currentlyAvailableAuthenticators|.
+1. For each |authenticator| currently available on this platform:
+    1. If <code>|options|.{{MakeCredentialOptions/attachment}}</code> is [=present|not present=] or its value matches
+        |authenticator|'s attachment modality, continue to next step. Otherwise, do not [=set/append=] |authenticator|
```
jyasskin: "do not X" is odd to see in a spec algorithm. Not-doing things is the default. ;)

Maybe `If <code>|options|.{{MakeCredentialOptions/attachment}}</code> is [=present=] and its value does not match |authenticator|'s attachment modality, [=continue=].`

index.bs
```diff
@@ -551,9 +556,12 @@ When this method is invoked, the user agent MUST execute the following algorithm
...
+        to |currentlyAvailableAuthenticators|.
+    1. If <code>|options|.{{MakeCredentialOptions/requireResidentKey}}</code> flag is not [=present=] or its value
+        is set to [=true=], [=set/append=] |authenticator| to |currentlyAvailableAuthenticators|.
```
jyasskin: Why is present-and-false different from not-present?

index.bs
```diff
@@ -922,6 +931,12 @@ optionally evidence of [=user consent=] to a specific transaction.
         authenticators are eligible to participate in a {{CredentialsContainer/create()}} or {{CredentialsContainer/get()}}
         operation. See [[#attachment]] for a description of the attachment values and their meanings.
 
+    :   <dfn>requireResidentKey</dfn>
+    ::  This member describes the [=[RPS]=]' requirements regarding availability of the [=credential private keys=]. If the
+        parameter is set to <a>true</a>, the authenticator SHOULD create a [=resident key=] when creating a [=scoped credential=].
```
jyasskin: You need to write requirements in an algorithm that actually performs the operations. In this case, it'd be somewhere in authenticatorMakeCredential, not just the non-normative description of the members of MakeCredentialOptions.

I suspect you should s/SHOULD/MUST/. Is there a point in passing `requireResidentKey` if authenticators are still allowed to store the key in the credential ID?

index.bs
jyasskin: Something in https://w3c.github.io/webauthn/#op-get-assertion is missing a mention of the fact that keys might only be found if their credential IDs are passed in allowList. Without that, the whole idea of resident vs non-resident keys doesn't really make sense.

equalsJeffH:
I just merged master into the angelo-keystorage branch to facilitate reviewing. 

Also, these linking errors appear when building the spec:
```
> curl https://api.csswg.org/bikeshed/ -F file=@index.bs -F force=1 -F output=err
\033[1;33mLINK ERROR:\033[0m No 'dfn' refs found for 'true'.
&lt;a data-link-type="dfn" data-lt="true">true&lt;/a>
\033[1;33mLINK ERROR:\033[0m No 'dfn' refs found for 'attachment' that are marked for export.
&lt;a data-link-type="dfn" data-lt="attachment">attachment&lt;/a>
\033[1;33mLINK ERROR:\033[0m No 'dfn' refs found for 'requireresidentkey'.
&lt;a data-link-type="dfn" data-lt="requireResidentKey">requireResidentKey&lt;/a>
\033[7;32m ✔ \033[0m Successfully generated, with 3 linking errors
```

index.bs
```diff
@@ -324,6 +324,11 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
     that this does not constitute [=user verification=] because [=TUP=], by definition, is not capable of [=biometric
     recognition=], nor does it involve the presentation of a shared secret such as a password or PIN.
 
+: <dfn>Resident Key</dfn>
+: <dfn>RK</dfn>
+:: A [=resident key=] is a [=credential private key=] that can be found and used by the authenticator based on only the 
+    associated RP ID. 
+
```
AngeloKai: Normally I don't care about renaming. But in this case, I just cannot see any merit to the  name "requireClientSideResidentPrivateKey". Neither the name "ClientSideResidentPrivateKey" nor the name "ResidentKey" means anything to the developer. The developer will read documentation to understand what the method does anyway, just like attachment or timeout. The name should be a hint of the behavior, not a word-by-word description of the behavior of the API. I am perfectly happy with adding more description to the name resident key or add a description that equates "client-side credential private key". In addition, no web developer reads WhatWG/W3C spec. Using fully qualified term such as credential private key or client-side doesn't mean anything to developers. I'd love to spare developers the pain of writing a 35-characters name every time they wants to use a feature.

AngeloKai: I consider the possibility of changing the name to "requireClientResidentKey:. However, most developers' perception of the word "client" is user agent, aka the browser. Using the name "client" may lead to developers thinking the private keys are kept on the browser. Therefore I propose we keep the name "requireResidentKey"

index.bs
```diff
@@ -563,7 +572,7 @@ When this method is invoked, the user agent MUST execute the following algorithm
         1. Otherwise, [=list/Append=] |C| to |excludeList|.
     1. [=In parallel=], invoke the [=authenticatorMakeCredential=] operation on |authenticator| with |rpId|,
         |clientDataHash|, |options|.{{MakeCredentialOptions/rp}}, |options|.{{MakeCredentialOptions/user}},
-        |normalizedParameters|, |excludeList| and |clientExtensions| as parameters.
+        |normalizedParameters|, |excludeList|, |clientExtensions|, and |keyStorageDevice| as parameters.
```
AngeloKai: It is necessary to pass the bit in to authenticatorMakeCredential given that this option is part of the parameters of the authenticatorMakeCredential operation. keyStorageDevice is the name of that parameter in the CTAP spec.

index.bs
AngeloKai: It is not necessary for that algorithm to be written in this case. We are doing this everywhere in the spec when it comes to authenticator operation. The way the platform finds if the authenticator is available is actually by using the authenticatorGetInfo method, which is documented in the CTAP spec and hard-coded in new authenticators already. The method is not there today anyway because that's really an editorial issue. I am happy with spending sometime later to update that section to more precisely describe the CTAP spec. 
I propose we just pass in the bit for now and update the algorithm to be more precise later. That entire section is already outdated compared to the existing CTAP spec.

index.bs
```diff
@@ -922,6 +932,12 @@ optionally evidence of [=user consent=] to a specific transaction.
         authenticators are eligible to participate in a {{CredentialsContainer/create()}} or {{CredentialsContainer/get()}}
         operation. See [[#attachment]] for a description of the attachment values and their meanings.
 
+    :   <dfn>requireResidentKey</dfn>
+    ::  This member describes the [=[RPS]=]' requirements regarding availability of the [=credential private keys=]. If the
+        parameter is set to <a>true</a>, the authenticator SHOULD create a [=resident key=] when creating a [=scoped credential=].
+        Similiar to [=attachment=], the [=requireResidentKey=] parameter is an additional constraint on which authenticators
+        are eligible for {{CredentialsContainer/create()|navigator.credentials.create()}} operation. 
+
```
AngeloKai: See above.

index.bs
jyasskin: This ties into #410 then: if the authenticator model section didn't exist, and merely referred to CTAP, it'd be clear that we need to look there to see the normative effect of `requireResidentKey`.

You should still remove the normative statements from the descriptions of these fields, and just write factual statements about how the CTAP algorithms use their values. (e.g. "the authenticator creates ...") Otherwise, when implementers are reading the algorithm definitions to figure out what to implement, they're likely to miss the extra requirements we place in the field descriptions, leading to bugs.

index.bs
jyasskin: It might help with naming to focus on what this field means for calls to `.get()`. As Vijay [said](https://github.com/w3c/webauthn/pull/378#issuecomment-288467611), if this field is omitted or `false`, the RP must pass a credential ID in [`allowList`](https://w3c.github.io/webauthn/#dom-scopedcredentialrequestoptions-allowlist). If it's `true`, then the RP doesn't need to pass the credential ID. So `mayOmitCredentialId` or `storeCredentialId` could be good names, or if you're willing to invert the default, `willPassCredentialId` or `getRequiresCredentialId`. This is used as `create({publicKey: {storeCredentialId: true}})`, right?

Allowing sites to omit the credential ID also allows the site to rely on browser or authenticator UI to select which account to log into (for a first-factor authenticator), while requiring the site to pass it forces the site to either use the authenticator as a second factor or to have the user explicitly type their username. Given that, maybe `firstFactor` or `authenticatorCanSelectAccount`, although I don't love either of those.

equalsJeffH:
To be clear, I do not think this essentially ad-hoc approach to authenticator selection is the way to go, regardless of whether the various issues with these particular proposed changes are addressed. 

For example, @leshi is now proposing in issue #424 to also be able to select authenticators by whether or not they are capable of user verification (which is one of the further attributes I listed in [my review above](https://github.com/w3c/webauthn/pull/378#pullrequestreview-34141186)).

Issue #345, opened 2017-02-14
Add "willMakeCredentialWorkWithTheseConstraints()" method to the API


kpaulh:
We started discussing something like this in the San Francisco f2f meeting. This method would add the ability for an RP to promote the capability to use a built-in authenticator - when and only when the user actually has the appropriate hardware support, without being intrusive.

Very loosely: 
`  Promise<boolean> willMakeCredentialWorkWithTheseConstraints();`

As an example use case, a user logs in to the RP. If the call is successful, the browser would drop an info bar / bubble that informs capable users of the RP's ability to use the authenticator for log-in. The bar provides the option to set it up or defer ("not now") or dismiss all future prompts. 

Choosing the affirmative could create a pop up (or new tab?) that would direct the user through the set-up flow (makeCredential() would be called here).

Alternatives discussed:
- querying the device, which has privacy issues. 
- adding additional bits to makeCredential() that would simply check if the call would *potentially* succeed without actually making a credential. Overloading the method in this way is clunky, might confuse developers.

equalsJeffH:
@kpaulh wrote:
> As an example use case, a user logs in to the RP. If the call is successful..

This is assuming that the login is a legacy one (e.g. via [form-based ("password") authentication](https://en.wikipedia.org/wiki/Form-based_authentication)) not using WebAuthn/FIDO means, yes?
> ..the browser would drop an info bar / bubble that informs capable users of the RP's ability to use the authenticator for log-in..

By "the authenticator" do you mean to say "an available platform authenticator"?  During the discussion at the end of the F2F meeting I recall that roaming authenticators (i.e., those with a cross-platform attachment) were not considered candidates for this use case -- is that still the intention?

Also, this use case depends upon the RP webapp developers to explicitly take advantage of the availability of this method and to weave it properly into their login flows, yes?  If they do not do so, then perhaps there is still need for possibly throwing a "NotFoundError" as anticipated in #302  and #350 ?  @vijaybh ?

vijaybh:
>> As an example use case, a user logs in to the RP. If the call is successful..
> This is assuming that the login is a legacy one (e.g. via form-based ("password") authentication) not using WebAuthn/FIDO means, yes?

I'm not sure that logging in to the RP is even a required part of the use case. It could be that the RP offers this option right on their login page to anonymous users. @kpaulh do you agree?

[...]

PR #379, opened 2017-03-14
Add isPlatformAuthenticatorReady function to the API surface

AngeloKai:
As per https://github.com/w3c/webauthn/issues/345, there is a real need to gives RP the ability to detect whether the client is capable of registering credentials with platform authenticators while giving users control over whether they want to share this information. The reality is that the number of devices with platform authenticators attached and set up is a still a drop of water against the backdrop of all devices. Without this function, the RP won't be able to tell if the authenticator is ready. Then the user is left alone to find out if s/he can use the platform attached authenticator to log in to this website. 

The method currently only supports platform authenticator based on the assumption that if a user goes out of his/her way to purchase a standalone authenticator, the user has a significant interest in testing out the functionality and the RP doesn't need to show a UX signal to attract user to use this feature. 

During discussion, one idea that pops up often is whether this is essentially the same as the [Permission API](http://www.w3.org/TR/permissions/). My current thinking is that they are different due to the intent user expresses while using the API. When a RP uses the permission API, user gives permission to share camera information. In this case, the user is indicating a strong interest to take action after s/he clicks yes. 

Resolving the promise with False can be a fingerprinting concern. However, what we have learned while implementing similar API like this before is that fingerprinting is less of a concern if it legitimately cripples the API. In this case, if we don't resolve the promise, developers won't be able to react based on user saying no. In addition, web developers are interested in having at least a control over the UI being shown.

***
[Preview](https://s3.amazonaws.com/pr-preview/AngeloKai/webauthn/angelo-findAuthnr.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/webauthn/a9fad51...AngeloKai:e924709.html)


index.bs
```diff
@@ -696,6 +698,39 @@ authorizing an authenticator with which to complete the operation.
 </div>
 
 
+### Signal platform authenticator capability - isPlatformAuthenticatorReady() method ### {#isPlatformAuthenticatorReady}
+
+<div link-for-hint="WebAuthentication/isPlatformAuthenticatorReady">
+This method is used to signal users whether they can log in to the [RP] with the <a>platform-attached</a> authenticator.
```
vijaybh: "Whether they can log in" implies that it tells the user a yes/no answer. In reality this is (potentially, if the platform is capable) telling the user that the RP would like to help them set up their platform authenticator.

index.bs
```diff
@@ -696,6 +698,39 @@ authorizing an authenticator with which to complete the operation.
...
 
 
+### Signal platform authenticator capability - isPlatformAuthenticatorReady() method ### {#isPlatformAuthenticatorReady}
+
+<div link-for-hint="WebAuthentication/isPlatformAuthenticatorReady">
+This method is used to signal users whether they can log in to the [RP] with the <a>platform-attached</a> authenticator. 
+The method gives [RP] the ability to detect whether the client is capable of registering credentials with 
+<a>platform authenticators</a> while giving users control over whether they want to share this information. When the 
+method is invoked, the <a>client</a> will search for the existence of an <a>platform authenticator</a> capable of 
+registering credentials. Upon finding one, the <a>client</a> will prompt the user for whether the user is interested
```
vijaybh: What if it doesn't find one? What happens to the promise?

index.bs
```diff
@@ -696,6 +698,39 @@ authorizing an authenticator with which to complete the operation.
...
+to register a credential with the authenticator. If so, the platform will redirect the user to a new tab to register
```
vijaybh: Why redirect to a new tab? Why not just resolve the promise and let the RP decide whether to redirect, remember the choice for next time, defer, or whatever?

index.bs
```diff
@@ -696,6 +698,39 @@ authorizing an authenticator with which to complete the operation.
...
+When this method is invoked, the user agent MUST execute the following algorithm: 
+
+1. Let |promise| be a [=a new Promise=] and return the |promise|.
+
+1. Let |isInterested| be a [=boolean=] value. 
+
+1. For all |authenticators| currently available on this platform, if there exists an |authenticator| with 
+    the value of "platform" for its attachment modality, the <a>user agent</a> may take this opportunity to 
+    ask the user whether the user is interested in registering a credential with a <a>platform authenticator</a>.
+    The <a>user agent</a> should show three optional actions to the user:
```
vijaybh: A few comments:
- This seems to go too far in specifying UI, IMO. For example, it implies there is necessarily a dialog box.
- It's not clear to me why this promise always needs to return. What harm is done by just letting it hang out there and never resolving it? You may have to do that anyways in the case there is no authenticator.
- Why are there exactly these three options? I would prefer for this to be in informative text (e.g. a note below the actual normative steps) rather than here. There may be a smaller or larger number of options that's appropriate, depending on the platform.

It seems to me that this would be much simpler if we only said:

1. Let |promise| be a [=a new Promise=]. Return |promise| and execute the following steps [=in parallel=].

1. Enumerate all the [=platform authenticators=] on this platform. If there are no such authenticators, take no further action and terminate this algorithm.

1. If one or more platform authenticators were found, make a determination of the user's willingness to create a credential for this [RP] on such an authenticator. Factors used in this determination may include, for example:
    - Whether the user is running in private or incognito mode
    - Whether the user has configured the browser to never create such credentials
    - Whether the user has on a previous occasion expressed an unwillingness to create a new credential for this [RP], either through configuration or by declining a user interface prompt.
    - The user's explicitly stated intention, as obtained through a browser-specific user interface prompt.

1. If the above procedure determines that the user is not willing to create a new credential on a [=platform authenticator=], take no further action and terminate this algorithm. Otherwise, resolve |promise| with a value of `True`.

index.bs
AngeloKai: I am a bit confused about "help them set up". AFAIK, the reason why this method is needed is because users who own a platform authenticator often don't know they can log in to a website with the authenticator. The goal of the method is to promote usage of the WebAuthn function among users who own platform authenticators, not to increase the number of users who activated the authenticator, which should be handled by the platform.

index.bs
AngeloKai: Good point. I am thinking maybe reject the promise with a "NotAllowedError", which would be the default error for all situations? This way the RP won't know exactly whether it is the lack of authenticator that triggers the failure, thus preventing fingerprinting attack.

index.bs
AngeloKai: That was a typo from my previous version of the draft. Changing it back to resolve the promise with True.

index.bs
jyasskin: FWIW, I like the idea of returning a `Promise` that resolves to indicate the presence of an authenticator or hangs to indicate non-presence. This could help deal with the case where the user attaches an authenticator for the first time after a website calls this function. Without a hanging `Promise`, we'd likely want another way to notify the website that the state has changed. With a hanging `Promise`, the UA can just resolve it discovers that the user has an authenticator. Hanging also avoids the potential information leak from the time it takes to return false.

index.bs
```diff
@@ -237,7 +237,7 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
 :: A cryptographic device used by a [=[WAC]=] to (i) generate a [=scoped credential=] and register it with a [=[RP]=],
     and (ii) subsequently used to cryptographically sign and return, in the form of an [=Authentication Assertion=], a
     challenge and other data presented by a [=[RP]=] (in concert with the [=[WAC]=]) in order to effect
-    [=authentication=].
+    [=authentication=].
```
vijaybh: Changed by mistake?

index.bs
```diff
@@ -696,6 +698,38 @@ authorizing an authenticator with which to complete the operation.
 </div>
 
 
+### Inform user platform authenticator capability - isPlatformAuthenticatorReady() method ### {#isPlatformAuthenticatorReady}
+
+<div link-for-hint="WebAuthentication/isPlatformAuthenticatorReady">
+This method can be used by [RPS] to inform user they can create a new credential with one or more platform authenticators.
+The method gives [RPS] the ability to detect whether the client is capable of registering credentials with
```
vijaybh: AIUI it only gives the RP the ability to tell the client that it (the RP) supports use of WebAuthn. The RP does not get to detect anything.

selfissued:
We should decide whether to make this API change before the Implementer's Draft.

AngeloKai:
@mikewest when you have a chance, can you please take a look? I am not sure if I should submit a PR to the CredMan spec for the IDL stuff as well.

AngeloKai:
the

mikewest:
Is this something you expect other credentials types to need in the future? It seems to be to be fairly specific to the `PublicKeyCredential` and the WebAuthn use case. Putting it on `navigator.credentials` seems overly generic.

Did you consider making it an attribute on `PublicKeyCredential`'s interface object instead, something kinda like what HTML is thinking about in https://github.com/whatwg/html/pull/1936?

```
  [Unscopable] readonly attribute Promise<void> platformAuthenticatorReady;
}
```

Used as:

```
PublicKeyCredential.platformAuthenticatorReady.then(_ => {
  // Do stuff here.
});
```

AngeloKai:
@jyasskin Added the examples as requested. Please let me know if it is still confusing. I do feel like I could have added more stuff on how the whole flow runs together. But I also feel like that could have happened after I submit the PR to move the example section to the beginning use case section.

equalsJeffH:
@AngeloKai -- AFAICT, this branch needs to be merged-from-master (again), thx.

equalsJeffH:
a rendered html diff for this PR is here: 

http://kingsmountain.com/doc/diff/diff-webauthn-index-angelokai-angelo-findAuthnr-9d073c0--from--index-master-c7f1592.html

equalsJeffH:
>a rendered html diff for this PR is here: 
> http://kingsmountain.com/doc/diff/diff-webauthn-index-angelokai-angelo-findAuthnr-9d073c0--from--index-master-c7f1592.html

the relevant changes are in this section:
4.1.5. Inform user platform authenticator capability - isPlatformAuthenticatorReady() method

index.bs
```diff
@@ -3242,6 +3287,43 @@ The sample code for generating and registering a new key follows:
...
+## Registration with Platform Authenticator ## {#sample-registration-with-platform-authenticator}
+
+This is flow for when the [=[RP]=] is only interested in creating a public key credential with 
+[=platform authenticators=].
+
+1. The user visits example.com. At this point, the user likely already logged in. 
+
+1. The [=[RP]=] script runs the code snippet below. 
+
+1. The client asks the user if s/he is interested in registering a credenital with platform authenticators.
```
The user agent asks the user whether they are willing to register with the [=[RP]=] using an available [=platform authenticator=].

index.bs
```diff
@@ -3242,6 +3287,43 @@ The sample code for generating and registering a new key follows:
...
+This is flow for when the [=[RP]=] is only interested in creating a public key credential with 
+[=platform authenticators=].
+
+1. The user visits example.com. At this point, the user likely already logged in. 
+
+1. The [=[RP]=] script runs the code snippet below. 
+
+1. The client asks the user if s/he is interested in registering a credenital with platform authenticators. 
+
+1. If the user indicates interest, proceed as described below. Otherwise, stop asking the user for interest.
```
If the user not willing, terminate this flow.

index.bs
```diff
@@ -3242,6 +3287,43 @@ The sample code for generating and registering a new key follows:
...
+
+1. The user visits example.com. At this point, the user likely already logged in. 
+
+1. The [=[RP]=] script runs the code snippet below. 
+
+1. The client asks the user if s/he is interested in registering a credenital with platform authenticators. 
+
+1. If the user indicates interest, proceed as described below. Otherwise, stop asking the user for interest. 
+
+1. The authenticator shows appropriate UI and direct user to create a credential. Once the new credential is created,
```
equalsJeffH: at this point, this scenario ought to have steps similar to, or the same as, [11.1 Registration](https://w3c.github.io/webauthn/#sample-registration) step 5 and onwards it seems.

index.bs
```diff
@@ -3242,6 +3287,43 @@ The sample code for generating and registering a new key follows:
...
 </pre>
 
+## Registration with Platform Authenticator ## {#sample-registration-with-platform-authenticator}
+
+This is flow for when the [=[RP]=] is only interested in creating a public key credential with 
+[=platform authenticators=].
+
+1. The user visits example.com. At this point, the user likely already logged in. 
+
+1. The [=[RP]=] script runs the code snippet below.
```
equalsJeffH: ought this scenario have steps similar to, or the same as, [11.1 Registration](https://w3c.github.io/webauthn/#sample-registration) step 3 and 4 after the above step ? (ie, for consistency)

index.bs
```diff
@@ -905,6 +905,49 @@ authorizing an authenticator with which to complete the operation.
 </div>
 
 
+
+### Inform user platform authenticator capability - isPlatformAuthenticatorReady() method ### {#isPlatformAuthenticatorReady}
```
jcjones: Maybe: "Advertise WebAuthn Support to Users - isPlatformAuthenticatorReady() method"?

index.bs
```diff
@@ -905,6 +905,49 @@ authorizing an authenticator with which to complete the operation.
...
+
+<div link-for-hint="WebAuthentication/isPlatformAuthenticatorReady">
+This method can be used by [=[RPS]=] to inform user they can create a new credential with one or more platform authenticators.
+The method gives [=[RPS]=] the ability to inform the client that they support registering credentials with <a>platform 
+authenticators</a> while giving users control over whether they want to share this information. When the
```
jcjones: Recommend: "while giving users control over whether this information is appropriate to provide to the RP."

index.bs
```diff
@@ -905,6 +905,49 @@ authorizing an authenticator with which to complete the operation.
...
+        [Unscopable] Promise < boolean > isPlatformAuthenticatorReady();
+    };
+</pre>
+
+When this method is invoked, the user agent MUST execute the following algorithm: 
+
+1. Let |promise| be a [=a new Promise=]. Return |promise| and execute the following steps [=in parallel=].
+
+1. Enumerate all the [=platform authenticators=] on this platform. If there are no such authenticators, resolve |promise| with
+    a value of false and terminate this algorithm.
```
jcjones: This should probably just not resolve, not resolve in a way leaking information without user consent.

index.bs
```diff
@@ -905,6 +905,49 @@ authorizing an authenticator with which to complete the operation.
...
+    partial interface PublicKeyCredential {
+        [Unscopable] Promise < boolean > isPlatformAuthenticatorReady();
+    };
+</pre>
+
+When this method is invoked, the user agent MUST execute the following algorithm: 
+
+1. Let |promise| be a [=a new Promise=]. Return |promise| and execute the following steps [=in parallel=].
+
+1. Enumerate all the [=platform authenticators=] on this platform. If there are no such authenticators, resolve |promise| with
```
jcjones: Perhaps this is more like, "evaluate if there are platform authenticators currently available, or known to be previously available to this platform" in some language to imply that it's a SHOULD and leave it to the UA to figure out the messy bits.

index.bs
```diff
@@ -905,6 +905,49 @@ authorizing an authenticator with which to complete the operation.
...
+    a value of false and terminate this algorithm.
+
+1. If one or more [=platform authenticators=] were found, make a determination of the user's willingness to create a credential
+    for this [RP] on such an authenticator. Factors used in this determination may include, for example:
+
+    1. Whether the user is running in private or incognito mode
+    1. Whether the user has configured the [=client=] to never create such credentials
+    1. Whether the user has on a previous occasion expressed an unwillingness to create a new credential for this [=[RP]=], either
+        through configuration or by declining a user interface prompt.
+    1. The user's explicitly stated intention, as obtained through a browser-specific user interface prompt.
```
jcjones: I think we usually say "user agent" instead of "browser"

index.bs
jcjones: False should be capitalized for consistency.

index.bs
```diff
@@ -905,6 +905,49 @@ authorizing an authenticator with which to complete the operation.
...
+
+1. If the above procedure determines that the user is not willing to create a new credential on a [=platform authenticator=],
+    resolve |promise| with a value of False and terminate this algorithm. Otherwise, resolve |promise| with a value of
+    true.
```
jcjones: capitalize "true", and I agree with @equalsJeffH re: not resolving if False is the case.

index.bs
```diff
@@ -3168,10 +3211,12 @@ platform to enumerate all the authenticator's credentials so that the client can
 
 ## Registration ## {#sample-registration}
 
-This is the first-time flow, in which a new credential is created and registered with the server.
+This is the first-time flow, in which a new credential is created and registered with the server. In this flow, the [=[RP]=]
```
jcjones: Do we need to do something to mark this as non-normative?

index.bs
```diff
@@ -3242,6 +3287,43 @@ The sample code for generating and registering a new key follows:
...
+
+            // If user is interested in the API, start registering credentials.
+            if (userIntent) {
+                var publicKeyOptions = { /* Public key credential creation options. */};
+
+                // Create and register credentials.
+                return navigator.credentials.create({ "publicKey": publicKeyOptions });
+            }
+
+            // If user is not interested, don't ask the user to register.
```
jcjones: In the sample code, we should put a note that this may never resolve; the catch/then aren't guaranteed to run, etc.

selfissued:
In today's checkin, I tried to incorporate the bulk of the suggestions by @equalsJeffH and @jcjones .  Please re-review.

equalsJeffH:
@mikewest -- If you can, please review the promise this specifies at https://github.com/w3c/webauthn/pull/379/files#diff-ec9cfa5f3f35ec1f84feb2e59686c34dR949 -- it returns a bool, rather than nothing as you suggested in https://github.com/w3c/webauthn/pull/379#issuecomment-308037955, and it also does not qualify the interface attribute as you did. Is it OK as written?  (It seems to me the fashion you suggested is more elegant)

equalsJeffH:
@selfissued wrote in https://github.com/w3c/webauthn/pull/379#pullrequestreview-53061221:
> ...if someone could figure out what the bikeshed/markdown error is, I'd appreciate it

by this you mean this error:
```
\033[1;31mFATAL ERROR:\033[0m Line 922 isn't indented enough (needs 1 indent) to be valid Markdown:
"&lt;/div>"
```
..?

I attempted to figure that out but was unable.

equalsJeffH:
wrt [my recent comment above](https://github.com/w3c/webauthn/pull/379#issuecomment-319226748)...

@AngeloKai [originally said](https://github.com/w3c/webauthn/pull/379#issue-214232061):
> ...if we don't resolve the promise, developers won't be able to react based on user saying no. In addition, web developers are interested in having at least a control over the UI being shown.

ok, now I think I'm getting what @AngeloKai is saying -- to test my understanding: 

if one crafts this method as @mikewest suggested such that it does not return a result and an affirmative answer is signaled by simply resolving the promise, and otherwise (no platform authnr available, user never answered, universe heat death, etc.) it never resolves, then there are not signals for crafting UX that can discriminate between success and various "failure" or "uh, i didn't see the prompt...." etc. cases.  Yes?

mikewest:
> The working group needs to figure out whether the promise should return False when a platform authenticator is not ready or whether to just never resolve the promise.

Based on the responses here, I guess I might not be understanding the feature. Do you expect this to raise UI to the user in order to make an authenticator available? How do you expect that flow to work? What do you expect developers to be doing with the yes/no response?

I had imagined this as being useful only in the positive direction: that is, if an authenticator is available (e.g. if the Promise resolves), you would use one of the other APIs (`create()`, `get()`, etc) in order to make use of that authenticator (and thereby prompt the user). If it wasn't available, you wouldn't. That is, you'd assume `false`, and know `true` iff the Promise resolved. What's the workflow y'all have in mind?

index.bs
```diff
@@ -931,6 +933,50 @@ authorizing an authenticator with which to complete the operation.
...
+
+1. If one or more [=platform authenticators=] were found, determine the user's willingness to create a credential
+    for this [RP] on such an authenticator. Factors used in this determination may include, for example:
+
+    1. Whether the user is running in private or incognito mode.
+    1. Whether the user has configured the [=client=] to never create such credentials.
+    1. Whether the user has previously expressed an unwillingness to create a new credential for this [=[RP]=],
+        either through configuration or by declining a user interface prompt.
+    1. The user's explicitly stated intentions, as obtained through user interaction with the user agent.
+
```
equalsJeffH: suggest compressing the above (lines 963 to 972) down to..
>Otherwise, assess whether the user is willing to create a credential for this [RP] using one of the available  [=platform authenticators=] (see above for a discussion of the factors that may be employed by the [=client=] in making this assessment).

note: further above, i suggest moving the example assessment criteria up to the alg introduction text.

index.bs
```diff
@@ -3271,6 +3319,43 @@ The sample code for generating and registering a new key follows:
...
+1. The user visits example.com. At this point, the user is likely already logged in.
+
+1. The [=[RP]=] script runs the code snippet below.
+
+1. The user agent asks the user whether they are willing to register with the [=[RP]=] using an available [=platform authenticator=].
+
+1. If the user is not willing, terminate this flow.
+
+1. The authenticator shows appropriate UI and direct user to create a credential. Once the new credential is created,
```

>1. The user is shown appropriate UI and guided in creating a credential using one of the available platform authenticators. Upon successful credential creation the RP script conveys the new credential to the server.

AngeloKai:
While the pull request was merged, I believe we discussed on the call that it'd be good if @kpaulh can review the pull request. After all, this is additional work on the browser end. And @kpaulh raised the issue initially. In cases of disagreements, revert is an option. 

@jcjones, @kpaulh, @leshi, at the interop meeting, perhaps we should chat about roughly how we want the method to work. I intentionally structured the PR in a very loose way. But I know that as a web dev myself, I'd hate if the behavior is vastly different between browsers.

index.bs
```diff
@@ -931,6 +933,50 @@ authorizing an authenticator with which to complete the operation.
...
+creating a credential based on a number of factors, which may include the user's explicitly stated intentions.
+If such interest is found, the promise will be resolved with the value of <i>True</i>.
+Based on the result, the [=[RP]=] can take further actions to guide the user to create a credential.
+
+This method has no arguments and returns a boolean value.
+
+<pre class="idl">
+    [SecureContext]
+    partial interface PublicKeyCredential {
+        [Unscopable] Promise < boolean > isPlatformAuthenticatorReady();
```
mikewest: My suggestion was to make this an attribute, not a method. :)

As I noted in https://github.com/w3c/webauthn/pull/379#issuecomment-319299153, it's not at all clear to me what workflow you're expecting here, so it's hard to give opinionated recommendations.

index.bs
```diff
@@ -931,6 +933,50 @@ authorizing an authenticator with which to complete the operation.
...
+This method can be used by [=[RPS]=] to inform the user that they can create a new credential with one or more platform authenticators.
+The method gives [=[RPS]=] the ability to inform the client that they support registering credentials with [=platform authenticators=]
+while giving users control over whether this information is appropriate to provide to the RP.
+When the method is invoked, the [=client=] will search for the existence of a [=platform authenticator=]
+capable of registering credentials.
+Upon finding one, the [=client=] will determine whether the user would be interested in
+creating a credential based on a number of factors, which may include the user's explicitly stated intentions.
+If such interest is found, the promise will be resolved with the value of <i>True</i>.
+Based on the result, the [=[RP]=] can take further actions to guide the user to create a credential.
+
```
mikewest: > Whether the user is running in private or incognito mode.

I would be very sad indeed if this method was useful to detect incognito mode. We're trying to remove those vectors, not to add new ones. Would y'all mind adding this to the privacy considerations section?

index.bs
selfissued: I'm far from an expert in Web IDL syntax.  I tried adding the `readonly` method and it gave me an IDL syntax error, so I removed it.  I just found your old comment from June 13th, @mikewest , in which you suggested this syntax:

`partial interface PublicKeyCredential {
  [Unscopable] readonly attribute Promise<void> platformAuthenticatorReady; }
`

I could change this to being a readonly attribute if people are good with that.

Comments?

AngeloKai: @selfissued @mikewest I thought about this question a few times in the past. The one question to be answered here is whether sharing the PlatformAuthenticatorReady information counts as fingerprinting and how much we want to tolerate the fingerprinting concern. 

Typically, for any personally identifiable information to be shared on the web, we would want to make sure we get user consent, such as prompting permission for geolocation etc. In case, the information is even more sensitive than geo. Unlike geo where the device almost always functions when it is hooked up to the client, most biometric authenticators require the user to enter their biometric info for a few times before the authenticators become ready. For example, Windows would ask the user to create a PIN first and then swipe their fingers multiple times. If the function resolves without UX or becomes an attribute that can be silently retrieved by any script, any site on the world would know (or at least have a good guess) the user has entered some sort of biometric info on the client. Of course we then enter into the discussion of whether the user actually cares we share this info. 

If the method becomes an attribute, the method would not prompt UI when being requested for this information and can become a privacy concern. 

It'd probably make sense to define an attribute on top of a method. In this sense, it means that after the user shows the interest, the platform will remember the choice. When the page loads in the future, the developer can check the attribute and can decide what UI to show. 

A note on the syntax proposed above, I am not sure if there needs to be a promise in there when the info is on the device. But I guess async is always good on the web.

AngeloKai: I can also imagine people start asking the question of why we shouldn't just extend the Permission API. There are two reasons here: 1) as stated above, the information shared includes not only the fact that the client has a device connected but also the fact that the user has used the device. 2) The user intent behind giving permission is more along the line of allowing the client to do something. But here the intent is more about user interested in using the device. I do go back and forth on the second point so I am hoping someone else brings a stronger counter argument.

kpaulh:
Is the example correct? Wouldn't this be called from PublicKeyCredential rather than from navigator.credentials? (per Mike's comments [here](https://github.com/w3c/webauthn/pull/379#issuecomment-308037955)). 

I must have missed the decision to resolve the promise no matter what. My initial thoughts were more in line with what Mike described [here](https://github.com/w3c/webauthn/pull/379#issuecomment-319299153).
Did we throw out the scenario in which the relying party would redirect the user to an enrollment page? The current example depicts more of a 'drive-by' enrollment. I guess I'd like to have a better understanding of the intended workflow as well.

On the other hand, the PR is written fairly loosely and I don't think it back us into any corners.
@leshi @balfanz, can we get your input?

selfissued:
I recently learned that there were in-person discussions in Barcelona in which it was agreed that having an interface that sometimes leaves a promise data structure in place that will never be resolved is poor systems design.  (It's like a C program that calls malloc() creating data structures that will never be either free()d or used.)  I agree with this conclusion.

balfanz:
One of the problems that this new method was trying to address was to not reveal whether or not a platform authenticator was available on the device, unless the user approves.

If the result is a `Promise<bool>`, however, then the caller will know:

- if it gets immediately resolved with false, it means there is no platform authenticator

All other cases mean that there is a platform authenticator:

- if it gets resolved after a while with false, it means there is a platform authenticator (user saw the system UI - which is only shown if there is, in fact, a platform authenticator, and declined)
- if it gets (eventually) resolved with true, it means there is a platform authenticator
- if it gets never resolved, it means that there is a platform authenticator (user saw the system UI - which is only shown if there is, in fact, a platform authenticator, and never takes action)

If the result is a `Promise<void>`, on the other hand, there is ambiguity:

- if the promise never gets resolved, we don't know whether the user declined, or whether there is no platform authenticator.

I will also point out that in the most likely case (the user approves use of the platform authenticator), the promise resolves in _both_ approaches. And in the case where the user doesn't interact with the system UI, the promise resolves in _neither_ approach. You only get the benefit of not having a dangling promise in the `Promise<bool>`-variant if the user actively denies the use of the platform authenticator in the system UI.

I'm in favor of addressing @mikewest 's concerns in another PR.

index.bs
```diff
@@ -933,6 +935,35 @@ authorizing an authenticator with which to complete the operation.
...
+Upon invocation, the [=client=] employs a platform-specific procedure to discover available [=platform authenticators=].
+If successful, the [=client=] then assesses whether the user is willing to create a credential using one of the available [=platform authenticators=].
+This assessment may include various factors, such as:
+    - Whether the user is running in private or incognito mode.
+    - Whether the user has configured the [=client=] to not create such credentials.
+    - Whether the user has previously expressed an unwillingness to create a new credential for this [=[RP]=],
+        either through configuration or by declining a user interface prompt.
+    - The user's explicitly stated intentions, determined through user interaction.
+If this assessment is affirmative, the promise is resolved with the value of <i>True</i>.
+Otherwise, the promise is resolved with the value of <i>False</i>.
```
kpaulh: I think that it's possible to both
a) avoid dangling promises
b) preserve privacy in all cases other than the user's explicit consent
by adding wording that mandates that the agent must determine and use a static timeout for reporting any and all false responses. i.e,:
If the user consents, the promise immediately resolves with True.
In any other situation (responds no, ignores the prompt, doesn't have platform authenticator), the promise holds onto that response and resolves with False only at the appointed time.
(Maybe this was what you intended, but it probably should be specifically stated.)

index.bs
```diff
@@ -3292,6 +3325,43 @@ The sample code for generating and registering a new key follows:
...
+
+1. If the user is not willing, terminate this flow.
+
+1. The user is shown appropriate UI and guided in creating a credential using one of the available platform authenticators.
+    Upon successful credential creation, the RP script conveys the new credential to the server.
+
+<pre class="example" highlight="js">
+    if (!PublicKeyCredential) { /* Platform not capable of the API. Handle error. */ }
+
+    navigator.credentials.isPlatformAuthenticatorAvailable()
```
kpaulh: Should hang off PublicKeyCredential, right?

index.bs
AngeloKai: I am not sure if adding a timer here helps much. Based on past experience, it appears that timer can be quite a battery drain on the platform. @jyasskin can potentially weigh in. Since the promise will be resolved eventually, The timer here also won't buy us much privacy too. The timer simply delays the exposure of the information but doesn't actually preserve privacy. If a bad guy truly wants to fingerprint, s/he can still do so.

index.bs
AngeloKai: For consistency purposes, yes. But I don't think it actually matters. They mean the same thing anyway.

index.bs
selfissued: Adding arbitrary delays seems like it could result in some truly terrible user experiences as well.

kpaulh: The delay shouldn't be user visible in any sense, since the rp's action upon receiving "false" is to do nothing. A "true" response, which does require some user action or interaction, would be conveyed immediately. The concept of the delay is just to hide which of the three failure cases is resolving the promise. 
(I don't disagree that it might cause possible battery drain.)

I guess my real question might be - have we agreed to give up on the idea that we shouldn't try to prevent fingerprinting?

selfissued: I question whether the RP's action upon receiving "false" is to do nothing.  In at least some use cases, I expect that the RP's action would be to immediately try to use non-platform authenticators.  This "immediately" would include an intervening long delay, making for a terrible UX, if the proposal to wait is adopted.

index.bs
leshi: Wat?

Aren't you modifying navigator.credentials in one case (like for all navigator.credentials) and just the webauthn stack in the other case?  Or am I totally confused?

index.bs
balfanz: I don't think that use cases in which someone would use a user-verifying platform authenticator overlap with use cases in which someone would use a non-platform authenticator.

To repeat @kpaulh 's question: Do we no longer have the goal of hiding the existence-vs-non-existence of authenticators from RPs unless the user approves? If we don't care, then I think we should take a different approach altogether - a GetAuthenticatorInfo() call or some such thing. But I think we *should* care, and that means that the two different error conditions (authenticator-doesn't-exist vs authenticator-exists-but-user-declines-use) should be indistinguishable.

index.bs
mikewest: 1. It does matter.
2. They don't mean the same thing.

I'd suggest that we update this example to match the IDL in https://w3c.github.io/webauthn/#isPlatformAuthenticatorAvailable.

index.bs
mikewest: A few small questions:

1.  If the RP is expected to act only in the case of a positive response, then it's not clear that providing a negative response has any impact over and above the fingerprinting risk. I'd suggest that we could simply not send a negative response (e.g. not resolve the `Promise`).

2.  Is the RP's response to a positive response defined? Talking to @kpaulh, it's not clear that there's agreement between vendors about what question we'd be asking users, and what the response to that question would actually mean. I worry a bit about the UI we'd present in that case, and what we'd actually ask users to approve.

3.  Given the amount of time folks generally spend on login pages, it seems like a delay long enough to be meaningful would be functionally indistinguishable from just not resolving the promise. What kind of timeout were you considering, @kpaulh? (On this point, it's not really clear to me why it would be bad for the promise not to resolve... That's totally something that can happen, and that callers need to deal with (e.g. `return new Promise(_ => {});`).

index.bs
selfissued: So we get the proposed change right the first time, can someone who's advocating for the change please propose the exact change wanted - along the lines of s/abc/xyz/ ?  Thanks.

equalsJeffH: Is someone crafting a new PR with which to propose a resolution to this discussion?

equalsJeffH:
I agree with @mikewest and @balfanz analysis and feel that the manner in which this PR works is suboptimal.

selfissued:
The PR adds the timeout that Dirk requested, so I believe it address their fingerprinting issue.

equalsJeffH:
what timeout, there's no timeout here in #379 (which is already merged to master)

selfissued:
https://github.com/w3c/webauthn/pull/523

Issue #233, opened 2016-10-17
detail-level issues in signature format, attestation format(s), attestation statement


equalsJeffH:
There's various detail-level issues in the signature format, attestation format(s), attestation statement sections. Here's at least some of them..

1)  the term `DAA root key` is not defined.

2)  `authenticatorData` aka "authenticator data" -- is inconsistently named in both latter fashions. Plus, its data structure fields are not formally named.

3)  `attestation data` -- this data structure is not formally named (ie no `<dfn>` for it). Plus its fields are not formally named. 

4)  `daaKey` is not defined/described in the **Syntax** subsections of {#packed-attestation} nor {#tpm-attestation}

5)  **Verification procedure** for {#packed-attestation}  does not stipulate behavior if both `x5c` and `daaKey` are present  (throw error?)

6)  **Verification procedure** for {#tpm-attestation} does not stipulate behavior if both `x5c` and `daaKey` are present, or if neither are present  (throw error?)

Issues (2) & (3) lead to trying to denote items in the data structure with imprecise names such as: `ScopedCredentialInfo.attestation.authenticatorData."Attestation data"."public key"`

One way of addressing (2) & (3) is to employ ABNF to formally define both `authenticatorData` and "Attestation data", like so..

<pre>
;--------  Authenticator Data ----------

authenticatorData = rpIdHash flags signCount [attData] [extData]

rpIdHash    = 32OCTET   ; 32 octets

flagByte    = 1OCTET    ; 1 octet
flags       =  flagByte
flags       =/ ED AT RFU TUP

    ED  = BIT   ; Extension Data present (most significant bit)

    AT  = BIT   ; ATtestation data present

    RFU = 5BIT  ; Reserved for Future Use

    TUP = BIT   ; Test of User Presence (least significant bit)

signCount   = 4OCTET    ; 4 octets

;--------  Attestation Data ----------

attData     = AAGUID credIDLen credID pkAlg credPkLen credPk

AAGUID      = 16OCTET   ; 16 octets

credIDLen   = 2OCTET    ; 2 octets

credID      = < sequence of credIDLen octets >

pkAlg       = EC / RSA  ; public key alg & encoding

    EC  = %x0100    ; ANSI X9.62 formatted EC public key

    RSA = %x0102    ; RSA PKCS1 or RSASSA-PSS public key

credPkLen   = 2OCTET    ; 2 octets

credPk      = < sequence of uauthPkLen octets >

;--------  Extension Data ----------

[...]

PR #381, opened 2017-03-15
replaced DAA root key by daaKeyId.  Added proper references

rlin1:
covers the DAA root key aspect as mentioned in #233.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1454,7 +1454,7 @@ authenticator over a set of contextual bindings. Each attestation statement form
     [=hash of the serialized client data=], and returns either:
     - An error indicating that the attestation is invalid, or
     - The attestation type, and the trust path of the attestation. This trust path is either empty (in case of
-        self-attestation), a DAA root key (in the case of Direct Anonymous Attestation), or a set of X.509 certificates.
+        self-attestation), an identifier of a [=DAA Issuer public key=] (in the case of Direct Anonymous Attestation), or a set of X.509 certificates.
 
 The initial list of supported attestation statement formats is in [[#defined-attestation-formats]].
 
@@ -1551,8 +1551,8 @@ in several ways, including:
     attestation keys and attestation certificates.
 
 - A WebAuthn Authenticator can implement [=direct anonymous attestation=] (see [[FIDOEcdaaAlgorithm]]).  Using this scheme,
-    the authenticator generates a blinded attestation signature.  This allows the [RP] to verify the signature using the DAA
-    root key, but the attestation signature doesn't serve as a global correlation handle.
+    the authenticator generates a blinded attestation signature.  This allows the [RP] to verify the signature using the [=DAA Issuer public key=], 
+    but the attestation signature doesn't serve as a global correlation handle.
 
 
 #### Attestation Certificate and Attestation Certificate CA Compromise #### {#ca-compromise}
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -1454,7 +1454,7 @@ authenticator over a set of contextual bindings. Each attestation statement form
     [=hash of the serialized client data=], and returns either:
     - An error indicating that the attestation is invalid, or
     - The attestation type, and the trust path of the attestation. This trust path is either empty (in case of
-        self-attestation), a DAA root key (in the case of Direct Anonymous Attestation), or a set of X.509 certificates.
+        self-attestation), an identifier of a [=DAA Issuer public key=] (in the case of Direct Anonymous Attestation), or a set of X.509 certificates.
```
equalsJeffH: the string used in [[FIDOEcdaaAlgorithm]], for `[=DAA Issuer public key=]`, is "ECDAA-Issuer public key".  We should replicate that herein, yes?

Are all the references to "Direct Anonymous Attestation (DAA)" herein actually in regards to "elliptic curve DAA" ?  if so, that should be made clear, eg using "ECDAA" where we use "DAA" (regardless of case), yes?

equalsJeffH:
also...

in [[!FIDOEcdaaAlgorithm]], these terms..

    DAA-Issuer
    DAA-Verify
    DAA-Sign

..are actually of the form "ECDAA-foo"

I suggest editing this parag from the {#attestation-formats} section..

>: <dfn>Direct Anonymous Attestation</dfn> (DAA)
:: In this case, the Authenticator receives DAA credentials from a single DAA-Issuer. These DAA credentials are used along with
    blinding to sign the [=attestation data=]. The concept of blinding avoids the DAA credentials being misused as global
    correlation handle. WebAuthn supports DAA using elliptic curve cryptography and bilinear pairings, called ECDAA (see
    [[FIDOEcdaaAlgorithm]]) in this specification.

..to be..

: <dfn>Direct Anonymous Attestation</dfn> (DAA)
:: In this case, an Authenticator receives DAA credentials from a given DAA-Issuer. These DAA credentials are used along with blinding to sign the [=attestation data=]. The concept of blinding avoids the DAA credentials being misused as a global correlation handle. WebAuthn employs the DAA variant based on elliptic curve cryptography and bilinear pairings, known as <dfn>ECDAA</dfn>, the details of which are specified in [[FIDOEcdaaAlgorithm]]. See the latter specification for the detailed definitions of <dfn>ECDAA-Issuer</dfn>, as well as the <dfn>ECDAA-Verify</dfn> and <dfn>ECDAA-Sign</dfn> operations. 

..and then have all subsequent occurrences of "DAA-foo" become "[=ECDAA-foo=]", as well as subsequent standalone occurrences of "DAA" become "[=ECDAA=]".

index.bs
rlin1: Yes.  Someone changed it from ECDAA to DAA some time ago.  Will revert back to ECDAA.

index.bs
```diff
@@ -1778,8 +1778,10 @@ implementable by [=authenticators=] with limited resources (e.g., secure element
     :: The elements of this array contain the attestation certificate and its certificate chain, each encoded in X.509 format.
         The attestation certificate must be the first element in the array.
 
-    : daaKey
-    :: The DAA root key. The syntax for `eccPubKey` is defined in [[#sec-attestation-data]].
+    : daaKeyId
+    :: The <dfn>identifier of the DAA Issuer public key</dfn>.  This is the  
+        BigNumberToB encoding of the component "c" of the <dfn>DAA Issuer public key</dfn> 
+        as defined section 3.3, step 3.5 in [[!FIDOEcdaaAlgorithm]].
```
rlin1: The challenge is that that c is a different object!

rlin1: Step 3.5 in section 3.3. is the right one.

rlin1:
Changes DAA to ECDAA since this is the only variant we are using.
@equalsJeffH: please verify that this addresses your concerns.

Issue #249, opened 2016-11-04
Should the "authentication" attribute on Navigator be [SameObject]?


bzbarsky:
And in general, should it have some sort of processing model defined?  Right now nothing defines what this attribute actually returns.  Does the returned value depend on the "this" value, or on the Realm of the attribute getter, or something else?

Presumably this should be [SameObject] hence and the return value should depend only on the "this" value, and should be associated to the same global as the "this" value.

equalsJeffH:
It seems to me that `[SameObject]` is appropriate given the description at https://heycam.github.io/webidl/#SameObject -  thoughts?

selfissued:
Until it is specified what this returns, it can be a source of interop issues.  We need to specify this for the Implementer's Draft.

selfissued:
From discussion on the call, it sounds like we should do this.

equalsJeffH:
@bzbarsky -- might you have a ptr to example text that addresses: "the return value should depend only on the "this" value, and should be associated to the same global as the "this" value." ?

bzbarsky:
Addresses in what sense?

The obvious way to do that is to have it stored in a slot of "this", create when needed, when creating use the relevant global of "this".

mikewest:
I think this is obviated by #384, which moves things to `navigator.credentials`, which is already `[SameObject]`: https://w3c.github.io/webappsec-credential-management/#framework-credential-management

equalsJeffH:
@mikewest 
> I think this is obviated by #384 

agreed, will close with PR #384 merge.

PR #384, opened 2017-03-16
Strawman of an integration between WebAuthn and Credential Management.

mikewest:
@battre: I took a different tack in this patch, leveraging most of the work you've already done.

The main distinction between what we talked about yesterday and what I ended up with here is the value of the `ScopedCredential`'s `id` parameter. I no longer agree with myself from yesterday; it doesn't make much sense to treat this as the `Account`'s ID. I think we should keep the existing relationship between the credential's ID and the "credential identifier" that the authenticator produces. I've just stored that `ArrayBuffer` as an internal property, and turned the `id` getter into a base64url encoded version of that buffer.

Also note that this set of patches focuses on `getAssertion()`, but doesn't make much change to `makeCredential()` (other than the rename). I'd still like to turn that into a dictionary, but this is already a big patch, so...

WDYT?


mikewest:
Whoops! I meant to make this a PR against @battre's fork of this repo, as there are still some open issues to work out. Ah well. :(

So, hello, webauthn! This is a slightly-more-fleshed-out strawman than we discussed on the call last week, but the broad strokes are the same.

equalsJeffH:
see also: 

Splitting "Credential Management"? (Mike West)
https://lists.w3.org/Archives/Public/public-webappsec/2017Mar/0016.html

https://w3c.github.io/webappsec-credential-management/base.html

https://w3c.github.io/webappsec-credential-management/sitebound.html

index.bs
```diff
@@ -357,59 +356,102 @@ The client facilitates these security measures by providing correct origins and
...
+    :   <dfn method>create(accountInformation, cryptoParameters, attestationChallenge, options)</dfn>
+    ::  This method allows a developer to request the User Agent to create a new credential, and persist it to the underlying
+        platform, which may involve data storage managed by the browser or the OS. The user agent will prompt the user to
+        approve this operation. On success, the promise will be resolved with a {{ScopedCredential}} which contains an
+        {{AuthenticatorAttestationResponse}} object. Implementation details are found in [#createCredential].
+
+    :   <dfn method>\[[Retrieve]](options, mediation)</dfn>
+    ::  The {{ScopedCredential}} [=interface object=] defines this internal method to override {{Credential}}'s implementation,
+        which will enable support for assertion generation through {{CredentialsContainer/get()}}. Implementation details are
+        found in [[#getAssertion]].
```
vijaybh: Perhaps it's my lack of knowledge of WebIDL, but I have many questions about this.

How does this actually show up in the author's code? We're saying here that a method in an internal slot of the ScopedCredential object overrides a method of the CredentialsContainer interface. Does this then constitute a monkey patch on section 4.1.1 of Credentials Management? Maybe this should be broken out of here and into a real extension point?

index.bs
```diff
@@ -357,59 +356,102 @@ The client facilitates these security measures by providing correct origins and
...
+    ::  This internal slot contains an identifier for the credential, chosen by the platform with help from the authenticator.
+        This identifier is used to look up credentials for use, and is therefore expected to be globally unique with high
+        probability across all credentials of the same type, across all authenticators. This API does not constrain the format
+        or length of this identifier, except that it must be sufficient for the platform to uniquely select a key. For example,
+        an authenticator without on-board storage may create identifiers that consist of the key material wrapped with a key
+        that is burned into the authenticator.
+
+    :   <dfn method>create(accountInformation, cryptoParameters, attestationChallenge, options)</dfn>
+    ::  This method allows a developer to request the User Agent to create a new credential, and persist it to the underlying
+        platform, which may involve data storage managed by the browser or the OS. The user agent will prompt the user to
```
vijaybh: Re: "persist it to the underlying platform..." - I think it would be more accurate and consistent to say "persist it to an authenticator" which may be an external device or built into the client. I think referring to "data storage" is misleading in either case as it implies general-purpose data stores.

vijaybh:
I have to run now, so I just published a half-done review. I will go through the rest of the PR later (likely this weekend) but please feel free to respond to these comments now if you'd like.

index.bs
```diff
@@ -357,59 +356,102 @@ The client facilitates these security measures by providing correct origins and
...
 
-## <dfn interface>WebAuthentication</dfn> Interface ## {#iface-credential}
+The {{ScopedCredential}} interface inherits from {{Credential}} [[!CREDENTIAL-MANAGEMENT-1]], and contains the attributes that
+are returned to the caller when a new credential is created, or a new assertion is requested.
 
 <xmp class="idl">
     [SecureContext]
-    interface WebAuthentication {
-        Promise<ScopedCredentialInfo> makeCredential(
+    interface ScopedCredential : Credential {
```
equalsJeffH: hm, https://w3c.github.io/webappsec-credential-management/base.html seems to have dropped "Transferable"

fyi/fwiw, part of our webauthn's  rationale for not declaring "transferable" is that making credentials transferable means that they can be postMessage'd to another origin (as we understand it) and that seemed to violate our security & privacy requirements.

mikewest: `Transferable` basically means that when the object is `postMessage`d, it doesn't copy the data, but moves it out of your window. We don't need that behavior for any of the current credential types, and if y'all don't need it either, then I'm happy to drop `Transferable` from the CM spec.

That said, there's nothing stopping a malicious developer from copying or serializing the `ScopedCredential` object and using `postMessage()` to transfer that data to a different window. What's the threat you'd like to defend against?

index.bs
```diff
@@ -357,59 +356,102 @@ The client facilitates these security measures by providing correct origins and
...
-    };
-</pre>
+The Web Authentication API is defined by the union of the Web IDL fragments presented in the following sections. A combined IDL
+listing is given in the [[#idl-index]].
 
+## <dfn interface>ScopedCredential</dfn> Interface ## {#iface-scopedcredential}
 
-## <dfn interface>WebAuthentication</dfn> Interface ## {#iface-credential}
+The {{ScopedCredential}} interface inherits from {{Credential}} [[!CREDENTIAL-MANAGEMENT-1]], and contains the attributes that
+are returned to the caller when a new credential is created, or a new assertion is requested.
```
mikewest: I agree that there's a lot of prose that needs rethinking. This PR really only concentrates on the mechanics right now, and one explicit goal is to work through y'all's thoughts on the confusing bits that remain unaddressed. :)

index.bs
mikewest: I borrowed this text from the current section 4.1.1: https://w3c.github.io/webauthn/#makeCredential. I'm happy to change it to something more accurate, though! I'll do that in a distinct patch.

index.bs
mikewest: > How does this actually show up in the author's code?

It doesn't. The `[[...]]` syntax is what the ECMAScript spec uses for internal properties... slots and methods that exist on an object for explanation purposes, but don't have IDL bindings).

> We're saying here that a method in an internal slot of the ScopedCredential object overrides a method of the CredentialsContainer interface.

No, sorry for the confusion: here, we're overriding `Credential`'s implementation of `[[Retreive]]`. `[[Retrieve]]` is _used_ by `CredentialsContainer/get()`, but isn't monkeying around with `CredentialsContainer`'s implementation. I'll walk through the new flow below:

> Does this then constitute a monkey patch on section 4.1.1 of Credentials Management?

In https://w3c.github.io/webappsec-credential-management/base.html, I'm fiddling around with how `get()` is wired up to specific kinds of credential objects. The existing document is pretty hand-wavey about how it knows that `get({ 'password': true })` is a request that operates on `PasswordCredential` objects. That new document attempts to be a little more rigorous. I'll try to spell it out briefly here:

1.  All `Credential` interface objects have an internal (e.g. not exposed to the web) implementation of `[[Retrieve]](options, mediation)` (https://w3c.github.io/webappsec-credential-management/base.html#credential-internal-methods). 

2.  During `get()`, we walk through all the interface objects that derive from `Credential` directly or indirectly (step 5.1 of https://w3c.github.io/webappsec-credential-management/base.html#abstract-opdef-request-a-credential), and examine their `[[type]]` slot's value. If that value is the same as one of the keys in the dictionary that we passed into `get()`, then we'll grab the interface object's `[[Retrieve]]` method.

3.  If, after walking through all such objects, we've only gotten one `[[Retrieve]]` method, then we'll execute it and return the resulting `Credential` to the developer.

    If we got more than one hit (e.g. the developer passed in `{ 'password': true, 'scoped': { ... } }`), then we'll error out.

It might be easier to explain this on a call. It's a bit convoluted, but I do think it's more clearly explained than the last attempt, at least from a mechanical perspective of what the browser would actually do. :)

mikewest:
I appreciate you taking a look, @vijaybh. I've answered your comments inline, and I'll clean up the nits you've noted later today. Good eye!

index.bs
```diff
@@ -357,59 +356,102 @@ The client facilitates these security measures by providing correct origins and
...
+        probability across all credentials of the same type, across all authenticators. This API does not constrain the format
+        or length of this identifier, except that it must be sufficient for the platform to uniquely select a key. For example,
+        an authenticator without on-board storage may create identifiers that consist of the key material wrapped with a key
+        that is burned into the authenticator.
+
+    :   <dfn method>create(accountInformation, cryptoParameters, attestationChallenge, options)</dfn>
+    ::  This method allows a developer to request the User Agent to create a new credential, and persist it to the underlying
+        platform, which may involve data storage managed by the browser or the OS. The user agent will prompt the user to
+        approve this operation. On success, the promise will be resolved with a {{ScopedCredential}} which contains an
+        {{AuthenticatorAttestationResponse}} object. Implementation details are found in [#createCredential].
```
mikewest: Fixed in https://github.com/w3c/webauthn/pull/384/commits/b1e87ca8f5feb2685dc9d47f033207e5972f86cc.

index.bs
mikewest: Changed in https://github.com/w3c/webauthn/pull/384/commits/b1e87ca8f5feb2685dc9d47f033207e5972f86cc. The bit about prompting might also be something to address, as it's not clear to me whether that's actually a requirement imposed by this spec.

index.bs
```diff
@@ -186,8 +186,7 @@ below and in [[#index-defined-elsewhere]].
 
 : HTML
 :: The concepts of [=relevant settings object=], [=origin=],
-    [=opaque origin=], [=is a registrable domain suffix of or is equal to=], and the {{Navigator}} interface are
-    defined in [[!HTML52]].
+    [=opaque origin=], [=is a registrable domain suffix of or is equal to=].
```
mikewest: Fixed in https://github.com/w3c/webauthn/pull/384/commits/b1e87ca8f5feb2685dc9d47f033207e5972f86cc

index.bs
```diff
@@ -92,8 +89,11 @@ is [=Registration=], where a [=scoped credential=] is created on an [=authentica
 with the present user's account (the account may already exist or may be created at this time). The second is
 [=Authentication=], where the [=[RP]=] is presented with an <em>[=Authentication Assertion=]</em> proving the presence
 and consent of the user who registered the [=scoped credential=]. Functionally, the [=Web Authentication API=] comprises
-two methods (along with associated data structures): {{makeCredential()}} and {{getAssertion()}}. The former is used
-during [=Registration=] and the latter during [=Authentication=].
+a {{ScopedCredential}} which extends the Credential Management API, and infrastructure which allows those credentials to be
```
mikewest: Added in https://github.com/w3c/webauthn/pull/384/commits/b1e87ca8f5feb2685dc9d47f033207e5972f86cc.

index.bs
```diff
@@ -2246,16 +2282,19 @@ authenticator.
...
 
 A [RP] simultaneously requests the use of an extension and sets its client argument by including an entry in the
-{{ScopedCredentialOptions/extensions}} option to the {{makeCredential()}} or {{getAssertion()}} call. The entry key MUST be the
+{{ScopedCredentialOptions/extensions}} option to the {{create()}} or {{CredentialsContainer/get()}} call. The entry key MUST be the
 extension identifier, and the value MUST be the [=client argument=].
 
 <pre class="example" highlight="js">
-    var assertionPromise = credentials.getAssertion(..., /* extensions */ {
-        "webauthnExample_foobar": 42
+    var assertionPromise = navigator.credentials.get({
```
battre: nit: navigator.credentials.get({"scoped": {

also below...

mikewest: Taken care of in https://github.com/w3c/webauthn/pull/384/commits/c48c70bc9ca6183e4e473d4c02fcc5cbd6a6562a.

index.bs
```diff
@@ -362,59 +361,102 @@ The client facilitates these security measures by providing correct origins and
...
-        );
     };
 </xmp>
+<dl dfn-type="attribute" dfn-for="ScopedCredential">
+    :   {{Credential/id}}
+    ::  {{ScopedCredential}} overrides {{Credential}}'s getter for this attribute, returning the [=base64url encoding=] of the
+        data contained in the object's {{ScopedCredential/[[identifier]]}} slot.
+
+    :   {{ScopedCredential/rawID}}
+    ::  This attribute returns the {{ArrayBuffer}} contained in the {{ScopedCredential/[[identifier]]}} internal slot.
```
jyasskin: Note that ArrayBuffers are mutable, so folks will be able to change the value of that internal slot. Web Bluetooth invented a ["read only ArrayBuffer"](https://webbluetoothcg.github.io/web-bluetooth/#read-only-arraybuffer) variant to handle this, but I don't think we've implemented the read-only aspect yet.

index.bs
```diff
@@ -674,23 +723,22 @@ When this method is invoked, the user agent MUST execute the following algorithm
...
-            :: A new {{ArrayBuffer}}, created using |global|'s [=%ArrayBuffer%=], containing the bytes of the returned
-                {{authenticatorData}}
-            : {{AuthenticationAssertion/signature}}
-            :: A new {{ArrayBuffer}}, created using |global|'s [=%ArrayBuffer%=], containing the bytes of the returned
-                {{signature}}
+        1.  [=set/Remove=] |authenticator| from |issuedRequests|.
+        2.  Let |value| be a new {{ScopedCredential}} associated with |global| whose fields are:
+
+            :   {{ScopedCredential/[[identifier]]}}
+            ::  A new {{ArrayBuffer}} containing the bytes of the credential ID returned from the successful
```
jyasskin: Keep the "created using |global|'s [=%ArrayBuffer%=]" phrase.

index.bs
```diff
@@ -703,34 +751,74 @@ During the above process, the user agent SHOULD show some UI to the user to guid
 authorizing an authenticator with which to complete the operation.
 </div>
 
+### <dfn interface>AuthenticatorResponse</dfn> interfaces ### {#iface-authenticatorresponse}
 
-## Information about Scoped Credential (interface <dfn interface>ScopedCredentialInfo</dfn>) ## {#iface-credentialInfo}
+The {{AuthenticatorResponse}} interface and its derieved interfaces represent the data
```
jyasskin: I think you're missing the end of a sentence here.

index.bs
```diff
@@ -703,34 +751,74 @@ During the above process, the user agent SHOULD show some UI to the user to guid
...
+<xmp class="idl">
+  [SecureContext]
+  interface AuthenticatorResponse {
+    readonly attribute ArrayBuffer clientDataJSON;
+  };
+</xmp>
+<dl dfn-for="AuthenticatorResponse">
+    :   <dfn attribute>clientDataJSON</dfn>
+    ::  This attribute contains a representation of the data sent from the client to the authenticator in order to generate
+        a given response. The {{AuthenticatorAttestationResponse}} and {{AuthenticatorAssertionResponse}}
```
jyasskin: Another truncated sentence here.

index.bs
mikewest: This is equally a concern in the status quo API, right? `readonly attribute` doesn't prevent the contents of the `ArrayBuffer` stored in `ScopedCredential/id` from being modified.

If we need that kind of guarantee, then we probably need to chat with @tobie/@bzbarsky/@dominic about WebIDL and ECMAScript bindings.

index.bs
mikewest: Thanks! Fixed in https://github.com/w3c/webauthn/pull/384/commits/fedccedb468282665936264105fac6ee09e577cf.

index.bs
mikewest: Thanks! Fixed in https://github.com/w3c/webauthn/pull/384/commits/fedccedb468282665936264105fac6ee09e577cf.

index.bs
mikewest: Thanks! Fixed in https://github.com/w3c/webauthn/pull/384/commits/fedccedb468282665936264105fac6ee09e577cf.

index.bs
jyasskin: Yep.

tobie: Does ES even let you freeze Typed Arrays?

bzbarsky: Typed arrays are "Integer Indexed Exotic Objects".  Per https://tc39.github.io/ecma262/#sec-integer-indexed-exotic-objects that means their integer-named properties are writable and non-configurable, so you can't make them non-writable.

But that's a red herring anyway, because you can just create a _new_ typed array of your own using the same backing arraybuffer.

If you actually perform a set on a typed array, you end up in https://tc39.github.io/ecma262/#sec-integer-indexed-exotic-objects-set-p-v-receiver and thence in https://tc39.github.io/ecma262/#sec-integerindexedelementset which will do some validation and call https://tc39.github.io/ecma262/#sec-setvalueinbuffer

There are no provisions for readonlyness of arraybuffers and therefore no provisions for readonlyness of typed arrays.  The bluetooth spec is basically unimplementable as written.

Oh, and adding such provisions is likely to cause prohibitive performance regressions for code that uses typed arrays.  Which means JS engine implementors will push back very hard on attempts to add something like this.  They have before.

The right way to expose binary data that you don't want mutated is to expose a copy.

mikewest: I agree with @bzbarsky. I'm also not sure I understand why it's necessary to freeze these buffers in the first place, as they're all already munged copies of things passed into other methods. I haven't looked at the bluetooth usage, but it's not clear to me why you need read-onlyness there either. *shrug*

index.bs
```diff
@@ -191,8 +206,7 @@ below and in [[#index-defined-elsewhere]].
 
 : HTML
 :: The concepts of [=relevant settings object=], [=origin=],
-    [=opaque origin=], [=is a registrable domain suffix of or is equal to=], and the {{Navigator}} interface are
-    defined in [[!HTML52]].
+    [=opaque origin=], and [=is a registrable domain suffix of or is equal to=] are defined in [[!HTML52]].
 
 : Web Cryptography API
 :: The {{AlgorithmIdentifier}} type and the method for [=normalizing an algorithm=] are defined in
```
mikewest: Added in https://github.com/battre/webauthn/commit/f49e915768636bae1d31fd5d74fe8c7cffff1b6b.

index.bs
mikewest: \*cough* To make things more confusing (but hopefully less confusing in the long term!), I've now folded those two documents back into https://w3c.github.io/webappsec-credential-management/. :)

selfissued:
This is the biggest proposed API change of all (for good reasons).  We should therefore definitely decide whether to do this before the Implementer's Draft.

equalsJeffH:
@selfissued 
> This is the biggest proposed API change of all (for good reasons). We should therefore definitely decide whether to do this before the Implementer's Draft.

though it may be tough to get this right ('this' == webauthn + [credman's "Core API"](https://w3c.github.io/webappsec-credential-management/#core)) in the  desired timeframe for the Impl Draft.

equalsJeffH:
I am overall supportive of leveraging credman in the context of WebAuthn. 

I (and perhaps others) have some concern wrt timeframe impact of progressing WebAuthn plus a monolithic CredMan (ie core (nee base) + "password+Fed" (nee 'sitebound')) [see here](https://lists.w3.org/Archives/Public/public-webappsec/2017Apr/0004.html). 

Tho, @mikewest [asserts his confidence that the newly-remerged and more-polished CredMan is able to support "external" specs such as WebAuthn in the near term](https://lists.w3.org/Archives/Public/public-webappsec/2017Apr/0006.html). 

thoughts?

selfissued:
There is conditional WG approval to merge this.  Unless substantial issues are identified by Monday, we will merge it.

equalsJeffH:
in case it is helpful, here is a textual diff of @battre and @mikewest's modifications to the spec (diff'd from the current state of webauthn's master branch):  http://kingsmountain.com/doc/diff/diff-webauthn-battre-cm-api-strawman-f49e915--from--index-master-94a30ff.pdf

index.bs
```diff
@@ -860,55 +967,30 @@ a minimum of friction, e.g., the user will not have to dig around in their pocke
...
     };
 </xmp>
 
-<div dfn-type="dict-member" dfn-for="AssertionOptions">
-    This dictionary is used to supply additional options when generating an assertion. All these parameters are optional.
+<div dfn-type="dict-member" dfn-for="ScopedCredentialRequestOptions">
+    This dictionary is used to supply additional options when generating an assertion. All these parameters are optional, except
+    {{ScopedCredentialRequestOptions/challenge}}, which is required:
+    
+    - The <dfn>challenge</dfn> parameter specifies a challenge that the selected authenticator is expected to sign to produce
```
equalsJeffH: s/is expected to sign to produce the/signs, along with other data, when producing an/

index.bs
```diff
@@ -860,55 +967,30 @@ a minimum of friction, e.g., the user will not have to dig around in their pocke
...
+        the assertion.
 
     - The optional <dfn>timeout</dfn> parameter specifies a time, in milliseconds, that the caller is willing to wait for the
         call to complete. This is treated as a hint, and may be overridden by the platform.
 
     - The optional <dfn>rpId</dfn> parameter specifies the rpId claimed by the caller. If it is omitted, it will be assumed to
-        be equal to the [=origin=] specified by the {{WebAuthentication}} object's [=relevant settings object=].
+        be equal to the [=environment settings object/origin=] specified by the [=current settings object=].
```
equalsJeffH: as noted above, using [=current settings object=] may not be correct, per @bzbarsky and @domenic in issue #271

index.bs
```diff
@@ -927,8 +1009,8 @@ user consent to a specific transaction. The structure of these signatures is def
 </pre>
 
 This is a dictionary containing zero or more extensions as defined in [[#extensions]]. An extension is an additional parameter
-that can be passed to the {{getAssertion()}} method and triggers some additional processing by the client platform and/or the
-authenticator.
+that can be passed to the {{CredentialsContainer/get()}} method as part of a {{ScopedCredentialRequestOptions}} object, and
+triggers some additional processing by the client platform and/or the authenticator.
```
equalsJeffH: s/some additional/additional/

index.bs
```diff
@@ -976,7 +1058,7 @@ following Web IDL.
 
     : <dfn dfn>JSON-serialized client data</dfn>
     :: This is the [=UTF-8 encoding=] of the result of calling the initial value of {{JSON/stringify|JSON.stringify}} on a
-        {{CollectedClientData}} dictionary. To avoid ambiguity, the {{ScopedCredentialInfo}} and {{AuthenticationAssertion}} structures
+        {{CollectedClientData}} dictionary. To avoid ambiguity, the {{AuthenticatorAttestationResponse}} and {{AuthenticatorAssertionResponse}} structures
         contain the actual serializations used by the client to generate them.
```
equalsJeffH: I am thinking that the sentence:
> To avoid ambiguity, the {{AuthenticatorAttestationResponse}} and {{AuthenticatorAssertionResponse}} structures contain the actual serializations used by the client to generate them.

..is confusing and redundant since (per suggestions above) we explain in [[#iface-authenticatorattestationresponse]] and [[#iface-authenticatorassertionresponse]] the reason for and importance of conveying the clientDataJSON intact.  So I suggest deleting the above sentence.

index.bs
```diff
@@ -1038,8 +1094,7 @@ by the caller to select a credential for use.
 </xmp>
 
 This dictionary contains the attributes that are specified by a caller when referring to a credential as an input parameter to
-the {{makeCredential()}} or {{getAssertion()}} method. It mirrors the fields of the {{ScopedCredential}} object returned by
-these methods.
+the {{create()}} or {{CredentialsContainer/get()}} methods. It mirrors the fields of the {{ScopedCredential}} object returned by these methods.
```
equalsJeffH: s/these methods/the latter methods/

index.bs
```diff
@@ -1604,22 +1659,22 @@ should be specified in the attestation certificate itself, so that it can be ver
 
 # [RP] Operations # {#rp-operations}
 
-Upon successful execution of a {{makeCredential()}} or {{getAssertion()}} call, the [RP]'s script receives a
-{{ScopedCredentialInfo}} or {{AuthenticationAssertion}} structure respectively from the client. It must then deliver the
-contents of this structure to the [=[RP]=], using methods outside the scope of this specification. This section describes the
-operations that the [RP] must perform upon receipt of these structures.
+Upon successful execution of a {{create()}} or {{CredentialsContainer/get()}} call, the [RP]'s script receives a {{ScopedCredential}} containing a
```
equalsJeffH: s/call/method invocation/

index.bs
```diff
@@ -1604,22 +1659,22 @@ should be specified in the attestation certificate itself, so that it can be ver
...
+{{AuthenticatorAttestationResponse}} or {{AuthenticatorAssertionResponse}} structure respectively from the client. It must then
```
equalsJeffH: s/respectively/, respectively,/

index.bs
```diff
@@ -1604,22 +1659,22 @@ should be specified in the attestation certificate itself, so that it can be ver
...
+deliver the contents of this structure to the [=[RP]=], using methods outside the scope of this specification. This section
```
equalsJeffH: s/[=[RP]=]/[=[RP]=] server/

index.bs
```diff
@@ -1604,22 +1659,22 @@ should be specified in the attestation certificate itself, so that it can be ver
...
+describes the operations that the [RP] must perform upon receipt of these structures.
 
 
 ## Registering a new credential ## {#registering-a-new-credential}
 
-When requested to register a new credential, represented by a {{ScopedCredentialInfo}} structure, as part of a registration
+When requested to register a new credential, represented by a {{AuthenticatorAttestationResponse}} structure, as part of a registration
```
equalsJeffH: s/When requested to register a new/When registering a new/

index.bs
```diff
@@ -1604,22 +1659,22 @@ should be specified in the attestation certificate itself, so that it can be ver
...
 ceremony, a [RP] MUST proceed as follows:
 
-1. Perform JSON deserialization on the {{ScopedCredentialInfo/clientDataJSON}} field of the {{ScopedCredentialInfo}} object to
+1. Perform JSON deserialization on the {{AuthenticatorResponse/clientDataJSON}} field of the {{AuthenticatorAttestationResponse}} object to
     extract the [=client data=] |C| claimed to have been used for the credential's attestation.
```
equalsJeffH: s/claimed to have been used for the credential's attestation/claimed as collected during the credential creation/

index.bs
equalsJeffH: s/registration ceremony/[=registration ceremony=]/

index.bs
```diff
@@ -1683,20 +1738,21 @@ or it MAY decide to accept the registration, e.g. while deleting the older regis
 
 ## Verifying an authentication assertion ## {#verifying-assertion}
 
-When requested to authenticate a given {{AuthenticationAssertion}} structure as part of an authentication ceremony, the [RP]
-MUST proceed as follows:
+When requested to authenticate a given {{ScopedCredential}} structure (|credential|) as part of an [=authentication=]
```
equalsJeffH: s/When requested to authenticate a given/When verifying a given/

index.bs
```diff
@@ -1683,20 +1738,21 @@ or it MAY decide to accept the registration, e.g. while deleting the older regis
...
+1. Using |credential|'s {{Credential/id}} attribute (or the corresponding {{ScopedCredential/rawID}}, if
+    [=base64url encoding=] is inappropriate for your use case), look up the corresponding credential public key.
 
-2. Let |cData|, |aData| and |sig| denote the {{AuthenticationAssertion/clientDataJSON}},
-    {{AuthenticationAssertion/authenticatorData}} and {{AuthenticationAssertion/signature}} attributes of the given
-    {{AuthenticationAssertion}} structure, respectively.
+2. Let |cData|, |aData| and |sig| denote the value of |credential|'s {{ScopedCredential/response}}'s
+    {{AuthenticatorResponse/clientDataJSON}}, {{AuthenticatorAssertionResponse/authenticatorData}}, and
+    {{AuthenticatorAssertionResponse/signature}} respectively.
+    {{AuthenticatorAssertionResponse}} structure in |credential|'s, respectively.
```
equalsJeffH: "{{AuthenticatorAssertionResponse}} structure in |credential|'s, respectively." appears to be a left-over fragment to delete.

balfanz:
While I appreciate (and share!) the desire to clarify the relationship between CredMan and webauthn, I think this PR would get us further away from, rather than closer to, a converged webauthn spec. It opens up too many questions that aren't open in the current spec. Examples below:

- How requireUserMediation() and the various levels of tests-of-user-presence in webauthn will collide or merge is not clear.

- We're thinking of introducing new methods to webauthn, such as a cancel() method, and a promoteAuthenticatorIfAvailable() method. For each of those, we'd have to think through what they mean for the other type of credentials, and will have a problem if we realize that they don't apply well to non-ScopedCredentials.

- Similar to requireUserMediation(), store() seems to be a no-op for ScopedCredentials. There are proposals for a merged createAndStore() method circulating on the list, but the necessity to even come up, and consider, and discuss, and argue over, and get consensus on, these kinds of new proposals arises only if we accept this PR.

- Create() is a no-op for passwords. In here it's therefore a static method on ScopedCredential. That seems a bit strange. Again, a merged createAndStore() method may rectify this situation, but see above for the problems with that.

More details here: https://lists.w3.org/Archives/Public/public-webauthn/2017Apr/0138.html

equalsJeffH:
As ought to be apparent, my review above was pretty much entirely syntactical and editorial, while Dirk's (@balfanz) [review](https://github.com/w3c/webauthn/pull/384#issuecomment-292658726) is addressing the higher-level functionality and semantics. Dirk makes very good points which I overall agree with, and I'm glad we did not simply merge this PR last Wed as some were wont to do.  Also, I've expressed concerns with getting "this" right in a near-term timeframe ([here](https://github.com/w3c/webauthn/pull/384#issuecomment-291912912) and [here](https://github.com/w3c/webauthn/pull/384#issuecomment-291915159), above). 

Properly aligning the overall "Credential" notion, such that we can appropriately craft both symmetric-key-based and public-key-based methods and thus flows, will obviously take further dedicated thought and effort it seems. See Dirk's detailed thoughts from yesterday for example: https://lists.w3.org/Archives/Public/public-webauthn/2017Apr/0138.html

Some folks are expressing urgency to "get something" onto various browser release trains slated to leave their stations this coming summer.  Is it worth getting something poorly-designed into those browsers? 

[As Dirk noted in his preso](https://docs.google.com/presentation/d/1RyfQS3f-Dk7xU8S6pCSBzWl3jGGGrkF1zWkUypVUnik) during [our Feb-2017 F2F (minutes)](https://www.w3.org/2017/02/13-webauthn-minutes.html#item03), we have some options to consider :

A. Just Rename (slides 8, 9)
( as noted in the F2F minutes, this is to just "'rename' scopedCredential" such that webauthn (WA) does not use the term 'cedential' in its API)

B. Join credman class hierarchy, keep webauthn methods  (slides 10..14)

C. Join credman (CM) class hierarchy, use CM methods (slides 15..18)

Plus there is also the status-quo:

D.  Leave [credman](https://w3c.github.io/webappsec-credential-management/) and [webauthn](https://w3c.github.io/webauthn/) entirely separate for their "level 1" (ie initial version) incarnations (leaves door open to address some sort of merger in level 2 incarnations).

See also Dirk's summary slide.  It seems that this PR #384 is essentially option B with CM.get() and WA.getAssertion() merged (which makes it essentially option C per the latter slide) -- though, there are various unaddressed subtle-but-important aspects to this as Dirk has noted in: https://lists.w3.org/Archives/Public/public-webauthn/2017Apr/0138.html

there are various costs and risks associated with each of the above alternatives that we need to sort out.

mikewest:
I've responded on the thread, thanks @balfanz! When we come to agreement there, I'll happily come back and deal with your editorial comments, @equalsJeffH.

(Is that the right work-mode for the group? It's not really clear to me where y'all expect decisions to be discussed and made: GitHub? The weekly call? Email?)

equalsJeffH:
On 4/10/17, 4:54 AM, @mikewest wrote:
> 
> I've responded on the thread, thanks @balfanz! When we come to
> agreement there, I'll happily come back and deal with your editorial
> comments, @equalsJeffH.

thanks, that sounds fine overall.  note that not all are editorial. there's also some detail-level technical items in there. 

> (Is that the right work-mode for the group? It's not really clear to
> me where y'all expect decisions to be discussed and made: GitHub? The
> weekly call? Email?)

we seem to be sorta all over the map wrt discussions, tho in terms of accepting PRs and merging: our tacit process seems to be to discuss PRs on the calls until there is consensus on the calls and also in reviews (and comments) on github to merge.

index.bs
jyasskin: If a user mutates the `ArrayBuffer` returned by `rawId`, does that change the value returned by `id`? If so (which is what the spec implies to me right now), then `cred.id !== cred.id`. If not, does the result of `id` reflect the initial value of `[[identifier]]` or the value at the point where `id` was first retrieved? I think we wind up needing to store both values eagerly and separately.

index.bs
```diff
@@ -806,8 +914,7 @@ authorizing an authenticator with which to complete the operation.
         complete. This is treated as a hint, and may be overridden by the platform.
 
     - The <dfn>rpId</dfn> parameter explicitly specifies the RP ID that the credential should be associated with. If it is
-        omitted, the RP ID will be set to the [=origin=] specified by the {{WebAuthentication}} object's [=relevant settings
-        object=].
+        omitted, the RP ID will be set to the [=environment settings object/origin=] specified by the [=current settings object=].
```
mikewest: (Fixed in #397)

index.bs
mikewest: Fixed in #397.

domenic:
Drive-by comment: it should be rawId, not rawID: https://w3ctag.github.io/design-principles/#casing-rules

AngeloKai:
I looked through the PR and haven't found problem with it.

index.bs
```diff
@@ -362,59 +379,105 @@ The client facilitates these security measures by providing correct origins and
...
+    ::  This attribute returns the {{ArrayBuffer}} contained in the {{ScopedCredential/[[identifier]]}} internal slot.
+
+    :   <dfn>response</dfn>
+    ::  This attribute contains the [=authenticator=]'s response to the client's assertion or attestation request. If the
+        {{ScopedCredential}} is created in response to {{ScopedCredential/create()}}, this attribute's
+        value will be an {{AuthenticatorAttestationResponse}}, otherwise, the {{ScopedCredential}} was created in response to
+        {{CredentialsContainer/get()}}, and this attribute's value will be an {{AuthenticatorAssertionResponse}}.
+
+    :   <dfn>\[[type]]</dfn>
+    ::  The {{ScopedCredential}} [=interface object=]'s {{Credential/[[type]]}} slot's value is the string "`scoped`".
```
equalsJeffH: thanks :)

index.bs
jyasskin: I think dropping the internal slot makes sense here. It's also not clear to me if we need to keep the `ArrayBuffer` version.

index.bs
```diff
@@ -290,9 +308,9 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
 : <dfn>RP ID</dfn>
 :: An identifier for the [RP] on whose behalf a given registration or authentication ceremony is being performed. Scoped
     credentials can only be used for authentication by the same entity (as identified by RP ID) that created and registered
-    them. By default, the RP ID for a WebAuthn operation is set to the [=origin=] specified by the {{WebAuthentication}}
-    object's [=relevant settings object=]. This default can be overridden by the caller subject to certain restrictions, as
-    specified in [[#makeCredential]] and [[#getAssertion]].
+    them. By default, the RP ID for a WebAuthn operation is set to the [=environment settings object/origin=] specified by the
+    [=current settings object=]. This default can be overridden by the caller subject to certain restrictions, as
```
equalsJeffH: wrt `[=current settings object=]` --   given the language a ways below of:
> Let |callerOrigin| be the [=environment settings object/origin=] specified by this {{ScopedCredential}} object's [=relevant settings object=]. 


..is using `[=current settings object=]` here correct?

index.bs
```diff
@@ -651,8 +716,8 @@ When this method is invoked, the user agent MUST execute the following algorithm
     
     1. [=set/Append=] |authenticator| to |issuedRequests|.
 
-1. Let |promise| be [=a new Promise=]. Return |promise| and start a timer for |adjustedTimeout| milliseconds. Then execute the
-    following steps [=in parallel=]. The [=task source=] for these [=tasks=] is the [=dom manipulation task source=].
+1. Start a timer for |adjustedTimeout| milliseconds. Then execute the following steps [=in parallel=]. The [=task source=] for
```
equalsJeffH: alrightie -- i wonder whether that rationale ought to be captured somewhere in these spec(s)?

index.bs
```diff
@@ -675,63 +740,106 @@ When this method is invoked, the user agent MUST execute the following algorithm
...
+                    {{signature}}
         3. [=set/For each=] remaining |authenticator| in |issuedRequests| invoke the [=authenticatorCancel=] operation on
             |authenticator| and [=set/remove=] it from |issuedRequests|.
-        4. [=Resolve=] |promise| with |value| and terminate this algorithm.
+        4. Return |value| and terminate this algorithm.
     </dd>
     </dl>
 
-1. [=Reject=] |promise| with a {{DOMException}} whose name is "{{NotAllowedError}}".
+1. Return a {{DOMException}} whose name is "{{NotAllowedError}}".
```
equalsJeffH: ACK

index.bs
```diff
@@ -675,63 +740,106 @@ When this method is invoked, the user agent MUST execute the following algorithm
...
 During the above process, the user agent SHOULD show some UI to the user to guide them in the process of selecting and
 authorizing an authenticator with which to complete the operation.
 </div>
 
+### <dfn interface>AuthenticatorResponse</dfn> interfaces ### {#iface-authenticatorresponse}
 
-## Information about Scoped Credential (interface <dfn interface>ScopedCredentialInfo</dfn>) ## {#iface-credentialInfo}
+The {{AuthenticatorResponse}} interface represents the data returned from an [=authenticator=] in response to a client's
+request. This base interface contains common attributes for all response types, while derived interfaces like
+{{AuthenticatorAttestationResponse}} and {{AuthenticatorAssertionResponse}} add attribute specific to a particular kind
```
equalsJeffH: ACK

index.bs
```diff
@@ -870,7 +914,7 @@ use a [=roaming authenticator=] which was originally registered with the [RP] us
 
     :   <dfn>attachment</dfn>
     ::  This member contains authenticator attachment descriptions, which are used as an additional constraint on which
-        authenticators are eligible to participate in a [[#makeCredential]] or [[#getAssertion]] operation. See [[#attachment]]
+        authenticators are eligible to participate in a {{ScopedCredential/create()}} operation. See [[#attachment]]
```
equalsJeffH: `[=attachment=]` is not used in get() ?

index.bs
```diff
@@ -1066,8 +1117,8 @@ by the caller to select a credential for use.
 </xmp>
 
 This dictionary contains the attributes that are specified by a caller when referring to a credential as an input parameter to
-the {{makeCredential()}} or {{getAssertion()}} method. It mirrors the fields of the {{ScopedCredential}} object returned by
-these methods.
+the {{create()}} or {{CredentialsContainer/get()}} methods. It mirrors the fields of the {{ScopedCredential}} object returned by
+latter method.
```
equalsJeffH: s/latter/the latter/

index.bs
equalsJeffH: ACK

index.bs
equalsJeffH: ACK

equalsJeffH:
I've re-reviewed and changed my "changes requested" to "approved"  given our decision on call last Fri 12-Apr-2017 (https://www.w3.org/2017/04/14-webauthn-minutes.html) to "merging in, in order to get the namespaces nailed down, and deal with problems afterwards".

index.bs
mikewest: Probably not. Fixing.

index.bs
mikewest: It's only passed into a `create()` operation, but I suppose it is used in the backend decisions for `get()`? Do we need to pass it into `get()` as well?

index.bs
```diff
@@ -1688,8 +1745,8 @@ ceremony, a [RP] MUST proceed as follows:
 
 13. If the attestation statement |attStmt| verified successfully and is found to be trustworthy, then register the new
     credential with the account that was denoted in the
-    {{WebAuthentication/makeCredential(options)/options}}.{{MakeCredentialOptions/user}} passed to {{makeCredential()}}, by
-    associating it with the credential ID and credential public key contained in |authData|'s [=attestation data=], as
+    {{ScopedCredential/[[Create]](options)/options}}.{{MakeCredentialOptions/user}} passed to {{CredentialsContainer/create()}},
+    by associating it with the credential ID and credential public key contained in |authData|'s [=attestation data=], as
```
vijaybh: After the CM API merge, the implication of the current text is that the RP should take the credential ID directly from the attestation data, and ignore the id in the credential object itself. This seems like the right thing to do and actually a security measure. Should we make this explicit in the text here?

mikewest: They should be the same, shouldn't they? I suppose there's additional confidence that the signed attestation data is more trustworthy...

equalsJeffH: > RP should take the credential ID directly from the attestation data, and ignore the id in the credential object itself. This seems like the right thing to do...

agreed.
> ...and actually a security measure. Should we make this explicit in the text here?

I think so, and perhaps discussed in a sec-cons section.

index.bs
```diff
@@ -619,10 +619,11 @@ authorizing an authenticator.
...
 
 <div link-for-hint="ScopedCredential/[[DiscoverFromExternalSource]](options)">
-This method is used to discover and use an existing scoped credential, with the user's consent. The script optionally specifies
-some criteria to indicate what credentials are acceptable to it. The user agent and/or platform locates credentials matching the
-specified criteria, and guides the user to pick one that the script should be allowed to use. The user may choose not to provide
-a credential even if one is present, for example to maintain privacy.
+The <dfn for="ScopedCredential" method>\[[DiscoverFromExternalSource]](options)</dfn> method is used to discover and use an
+existing scoped credential, with the user's consent. The script optionally specifies some criteria to indicate what credentials
+are acceptable to it. The user agent and/or platform locates credentials matching the specified criteria, and guides the user
+to pick one that the script should be allowed to use. The user may choose not to provide a credential even if one is present,
```
equalsJeffH: s/should be allowed/will be allowed/ ?

index.bs
```diff
@@ -2913,41 +2915,47 @@ The sample code for generating and registering a new key follows:
...
 
-    var userAccountInformation = {
-        rpDisplayName: "Acme",
-        displayName: "John P. Smith",
-        name: "johnpsmith@example.com",
-        id: "1098237235409872",
-        imageURL: "https://pics.acme.com/00/p/aBjjjpqPb.png"
-    };
+    var scoped = {
+      challenge: new TextEncoded().encode("climb a mountain"),
```
equalsJeffH: I have angst wrt this bogus client-side non-random setting of challenge (see #404) -- perhaps we should change it to (if I have it correct)..
```js
var challenge = Uint8Array.from(window.atob("PGifxAoBwCkWkm4b1CiIl5otCphiIh6MijdjbWFjomA="), c=>c.charCodeAt(0))
``` 
for now as we await #404 to be decided?

mikewest: Sure (although that's hideous). ;)

equalsJeffH:
the new changes in the latest commits today look overall ok to me -- thx @mikewest -- i did make a couple detail-level comments.

mikewest:
@equalsJeffH: https://github.com/w3c/webauthn/pull/384/commits/a042dd929fd9090e66c043f1538ac3c72c0e0768 should address your two nits.

When we resolve #404, we'll certainly want to update the other `challenge` examples... https://w3c.github.io/webauthn/#example-33d07a93 in particular. But let's address those separately. :)

equalsJeffH:
@mikewest thanks for your patience & perseverance.  will merge forthwith.  

also, closes #249, improves #404

PR #386, opened 2017-03-21
Align registries draft and WebAuthn draft and address extension issues

selfissued:
This prepares draft-hodges-webauthn-registries for submission as an Internet Draft at the beginning of the IETF meeting next week, aligning the contents of the registries and WebAuthN drafts.  It also clears up some ambiguities in the general-purpose extension semantics, which should enable us to close issues #270 and #295 , among others.

PR #389, opened 2017-03-23
Separated proposed changes to extension semantics from PR #386 and use TypeError, per @jyasskin

selfissued:
Per discussions on the 22-Mar-18 WG call, these semantic changes previously part of the mbj-registries-and-extensions branch proposed as pull request #386 are now in the separate branch mbj-extension-semantics.  Thanks to @jyasskin for suggesting the use of the TypeError exception.


index.bs
```diff
@@ -475,7 +475,8 @@ When this method is invoked, the user agent MUST execute the following algorithm
     1. If |extension| is not supported by this client platform, then [=continue=].
 
     1. Otherwise, let |result| be the result of running |extension|'s [=client processing=] algorithm on |argument|. If the
-        algorithm returned an error, [=continue=].
+        algorithm returned an error, return [=a promise rejected with=] a {{DOMException}} whose name is
+        "{{TypeError}}", and terminate this algorithm.
```
equalsJeffH: this adds signaling of errors from running |extension|'s [=client processing=] algorithm, and aborting of the entire makeCred() operation, where the prior behavior (whether actually good or not) is to ignore errors and go on to the next extension and not abort the overall operation.

Others e.g. @vijaybh @jcjones @jyasskin ought to weigh in on this algorithm change.

index.bs
```diff
@@ -605,7 +606,9 @@ When this method is invoked, the user agent MUST execute the following algorithm
     |extension| → |argument| of <code>{{options}}.{{AssertionOptions/extensions}}</code>:
     1. If |extension| is not supported by this client platform, then [=continue=].
     1. Otherwise, let |result| be the result of running |extension|'s [=client processing=] algorithm on |argument|. If the
-        algorithm returned an error, [=continue=].
+        algorithm returned an error, return [=a promise rejected with=] a {{DOMException}} whose name is
+        "{{TypeError}}", and terminate this algorithm.
```

index.bs
```diff
@@ -2246,6 +2249,13 @@ the [RP] know that the extension was honored by the client. Similarly, any exten
 specify a method of augmenting the [=authenticator data=] to let the [RP] know that the extension was honored by the
 authenticator.
 
+If an extension is designed to enable the [RP] to know whether it was understood and acted on,
+the extension MUST return some data from the client to the [RP] for this purpose.
+If an extension does not otherwise require any result values, it SHOULD be defined as returning a Boolean extension result,
+set to `true` to signify that the extension was understood and processed.
+Likewise, such an extension that is processed by the authenticator that does not otherwise require any result values
+SHOULD also return a Boolean extension result,
+set to `true` to signify that the extension was understood and processed.
```
equalsJeffH: per prior discussion with MikeJ @selfissued, I am inclined to model the above-suggested "the extension was honored" return value as a single bit (or byte) signifying "acknowledged", i.e., a value signifying "ACK" and not as a boolean, because we never intend to return false, and that seems potentially confusing to me.

equalsJeffH:
suggest we retitle this PR to be something like "changes to makeCred() and getAssn() extension handling"

index.bs
selfissued: The choice of "true" as the suggested default extension result was to be parallel with the choice of "true" as the client argument value at https://w3c.github.io/webauthn/#extension-request-parameters.  It would be puzzling to use different conventions in the same cases when the reason for these default values is the same.

index.bs
```diff
@@ -475,7 +475,8 @@ When this method is invoked, the user agent MUST execute the following algorithm
     1. If |extension| is not supported by this client platform, then [=continue=].
 
     1. Otherwise, let |result| be the result of running |extension|'s [=client processing=] algorithm on |argument|. If the
-        algorithm returned an error, [=continue=].
+        algorithm returned an error, return [=a promise rejected with=] a {{DOMException}} whose name is
```
jyasskin: I think you want just "... rejected with=] a {{[TypeError](https://heycam.github.io/webidl/#exceptiondef-typeerror)}}, ..."

index.bs
jyasskin: This makes sense to me, although I don't have a strong opinion.

If an extension is critical for security, the RP needs to check for it in the response regardless, because of the previous "if not supported then continue" step. If the extension is optional, you still want some way for developers to notice that their argument is malformed and so won't work on any browser. Throwing a TypeError here is plausible. Checking the "exts" extension and then for a missing response for supported extensions would also work, but takes a lot more developer effort.

vijaybh: I can understand the argument that a programmer error (such as a badly formed client argument to an extension) should be flagged. However, this has a couple of problems:
1. If the specification of an extension evolves in future (e.g. extension adds a new version and the version number is given in the client argument) then this will either leak data or break an operation that could have been completed correctly by ignoring the extension.
1. Client processing can also fail for a number of other reasons other than malformed arguments (e.g. authenticator selection not finding authenticators of the requested type). In those cases this error becomes misleading and also leaks privacy-sensitive info (which can be used to fingerprint the system).

index.bs
vijaybh: This paragraph seems redundant with the preceding one.

IMO there is no such thing as an extension that is not designed to enable the RP to know if it was understood and acted on. The rest of the paragraph is just recommending returning "true" as a specific way of letting the RP know the extension was honored (the latter part was specified in the previous paragraph). If we do want to add this, it seems like it could be added to the previous paragraph.

index.bs
selfissued: Per your point 1, Vijay, if an extension is updated in an incompatible way, I'd argue that it should instead be a new extension.

Per your point 2, I'm good with clients throwing appropriate errors when they detect errors.

The point of this change is that it's, in general, a good architectural choice to enable errors to be caught and to fail early a the time they are caught.  Ignoring errors makes things hard to debug and allows incorrect code to appear to work, which is almost always a bad thing.  Fail fast.

index.bs
selfissued: OK, I'll restructure this text as you suggest, Vijay.

index.bs
```diff
@@ -2497,12 +2503,17 @@ the credential. It is intended primarily for [RPS] that wish to tightly control
     the available authenticators match a provided AAGUID, the client MUST select an authenticator from among the available
     authenticators to generate the credential.
 
+: Client data
+:: Returns the JSON value `true` to indicate to the RP that the extension was acted upon
+
```
vijaybh: Should specify both key and value. Ideally in such a way that the server does not need to do complicated matching of request and response. May be better here just to include the supplied client argument in ClientData, with the same key.

index.bs
```diff
@@ -2497,12 +2503,17 @@ the credential. It is intended primarily for [RPS] that wish to tightly control
...
 : Authenticator argument
-:: There is no authenticator argument.
+:: none
 
 : Authenticator processing
-:: None.
+:: none
```
vijaybh: Editorial nit: Why omit capitalization and punctuation in this one item when others are capitalized and punctuated?

index.bs
```diff
@@ -2517,6 +2528,9 @@ This [=registration extension=] enables the [RP] to determine which extensions t
 : Client processing
 :: None, except default forwarding of client argument to authenticator argument.
 
+: Client data
+:: Returns the list of supported extensions as a JSON array of extension identifier strings
+
```
vijaybh: ClientData (which must be hashed and fed into the authenticator for signature) cannot depend on AuthenticatorData (which comes out of the authenticator after the signature has been computed). In this case it is better to include no client data since the authenticator data will reflect the option anyways.

index.bs
```diff
@@ -2541,6 +2555,9 @@ This [=registration extension=] and [=authentication extension=] enables use of
 : Client processing
 :: None, except default forwarding of client argument to authenticator argument.
 
+: Client data
+:: Returns a JSON string containing the base64url encoding of the authenticator data
+
```
vijaybh: ClientData (which must be hashed and fed into the authenticator for signature) cannot depend on AuthenticatorData (which comes out of the authenticator after the signature has been computed). TShould be "none".

index.bs
```diff
@@ -2594,6 +2611,10 @@ party.
 : Client processing
 :: None, except default forwarding of client argument to authenticator argument.
 
+: Client data
+:: Returns a JSON object that encodes the location information in the authenticator data as a Coordinates value,
+    as defined by [The W3C Geolocation API Specification](https://dev.w3.org/geo/api/spec-source.html#coordinates_interface).
+
```
vijaybh: ClientData (which must be hashed and fed into the authenticator for signature) cannot depend on AuthenticatorData (which comes out of the authenticator after the signature has been computed). Should be "none".

index.bs
```diff
@@ -2644,6 +2664,9 @@ This [=registration extension=] and [=authentication extension=] enables use of
 : Client processing
 :: None, except default forwarding of client argument to authenticator argument.
 
+: Client data
+:: Returns a JSON array of 3-element arrays of numbers that encodes the factors in the authenticator data
+
```
vijaybh: ClientData (which must be hashed and fed into the authenticator for signature) cannot depend on AuthenticatorData (which comes out of the authenticator after the signature has been computed). Should be "none".

index.bs
```diff
@@ -475,7 +475,7 @@ When this method is invoked, the user agent MUST execute the following algorithm
     1. If |extension| is not supported by this client platform, then [=continue=].
 
     1. Otherwise, let |result| be the result of running |extension|'s [=client processing=] algorithm on |argument|. If the
-        algorithm returned an error, return [=a promise rejected with=] a {{DOMException}} whose name is
+        algorithm returned an error, return [=a promise rejected with=] a
         "{{TypeError}}", and terminate this algorithm.
```
selfissued: I'm fine with your suggestion to terminate the algorithm and propagate the error returned from the client processing.  The main thing I was trying to achieve was to enable errors to be caught and cause the operation to fail at that time, rather than being silently ignored.

index.bs
```diff
@@ -2400,6 +2397,9 @@ error.
     the value of |rpId| for all client processing should be replaced by the
     value of |appid|.
 
+: Client data
+:: Returns the JSON value `true` to indicate to the RP that the extension was acted upon
+
```
selfissued: Might this be clearer if the six standard sections had these titles, rather than the current ones?
 - Client input data
 - Client processing
 - Client output data
 - Authenticator input data
 - Authenticator processing
 - Authenticator output data

What you're mostly talking about I think is client processing, rather than the client output data.  Of course, you could be proposing that a particular extension, the "appid" extension, have a different return value than "true".  That's fine with me.  If so, make a proposal.

index.bs
```diff
@@ -2439,20 +2461,23 @@ prompt string, intended for display on a trusted device on the authenticator.
...
+: Client extension input
 :: A single [=UTF-8 encoded=] string prompt.
 
-: Client processing
-:: None, except default forwarding of client argument to authenticator argument.
+: Client extension processing
+:: None, except creating the authenticator extension input from the client extension input.
+
+: Client extension output
+:: Returns the authenticator extension output string as JSON string
```
jyasskin: Assuming this is the algorithm to convert the authenticator extension output into the client extension output, you'll want to mention that it [=utf-8 decodes=] the CBOR into a DOMString.

index.bs
```diff
@@ -2360,53 +2372,60 @@ The entry key MUST be the extension identifier, and the value MUST be the [=clie
...
+processing MUST specify the method of computing the [=authenticator extension input=] from the [=client extension input=]. For extensions that do not
+require input parameters and are defined as taking a Boolean [=client extension input=] value set to `true`, this method SHOULD consist of
+passing an [=authenticator extension input=] value of `true` (CBOR major type 7, value 21).
 
 Note: Extensions should aim to define authenticator arguments that are as small as possible. Some authenticators communicate
     over low-bandwidth links such as Bluetooth Low-Energy or NFC.
 
 
-## Extending <dfn>client processing</dfn> ## {#extension-client-processing}
+## <dfn>Client extension processing</dfn> ## {#client-extension-processing}
```
equalsJeffH: there's a couple problems with the change in this line:
1) it removes the dfn for "client processing", thus occurrences of `[=client processing=]` are dangling, and, 
2) `<dfn>Client extension processing</dfn>` and `{#client-extension-processing}` both cause bikeshed to render html elements with ids of "client-extension-processing"

index.bs
```diff
@@ -2360,53 +2372,60 @@ The entry key MUST be the extension identifier, and the value MUST be the [=clie
...
-the authenticator argument.
+Likewise, the extension output is represented in the [=client data=] as a CBOR map
+with [=extension identifier=] values as keys, and the [=client extension output=] value of each extension as the value.
 
+Extensions that require authenticator processing MUST define
+the process by which the [=client extension input=] can be used to determine the [=authenticator extension input=] and
+the process by which the [=authenticator extension output=] can be used to determine the [=client extension output=].
 
-## Extending authenticator processing ## {#extension-authenticator-processing}
+## <dfn>Authenticator extension processing</dfn> ## {#authenticator-extension-processing}
```
equalsJeffH: similar issues in this line as in comment above on L2388

selfissued:
Thanks, Jeff.  I've corrected the indexing errors.

index.bs
```diff
@@ -2360,53 +2372,60 @@ The entry key MUST be the extension identifier, and the value MUST be the [=clie
...
+the process by which the [=authenticator extension output=] can be used to determine the [=client extension output=].
 
-## Extending authenticator processing ## {#extension-authenticator-processing}
+## <dfn>Authenticator extension processing</dfn> ## {#sctn-authenticator-extension-processing}
 
-Extensions that define additional authenticator processing may similarly define an [=authenticator data=] value. The value may
+Extensions that define additional authenticator processing define an [=authenticator extension input=] value. The value may
 be any data that can be encoded in CBOR. An authenticator that processes an [=authentication extension=] that defines such a
-value must include it in the [=authenticator data=].
+value must include it in the [=authenticator data=] extensions section.
```
equalsJeffH: it seems that this sentence..
> An authenticator that processes an [=authentication extension=] that defines such a value must include it in the [=authenticator data=] extensions section.

..and this sentence..
> As specified in [[#sec-authenticator-data]], the [=authenticator extension input=] value of each processed extension is included in the
-extended data part of the [=authenticator data=]. This part is a CBOR map, with extension identifiers as keys, and the
+extensions data part of the [=authenticator data=].

..are redundant?

index.bs
```diff
@@ -2212,36 +2216,38 @@ The mechanism for generating scoped credentials, as well as requesting and gener
...
 
 Clients wishing to support the widest possible range of extensions may choose to pass through any extensions that they do not
-recognize to authenticators, generating the authenticator argument by simply encoding the client argument in CBOR. All
+recognize to authenticators, generating the [=authenticator extension input=] by simply encoding the [=client extension input=] in CBOR. All
 WebAuthn extensions MUST be defined in such a way that this implementation choice does not endanger the user's security or
 privacy. For instance, if an extension requires client processing, it could be defined in a manner that ensures such a naïve
-pass-through will produce a semantically invalid authenticator argument, resulting in the extension being ignored by the
+pass-through will produce a semantically invalid [=authenticator extension input=] value, resulting in the extension being ignored by the
 authenticator. Since all extensions are optional, this will not cause a functional failure in the API operation.
+Likewise, clients can choose to produce a [=client extension output=] value for an extension that it does not understand by encoding the
```
equalsJeffH: > the client extension output is created from the authenticator extension output, not from the result of the client processing steps?

It seems the answer is "no, see 1st parag of [[sctn-client-extension-processing]]"

> Does it mean that we need two client processing algorithms, one to process the input, and another to process the output?

that seems to be addressed by the last parag of [[sctn-client-extension-processing]] in combination with section [[#sctn-authenticator-extension-processing]].

we could perhaps have better linking to the latter sections from this opening [[#extensions]] section.

index.bs
```diff
@@ -527,8 +527,9 @@ When this method is invoked, the user agent MUST execute the following algorithm
     |extension| → |argument| of <code>|options|.{{MakeCredentialOptions/extensions}}</code>:
     1. If |extension| is not supported by this client platform, then [=continue=].
 
-    1. Otherwise, let |result| be the result of running |extension|'s [=client processing=] algorithm on |argument|. If the
-        algorithm returned an error, [=continue=].
+    1. Otherwise, let |result| be the result of running |extension|'s [=client extension processing=] algorithm on |argument|. If the
+        algorithm returned an error, return [=a promise rejected with=] either that error or a
```
equalsJeffH: also, this is still fundamentally changing how extensions returning errors are handled, see https://github.com/w3c/webauthn/pull/389#discussion_r107807105 and https://github.com/w3c/webauthn/pull/389#discussion_r109304292.

Is there some way to note the error and [=continue=] with remaining extensions?

selfissued: Since this aspect of the PR still seems to be generating discussion, I'll plan to remove this part and file it as a separate issue, letting us continue discussing it, while applying the changes in the PR that are non-controversial clear improvements over what's there.

index.bs
```diff
@@ -666,8 +667,9 @@ When this method is invoked, the user agent MUST execute the following algorithm
 1. If the {{ScopedCredentialRequestOptions/extensions}} member of |scopedOptions| is [=present=], then [=map/for each=]
     |extension| → |argument| of <code>|scopedOptions|.{{ScopedCredentialRequestOptions/extensions}}</code>:
     1. If |extension| is not supported by this client platform, then [=continue=].
-    1. Otherwise, let |result| be the result of running |extension|'s [=client processing=] algorithm on |argument|. If the
-        algorithm returned an error, [=continue=].
+    1. Otherwise, let |result| be the result of running |extension|'s [=client extension processing=] algorithm on |argument|. If the
```
selfissued: Ditto

index.bs
selfissued: The next commit will make it clear that the client processing steps take in the [=client extension input=] and produce the [=authenticator extension input=] and they also take in the [=authenticator extension output=] and produce the [=client extension output=].

equalsJeffH:
merged it since we were authz'd to do so on the Wed 19-Apr-2017 webauthn call.

index.bs
```diff
@@ -2495,7 +2509,7 @@ error.
 :: `appid`
 
 : Client extension input
-:: A single [=UTF-8 encoded=] string specifying a FIDO |appId|.
+:: A single JSON string specifying a FIDO |appId|.
```
jyasskin: To be nitpicky, since the extension input is a Javascript object, it's a [`{{DOMString}}`](https://heycam.github.io/webidl/#idl-DOMString) rather than a JSON string. JSON is a serialization of many Javascript objects, but it's not the type of things that get passed to web APIs.

No need to fix this before merging, just mentioning it for future patches.

PR #390, opened 2017-03-24
fix webauthn ref in -webauthn-registries to link to webauthn editors' draft

equalsJeffH:
fix webauthn ref in -webauthn-registries to link to webauthn editors' draft

add up-to-date rendered copies of -webauthn-registries-00d


selfissued:
If we're going to check in compiled output, let's at least use a name that's not going to change with every new version.  Therefore, please rename draft-hodges-webauthn-registries-00d.html to draft-hodges-webauthn-registries.html.  After that, the rest of this looks good.

equalsJeffH:
done.

Issue #278, opened 2016-11-04
PublicKeyCredentialRequestOptions and MakeCredentialOptions could both inherit from a dictionary which has their shared members


bzbarsky:
In particular, everything except excludeList/allowList.

selfissued:
Am I correct that this won't change the API, only the way it is described?  If it would change the API, we need to decide this before we declare an Implementer's Draft.

selfissued:
There was agreement on the call to do this.

equalsJeffH:
this may be addressed by the credman alignment

equalsJeffH:
We've done some renaming since this issue was submitted. Presently:
* AssertionOptions -> PublicKeyCredentialRequestOptions
* ScopedCredentialOptions ->  MakeCredentialOptions

..so I will update the title of this issue. 

Also, the alignment of WebAuthn with [CredMan](https://w3c.github.io/webappsec-credential-management/) _did not_ address this issue as I'd been (incorrectly) suspecting. 

PublicKeyCredentialRequestOptions and MakeCredentialOptions do inherit from different dictionaries defined in CredMan. 

Also, PublicKeyCredentialRequestOptions and MakeCredentialOptions continue to share three common members:
```
required BufferSource                challenge;

unsigned long                        timeout;

AuthenticationExtensions             extensions;
```

So, if this is indeed something that we ought to address, then we should define a "PublicKeyCredentialOptions"dictionary, say, whose members are the above three items, and then have PublicKeyCredentialRequestOptions and MakeCredentialOptions inherit from it. 

Also, rpId is common between PublicKeyCredentialRequestOptions and MakeCredentialOptions, but at different depths within the dictionaries.  Is this also something to deal with or is it fine as it is?

See also #488

equalsJeffH:
@bzbarsky @jyasskin -- your thoughts on above https://github.com/w3c/webauthn/issues/278#issuecomment-306321898, should we just do this, and do we do something with rpId or just leave it as-is ?

bzbarsky:
> PublicKeyCredentialRequestOptions and MakeCredentialOptions inherit from different dictionaries defined in CredMan.

I'm not obviously seeing that in https://w3c.github.io/webauthn/ as of today.

> then have PublicKeyCredentialRequestOptions and MakeCredentialOptions inherit from it

You can only inherit from one thing.

[...]

PR #397, opened 2017-04-12
Introduce authenticator response interfaces.

mikewest:
This patch adds an 'AuthenticatorResponse' interface, representing the
generic attributes of responses from authenticators. It then redefines
'ScopedCredentialInfo' and 'AuthenticatorAssertion' to derive from this
interface, and renames them to 'AuthenticatorAttestionResponse' and
'AuthenticatorAssertionResponse' respectively.

These new interfaces are a drop-in replacement for the old interfaces,
no normative changes are intended in this patch, other than the
renaming.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -235,7 +235,7 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
     verification=].
 
 : <dfn>Authentication Assertion</dfn>
-:: The cryptographically signed {{AuthenticationAssertion}} object returned by an [=authenticator=] as the result of a
+:: The cryptographically signed {{AuthenticatorAssertionResponse}} object returned by an [=authenticator=] as the result of a
     [=authenticatorGetAssertion=] operation.
 
 : <dfn>Authenticator</dfn>
@@ -377,14 +377,14 @@ The Web Authentication API is defined by the union of the Web IDL fragments pres
 <xmp class="idl">
     [SecureContext]
     interface WebAuthentication {
-        Promise<ScopedCredentialInfo> makeCredential(
+        Promise<AuthenticatorAttestationResponse> makeCredential(
             RelyingPartyUserInfo                 accountInformation,
             sequence<ScopedCredentialParameters> cryptoParameters,
             BufferSource                         attestationChallenge,
             optional ScopedCredentialOptions     options
         );
 
-        Promise<AuthenticationAssertion> getAssertion(
+        Promise<AuthenticatorAssertionResponse> getAssertion(
             BufferSource                    assertionChallenge,
             optional AssertionOptions       options
         );
@@ -399,7 +399,7 @@ This interface has two methods, which are described in the following subsections
 <div link-for-hint="WebAuthentication/makeCredential(accountInformation, cryptoParameters, attestationChallenge, options)">
 With this method, a script can request the User Agent to create a new credential of a given type and persist it to the
 underlying platform, which may involve data storage managed by the browser or the OS. The user agent will prompt the user to
-approve this operation. On success, the promise will be resolved with a {{ScopedCredentialInfo}} object describing the newly
+approve this operation. On success, the promise will be resolved with a {{AuthenticatorAttestationResponse}} object describing the newly
 created credential.
 
 <div class="note">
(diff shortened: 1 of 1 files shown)


mikewest:
This patch is an attempt to extract one point of agreement between @leshi's `MakeCredentialResponse`/`AssertionResponse` in https://lists.w3.org/Archives/Public/public-webauthn/2017Apr/0157.html, and the pretty similar `AuthenticatorAttestationResponse`/`AuthenticatorAssertionResponse` in my https://github.com/w3c/webauthn/pull/384/commits/6c38d023cb73df5a84a73432ac3d9f19dc736d7c.

I used my names here, both because I an enamored with my own cleverness and because I could easily extract them from that existing commit. I don't think I'd be terribly sad if folks would prefer something more along the lines of @leshi's proposal. :)

AngeloKai:
This will resolve https://github.com/w3c/webauthn/issues/278

jcjones:
On-call, no one has any significant issues with this, but we want to get some detailed reviews before merging. Goal of merging on Friday.

index.bs
```diff
@@ -704,35 +704,85 @@ During the above process, the user agent SHOULD show some UI to the user to guid
...
+        over it.
+
+    :   <dfn>attestationObject</dfn>
+    ::  This attribute contains an [=attestation object=]. The contents of the object are determined by the
+        [=attestation statement format=] used by the authenticator. The object is opaque to, and cryptographically
+        protected against tampering by, the client. It contains the credential's unique identifier, the [=credential
+        public key=], and the attestation statement. It also contains any additional information that the [RP]'s
+        server requires to validate the attestation statement, as well as to decode and validate the bindings of both
+        the client and authenticator data. For more details, see [[#cred-attestation]] as well as
+        [Figure 3](#fig-attStructs).
```
equalsJeffH: might as well fix this entire parag now -- there's various issues with it. Here's a suggested rewrite:

This attribute contains an [=attestation object=], which is opaque to, and cryptographically protected against tampering by, the client. The [=attestation object=] contains both [=authenticator data=] and an attestation statement. The former contains the AAGUID, a unique credential ID, and the [=credential public key=]. The contents of the attestation statement are determined by the [=attestation statement format=] used by the [=authenticator=]. It also contains any additional information that the [RP]'s server requires to validate the attestation statement, as well as to decode and validate the [=authenticator data=] along with the [=JSON-serialized client data=]. For more details, see [[#cred-attestation]] as well as \[Figure 3]\(#fig-attStructs).

index.bs
```diff
@@ -704,35 +704,85 @@ During the above process, the user agent SHOULD show some UI to the user to guid
...
-    [=hash of the serialized client data=] has been computed over it.
-
-    The <dfn>attestationObject</dfn> attribute contains an [=attestation object=]. The contents of this object are
-    determined by the [=attestation statement format=] used by the authenticator. This object is opaque to, and
-    cryptographically protected against tampering by, the client. It contains the credential's unique identifier, [=credential
-    public key=], and attestation statement. It also contains any additional information that the [RP]'s server requires to
-    validate the attestation statement, as well as to decode and validate the bindings of both the client and authenticator
-    data. For more details, see [[#cred-attestation]].
+The {{AuthenticatorAttestationResponse}} interface represents the [=authenticator=]'s response to a client's request
+to create a new [=scoped credential=]. It contains information about the credential that can be used to identify it
```
s/request to create a new/request for the creation of a new/

s/about the credential/about the new credential/

index.bs
```diff
@@ -704,35 +704,85 @@ During the above process, the user agent SHOULD show some UI to the user to guid
...
+        the client and authenticator data. For more details, see [[#cred-attestation]] as well as
+        [Figure 3](#fig-attStructs).
 </div>
 
 
+### Web Authentication Assertion (interface <dfn interface>AuthenticatorAssertionResponse</dfn>) ### {#iface-authenticatorassertionresponse}
+
+The {{AuthenticatorAssertionResponse}} interface represents that [=authenticator=]'s response to a client's request
+to generate an [=authentication assertion=]. This response contains a cryptographic signature proving possession of
+the [=credential private key=], and optionally evidence of [=user consent=] to a specific transaction.
```
The {{AuthenticatorAssertionResponse}} interface represents an [=authenticator=]'s response to a client's request for generation of a new [=authentication assertion=] given the [=[RP]=]'s challenge and optional list of credentials it is aware of.  This response contains a cryptographic signature proving possession of the [=credential private key=], and optionally evidence of [=user consent=] to a specific transaction.

index.bs
```diff
@@ -704,35 +704,85 @@ During the above process, the user agent SHOULD show some UI to the user to guid
...
-## Information about Scoped Credential (interface <dfn interface>ScopedCredentialInfo</dfn>) ## {#iface-credentialInfo}
+[=Authenticators=] respond to relying party requests by returning an object derived from the
+{{AuthenticatorResponse}} interface:
 
 <pre class="idl">
     [SecureContext]
-    interface ScopedCredentialInfo {
-        readonly    attribute ArrayBuffer   clientDataJSON;
-        readonly    attribute ArrayBuffer   attestationObject;
+    interface AuthenticatorResponse {
```
leshi: \<bikeshed>Maybe call this `BaseResponse` since this doesn't actually come from the Authenticator?  And then call the other two `RegistrationResponse` and `AuthenticationResponse` or `MakeCredentialRespone` and `GetAssertionResponse`\</bikeshed>

leshi: I hereby officially don't care about this bikeshed :)

jcjones:
Merge after @equalsJeffH's changes.

index.bs
jyasskin: No opinion on the overall direction of @leshi's suggestion, but in general be careful to keep `interface` names distinct from everything else that might ever appear on the web platform. So, "BaseResponse" is bad because it doesn't identify what kind of response. Including "Authenticator" or "PublicKeyCredential" somewhere in the name would fix it.

mikewest:
I think https://github.com/w3c/webauthn/pull/397/commits/3cdf1bfbf3252e33354549275566be70e73e7f18 addresses the feedback.

equalsJeffH:
@mikewest replied:
> I think 3cdf1bf addresses the feedback.

Yep, it does from my perspective -- thanks for your contributions @mikewest !

PR #398, opened 2017-04-12
Move `getAssertion()`'s `challenge` into `AssertionOptions`

mikewest:
Passing a single dictionary parameter into `getAssertion()` provides
for greater forward compatibility, as new data can be flexibly added
to the method invocation without restructuring the existing
structure. It also helps developers understand what they're passing
in. This is less important for `getAssertion()` than it is for
`makeCredential()`, obviously, but aligning both in a similar
structure seems like a good change to make.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -384,7 +384,7 @@ The Web Authentication API is defined by the union of the Web IDL fragments pres
             optional ScopedCredentialOptions     options
         );
 
-        Promise<AuthenticationAssertion> getAssertion(AssertionRequest request);
+        Promise<AuthenticationAssertion> getAssertion(AssertionOptions options);
     };
 </xmp>
 
@@ -409,9 +409,9 @@ This method takes the following parameters:
     authenticator already has a credential for the specified value of {{RelyingPartyUserInfo/id}} in {{accountInformation}}, and if this
     credential is not listed in the {{ScopedCredentialOptions/excludeList}} member of {{options}}, then after successful
     execution of this method:
-    - Any calls to {{getAssertion()}} that do not specify {{AssertionRequest/allowList}} will not result in the older
+    - Any calls to {{getAssertion()}} that do not specify {{AssertionOptions/allowList}} will not result in the older
         credential being offered to the user.
-    - Any calls to {{getAssertion()}} that specify the older credential in the {{AssertionRequest/allowList}} may also not
+    - Any calls to {{getAssertion()}} that specify the older credential in the {{AssertionOptions/allowList}} may also not
         result in it being offered to the user.
 
 - The <dfn>cryptoParameters</dfn> parameter supplies information about the desired properties of the credential to be created.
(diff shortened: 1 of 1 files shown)


mikewest:
As @vijaybh suggeted in #2 of https://lists.w3.org/Archives/Public/public-webauthn/2017Apr/0158.html, moving `getAssertion()` and `makeCredential()` to dictionary-based invocation rather than list-of-parameter-based invocation seems pretty reasonable. This patch migrates `getAssertion()`.

index.bs
```diff
@@ -892,8 +883,12 @@ user consent to a specific transaction. The structure of these signatures is def
 
 ## Additional options for Assertion Generation (dictionary <dfn dictionary>AssertionOptions</dfn>) ## {#assertion-options}
```
vijaybh: Editorial: Could we remove the word "additional here, since this is now all the options?

mikewest: Good call. Done in https://github.com/w3c/webauthn/pull/398/commits/0039a138e2d27567e7d32f0ddd203748bedb9cbb.

mikewest:
> I'd be inclined to term it "params" or "parameters" but it is not a big deal

`data` or `params` or anything else is fine with me. I'm easy. :)

> My recollection is that we had kept the challenge separate from the truly optional options because the challenge is a must-have.

As I'm sure you noted, the `challenge` is marked as `required`, so this will throw a `TypeError` if it's not present on the dictionary that's passed in. I think the guarantees are much the same as the existing code.

mikewest:
How about `AssertionRequest` in https://github.com/w3c/webauthn/pull/398/commits/3a5fefb7b0603cab4c7a0bdf7476575067561462?

equalsJeffH:
>> My recollection is that we had kept the challenge separate from the truly optional options because the challenge is a must-have.

> As I'm sure you noted, the `challenge` is marked as `required`, so this will throw a `TypeError` if it's not present on the dictionary that's passed in. I think the guarantees are much the same as the existing code.

understood and agreed.

equalsJeffH:
I noted in https://github.com/w3c/webauthn/pull/398#issuecomment-293641860:
>> I'd be inclined to term [options] "params" or "parameters" but it is not a big deal

@mikewest replied:
> data or params or anything else is fine with me. I'm easy. :)
> How about AssertionRequest in 3a5fefb?

Well, in thinking about it let's leave that argument/parameter name as "options" -- "request", as attempted 3a5fefb, seems misleading because the dictionary semantically contains options/arguments/parameters and "options" is shorter. 

So going with the two commits -- https://github.com/w3c/webauthn/pull/398/commits/a9da99223c2da0bb07413af2d07b6aa36236421f and https://github.com/w3c/webauthn/pull/398/commits/0039a138e2d27567e7d32f0ddd203748bedb9cbb -- works for me.

index.bs
```diff
@@ -557,30 +554,24 @@ authorizing an authenticator.
 
 ### Use an existing credential - getAssertion() method ### {#getAssertion}
 
-<div link-for-hint="WebAuthentication/getAssertion(assertionChallenge, options)">
+<div link-for-hint="WebAuthentication/getAssertion(options)">
```
leshi: shouldn't `options` be `request` here?

index.bs
```diff
@@ -557,30 +554,24 @@ authorizing an authenticator.
...
-- The <dfn>assertionChallenge</dfn> parameter contains a challenge that the selected authenticator is expected to sign to
-    produce the assertion.
+<div class="note" dfn-type="argument" dfn-for="WebAuthentication/getAssertion(request)">
+    This method takes the following parameters:
 
-- The optional <dfn>options</dfn> parameter specifies additional options, as described in
-    [[#assertion-options]].
-
-</ul>
+    :   <dfn>options</dfn>
```
leshi: Shouldn't this be `request`?

equalsJeffH: I suggest not naming this `request` and just using `options` or `parameters` here (see https://github.com/w3c/webauthn/pull/398#issuecomment-294038181).

jcjones:
On-call: Drop the 3rd commit, and then clear to merge.

mikewest:
Dropped https://github.com/w3c/webauthn/pull/398/commits/3a5fefb7b0603cab4c7a0bdf7476575067561462 and merged in the changes from #397 (land this one after that one).

equalsJeffH:
LGTM, thanks!

PR #399, opened 2017-04-12
Convert `makeCredential()`'s parameters into a dictionary.

mikewest:
Passing a single dictionary parameter into `makrCredential()` provides
for greater forward compatibility, as new data can be flexibly added
to the method invocation without restructuring the existing
structure. It also helps developers understand what they're passing
in, as each parameter will be labeled.

This patch restructures the data passed into `makeCredential()`
substantially, moving from four parameters to a single dictionary,
and merging some existing types into a simpler structure. Most of
it is straightforward; the only bit I know will be controversial is
dropping `RelyingPartyUserInfo` in favor of two instances of a
simpler `ScopedCredentialEntity` object: one for the RP, one for the
user.

Let's chat about how (un)reasonable this approach might be.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -406,6 +406,11 @@ created credential.
 
 When this method is invoked, the user agent MUST execute the following algorithm:
 
+1. If any of the {{ScopedCredentialEntity/name}} member of {{options}}.{{MakeCredentialOptions/rp}}, the
+    {{ScopedCredentialEntity/name}} member of {{options}}.{{MakeCredentialOptions/user}}, or the {{ScopedCredentialEntity/id}}
+    member of {{options}}.{{MakeCredentialOptions/user}} are [=present|not present=], return [=a promise rejected with=] a
+    {{TypeError}} [=simple exception=].
+
 1. If the {{MakeCredentialOptions/timeout}} member of {{options}} is [=present=], check if its value lies within a
     reasonable range as defined by the platform and if not, correct it to the closest value lying within that range. Set
     |adjustedTimeout| to this adjusted value. If the {{MakeCredentialOptions/timeout}} member of {{options}} is [=present|not
@@ -420,7 +425,8 @@ When this method is invoked, the user agent MUST execute the following algorithm
 1. If the {{ScopedCredentialEntity/id}} member of {{options}}.{{MakeCredentialOptions/rp}} is [=present|not present=], then set
     |rpId| to |callerOrigin|.
 
-   Otherwise:
+    Otherwise:
+
     1. Let |effectiveDomain| be the |callerOrigin|'s [=effective domain=].
     1. If |effectiveDomain| is null, then return [=a promise rejected with=] a {{DOMException}} whose name is
         "{{SecurityError}}" and terminate this algorithm.
@@ -481,7 +487,7 @@ When this method is invoked, the user agent MUST execute the following algorithm
 
 1. [=set/For each=] |authenticator| in |currentlyAvailableAuthenticators|:
     1. Let |excludeList| be a new [=list=].
-    1. [=list/For each=] credential |C| in <code>{{options}}.{{MakeCredentialOptions/exclude}}</code>:
+    1. [=list/For each=] credential |C| in <code>{{options}}.{{MakeCredentialOptions/excludeList}}</code>:
         1. If <code>|C|.{{transports}}</code> [=list/is not empty=], and |authenticator| is connected over a transport not
             mentioned in <code>|C|.{{transports}}</code>, the client MAY [=continue=].
         1. Otherwise, [=list/Append=] |C| to |excludeList|.
@@ -796,17 +802,14 @@ user consent to a specific transaction. The structure of these signatures is def
 
 <xmp class="idl">
     dictionary MakeCredentialOptions {
-        // Relying Party / User Metadata:
         required ScopedCredentialEntity rp;
         required ScopedCredentialEntity user;
 
-        // Attestation configuration:
         required BufferSource                         challenge;
         required sequence<ScopedCredentialParameters> parameters;
 
-        // Optional constraints:
         unsigned long                        timeout;
-        sequence<ScopedCredentialDescriptor> exclude;
+        sequence<ScopedCredentialDescriptor> excludeList;
         Attachment                           attachment;
         AuthenticationExtensions             extensions;
     };
@@ -835,7 +838,7 @@ user consent to a specific transaction. The structure of these signatures is def
         same {{ScopedCredentialEntity/id}}.
 
     :   <dfn>challenge</dfn>
-    ::  This member contains a channelge intended to be used for generating the newly created credential's [=attestation object=].
+    ::  This member contains a challenge intended to be used for generating the newly created credential's [=attestation object=].
 
     :   <dfn>parameters</dfn>
     ::  This member contains information about the desired properties of the credential to be created. The sequence is ordered
@@ -846,7 +849,7 @@ user consent to a specific transaction. The structure of these signatures is def
     ::  This member specifies a time, in milliseconds, that the caller is willing to wait for the call to complete. This is
         treated as a hint, and may be overridden by the platform.
 
-    :   <dfn>exclude</dfn>
+    :   <dfn>excludeList</dfn>
     ::  This member is intended for use by [=[RPS]=] that wish to limit the creation of multiple credentials for the same
         account on a single authenticator. The platform is requested to return an error if the new credential would be created
         on an authenticator that also contains one of the credentials enumerated in this parameter.
--- a/index.bs
+++ b/index.bs
@@ -377,12 +377,7 @@ The Web Authentication API is defined by the union of the Web IDL fragments pres
 <xmp class="idl">
     [SecureContext]
     interface WebAuthentication {
-        Promise<ScopedCredentialInfo> makeCredential(
-            RelyingPartyUserInfo                 accountInformation,
-            sequence<ScopedCredentialParameters> cryptoParameters,
-            BufferSource                         attestationChallenge,
-            optional ScopedCredentialOptions     options
-        );
+        Promise<ScopedCredentialInfo> makeCredential(MakeCredentialOptions options);
 
         Promise<AuthenticationAssertion> getAssertion(
             BufferSource                    assertionChallenge,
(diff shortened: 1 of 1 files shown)


mikewest:
As @vijaybh suggeted in #2 of https://lists.w3.org/Archives/Public/public-webauthn/2017Apr/0158.html, moving getAssertion() and makeCredential() to dictionary-based invocation rather than list-of-parameter-based invocation seems pretty reasonable. This patch migrates `makeCredential()`, and makes some potentially controversial suggestions in doing so. I've kept it separate from the `getAssertion()` migration for that reason.

index.bs
```diff
@@ -889,6 +792,102 @@ user consent to a specific transaction. The structure of these signatures is def
     [[#op-get-assertion]].
 </div>
 
+## Additional options for Attestation (dictionary <dfn dictionary>MakeCredentialOptions</dfn>) ## {#dictionary-makecredentialoptions}
```
vijaybh: Editorial: this isn't "additional options" any more, it's all the options. Maybe say something like "Parameters for credential creation" instead?

mikewest: Picked up in https://github.com/w3c/webauthn/pull/399/commits/6dca154104884ed814f2b0ded74a959bdf28e3ef.

equalsJeffH: yeah, as i noted in review of PR #398 maybe these are not "options" but are parameters, some of which are optional ?

index.bs
```diff
@@ -377,12 +377,7 @@ The Web Authentication API is defined by the union of the Web IDL fragments pres
 <xmp class="idl">
     [SecureContext]
     interface WebAuthentication {
-        Promise<ScopedCredentialInfo> makeCredential(
-            RelyingPartyUserInfo                 accountInformation,
-            sequence<ScopedCredentialParameters> cryptoParameters,
-            BufferSource                         attestationChallenge,
-            optional ScopedCredentialOptions     options
-        );
+        Promise<ScopedCredentialInfo> makeCredential(MakeCredentialOptions options);
```
leshi: \<bikeshed> I like the `request` terminology that you had in PR #398.  Maybe also rename `MakeCredentialOptions` to `MakeRequest` or `CredentialRequest`? \</bikeshed>

index.bs
```diff
@@ -889,6 +792,102 @@ user consent to a specific transaction. The structure of these signatures is def
...
+
+<xmp class="idl">
+    dictionary MakeCredentialOptions {
+        // Relying Party / User Metadata:
+        required ScopedCredentialEntity rp;
+        required ScopedCredentialEntity user;
+
+        // Attestation configuration:
+        required BufferSource                         challenge;
+        required sequence<ScopedCredentialParameters> parameters;
```
leshi: I don't think these are the attestation parameters -- they're the parameters for the type of credential.  We don't have parameters that allow an RP to request a particular type of attestation, though maybe we should.

index.bs
```diff
@@ -889,6 +792,102 @@ user consent to a specific transaction. The structure of these signatures is def
     [[#op-get-assertion]].
 </div>
 
+## Parameters for Credential Creation (dictionary <dfn dictionary>MakeCredentialOptions</dfn>) ## {#dictionary-makecredentialoptions}
+
+<xmp class="idl">
+    dictionary MakeCredentialOptions {
+        // Relying Party / User Metadata:
+        required ScopedCredentialEntity rp;
```
leshi: Interesting, so we now have the same data for an rp and the user.  Cool!

index.bs
```diff
@@ -889,6 +792,102 @@ user consent to a specific transaction. The structure of these signatures is def
...
+        of registered WebAuthn Extensions.
+</div>
+
+### Entity Description ### {#dictionary-scopedcredentialentity}
+
+The {{ScopedCredentialEntity}} dictionary describes a user account or a [=relying party=] with which a credential is associated.
+
+<xmp class="idl">
+    dictionary ScopedCredentialEntity {
+        DOMString id;
```
leshi: In `RelyingPartyUserInfo` some of these were required, why the change?

index.bs
equalsJeffH: But, I do not like the term `request` here because that dict instance is not a "request" in an of itself.  I much prefer just using `options` or `parameters` here (see https://github.com/w3c/webauthn/pull/398#issuecomment-294038181).  If I were to implement, I'd find `request` misleading/confusing.

index.bs
```diff
@@ -443,19 +417,22 @@ When this method is invoked, the user agent MUST execute the following algorithm
     |callerOrigin| is an [=opaque origin=], return [=a promise rejected with=] a {{DOMException}} whose name is
     "{{NotAllowedError}}", and terminate this algorithm.
 
-1. If the {{ScopedCredentialOptions/rpId}} member of {{options}} is [=present|not present=], then set |rpId| to |callerOrigin|.
-    Otherwise:
+1. If the {{ScopedCredentialEntity/id}} member of {{options}}.{{MakeCredentialOptions/rp}} is [=present|not present=], then set
+    |rpId| to |callerOrigin|.
+
+   Otherwise:
```
equalsJeffH: unfortunately, the as-rendered indent of this "otherwise" block is now lost and so it is visually confusing that this is a block of substeps. we need to fix this.

index.bs
```diff
@@ -443,19 +417,22 @@ When this method is invoked, the user agent MUST execute the following algorithm
...
-        return [=a promise rejected with=] a {{DOMException}} whose name is "{{SecurityError}}", and terminate this algorithm.
-    1. Set |rpId| to the {{ScopedCredentialOptions/rpId}}.
+    1. If {{options}}.{{MakeCredentialOptions/rp}}.{{ScopedCredentialEntity/id}} [=is not a registrable domain suffix of and is
+        not equal to=] |effectiveDomain|, return [=a promise rejected with=] a {{DOMException}} whose name is
+        "{{SecurityError}}", and terminate this algorithm.
+    1. Set |rpId| to {{options}}.{{MakeCredentialOptions/rp}}.{{ScopedCredentialEntity/id}}.
 
-1. Let |normalizedParameters| be a new [=list=] whose [=list/items=] are pairs of ScopedCredentialType and a [=dictionary=] type
-    (as returned by [=normalizing an algorithm=]).
+1. Let |normalizedParameters| be a new [=list=] whose [=list/items=] are pairs of {{ScopedCredentialType}} and a [=dictionary=]
```
equalsJeffH: unfortunately, now the as-rendered step numbering is re-started here, rather than properly continuing.

index.bs
equalsJeffH: Hm. the challenge is not a part of attestation per-se -- it provides per-protocol-run roundtrip integrity whether the protocol run is a registration operation or an authn operation.  

ScopedCredentialParameters are params for the creation of the underlying credential key pair, which then is attested.

index.bs
```diff
@@ -889,6 +792,102 @@ user consent to a specific transaction. The structure of these signatures is def
     [[#op-get-assertion]].
 </div>
 
+## Parameters for Credential Creation (dictionary <dfn dictionary>MakeCredentialOptions</dfn>) ## {#dictionary-makecredentialoptions}
```
equalsJeffH: nominally, I think I'm ok with this overall restructuring of the makeCredential parameters/options, tho I have detail-level comments below.

index.bs
```diff
@@ -889,6 +792,102 @@ user consent to a specific transaction. The structure of these signatures is def
...
+        required ScopedCredentialEntity rp;
+        required ScopedCredentialEntity user;
+
+        // Attestation configuration:
+        required BufferSource                         challenge;
+        required sequence<ScopedCredentialParameters> parameters;
+
+        // Optional constraints:
+        unsigned long                        timeout;
+        sequence<ScopedCredentialDescriptor> exclude;
```
equalsJeffH: I would keep the name of this as `excludeList`.  I note that the other PRs do not alter `allowList`'s name.

index.bs
```diff
@@ -889,6 +792,102 @@ user consent to a specific transaction. The structure of these signatures is def
...
+
+    :   <dfn>user</dfn>
+    ::  This member contains data about the user account for which the [=relying party=] is requesting attestation.
+
+        Its value's {{ScopedCredentialEntity/name}} member is required, and contains a friendly name for the user account (e.g.
+        "john.p.smith@example.com", or "John P. Smith").
+
+        Its value's {{ScopedCredentialEntity/id}} member is required, and contains an identifier for the account, specified by
+        the [=relying party=]. This is not meant to be displayed to the user, but is used by the relying party to control the
+        number of credentials - an authenticator will never contain more than one credential for a given relying party under the
```
equalsJeffH: aside: see https://github.com/w3c/webauthn/issues/403

index.bs
```diff
@@ -889,6 +792,102 @@ user consent to a specific transaction. The structure of these signatures is def
...
+    :   <dfn>parameters</dfn>
+    ::  This member contains information about the desired properties of the credential to be created. The sequence is ordered
+        from most preferred to least preferred. The platform makes a best-effort to create the most preferred credential that it
+        can.
+
+    :   <dfn>timeout</dfn>
+    ::  This member specifies a time, in milliseconds, that the caller is willing to wait for the call to complete. This is
+        treated as a hint, and may be overridden by the platform.
+
+    :   <dfn>exclude</dfn>
```
equalsJeffH: s/exclude/excludeList/

index.bs
```diff
@@ -889,6 +792,102 @@ user consent to a specific transaction. The structure of these signatures is def
...
+        Its value's {{ScopedCredentialEntity/name}} member is required, and contains a friendly name for the user account (e.g.
+        "john.p.smith@example.com", or "John P. Smith").
+
+        Its value's {{ScopedCredentialEntity/id}} member is required, and contains an identifier for the account, specified by
+        the [=relying party=]. This is not meant to be displayed to the user, but is used by the relying party to control the
+        number of credentials - an authenticator will never contain more than one credential for a given relying party under the
+        same {{ScopedCredentialEntity/id}}.
+
+    :   <dfn>challenge</dfn>
+    ::  This member contains a channelge intended to be used for generating the newly created credential's [=attestation object=].
```
equalsJeffH: s/channelge/challenge/ :)

need to expand the explanation (but not necessarily now) -- challenge is used in both regstn and authn protocol runs and has security implications. have submitted https://github.com/w3c/webauthn/issues/404

index.bs
equalsJeffH: +1 to @leshi's question "In `RelyingPartyUserInfo` some of these were required, why the change?"

index.bs
leshi: I'm fine with whatever term we use, as long as it's consistent

index.bs
leshi: Suggest making `id` and `name` required.

leshi: On the call, Jeffrey & Mike mentioned that requiring developers to set RP is a bad idea and the params that were required in RelyingPartyUserInfo can be enforced in the algorithm.

index.bs
mikewest: If y'all have suggestions for both `parameters` and `Attestation configuration`, I'm all ears. I don't have strong opinions about either.

index.bs
mikewest: I think we agreed that we're running with `options` for both.

index.bs
mikewest: I think we agreed on the call to drop the comment, and leave the names.

mikewest:
Addressed the comments based on my understanding of what we agreed to in the call via https://github.com/w3c/webauthn/pull/399/commits/805891752a579d84c677f512539c34f05658cf8d, and merged in #398 (and therefore #397) via https://github.com/w3c/webauthn/pull/399/commits/ae33b140cd808f849c9c238d3b703099bc3cd9df. Land those two before you land this one. :)

index.bs
equalsJeffH: i think "drop the comment, and leave the names" is fine for now", thx.

equalsJeffH:
LGTM, thx @mikewest !

Issue #296, opened 2016-11-06
Refine meaning of PublicKeyCredentialType to be "signature & assertion format (and version thereof)"


equalsJeffH:
**Summary:** Refine meaning of ScopedCredentialType to be "signature & assertion format (and version thereof)"

**Background:**
See (short three-message) thread beginning here: https://lists.w3.org/Archives/Public/public-webauthn/2016Jul/0043.html

From the initial message in thread from @equalsJeffH :

Subject: CredentialType (was: TAG review feedback: Align Credential  interface with  Credential Management?

On 7/6/16, 9:05 AM, "Hodges, Jeff" <jeff.hodges@paypal.com> wrote:
> On 7/5/16, 11:59 PM, "Vijay Bharadwaj via GitHub" <sysbot+gh@w3.org>
wrote:
>>- Possibly rename CredentialType to version to make it clearer what
this field signifies (since ScopedCredential is itself arguably a
credential type and ScopedCredentialType is unwieldy)

[ tho Mike West has replied to the original msg, the below is a
tangentially related subitem of the latter ]

the WebAuthentication API <https://w3c.github.io/webauthn/#idl-index> is
only about what we've so far called "ScopedCredentials", which have
various overall characteristics such as utilizing asymmetric key pairs,
plus associated signature and hash algs, and being scoped to RPs based on
eTLD+1.

So I agree with Vijay that we should rename CredentialType. As Vijay noted
(on the call today) that someone(s) have mentioned: perhaps it could be
used to signify assertions' particular construction aka "signature &
assertion format", in similar fashion to "attestation format". I think we
should look into doing that, which inherently incorporates a notion of
"version" because if in the future a particular  "signature & assertion
format" is altered, we can assign the new format a new name.

Also, presently, CredentialType (aka SignatureFormat) and
AlgorithmIdentifier (i.e., the cryptographic parameters) are separate
items..
```
  dictionary ScopedCredentialParameters {
      required CredentialType         type;
      required AlgorithmIdentifier    algorithm;
  };
```
..and, AFAICT, the signature format section [1] of the spec is orthogonal
to the particular crypto key pair type + sig alg, which are specified by
ScopedCredentialParameters.algorithm.

So perhaps we can rename CredentialType to something like
SignatureAndAssertionFormat, and term the present construction as
"ScopedCred1" or something like that..
```
  enum SignatureAndAssertionFormat {
      "ScopedCred1"
  };
```
[...]

PR #407, opened 2017-04-18
Add credential type uaf

rlin1:
***
[Preview](https://s3.amazonaws.com/pr-preview/w3c/webauthn/credential-type-uaf.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/webauthn/c7f1592...5551ad8.html)


jyasskin:
What does it mean to add a credential type instead of just adding a new attestation format?

equalsJeffH:
@selfissued wrote:
> If we're going to be adding credential types, then it seems like we should probably add a Credential Types registry to our registry document. Do you agree with this, @equalsJeffH ?

Yes.  Though, it ought to be called a "signature and assertion format registry", see #296

@jyasskin asked:
> What does it mean to add a credential type instead of just adding a new attestation format?

"credential type" is a misnomer, it really should be named something like "assertionAndSignatureFormat" -- see #296 and #233 

@selfissued proposed:
> it should be possible to add the UAF credential type and attestation format in a separate document, rather than requiring that they be added to the WebAuthn spec. That's the approach that I think we should pursue for this one and PR #408 

I disagree _**in this specific case of UAF**_ because: 
1. there are millions of already-deployed UAF-capable smartphones, which are upgradeable to speak CTAP and thus be usable with WebAuthn-enabled browsers via CTAP (in the nearish term), and, 
2. the amount of normative spec prose to enable handling UAF signature and assertion format is small, and largely consists of changes we will need to introduce anyway in order to prepare the webauthn spec to properly leverage a "signature and assertion format registry". Essentially the webauthn spec will just reference the UAF spec set for the bulk of the normative details. 
3. we have updated the webauthn spec already to handle U2F authenticators for essentially the same reasons.

If someone later desires to register and have WebAuthn support yet another signature-and-assertion-format, then I agree that the approach they should take is to formally register such and specify it in self-contained specs separate from the present WebAuthn spec.

equalsJeffH:
This PR is concise, largely does not affect other portions of the webauthn spec, incorporates some changes that we will need to do any way in  order to have the spec properly handle separately-defined assertion-and-signature-formats.

We ought to refine this PR and #408 appropriately and merge them for WD-06.

Fixes #465

index.bs
```diff
@@ -1216,7 +1216,8 @@ following Web IDL.
 
 <pre class="idl">
     enum PublicKeyCredentialType {
-        "public-key"
+        "public-key",
```
jyasskin: Line up the indentation.

index.bs
```diff
@@ -1499,7 +1500,7 @@ input parameters:
 - A list of credentials acceptable to the [RP] (possibly filtered by the client).
 - Extension data created by the client based on the extensions requested by the [RP].
 
-When this method is invoked, the authenticator must perform the following procedure:
+When this method is invoked on a credential with ScopedCredentialType "ScopedCred", the authenticator must perform the following procedure:
```
jyasskin: "ScopedCredential" doesn't exist anymore.

index.bs
```diff
@@ -1534,6 +1535,17 @@ terminates the operation and returns an error.
 
 If the user refuses consent, the authenticator returns an appropriate error status to the client.
 
+When this method is invoked on a credential with ScopedCredentialType "ScopedCred_FIDO_UAF", the authenticator must perform the procedure as specified in [[FIDO-UAF-Authenticator-Commands]] section 6.3.
```
jyasskin: Format the two "When this method is" sections using `<dl class=switch>` to be clear where the boundary between the two options is. Something like:

```
1. If the quabbledyblargh is:
    <dl class="switch">
    <dt>"public-key"</dt>
    <dd>
    1. Check if all the supplied parameters …
    </dd>
    <dt>"public-key-uaf"</dt>
    <dd>
    1. perform the procedure as specified in [FIDO-UAF-Authenticator-Commands] section 6.3. 
    2. …
    </dd>
    </dl>
```

You'll have to fill in "quabbledyblargh" with some value computed from the arguments to this operation, which don't actually seem to have enough information in them to identify which algorithm to run. So … figure that out. 😃

You'll also have to play with the indentation and blank lines to get Markdown happy with that `<dl>`. I think there are some other examples you can try to imitate.

index.bs
```diff
@@ -1901,8 +1913,8 @@ or it MAY decide to accept the registration, e.g. while deleting the older regis
 
 ## Verifying an authentication assertion ## {#verifying-assertion}
 
-When verifying a given {{PublicKeyCredential}} structure (|credential|) as part of an [=authentication=] [=ceremony=], the [RP]
-MUST proceed as follows:
+When verifying a given {{PublicKeyCredential}} structure (|credential|) with PublicKeyCredentialType <a>public-key</a>
```
jyasskin: Again, format this decision as part of the algorithm. `If |credential|.type is: <dl class="switch"><dt>"public-key"</dt> …`

index.bs
jyasskin: [Section 6.3](https://fidoalliance.org/specs/fido-uaf-v1.1-id-20170202/fido-uaf-authnr-cmds-v1.1-id-20170202.html#sign-command) specifies parameter and response formats, which don't seem to be used here. Could you be clearer about how the inputs get from here into the [6.3.4](https://fidoalliance.org/specs/fido-uaf-v1.1-id-20170202/fido-uaf-authnr-cmds-v1.1-id-20170202.html#command-description-2) algorithm?

index.bs
rlin1: Done.

index.bs
rlin1: Merged the recent terminology changes into this branch.

index.bs
rlin1: Done in commit c7eee4a

rlin1: Done in commit 2d5252f

rlin1:
Nominally addressed all comments (except merging PR #407 with PR #408).

index.bs
```diff
@@ -1499,7 +1500,10 @@ input parameters:
 - A list of credentials acceptable to the [=[RP]=] (possibly filtered by the client).
 - Extension data created by the client based on the extensions requested by the [=[RP]=].
 
-When this method is invoked, the [=authenticator=] must perform the following procedure:
+The [=authenticator=] must perform the following procedure depending on the PublicKeyCredentialType of the related [=credential=]:
```
jyasskin: How does the authenticator find the related credential in order to make this decision? That's not discovered until step 4 of the `"public-key"` steps.

index.bs
```diff
@@ -1524,16 +1528,35 @@ When this method is invoked, the [=authenticator=] must perform the following pr
...
+- On successful completion, the [=authenticator=] returns to the user agent:
+    * The identifier of the credential used to generate the signature.
+    * binary authenticator generated data according to the structure as specified in [[FIDO-UAF-Authenticator-Commands]] section 5.2.2.
+    * The assertion signature.
+
+If the [=authenticator=] cannot find any credential corresponding to the specified [RP] that matches the specified criteria, it
+terminates the operation and returns an error.
 
+If the user refuses consent, the [=authenticator=] returns an appropriate error status to the client.
+    </dd>
```
jyasskin: You're missing a `</dl>` here, discovered by looking at the indentation of https://s3.amazonaws.com/pr-preview/w3c/webauthn/credential-type-uaf.html#op-cancel.

index.bs
```diff
@@ -1942,7 +1967,23 @@ When verifying a given {{PublicKeyCredential}} structure (|credential|) as part
...
+    {{AuthenticatorResponse/clientDataJSON}}, {{AuthenticatorAssertionResponse/authenticatorData}}, and
+    {{AuthenticatorAssertionResponse/signature}} respectively.
+
+3. Perform JSON deserialization on |cData| to extract the [=client data=] |C| used for the signature.
+
+4. Perform the verification as described in [[FIDO-UAF-Protocol]] section 3.5.7.5.
+
+5. If all the above steps are successful, continue with the authentication ceremony as appropriate. Otherwise, fail the
+    authentication ceremony.
+    </dd>
```
jyasskin: Here too.

index.bs
```diff
@@ -1524,16 +1528,35 @@ When this method is invoked, the [=authenticator=] must perform the following pr
...
 
-If the authenticator cannot find any credential corresponding to the specified [=[RP]=] that matches the specified criteria, it
+If the [=authenticator=] cannot find any credential corresponding to the specified [=[RP]=] that matches the specified criteria, it
 terminates the operation and returns an error.
 
 If the user refuses consent, the authenticator returns an appropriate error status to the client.
+    </dd>
+    <dt>"public-key-uaf"</dt>
+    <dd>
+- Follow the procedure as specified in [[FIDO-UAF-Authenticator-Commands]] section 6.3. Using
```
jyasskin: I haven't double-checked the argument passing here.

index.bs
rlin1: Not sure I understand what you are aiming for. 
Looks like you want the first 4 steps under "public-key" to be moved before the "switch" based on on the credential type.  In which case the credential and its type are known.

index.bs
rlin1: added missing /dl

index.bs
rlin1: added missing /dl

index.bs
```diff
@@ -1524,17 +1532,36 @@ When this method is invoked, the [=authenticator=] must perform the following pr
...
```
jyasskin: Step 4 of this algorithm (above) is now "Prompt the user to select a credential from among the above list."
Step 7.1 of https://fidoalliance.org/specs/fido-uaf-v1.1-id-20170202/fido-uaf-authnr-cmds-v1.1-id-20170202.html#command-description-2 (which you're saying to execute here) is "If this authenticator has a user interface and wants to use it for this purpose: Ask the user which of the usernames he wants to use for this operation. Select the related RawKeyHandle and jump to step #8."

So you're specifying that a UAF authenticator needs to ask the user twice. That's obviously wrong.

Please imagine that you don't know anything about FIDO or UAF, but you want to implement a client and authenticator anyway. Write down the steps your implementation would need to follow in order to behave the same as all the other implementations folks are producing.

index.bs
jyasskin: Basically, you need to walk through the steps you're asking folks to implement, one by one, and make sure that they're what you intend and that each step only uses information that previous steps have already said how to compute. I can't write the spec for you or give you a precise recipe for how to fix it because I don't know what you intend to specify.

index.bs
rlin1: Please look into step 7 (in UAF Authenticator Command spec), it specifies that 7.1 will only happen "If number of remaining RawKeyHandles is > 1".  This cannot be the case once the user has selected the preferred credential (note that the credential rawID translates into KeyHandle which is uwrapped/mapped into RawKeyHandle in step #4 in UAF Authenticator command).
So we are NOT running into a second credential selection process - we are just leveraging the fact the in the Web Authentication world the platform asks the user to disambiguate the credential.
So I think we are fine here.

rlin1: BTW: Thanks for your feedback - very helpful.

index.bs
```diff
@@ -1524,17 +1532,36 @@ When this method is invoked, the [=authenticator=] must perform the following pr
...
+    * The caller's RP ID as "AppID"
```
equalsJeffH: RP ID is only a hostname in webauthn (at this time).  In UAF, the RP server supplies the AppID in the OperationHeader.  Perhaps we should use the AppID extension to allow the RP to send its AppID and then pass that down to the authnr?  Though, the AppID extension has issues, see issue #491

leshi:
Please resolve the discussion in #554 before merging.

nadalin:
@leshi This is a next level issue so it will be looked at after we produce the REC

nadalin:
Per 03/07/19 F2F Close

Issue #413, opened 2017-04-20
Account information merge


leshi:
After the merge, the account information changed more than I initially realized.

We lost the "User displayName" and gained an "RP id".  Let make sure this is the right thing to do.

AngeloKai:
We lost an RP id. RP id has always been there in options. The merge didn't change the fact that RP id is optional either.

leshi:
Thanks, Angelo.


We had:
```
dictionary RelyingPartyAccount {
    required DOMString rpDisplayName;
    required DOMString displayName;
    DOMString id; // account identifier, not rp identifier
    DOMString name;
    DOMString imageURL;
```
So an **RP** name.  And **user** display name, name, id, and icon URL.


We have:

```
dictionary ScopedCredentialEntity {
    DOMString id;
    DOMString name;
    USVString icon;
};
```

So an **RP** id, name, icon. And a **User** id, name, icon.

So just by looking at that, the delta is that we gained an **RP** id & icon and lost a **user** display name.

I think you're pointing out that **rpID** was removed from options.  You are right, thanks for pointing that out.

So the correct statement is: we gained an **RP** icon and lost a **user** display name.

equalsJeffH:
to fix the lost user display name, perhaps we ought to define this:
```
dictionary ScopedCredentialUserEntity : ScopedCredentialEntity {
    DOMString displayName;
};
```
..and then alter  MakeCredentialOptions..
```
dictionary MakeCredentialOptions {
    required ScopedCredentialEntity      rp;
    required ScopedCredentialUserEntity  user;
    ...
```

mikewest:
Hrm. I missed the display name. Thanks for catching it!

That said, on the one hand, it's easy to add a display name back, on the other, it's not clear to me how many names we need for a user. :) How do y'all expect this data to be used (by the authenticator, I assume)? Is the distinction between "name" and "display name" relevant for the UI that the authenticator will produce?

equalsJeffH:
I suspect our original intention for `Account.name` (see WD-04: https://www.w3.org/TR/webauthn/#iface-account) was for it to effectively be used for the unique-at-the-RP account identifier (i.e., in essentially the sense of `id` in the SCIM schema: https://tools.ietf.org/html/rfc7643#section-3.1).  Perhaps `name` should be named differently.

`displayName`, as spec'd in WD-04, would AFAIK be used in events such as showing the user a pick list of accounts having creds registered with an RP the user is interacting with. 

Yes?

[...]

PR #423, opened 2017-04-24
Add displayName for the user account

selfissued:
This addresses the missing displayName issue being discussed in issue #413 .  The `name` is actually the display name, so isn't missing.  It's the detailed name that was missing.  This PR restores it.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -478,7 +478,9 @@ When this method is invoked, the user agent MUST execute the following algorithm
 1. Let |options| be the value of |options|["{{CredentialRequestOptions/scoped}}"].
 
 1. If any of the {{ScopedCredentialEntity/name}} member of |options|.{{MakeCredentialOptions/rp}}, the
-    {{ScopedCredentialEntity/name}} member of |options|.{{MakeCredentialOptions/user}}, or the {{ScopedCredentialEntity/id}}
+    {{ScopedCredentialEntity/name}} member of |options|.{{MakeCredentialOptions/user}},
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -478,7 +478,9 @@ When this method is invoked, the user agent MUST execute the following algorithm
 1. Let |options| be the value of |options|["{{CredentialRequestOptions/scoped}}"].
 
 1. If any of the {{ScopedCredentialEntity/name}} member of |options|.{{MakeCredentialOptions/rp}}, the
-    {{ScopedCredentialEntity/name}} member of |options|.{{MakeCredentialOptions/user}}, or the {{ScopedCredentialEntity/id}}
+    {{ScopedCredentialEntity/name}} member of |options|.{{MakeCredentialOptions/user}},
+    the {{ScopedCredentialUserEntity/displayName}} member of |options|.{{MakeCredentialOptions/user}},
+    or the {{ScopedCredentialEntity/id}}
```
equalsJeffH: do we really want to require RPs to pass all this in regardless of the authenticator type ?  This is needed for a multi-factor (aka 1st-factor) authnr, but not necessarily for single-factor (aka 2nd factor) authnr, yes?

selfissued: The problem with not requiring this information is that the RP doesn't know whether the authenticator needs it or not.  I think the PR does what we need, as-is.

equalsJeffH: OK.

PR #425, opened 2017-04-24
Send authenticator extension inputs to authenticators and send client extension outputs to RPs

selfissued:
This corrects some omissions noticed by @jyasskin that meant that authenticator extensions weren't getting their inputs and client extensions weren't getting their outputs to the RP.  Because this affects the API, we should do it now, rather than later, so we won't have future breaking changes.


index.bs
```diff
@@ -521,16 +522,22 @@ When this method is invoked, the user agent MUST execute the following algorithm
...
+    |extensionId| → |clientExtensionInput| of <code>|options|.{{MakeCredentialOptions/extensions}}</code>:
+    1. If |extensionId| is not supported by this client platform or is not a [=registration extension=], then [=continue=].
 
-    1. Otherwise, let |result| be the result of running |extension|'s [=client extension processing=] algorithm on |argument|. If the
+    1. [=list/Append=] the entry |extensionId| → |clientExtensionInput| to |clientExtensions|.
+
+    1. If |extensionId| is not an [=authenticator extension=], then [=continue=].
+
+    1. Let |result| be the base64url encoded representation of 
+        the result of running |extensionId|'s [=client extension processing=] algorithm
```
jyasskin: To base64url encode something, it needs to be an array of bytes, but I don't see a restriction in https://api.csswg.org/bikeshed/?url=https://raw.githubusercontent.com/selfissued/webauthn/mbj-extensions-connective-tissue/index.bs#sctn-client-extension-processing that all extensions return arrays of bytes from their client extension processing. [appid](https://api.csswg.org/bikeshed/?url=https://raw.githubusercontent.com/selfissued/webauthn/mbj-extensions-connective-tissue/index.bs#sctn-appid-extension) seems not to return anything, for example.

index.bs
```diff
@@ -661,15 +674,22 @@ When this method is invoked, the user agent MUST execute the following algorithm
         |effectiveDomain|, return a {{DOMException}} whose name is "{{SecurityError}}", and terminate this algorithm.
     1. Set |rpId| to the {{ScopedCredentialRequestOptions/rpId}}.
 
-1. Let |clientExtensions| be a new [=list=].
+1. Let |clientExtensions| be a new [=list=] and let |authenticatorExtensions| be a new [=list=].
```
jyasskin: Should these be [maps](https://infra.spec.whatwg.org/#maps) now?

index.bs
```diff
@@ -681,8 +701,10 @@ When this method is invoked, the user agent MUST execute the following algorithm
         [=hash of the serialized client data=]
     : {{tokenBinding}}
     :: The [=Token Binding ID=] associated with |callerOrigin|, if one is available.
-    : {{CollectedClientData/extensions}}
+    : {{CollectedClientData/clientExtensions}}
     :: |clientExtensions|
```
jyasskin: Since [`AuthenticationExtensions`](https://api.csswg.org/bikeshed/?url=https://raw.githubusercontent.com/selfissued/webauthn/mbj-extensions-connective-tissue/index.bs#dictdef-authenticationextensions) doesn't have any fields, it doesn't quite work to initialize it from a map. Either, we should add [partial dictionary](https://heycam.github.io/webidl/#dfn-partial-dictionary) definitions for each extension, or maybe it should be a [record](https://heycam.github.io/webidl/#idl-record) instead?

index.bs
```diff
@@ -604,6 +613,10 @@ When this method is invoked, the user agent MUST execute the following algorithm
                 ::  A new {{ArrayBuffer}}, created using |global|'s [=%ArrayBuffer%=], containing the bytes of |clientDataJSON|.
                 :   {{AuthenticatorAttestationResponse/attestationObject}}
                 ::  |attestationObject|
+            :   {{ScopedCredential/clientExtensionResults}}
+            ::  A new {{AuthenticationExtensions}} object containing the [=extension identifier=] → [=client extension output=] entries
+                created by running each extension's [=client extension processing=] algorithm to create the [=client extension outputs=],
+                for each [=client extension=] in {{AuthenticatorResponse/clientDataJSON}}.clientExtensions.
```
selfissued: The authenticator extension outputs are returned in the authenticator data in CBOR map.

index.bs
```diff
@@ -604,6 +613,10 @@ When this method is invoked, the user agent MUST execute the following algorithm
                 ::  A new {{ArrayBuffer}}, created using |global|'s [=%ArrayBuffer%=], containing the bytes of |clientDataJSON|.
                 :   {{AuthenticatorAttestationResponse/attestationObject}}
                 ::  |attestationObject|
+            :   {{ScopedCredential/clientExtensionResults}}
```
selfissued: Sigh.  This was another part of the legacy extensions text that I'd failed to catch and correct earlier.  Hopefully this is the last of it.  I significantly rewrote the offending paragraph, in part, by deleting content that duplicated things said elsewhere.

index.bs
selfissued: It's now as follows, per your suggestion:
typedef record<DOMString, any> AuthenticationExtensions;

index.bs
```diff
@@ -1062,7 +1084,8 @@ following Web IDL.
         required DOMString           origin;
         required DOMString           hashAlg;
         DOMString                    tokenBinding;
-        AuthenticationExtensions     extensions;
+        AuthenticationExtensions     clientExtensions;
```
selfissued: Updated.  Thanks for the catch.

index.bs
```diff
@@ -2328,10 +2327,11 @@ An extension can also be an <dfn>authenticator extension</dfn>, meaning that the
 For [=authenticator extensions=], as part of the [=client extension processing=],
 the client also creates the [=authenticator extension input=] value for each extension (often based on the corresponding [=client extension input=] value),
 and passes them to the authenticator in the {{CredentialsContainer/create()}} call (for [=registration extensions=]) or
-the {{CredentialsContainer/get()}} call (for [=authentication extensions=]). These [=authenticator extension input=] values are passed as name-value
+the {{CredentialsContainer/get()}} call (for [=authentication extensions=]). These [=authenticator extension input=] values are represented in [=CBOR=] and passed as name-value
 pairs, with the [=extension identifier=] as the name, and the corresponding [=authenticator extension input=] as the value. The authenticator,
 in turn, performs additional processing for the extensions that it supports, and returns the [=authenticator extension output=] for each as
 specified by the extension.
+The [=authenticator extension output=] is represented in [=CBOR=].
```
jyasskin: I think I'd prefer to specify the type of data represented by a term by that term's `<dfn>` so that anyone curious about it just follows the link to get to the type, rather than having to read through the whole rest of the spec.

index.bs
```diff
@@ -2432,17 +2432,15 @@ Note: Extensions should aim to define authenticator arguments that are as small
...
 Extensions may define additional processing requirements on the client platform during the creation of credentials or the
-generation of an assertion. In order for the [=[RP]=] to verify the processing took place, or if the processing has a result
-value that the [RP] needs to be aware of, the extension specifies a extension key-value pair to be included in the [=client data=] extensions field,
-which is returned to the [RP] via {{AuthenticatorResponse/clientDataJSON}}.
-
-If any extension processed by a client defines such a
-value, the client includes a dictionary in its [=client data=] with the key {{CollectedClientData/extensions}}. For each such
+generation of an assertion.
+The [=client extension input=] for the extension is used an input to this client processing.
+The [=client extension input=] value may be any value that can be encoded using JSON.
```
jyasskin: I'd rather see the restriction that extensions must define their client extension input as a JSON'able value, by the definition of [client extension input](https://api.csswg.org/bikeshed/?url=https://raw.githubusercontent.com/selfissued/webauthn/mbj-extensions-connective-tissue/index.bs#client-extension-input), rather than in the client processing section.

index.bs
jyasskin: Ok, that's cool. It looks like some places in the spec also expect to return authenticator extension outputs in the client extension output, but because this step calls the "client extension processing" algorithm, which in all the authenticator extensions only specifies how to turn client inputs into authenticator inputs, it doesn't actually get the outputs transformed.

Looking at the text in [Client extension processing](https://api.csswg.org/bikeshed/?url=https://raw.githubusercontent.com/selfissued/webauthn/mbj-extensions-connective-tissue/index.bs#sctn-client-extension-processing) that

> Extensions that require authenticator processing MUST define … the process by which the authenticator extension output can be used to determine the client extension output.

it seems like extensions were supposed to define what this step expects, but for example [txAuthSimple](https://api.csswg.org/bikeshed/?url=https://raw.githubusercontent.com/selfissued/webauthn/mbj-extensions-connective-tissue/index.bs#sctn-simple-txauth-extension), specifies this transformation in the "Client extension output" section instead.

This might tie together better if this step called the extension's [=client extension output=] algorithm, which can use both the client input and authenticator output as inputs?

selfissued: I'd rather handle the (I agree needed) changes to txAuthSimple as a separate PR.  Frankly, I need to review all the actual extensions, but that's not on the top of my queue at the moment.

Elsewhere in the spec it says more about how the client processing works.  I didn't want to repeat it here.  I don't see anyplace that expects to return authenticator extension outputs in client extension outputs.  (There *is* language saying that the client processing rules can *produce* client extension outputs from authenticator extension outputs.)  If there are any ambiguities left, by all means, let me know.

index.bs
```diff
@@ -388,6 +388,7 @@ are returned to the caller when a new credential is created, or a new assertion
     interface ScopedCredential : Credential {
         readonly attribute ArrayBuffer           rawId;
         readonly attribute AuthenticatorResponse response;
+        readonly attribute AuthenticationExtensions clientExtensionResults;
```
jyasskin: Could you add a description of this field after the description of [response](https://api.csswg.org/bikeshed/?url=https://raw.githubusercontent.com/selfissued/webauthn/mbj-extensions-connective-tissue/index.bs#dom-scopedcredential-response)? Those descriptions are non-normative, but they help folks understand the spec. Something like
> This attribute contains the [=client extension output=] produced by the extension's [=client extension processing=]. The [=authenticator extension output=] is returned through the {{AuthenticatorAttestationResponse/attestationObject|AuthenticatorAttestationResponse.attestationObject}} or {{AuthenticatorAssertionResponse/authenticatorData|AuthenticatorAssertionResponse.authenticatorData}} depending on the call that produced this credential.

index.bs
jyasskin: Sounds good.

index.bs
```diff
@@ -2409,31 +2431,29 @@ Note: Extensions should aim to define authenticator arguments that are as small
...
 Extensions may define additional processing requirements on the client platform during the creation of credentials or the
-generation of an assertion. In order for the [=[RP]=] to verify the processing took place, or if the processing has a result
-value that the [RP] needs to be aware of, the extension specifies a extension key-value pair to be included in the [=client data=] extensions field,
-which is returned to the [RP] via {{AuthenticatorResponse/clientDataJSON}}.
-
-If any extension processed by a client defines such a
-value, the client includes a dictionary in its [=client data=] with the key {{CollectedClientData/extensions}}. For each such
+generation of an assertion.
+The [=client extension input=] for the extension is used an input to this client processing.
+Supported [=client extensions=] are recorded as a dictionary in the [=client data=] with the key {{CollectedClientData/extensions}}.
```
jyasskin: s/extensions/clientExtensions/

index.bs
```diff
@@ -2409,31 +2431,29 @@ Note: Extensions should aim to define authenticator arguments that are as small
...
+For each such
 extension, the client adds an entry to this dictionary with the [=extension identifier=] as the key, and the extension's
 [=client extension input=] as the value.
-The [=client extension input=] value may be any value that can be encoded using JSON.
 
-Likewise, the extension output is represented in the [=client data=] as a CBOR map
+Likewise, the [=client extension outputs=] are represented as a dictionary in the {{ScopedCredential/clientExtensionResults}}
 with [=extension identifiers=] as keys, and the <dfn>client extension output</dfn> value of each extension as the value.
+Like the [=client extension input=], the [=client extension ouput=] is a value that can be encoded in JSON.
```
jyasskin: Typo: "ouput".

You can check https://api.csswg.org/bikeshed/?url=https://raw.githubusercontent.com/selfissued/webauthn/mbj-extensions-connective-tissue/index.bs&output=err for this sort of error.

PR #427, opened 2017-04-26
address empty allowlist in 'use existing cred' alg, fixes #387

### LINKED ISSUE #387: Consider empty allowLists
https://w3c.github.io/webauthn/#getAssertion currently says:

> 11. For each authenticator currently available on this platform, perform the following steps:
>     1. Let *credentialList* be a new list.
>     1. If `options.allowList` is not empty, execute a platform-specific procedure to determine which, if any, credentials in `options.allowList` are present on this authenticator by matching with `options.allowList.id` and `options.allowList.type`, and set *credentialList* to this filtered list.
>     1. If *credentialList* is empty then continue.

If `allowList` is empty, `credentialList` remains empty, and every authenticator is skipped, which makes the call less than useful. However, not all examples include an `allowList` (e.g. https://w3c.github.io/webauthn/#sample-extensions), and the [`AssertionOptions`](https://w3c.github.io/webauthn/#dictdef-assertionoptions) definition defaults it to the empty list instead of making it required. #378 also wants to remove the technical need to pass an ID through the `allowList` for every `getAssertion()` call.

I think the spec should either consistently avoid empty `allowList`s or make them work, but I don't have an opinion which.

equalsJeffH:
this addresses empty allowlist in 'use existing cred' alg, fixes #387

please review.

***
[Preview](https://s3.amazonaws.com/pr-preview/w3c/webauthn/jeffh-empty-allowlist-387.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/webauthn/46b3933...7cd5ebf.html)


index.bs
```diff
@@ -696,10 +696,16 @@ When this method is invoked, the user agent MUST execute the following algorithm
 
     1. If <code>|scopedOptions|.{{ScopedCredentialRequestOptions/allowList}}</code> [=list/is not empty=], execute a 
         platform-specific procedure to determine which, if any, credentials in
-        <code>|scopedOptions|.{{ScopedCredentialRequestOptions/allowList}}</code> are present on this |authenticator| by matching with 
-        <code>|scopedOptions|.{{ScopedCredentialRequestOptions/allowList}}.{{ScopedCredentialDescriptor/id}}</code> and 
-        <code>|scopedOptions|.{{ScopedCredentialRequestOptions/allowList}}.{{ScopedCredentialDescriptor/type}}</code>, and set |credentialList| to
-        this filtered list.
+        <code>|scopedOptions|.{{ScopedCredentialRequestOptions/allowList}}</code> are present on this |authenticator| by
```
jyasskin: This is a big step. ;) I might split it up something like:

[Remove "Let credentialList be a new list."]

```html
1. If scopedOptions.allowList:
<dl class="switch">
    : [=list/is empty=]
    :: Let |credentialList| be a new [=list=] consisting of the
        credentials on |authenticator| that are mapped to
        rpId and also have a type within scopedOptions.allowList.
        [Or you could use "match" from below.]
        <span class="note">Note: the procedure to identify these
        credentials is platform-specific.</span>
    : [=list/is not empty=]
    :: Let |credentialList| be a new [=list=] consisting of the
        credentials on |authenticator| that match [Ideally, "match"
        would be a link to a definition of how a credential matches
        a ScopedCredentialDescriptor.] an item in
        scopedOptions.allowList. <span class="note">Note: the
        procedure to identify these credentials is
        platform-specific.</span>
```

equalsJeffH: nice suggestion, thanks :)

jyasskin: D'oh: In the "is empty" case, the "also have a type within scopedOptions.allowList." bit shouldn't be there, since there are no types in an empty allowList.

equalsJeffH:
oops, please ignore commit https://github.com/w3c/webauthn/pull/427/commits/5535f4f8663b3ecb27621afa3b992a4d801871c2 -- it is broken something ugly. Am working to rectify.... pass that torch please... where's my saw?

equalsJeffH:
Ok, please review https://github.com/w3c/webauthn/pull/427/commits/e1ffe61e72d5ead3bd7c0ee5fb3d4067daf2cd08 -- a rendered spec copy is here: http://kingsmountain.com/doc/index-jeffh-empty-allowlist-387-e1ffe61.html

index.bs
```diff
@@ -692,14 +692,33 @@ When this method is invoked, the user agent MUST execute the following algorithm
...
-        this filtered list.
+    <span class="note">Note: the procedure to determine the [=authenticators=] present on a platform is 
+        platform-specific.</span>
+
+    1. Let |credentialList| be a new [=ordered set=].
+
+    1. If <code>|scopedOptions|.{{ScopedCredentialRequestOptions/allowList}}</code>:
+        <dl class="switch">
+            : [=list/is not empty=]
+            :: [=set/For each=] credential |C| in <code>|scopedOptions|.{{ScopedCredentialRequestOptions/allowList}}</code>:
```
jyasskin: *C*s are `ScopedCredentialDescriptor`s, not credentials.

index.bs
```diff
@@ -692,14 +692,33 @@ When this method is invoked, the user agent MUST execute the following algorithm
...
+                1. If there is any credential |X| present on this |authenticator| that:
```
jyasskin: Not for this PR: *X*s are actually not [credentials](https://w3c.github.io/webappsec-credential-management/#concept-credential) themselves either, they're [credential sources](https://w3c.github.io/webappsec-credential-management/#credential-source) (although that term is new and could change if you prefer a better option). This algorithm is about to use authenticatorGetAssertion to generate a credential from *X*.

This is a wide-ranging conflation in this spec, so calling *X*s "credentials" is fine for this PR.

index.bs
```diff
@@ -692,14 +692,33 @@ When this method is invoked, the user agent MUST execute the following algorithm
...
+                        <code>|C|.{{ScopedCredentialDescriptor/type}}</code>,
+
+                    then [=set/append=] |X| to |credentialList|.
+
+                <span class="note">Note: the procedures to determine which credentials present on |authenticator| map to 
+                    |rpId|, and to match credentials present on |authenticator| with values in
+                    {{ScopedCredentialRequestOptions/allowList}}, are platform-specific.</span>
+
+            : [=list/is empty=]
+            :: If there is any credential |X| present on this |authenticator| that is mapped to |rpId| then [=set/append=] |X|
```
jyasskin: This seems to assume that there will only be one credential on an authenticator for any given |rpId|. If that's not true, then it should probably add all such credentials to |credentialList| so that authenticatorGetAssertion can select from all of them in the next step.

index.bs
```diff
@@ -692,14 +692,33 @@ When this method is invoked, the user agent MUST execute the following algorithm
...
+    1. If <code>|scopedOptions|.{{ScopedCredentialRequestOptions/allowList}}</code>:
+        <dl class="switch">
+            : [=list/is not empty=]
+            :: [=set/For each=] credential |C| in <code>|scopedOptions|.{{ScopedCredentialRequestOptions/allowList}}</code>:
+                1. If there is any credential |X| present on this |authenticator| that:
+                    - is mapped to |rpId|, and,
+                    - matches with <code>|C|.{{ScopedCredentialDescriptor/id}}</code> and 
+                        <code>|C|.{{ScopedCredentialDescriptor/type}}</code>,
+
+                    then [=set/append=] |X| to |credentialList|.
```
jyasskin: Being clearer about what kind of thing *credentialList* holds made me notice that the next step, "If `C.transports` is not empty", doesn't have the right types. `.transports` is a field on `ScopedCredentialDescriptor`, not arbitrary credentials. The next step doesn't actually use the credential for calling authenticatorGetAssertion, just the set of credentials that match `allowList`. And does it really intend to call authenticatorGetAssertion multiple times on the same authenticator, or would it rather call it once per authenticator or maybe authenticator/transport pair, and let the authenticator choose the credential from the credential list?

index.bs
equalsJeffH: d'oh. thanks, fixing...

index.bs
equalsJeffH: hm, ok, thanks for the heads-up.  I've submitted issue #430

index.bs
selfissued: This seems like a clear improvement that should be merged in.

jyasskin: Yep, a separate PR is fine with me, and that separate PR probably isn't at the top of the priority queue.

index.bs
equalsJeffH: > The next step doesn't actually use the credential for calling authenticatorGetAssertion, just the set of credentials that match allowList

well, if `allowList` is empty, then it calls authenticatorGetAssertion with the set of `ScopedCredentialDescriptor` on that |authenticator| that are associated with rpId, yes?  I.e.,[ |credentialList| is passed to authenticatorGetAssertion](https://w3c.github.io/webauthn/#createCredential). [authenticatorGetAssertion in CTAP takes a sequence](https://fidoalliance.org/specs/fido-v2.0-rd-20161004/fido-client-to-authenticator-protocol-v2.0-rd-20161004.html#authenticatorgetassertion) for that arg (updated CTAP spec will have the type corrected :)

> does it really intend to call authenticatorGetAssertion multiple times on the same authenticator, or would it rather call it once per authenticator or maybe authenticator/transport pair, and let the authenticator choose the credential from the credential list?

yeah, good catch and good questions -- as-written it seems broken.  [authenticatorGetAssertion in CTAP takes a sequence](https://fidoalliance.org/specs/fido-v2.0-rd-20161004/fido-client-to-authenticator-protocol-v2.0-rd-20161004.html#authenticatorgetassertion) for that arg -- so maybe what we should do here is "for each unique `ScopedCredentialDescriptor.tranport` value found in |credentialDescirptorList|, call |authnr| with a sequence of `ScopedCredentialDescriptor` who all have the same `transport` value."

index.bs
equalsJeffH: Yep, another good catch, thanks.

equalsJeffH: btw, I have fixes to all of the above in-flight to be added as commits to this PR.   am just waiting for thoughts on https://github.com/w3c/webauthn/pull/427#discussion_r113978961

index.bs
jyasskin: Semantically, a [`ScopedCredentialDescriptor`](https://w3c.github.io/webauthn/#dictdef-scopedcredentialdescriptor) with no transports is equivalent to one that lists [all three](https://w3c.github.io/webauthn/#enumdef-transport), right? Remember to deal with that in the text.

Is it possible for an authenticator to be attached via more than one transport? Step 18 in [[[Create]]](https://w3c.github.io/webauthn/#createCredential) seems to assume not. Maybe the client would just treat it as two separate authenticators in that case? If we can assume that an authenticator has exactly one transport, then maybe you can replace 
> For each credential C in scopedOptions.[allowList](https://api.csswg.org/bikeshed/?url=https://raw.githubusercontent.com/w3c/webauthn/jeffh-empty-allowlist-387/index.bs#dom-scopedcredentialrequestoptions-allowlist):

with
> For each credential C in scopedOptions.[allowList](https://api.csswg.org/bikeshed/?url=https://raw.githubusercontent.com/w3c/webauthn/jeffh-empty-allowlist-387/index.bs#dom-scopedcredentialrequestoptions-allowlist) such that either C.[transports](https://api.csswg.org/bikeshed/?url=https://raw.githubusercontent.com/w3c/webauthn/jeffh-empty-allowlist-387/index.bs#dom-scopedcredentialdescriptor-transports) is empty or *authenticator*'s transport appears in C.[transports](https://api.csswg.org/bikeshed/?url=https://raw.githubusercontent.com/w3c/webauthn/jeffh-empty-allowlist-387/index.bs#dom-scopedcredentialdescriptor-transports):

And then the iteration over *credentialList* could be replaced by "Invoke the authenticatorGetAssertion operation on authenticator, with rpId, clientDataHash, credentialList, and clientExtensions as parameters."

@kpaulh @AngeloKai what are your implementations actually doing here? Do you deal with authenticators that have multiple transports and `allowList`s that constrain their credentials' transports?

equalsJeffH: Hm. `{{Transport}}` values are hints, and  `{{ScopedCredentialDescriptor}}.transports` is a `sequence<Transport>`, so we have so far assumed that an authenticator MAY be available over more than one transport, and that the client MAY select one such transport to use in an attempt to invoke `authenticatorGetAssertion` on |authenticator|. 

wrt..
> Semantically, a ScopedCredentialDescriptor with no transports is equivalent to one that lists all three...

..hm, I suppose so, modulo "using local configuration knowledge of the appropriate transport to use with |authenticator|"...

equalsJeffH:
Further surgery in https://github.com/w3c/webauthn/pull/427/commits/c2126831b0ef60b049a4da9f1ecc0aee8475901a -- hopefully in correct direction...

index.bs
```diff
@@ -692,44 +692,75 @@ When this method is invoked, the user agent MUST execute the following algorithm
...
+            :: [=set/For each=] {{ScopedCredentialDescriptor}} |C| in
+                <code>|scopedOptions|.{{ScopedCredentialRequestOptions/allowList}}</code>:
+                1. For each credential |X| present on this |authenticator| that:
                     - is mapped to |rpId|, and,
                     - matches with <code>|C|.{{ScopedCredentialDescriptor/id}}</code> and 
                         <code>|C|.{{ScopedCredentialDescriptor/type}}</code>,
 
-                    then [=set/append=] |X| to |credentialList|.
+                    then:
+                        1. let |Y| be a new {{ScopedCredentialDescriptor}}.
```
jyasskin: It seems like you could just append |C| to |credentialDescriptorList| here. You could copy it if you're worried about modifications later, but dictionaries are already copied on their way into IDL algorithms. And |Y| differs from |C| in its value for `transports`, but it seems like you'd rather use |C|'s.

index.bs
```diff
@@ -692,44 +692,75 @@ When this method is invoked, the user agent MUST execute the following algorithm
...
+                        1. [=set/append=] |Y| to |credentialDescriptorList|.
 
             : [=list/is empty=]
-            :: If there is any credential |X| present on this |authenticator| that is mapped to |rpId| then [=set/append=] |X| 
-                to |credentialList|.
+            :: For each credential |X| present on this |authenticator| that is mapped to |rpId| then:
+
+                1. let |Y| be a new {{ScopedCredentialDescriptor}}.
+
+                1. Set |Y|'s members according to the corresponding properties of |X|.
```
jyasskin: This looks good, except it's unclear about |Y|.transports, since |X| doesn't have any. I suspect you want |Y|.transports to be empty? Then I'd spell this:
> [=set/Append=]  «[ "id" → |X|'s credential ID, "type" → |X|'s type ]» to |credentialDescriptorList|.

index.bs
```diff
@@ -692,44 +692,75 @@ When this method is invoked, the user agent MUST execute the following algorithm
...
+                    <code>|credentialsPerTransportMap|[|T|]</code>.
+
+                : [=list/is empty=]
+                :: [=list/For each=] value |T| of {{Transport}}, [=set/append=] |C| to
+                    <code>|credentialsPerTransportMap|[|T|]</code>.
+            </dl>
+
+    1. If |credentialsPerTransportMap|
+        <dl class="switch">
+            : [=map/size=] is not 0
```
jyasskin: I think we're guaranteed that the size isn't 0, since |credentialDescriptorList| isn't empty, and any |C| leads to adding a value to |credentialsPerTransportMap|.

You can write
1. Assert: The [=map/size=] of |credentialsPerTransportMap| is not 0.

to get nice Bikeshed formatting of that guarantee.

index.bs
```diff
@@ -692,44 +692,75 @@ When this method is invoked, the user agent MUST execute the following algorithm
...
+            :: the client SHOULD:
+                1. Let |transportHintsList| be the result of [=map/getting the keys=] of |credentialsPerTransportMap|.
+                1. Set |transport| to one of the values in |transportHintsList|. The client MAY use local configuration
```
jyasskin: I think this means that an allowList of [{id: 1, transports: ["usb"]}, {id: 2, transports: ["nfc"]}] will call authenticatorGetAssertion on only one of credentials 1 or 2. I suspect that's not what was intended.

equalsJeffH:
@jyasskin 
> I think it probably makes sense to give up on getting the transport bits right in this PR. 

to be clear, are you suggesting to not merge this PR at this time, and wait for feedback from developers before working on it further?  that's ok by me.  thanks much for your comments. 

/cc @kpaulh @AngeloKai @jcjones

selfissued:
When thinking about merge/no-merge decisions, I'd have us ask the question "Is what's in the PR better than what's in the master or is it in some ways worse?"  If it's strictly better, I'd do the merge, even if we think the functionality might be subject to future feedback from implementers.  If it demonstrably makes some things worse, that's an argument to wait.  But if it only makes things better, I'd rather that we incorporated the good work by @jyasskin and @equalsJeffH and not leave it languishing.

equalsJeffH:
@jyasskin and other implementors will have to say whether this actually is any better than the prior definitely broken-ness...

jyasskin:
I was thinking "stop trying to get transport right, and just merge the PR as it is, or as it was before we started messing with transport."

selfissued:
Jeff, can you address the merge conflicts with this one?  Then I'd be happy to do the merge.

equalsJeffH:
please lets not merge until Mon so i have time to digest @jyasskin comments and figure out which commit(s) to merge.

index.bs
equalsJeffH: > it's unclear about |Y|.transports, since |X| doesn't have any. I suspect you want |Y|.transports to be empty?

Hm, I suppose I was/am imagining that the platform-specific operation that satisfies the "For each credential |X| present on this |authenticator| ..." statement returns authnr attributes as well as the credentials mapped to RP ID.  Could we perhaps write something like:
> [=set/Append=]  «[ "id" → |X|'s credential ID, "type" → |X|'s type, "transports"  → |authenticator|'s transports ]» to |credentialDescriptorList|.

..?

index.bs
equalsJeffH: > I think we're guaranteed that the size isn't 0, since |credentialDescriptorList| isn't empty

d'oh!

> You can write ... to get nice Bikeshed formatting of that guarantee

ah, cool, thx.

index.bs
equalsJeffH: urm... shucks... looks to me from [CTAP spec](https://fidoalliance.org/specs/fido-v2.0-rd-20161004/fido-client-to-authenticator-protocol-v2.0-rd-20161004.html#authenticatorgetassertion), and from [WebAuthn's authenticatorGetAssertion](https://w3c.github.io/webauthn/#op-get-assertion) (abstract) operation, that we do not need to be quite as persnickety wrt filtering the allowList or gathering "credentials on this authnr" because it seems the latter routines are intended to do it. If so, we can simplify this portion of "use existing cred(s)" (aka [[DiscoverFromExternalSource]](options)), yes?

equalsJeffH: Hm. It seems to me that whomever is examining the allowList, of type sequence<ScopedCredentialDescriptor>, ought to handle everything, i.e., selecting credDescriptors  by rpId as well as transports, and if that's down within  authenticatorGetAssertion, then OK. Thoughts?

equalsJeffH:
Ok, I've pushed new commits, see  912c4ec

In spirit of @jyasskin's https://github.com/w3c/webauthn/pull/427#issuecomment-298122535, and additionally given that the AuthenticatorGetAssertion operation and CTAP ostensibly do a bunch of this allowList processing as I noted in https://github.com/w3c/webauthn/pull/427#discussion_r114213505,   I backed my over-the-top changes out to close to what is in the master branch, and then modified to just fix (i hope) the `|credentialList|` type issue and the empty-allowList-foregoing-all-processing issue (which is crux of issue #387). This may improve issue #481.

Here's some rationale wrt the empty `allowList` (aka `credentialDescriptorList` in the actual algorithm) from issue #387 (captured here to such that it is handy):

https://github.com/w3c/webauthn/issues/387#issuecomment-294056421  @equalsJeffH
> My understanding is that we need to make empty `allowList` work, and the semantics ought to be that if the `allowList`is empty, the RP is saying "please use any credential you may have associated with my RP ID", and on the client side a platform-specific procedure is used to determine whether any such credentials exist. 

https://github.com/w3c/webauthn/issues/387#issuecomment-294058725 @jyasskin 
https://github.com/w3c/webauthn/issues/387#issuecomment-294176048 @vijaybh 

please review.  In the meantime I'm looking into addressing issues #481 and #480 in here...

rlin1:
It fixes #481 in my opinion.

rlin1:
I think there is one issue left:
line 834 and line 839: credentialDescriptorList appears as parameter in the call to the Authenticator.
But: This appear inside our "in parallel for each credential C in credentialDescriptorList".

So I think we should replace credentialDescriptorList by C in lines 834 and 839.

rlin1:
One more comment: the "in parallel for each credential C" seem to implicitly assume that each Authenticator related to the different credentials C could "display" (e.g. flash LED etc.) something to the use *in parallel*.
Note: If you have two bound authenticators in one smartphone (e.g. Fingerprint based and one Iris based one) this assumption might not be correct.

equalsJeffH:
thanks @rlin1 

> It fixes #481 in my opinion.

great, that issue ought to get auto-closed upon merge. 

> I think there is one issue left:
line 834 and line 839: credentialDescriptorList appears as parameter in the call to the Authenticator.
But: This appear inside our "in parallel for each credential C in credentialDescriptorList".

Yes, tho that is the topic of issue #480 which I am intending to work on addressing after this PR is merged.

I'll "move" your 2nd https://github.com/w3c/webauthn/pull/427#issuecomment-306543869 over to issue #480

equalsJeffH:
@selfissued approved the present state, and @rlin feels it closes #481. in conversation, @jyasskin has confidence in @selfissued's review, tho is unavailable today.   Will merge this now.

Issue #422, opened 2017-04-22
authenticator taxonomy


equalsJeffH:
Given that the spec has explicit notions of platform and roaming authenticators, and we are now thinking about being able to denote other attributes of authenticators-cum-platform, such as whether they manage credential private keys, we may want to explicitly write down an authenticator taxonomy. 

Here's a swipe at it:

* first-factor Bound Authenticator
  * create(): credential private keys are stored client-side.
  * get(): When there is no user session (no cookies, a clear machine), the platform+UA+authenticator work together to display a pick list of any existing credentials registered with this RP.   During step-up authentication (when there is a user session) the server can supply credential IDs to the authnr.

* second-factor Bound Authenticator
  * create(): credential private keys are stored client-side.
  * get(): When there is no user session (no cookies, a clear machine), the RP needs to prompt the user for their account information, otherwise the user cannot use this authenticator at this time. During step-up authentication (when there is a user session) the server can supply credential IDs to the authnr.

* first-factor Roaming Authenticator
  * create(): credential private keys are stored on the discrete authenticator.
  * get(): When there is no user session (no cookies, a clear machine), the platform+UA+authenticator work together to display a pick list of any existing credentials registered with this RP. During step-up authentication (when there is a user session) the server can supply credential IDs to the authnr.

* second-factor Roaming Authenticator
  * create(): credential private keys are stored server-side (wrapped within credential IDs).
  * get(): When there is no user session (no cookies, a clear machine), the RP needs to prompt the user for their account information, otherwise the user cannot use this authenticator at this time. During step-up authentication (when there is a user session) the server can supply credential IDs to the authnr.

equalsJeffH:
Oops, I committed a terminology faux pas in my original post: "bound authenticator" corrected to be "platform authenticator", [see above](https://github.com/w3c/webauthn/issues/422#issue-223531685).

also listed terms we perhaps ought to define in the spec, [see above](https://github.com/w3c/webauthn/issues/422#issue-223531685).

equalsJeffH:
WRT the term "first-factor" -- it is one of those invented-during-a-meeting-while-in-a-rush things that was not carefully thought-through or researched prior to being employed. It turns out we perhaps ought to have simply referred to NIST SP800-63-2 and used "multi-factor authenticator".  [See S 6.1.1](http://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-63-2.pdf). 

[...]

PR #429, opened 2017-04-26
3rd approach: Add authenticator selection dictionary to create with attachment

AngeloKai:
In this PR, the attachment dictionary is part of the authenticator selection dictionary and is removed from MakeCredentialOptions. One thing to note in this PR is that the authenticatorMakeCredential operation is invoked with all the options in AuthenticatorSelectionCriteria when in reality only one (requireResidentKey) is passed in.


index.bs
```diff
@@ -551,9 +561,19 @@ When this method is invoked, the user agent MUST execute the following algorithm
 
 1. Let |issuedRequests| and |currentlyAvailableAuthenticators| be new [=ordered sets=].
 
-1. For each |authenticator| currently available on this platform, if
-    <code>|options|.{{MakeCredentialOptions/attachment}}</code> is [=present|not present=] or its value matches
-    |authenticator|'s attachment modality, [=set/append=] |authenticator| to |currentlyAvailableAuthenticators|.
+1. For each [=authenticator=] currently available on this platform, process each option in 
+    |options|.{{MakeCredentialOptions/authenticatorSelection}} by checking if the [=authenticator=] is satisfies the
```
leshi: "is satisfies" -- something weird here

index.bs
```diff
@@ -551,9 +561,19 @@ When this method is invoked, the user agent MUST execute the following algorithm
...
-1. For each |authenticator| currently available on this platform, if
-    <code>|options|.{{MakeCredentialOptions/attachment}}</code> is [=present|not present=] or its value matches
-    |authenticator|'s attachment modality, [=set/append=] |authenticator| to |currentlyAvailableAuthenticators|.
+1. For each [=authenticator=] currently available on this platform, process each option in 
+    |options|.{{MakeCredentialOptions/authenticatorSelection}} by checking if the [=authenticator=] is satisfies the 
+    requirement of the option. 
+        1. Check if |attachment| is [=present|not present=] or its value matches
+            |authenticator|'s attachment modality, [=continue=].
+        1. If |requireResidentKey| is set to |true|, check if the |authenticator| is capable of storing 
+            [=Client-Side-Resident Credential Private Key=]. If so, [=continue=]. If not, return a {{DOMException}} whose name
```
leshi: QuotaExceededError doesn't seem right.  "Quota exceeded" can mean lots of things -- like "retries exceeded", "requests exceeded".  Can we be more specific?

selfissued:
I think that this is the cleanest approach and should be chosen over PR #378 and PR #428 .  It puts all the authenticator selection criteria in one place.  It also means that we could revert PR #409 - removing this bit from the authenticator data.

index.bs
AngeloKai: I am happy with changing the DOMException error into other things. QuotaExceededError is just the closest one I can find on https://heycam.github.io/webidl/#idl-DOMException-error-names. We can certainly add a new error name in the Web IDL if necessary.

index.bs
```diff
@@ -568,13 +578,23 @@ When this method is invoked, the user agent MUST execute the following algorithm
...
 1. If |currentlyAvailableAuthenticators| [=list/is empty=], return a {{DOMException}} whose name is 
     "{{NotFoundError}}", and terminate this algorithm.
 
+1. For each [=authenticator=] currently available on this platform, process each option in 
+    |options|.{{MakeCredentialOptions/authenticatorSelection}} by checking if the [=authenticator=] satisfies the 
+    requirement of the option. 
+        1. Check if |attachment| is [=present|not present=] or its value matches
+            |authenticator|'s attachment modality, [=continue=].
+        1. If |requireResidentKey| is set to |true|, check if the |authenticator| is capable of storing 
+            [=Client-Side-Resident Credential Private Key=]. If so, [=continue=]. If not, return a {{DOMException}} whose name
```
jyasskin: If the user has two authenticators available, and only one of them supports resident keys, this step will fail the entire algorithm when it gets to the other one. Is that really what you want to happen?

index.bs
```diff
@@ -568,13 +578,23 @@ When this method is invoked, the user agent MUST execute the following algorithm
...
-
 1. If |currentlyAvailableAuthenticators| [=list/is empty=], return a {{DOMException}} whose name is 
     "{{NotFoundError}}", and terminate this algorithm.
 
+1. For each [=authenticator=] currently available on this platform, process each option in 
+    |options|.{{MakeCredentialOptions/authenticatorSelection}} by checking if the [=authenticator=] satisfies the 
+    requirement of the option. 
+        1. Check if |attachment| is [=present|not present=] or its value matches
+            |authenticator|'s attachment modality, [=continue=].
+        1. If |requireResidentKey| is set to |true|, check if the |authenticator| is capable of storing
```
jyasskin: s/storing/creating a/?

Also, I wonder if this filter should be happening on the CTAP side. Could you give me read access to the CTAP repository so I can check if the latest version has a good place to put this filter?

index.bs
```diff
@@ -568,13 +578,23 @@ When this method is invoked, the user agent MUST execute the following algorithm
...
-    <code>|options|.{{MakeCredentialOptions/attachment}}</code> is [=present|not present=] or its value matches
-    |authenticator|'s [=attachment modality=], [=set/append=] |authenticator| to |currentlyAvailableAuthenticators|. 
-
 1. If |currentlyAvailableAuthenticators| [=list/is empty=], return a {{DOMException}} whose name is 
     "{{NotFoundError}}", and terminate this algorithm.
 
+1. For each [=authenticator=] currently available on this platform, process each option in 
+    |options|.{{MakeCredentialOptions/authenticatorSelection}} by checking if the [=authenticator=] satisfies the 
+    requirement of the option. 
+        1. Check if |attachment| is [=present|not present=] or its value matches
```
jyasskin: Try to avoid "Check if *X*" in specifications. That sort of sentence tends to fail to say what to do if *X* is true or false. Just "If *X*, then *Y*. Otherwise, *Z*." is clearer.

index.bs
```diff
@@ -568,13 +578,23 @@ When this method is invoked, the user agent MUST execute the following algorithm
...
+            |authenticator|'s attachment modality, [=continue=].
```
jyasskin: I think you're using "continue" with a meaning opposite its actual definition at https://infra.spec.whatwg.org/#iteration-continue.

index.bs
```diff
@@ -568,13 +578,23 @@ When this method is invoked, the user agent MUST execute the following algorithm
...
 1. Let |issuedRequests| and |currentlyAvailableAuthenticators| be new [=ordered sets=].
 
-1. For each |authenticator| currently available on this platform, if
-    <code>|options|.{{MakeCredentialOptions/attachment}}</code> is [=present|not present=] or its value matches
-    |authenticator|'s [=attachment modality=], [=set/append=] |authenticator| to |currentlyAvailableAuthenticators|. 
-
 1. If |currentlyAvailableAuthenticators| [=list/is empty=], return a {{DOMException}} whose name is 
     "{{NotFoundError}}", and terminate this algorithm.
 
+1. For each [=authenticator=] currently available on this platform, process each option in
```
jyasskin: You can omit everything after "platform," in this sentence.

index.bs
```diff
@@ -568,13 +578,23 @@ When this method is invoked, the user agent MUST execute the following algorithm
...
+    |options|.{{MakeCredentialOptions/authenticatorSelection}} by checking if the [=authenticator=] satisfies the 
+    requirement of the option. 
+        1. Check if |attachment| is [=present|not present=] or its value matches
+            |authenticator|'s attachment modality, [=continue=].
+        1. If |requireResidentKey| is set to |true|, check if the |authenticator| is capable of storing 
+            [=Client-Side-Resident Credential Private Key=]. If so, [=continue=]. If not, return a {{DOMException}} whose name
+            is "{{QuotaExceededError}}".
+        1. If the [=authenticator=] is capable of all criterias as listed above, [=set/append=] [=authenticator=] to 
+            |currentlyAvailableAuthenticators|. 
+        1. If there is no [=authenticator=] that satisfies the criterias listed in
```
jyasskin: You don't actually intend this to be done for each *authenticator*, but that's the implication if you put it in this group of substeps.

index.bs
```diff
@@ -568,13 +578,23 @@ When this method is invoked, the user agent MUST execute the following algorithm
...
 
+1. For each [=authenticator=] currently available on this platform, process each option in 
+    |options|.{{MakeCredentialOptions/authenticatorSelection}} by checking if the [=authenticator=] satisfies the 
+    requirement of the option. 
+        1. Check if |attachment| is [=present|not present=] or its value matches
+            |authenticator|'s attachment modality, [=continue=].
+        1. If |requireResidentKey| is set to |true|, check if the |authenticator| is capable of storing 
+            [=Client-Side-Resident Credential Private Key=]. If so, [=continue=]. If not, return a {{DOMException}} whose name
+            is "{{QuotaExceededError}}".
+        1. If the [=authenticator=] is capable of all criterias as listed above, [=set/append=] [=authenticator=] to
```
jyasskin: If we got here in the substeps, "If the [=authenticator=] is capable of all criterias as listed above," is known to be true, so you don't need to say it.

index.bs
```diff
@@ -974,10 +994,10 @@ optionally evidence of [=user consent=] to a specific transaction.
         account on a single authenticator. The platform is requested to return an error if the new credential would be created
         on an authenticator that also contains one of the credentials enumerated in this parameter.
 
-    :   <dfn>attachment</dfn>
-    ::  This member contains authenticator attachment descriptions, which are used as an additional constraint on which
-        authenticators are eligible to participate in a {{CredentialsContainer/create()}} or {{CredentialsContainer/get()}}
-        operation. See [[#attachment]] for a description of the attachment values and their meanings.
+    :   <dfn>authenticatorSelection</dfn>
+    ::  This member is intended for use by [=[RPS]=] that wish to select the appropriate authenticators to participate in 
+        the {{CredentialsContainer/create()}} or {{CredentialsContainer/get()}} operation based on their unique needs.
```
jyasskin: s/based on their unique needs//

index.bs
```diff
@@ -974,10 +994,10 @@ optionally evidence of [=user consent=] to a specific transaction.
...
         on an authenticator that also contains one of the credentials enumerated in this parameter.
 
-    :   <dfn>attachment</dfn>
-    ::  This member contains authenticator attachment descriptions, which are used as an additional constraint on which
-        authenticators are eligible to participate in a {{CredentialsContainer/create()}} or {{CredentialsContainer/get()}}
-        operation. See [[#attachment]] for a description of the attachment values and their meanings.
+    :   <dfn>authenticatorSelection</dfn>
+    ::  This member is intended for use by [=[RPS]=] that wish to select the appropriate authenticators to participate in 
+        the {{CredentialsContainer/create()}} or {{CredentialsContainer/get()}} operation based on their unique needs. 
+        See [[#Authenticator Selection Criteria]] for a description of all of the criterias that the [=[RPS]=] may impose.
```
jyasskin: This didn't link in https://api.csswg.org/bikeshed/?url=https://raw.githubusercontent.com/AngeloKai/webauthn/angelo-secondAuthnrSelect/index.bs#dom-makecredentialoptions-authenticatorselection. I think you can probably remove the entire sentence because it just links to the `AuthenticatorSelectionCriteria` type that's already linked by the definition of `authenticatorSelection`.

index.bs
```diff
@@ -1012,6 +1032,33 @@ The {{ScopedCredentialEntity}} dictionary describes a user account or a [=relyin
         a user's avatar or a [=relying party=]'s logo.
 </div>
 
+
+### Authenticator Selection Criteria ### {#authenticatorSelection}
+
+The {{AuthenticatorSelectionCriteria}} dictionary describes a set of criterias for which the [RPS] can use to find the appropriate
```
jyasskin: s/for//

index.bs
```diff
@@ -1012,6 +1032,33 @@ The {{ScopedCredentialEntity}} dictionary describes a user account or a [=relyin
...
+<xmp class="idl">
+    dictionary AuthenticatorSelectionCriteria {
+        Attachment    attachment;
+        boolean       requireResidentKey;
+    };
+</xmp>
+
+<div dfn-type="dict-member" dfn-for="AuthenticatorSelectionCriteria">
+    :   <dfn>attachment</dfn>
+    ::  This member contains authenticator attachment descriptions, which are used as an additional constraint on which
```
jyasskin: I'd shorten this to, "If present, filters to authenticators attached with the specified {{Attachment}}."

index.bs
```diff
@@ -1012,6 +1032,33 @@ The {{ScopedCredentialEntity}} dictionary describes a user account or a [=relyin
...
+
+### Authenticator Selection Criteria ### {#authenticatorSelection}
+
+The {{AuthenticatorSelectionCriteria}} dictionary describes a set of criterias for which the [RPS] can use to find the appropriate
+authenticator 
+
+<xmp class="idl">
+    dictionary AuthenticatorSelectionCriteria {
+        Attachment    attachment;
+        boolean       requireResidentKey;
```
jyasskin: Probably add `=false` so algorithms don't have to deal with the possibility of this field being not-present.

index.bs
```diff
@@ -1012,6 +1032,33 @@ The {{ScopedCredentialEntity}} dictionary describes a user account or a [=relyin
...
+
+<div dfn-type="dict-member" dfn-for="AuthenticatorSelectionCriteria">
+    :   <dfn>attachment</dfn>
+    ::  This member contains authenticator attachment descriptions, which are used as an additional constraint on which
+        authenticators are eligible to participate in a {{CredentialsContainer/create()}} or {{CredentialsContainer/get()}}
+        operation. See [[#attachment]] for a description of the attachment values and their meanings.
+
+    :   <dfn>requireResidentKey</dfn>
+    ::  This member describes the [=[RPS]=]' requirements regarding availability of the [=Client-side-resident Credential 
+        Private Key=]. If the parameter is set to <a>true</a>, the authenticator SHOULD create a
```
jyasskin: Don't use "SHOULD" here. If it's not a guarantee, it seems like it'll be useless to RPs.

index.bs
```diff
@@ -1407,6 +1455,11 @@ When this operation is invoked, the authenticator must perform the following pro
     not, return an error code equivalent to NotSupportedError and terminate the operation.
 - Check if a credential matching any of the supplied {{ScopedCredential}} identifiers is present on this authenticator. If so,
     return an error code equivalent to NotAllowedError and terminate the operation.
+- Check if the |requireResidentKey| flag is set to |true|, check if the authenticator can store
```
jyasskin: Again, avoid the "check if" construction.

index.bs
```diff
@@ -1407,6 +1455,11 @@ When this operation is invoked, the authenticator must perform the following pro
     not, return an error code equivalent to NotSupportedError and terminate the operation.
 - Check if a credential matching any of the supplied {{ScopedCredential}} identifiers is present on this authenticator. If so,
     return an error code equivalent to NotAllowedError and terminate the operation.
+- Check if the |requireResidentKey| flag is set to |true|, check if the authenticator can store 
+    [=Client-side-resident Credential Private Key=]. If not, return an error code equivalent to QuotaExceededError and  
+    terminate the operation.   
+- Check if the [=authenticator=] is capable of storing  return a {{DOMException}} whose name is equivalent to
```
jyasskin: You're missing something between "storing" and "return".

index.bs
jyasskin: In this terminology, and in the ["Client-side-resident Credential Private Key"](https://api.csswg.org/bikeshed/?url=https://raw.githubusercontent.com/AngeloKai/webauthn/angelo-secondAuthnrSelect/index.bs#client-side-resident-credential-private-key) term above, I'd still rather focus on the fact that this allows creating credentials using just the RP Id without passing a Credential Id. The fact that the key will be stored client-side is just an implementation detail of that RP-visible feature.

index.bs
equalsJeffH: > Also, I wonder if this filter should be happening on the CTAP side.

yes, something to wonder about and is it not a similar question to this one (and the one following it):  https://github.com/w3c/webauthn/pull/427#discussion_r114213505  ?

WRT CTAP, at this point in time, Section 4 "authenticator API" of the published CTAP spec is stable (tho there may be outstanding proposals to alter it): https://fidoalliance.org/specs/fido-v2.0-rd-20161004/fido-client-to-authenticator-protocol-v2.0-rd-20161004.html#authenticator-api

index.bs
```diff
@@ -325,6 +325,16 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
     that this does not constitute [=user verification=] because [=TUP=], by definition, is not capable of [=biometric
     recognition=], nor does it involve the presentation of a shared secret such as a password or PIN.
 
+: <dfn>Client-side-resident Credential Private Key</dfn>
+:: A [=Client-side-resident Credential Private Key=] is stored either on the client platform, or in some cases on the authenticator
```
rlin1: This is confusing.  By definition the key always resides in the authenticator.  The authenticator might be bound to the platform or it might be a roaming authenticator.

index.bs
rlin1: Why do we need two different fields for attachment and requireResidentKey?
If the Authenticator is roaming the key is roaming (by definition of Authenticator).
If the Authenticator is bound to the platform, the key is bound to the platform as well (called "resident" here).

index.bs
rlin1: If we want to differentiate between Authenticators offloading storage of the wrapped key material, we need to be more specific in the description here.
For example:
By definition, the private key is always exclusively controlled by the Authenticator. In the case of a [=Client-side-resident Credential Private Key=], the Authenticator might offload storage of wrapped key material to the client platform, but the client platform is not expected to offload the key storage to remote entities (e.g. RP Server).

index.bs
rlin1: Would requireResidentKey mean that the Authenticator is NOT allowed to offload storage of wrapped key material to (a) the client platform or to (b) remote entities?

index.bs
jyasskin: My understanding is that the key may *not* always reside in the authenticator. Specifically, the authenticator may encrypt the key to itself, and store the encrypted key in the Credential ID that the RP stores, expecting to receive it back from the RP in the get() call.

For roaming authenticators, storing it in the browser or OS isn't an option, since it needs to be available for another login to the same RP when the same authenticator is attached to a different browser+OS. Having the RP store it with the Credential ID is still viable in that case.

@rlin1, since you weren't already thinking of that model, that makes you really valuable in wording the spec, so that it can explain this option to other newcomers. Can you think of a clear way to specify the requirement that the authenticator *not* use the scheme I just described? That's the intent of setting `requireResidentKey` to true.

rlin1: This is where it gets tricky.  The Authenticator always controls/protects/owns the key, but it might offload the persistent storage of the wrapped key to some external component (if this is what is meant by resident).
We need a description that makes this distinction clear. 
My proposal:
By definition, the private key is always exclusively controlled by the Authenticator. In the case of a [=Client-side-resident Credential Private Key=], the Authenticator might offload storage of wrapped key material to the client platform, but the client platform is not expected to offload the key storage to remote entities (e.g. RP Server).

jyasskin: The client platform *is* expected to offload the key storage to remote entities. The storage is in the Credential ID, which the RP Server stores.

index.bs
equalsJeffH: > If the Authenticator is roaming the key is roaming (by definition of Authenticator).

We need to refine the webauthn definition for [=authenticator=] to clarify this. also, there is not a dfn for a "roaming credential private key" (as yet).

index.bs
jyasskin: Whoops, yes, you've made the right distinction about client-side-resident keys.

equalsJeffH:
see also #422 -- having an explicitly defined authenticator taxonomy will be helpful for the other definitions we need to make in order to specify this functionality.

index.bs
```diff
@@ -584,16 +584,14 @@ When this method is invoked, the user agent MUST execute the following algorithm
...
-        1. If |requireResidentKey| is set to |true|, check if the |authenticator| is capable of storing 
-            [=Client-Side-Resident Credential Private Key=]. If so, [=continue=]. If not, return a {{DOMException}} whose name
-            is "{{QuotaExceededError}}".
-        1. If the [=authenticator=] is capable of all criterias as listed above, [=set/append=] [=authenticator=] to 
-            |currentlyAvailableAuthenticators|. 
-        1. If there is no [=authenticator=] that satisfies the criterias listed in 
-            |options|.{{MakeCredentialOptions/authenticatorSelection}}, return a {{DOMException}} whose name is 
-            "{{ConstraintError}}", and terminate this algoritm.
+        1. If |attachment| is [=present|not present=] or its value matches
+            |authenticator|'s attachment modality, then [=iteration/continue=]. Otheriwse, [=iteration/break=].
```
jyasskin: Both continue and break will skip the remaining substeps in this sequence. They act like the same keywords in C, Java, Python, etc.

rlin1:
We also have the authnSel extension.  I propose adding the AAGUID-list to this AuthenticatorSelectionCriteria.  I am fine with adding it in a different PR.

gmandyam:
@rlin1 

After merging this, we should propose additional criteria.  I was hoping to add criteria derived from a subset of existing FIDO Metadata, e.g. keyProtection = ... .

equalsJeffH:
decision on webauthn call today is to merge this upon resolving review comments eg @jyasskin , @rlin's  https://www.w3.org/2017/05/03-webauthn-minutes.html

equalsJeffH:
fixes #367

equalsJeffH:
fixes #367

equalsJeffH:
fixes #245 ?

index.bs
AngeloKai: Is it ok if I add another PR to explain "roaming credential private key" instead of this one? We can address the editorial issues in WD-06.

AngeloKai:
@leshi @equalsJeffH @jyasskin can you please review?

index.bs
```diff
@@ -569,13 +582,21 @@ When this method is invoked, the user agent MUST execute the following algorithm
...
+1. If |currentlyAvailableAuthenticators| [=list/is empty=], return a {{DOMException}} whose name is 
     "{{NotFoundError}}", and terminate this algorithm.
 
+1. If |options|.{{MakeCredentialOptions/authenticatorSelection}} is [=present|present=], iterate through 
+        |currentlyAvailableAuthenticators| and doing the following for each [=authenticator=]:
+    1. If |attachment| is [=present|not present=] or its value matches |authenticator|'s attachment modality, continue
+        to the next step. Otherwise, [=iteration/continue=] to the next [=authenticator=].
+    1. If |requireResidentKey| is set to |true| and the |authenticator| is capable of storing 
+        [=Client-Side-Resident Credential Private Key=], continue to the next step. Otherwise, return a {{DOMException}} 
+        whose name is "{{QuotaExceededError}}" and [=iteration/continue=] to the next [=authenticator=].
```
jyasskin: "return a {{DOMException}} ... and [=iteration/continue=]" is self-inconsistent because "return" and "continue" direct control flow to different places. What step is supposed to do something with the QuotaExceededError? How does it flow through to affect the eventual return value of this whole algorithm, which has to return just one of a PublicKeyCredential or an exception?

index.bs
```diff
@@ -569,13 +582,21 @@ When this method is invoked, the user agent MUST execute the following algorithm
...
-1. If |currentlyAvailableAuthenticators| [=list/is empty=], return a {{DOMException}} whose name is
+1. If |currentlyAvailableAuthenticators| [=list/is empty=], return a {{DOMException}} whose name is 
     "{{NotFoundError}}", and terminate this algorithm.
 
+1. If |options|.{{MakeCredentialOptions/authenticatorSelection}} is [=present|present=], iterate through 
+        |currentlyAvailableAuthenticators| and doing the following for each [=authenticator=]:
+    1. If |attachment| is [=present|not present=] or its value matches |authenticator|'s attachment modality, continue
+        to the next step. Otherwise, [=iteration/continue=] to the next [=authenticator=].
+    1. If |requireResidentKey| is set to |true| and the |authenticator| is capable of storing 
+        [=Client-Side-Resident Credential Private Key=], continue to the next step. Otherwise, return a {{DOMException}}
```
jyasskin: Like above, continuing to the next step is the default, so you should invert the condition and only mention the other case. Unlike above, the condition you have isn't actually right, in that it'll treat a `false` |requireResidentKey| as a failure. How about:
> If |requireResidentKey| is set to |true| and the |authenticator| is <ins>not</ins> capable of storing <ins>a</ins> [=Client-Side-Resident Credential Private Key=], *failure action*.

index.bs
```diff
@@ -569,13 +582,21 @@ When this method is invoked, the user agent MUST execute the following algorithm
...
+1. If |options|.{{MakeCredentialOptions/authenticatorSelection}} is [=present|present=], iterate through 
+        |currentlyAvailableAuthenticators| and doing the following for each [=authenticator=]:
+    1. If |attachment| is [=present|not present=] or its value matches |authenticator|'s attachment modality, continue
+        to the next step. Otherwise, [=iteration/continue=] to the next [=authenticator=].
+    1. If |requireResidentKey| is set to |true| and the |authenticator| is capable of storing 
+        [=Client-Side-Resident Credential Private Key=], continue to the next step. Otherwise, return a {{DOMException}} 
+        whose name is "{{QuotaExceededError}}" and [=iteration/continue=] to the next [=authenticator=]. 
+    1. [=set/Append=] [=authenticator=] to |currentlyAvailableAuthenticators|. 
+
+1. If there is no [=authenticator=] that satisfies the criterias listed in |options|.{{MakeCredentialOptions/authenticatorSelection}}
```
jyasskin: This "If" is vague, and I think it's subsumed by the loop immediately preceding it. Does "If |currentlyAvailableAuthenticators| is [=set/empty=] ..." produce the right outcome?

index.bs
```diff
@@ -930,8 +951,8 @@ optionally evidence of [=user consent=] to a specific transaction.
         required sequence<PublicKeyCredentialParameters> parameters;
 
         unsigned long                        timeout;
-        sequence<PublicKeyCredentialDescriptor> excludeList;
-        Attachment                           attachment;
+        sequence<ScopedCredentialDescriptor> excludeList;
```
jyasskin: Looks like a merge error vs the change from Scoped to PublicKey.

index.bs
```diff
@@ -1018,6 +1038,32 @@ associated.
...
 
+
+### Authenticator Selection Criteria ### {#authenticatorSelection}
+
+The {{AuthenticatorSelectionCriteria}} dictionary describes a set of criterias for which the [RPS] can use to find the appropriate
+authenticator 
+
+<xmp class="idl">
+    dictionary AuthenticatorSelectionCriteria {
+        Attachment    attachment = false;
```
jyasskin: `false` isn't a valid [`Attachment`](https://api.csswg.org/bikeshed/?url=https://raw.githubusercontent.com/AngeloKai/webauthn/angelo-secondAuthnrSelect/index.bs#enumdef-attachment) value, so it can't be the default here. I think you want to leave this field without a default, since the algorithm treats not-present differently from either present value.

index.bs
```diff
@@ -1018,6 +1038,32 @@ associated.
...
+        boolean       requireResidentKey = false;
+    };
+</xmp>
+
+<div dfn-type="dict-member" dfn-for="AuthenticatorSelectionCriteria">
+    :   <dfn>attachment</dfn>
+    :   If this memeber is [=present|present=], eligible authenticators are filter to only authenticators attached with the
```
jyasskin: spelling: memeber and s/filter/filtered/

index.bs
```diff
@@ -1018,6 +1038,32 @@ associated.
...
+    dictionary AuthenticatorSelectionCriteria {
+        Attachment    attachment = false;
+        boolean       requireResidentKey = false;
+    };
+</xmp>
+
+<div dfn-type="dict-member" dfn-for="AuthenticatorSelectionCriteria">
+    :   <dfn>attachment</dfn>
+    :   If this memeber is [=present|present=], eligible authenticators are filter to only authenticators attached with the 
+        specified [[#attachment]] are eligible
```
jyasskin: s/are eligible// I think, or s/eligible authenticators are filter to//

index.bs
jyasskin: Remember to check the rendering: https://api.csswg.org/bikeshed/?url=https://raw.githubusercontent.com/AngeloKai/webauthn/angelo-secondAuthnrSelect/index.bs#dom-authenticatorselectioncriteria-attachment reveals you need two ":"s for this line.

index.bs
```diff
@@ -1401,14 +1447,18 @@ input parameters:
 - A list of {{PublicKeyCredential}} objects provided by the [RP] with the intention that, if any of these are known to the
     authenticator, it should not create a new credential.
 - Extension data created by the client based on the extensions requested by the [RP].
+- The |requireResidentKey| parameter of the |options|.{{MakeCredentialOptions/authenticatorSelection}} dictionary.  
 
 When this operation is invoked, the authenticator must perform the following procedure:
 - Check if all the supplied parameters are syntactically well-formed and of the correct length. If not, return an error code
-    equivalent to UnknownError and terminate the operation.
+    equivalent to UnknownError and terminate the operation.
```
jyasskin: Try not to introduce trailing whitespace.

index.bs
```diff
@@ -1401,14 +1447,18 @@ input parameters:
...
 
 When this operation is invoked, the authenticator must perform the following procedure:
 - Check if all the supplied parameters are syntactically well-formed and of the correct length. If not, return an error code
-    equivalent to UnknownError and terminate the operation.
+    equivalent to UnknownError and terminate the operation.    
 -  Check if at least one of the specified combinations of {{PublicKeyCredentialType}} and cryptographic parameters is supported.
     If not, return an error code equivalent to NotSupportedError and terminate the operation.
 - Check if a credential matching any of the supplied {{PublicKeyCredential}} identifiers is present on this authenticator. If
     so, return an error code equivalent to NotAllowedError and terminate the operation.
+- If the |requireResidentKey| flag is set to |true| and the authenticator can store [=Client-side-resident Credential
```
jyasskin: s/can store/can store a/

index.bs
```diff
@@ -1401,14 +1447,18 @@ input parameters:
...
 When this operation is invoked, the authenticator must perform the following procedure:
 - Check if all the supplied parameters are syntactically well-formed and of the correct length. If not, return an error code
-    equivalent to UnknownError and terminate the operation.
+    equivalent to UnknownError and terminate the operation.    
 -  Check if at least one of the specified combinations of {{PublicKeyCredentialType}} and cryptographic parameters is supported.
     If not, return an error code equivalent to NotSupportedError and terminate the operation.
 - Check if a credential matching any of the supplied {{PublicKeyCredential}} identifiers is present on this authenticator. If
     so, return an error code equivalent to NotAllowedError and terminate the operation.
+- If the |requireResidentKey| flag is set to |true| and the authenticator can store [=Client-side-resident Credential 
+    Private Key=], continue to the next step. Otherwise, return an error code equivalent to QuotaExceededError and
```
jyasskin: This repeats the mistake in the *requireResidentKey* step above: a false value will cause a `QuotaExceededError`.

index.bs
jyasskin: This one's still bad grammar.

index.bs
jyasskin: FWIW, I don't hate `QuotaExceededError` here. It's not exactly what you expect to be reported, but it's actually plausible that the UA+authenticator would limit the number of keys a single domain can store even if the authenticator still has space, which would make "quota exceeded" exactly right.

I haven't checked this with other web platform folks, so it's possible someone will want to add a new WebIDL error type instead, but it seems fine to go with `QuotaExceededError` for now.

index.bs
```diff
@@ -569,13 +582,21 @@ When this method is invoked, the user agent MUST execute the following algorithm
 
 1. Let |issuedRequests| and |currentlyAvailableAuthenticators| be new [=ordered sets=].
 
-1. For each |authenticator| currently available on this platform, if
-    <code>|options|.{{MakeCredentialOptions/attachment}}</code> is [=present|not present=] or its value matches
-    |authenticator|'s [=attachment modality=], [=set/append=] |authenticator| to |currentlyAvailableAuthenticators|.
-
-1. If |currentlyAvailableAuthenticators| [=list/is empty=], return a {{DOMException}} whose name is
+1. If |currentlyAvailableAuthenticators| [=list/is empty=], return a {{DOMException}} whose name is
```
AngeloKai: |currentlyAvailableAuthenticators| won't be empty at this point. At this point, the client goes down to search and find all the authenticators at the platform. Then the client checks to see if any of them meets the criteria set by the RPs. 
This way, if an RP sees NotFoundError, the RP knows there's no authenticator whatsoever, If the RP sees ConstraintError, the RP knows it hasn't set the right constraint yet or its constraint is limiting. If the RP sees QuotaExceededError, the RP knows the authenticator doesn't have storage in it.

index.bs
equalsJeffH: yep. how 'bout:
> [RPS] may use the {{AuthenticatorSelectionCriteria}} dictionary to specify their requirements regarding authenticator attributes. 

..?

index.bs
equalsJeffH: yes, please delete "based on their unique needs".

index.bs
jyasskin: How about splitting the previous step to:

> 1. Let |issuedRequests| be a new empty [=ordered set=].
> 2. Let |currentlyAvailableAuthenticators| be a new [=ordered set=] consisting of all [=authenticators=] available on this platform.

index.bs
AngeloKai: I was thinking we can send one DOMException but not breaking the algorithm. In another word, this will let developer know one authenticator on the platform doesn't have the storage. If there's another authenticator that does have the storage, then the authentication will still work. Is that not the right web behavior? I can also remove the QuotaExceededError.

AngeloKai: If we end up doing silent authentication in the next iteration, it could be an attack by the RP to fill up the memory allocated and stop user from logging in.

jyasskin: This algorithm ultimately controls how [`navigator.credentials.create()`](https://w3c.github.io/webappsec-credential-management/#credentialscontainer) resolves the `Promise` it returned, and that `Promise` either rejects with an exception or fulfils with a `PublicKeyCredential`. There's no mechanism specified to return a stream of exceptions in addition to a `PublicKeyCredential`.

I think the `QuotaExceededError` solves one problem that @leshi is worried about, so it's worth keeping.

How about, add 
> 17. Let |selectionReasons| be a new list.

above this loop; change this step to:
> 2. If |requireResidentKey| is set to |true| and the |authenticator| is not capable of storing a [=Client-Side-Resident Credential Private Key=], append a new "{{QuotaExceededError}}" {{DOMException}} to |selectionReasons| and [=continue=].

; then change the step after this loop to:

> 19. If |selectedAuthenticators| is [=empty=], then if |selectionReasons| is:
    <dl class="switch">
    : [=empty=]
    :: Return a DOMException whose name is "ConstraintError", and terminate this algorithm.
    : [=not empty=]
    :: Return an arbitrary item of |selectionReasons|, and terminate this algorithm.
    </dl>

(I haven't double-checked that markdown; please make sure it renders sensibly.)

index.bs
jyasskin: I've noticed that step 18 iterates over |currentlyAvailableAuthenticators| and also inserts into it. It should insert into a different list that you initialize here, perhaps
> 18. Let |selectedAuthenticators| be a new [=ordered set=].

AngeloKai: Updated the algorithm with selectedAuthenticators.

index.bs
AngeloKai: Updated with selectionReasons.

AngeloKai:
@equalsJeffH @jyasskin can you please review the new changes?

index.bs
equalsJeffH: OK. tho submit issue first please.

index.bs
```diff
@@ -1041,20 +1051,19 @@ associated.
...
-The {{AuthenticatorSelectionCriteria}} dictionary describes a set of criterias for which the [RPS] can use to find the appropriate
-authenticator 
+[=[RPS]=] may use the {{AuthenticatorSelectionCriteria}} dictionary to specify their requirements regarding authenticator attributes.
 
 <xmp class="idl">
     dictionary AuthenticatorSelectionCriteria {
-        Attachment    attachment = false;
-        boolean       requireResidentKey = false;
+        Attachment    attachment;
+        boolean       requireResidentKey;
```
equalsJeffH: I.e. I believe @jyasskin is suggesting:
```
       Attachment    attachment;
       boolean       requireResidentKey = false; 
```

AngeloKai: Just updated with requireResidentKey default set to false.

jyasskin:
FWIW, @leshi and I represent different parts of Chromium: I can say that the spec looks in line with the rest of the web platform, but @leshi owns whether the semantics are what we want. I could hold off on approving that the spec's sane until I've checked with him for Chrome's overall opinion, but I think it's more helpful to just treat our opinions as separate.

AngeloKai:
I was not aware of that. In future PRs, I will treat your opinions as separate then.

PR #432, opened 2017-05-01
public key cred - fixes #406

### LINKED ISSUE #406: Rename ScopedCredential to PublicKeyCredential
"Scoped" doesn't identify what's unique about WebAuthn credentials, and on the calls, folks seem to be happy with "public key" instead.

For the key in [CredentialRequestOptions](https://w3c.github.io/webappsec-credential-management/#dictdef-credentialrequestoptions), I suggest "publicKey" following the convention for properties in https://w3ctag.github.io/design-principles/#casing-rules.

equalsJeffH:
"scoped cred" -> "public key cred" - fixes #406

whatchay'all think?

/cc @mikewest

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1,5 +1,5 @@
 <pre class='metadata'>
-Title: Web Authentication: An API for accessing Scoped Credentials
+Title: Web Authentication: An API for accessing Public Key Credentials
 Status: ED
 Prepare for TR: true
 TR: https://www.w3.org/TR/webauthn/
@@ -28,13 +28,13 @@ Text Macro: RPS Relying Parties
 Text Macro: INFORMATIVE <em>This section is not normative.</em>
 Text Macro: WAC WebAuthn Client
 Ignored Vars: op, alg, type, algorithm
-Abstract: This specification defines an API enabling the creation and use of strong, attested, cryptographic scoped credentials
- by web applications, for the purpose of strongly authenticating users. Conceptually, one or more credentials, each scoped to a
- given Relying Party, are created and stored on an authenticator by the user agent in conjunction with the web application. The
- user agent mediates access to scoped credentials in order to preserve user privacy. Authenticators are responsible for ensuring
- that no operation is performed without user consent. Authenticators provide cryptographic proof of their properties to relying
- parties via attestation. This specification also describes the functional model for WebAuthn conformant authenticators,
- including their signature and attestation functionality.
+Abstract: This specification defines an API enabling the creation and use of strong, attested, scoped, public key-based
+ credentials by web applications, for the purpose of strongly authenticating users. Conceptually, one or more credentials, each
+ scoped to a given Relying Party, are created and stored on an authenticator by the user agent in conjunction with the web
+ application. The user agent mediates access to [=public key credentials=] in order to preserve user privacy. [=Authenticators=]
+ are responsible for ensuring that no operation is performed without [=user consent=]. [=Authenticators=] provide cryptographic
+ proof of their properties to relying parties via [=attestation=]. This specification also describes the functional model for
+ WebAuthn conformant authenticators, including their signature and attestation functionality.
 Boilerplate: omit conformance, omit feedback-header
 Markup Shorthands: css off, markdown on
 </pre>
@@ -91,25 +91,25 @@ spec:html; type:dfn; for:environment settings object; text:global object
 
 [INFORMATIVE]
 
-This specification defines an API enabling the creation and use of strong, attested, cryptographic <em>[=scoped
-credentials=]</em> by web applications, for the purpose of strongly authenticating users. A [=scoped credential=] is
+This specification defines an API enabling the creation and use of strong, attested, scoped, public key-based
+credentials by web applications, for the purpose of strongly authenticating users. A [=public key credential=] is
 created and stored by an <em>[=authenticator=]</em> at the behest of a <em>[=[RP]=]</em>, subject to <em>[=user 
-consent=]</em>. Subsequently, the scoped credential can only be accessed by origins belonging to that [=[RP]=].
+consent=]</em>. Subsequently, the [=public key credential=] can only be accessed by [=origins=] belonging to that [=[RP]=].
 This scoping is enforced jointly by <em>[=conforming User Agents=]</em> and <em>[=authenticators=]</em>.
 Additionally, privacy across [=[RPS]=] is maintained; [RPS] are not able to detect any properties, or even
 the existence, of credentials scoped to other [RPS].
 
 [RPS] employ the [=Web Authentication API=] during two distinct, but related, [=ceremonies=] involving a user. The first
-is [=Registration=], where a [=scoped credential=] is created on an [=authenticator=], and associated by a [=[RP]=]
+is [=Registration=], where a [=public key credential=] is created on an [=authenticator=], and associated by a [=[RP]=]
 with the present user's account (the account may already exist or may be created at this time). The second is
 [=Authentication=], where the [=[RP]=] is presented with an <em>[=Authentication Assertion=]</em> proving the presence
-and consent of the user who registered the [=scoped credential=]. Functionally, the [=Web Authentication API=] comprises
-a {{ScopedCredential}} which extends the Credential Management API [[CREDENTIAL-MANAGEMENT-1]], and infrastructure which
+and consent of the user who registered the [=public key credential=]. Functionally, the [=Web Authentication API=] comprises
+a {{PublicKeyCredential}} which extends the Credential Management API [[CREDENTIAL-MANAGEMENT-1]], and infrastructure which
 allows those credentials to be used with {{CredentialsContainer/create()|navigator.credentials.create()}} and
 {{CredentialsContainer/get()|navigator.credentials.get()}}. The former is used during [=Registration=], and the
 latter during [=Authentication=].
 
-Broadly, compliant [=authenticators=] protect [=scoped credentials=], and interact with user agents to implement the
+Broadly, compliant [=authenticators=] protect [=public key credentials=], and interact with user agents to implement the
 [=Web Authentication API=]. Some authenticators may run on the same computing device (e.g., smart phone, tablet, desktop PC) as
 the user agent is running on. For instance, such an authenticator might consist of a Trusted Execution Environment (TEE) applet,
 a Trusted Platform Module (TPM), or a Secure Element (SE) integrated into the computing device in conjunction with some means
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -364,79 +364,80 @@ deletion are considered to be the responsibility of such a user interface and ar
...
 
     :   <dfn>clientExtensionResults</dfn>
     ::  This attribute contains a [=map=] containing [=extension identifier=] → [=client extension output=] entries
         produced by the extension's [=client extension processing=].
 
     :   <dfn>\[[type]]</dfn>
-    ::  The {{ScopedCredential}} [=interface object=]'s {{Credential/[[type]]}} [=internal slot=]'s value is the string
-        "`scoped`".
+    ::  The {{PublicKeyCredential}} [=interface object=]'s {{Credential/[[type]]}} [=internal slot=]'s value is the string
+        "`publickey`".
```
jyasskin: I believe https://w3ctag.github.io/design-principles/#casing-rules recommends "public-key" here.

index.bs
```diff
@@ -364,79 +364,80 @@ deletion are considered to be the responsibility of such a user interface and ar
...
         "{{Credential/[[discovery]]/remote}}".
 
     :   <dfn>\[[identifier]]</dfn>
-    ::  This [=internal slot=] contains an identifier for the credential, chosen by the platform with help from the authenticator.
-        This identifier is used to look up credentials for use, and is therefore expected to be globally unique with high
-        probability across all credentials of the same type, across all authenticators. This API does not constrain the format
-        or length of this identifier, except that it must be sufficient for the platform to uniquely select a key. For example,
-        an authenticator without on-board storage may create identifiers containing a [=credential private key=] wrapped with a
-        symmetric key that is burned into the authenticator.
+    ::  This [=internal slot=] contains an identifier for the credential, chosen by the platform with help from the
```
jyasskin: I think the file is mostly wrapped to 130 characters. To avoid churn, maybe don't rewrap to less than that when you're not otherwise changing the paragraph?

index.bs
```diff
@@ -1149,38 +1157,38 @@ following Web IDL.
...
 
 
-### Credential Type enumeration (enum <dfn enum>ScopedCredentialType</dfn>) ### {#credentialType}
+### Credential Type enumeration (enum <dfn enum>PublicKeyCredentialType</dfn>) ### {#credentialType}
 
 <pre class="idl">
-    enum ScopedCredentialType {
-        "scoped"
+    enum PublicKeyCredentialType {
+        "publickey"
```
jyasskin: Probably "public-key" per https://w3ctag.github.io/design-principles/#casing-rules.

index.bs
```diff
@@ -1149,38 +1157,38 @@ following Web IDL.
...
 </pre>
 
-<div dfn-type="enum-value" dfn-for="ScopedCredentialType">
+<div dfn-type="enum-value" dfn-for="PublicKeyCredentialType">
     This enumeration defines the valid credential types. It is an extension point; values may be added to it in the future, as
     more credential types are defined. The values of this enumeration are used for versioning the Authentication Assertion and
     attestation structures according to the type of the authenticator.
 
-    Currently one credential type is defined, namely "<dfn>scoped</dfn>".
+    Currently one credential type is defined, namely "<dfn>publickey</dfn>".
```
jyasskin: And here, so you don't have to re-find this use.

index.bs
```diff
@@ -2443,7 +2466,7 @@ The entry key is the [=extension identifier=] and the value is the [=client exte
 
 <pre class="example" highlight="js">
     var assertionPromise = navigator.credentials.get({
-        scoped: {
+        publickey: {
```
jyasskin: `publicKey`

index.bs
```diff
@@ -2499,17 +2522,17 @@ from the [=authenticator extension input=], and possibly also other inputs, for
...
-of the extension. The [RP] sets this value in its request for an assertion:
+The [=extension identifier=] is chosen as `webauthnExample_geo`. The [=client extension input=] is the constant value `true`,
+since the extension does not require the [=[RP]=] to pass any particular information to the client, other than that it requests
+the use of the extension. The [RP] sets this value in its request for an assertion:
 
 <pre highlight="js">
     var assertionPromise =
         navigator.credentials.get({
-            scoped: {
+            publickey: {
```
jyasskin: `publicKey`

index.bs
```diff
@@ -3038,9 +3064,9 @@ This is the first-time flow, in which a new credential is created and registered
 The sample code for generating and registering a new key follows:
 
 <pre class="example" highlight="js">
-    if (!ScopedCredential) { /* Platform not capable. Handle error. */ }
+    if (!PublicKeyCredential) { /* Platform not capable. Handle error. */ }
 
-    var scoped = {
+    var publickey = {
```
jyasskin: `publicKey`

index.bs
```diff
@@ -3060,11 +3086,11 @@ The sample code for generating and registering a new key follows:
       // prefers an ES256 credential.
       parameters: [
         {
-          type: "scoped",
+          type: "publickey",
```
jyasskin: `public-key`

index.bs
```diff
@@ -3076,7 +3102,7 @@ The sample code for generating and registering a new key follows:
     };
 
     // Note: The following call will cause the authenticator to display UI.
-    navigator.credentials.create({ scoped })
+    navigator.credentials.create({ publickey })
```
jyasskin: `publicKey`

index.bs
```diff
@@ -3126,15 +3152,15 @@ If the [RP] script does not have any hints available (e.g., from locally stored
...
 
 <pre class="example" highlight="js">
-    if (!ScopedCredential) { /* Platform not capable. Handle error. */ }
+    if (!PublicKeyCredential) { /* Platform not capable. Handle error. */ }
 
     var options = {
                     challenge: new TextEncoder().encode("climb a mountain"),
                     timeout: 60000,  // 1 minute
-                    allowList: [{ type: "scoped" }]
+                    allowList: [{ type: "publickey" }]
```
jyasskin: `public-key`

index.bs
```diff
@@ -3126,15 +3152,15 @@ If the [RP] script does not have any hints available (e.g., from locally stored
...
                   };
 
-    navigator.credentials.get({ "scoped": options })
+    navigator.credentials.get({ "publickey": options })
```
jyasskin: `publicKey`

index.bs
```diff
@@ -3147,15 +3173,15 @@ performing such an authentication might look like the following. Note that this
 extension for transaction authorization.
 
 <pre class="example" highlight="js">
-    if (!ScopedCredential) { /* Platform not capable. Handle error. */ }
+    if (!PublicKeyCredential) { /* Platform not capable. Handle error. */ }
 
     var encoder = new TextEncoder();
     var acceptableCredential1 = {
-        type: "scoped",
+        type: "publickey",
```
jyasskin: Here too.

index.bs
equalsJeffH: yeah, it did seem to me that casing-rules can be interpreted to apply here, but it sorta seems an "internal slot" is only implicitly an enum. If up to me I'd use "publickey" rather than "public-key", but it's not a big deal. this was sort of an experiment to see if anyone would catch it :)

index.bs
equalsJeffH: @vijaybh wrapped it to 128 chars when initially converting to BS format: https://lists.w3.org/Archives/Public/public-webauthn/2016Apr/0116.html

index.bs
equalsJeffH: ja, shur, u betcha, see above.

index.bs
jyasskin: :) The only things that just concatenate lower-case words are [events and HTML elements and attributes](https://w3ctag.github.io/design-principles/#casing-rules), and this definitely isn't one of those.

index.bs
equalsJeffH: personally, I'd much prefer 80-char lines, but can live with it as-is.

Also, I tried hard-wrap+indent of header lines "#..." and BS complained so did not wrap them.

selfissued:
I agree that we shouldn't re-wrap lines unless we're actually changing them to prevent unnecessary merge conflicts.

selfissued:
BTW, I'm not suggesting unwrapping any lines that were already re-wrapped by this set of edits.  What's done is done.

equalsJeffH:
@selfissued said
> we shouldn't re-wrap lines unless we're actually changing them to prevent unnecessary merge conflicts.

fyi/fwiw: AFAIK simply re-hard-wrapping lines does not itself cause merge conflicts -- e.g. it did not when I merged master into this branch, nor when I've done prior line-wrapping cleanup in prior PRs.

equalsJeffH:
Ok folks -- given the three positive reviews so far, unless anyone objects by noon PT, I have it good authority that @selfissued is gonna merge this :)

AngeloKai:
@selfissued @equalsJeffH I still need to review them. I may submit a formal objection later today.

nadalin:
I really would like to hear from all the implementers on this one before it is just merged @jcjones and @AngeloKai

AngeloKai:
@equalsJeffH My main concern with the change is that the new name, "PublicKeyCredentials", isn't easier to understand for web developers than the old name "ScopedCredential". The former implies the credential has a public/private key pair, which is a common concept found in every crypto API. The latter implies the credential is scoped to something, which most web dev immediately think of scoping to origin. I took an informal vote around my office full of web developers and none of them knows what either means. I am afraid we are tinkering without real feedbacks.

leshi:
My 2 cents is that the spec is SO complex that there is no way anyone is going to understand it enough to implement anything useful by skimming it.  The developers at the very least will have to understand how to verify assertions.  IMHO, that will be easier if they have the "asymmetric key" big picture rather than the "scoped to something" big picture.

I support this change.

jyasskin:
"Scoped" will never survive TAG review because it doesn't distinguish this kind of credential from any other kind of web credential. PublicKey might not survive either, due to some of the other public-key-ish systems mentioned in https://github.com/w3c/webauthn/issues/406#issuecomment-294258206, but it at least has a chance. We should get feedback on a name that has a chance, rather than getting feedback on one that doesn't.

I'm not wedded to "PublicKey" if @AngeloKai has another suggestion, but we do need a concrete suggestion.

AngeloKai:
@jyasskin @leshi How about DeviceBoundCredentials or AuthenticatorBoundCredentials? One of the biggest value add of the API is that credentials are now bound with the device, ensuring at least one barrier of security. Also, "scoped" already survived the TAG review: https://github.com/w3ctag/design-reviews/issues/97

jyasskin:
@AngeloKai In your informal vote, what did your coworkers think a device- or authenticator-bound credential would imply? What did they think a public-key credential would imply?

IIUC, with self-attested keys, there's not actually a guarantee that the key is bound to the client? Whereas there is always a public/private key pair involved?

(None of the ScopedCredential interface survived the first TAG review: they said to merge it into Credential Manager.)

jyasskin:
"DeviceBoundCredential" would be acceptable to me. I prefer "PublicKeyCredential" because it gives developers more hint about how they need to use the credential, and I don't think @AngeloKai has given an argument against it (["implies the credential has a public/private key pair"](https://github.com/w3c/webauthn/pull/432#issuecomment-298748100) is correct), but "DeviceBound" does distinguish this from other kinds of credentials on the platform.

rlin1:
I like publicKeyCredential (as it is more specific than ScopedCredential).  Good to be merged.

equalsJeffH:
decision on the webauthn call today is to merge this for WD-05:
https://www.w3.org/2017/05/03-webauthn-minutes.html

PR #440, opened 2017-05-04
Create .pr-preview.json -- set up 'pr-preview'

equalsJeffH:
see: [[pr-preview] announcement](https://lists.w3.org/Archives/Public/spec-prod/2017JanMar/0047.html)  as well as [this one](https://lists.w3.org/Archives/Public/spec-prod/2017JanMar/0048.html)

### DIFF
--- a/.pr-preview.json
+++ b/.pr-preview.json
@@ -1,0 +1,7 @@
+{
+    "src_file": "index.bs",
+    "type": "bikeshed",
+    "params": {
+        "force": 1
+    }
+}


jcjones:
It looks OK to me. Have we gotten [the integration installed yet](https://github.com/integration/pr-preview)?

equalsJeffH:
I _think_ that the one-button integration u point to is only for top-level accounts, eg W3C, jcjones.  Tobie's instructions hint that just having the .pr-preview.json is sufficient, but perhaps not?

jcjones:
Let's find out!

PR #442, opened 2017-05-04
move AAGUID for selecting authnr into new AuthenticatorSelectionCriteria

### LINKED ISSUE #441: Make AuthenticatorSelectionCriteria more complete.
We have an extension for selecting the authenticator based on a list of AAGUIDs, then we recently added the more generic AuthenticatorSelectionCriteria.

We should move the extension into this new AuthenticatorSelectionCriteria in order to concentrate all authenticator selection related criteria (authenticator policy) into one single entity.

rlin1:
Close #441

***
[Preview](https://s3.amazonaws.com/pr-preview/w3c/webauthn/authnr-sel-addtl-fields.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/webauthn/a2a4210...72f82e9.html)


rlin1:
Close #446 
Close #447

index.bs
```diff
@@ -1100,6 +1124,24 @@ a minimum of friction, e.g., the user will not have to dig around in their pocke
 example of the latter, when the user is accessing the [=[RP]=] from a given client for the first time, they may be required to
 use a [=roaming authenticator=] which was originally registered with the [=[RP]=] using a different client.
 
+### Key Protection enumeration (enum <dfn enum>KeyProtection</dfn>) ### {#key-protection}
+
+<pre class="idl">
+    enum KeyProtection {
+       "SW",   // key protected by Rich-OS software
```
jyasskin: A comment is not enough to specify what these fields do.

jyasskin:
More generally, if you actually want these fields added in a timely manner, you should send them in separate PRs so that a field that gets consensus isn't held up behind more contentious fields. Showing a unified framework for these fields in an issue could help build support for them, but even then it's wise to add them separately.

equalsJeffH:
@jyasskin noted on call just now something about this interacting with pr #460

equalsJeffH:
@jyasskin  notes on the call just now: this PR has 3 components: AAGUID, key protection, matcher protection. all should be evaluated independently (separate PRs). matcher protection _could_ be moved into PR #460 but not necessarily.

selfissued:
I agree with @jyasskin that each independent piece of functionality should be placed into its own PR so that it can be evaluated independently.  This was agreed to on the 24-May-17 WG call.

rlin1:
Will close (unmerged) this PR and create individual PRs for adding aaguidList, keyProtection and matcherProtection to Authenticator SelectionCriteria.

PR #449, opened 2017-05-04
Remove QuotaExceededError from spec

AngeloKai:
Remove QuotaExceededError per request by @leshi

***
[Preview](https://s3.amazonaws.com/pr-preview/AngeloKai/webauthn/changeError.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/webauthn/0f8aa1d...AngeloKai:d326645.html)

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1,6 +1,6 @@
 <pre class='metadata'>
 Title: Web Authentication: An API for accessing Public Key Credentials
-Status: WD
+Status: ED
 Prepare for TR: true
 TR: https://www.w3.org/TR/webauthn/
 ED: https://w3c.github.io/webauthn/
--- a/index.bs
+++ b/index.bs
@@ -1,6 +1,6 @@
 <pre class='metadata'>
 Title: Web Authentication: An API for accessing Public Key Credentials
-Status: ED
+Status: WD
 Prepare for TR: true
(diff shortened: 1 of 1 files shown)


jyasskin:
@leshi, I thought `QuotaExceededError` solved (or improved) one problem you pointed out with requiring resident keys: it lets the RP tell the user that their key storage is full. It's true that [it's not as precise as one might want](https://github.com/w3c/webauthn/pull/429#discussion_r114042267), but we can solve that by proposing a more precise error name into [WebIDL](https://github.com/heycam/webidl/pulls).

https://w3c.github.io/mediacapture-main/getusermedia.html#overconstrainederror-object has another, more complex, precedent.

Or do you actually think that we should leave this vague for now (as this PR does), and come up with a more solid solution with more time to think?

AngeloKai:
@jyasskin This seems like a small issue. I prefer not letting this drag out our publication schedule. I'd preferring this vague for now and making it more solid later.

leshi:
@jyasskin,

I think the `SelectionResponse` variable was internal anyway.  So RP wouldn't learn why things failed.

AngeloKai:
Given that neither @leshi nor @jyasskin has a strong opinion on this, I will go ahead and merge it then.

PR #460, opened 2017-05-10
Adds requireUserVerification option in AuthenticatorSelectionCriteria

### LINKED ISSUE #459: requireUserVerification option in AuthenticatorSelectionCriteria
We seem to have accidentally removed the `requireUserVerification` option in the `AuthenticatorSelectionCriteria`.  We may want to add it back in.

leshi:
Closes #459

***
[Preview](https://s3.amazonaws.com/pr-preview/w3c/webauthn/acz.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/webauthn/52b4422...25889a4.html)

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1153,8 +1153,8 @@ attributes.
 <xmp class="idl">
     dictionary AuthenticatorSelectionCriteria {
         AuthenticatorAttachment      attachment;
-        boolean       requireResidentKey = false;
-        boolean       requireUserVerification = false;
+        boolean                      requireResidentKey = false;
+        boolean                      requireUserVerification = false;
     };
 </xmp>
 
--- a/index.bs
+++ b/index.bs
@@ -1486,9 +1486,13 @@ The [=authenticator data=] structure is a byte array of 37 bytes or more, as fol
         <td>1</td>
         <td>
             Flags (bit 0 is the least significant bit):
-            - Bit 0: [=Test of User Presence=] (`TUP`) result.
+            - Bit 0: [=test of user presence/User Present=] (`UP`) result.
+                `1` means the user is [=test of user presence/present=].
+                `0` means the user is not [=test of user presence/present=].
             - Bit 1: Reserved for future use (`RFU1`).
-            - Bit 2: [=User Verification=] (`UV`) result.
+            - Bit 2: [=user verification/User Verified=] (`UV`) result.
+                `1` means [=user verification=] has been performed.
+                `0` means [=user verification=] has not been performed.
             - Bits 3-5: Reserved for future use (`RFU2`).
             - Bit 6: [=Attestation data=] included (`AT`). Indicates whether the authenticator added [=attestation data=].
             - Bit 7: Extension data included (`ED`). Indicates if the [=authenticator data=] has extensions.
--- a/index.bs
+++ b/index.bs
@@ -1170,7 +1170,7 @@ attributes.
     :   <dfn>requireUserVerification</dfn>
     ::  This member describes the [=[RPS]=]' requirements regarding the authenticator being capable of performing user verification.
         If the parameter is set to true, the authenticator MUST perform user verification when performing
-	the {{CredentialsContainer/create()}} operation and future [[#getAssertion]] operations when it is
+        the {{CredentialsContainer/create()}} operation and future [[#getAssertion]] operations when it is
         requested to verify the credential.
 </div>
 
@@ -1487,7 +1487,7 @@ The [=authenticator data=] structure is a byte array of 37 bytes or more, as fol
             Flags (bit 0 is the least significant bit):
             - Bit 0: [=Test of User Presence=] (`TUP`) result.
             - Bit 1: Reserved for future use (`RFU1`).
-	    - Bit 2: [=User Verification=] (`UV`) result.
+            - Bit 2: [=User Verification=] (`UV`) result.
             - Bits 3-5: Reserved for future use (`RFU2`).
             - Bit 6: [=Attestation data=] included (`AT`). Indicates whether the authenticator added [=attestation data=].
             - Bit 7: Extension data included (`ED`). Indicates if the [=authenticator data=] has extensions.
--- a/index.bs
+++ b/index.bs
@@ -1152,9 +1152,9 @@ attributes.
 
 <xmp class="idl">
     dictionary AuthenticatorSelectionCriteria {
-        AuthenticatorAttachment      AA;         // authenticatorAttachment
-        boolean                      RK = false; // requireResidentKey
-        boolean                      UV = false; // requireUserVerification
+        AuthenticatorAttachment      aa;         // authenticatorAttachment
+        boolean                      rk = false; // requireResidentKey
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -595,6 +595,8 @@ When this method is invoked, the user agent MUST execute the following algorithm
         to |authenticator|'s attachment modality, [=iteration/continue=].
     1. If {{AuthenticatorSelectionCriteria/requireResidentKey}} is set to |true| and the |authenticator| 
         is not capable of storing a [=Client-Side-Resident Credential Private Key=], [=iteration/continue=].
+    1. If {{AuthenticatorSelectionCriteria/requireUserVerification}} is set to |true| and the
```
jyasskin: This needs to also get passed on to the authenticator, which needs to be specified to use it to ensure that future `get()` calls are authorized by the same user as authorized the `create()` call.

index.bs
```diff
@@ -1064,6 +1067,12 @@ associated.
     ::  This member describes the [=[RPS]=]' requirements regarding availability of the [=Client-side-resident Credential 
         Private Key=]. If the parameter is set to true, the authenticator MUST create a 
         [=Client-side-resident Credential Private Key=] when creating a [=public key credential=].
+
+    :   <dfn>requireUserVerification</dfn>
+    ::  If this parameter is set to <i>true</i>, only authenticators capable of user verification would participate in the 
+        {{CredentialsContainer/create()}} operation, thereby ensuring the authenticator will perform the user verification
```
jyasskin: It's not enough to just perform [user verification](https://s3.amazonaws.com/pr-preview/w3c/webauthn/acz.html#user-verification). Even if the authenticator can "distinguish individual users", it needs to record which user authorized create() and only sign credentials when the same user authorizes a get() call.

I'd like to use a more precise name than `requireUserVerification` for this bit. `constrainGetToSameUser` captures the idea, but there may be a shorter expression of it.

equalsJeffH:
hey folks -- it looks to me like `requireUserVerification` and associated text was inadvertently removed by a cockpit error when resolving merge conflicts, in this commit:

https://github.com/w3c/webauthn/commit/5e860cd2c4c028d49601847022a84c465cae7815#diff-ec9cfa5f3f35ec1f84feb2e59686c34d

If you view the diff rendered by the above, then search for `requireUserVerif` and/or `<<<<<<< HEAD` you will find what seems to be a chunk of spec that ought not to have been removed when correcting merge conflicts.  

I note that the removed text was apparently added with the merge of PR #348 on Mon 24-Apr-2017. We discussed PR #348 on the [2017/04/19-webauthn call](https://www.w3.org/2017/04/19-webauthn-minutes.html)  and I do not see mentioned explicitly in those minutes that we agreed to merge #348 as-is (there are unresolved raised issues with it noted in the PR), but it was merged in any case. 

Thus I am thinking that it is OK to attempt to re-insert `requireUserVerification` and associated text in this new PR rather than try to go back and fix the merge-conflict-resolution-error, because we may or may not "need" this `requireUserVerification` now that we have a more refined notion of "authenticator selection" to work with.  

I'll try to catch up on @jyasskin's comments tomorrow and see about contributing to the discussion (it's getting to be past dinner time for me tonight here in CET)...

selfissued:
I support restoring this important functionality.

AngeloKai:
I noticed the missing of this functionality when I did the merge but skipped it because I feel more discussion is needed around this. On the top of my head, the first question I have to ask is how sure are we that we will only stop at one factor biometric authentication? Last time I heard, multi-factor auth is slowing becoming a trend. I am wondering if we should change the Boolean to a dictionary.

index.bs
selfissued: @jyasskin is right.  The client doesn't know whether the authenticator can do user verification or not.  This needs to be passed on to the authenticator - not acted upon by the client.

leshi: The client will call `getInfo()` to get authenticator capabilities.

index.bs
leshi: I'm happy to change the name to what you propose.  Any objection from others?

leshi:
Alexei to clarify authenticator behavior

index.bs
equalsJeffH: I disagree with @jyasskin here. The behavior he describes wrt user verification is how things work today and everyone understands that. however, after grovelling through UAF, U2F, and WebAuthn, AFAICT we do not explicitly state it. I submited #493.

gmandyam:
"If this parameter is set to true, only authenticators capable of user verification would participate ..."

Current text implies that if parameter is set to false that silent authenticators would be permissible.  If that is not the intent, then this text should be modified to indicate desired client behavior when requireUserVerification is set to false.

index.bs
```diff
@@ -1477,7 +1486,9 @@ The [=authenticator data=] structure is a byte array of 37 bytes or more, as fol
         <td>
             Flags (bit 0 is the least significant bit):
             - Bit 0: [=Test of User Presence=] (`TUP`) result.
-            - Bits 1-5: Reserved for future use (`RFU`).
+            - Bit 1: Reserved for future use (`RFU1`).
```
AngeloKai: Can you explain a little why you aren't using Bit 1? I hold the position that the web spec shouldn't define the underlying protocol too specifically and this is an instance where the web spec is being more descriptive. Therefore if the reasoning here comes from the CTAP world, that's fine. But if there's discussion that leads to this justification, please do let us know. At least JC and I are far removed from the CTAP world.

selfissued: In private communication, Alexei Czeskis @leshi let me know that Google is already privately using bit 1 for some purpose, and asked me to use bit 2 for UV.  If Google is willing to move their private use bit elsewhere, I'd be glad to use bit 1 for UV instead, as it would seem more natural.

nadalin: Seems we should be keeping to what Angelo has indicated as making this bit 2 a little bit off

index.bs
```diff
@@ -1556,6 +1567,7 @@ input parameters:
     these are known to the authenticator, it should not create a new credential.
 - Extension data created by the client based on the extensions requested by the [=[RP]=], if any.
 - The |requireResidentKey| parameter of the |options|.{{MakePublicKeyCredentialOptions/authenticatorSelection}} dictionary.
+- The |requireUserVerification| parameter of the |options|.{{MakePublicKeyCredentialOptions/authenticatorSelection}} dictionary.
```
equalsJeffH: please fix the ordering of the params as listed here and as listed in #createCredential alg, per comment here: https://github.com/w3c/webauthn/pull/502#pullrequestreview-46415425  -- i.e., i suggest keeping extension data as the last item in the params list, please see comment above.

index.bs
leshi: Hi there, we use the ever-so-mysterious Bit 1 to indicate that the authenticator believes that the assertion was sent over a wireless transport (rather than a wired transport).

selfissued: How does this bit affect the client's and/or RP's behaviors? What "if" statements do they write based on this condition? Is there valuable interoperable behavior we should be specifying based on this result?

PR #435, opened 2017-05-03
s/daaKey/ecdaaKeyId/ in fig 3

equalsJeffH:
s/daaKey/ecdaaKeyId/ in "attestation statement" in fig 3

PR #463, opened 2017-05-15
redraw fig 3, polish attestation & assertion signature definitions and prose

equalsJeffH:
this PR supersedes PR #435 (which I will close).  I redrew fig 3 in order to hopefully clarify it and the attestation object's relationship with its various component data structures.  Did various editorial/terminological cleaning up.

***
[Preview](https://s3.amazonaws.com/pr-preview/w3c/webauthn/jeffh-redraw-fig3.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/webauthn/177f1ec...780ff7f.html)


index.bs
```diff
@@ -228,18 +230,18 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
 :: See [=Authentication Assertion=].
 
 : <dfn>Attestation</dfn>
-:: Generally, a statement that serves to bear witness, confirm, or authenticate.
-    In the WebAuthn context, attestation is employed to attest to the provenance of an authenticator and the data it emits;
-    including, for example: credential IDs, credential key pairs, signature counters, etc. <dfn>Attestation information</dfn> is
-    conveyed in <dfn>attestation objects</dfn>.
-    See also [=attestation statement format=], and [=attestation type=].
+:: Generally, <em>attestation</em> is a statement serving to bear witness, confirm, or authenticate.
+    In the WebAuthn context, [=attestation=] is employed to <em>attest</em> to the <em>provenance</em> of an [=authenticator=]
```
jyasskin: 1. I'd like to be clearer about *who* is attesting to this sort of thing. Attestations aren't just handed down by God, they can be more or less trustworthy, and RPs ought to pay some attention to that.
2. Isn't it less the provenance of the data an authenticator emits (and stores, for private keys), than the security guarantees provided for that data?

index.bs
```diff
@@ -310,14 +312,15 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
...
-:: Generically, a credential is data one entity presents to another in order to authenticate the former's identity [[RFC4949]].
-    A WebAuthn <em>[=public key credential=]</em> is a <code>{ identifier, type }</code> pair identifying authentication
-    information established by the authenticator and the [RP], together, at [=registration=] time. The authentication
-    information consists of an asymmetric key pair, where the public key portion is returned to the [RP], which stores it in
-    conjunction with the present user's account. The authenticator maps the private key to the [RP]'s [=RP ID=] and stores
-    it. Subsequently, only that [RP], as identified by its [=RP ID=], is able to employ the [=public key credential=] in
-    [=authentication=] ceremonies, via the {{CredentialsContainer/get()}} method. The [=[RP]=] uses its copy of the stored public
-    key to verify the resultant [=Authentication Assertion=].
+:: Generically, a <em>credential</em> is data one entity presents to another in order to <em>authenticate</em> the former to the
+    latter [[RFC4949]]. A WebAuthn <em>[=public key credential=]</em> is a <code>{ identifier, type }</code> pair identifying
```
jyasskin: This is less true now that WebAuthn subclasses Credential Manager's `Credential` type, which includes the RFC4949 "signed value" in addition to the RFC4949 "credential".

This and the previous sentence are also inconsistent. The {identifier, type} pair is *not* the "data one entity presents to another in order to authenticate the former to the latter"; that also has to include the signed challenge. (I favor moving away from RFC4949's definitions for reasons including that it has no term for the data actually used to authenticate.)

Issue #416, opened 2017-04-20
rpID seems to have changed meaning a bit


leshi:
Previously, the rpID was an optional field that you only overwrite if you want to change the security semantics of how the credential was bound.

However, the rpID has now moved into the `ScopedCredentialEntity` structure:
```
dictionary ScopedCredentialEntity {
    DOMString id;
    DOMString name;
    USVString icon;
};
```

But the general meaning of this structure is "display information for the authenticator account chooser".  It feels like the move resulted in a semantic change from "security knob" to "metadata".

mikewest:
Technically, the field is still optional, so I think it has the same semantics as it used to. That said, I can see how the way it's presented to developers might have shifted in a way that makes it less likely to be used correctly. *shrug* I'm open to recommendations. What would you like to see change?

equalsJeffH:
I have several rpId/origin issues to address, will add this one to my list...(they have percolated up to the top of my stack I'm hoping...)

equalsJeffH:
things were further refined since @leshi submitted this issue, and the RP ID is now conveyed in via `MakeCredentialOptions.rp.id`, and all of `MakeCredentialOptions.rp` is RP-specific, plus, as a part of  PR #464 we have clarified the optionality of setting the RP ID (the default is the caller's effective domain and origin).

equalsJeffH:
things were further refined since @leshi submitted this issue, and the RP ID is now conveyed in via `MakeCredentialOptions.rp.id`, and all of `MakeCredentialOptions.rp` is RP-specific, plus, as a part of  PR #464 we have clarified the optionality of setting the RP ID (the default is the caller's effective domain and origin).  So I assert that we can close this upon merge of PR #464.

equalsJeffH:
this is fixed by PR 464, closing.

Issue #260, opened 2016-11-04
_rpId_ generation allows more relaxation of same-origin restrictions than document.domain does


bzbarsky:
Providing an rpId of "foo.com" would lead to a single rpId being allowed to be shared by "http://foo.com", "https://foo.com", and "https://foo.com:8080".  The first may not be a problem because this API is secure context only, but is it intended that "https://foo.com" and "https://foo.com:8080" be able to share an rpId?  If not, this needs to be fixed in the spec.  If this is intended, it may be worth a note calling it out, because this is a quite surprising deviation from how same-origin policy normally works.

selfissued:
@equalsJeffH plans to fix this.

Issue #259, opened 2016-11-04
Is _rpId_ supposed to look like an origin serialization, or like a hostname?


bzbarsky:
https://w3c.github.io/webauthn/#dom-webauthentication-makecredential step 3 has two branches.  the "rpId is not specified" branch will set _rpId_ to a string like "https://foo.bar.com".  The "rpId is specified" branch will set it to a string like "bar.com".

Is this difference purposeful?

jyasskin:
@leshi, are there constraints here based on the old FIDO implementation?

Without constraints, I'd suggest that the origin relaxation steps replace the [host](https://html.spec.whatwg.org/multipage/browsers.html#concept-origin-host) in the original origin, but leave the scheme and port alone.

selfissued:
@equalsJeffH plans to fix this.

equalsJeffH:
fixed by PR #464, closing.

PR #464, opened 2017-05-17
reconciling origin and RP ID handling

equalsJeffH:
I think I have the RP ID & origin issues mostly figured out (see issues #255, #259, #260, #416).  We have been inconsistently treating RP ID as an [origin](https://html.spec.whatwg.org/#concept-origin) or (implicitly) as a hostname (i.e., a [valid-domain-string](https://url.spec.whatwg.org/#valid-domain-string)).

After poking throughout the spec, I believe we want the RP ID to consistently be the latter (i.e., a [valid-domain-string](https://url.spec.whatwg.org/#valid-domain-string) that [is-a-registrable-domain-suffix-of-or-is-equal-to](https://html.spec.whatwg.org/#is-a-registrable-domain-suffix-of-or-is-equal-to) the RP's origin's [effective domain](https://html.spec.whatwg.org/#concept-origin-effective-domain)), since we otherwise are consistently conveying the full RP origin (in [collectedClientData](https://w3c.github.io/webauthn/#dictdef-collectedclientdata)) in any case. 

This PR at this point is for review-only -- I've only modified the #createCredential algorithm (beginning on new line 527) and am holding off on #getAssertion until have gotten some review. Though, I did go thru the entire spec to normalize descriptions of origin and RP ID.

***
[Preview](https://s3.amazonaws.com/pr-preview/w3c/webauthn/jeffh-origin-rpid.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/webauthn/a2a4210...0637185.html)


index.bs
```diff
@@ -1060,12 +1072,12 @@ associated.
 
 <div dfn-type="dict-member" dfn-for="AuthenticatorSelectionCriteria">
     :   <dfn>attachment</dfn>
-    ::  If this memeber is [=present|present=], eligible authenticators are filtered to only authenticators attached with the 
+    ::  If this memeber is [=present|present=], eligible authenticators are filtered to only authenticators attached with the
```
jcjones: spelling: s/memeber/member/ ...unless this was actually calling for bear memes, in which case we should make this a MUST imo.

index.bs
```diff
@@ -1447,16 +1459,17 @@ The following operations can be invoked by the client in an authenticator sessio
 This operation must be invoked in an authenticator session which has no other operations in progress. It takes the following
 input parameters:
 
-- The caller's RP ID, as determined by the user agent and the client.
+- The caller's [=RP ID=], as optionally decared by the [=[RP]=] and vetted by the user agent and the client, or defaulted to the
```
jcjones: nit- I almost wonder at the language here, since the definition of RP ID includes how it's vetted... but I think that's more style than substance, so ... proceed!

index.bs
```diff
@@ -569,7 +579,8 @@ When this method is invoked, the user agent MUST execute the following algorithm
     : {{CollectedClientData/challenge}}
     :: The [=base64url encoding=] of |options|.{{MakeCredentialOptions/challenge}}
     : {{origin}}
-    :: The [=unicode serialization of an origin|unicode serialization=] of |rpId|
+    :: |effectiveDomain|
+    <!-- might need to have both {{origin}} and {{rpId}} in collectedClientData... ? -->
```
jcjones: hrrrrmmm. So if we only include `effectiveDomain` in the client data, then the RP doesn't get looped-back what the `rpId` calculated was. But it still can tell what the origin of the requesting page was, and of course the RP's software is _probably_ the one setting the `rpId` in the first place.

I believe this works from the RP's perspective, and clearly it's not necessary for the authenticator. The question is, I guess: is there any security impact to the `rpId` not being in the signed hash?

Not being signed means the RP can't rely on the `rpId` being accurate to its intentions. If it were forged (through a MITM), it would allow a forger to potentially cause a signature with a different private key, since the `rpId` is often mixed into the key wrapping functions. But the MITM could also replace the included/excluded keys, which accomplishes the same thing without regard to the `rpId`. (To what end, I don't know)

So I think then it's OK to not sign the `rpId`, but I suggest others think through this too.

index.bs
```diff
@@ -521,19 +527,23 @@ When this method is invoked, the user agent MUST execute the following algorithm
     [=relevant settings object=]. If |callerOrigin| is an [=opaque origin=], return a {{DOMException}} whose name is
     "{{NotAllowedError}}", and terminate this algorithm.
 
-1. If the {{PublicKeyCredentialEntity/id}} member of |options|.{{MakeCredentialOptions/rp}} is [=present|not present=], then set
-    |rpId| to |callerOrigin|.
+1. Let |effectiveDomain| be the |callerOrigin|'s [=effective domain=]. If |effectiveDomain| is null, then return a
```
equalsJeffH: ok. Tho note that we /are/ going to have normative deltas from WD-05 given the latter was a rush job with known normative issues outstanding....

index.bs
equalsJeffH: ok, a couple items to note in reply here. first, setting `{{origin}}` to |effectiveDomain| is a mistake. mea culpa, sorry, dunno what I was thinking.  it should be set to the serialization of |callerOrigin|, yes?

>  is there any security impact to the rpId not being in the signed hash [of collectedClientData]

ans: there is none, because hash(RP ID) is the first 32 bytes of `authenticator data` -- i.e., we're signing it in any case.  :)

index.bs
equalsJeffH: LOL, cute. I'll correct the misspelling...

index.bs
equalsJeffH: well, am trying to remind the reader explicitly about its magic properties...

index.bs
jcjones: image

index.bs
```diff
@@ -1400,19 +1412,19 @@ The [=authenticator data=] structure is a byte array of 37 bytes or more, as fol
...
+requested credential exactly matches the [=RP ID=] supplied by the client.
 
 The `TUP` flag SHALL be set if and only if the authenticator detected a user through an authenticator specific gesture. The `RFU` bits
 SHALL be set to zero.
 
 For attestation signatures, the authenticator MUST set the AT flag and include the [=attestation data=]. For authentication
 signatures, the AT flag MUST NOT be set and the [=attestation data=] MUST NOT be included.
 
-If the authenticator does not include any extension data, it MUST set the `ED` flag in the first byte to zero, and to one if
+If the authenticator does not include any extension data, it MUST set the `ED` flag to zero; it is set to one if
```
jyasskin: I prefer the original wording here (possibly with s/and/or/): the modified wording uses MUST for one branch, but just a factual claim for the second, and I'd rather they be consistent.

index.bs
```diff
@@ -1441,16 +1453,17 @@ The following operations can be invoked by the client in an authenticator sessio
 This operation must be invoked in an authenticator session which has no other operations in progress. It takes the following
 input parameters:
 
-- The caller's RP ID, as determined by the user agent and the client.
+- The caller's [=RP ID=], as optionally decared by the [=[RP]=] and vetted by the user agent and the client, or defaulted to the
```
jyasskin: I think I'd prefer the original wording with "determined" linked to the step that determines it (you can put `id`s on `<li>` tags). The authenticator operations don't need to care much how the RP ID was determined, so we don't need to clutter their interfaces by re-describing it.

index.bs
```diff
@@ -1034,8 +1046,8 @@ associated.
 </xmp>
 <div dfn-type="dict-member" dfn-for="PublicKeyCredentialEntity">
     :   <dfn>id</dfn>
-    ::  A unique identifier for the entity. This will be the [=ASCII serialization of an origin=] for a [=[RP]=],
-        and an arbitrary string specified by the [=[RP]=] for user accounts.
+    ::  A unique identifier for the entity. For a [=relying party=] entity, this will be the [=RP ID=]. For a user account
```
jyasskin: s/will be/sets/?

index.bs
```diff
@@ -1034,8 +1046,8 @@ associated.
...
+        entity, this will be an arbitrary string specified by the [=relying party=].
```
jyasskin: Try to say how this string is used, although probably in a different PR since it's not related to cleaning up RP IDs.

index.bs
```diff
@@ -1447,16 +1459,17 @@ The following operations can be invoked by the client in an authenticator sessio
...
 - The [=[RP]=]'s {{PublicKeyCredentialEntity}}.
-- The user account's {{PublicKeyCredentialEntity}}.
+- The user account's {{PublicKeyCredentialUserEntity}}.
 - The {{PublicKeyCredentialType}} and cryptographic parameters requested by the [=[RP]=], with the cryptographic algorithms
     normalized as per the procedure in [[WebCryptoAPI#algorithm-normalization-normalize-an-algorithm]].
-- A list of {{PublicKeyCredential}} objects provided by the [=[RP]=] with the intention that, if any of these are known to the
-    authenticator, it should not create a new credential.
-- Extension data created by the client based on the extensions requested by the [=[RP]=].
-- The |requireResidentKey| parameter of the |options|.{{MakeCredentialOptions/authenticatorSelection}} dictionary.  
+- An optional list of {{PublicKeyCredential}} objects provided by the [=[RP]=] with the intention that, if any of these are
```
jyasskin: Since [CTAP](https://github.com/fido-alliance/fido-2-specs/pull/226/files#diff-ad05d352dbf182e60e46549766441e10R243) refers to these parameters by name, WebAuthn probably should give them the same names.

index.bs
```diff
@@ -291,23 +306,27 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
...
 
 : <dfn>Relying Party Identifier</dfn>
 : <dfn>RP ID</dfn>
-:: An identifier for the [=[RP]=] on whose behalf a given registration or authentication ceremony is being performed. Public Key
-    credentials can only be used for authentication by the same entity (as identified by RP ID) that created and registered
-    them. By default, the RP ID for a WebAuthn operation is set to the [=environment settings object/origin=] specified by the
-    [=relevant settings object=] of the {{CredentialsContainer}} object. This default can be overridden by the caller subject
-    to certain restrictions, as specified in [[#createCredential]] and [[#getAssertion]].
+:: A [=domain=] identifier for the [=[RP]=] on whose behalf a given [=registration=] or
+    [=authentication|authentication ceremony=] is being performed. A [=RP ID=] MUST be a [=valid domain string=]. A [=public key
```
equalsJeffH: done.

equalsJeffH:
@jyasskin wrote:
> I agree with @bzbarsky's concern in #260 that using just the domain name without a port could lead to problems. 

That is intentional in order to match the behavior of pervasively deployed ambient credentials (e.g., cookies, [RFC6265]). We have addressed this topic with the TAG, see:
* https://github.com/w3ctag/design-reviews/issues/97#issuecomment-171564611 (@vijaybh)
* https://github.com/w3ctag/design-reviews/issues/97#issuecomment-175766580 (@balfanz)

> It's probably fine to save the 8 bytes from omitting https:// since this is secure-contexts only ... 

It is not about saving bytes -- it is about scoping the PKCreds to a domain name, as cookies are.

> and it's intentional to implicitly share credentials across suborigins?

ah, thx. will have to investigate https://w3c.github.io/webappsec-suborigins/ -- what is the status of that spec? I note it is not as yet in the /TR namespace.  I also note long threads on public-webappsec@ to review...

> If we don't use an origin serialization uniformly, there should be a note explaining which parts of the origin we're intentionally omitting and why.

I added such in response to #260 in the commits I pushed today.

index.bs
```diff
@@ -291,23 +306,27 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
...
+    credential=] can only be used for [=authentication=] with the same entity (as identified by [=RP ID=]) it was registered
+    with. By default, the [=RP ID=] for a WebAuthn operation is set to the {{CredentialsContainer}} object's [=relevant settings
+    object=]'s [=environment settings object/origin=]'s [=effective domain=], i.e., the caller's [=environment settings
```
equalsJeffH: ok, done.

index.bs
```diff
@@ -581,23 +592,23 @@ When this method is invoked, the user agent MUST execute the following algorithm
 
 1. Let |clientDataHash| be the [=hash of the serialized client data=] represented by |clientDataJSON|.
 
-1. Let |currentlyAvailableAuthenticators| be a new [=ordered set=] consisting of all [=authenticators=] 
+1. Let |currentlyAvailableAuthenticators| be a new [=ordered set=] consisting of all [=authenticators=]
```
equalsJeffH: well, various people complain about spurious whitespace lingering at end-of-lines, or when parags are reformatted (we agreed to a 128-char line length way back when). 

I note that viewing the github-generated diff using `?w=1` at end of URI, eg <https://github.com/w3c/webauthn/pull/464/files?w=1> helps filter-out whitespace-only changes.   

when using git diff locally, there's `git diff -w`

index.bs
equalsJeffH: ok, done.

index.bs
equalsJeffH: submit an issue?

index.bs
equalsJeffH: Ok, done. tho the "..in the first byte.." portion of the original wording is incorrect and i've left it out.

index.bs
equalsJeffH: Ok. done.

index.bs
equalsJeffH: Well, in the specific case of this authenticatorMakeCredential parameter, the `MakeCredentialOptions.excludeList` is different from the `|excludeList|` ephemeral variable in the `[[Create]](options)` method alg (aka #[createCredential](https://w3c.github.io/webauthn/#createCredential)). The former is filtered into the latter in "present #[createCredential](https://w3c.github.io/webauthn/#createCredential) step 23".

Also, the [webauthn "authenticator model"](https://w3c.github.io/webauthn/#authenticator-model) is hand-wavy "non-normative", as is CTAP itself. [CTAP section 4](https://fidoalliance.org/specs/fido-v2.0-rd-20161004/fido-client-to-authenticator-protocol-v2.0-rd-20161004.html#authenticator-api) says
> Note that this API level is conceptual and does not represent actual APIs. The actual APIs will be provided by each implementing platform.

see also issues #176 and #410.

So, perhaps we could leave the above list of authenticatorMakeCredential input params largely as-is, where it defines the values passed rather than using CTAP-specific parameter names?

Once we have a publicly-released updated CTAP spec, we could update this parameter list to say "for example, in CTAP this is foo parameter" (just a thought).

equalsJeffH:
additional commits pushed, please review, thx :)

index.bs
```diff
@@ -293,23 +309,38 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
...
+    key credential=] can only be used for [=authentication=] with the same entity (as identified by [=RP ID=]) it was registered
+    with. By default, the [=RP ID=] for a WebAuthn operation is set to the caller's [=environment settings object/origin=]'s
+    [=effective domain=]. This default MAY be overridden by the caller, as long as the caller-specified [=RP ID=] value [=is a
+    registrable domain suffix of or is equal to=] the caller's [=environment settings object/origin=]'s [=effective domain=].
+    See also [[#createCredential]] and [[#getAssertion]].
+
+    <div class="note">
+        Note: [=Public key credential=]'s scope is for a [=[RP]=]'s [=origin=], with the following <em>restrictions</em> and
+        <em>relaxations</em>:
+            - The  scheme MUST be `https` (i.e., <em>a restriction</em>), and,
```
equalsJeffH: thx, fixed in my local copy.

index.bs
```diff
@@ -1722,7 +1775,8 @@ WebAuthn supports multiple attestation types:
 
 ### Generating an Attestation Object ### {#generating-an-attestation-object}
 
-This section specifies the algorithm for generating an [=attestation object=] for any [=attestation statement format=].
+This section specifies the algorithm for generating an [=attestation object=] (see: [Figure 3](#fig-attStructs)) for any
```
selfissued: Including a literal figure number "Figure 3" is likely to cause numbering problems later.  If possible, please have this number be automatically generated by the tool, rather than included literally.

index.bs
```diff
@@ -293,23 +309,38 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
...
+    [=effective domain=]. This default MAY be overridden by the caller, as long as the caller-specified [=RP ID=] value [=is a
+    registrable domain suffix of or is equal to=] the caller's [=environment settings object/origin=]'s [=effective domain=].
+    See also [[#createCredential]] and [[#getAssertion]].
+
+    <div class="note">
+        Note: [=Public key credential=]'s scope is for a [=[RP]=]'s [=origin=], with the following <em>restrictions</em> and
+        <em>relaxations</em>:
+            - The  scheme MUST be `https` (i.e., <em>a restriction</em>), and,
+            - the host MAY be equal to the [=[RP]=]'s [=origin=]'s [=effective domain=], or it MAY be equal to a registrable
+                domain suffix of the [=[RP]=]'s [=origin=]'s [=effective domain=] (i.e., <em>an available relaxation</em>), and,
```
equalsJeffH: See: https://github.com/w3c/webauthn/pull/464#issuecomment-303785687

index.bs
equalsJeffH: I trust you realize that I realize the issue here :)
The `<figure>` and  `<figcaption>` elements are a part of HTML, and AFAICT neither bikeshed, nor HTML, at this time provides auto-numbering for `<figure>`.  Please correct me if I've got this wrong...

index.bs
```diff
@@ -386,14 +404,15 @@ a different [=origin=], by incorporating the [=origin=] in its responses. Specif
...
 is created as well as in all assertions produced by WebAuthn credentials.
 
-Additionally, to maintain user privacy and prevent malicious [=[RPS]=] from probing for the presence of credentials belonging to
-other [RPS], each credential is also associated with a Relying Party Identifier, or RP ID. This RP ID is provided by the client
-to the authenticator for all operations, and the authenticator ensures that credentials created by a [=[RP]=] can only be used
-in operations requested by the same RP ID. Separating the [=origin=] from the RP ID in this way allows the API to be used in
-cases where a single [=[RP]=] maintains multiple [=origins=].
+Additionally, to maintain user privacy and prevent malicious [=[RPS]=] from probing for the presence of [=public key
+credentials=] belonging to other [=[RPS]=], each [=public key credential|credential=] is also associated with a [=Relying Party
+Identifier=], or [=RP ID=]. This [=RP ID=] is provided by the client to the [=authenticator=] for all operations, and the
```
equalsJeffH: this is a topic for a discussion of our overall so-far ad-hoc spec source editing conventions and techniques (e.g. the use of regex for search and replace)

equalsJeffH:
@AngeloKai said:
> I am just hoping someone can explain to me why an RP should allow credential to be relaxed to a registerable domain suffix?

please see #241 "enforce strict same-origin policy on rpid" and your comment on it before we closed it: https://github.com/w3c/webauthn/issues/241#issuecomment-264337464

see also: https://github.com/w3c/webauthn/pull/464#issuecomment-303785687

AngeloKai:
@equalsJeffH I am fully supportive of the RP ID work and the ability of sharing credentials across sub-domains. AFAIK registerable domain suffix means new domain suffix, such as .aux instead of top level domain such as google.com. Is it necessary to allow credentials' origin to be relaxed to .aux?

index.bs
selfissued: Please proceed to merge then.  If we find better syntax for this, we can always add it later.

equalsJeffH:
@AngeloKai replied:
> AFAIK registerable domain suffix means new domain suffix, such as .aux instead of top level domain such as google.com

nope, a "registrable domain", as it has become to be used (AFAIK largely due to @mikewest) is equivalent to eTLD+1, though the present defintions and such at https://publicsuffix.org/list/ could use some polishing.  I've attempted to do such, see the updated definitions I've suggested at https://github.com/publicsuffix/publicsuffix.org/pull/2#issuecomment-281520351

index.bs
```diff
@@ -293,23 +309,38 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
...
+:: A [=domain=] identifier for the [=[RP]=] on whose behalf a given [=registration=] or
+    [=authentication|authentication ceremony=] is being performed. An [=RP ID=] MUST be a [=valid domain string=]. A [=public
+    key credential=] can only be used for [=authentication=] with the same entity (as identified by [=RP ID=]) it was registered
+    with. By default, the [=RP ID=] for a WebAuthn operation is set to the caller's [=environment settings object/origin=]'s
+    [=effective domain=]. This default MAY be overridden by the caller, as long as the caller-specified [=RP ID=] value [=is a
+    registrable domain suffix of or is equal to=] the caller's [=environment settings object/origin=]'s [=effective domain=].
+    See also [[#createCredential]] and [[#getAssertion]].
+
+    <div class="note">
+        Note: [=Public key credential=]'s scope is for a [=[RP]=]'s [=origin=], with the following <em>restrictions</em> and
```
jyasskin: "A [=public key credential=] ..."?

index.bs
```diff
@@ -293,23 +309,38 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
...
+        <em>relaxations</em>:
+            - The scheme MUST be `https` (i.e., <em>a restriction</em>), and,
```
jyasskin: Don't use MUST, etc. from non-normative sections, like Notes. Lower-case "must" is ok, but it's even better to avoid RFC2119 terms entirely when possible. e.g. "The scheme is always `https`. The host can be ..."

index.bs
```diff
@@ -293,23 +309,38 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
...
+            - the host MAY be equal to the [=[RP]=]'s [=origin=]'s [=effective domain=], or it MAY be equal to a registrable
+                domain suffix of the [=[RP]=]'s [=origin=]'s [=effective domain=] (i.e., <em>an available relaxation</em>), and,
+            - all (TCP) ports on that host (i.e., <em>a relaxation</em>).
```
jyasskin: nit: this bullet isn't parallel with the other two. Use a complete sentence here.

index.bs
```diff
@@ -515,25 +536,37 @@ When this method is invoked, the user agent MUST execute the following algorithm
...
+    then return a {{DOMException}} whose name is "{{SecurityError}}" and terminate this algorithm.
+<!-- PLACEHOLDER - https://github.com/w3c/webauthn/issues/474
+    Note: An [=effective domain=] MUST be a [=valid domain=]. Formats other than [=domain=] -- i.e., [=ipv4 address=],
+        [=ipv6 address=], [=opaque host=], or [=empty host=] -- are disallowed.
+-->
+
+<!-- Note: this next step is actually a top-level step, but bikeshed wanted it indented this much in order to compile w/o errors
+-->
+       <li id='CreateCred-DetermineRpId'>
+           If |options|.{{MakeCredentialOptions/rp}}.{{PublicKeyCredentialEntity/id}} is [=present|not present=], then set
```
jyasskin: Minor nit: I'd probably write this as:
```
1. Let |rpId| be |effectiveDomain|.
<li id='CreateCred-DetermineRpId'>
If |options|.{{MakeCredentialOptions/rp}}.{{PublicKeyCredentialEntity/id}} is [=present=]:

    1. ... Set |rpId| to ...
</li>
```

That makes it a little clearer what the default is, [uses "let" to introduce the variable](https://infra.spec.whatwg.org/#variables), and could let you omit the Note about the default a couple lines down.

index.bs
jyasskin: Unfortunately, ?w=1 doesn't let me leave comments in the same view.

index.bs
```diff
@@ -681,68 +714,80 @@ credentials are acceptable to it. The user agent and/or platform locates credent
...
-    This method takes the following parameters:
+This method accepts a single argument:
 
-    :   |options|
-    ::  A {{CredentialRequestOptions}} object, containing a challenge that the selected authenticator is expected to
-        sign to produce the assertion, and additional options as described in [[#assertion-options]]
-</div>
+<dl dfn-type="argument" dfn-for="PublicKeyCredential/[[DiscoverFromExternalSource]](options)">
+    :   <dfn>options</dfn>
+    ::  This argument is a  {{CredentialRequestOptions}} object whose |options|["{{CredentialRequestOptions/publicKey}}"] member
```
jyasskin: nit: You have an extra space after 'a', and I'd usually write options.publicKey instead of options["publicKey"].

index.bs
```diff
@@ -762,12 +807,12 @@ When this method is invoked, the user agent MUST execute the following algorithm
 
     1. Let |credentialList| be a new [=list=].
 
-    1. If <code>|publicKeyOptions|.{{PublicKeyCredentialRequestOptions/allowList}}</code> [=list/is not empty=], execute a
+    1. If <code>|options|.{{PublicKeyCredentialRequestOptions/allowList}}</code> [=list/is not empty=], execute a
```
jyasskin: This is as far as I got today. If the rest is just editorial cleanups, the stuff I looked at is good to merge.

index.bs
equalsJeffH: I submitted https://github.com/tabatkins/bikeshed/issues/1032

index.bs
```diff
@@ -293,23 +309,38 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
...
 
 : <dfn>Relying Party Identifier</dfn>
 : <dfn>RP ID</dfn>
-:: An identifier for the [=[RP]=] on whose behalf a given registration or authentication ceremony is being performed. Public Key
-    credentials can only be used for authentication by the same entity (as identified by RP ID) that created and registered
-    them. By default, the RP ID for a WebAuthn operation is set to the [=environment settings object/origin=] specified by the
-    [=relevant settings object=] of the {{CredentialsContainer}} object. This default can be overridden by the caller subject
-    to certain restrictions, as specified in [[#createCredential]] and [[#getAssertion]].
+:: A [=domain=] identifier for the [=[RP]=] on whose behalf a given [=registration=] or
+    [=authentication|authentication ceremony=] is being performed. An [=RP ID=] MUST be a [=valid domain string=]. A [=public
```
equalsJeffH: Ok, sure, thx. done in my local copy.

index.bs
```diff
@@ -293,23 +309,38 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
...
+    key credential=] can only be used for [=authentication=] with the same entity (as identified by [=RP ID=]) it was registered
+    with. By default, the [=RP ID=] for a WebAuthn operation is set to the caller's [=environment settings object/origin=]'s
+    [=effective domain=]. This default MAY be overridden by the caller, as long as the caller-specified [=RP ID=] value [=is a
+    registrable domain suffix of or is equal to=] the caller's [=environment settings object/origin=]'s [=effective domain=].
+    See also [[#createCredential]] and [[#getAssertion]].
+
+    <div class="note">
```
equalsJeffH: done in my local copy.

index.bs
equalsJeffH: I somewhat disagree w/ "...avoid RFC2119 terms entirely..."  -- they are helpful when properly employed, IMHO (I've written a few RFCs and this is my first W3C spec).  agreed that this Note is not the correct place to wield them.

equalsJeffH: done in my local copy.

index.bs
equalsJeffH: not sure what u mean by "parallel".

index.bs
equalsJeffH: done in my local domain tho I'm inclined to leave the Note intact in order to provide explicit context to the reader.

index.bs
equalsJeffH: ok, well, I need to leave early today and likely wont get back to this until tmrw morning, so there's time for further review...

index.bs
equalsJeffH: hm, I'm not sure who wrote `|options|["{{CredentialRequestOptions/publicKey}}"] member` -- possibly @mikewest as part of PR #384 et al  (?) .... so I've just been leaving that alone since I'm not yet an Adept.  in source form would it be `|options|.{{CredentialRequestOptions/publicKey}}` ?

index.bs
equalsJeffH: mumble.  from an editing perspective, to me, folding lines at some agreed-to length, and re-formatting them as we go along such that the prose parags are easier to read trumps pretty much everything else.  this is part of why I do kdiff3 text-only diffs and also use daisydiff and do not rely on only github's source diff display....

index.bs
jyasskin: I usually use `<code>|options|.{{CredentialRequestOptions/publicKey}}</code>`.

index.bs
jyasskin: You have clauses for two bullets, but only a noun phrase for the third, and you ought to have the same grammatical form for all of them. Also, "The scheme must be https, the host may be equal ..., and all ports on that host." doesn't read well.

index.bs
jyasskin: I don't think the IETF and the W3C are really different here, so you're probably right that there are cases in non-normative contexts where the terms make sense. Most uses I've seen have been confusing.

index.bs
equalsJeffH: fixed in my local copy.

index.bs
equalsJeffH: I think u misunderstood me -- the capitalized RFC2119 terms must only be used in a normative context.   We _could_ declare the Terminology section to be normative, dunno about that. but in any case, keeping RFC2119 terms contained to the algorithms and other gnarly innards is fine by me.

index.bs
equalsJeffH: oh shiza -- now I'm regretting shining-on sentence parsing in 7th grade. my teacher was right! it came back to haunt me! [I'm doomed](https://cdn.meme.am/cache/instances/folder53/500x/57742053/enraged-spongebob-oh-no-im-doomed.jpg)....

index.bs
equalsJeffH: alrightie, thx.

equalsJeffH: just love all those angle brackets... is there a bikeshed issue addressing this? like, we shudn't havta use no stinkin' html....

equalsJeffH: done in 0637185

equalsJeffH:
@jyasskin -- I just pushed https://github.com/w3c/webauthn/pull/464/commits/06371859d0fe10fe23c4b1862758a67cd16c686e -- also you did not get thru entire spec yesterday -- there are some normative changes in range of new line 810 to new line 2000, but most of them you probably saw already.  changes after new line 2000 are essentially editorial markup and source reformatting. 

do you wish to further review prior to merging this?

index.bs
```diff
@@ -681,68 +712,81 @@ credentials are acceptable to it. The user agent and/or platform locates credent
...
-    1. Set |rpId| to the {{PublicKeyCredentialRequestOptions/rpId}}.
+1. Let |effectiveDomain| be the |callerOrigin|'s [=effective domain=]. If |effectiveDomain| is null, then return a
+    {{DOMException}} whose name is "{{SecurityError}}" and terminate this algorithm.
+<!-- PLACEHOLDER - see https://github.com/w3c/webauthn/issues/474
+    Note: An [=effective domain=] MUST be a [=valid domain=]. Formats other than [=domain=] -- i.e., [=ipv4 address=],
+        [=ipv6 address=], [=opaque host=], or [=empty host=] -- are disallowed.
+-->
+
+        <li id='GetAssn-DetermineRpId'>
+            If |options|.{{PublicKeyCredentialRequestOptions/rpId}} is [=present|not present=], then set |rpId| to |effectiveDomain|.
```
jyasskin: Same comment here as above, that using "Let |rpId| be |effectiveDomain|. If rpId is present ... <change it>" may be a bit clearer, and they should match in any case.

index.bs
```diff
@@ -1406,19 +1452,20 @@ The [=authenticator data=] structure is a byte array of 37 bytes or more, as fol
...
-However, it differs from other [=client data=] in some important ways. First, unlike the client data, the RP ID of a credential
-does not change between operations but instead remains the same for the lifetime of that credential. Secondly, it is validated
-by the authenticator during the [=authenticatorGetAssertion=] operation, by verifying that the RP ID associated with the
-requested credential exactly matches the RP ID supplied by the client.
+The [=RP ID=] is originally received from the client when the credential is created, and again when an assertion is generated.
+However, it differs from other [=client data=] in some important ways. First, unlike the client data, the [=RP ID=] of a
+credential does not change between operations but instead remains the same for the lifetime of that credential. Secondly, it is
+validated by the authenticator during the [=authenticatorGetAssertion=] operation, by verifying that the [=RP ID=] associated
+with the requested credential exactly matches the [=RP ID=] supplied by the client, and that the [=RP ID=] [=is a registrable
+domain suffix of or is equal to=] the [=effective domain=] of the RP's [=origin=]'s [=effective domain=].
```
jyasskin: "the [=effective domain=] of the RP's [=origin=]'s [=effective domain=]." has two "effective domain"s. Pick one. :) I'd probably use "origin's effective domain" just because it's a bit shorter, but both are correct.

index.bs
```diff
@@ -456,53 +475,55 @@ that are returned to the caller when a new credential is created, or a new asser
...
-To support obtaining assertions via {{CredentialsContainer/get()|navigator.credentials.get()}}, this document extends the
-{{CredentialRequestOptions}} dictionary as follows:
+To support registration via {{CredentialsContainer/create()|navigator.credentials.create()}}, this document extends
+the {{CredentialCreationOptions}} dictionary as follows:
 
 <pre class="idl">
-    partial dictionary CredentialRequestOptions {
-        PublicKeyCredentialRequestOptions? publicKey;
+    partial dictionary CredentialCreationOptions {
+        MakeCredentialOptions? publicKey;
```
battre: Just noticed this after landing, but shouldn't this be scoped in some way? e.g. PublicKeyMakeCredentialsOptions

PR #476, opened 2017-05-29
Authnr sel keyprotection

### LINKED ISSUE #446: Add required key protection to AuthenticatorSelectionCriteria
Regarding https://w3c.github.io/webauthn/#authenticatorSelection, add criterion for required key protection.

Suggestion is to define
enum KeyProtection {
"SW",
"HW",
"TEE",
"SE"}

If requireResidentKey=false, then this criterion will be ignored.  If requireResidentKey=true, then the UA will raise exception if the key protection level does not meet level specified.

rlin1:
Close #446

***
[Preview](https://s3.amazonaws.com/pr-preview/w3c/webauthn/authnr-sel-keyprotection.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/webauthn/a2a4210...e53d515.html)


rlin1:
The big PR 442 was split into smaller PRs.  This is the one addressing keyProtection

index.bs
```diff
@@ -1053,8 +1053,9 @@ associated.
 
 <xmp class="idl">
     dictionary AuthenticatorSelectionCriteria {
-        Attachment    attachment;
```
jyasskin: It's not worth lining up the dictionary members.

balfanz:
What is the use case for this? I don't see the need for this functionality in the first version of the spec.

jcjones:
I don't see the usefulness of this even in L2. This seems like it would fragment adoption, which is bad.

nadalin:
Per 03/07/19 F2F Close

PR #477, opened 2017-05-29
Authnr sel matcherprotection

### LINKED ISSUE #447: Add required matcher protection to AuthenticatorSelectionCriteria
Regarding https://w3c.github.io/webauthn/#authenticatorSelection, add criterion for required matcher protection.  This field will describe the RP requirement for what type of environment the matching algorithm used for user verification is performed. 

Suggestion is to define
enum MatcherProtection {
"SW",
"TEE",
"HW"}

rlin1:
Close #447

***
[Preview](https://s3.amazonaws.com/pr-preview/w3c/webauthn/authnr-sel-matcherprotection.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/webauthn/a2a4210...a3521f0.html)


rlin1:
The big PR 442 was split into smaller PRs.  This is the one addressing matcherProtection

rlin1:
The recent commits should address that

balfanz:
I don't think we need this functionality in the first version of the spec. Let's table this for a later version of the spec.

index.bs
```diff
@@ -3409,6 +3434,13 @@ Axel Nennker, Kimberly Paulhamus, Adam Powers, Yaron Sheffer, Mike West, Jeffrey
     "status": "FIDO Alliance Implementation Draft"
   },
 
+  "FIDO-Allowed-Restricted-Operating-Environments": {
+    "authors": ["L. Lundblade", "M. Karlsson"],
+    "title": "FIDO Authenticator Allowed Restricted Operating Environments List",
+    "href": "https://fidoalliance.org/wp-content/uploads/fido-authenticator-allowed-restricted-operating-environments-list-2.pdf",
+    "status": "FIDO Alliance Final Document"
```
jyasskin: The link you have doesn't look like a final document: is it really an editor's draft until it's published to https://fidoalliance.org/specs?

index.bs
```diff
@@ -598,6 +598,8 @@ When this method is invoked, the user agent MUST execute the following algorithm
         to |authenticator|'s attachment modality, [=iteration/continue=].
     1. If {{AuthenticatorSelectionCriteria/requireResidentKey}} is set to |true| and the |authenticator| 
         is not capable of storing a [=Client-Side-Resident Credential Private Key=], [=iteration/continue=].
+    1. If {{AuthenticatorSelectionCriteria/matcherProtection}} is [=present|present=] and the
+        to |authenticator|'s matcher protection method is not included in that list, [=iteration/continue=].
```
jyasskin: I don't see the matcher protection method in the getInfo operation. Does the client really know it, or does this information need to be sent down to the authenticator for it to do the filtering?

index.bs
```diff
@@ -1104,6 +1112,29 @@ a minimum of friction, e.g., the user will not have to dig around in their pocke
...
+
+<pre class="idl">
+    enum MatcherProtection {
+       "rich-operating-system-software", // need to break Rich-OS kernel in order to by-pass user verification for miusing the key
+       "trusted-execution-environment",  // need to break the Trusted-Execution-Environment 
+                                         //  in order to by-pass user verification for miusing the key
+       "external-chip"};                 // need to break the external chip in order to by-pass user verification for miusing the key
+</pre>
+
+The term <dfn>trusted execution environment</dfn> (<dfn>TEE</dfn>) refers to any <a>restricted operating environment</a> that
```
jyasskin: I don't see "trusted execution environment" defined in https://fidoalliance.org/wp-content/uploads/fido-authenticator-allowed-restricted-operating-environments-list-2.pdf.

index.bs
```diff
@@ -1104,6 +1112,29 @@ a minimum of friction, e.g., the user will not have to dig around in their pocke
 example of the latter, when the user is accessing the [=[RP]=] from a given client for the first time, they may be required to
 use a [=roaming authenticator=] which was originally registered with the [=[RP]=] using a different client.
 
+### User Verification Protection enumeration (enum <dfn enum>UserVerificationProtection</dfn>) ### {#user-verification-protection}
+
+<pre class="idl">
+    enum MatcherProtection {
+       "rich-operating-system-software", // need to break Rich-OS kernel in order to by-pass user verification for miusing the key
```
jyasskin: You define "restricted operating environment" below, rather than "rich operating system software". You should use the same terms in both places. I like "restricted operating environment" better, although I'm not confident that's the best term for this.

index.bs
jyasskin: Instead of "the term X refers to Y", which makes it hard to tell whether X is an "authenticator’s matcher protection method", try to word this as "An authenticator's user matching is protected by `"external-chip"` if ..."

index.bs
```diff
@@ -598,6 +598,8 @@ When this method is invoked, the user agent MUST execute the following algorithm
         to |authenticator|'s attachment modality, [=iteration/continue=].
     1. If {{AuthenticatorSelectionCriteria/requireResidentKey}} is set to |true| and the |authenticator| 
         is not capable of storing a [=Client-Side-Resident Credential Private Key=], [=iteration/continue=].
+    1. If {{AuthenticatorSelectionCriteria/matcherProtection}} is [=present|present=] and the
```
jyasskin: Typo: "and the to authenticator's"

index.bs
```diff
@@ -1104,6 +1112,29 @@ a minimum of friction, e.g., the user will not have to dig around in their pocke
 example of the latter, when the user is accessing the [=[RP]=] from a given client for the first time, they may be required to
 use a [=roaming authenticator=] which was originally registered with the [=[RP]=] using a different client.
 
+### User Verification Protection enumeration (enum <dfn enum>UserVerificationProtection</dfn>) ### {#user-verification-protection}
+
+<pre class="idl">
+    enum MatcherProtection {
```
jyasskin: Seeing you call this "matcher protection" reinforces my belief that we should rename "user verification" to something like "user matching" or "person matching" since it's all about ensuring that the person getting an assertion matches the person who created the key pair.

index.bs
rlin1: It is the final one.

index.bs
rlin1: Corrected.

akshayku:
I don't think we need this. I would vote to close this.

rlin1:
as decided on the call: close with no action.

Issue #445, opened 2017-05-04
Add biometric accuracy to AuthenticatorSelectionCriteria


gmandyam:
Regarding https://w3c.github.io/webauthn/#authenticatorSelection, add criteria for biometric accruacy.

Suggestion is 

dictionary BiometricAccuracy {
double FRR;  //Maximum allowable False Rejection Rate
double FAR; //Maximum allowable False Acceptance Rate
double EER;// Maximum allowable equal error rate
};

UA would raise exception if available authenticator is a biometric, its accuracy is known, and the authenticator cannot satisfy the limits provided in the selection criteria.

rlin1:
I suggest adding such changes to branch authnr-sel-addtl-fields and the related PR #442

nadalin:
Closing for now as there is no interest

PR #479, opened 2017-05-29
Authnr selection aaguidlist

### LINKED ISSUE #478: Move aaguidList from extensions to AuthenticatorSelectionCriteria
see branch authnr-sel-aaguidlist

rlin1:
Close #478

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/pull/479.html" title="Last updated on Dec 15, 2017, 9:44 PM GMT (34caca4)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/479/986d627...34caca4.html" title="Last updated on Dec 15, 2017, 9:44 PM GMT (34caca4)">Diff</a>


rlin1:
The big PR 442 was split into smaller PRs.  This is the one addressing aaguidList

index.bs
```diff
@@ -1067,8 +1068,21 @@ associated.
     ::  This member describes the [=[RPS]=]' requirements regarding availability of the [=Client-side-resident Credential 
         Private Key=]. If the parameter is set to true, the authenticator MUST create a 
         [=Client-side-resident Credential Private Key=] when creating a [=public key credential=].
+
+    :   <dfn>aaguidList</dfn>
+    ::  If this member is [=present|present=], eligible authenticators are filtered to the ones identified by matching
```
jyasskin: AAGUID filtering needs to happen as a step in https://w3c.github.io/webauthn/#op-make-cred or https://w3c.github.io/webauthn/#createCredential depending on whether the Client or Authenticator has the relevant information, not in the definition of the dictionary field.

rlin1:
>What has changed to start mandating it for all implementations?
In the beginning of the WebAuthn spec process, some people were convinced that RPs wouldn't have the need to influence the authenticator selection.

At this stage, we already have seen that there is a need to the RP to influence the authenticator selection in order to improve the user experience.  As a consequence we have been adding the AuthenticatorSelectionCriteria object to the core specification.

The discussion regarding the fields in AuthenticatorSelectionCriteria shows that it is difficult to determine now what the right criteria for authenticator selection will be in the future.

So in my opinion there is substantial value for one generic method allowing the RP to formulate *any* rule for influencing the authenticator selection.  
More specifically, I am convinced that the ability for providing a list of authenticator model names (i.e. AAGUIDs) provides exactly that.  It allows the RP to use whatever set of criteria they want on the server side to generate such (prioritized) list and then make it easy for the platform to follow that list.

>Also, why is the AAGUID the right thing to filter on rather than something like a certificate in the attestation chain?

The AAGUID was introduced as an abstraction for the authenticator model name.  I think it is less bulky to send over a (potentially large) list of AAGUIDs than sending over a (potentially) even larger list of acceptable attestation certificates.
Note: One authenticator model might even have several attestation certificates in use (for the various batches of authenticator instances of such model).

index.bs
```diff
@@ -598,6 +598,8 @@ When this method is invoked, the user agent MUST execute the following algorithm
         to |authenticator|'s attachment modality, [=iteration/continue=].
     1. If {{AuthenticatorSelectionCriteria/requireResidentKey}} is set to |true| and the |authenticator| 
         is not capable of storing a [=Client-Side-Resident Credential Private Key=], [=iteration/continue=].
+    1. If {{AuthenticatorSelectionCriteria/aaguidList}} is [=present|present=] and the
+        to |authenticator|'s AAGUID protection method is not included in that list, [=iteration/continue=].
```
leshi: what's a "AAGUID protection method"?

balfanz:
Can someone explain how AAGUIDs are used today? I personally am not aware of any authenticators that use AAGUIDs, so I would be inclined to not include this functionality in the spec. 

But I do think such authenticators exist. Can @rlin1 or someone else explain how this functionality would be used? What could you do that you can't do without this PR, and why is that important?

rlin1:
TODO: add reference to FIDO Metadata Service as an example on where/how to get information on authenticator models.

index.bs
rlin1: Copy&paste error. Corrected that

rlin1:
Response to https://github.com/w3c/webauthn/pull/479#issuecomment-305259394:

I am not aware of any webauthn authenticators today.  I am aware of UAF and U2F authenticator.
UAF authenticators use a slightly different notion of a "model name" called AAID - and they implement that.
In U2F there is no equivalent of a model name, the closest thing to it is the list of attestationCertificateKeyIdentifiers.

In UAF the RP is able to influence Authenticator selection for "makeCredential" based on the AAID (and other criteria).  The authenticator selection based AAID can be used to support arbitrary criteria to be implemented by the RP - even without requiring dedicated support for the in the API.
For example: the RP could select authenticator for registration that meet some specific security certification etc.

I don't know how the RP could do that in U2F or in WebAuthentication - without this aaguidList approach.

nadalin:
Please update this by 6/27 so we can discuss on 6/28 call

rlin1:
changes requested by @leshi have been addressed.

equalsJeffH:
please merge-from-master (into this branch) and fix build errors, to facilitate further reviews, thx.

equalsJeffH:
I merged-from-master and fixed conflicts.

akshayku:
Overall, this PR looks fine to me but this is breaking WD07 which we all agreed was feature complete and probably not needed. Correct me if I am wrong here. 

There are two kinds of RPs, one who care about particular authenticators (for whatever reason, legal or otherwise and I don't want to rehash that discussion) and ones who don't care. Both of these can be addressed by current spec via direct attestation or null attestation(default case). This PR will short circuit faster (platform deciding earlier instead of RP) but essentially both ways achieve the same thing. If I remember correctly, there are other selection criterias (in addition to AAGUID) also which were pushed to L2 sometime back.

Should we take care of this in L2 or do we need it now?

akshayku:
I also see few more selection criteria (Matcher Protection #447, Key Protection #446, Biometric accuracy #445) that we have moved to L2. I would suggest all new selection criteria PRs/Issues (Like this one, Biometric Criteria #510) to be addressed in L2.

rlin1:
Decided on call to move to L2.

nadalin:
Per 03/07/19 F2F Close

PR #487, opened 2017-06-05
Omitted default abstract header

herrjemand:
Default abstract header causes two abstract headers to appear.
issue

Added omit to boilerplate options.

***
[Preview](https://s3.amazonaws.com/pr-preview/herrjemand/webauthn/ommited-default-abstract-header.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/webauthn/46b3933...herrjemand:174656d.html)

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -37,7 +37,7 @@ Abstract: This specification defines an API enabling the creation and use of str
  [=Authenticators=] provide cryptographic proof of their properties to [=relying parties=] via [=attestation=]. This
  specification also describes the functional model for WebAuthn conformant [=authenticators=], including their signature and
  [=attestation=] functionality.
-Boilerplate: omit conformance, omit feedback-header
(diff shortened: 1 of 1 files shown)


equalsJeffH:
LGTM, thx!

herrjemand:
@equalsJeffH happy to help*)

equalsJeffH:
AFAICT we should simply merge this.

equalsJeffH:
we may not actually "need" this because rendered spec that I gen using 
`curl https://api.csswg.org/bikeshed/ -F file=@index.bs -F force=1 > index.html` 
does not have double "Abstract" header, and neither does https://www.w3.org/TR/webauthn/

however I tested adding it and processing thru the above `curl` command and the "Abstract" header is rendered correctly so there is perhaps no downside to having it.

Issue #392, opened 2017-04-03
U2F Attestation only lists Basic Attestation as supported


rlin1:
See section 7.6 "FIDO U2F Attestation Statement Format"
Attestation types supported: Basic

In principle the authenticator could also generate self-signed attestation certificates which meet the definition of self-attestation as described in section 5.3.3.
Plus the current Mozilla Firefox U2F implementation uses this concept.

So I suggest adding self-attestation of the list of supported attestation types in section 7.6.

equalsJeffH:
we should do this

emlun:
Now the spec instead doesn't say how to detect when the `fido-u2f` attestation statement format uses self attestation. It seems like that would be when `attStmt.x5c[0]` is self-signed, correct? Either way I think the verification procedure in [§7.6][1] should make some mention of self attestation, as currently

- the only occurence of the word "self" in §7.6 is in the list of supported attestation types; and
- the final step of the verification procedure says to always return attestation type Basic. While this is technically correct as the returned `x5c` is indeed the entire trust path, it is confusing.

[1]: https://www.w3.org/TR/webauthn/#fido-u2f-attestation

Issue #351, opened 2017-02-15
References to "algorithm" and "alg" should be same string


jovasco:
*ScopedCredentialParameters* uses "algorithm" and *Packed Attestation Statement* uses "alg".

The first string is passed to and the second string passed from the authenticator. They should use the same string either "algorithm" or "alg" mostly for consistency and readability but there are space considerations as well: now we need 2 string constants in the applet.

I understand this is done for brevity of the reply but I question if this actually makes a useful difference:  an attestation reply should be approaching 1k depending on the certificate, so what do the 6 bytes matter? or the 50 or so you might save over all the strings?

And if that matters, why not do it for the incoming data as well? Each byte there adds just as much to the round trip delay.

zcorpan:
One possible reason for spelling it "alg" is to avoid misspellings of "algorithm". 🙂

jovasco:
Excellent reason :)

selfissued:
Given space/bandwidth constraints of some authenticator communication channels, we should use the shorter form "alg" in both cases.

selfissued:
We should change this to "alg" in PublicKeyCredentialParameters since it's sent on the wire to the authenticator.

Issue #329, opened 2017-01-21
Too generic names?


bfgeek:
A lot of the IDL names are a little too generic in the spec?

E.g. 
`Attachment` -> `CredentialAttachment` ?
`Transport` -> ???
`Account` -> ???
The above don't particularly matter as they aren't exposed...

But a closer look at:
`clientData` -> ???
`getAssertion` - no idea what this is doing from the method name?

equalsJeffH:
see also #312

selfissued:
This (and other renaming issues) affects the API, and so should be decided for the Implementer's Draft.

selfissued:
@equalsJeffH will make proposals.

equalsJeffH:
only two left to address are..

Attachment -> CredentialAttachment ?
Transport -> ???

AngeloKai:
@equalsJeffH @bfgeek I'd prefer having a concrete proposal if we were to make the code change. CredentialAttachment seems fine to me. For transport, I don't see major problem with it. Let's set a timeline of a week. If no one proposes a concrete proposal about the name change within a week, we should just close this.

bfgeek:
I'm not going to propose names for this working group as it's outside my area of expertise. Generic names are generally bad, as you aren't within a silo when you write a spec, you need to play name with the rest of the web platform.

For example, which would you name `Transport` if RTC already claimed this name? E.g.  `RTCIceTransportPolicy` was named `Transport` for some reason.

The TAG will probably have similar feedback to this effect.

rlin1:
authenticatorAttachment and authenticatorTransport?

equalsJeffH:
@rlin -- works for me :)

Issue #327, opened 2017-01-21
excludeList and allowList should be excludeCredentials and allowCredentials


bfgeek:
In dictionary types a sequence list is typically a plural name, e.g.
https://webbluetoothcg.github.io/web-bluetooth/#dictdef-requestdeviceoptions

https://w3c.github.io/webauthn/#dom-assertionoptions-allowlist -> `allowCredentials`
https://w3c.github.io/webauthn/#dom-scopedcredentialoptions-excludelist -> `excludeCredentials`

?

selfissued:
This seems like a reasonable renaming request, for consistency reasons.  Let's do this for the Implementer's Draft and then close the issue.

PR #489, opened 2017-06-05
cleanup easy issues WD-06

equalsJeffH:
this fixes the easier of the "easy" issues: #285 #326 #327 #329 #351 #392 #490 #494, please review.

***
[Preview](https://s3.amazonaws.com/pr-preview/w3c/webauthn/jeffh-cleanup-easy-issues-wd-06.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/webauthn/d8b103a...fe2d92e.html)


equalsJeffH:
This is an easy-peasy one (i hope) -- pls review :)

index.bs
```diff
@@ -1219,7 +1219,7 @@ following Web IDL.
     dictionary CollectedClientData {
         required DOMString           challenge;
         required DOMString           origin;
-        required DOMString           hashAlg;
+        required DOMString           hashAlgorithm;
```
jyasskin: IIUC, #326 and #351 say that the CollectedClientData and PublicKeyCredentialParameters fields should have consistent names. Here you've switched which one uses the short form, but they're still inconsistent. Maybe unify on "alg"?

index.bs
```diff
@@ -973,24 +973,24 @@ optionally evidence of [=user consent=] to a specific transaction.
 
     The <dfn>type</dfn> member specifies the type of credential to be created.
 
-    The <dfn>algorithm</dfn> member specifies the cryptographic signature algorithm with which the newly generated credential
+    The <dfn>alg</dfn> member specifies the cryptographic signature algorithm with which the newly generated credential
```
equalsJeffH: done in my local copy.

index.bs
equalsJeffH: well, AFAICT "alg" is only used in the spec IDL or prose in two places:
```
> grep -iE "alg[;\s}\.]+" index.bs 
        <code>|current|.{{PublicKeyCredentialParameters/alg}}</code> and |op| set to `"generateKey"`. If an error occurs during
        required AlgorithmIdentifier          alg;
```

..(tho my egrep fu is so-so) and so I'm inclined to leave this as-is with the vast majority of the use of [Aa]lgrorithm (85 instances) being the spelled-out form. 

that ok?

jyasskin: I don't feel strongly. @jovasco @bfgeek, do you?

equalsJeffH:
thanks for your review @jyasskin :)

index.bs
jovasco: I'm not up to date enough to understand which instance this is but please don't use multiple strings for the same thing towards the authenticator. It's all data we have to store and we win nothing. I rather use my <256k of flash storage for code and key storage than pretty member names.

index.bs
```diff
@@ -482,7 +482,7 @@ the {{CredentialCreationOptions}} dictionary as follows:
 
 <pre class="idl">
     partial dictionary CredentialCreationOptions {
-        MakeCredentialOptions? publicKey;
+        MakeCredentialOptions      publicKey;
```
battre: See my comment on the bug, I think this needs to stay nullable.

index.bs
```diff
@@ -493,7 +493,7 @@ To support obtaining assertions via {{CredentialsContainer/get()|navigator.crede
 
 <pre class="idl">
     partial dictionary CredentialRequestOptions {
-        PublicKeyCredentialRequestOptions? publicKey;
+        PublicKeyCredentialRequestOptions      publicKey;
```
battre: This needs to stay nullable in my opinion as well.

index.bs
battre: Taking this back after being educated by @jyasskin in issue #490 .

index.bs
battre: Taking this back as well.

index.bs
equalsJeffH: AFAICT, the short form, "alg", is used in all the data structures handled by authenticators. 

improved egrep for "alg":
```
> grep -niE "alg[:;\s}\.| ]+" index.bs
579:    1. Let |normalizedAlgorithm| be the result of [=normalizing an algorithm=] [[!WebCryptoAPI]], with |alg| set to
580:        <code>|current|.{{PublicKeyCredentialParameters/alg}}</code> and |op| set to `"generateKey"`. If an error occurs during
989:        required AlgorithmIdentifier          alg;
1711:            ; All public key formats must include an alg name
1712:            pubKeyTemplate = { alg: text }
1716:            rsaPubKey = { alg: rsaAlgName, n: biguint, e: uint }
1720:            eccPubKey = { alg: eccAlgName, x: biguint, y: biguint }
2075:                           alg: rsaAlgName / eccAlgName,
2080:                           alg: "ED256" / "ED512",
2114:    mechanism. It sets |x5c| to the certificate chain of the attestation public key and |alg| to the algorithm of the
2119:    through an authenticator-specific mechanism (see [[!FIDOEcdaaAlgorithm]]). It sets |alg| to the algorithm of the
2123:    and signing the result using the credential private key. It sets |alg| to the algorithm of the credential private key, and
2134:        attestation public key in |x5c| with the algorithm specified in |alg|.
2146:    - Validate that |alg| matches the algorithm of the credential private key in |authenticatorData|.
2148:        credential public key with |alg|.
2200:                            alg: rsaAlgName / eccAlgName,
2204:                            alg:  "ED256" / "ED512",
2273:        algorithm specified in |alg|.
2932:        ...                                         -- all public key alg etc.
2982:        ...                                         -- all public key alg etc.
3056:        ...                    -- all public key alg etc.
```

equalsJeffH:
@jyasskin cautioned:
> But do give folks a chance to reply about how to resolve the alg/algorithm inconsistency.

Of course.  I  think we are good wrt "alg" and "algorithm" at this point (we use the latter preponderantly).  I note that web crypto uses "alg" in various of its var names -- it's a common abbreviation. 

thanks everyone for your reviews. 

Merge this?

AngeloKai:
I am hoping @jcjones, @leshi, or @kpaulh can take a look at this before we merge.

kpaulh:
These changes make me happy. For my part, I'm fine with the current usage of alg vs algorithm.

equalsJeffH:
thx @selfissued 
> can you also please file a CTAP bug listing the changes made in WebAuthn that need to be tracked in CTAP? This at least includes "algorithm" -> "alg". Are there any others?

will do. 

> Are there any others?

I do not think so. 

NOTE:  In working with @selfissued to review this PR, I noticed some rough edges that I neglected to polish when doing PR #427, so I submitted issue #494, fixed that just now, and just  those latter changes can be quickly reviewed here: https://github.com/w3c/webauthn/pull/489/commits/fe2d92e4c6978d0ecd3e6b355b9b3185789b2a9d

thanks!

equalsJeffH:
@selfissued requested:
> can you also please file a CTAP bug listing the changes made in WebAuthn that need to be tracked in CTAP? This at least includes "algorithm" -> "alg". 

done see fido-alliance/fido-2-specs#252

index.bs
jcjones: Since this doesn't go on the wire, I'd prefer the spelled-out form. Not willing to die on the hill though.

AngeloKai:
I was talking with @selfissued about the spelling out thing. It seemed like a good idea to keep all the parameter names in the CTAP spec as numbers and create a table/map in the spec about how the numbers map to the names. This way we don't have to keep struggling with keeping the names short for slow transmission protocols, such as BLEs. (Ideas credit to @selfissued and AlexR) Should we perhaps wait a little bit on the merge until this is mauled over in the CTAP space? I am not the expert there so please let me know what you think.

jyasskin:
If the only reason to use an abbreviated name here is to reduce the encoded CBOR size for protocols like CTAP, and CTAP encodes the name as an integer rather than the Javascript field name, then it seems like we shouldn't abbreviate the name in Javascript. So, #351 would just be fixed by unifying on "algorithm".

jyasskin:
I wouldn't wait on merging the rest of this change, just remove the changes to `PublicKeyCredentialParameters.algorithm` from it before merging.

equalsJeffH:
@jyasskin mused:
> ...CTAP encodes the name as an integer rather than the Javascript field name,

Well, at this time, that is an if-CTAP-is-modified-to-do-so kind of statement.   So I suggest we merge this as-is and if CTAP goes that direction we can then "just remove the changes to PublicKeyCredentialParameters.algorithm".

Ok?

selfissued:
Please proceed with the merge today.  If the easy stuff is taking this long, it makes me fear how long the hard stuff may take us to do.

jyasskin:
I'm happy merging either way. I prefer to avoid making two changes to these names, but it's not a huge deal to me.

jovasco:
@AngeloKai 
I am not against changing the fields names to numbers but this has been discussed in the past and it was rejected back then.

Issue #301, opened 2016-11-09
Move {#sample-scenarios} (currently Section 10) to the top of the doc


vijaybh:
See TAG review comment at https://github.com/w3ctag/spec-reviews/issues/97#issuecomment-257799820

nadalin:
@selfissued is this already covered by #501 ?

PR #501, opened 2017-06-23
Merge {#sample-scenarios} section with the {#usecase} section

AngeloKai:
Addresses https://github.com/w3c/webauthn/issues/301. I removed a number of the use cases that aren't intended for Level 1 of the API. We can add an explainer-style document titled future use case to the spec for those use cases. The removal will clarify things for readers who aren't familiar with the API.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -126,49 +126,176 @@ as Universal Serial Bus (USB), Bluetooth Low Energy (BLE) or Near Field Communic
 
 ## Use Cases ## {#use-cases}
 
-The below use case scenarios illustrate use of two very different types of [=authenticators=], as well as outline further
-scenarios. Additional scenarios, including sample code, are given later in [[#sample-scenarios]].
+The following use case scenarios illustrate a typical flow involving registration and authentication with the sample code 
+included. 
 
-### Registration ### {#usecase-registration}
+### Regirstration ### {#usecase-registration}
 
-- On a phone:
-    * User navigates to example.com in a browser and signs in to an existing account using whatever method they have been using
-        (possibly a legacy method such as a password), or creates a new account.
-    * The phone prompts, "Do you want to register this device with example.com?"
-    * User agrees.
-    * The phone prompts the user for a previously configured [=authorization gesture=] (PIN, biometric, etc.); the user
-        provides this.
-    * Website shows message, "Registration complete."
+This is the first-time flow, in which a new credential is created and registered with the server.
+
+1. The user visits example.com, which serves up a script. At this point, the user must already be logged in using a legacy
+    username and password, or additional authenticator, or other means acceptable to the [=[RP]=].
+
+2. The [=[RP]=] script runs the code snippet below.
+
+3. The client platform searches for and locates the authenticator.
+
+4. The client platform connects to the authenticator, performing any pairing actions if necessary.
+
+5. The authenticator shows appropriate UI for the user to select the authenticator on which the new credential will be
+    created, and obtains a biometric or other authorization gesture from the user.
+
+6. The authenticator returns a response to the client platform, which in turn returns a response to the [=[RP]=] script. If
+    the user declined to select an authenticator or provide authorization, an appropriate error is returned.
+
+7. If a new credential was created,
+    - The [=[RP]=] script sends the newly generated [=credential public key=] to the server, along with additional information
+        such as attestation regarding the provenance and characteristics of the authenticator.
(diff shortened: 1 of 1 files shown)


AngeloKai:
I can go either way regarding the milestone. But this seems like a easy fix for a TAG review issue.

index.bs
```diff
@@ -126,49 +126,176 @@ as Universal Serial Bus (USB), Bluetooth Low Energy (BLE) or Near Field Communic
 
 ## Use Cases ## {#use-cases}
 
-The below use case scenarios illustrate use of two very different types of [=authenticators=], as well as outline further
-scenarios. Additional scenarios, including sample code, are given later in [[#sample-scenarios]].
+The following use case scenarios illustrate a typical flow involving registration and authentication with the sample code 
+included. 
 
-### Registration ### {#usecase-registration}
+### Regirstration ### {#usecase-registration}
```
selfissued: Please remove the spelling error introduced above.

index.bs
```diff
@@ -126,49 +126,176 @@ as Universal Serial Bus (USB), Bluetooth Low Energy (BLE) or Near Field Communic
...
-    * User agrees.
-    * The phone prompts the user for a previously configured [=authorization gesture=] (PIN, biometric, etc.); the user
-        provides this.
-    * Website shows message, "Registration complete."
+This is the first-time flow, in which a new credential is created and registered with the server.
+
+1. The user visits example.com, which serves up a script. At this point, the user must already be logged in using a legacy
+    username and password, or additional authenticator, or other means acceptable to the [=[RP]=].
+
+2. The [=[RP]=] script runs the code snippet below.
```
selfissued: I believe we settled on the editorial convention of labelling all steps as "1."  Bikeshed/Markdown will then do the correct numbering for us.  Please revise accordingly.

index.bs
```diff
@@ -3153,213 +3280,6 @@ IANA "WebAuthn Extension Identifier" registry established by [[!WebAuthn-Registr
...
-    navigator.credentials.get({ "publicKey": options })
-        .then(function (assertion) {
-        // Send assertion to server for verification
-    }).catch(function (err) {
-        // No acceptable credential or user refused consent. Handle appropriately.
-    });
-</pre>
-
-
-## Decommissioning ## {#sample-decommissioning}
```
selfissued: Did we intend to delete the decommissioning description?

nadalin:
@AngeloKai Please update

index.bs
```diff
@@ -126,49 +126,176 @@ as Universal Serial Bus (USB), Bluetooth Low Energy (BLE) or Near Field Communic
...
+This is the first-time flow, in which a new credential is created and registered with the server.
+
+1. The user visits example.com, which serves up a script. At this point, the user must already be logged in using a legacy
+    username and password, or additional authenticator, or other means acceptable to the [=[RP]=].
+
+2. The [=[RP]=] script runs the code snippet below.
+
+3. The client platform searches for and locates the authenticator.
+
+4. The client platform connects to the authenticator, performing any pairing actions if necessary.
```
equalsJeffH: these two steps could be combined?  also, would not pairing occur out-of-band from this flow? i.e., being paired is a pre-condition?

index.bs
```diff
@@ -126,49 +126,176 @@ as Universal Serial Bus (USB), Bluetooth Low Energy (BLE) or Near Field Communic
...
-    * The phone prompts, "Do you want to register this device with example.com?"
-    * User agrees.
-    * The phone prompts the user for a previously configured [=authorization gesture=] (PIN, biometric, etc.); the user
-        provides this.
-    * Website shows message, "Registration complete."
+This is the first-time flow, in which a new credential is created and registered with the server.
+
+1. The user visits example.com, which serves up a script. At this point, the user must already be logged in using a legacy
+    username and password, or additional authenticator, or other means acceptable to the [=[RP]=].
+
```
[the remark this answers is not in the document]
equalsJeffH: Further pre-conditions:
* is this flow for registering either a platform or a roaming authnr?
* is this device new-to-user ie no RP cookies extant at legacy sign-in time?
* is this flow for either UV-capable or 2nd-factor-only ?
* what subsequent use cases will this authnr be used for?  I.e., is it the "only" authnr for the RP, or is it intended to be the "primary" authnr (in which case it should be roaming?)?

equalsJeffH:
following up on @selfissued   , this PR is not just a "merge" of #sample-scenarios section with the #usecase section, it is a wholesale replacement of the former by the latter, which is suboptimal I believe because those two sections serve different audiences.

PR #510, opened 2017-07-26
Biometric Criteria Extension

gmandyam:
Biometric Selection Criteria extension

***
<a href="https://pr-preview.s3.amazonaws.com/gmandyam/webauthn/pull/510.html" title="Last updated on Feb 9, 2018, 12:23 AM GMT (971b383)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/510/ca4cf0f...gmandyam:971b383.html" title="Last updated on Feb 9, 2018, 12:23 AM GMT (971b383)">Diff</a>


index.bs
```diff
@@ -3093,6 +3093,43 @@ This [=registration extension=] and [=authentication extension=] enables use of
                     01           -- Subitem 3: CBOR short for Matcher Protection Type Software
     </pre>
 
+## Authenticator Biometric Criteria Extension (biometricBound) ## {#sctn-authenticator-biometric-criteria-extension}
+
+This [=registration extension=] allows a [=[RP]=] to specify the desired performance of a biometric authenticator
+that could be leveraged when creating the credential. 
+
+: Extension identifier
+:: `biometricBound`
```
equalsJeffH: the extension title is "Authenticator Biometric Criteria", yet the extension identifier is "biometricBound" which is inconsistent.  FAR and FRR are "biometric matching criteria", yes? Perhaps the latter should be "biometricMatchingCriteria" ?   And perhaps the section title ought to be "Biometric Authenticator Matching Criteria" ?

index.bs
```diff
@@ -3093,6 +3093,43 @@ This [=registration extension=] and [=authentication extension=] enables use of
...
+
+: Client extension input
+:: Biometric performance bounds:
+
+    <pre class="idl">
+    dictionary biometricCriteria{
```
equalsJeffH: perhaps `biometricCriteria` should be `authenticatorBiometricMatchingCriteria` ?

index.bs
```diff
@@ -3093,6 +3093,43 @@ This [=registration extension=] and [=authentication extension=] enables use of
...
+        };
+    </pre>
+
+    The FAR is the maximum false acceptance rate for a biometric authenticator allowed by the [=[RP]=].
+
+    The FAR is the maximum false rejection rate for a biometric authenticator allowed by the [=[RP]=].
+
+: Client extension processing
+:: This extension can only be used during {{CredentialsContainer/create()}}. If the client supports the Authenticator Biometric Criteria
+    Extension and biometric authenticators are available, it MUST use the first available biometric authenticator whose FAR and FRR match the bounds as provided.
```
equalsJeffH: we perhaps ought to define "biometric authenticators" up in terminology section, and then use `[=biometric authenticators=]` here.

index.bs
```diff
@@ -3093,6 +3096,43 @@ This [=registration extension=] and [=authentication extension=] enables use of
...
+
+: Client extension input
+:: Biometric performance bounds:
+
+    <pre class="idl">
+    dictionary authenticatorBiometricMatchingCriteria{
+        float FAR;
+        float FRR;
+        };
+    </pre>
```
equalsJeffH: IIUC, we ought to define the input in terms of CDDL, rather than WebIDL, yes?
see also issue #679 "add CDDL to every extension" and issue #626 "Extensions need to define how their parameters convert to/from CBOR"

index.bs
```diff
@@ -3093,6 +3096,43 @@ This [=registration extension=] and [=authentication extension=] enables use of
                     01           -- Subitem 3: CBOR short for Matcher Protection Type Software
     </pre>
 
+## Biometric Authenticator Matching Criteria Extension (biometricBound) ## {#sctn-authenticator-biometric-criteria-extension}
```
equalsJeffH: I would title it: 
"Biometric Authenticator Performance Bounds Extension"

index.bs
```diff
@@ -3093,6 +3096,43 @@ This [=registration extension=] and [=authentication extension=] enables use of
...
+
+This [=registration extension=] allows a [=[RP]=] to specify the desired performance of a biometric authenticator
+that could be leveraged when creating the credential.
```
This [=registration extension=] allows [=[RPS]=] to specify the desired performance bounds for selecting [=biometric authenticators=] as candidates to be employed in a [=registration=] ceremony.

index.bs
```diff
@@ -3093,6 +3096,43 @@ This [=registration extension=] and [=authentication extension=] enables use of
...
+        float FRR;
+        };
+    </pre>
+
+    The FAR is the maximum false acceptance rate for a biometric authenticator allowed by the [=[RP]=].
+
+    The FAR is the maximum false rejection rate for a biometric authenticator allowed by the [=[RP]=].
+
+: Client extension processing
+:: This extension can only be used during {{CredentialsContainer/create()}}. If the client supports the Authenticator Biometric Criteria
```
equalsJeffH: s/Authenticator Biometric Criteria Extension/this extension/

index.bs
```diff
@@ -3093,6 +3096,43 @@ This [=registration extension=] and [=authentication extension=] enables use of
...
+that could be leveraged when creating the credential. 
+
+: Extension identifier
+:: `biometricMatchingCriteria`
+
+: Client extension input
+:: Biometric performance bounds:
+
+    <pre class="idl">
+    dictionary authenticatorBiometricMatchingCriteria{
```
s/ authenticatorBiometricMatchingCriteria/ authenticatorBiometricPerfBounds/

gmandyam:
@equalsJeffH 

Thanks for review.  Tried to address your comments in latest PR.

Regarding CDDL, this extension is intended to terminate in the client and therefore has no required authenticator input.  My interpretation of https://w3c.github.io/webauthn/#sctn-client-extension-processing is that CBOR definition is only required when a "client extension input can be used to determine the CBOR authenticator extension input."

index.bs
```diff
@@ -3861,6 +3864,45 @@ This [=registration extension=] and [=authentication extension=] enables use of
...
+        };
+    </pre>
+
+    The FAR is the maximum false acceptance rate for a biometric authenticator allowed by the [=[RP]=].
+
+    The FAR is the maximum false rejection rate for a biometric authenticator allowed by the [=[RP]=].
+
+: Client extension processing
+:: This extension can only be used during {{CredentialsContainer/create()}}. If the client supports the this
+    extension and at least one [=biometric authenticator=] is available, it MUST use the first available biometric authenticator whose FAR and FRR match the bounds as provided.
```
emlun: First off: s/the this/this/

---

Second:

>If the client supports the this extension and at least one biometric authenticator is available, it MUST use the first available biometric authenticator whose FAR and FRR match the bounds as provided.

This formulation would likely be confusing with the new hot-plugging approach for `create()`. We no longer assume that there will be a single instant when you can list all "available" authenticators. Instead they will appear at some point (possibly immediately, e.g., platform authenticators) during the timeout, and the client can't know if the user will plug in an eligible biometric authenticator after the client has found a non-biometric authenticator. I therefore suggest reformulating this to something more like:

>If the client supports this extension, it MUST NOT use a biometric authenticator whose FAR or FRR does not match the bounds as provided.

gmandyam:
@emlun 

Thanks.  Changes made; please review.

index.bs
```diff
@@ -3861,6 +3864,45 @@ This [=registration extension=] and [=authentication extension=] enables use of
                     01           -- Subitem 3: CBOR short for Matcher Protection Type Software
     </pre>
 
+
+## Biometric Authenticator Performance Bounds Extension (biometricBound) ## {#sctn-authenticator-biometric-criteria-extension}
```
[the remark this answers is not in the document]
equalsJeffH: suggest:
s/biometricBound/biometricPerfBounds/

index.bs
```diff
@@ -3861,6 +3864,45 @@ This [=registration extension=] and [=authentication extension=] enables use of
...
+
+This [=registration extension=] allows [=[RPS]=] to specify the desired performance bounds for selecting [=biometric authenticators=]
+as candidates to be employed in a [=registration=] ceremony. 
+
+: Extension identifier
+:: `biometricMatchingCriteria`
```
s/biometricMatchingCriteria/biometricPerfBounds/

equalsJeffH:
wrt https://github.com/w3c/webauthn/pull/510#issuecomment-350798958 -- what do others think?

emlun:
Re https://github.com/w3c/webauthn/pull/510#issuecomment-354453032 : Seems reasonable to me. The `appid` and `authnSel` extensions work in the same way.

index.bs
```diff
@@ -3861,6 +3864,45 @@ This [=registration extension=] and [=authentication extension=] enables use of
...
+        };
+    </pre>
+
+    The FAR is the maximum false acceptance rate for a biometric authenticator allowed by the [=[RP]=].
+
+    The FAR is the maximum false rejection rate for a biometric authenticator allowed by the [=[RP]=].
+
+: Client extension processing
+:: This extension can only be used during {{CredentialsContainer/create()}}. If the client supports this
+    extension, it MUST NOT use a biometric authenticator whose FAR or FRR does not match the bounds as provided.
```
akshayku: How does client know these values? Are you expecting client to contact some authority who verifies it? FIDO metadata service?

gmandyam: @akshayku 

Yes, that is certainly one potential approach.  Expectation is that authenticator would've gone through FIDO Biometric Cert, had its performance verified, and agreed to publish its performance data on the FIDO Metadata Service (MDS).

selfissued: This whole thing seems really brittle.  It's more likely to provide a false sense of security than to actually be accurate.

gmandyam:
@selfissued 

I am sorry; I don't understand your exact concern.

>Please consider creating a separate specification for this proposed extension. 

Why is https://fidoalliance.org/specs/fido-uaf-v1.2-rd-20171128/fido-metadata-statement-v1.2-rd-20171128.html#dictionary-biometricaccuracydescriptor-members insufficient for this purpose?  Do you believe FRR and FAR are not sufficiently defined in this specification, and therefore a new specification is needed?

>This whole thing seems really brittle. It's more likely to provide a false sense of security than to actually be accurate.

I don't understand this concern as written.  Perhaps you can explain further.  Do you believe that FIDO MDS information on biometric accuracy will be inaccurate and/or unverified, for instance?

index.bs
equalsJeffH: so the above text perhaps ought to additionally indicate that various platform-specific means (eg some possibly based on the MDS) are anticipated for the client platform to use in order to have knowledge of these values.

equalsJeffH:
@selfissued 
> Please consider creating a separate specification for this proposed extension

Hi Mike -- it seems to me that you are proposing a process here that the WG has not itself decided upon and is up to the WG to impose rather than individual members.

emlun:
@selfissued 

>Providing no data of this is kind is better than providing data that provides a false sense of security.

How would you feel about this if the specified requirement in the extension, instead of

>If the client supports this extension, it MUST NOT use a biometric authenticator whose FAR or FRR **does not match** the bounds as provided.

was something more like this?

>If the client supports this extension, it MUST NOT use a biometric authenticator whose FAR or FRR **is not certified by an authoritative source to match** the bounds as provided.

equalsJeffH:
building on @emlun's fine suggestion, we perhaps ought to add a pointer to where one might obtain such metadata (as we have done elsewhere in the spec):
> If the client supports this extension, it MUST NOT use a biometric authenticator whose FAR or FRR is not certified by an authoritative source to match the bounds as provided. For example, the FIDO Metadata Service [FIDOMetadataService] provides one way to access such information.

I note in [FIDO Metadata Service](https://fidoalliance.org/specs/fido-uaf-v1.2-rd-20171128/fido-metadata-statement-v1.2-rd-20171128.pdf) section 3.3 "BiometricAccuracyDescriptor dictionary" that there is provision for conveying FAR, FRR, ERR, FAAR and other applicable metadata values.

emlun:
By the way I'd also like to note that in the end it falls on the RP to verify any claimed performance certification by inspecting the attestation statement, and not blindly trust the client to do this verification. It is thus not _critical_ to the integrity of the protocol that the client can derive these properties perfectly or even reliably, but it is of course nice if the extension can enable failing early.

nadalin:
@equalsJeffH @emlun So the extentions are vendor spepcifc that is they can do what they want, so including specifics may not be the correct way to go as that may dictate a specific implementaion

equalsJeffH:
I think what @emlun is teasing-out is a possible impl-cons Note or something similar, that's all.

equalsJeffH:
After the merge conflicts are resolved, I'm now OK with this being merged.

nadalin:
@gmandyam can you fix the branch conflicts and merge

gmandyam:
@nadalin 

Yes - I am working on it.  I am getting a Travis build failure currently and am trying to debug.

Issue #512, opened 2017-07-28
Is AlgorithmIdentifier value a string or an object?


selfissued:
The examples for AlgorithmIdentifier use values such as "ES256" and "RS256" - both in WebAuthn and CTAP.  But if the WebCrypto normalization rules result in these always being objects, we should consider either using JOSE algorithm values to keep them strings or fix the examples.  I think that authenticator implementers will be much happier with strings than objects.

selfissued:
This was closed by PR #514 .

Issue #509, opened 2017-07-26
Attestation Data defines ECC key X, Y parameters two ways


jcjones:
Web Authentication defines the Attestation Data's Credential Public Key in the CBOR map as having parameters `x` and `y` of type `biguint` [1], and then refers to RFC7518 section 6.2.1.{2,3} [2] for the details:

> "The semantics and naming of the other fields (though not their encoding) follows the definitions in [RFC7518] section 6"

RFC7518 defines `x` and `y` as base64url-encoded strings; presumably the language "_though not their encoding_" means to ignore that and go with them as being type `biguint`, but it'd be nice to be more explicit, perhaps with an example or saying something like:

> "The semantics and naming of the other fields follow the definitions in [RFC7518] section 6, except where the CBOR-defined encoding differs, where Web Authentication's definition has supremacy."

[1] https://www.w3.org/TR/webauthn/#sec-attestation-data
[2] https://tools.ietf.org/html/rfc7518#section-6.2.1.2

AngeloKai:
@selfissued can we close this issue now that 514 is merged?

AngeloKai:
Accidentally click on the wrong button.

jcjones:
Solved by COSE key formats

Issue #366, opened 2017-02-27
Credential CBOR


apowers313:
A few thoughts on simplifying the attestation credential CBOR in [Section 5.3.1](https://www.w3.org/TR/webauthn/#sec-attestation-data):
* Convert the CBOR map to a CBOR array / list with values at well-known positions in the array: 0 = alg; 1 = x; 2 = y; etc. This makes the final size smaller, since CBOR maps use strings for the indexes. Currently the difference is only two bytes ("alg" = 3 bytes; "x" = 1 byte; "1" = 1 byte; total 5 bytes as opposed to 3 bytes when using array indexes), but if new algorithms are added that have long strings for keys it may make a bigger difference.
* The CDDL currently specifies using "biguint" and "uint" for the various crypto parameters. These are essentially [semantically tagged](https://tools.ietf.org/html/rfc7049#section-2.4) byte strings. I would suggest just using byte strings instead of "biguint" and "uint". Although the current spec is is more semantically correct, adding semantic tags is just extra overhead and only creates difficulties when trying to encode / decode with off-the-shelf CBOR libraries.

equalsJeffH:
this comment on now-merged-and-closed PR #321 is perhaps apropos to this issue: https://github.com/w3c/webauthn/pull/321#discussion_r96979991

selfissued:
When we're representing keys in CBOR, I believe that we should use the COSE_Key format defined in https://tools.ietf.org/html/draft-ietf-cose-msg-24 (which is about to become an RFC).  As a rule, we should avoid defining different data formats when standard ones are defined that will work well for our application.

apowers313:
@selfissued I assume you're talking about the COSE_Key format illustrated in [Section C.7.1](https://tools.ietf.org/html/draft-ietf-cose-msg-24#page-118)?

selfissued:
Yes, and defined in [Section 7](https://tools.ietf.org/html/draft-ietf-cose-msg-24#section-7) .

apowers313:
@selfissued Is it just me, or do the key parameters in [Section 13](https://tools.ietf.org/html/draft-ietf-cose-msg-24#section-13) (and specifically 13.1) only describe how to use COSE for EC and not RSA?

selfissued:
[...]

PR #514, opened 2017-08-02
Use COSE_Key and COSE Algorithm Identifiers

selfissued:
This implements the decision we documented in issue #366 to use the COSE_Key data structure to represent keys, rather than our current ad-hoc structure.  This PR also fixes issue #509 and issue #512 by using COSE algorithm identifiers, such as -7 for "ES256" - eliminating the current ambiguities and discrepancies in algorithm representations.

***
[Preview](https://s3.amazonaws.com/pr-preview/selfissued/webauthn/mbj-COSE_Key.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/webauthn/e1faf59...selfissued:b2b1793.html)


selfissued:
This fixes #509 .
This fixes #512 .

nadalin:
Please review @jyasskin , @equalsJeffH , @jcjones

index.bs
```diff
@@ -579,16 +571,14 @@ When this method is invoked, the user agent MUST execute the following algorithm
             {{CredentialsContainer/create()}}.
     </li>
 
-1. Let |normalizedParameters| be a new [=list=] whose [=list/items=] are pairs of {{PublicKeyCredentialType}} and a
-    [=dictionary=] type (as returned by [=normalizing an algorithm=]).
+1. Let |normalizedParameters| be a new [=list=] whose [=list/items=] are pairs of {{PublicKeyCredentialType}} and
+    an {{AlgorithmIdentifier}}.
```
equalsJeffH: the type `{{AlgorithmIdentifier}}` is now a dangling pointer in that it is not itself defined.

index.bs
```diff
@@ -1414,11 +1404,13 @@ the {{CredentialsContainer/create()}} or {{CredentialsContainer/get()}} methods.
 </div>
 
 
-### Cryptographic Algorithm Identifier (type {{AlgorithmIdentifier}}) ### {#alg-identifier}
-
-A string or dictionary identifying a cryptographic algorithm and optionally a set of parameters for that algorithm. This type is
-defined in [[!WebCryptoAPI]].
+### Cryptographic Algorithm Identifier (type <dfn>AlgorithmIdentifier</dfn>) ### {#alg-identifier}
 
+A number or string identifying a cryptographic algorithm.
```
equalsJeffH: there needs to be IDL here defining `AlgorithmIdentifier` -- I'm working on that now.

index.bs
```diff
@@ -1783,30 +1774,9 @@ credential. It has the following format:
...
-            eccAlgName = "ES256" / "ES384" / "ES512"
-            ```
-
-            Thus, each public key type is a CBOR map starting with an entry named `alg`, which contains a text string that
-            specifies the name of the signature algorithm associated with the credential private key, using values defined in
-            [[!RFC7518]] section 3.1. The semantics and naming of the other fields (though not their encoding) follows the
-            definitions in [[!RFC7518]] section 6. Specifically, for ECC keys, the semantics of the `x` and `y` fields are
-            defined in [[!RFC7518]] sections 6.2.1.2 and 6.2.1.3, while for RSA keys, the semantics of the `n` and `e` fields
-            are defined in [[!RFC7518]] sections 6.3.1.1 and 6.3.1.2.
+            The Credential ID is the credential public key encoded in COSE_Key format, as defined in Section 7 of [[!RFC8152]].
```
equalsJeffH: The Cred ID is separate from the Cred Pub Key. They are two different things.

index.bs
```diff
@@ -2534,7 +2504,7 @@ This attestation statement format is used with FIDO U2F authenticators using the
     :: The [=attestation signature=].
 
 : Signing procedure
-:: If the credential public key of the given credential is not of algorithm "ES256", stop and return an error.
+:: If the credential public key of the given credential is not of algorithm -7 ("ES256"), stop and return an error.
```
equalsJeffH: -7 should be `<code>-7</code>` -- I'm fixing that.

index.bs
```diff
@@ -3304,11 +3274,11 @@ The sample code for generating and registering a new key follows:
       pubKeyCredParams: [
         {
           type: "public-key",
-          alg: "ES256",
+          alg: -7 // "ES256" as registered in the IANA COSE Algorithms registry
```
equalsJeffH: we need refs for IANA COSE Algorithms registry and for JOSE Algs registry. I'm working on adding those.

index.bs
```diff
@@ -64,13 +64,6 @@ spec: TokenBinding; urlPrefix: https://tools.ietf.org/html/draft-ietf-tokbind-pr
         text: Token Binding
         text: Token Binding ID; url: section-3.2
 
-spec: WebCryptoAPI; urlPrefix: https://www.w3.org/TR/WebCryptoAPI/
-    type: dfn
-        text: normalizing an algorithm; url: dfn-normalize-an-algorithm
-        text: recognized algorithm name
-    type: dictionary
-        text: AlgorithmIdentifier; url: dfn-AlgorithmIdentifier
-
```
equalsJeffH: when one deletes stuff like this, one needs to run BikeShed and check for errors  -- in this case, we still use the "recognized algorithm name" from WebCrypto, and that is causing linking errors without at least that portion the above in the spec.

selfissued: Where we are using WebCrypto "recognized algorithm names" for the hash algorithms (which are listed at http://www.w3.org/TR/WebCryptoAPI/#sha-registration - such as "SHA-256"), maybe we should give an example value, such as "SHA-256".

Interestingly, the Travis-CI build succeeded for the PR without Bikeshed errors.

equalsJeffH:
fixes issue #366

Issue #527, opened 2017-08-09
revise approach of isPlatformAuthenticatorAvailable()


equalsJeffH:
@mikewest offers an alternative approach to isPlatformAuthenticatorAvailable() (embodied by the already-merged PR #379) here https://github.com/w3c/webauthn/pull/379#issuecomment-308037955

Also, discussion on pr #379's approach has been continuing post-merge (see @balfanz s https://github.com/w3c/webauthn/pull/379#issuecomment-320375807 and subsequent comments). 

I support @mikewest's and @balfanz's arguments and feel the spec should be updated to use the approach described in @mikewest's https://github.com/w3c/webauthn/pull/379#issuecomment-308037955

equalsJeffH:
discussion on webauthn call 9-Aug-2017 w/ @kpaulh @jyasskin @selfissued @AngeloKai , we decided to go with PR #523 rather than this approach.

PR #523, opened 2017-08-08
Fixed the example and other points listed in isPlatformAuthenticatorReady comment

AngeloKai:
The current state only fixes the example section. I chatted with @selfissued  who will add the timeout section.

***
[Preview](https://s3.amazonaws.com/pr-preview/AngeloKai/webauthn/angelo-findAuthnr.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/webauthn/73d4461...AngeloKai:7ac4b91.html)

PR #528, opened 2017-08-09
revision and polishing of the merged isPlatformAuthenticatorAvailable PR #379

equalsJeffH:
This addresses the issues described in #527 "revise approach of isPlatformAuthenticatorAvailable() subtype:algorithms"

***
[Preview](https://s3.amazonaws.com/pr-preview/w3c/webauthn/jeffh-revise-isplatformauthnrready-379.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/webauthn/73d4461...cc85505.html)

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -967,25 +967,25 @@ authorizing an authenticator with which to complete the operation.
 
 <div link-for-hint="WebAuthentication/isPlatformAuthenticatorAvailable">
 
-[=[RPS]=] use this method to determine whether they can create a new credential using a [=platform authenticator=].
+[=[RPS]=] may use this method to determine whether a [=platform authenticator=] is available and offer the user an opportunity
+to register a credential using it.
 Upon invocation, the [=client=] employs a platform-specific procedure to discover available [=platform authenticators=].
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -967,25 +967,25 @@ authorizing an authenticator with which to complete the operation.
...
+If this assessment is affirmative, the promise is resolved.
+Otherwise, the promise does not resolve.
 
 This method has no arguments and returns a boolean value.
 
 <pre class="idl">
     [SecureContext]
     partial interface PublicKeyCredential {
-        [Unscopable] Promise < boolean > isPlatformAuthenticatorAvailable();
+        [Unscopable] readonly attribute Promise<void> isPlatformAuthenticatorAvailable;
```
selfissued: Making this doesn't work because it means that there can't be UI.

selfissued: Making this an attribute doesn't work because then there can't be UI.

equalsJeffH:
discussion on webauthn call 9-Aug-2017 w/ @kpaulh @jyasskin @selfissued @AngeloKai , we decided to go with PR #523 rather than this approach.

Issue #104, opened 2016-05-13
Clarify which sign. algm.'s RP's must support (Sec. 4.3.2.1.2)


gmandyam:
Current text states "The following algorithms must be implemented by servers:" and goes on to list target signing methods.  But the text later says "WebAuthn Relying Parties must implement all the algorithms implemented by the authenticators that they support."

The two statements seem to contradict each other.  For instance, if an RP does not support any authenticators that implement ECDAA then it does not seem to be necessary for the RP to implement ECDAA.  Would recomment changing "The following algorithms must be implemented by servers:"  to "One or more of the following algorithms must be implemented by servers (depending on the authenticators that servers support):"

equalsJeffH:
at the time @rlin1 submitted this issue, S 4.3.2.1.2 mapped to this spot in index.src.html: `{#packed-attestation-signature}`.   we need to note such spec source anchors in issues because section #s will change over time as we move sections around.

vijaybh:
We already decided in Berlin to not specify server behavior in this spec.

Issue #537, opened 2017-08-21
Make create() and get() abortable


balfanz:
We need to make sure that create() and get() are abortable: For example, imagine a web page that gives users the ability to dismiss a dialog or popup that takes the user through an Authenticator registration or assertion-generation process. When the user dismisses that dialog or popup, the Authenticator should be notified to, say, stop blinking, callbacks waiting for the create() or get() results should be rejected, etc.

Presumably, we need to add an AbortSignal parameter to both calls to accomplish this: https://dom.spec.whatwg.org/#aborting-ongoing-activities

AngeloKai:
Adding this parameter is discussed here: https://github.com/w3c/webauthn/issues/380. I'd expect that on spec level, we just need to extend the WHATWG HTML spec.

equalsJeffH:
see also issue #283 
issue #316 may be related.

fwiw, terminology-wise, "abort(able)" is being used equivalently to "cancel(able)", both here in webauthn, and in whatwg-land, eg in the [abort/cancel proposal in DOM, Fetch, et al.](https://github.com/whatwg/fetch/issues/447): https://github.com/whatwg/fetch/issues/447

The latter is being implemented now, and is perhaps (ie needs investigation) what we ought to use our abort/cancel functionality. See: https://github.com/whatwg/fetch/issues/447#issuecomment-316378856, https://github.com/whatwg/fetch/issues/447#issuecomment-324038255

AngeloKai:
Resolving this issue would also resolve https://github.com/w3c/webauthn/issues/116

nadalin:
@AngeloKai Where do we stand with this ?

jcjones:
Closed by #544

PR #498, opened 2017-06-16
fixup algs contd 3

equalsJeffH:
this builds on PRs #371 and #495. it resolves these issues: 
improves #254 (#getAssertion section will be updated in a separate PR and issue #254 will be fixed at that time)
fix #466 
fix #472 
fix #536 
fix #559

***
[Preview](https://s3.amazonaws.com/pr-preview/w3c/webauthn/jeffh-fixup-algs-contd-3.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/webauthn/b8c6027...4b2193e.html)

Issue #380, opened 2017-03-14
Should the WebAuth API have a cancel() method?


kpaulh:
Hey gang,

The spec details an authenticatorCancel method, but the client doesn't necessarily know when to invoke it. We have seen a couple scenarios where not having a way to cancel an existing authenticator operation from Javascript might present an undesirable user experience. 

For example, having multiple tabs that are attempting operations and losing focus on one. The user opens two registration tabs. The device will blink until it receives two taps.
 
Since promises aren't themselves cancelable, it seems that a cancel() method is the best option - assuming these use cases are concerning enough to warrant one. (We're not entirely convinced that they are). We also briefly considered having smaller timeouts, say, 10 seconds, but this presents problems for bluetooth authenticators. 

Thoughts?

jyasskin:
I believe @jakearchibald is working on promise cancellation in the context of `fetch()` and may be able to say how WebAuth can match that.

jakearchibald:
We plan to represent cancellation by rejecting with an `AbortError`. Here's the latest proposal https://github.com/whatwg/fetch/issues/447#issuecomment-281731850.

vijaybh:
It seems that this puts good UI at the mercy of every web developer. This seems wrong.

The UA could for example cancel operations when a tab is occluded. But it seems like that would not require a Javascript method for the webpage to do so?

equalsJeffH:
see also #318 "Define what happens when the Document loses focus"

kpaulh:
I think you meant #316. I see. Right, the browser could certainly handle losing focus.
However, as another example, we also have scenarios involving in-page Javascript that can change UI without navigation or losing focus that I believe is currently being used for some security-key related flows.

kpaulh:
@mikewest boiled down the `fetch` proposal for canceling promises into a pretty simple mechanism with [https://github.com/whatwg/dom/pull/434](url). Rather than adding an explicit cancel() method, we would  `makeCredential` and `getAssertion` accept a `CancellationSignal`.

I believe this is a pretty good fit for our use-case - if we decided this was something we needed to do.

jyasskin:
The PR moved to https://github.com/whatwg/dom/pull/437.

equalsJeffH:
[...]

PR #544, opened 2017-09-06
Make create() and get() abortable 

AngeloKai:
NOTE: this is closely related to: https://github.com/w3c/webappsec-credential-management/pull/110 (and formerly #104), which is now (as of 2-Nov-2017) merged to master.

As @leshi explained to me, Google's current two step verification step involves prompting a wizard such as [this.](https://angelokai.github.io/Screenshot.png). If the user clicks cancel on the wizard, the UA needs to withdraw all the calls to authenticators, which should stop blinking subsequently. I'd call out the PR is not only to fulfill this niche use case but to give us a clean interface overall. 

The PR makes use of the https://dom.spec.whatwg.org/#abortsignal-aborted-flag) object to achieve such functionality. This would be an addition to the API interface. AFAIK, both Edge and Firefox have implemented AbortSignal. Chrome is working on the feature too. 

There is a corresponding PR standing at CredMan land: https://github.com/w3c/webappsec-credential-management/pull/104. The CredMan PR adds the abort functionality to the CredMan surface, though this may not be needed depending on @mikewest and Dominic's preference. 

The AbortSignal feature is new to HTML overall. Fetch is the first one to utilize the feature: https://github.com/whatwg/fetch/issues/447. However, we can't completely follow fetch because there are still nuances here and there.

***
[Preview](https://s3.amazonaws.com/pr-preview/AngeloKai/webauthn/abort.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/webauthn/bcb5c11...AngeloKai:78b0912.html)


AngeloKai:
The PR is based on JeffH's recent changes so there are many commits attached to this. After JeffH's PR is merged, I will rebase to make things clearer.

AngeloKai:
The PR aims at addressing the following issues: https://github.com/w3c/webauthn/issues/537, https://github.com/w3c/webauthn/issues/380, https://github.com/w3c/webauthn/issues/283, https://github.com/w3c/webauthn/issues/316, and finally https://github.com/w3c/webauthn/issues/292. 

https://github.com/w3c/webauthn/issues/292 will need both this PR and https://github.com/w3c/webauthn/pull/498

AngeloKai:
Admittedly, the process described in this PR are a little hand-wavy. Edge, for example, uses a fairly complicated heuristic model to avoid race condition and ensure operations are happening smoothly. For example, when multiple transports are hitting the same authenticators, there's a queuing mechanism in combination with timing heuristics to not have crazy errors. Because of the complication, I decided to not specify it further and limit the description to "loop through each authenticator and cancel things."

AngeloKai:
Regarding https://github.com/w3c/webauthn/issues/316, we have some clarification to do. Reading from Firefox's pull request and what I heard from U2F, the ask here is simply to stop authenticator operation and reject promise with an appropriate error when user navigates away from a page. If so, we should just hook up the abort operation with the https://html.spec.whatwg.org/multipage/webappapis.html#handler-window-onunload event. This would be the cleanest way to approach issues without touching hot buttons such as memory usage. If the ask is to automatically determine whether a button is visible and remove the authenticator operations once they stop being visible, we'd have many more complicated logic. 

In the second case, I'd suggest we'd leave it to the web developers to decide how they want to approach this, provided that we provide them with a hook such as the AbortSignal.

index.bs
```diff
@@ -53,6 +53,10 @@ Markup Shorthands: css off, markdown on
 
 <pre class="anchors">
 
+spec: DOM; urlPrefix: https://dom.spec.whatwg.org/
+    type: interface
+        text: AbortSignal; url: #abortsignal
```
annevk: This should not be needed as this is already exported.

index.bs
```diff
@@ -100,6 +104,7 @@ spec: mixed-content; urlPrefix: www.w3.org/TR/mixed-content/
 <pre class="link-defaults">
 spec:html; type:dfn; for:environment settings object; text:global object
 spec:infra; type:dfn; text:list
+spec:infra; type:dfn; for:struct; text:item
```
annevk: Same here.

index.bs
```diff
@@ -549,7 +554,8 @@ To support obtaining assertions via {{CredentialsContainer/get()|navigator.crede
 {{CredentialsContainer/create()|navigator.credentials.create()}} to request the creation of a new [=credential key pair=]
 and {{PublicKeyCredential}}, managed by an [=authenticator=].
 On success, the returned {{promise}} will be resolved with a {{PublicKeyCredential}} containing an
-{{AuthenticatorAttestationResponse}} object.
+{{AuthenticatorAttestationResponse}} object. This method can be <dfn id="concept-webauthn-terminated">terminated</dfn> 
+with flag [=AbortSignal/aborted flag=], which is unset unless otherwise specificed.
```
annevk: This seems wrong. How many reasons are there to be terminated? Just aborting?

index.bs
```diff
@@ -697,53 +702,68 @@ When this method is invoked, the user agent MUST execute the following algorithm
...
         and [=set/remove=] |authenticator| from |issuedRequests|.</dd>
 
-    <dt>If any |authenticator| returns a status indicating that the user cancelled the operation,</dt>
-    <dd>
-        1. [=set/Remove=] |authenticator| from |issuedRequests|.
+    :   If the <code>|options|.{{MakePublicKeyCredentialOptions/signal}}</code> is [=present=] and and the 
+        [=AbortSignal/aborted flag=] is set to false,
+    ::  [=set/For each=] |authenticator| in |issuedRequests| invoke the [=authenticatorCancel=] operation on |authenticator|
+        and [=set/remove=] |authenticator| from |issuedRequests|.</dd> Set the [=AbortSignal/aborted flag=] be true. Then 
+        reject the promise with an {{AbortError}} {{DOMException}}.
```
annevk: This seems wrong. If the passed in signal is not aborted, why would you abort?

index.bs
```diff
@@ -617,15 +621,16 @@ When this method is invoked, the user agent MUST execute the following algorithm
...
     1. If <code>|current|.{{PublicKeyCredentialParameters/type}}</code> does not contain a {{PublicKeyCredentialType}} supported
         by this implementation, then [=continue=].
     1. Let |alg| be <code>|current|.{{PublicKeyCredentialParameters/alg}}</code>.
     1. [=list/Append=] the pair of <code>|current|.{{PublicKeyCredentialParameters/type}}</code> and |alg| to
         |credTypesAndPubKeyAlgs|.
 
-1. If |credTypesAndPubKeyAlgs| [=list/is empty=] and |options|.{{MakePublicKeyCredentialOptions/pubKeyCredParams}}
-    [=list/is not empty=], cancel the timer started in step 2, return a {{DOMException}} whose name is "{{NotSupportedError}}",
+1. If |credTypesAndPubKeyAlgs| [=list/is empty=] and <code>|options|.{{MakePublicKeyCredentialOptions/pubKeyCredParams}}</code>
+    [=list/is not empty=], return a {{DOMException}} whose name is "{{NotSupportedError}}",
```
annevk: Return to what? Shouldn't this reject the promise?

index.bs
```diff
@@ -753,6 +773,54 @@ authorizing an authenticator.
 </div>
 
 
+#### Construct the resultant credential - `constructResultantCredentialCallback` method #### {#constructResultantCredentialCallback}
```
annevk: Should this really be part of this PR?

index.bs
```diff
@@ -697,53 +702,68 @@ When this method is invoked, the user agent MUST execute the following algorithm
...
     1. [=set/Append=] |authenticator| to |issuedRequests|.
 
-1. Start a timer for |adjustedTimeout| milliseconds. Then execute the following steps [=in parallel=]. The [=task source=] for
-    these [=tasks=] is the [=dom manipulation task source=].
+1. Start a timer for |adjustedTimeout| milliseconds.
 
-1. While |issuedRequests| [=list/is not empty=], perform the following actions depending upon the |adjustedTimeout| timer and
-    responses from the authenticators:
+1. [=While=] |issuedRequests| [=list/is not empty=], perform the following actions in parallel depending upon the |adjustedTimeout| 
+    timer and responses from the authenticators but abort if the operation is [=terminated=]:
```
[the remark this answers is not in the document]
jakearchibald: When synchronising parallel actions, instead of using a loop, I've "waited" on values. I'm not sure if this is a better system or not, @annevk what do you think?

In https://wicg.github.io/background-fetch/#attempt-a-background-fetch-algorithm I do something similar, but I set up variables like "completeFetches" (step 2) and "abandoned" (step 3). Step 6 does a load of parallel tasks, and may set the "abandoned" flag. Step 7 waits for these values to change.

index.bs
```diff
@@ -697,53 +702,68 @@ When this method is invoked, the user agent MUST execute the following algorithm
...
+            ::  whose value is |credentialCreationData|.
+
+            ISSUE: <code>[=constructResultantCredentialCallback=]</code> and
+                <code>[=credentialCreationDataInstance=]</code> should actually not be defined here (?) and link back
+                to their definitions in [[!CREDENTIAL-MANAGEMENT-1]] (?).
 
         5. [=set/For each=] remaining |authenticator| in |issuedRequests| invoke the [=authenticatorCancel=] operation on
             |authenticator| and [=set/remove=] it from |issuedRequests|.
+
         6. Return |value| and terminate this algorithm.
```
jakearchibald: You can't really "return" from in parallel steps.

index.bs
jakearchibald: It's a little odd, but this algorithm is run "in parallel", but it's started over at https://w3c.github.io/webappsec-credential-management/#abstract-opdef-create-a-credential. But yeah I think this algorithm should just take the promise.

index.bs
AngeloKai: There are two reasons why an create credential and get credential should be terminated: 1) JS code intentionally terminates the operation to provide better experience, 2) a document unloads and asks all authentication operations to be terminated. The second one is the one I am still not sure what to do with. Preferably there should be only one abort eventListener for the browser to know whether it should start aborting the request. Weird things like page unload should just set the abort flag.

index.bs
AngeloKai: @annevk It should return to the base algorithm by credential management, which then handles the rejection.

index.bs
```diff
@@ -697,53 +702,68 @@ When this method is invoked, the user agent MUST execute the following algorithm
         |credTypesAndPubKeyAlgs|, |excludeCredentialDescriptorList|, and |authenticatorExtensions| as parameters.
     1. [=set/Append=] |authenticator| to |issuedRequests|.
 
-1. Start a timer for |adjustedTimeout| milliseconds. Then execute the following steps [=in parallel=]. The [=task source=] for
-    these [=tasks=] is the [=dom manipulation task source=].
+1. Start a timer for |adjustedTimeout| milliseconds.
 
-1. While |issuedRequests| [=list/is not empty=], perform the following actions depending upon the |adjustedTimeout| timer and
-    responses from the authenticators:
+1. [=While=] |issuedRequests| [=list/is not empty=], perform the following actions in parallel depending upon the |adjustedTimeout|
```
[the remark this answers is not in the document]
AngeloKai: Well there are two main layers of parallel processing with this API: 
1. Parallel processing at the JS layer, i.e. the one where the cred man spec calls the PublicKeyCredential algorithm in parallel. 
1. Parallel processing at the authenticator layer. If more than one authenticators are connected to the platform, the PublicKeyCredential algorithm would send request in parallel to each authenticator. 

If you go a little extreme, you can say there is a third corner case where an native app may try to talk to authenticator while the web page is also talking to it. But it seems like very much a corner case and should be addressed at the device level. 

As far as aborting is concerned, what seems to be a reasonable is 1) check abort before credman goes parallel, 2) check abort before authenticator goes parallel, 3) check abort in each individual request out to authenticator and, if true, cancel each authenticator operation.

index.bs
AngeloKai: I see. https://wicg.github.io/background-fetch/#attempt-a-background-fetch-algorithm is really educational and helps quite a bit. But I prefer not to correct this in the abort PR. We can open up another PR to do parallel right. 

@mikewest @equalsJeffH does this mean we need to correct both credman and webauthn?

index.bs
AngeloKai: this is probably a leftover from a previous merge. I will take it out.

index.bs
equalsJeffH: > does this mean we need to correct both credman and webauthn?

I'm not sure.  This algorithm is invoked from credman as an internal method. The associated mods to credman are queued as credman PR https://github.com/w3c/webappsec-credential-management/pull/100  The credman steps that invoke this webauthn method/algorithm are themselves running in parallel. So saying "return" here seemed fine.

index.bs
equalsJeffH: you don't need to take it out if you merge from current master (which I think you did in a commit more recent than this one here)

index.bs
```diff
@@ -564,7 +562,8 @@ To support obtaining assertions via {{CredentialsContainer/get()|navigator.crede
 {{CredentialsContainer/create()|navigator.credentials.create()}} to request the creation of a new [=credential key pair=]
 and {{PublicKeyCredential}}, managed by an [=authenticator=].
 On success, the returned {{promise}} will be resolved with a {{PublicKeyCredential}} containing an
-{{AuthenticatorAttestationResponse}} object.
+{{AuthenticatorAttestationResponse}} object. This method can be <dfn id="concept-webauthn-terminated">terminated</dfn> 
+with flag [=AbortSignal/aborted flag=], which is unset unless otherwise specificed.
```
equalsJeffH: if we are modeling our termination and aborting functionality on [fetch](https://fetch.spec.whatwg.org), then this -- using [=AbortSignal/aborted flag=] here -- does not appear to be correct. In [the fetching alg](https://fetch.spec.whatwg.org/#fetching), `aborted` is a simple flag local to the fetch algorithm:
`... <dfn export for=fetch id=concept-fetch-terminate>terminated</dfn> with flag <var>aborted</var> ...`

equalsJeffH:
My initial review of this is here https://github.com/w3c/webauthn/pull/544#pullrequestreview-70971455 -- thx for taking this on Angelo

jcjones:
With agreement on the call, 25 Oct, we're going to make this the flagship PR to do in CR, and unblock WD-07 from this change. We're all in support that this needs to happen, but we still need more help on it.

index.bs
annevk: I wonder if we have a consistent story for unload. I suspect a lot of promises might end up forever pending whenever that happens.

index.bs
```diff
@@ -2183,6 +2206,8 @@ should be specified in the attestation certificate itself, so that it can be ver
 
 # [=[RP]=] Operations # {#rp-operations}
 
+[INFORMATIVE]
```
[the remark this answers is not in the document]
AngeloKai: It's not the job of this WG to define how RP does things. In fact it's not the job of the W3C or any organization to define how RP verifies things. The FIDO org can certify a server but doesn't define how the RP writes its code. No other W3C specs that have gone to REC attempt to define normative actions for the RP. No one can tell web developers they have to write their code in this way. 

I agree it should be a separate PR. I will start a separate one.

nadalin:
@angeloKai Why was this changed back to WD07 as the minutes say CR

jcjones:
@nadalin: We moved it back because we don't want breaking changes after WD-07, so this must block WD-07.

index.bs
```diff
@@ -3779,6 +3779,39 @@ extension for transaction authorization.
...
 
+## Aborting Authentication Operations ## {#sample-aborting}
+
+The below example showcases how a developer may use the AbortSignal parameter to abort an
+ongoing credential registration operation. Similiar procedure applies to an ongoing
+authentication operation. 
+
+<pre class="example" highlight="js">
+    const authAbortController = new AbortController(); 
+    const authAbortSignal = controller.signal;
```
jcjones: This'd be `authAbortController.signal`, yes?

index.bs
```diff
@@ -3754,6 +3779,39 @@ extension for transaction authorization.
...
+    // Assume widget shows up whenever auth occurs. 
+    if (widget = "disappear") {
+        authAbortSignal.abort(); 
+
+        authAbortSignal.onabort = function () {
+            // Migrate away from the page and other necessary clean-ups. 
+        }
+    }
+</pre>
+
```
equalsJeffH: the example is a nice addition.  IIUC, this example does not match up with the prior proposed spec mods further above. Is the plan to get the example correct and then we see what spec mods we actually need to accomodate it?

index.bs
```diff
@@ -3754,6 +3779,39 @@ extension for transaction authorization.
...
+<pre class="example" highlight="js">
+    const authAbortController = new AbortController(); 
+    const authAbortSignal = controller.signal;
+
+    var options = {
+        // A list  of options. 
+    }
+
+    navigator.credentials.create({"publicKey": options}, authAbortSignal)
```
equalsJeffH: IIUC, we were contemplating passing the signal as part of the `options`, rather than passing`authAbortSignal` as a separate arg?  

see also: https://github.com/w3c/webauthn/pull/544/files#r146071432

index.bs
AngeloKai: No, this exemplifies what the addition would look like toward the end. Please let me know what you believe the original proposal is.

index.bs
AngeloKai: This pull request demonstrates what it would look like to keep this in a separate arg. I intend to submit another pull request that shows what it'd look like to keep the arg inside option. 

The main argument for keeping it as a separate arg is that in the future when we allow multiple credentials to be requested in one operation, developers can abort them altogether. For example: 

    
    
    const authAbortController = new AbortController() 
    const authAbortSignal = controller.signal

    var publicKeyOptions = { // A list of public key related options.}
    var passwordOptions = { // A list of password related options.}
    var federatedOptions = { // A list of federated options.}

    navigator.credentials.get({
         {"publicKey": publicKeyOptions, 
          "password": passwordOptions,
          "federated": federatedOptions,
         }, authAbortSignal)
        .then(function (attestation) {
            // Register the user. 
        }).catch(function (error) {
            if (error = "AbortError") {
                // Let the server know a key hasn't been created. 
            }
        });

    // Assume widget shows up whenever auth occurs. 
    if (widget = "disappear") {
        authAbortSignal.abort(); 

        authAbortSignal.onabort = function () {
            // Migrate away from the page and other necessary clean-ups. 
        }
    }
`
I was convinced after discussion with @jyasskin. @jyasskin do you have more to add?

index.bs
```diff
@@ -570,7 +570,9 @@ To support obtaining assertions via {{CredentialsContainer/get()|navigator.crede
 {{CredentialsContainer/create()|navigator.credentials.create()}} to request the creation of a new [=credential key pair=]
 and {{PublicKeyCredential}}, managed by an [=authenticator=].
 On success, the returned {{promise}} will be resolved with a {{PublicKeyCredential}} containing an
-{{AuthenticatorAttestationResponse}} object.
+{{AuthenticatorAttestationResponse}} object. If an {{AbortSignal}} object is passed in,
+the developer can abort on going {{CredentialsContainer/create()|navigator.credentials.create()}} operation with
```
jyasskin: "can abort an ongoing"?

index.bs
```diff
@@ -735,6 +737,12 @@ When this method is invoked, the user agent MUST execute the following algorithm
         ::  [=set/For each=] |authenticator| in |issuedRequests| invoke the [=authenticatorCancel=] operation on |authenticator|
             and [=set/remove=] |authenticator| from |issuedRequests|.
 
+        :   If the <code>|options|.{{MakePublicKeyCredentialOptions/signal}}</code> is [=present=] and and the
```
jyasskin: "If <code>|options|..." and remove the second "and and".

index.bs
```diff
@@ -735,6 +737,12 @@ When this method is invoked, the user agent MUST execute the following algorithm
         ::  [=set/For each=] |authenticator| in |issuedRequests| invoke the [=authenticatorCancel=] operation on |authenticator|
             and [=set/remove=] |authenticator| from |issuedRequests|.
 
+        :   If the <code>|options|.{{MakePublicKeyCredentialOptions/signal}}</code> is [=present=] and and the 
+            [=AbortSignal/aborted flag=] is set to false,
```
jyasskin: If this is the step intended to handle the developer aborting the create() call, I think you mean "is set to true" here.

index.bs
jyasskin: Assuming you keep Jeff's suggestion to put the AbortSignal in CredentialCreationOptions instead of MakePublicKeyCredentialOptions (which I agree with, as mentioned in https://github.com/w3c/webappsec-credential-management/pull/104), you'll need to pull the signal out before step 2 where you overwrite |options|.

index.bs
```diff
@@ -735,6 +737,12 @@ When this method is invoked, the user agent MUST execute the following algorithm
...
+        ::  [=set/For each=] |authenticator| in |issuedRequests| invoke the [=authenticatorCancel=] operation on |authenticator|
+            and [=set/remove=] |authenticator| from |issuedRequests|.</dd> Set the [=AbortSignal/aborted flag=] be true. Then
```
jyasskin: But then I'm not sure why you'd change the aborted flag in this step.

jyasskin: If you just want a new paragraph inside the same `<dd>`, don't write `</dd>`, but instead put a blank line before this line.

index.bs
```diff
@@ -974,7 +982,8 @@ method is invoked, the user agent MUST:
 
     1. [=set/Append=] |authenticator| to |issuedRequests|.
 
-1. Execute the following steps [=in parallel=]. The [=task source=] for these [=tasks=] is the [=dom manipulation task source=].
+1. Execute the following steps [=in parallel=] but abort if the steps are [=terminated=]. The [=task source=] for these [=tasks=]
```
jyasskin: "Abort if the steps are terminated" isn't quite the right thing to say here. First, even Fetch says, ["if the ongoing fetch is terminated"](https://fetch.spec.whatwg.org/#concept-main-fetch), rather than saying steps might be terminated.

Second, aborting steps just stops running them. For WebAuthn, we need to run authenticatorCancel at the very least, which seems to mean you should use https://dom.spec.whatwg.org/#abortsignal-add. See the "Add the following abort steps" line in https://fetch.spec.whatwg.org/#dom-global-fetch for an example.

index.bs
```diff
@@ -985,6 +994,12 @@ method is invoked, the user agent MUST:
             ::  [=set/For each=] |authenticator| in |issuedRequests| invoke the [=authenticatorCancel=] operation on
                 |authenticator| and [=set/remove=] |authenticator| from |issuedRequests|.
 
+            :   If the {{PublicKeyCredentialRequestOptions/signal}} member is [=present=] and the [=AbortSignal/aborted flag=] is set to
```
jyasskin: Same comments here as for `MakePublicKeyCredentialOptions/signal` above.

index.bs
jyasskin: s/the/its/

index.bs
```diff
@@ -1770,7 +1797,7 @@ session to exist at any particular time, or by providing more complicated sessio
 The following operations can be invoked by the client in an authenticator session.
 
 
-<h4 id="op-make-cred" algorithm>The <dfn>authenticatorMakeCredential</dfn> operation</h4>
+<h4 id="op-make-cred" algorithm>The <dfn export>authenticatorMakeCredential</dfn> operation</h4>
```
jyasskin: Other specifications shouldn't be calling this operation, so we shouldn't export it. If it needs to be mentioned non-normatively, the other spec can [use a `spec` attribute to do so](https://tabatkins.github.io/bikeshed/#linking-to-unexported-definitions).

index.bs
```diff
@@ -3754,6 +3781,39 @@ extension for transaction authorization.
...
+
+<pre class="example" highlight="js">
+    const authAbortController = new AbortController(); 
+    const authAbortSignal = controller.signal;
+
+    var options = {
+        // A list  of options. 
+    }
+
+    navigator.credentials.create({"publicKey": options, authAbortSignal})
```
jyasskin: s/authAbortSignal/signal: authAbortSignal/

index.bs
```diff
@@ -3754,6 +3781,39 @@ extension for transaction authorization.
...
+The below example showcases how a developer may use the AbortSignal parameter to abort an
+ongoing credential registration operation. Similiar procedure applies to an ongoing
+authentication operation. 
+
+<pre class="example" highlight="js">
+    const authAbortController = new AbortController(); 
+    const authAbortSignal = controller.signal;
+
+    var options = {
+        // A list  of options.
```
jyasskin: Duplicated space here.

index.bs
```diff
@@ -3754,6 +3781,39 @@ extension for transaction authorization.
     });
 </pre>
 
+## Aborting Authentication Operations ## {#sample-aborting}
+
+The below example showcases how a developer may use the AbortSignal parameter to abort an
+ongoing credential registration operation. Similiar procedure applies to an ongoing
```
jyasskin: "A similar procedure".

index.bs
```diff
@@ -3754,6 +3781,39 @@ extension for transaction authorization.
     });
 </pre>
 
+## Aborting Authentication Operations ## {#sample-aborting}
+
+The below example showcases how a developer may use the AbortSignal parameter to abort an
```
jyasskin: s/showcases/shows/?

index.bs
```diff
@@ -3754,6 +3781,39 @@ extension for transaction authorization.
...
+
+    var options = {
+        // A list  of options. 
+    }
+
+    navigator.credentials.create({"publicKey": options, authAbortSignal})
+        .then(function (attestation) {
+            // Register the user. 
+        }).catch(function (error) {
+            if (error = "AbortError") {
```
jyasskin: `error.name ==`, right?

index.bs
```diff
@@ -3754,6 +3781,39 @@ extension for transaction authorization.
...
+                // Let the server know a key hasn't been created. 
+            }
+        });
+
+    // Assume widget shows up whenever auth occurs. 
+    if (widget = "disappear") {
```
jyasskin: `==`

index.bs
```diff
@@ -3754,6 +3781,39 @@ extension for transaction authorization.
...
+        authAbortSignal.abort(); 
+
+        authAbortSignal.onabort = function () {
```
jyasskin: This function will never run, since https://dom.spec.whatwg.org/#abortsignal-signal-abort fired the event 2 lines up.

index.bs
```diff
@@ -3754,6 +3781,39 @@ extension for transaction authorization.
...
+    var options = {
+        // A list  of options. 
+    }
+
+    navigator.credentials.create({"publicKey": options, authAbortSignal})
+        .then(function (attestation) {
+            // Register the user. 
+        }).catch(function (error) {
+            if (error = "AbortError") {
+                // Let the server know a key hasn't been created.
```
jyasskin: The exact error type probably doesn't affect what the server gets told. It might affect the UI shown to the user?

AngeloKai:
Btw the corresponding credman PR https://github.com/w3c/webappsec-credential-management/pull/110 is merged.

index.bs
AngeloKai: Discussed over the call that it matches expectation.

index.bs
```diff
@@ -1393,6 +1410,7 @@ an assertion. Its {{PublicKeyCredentialRequestOptions/challenge}} member must be
         unsigned long                        timeout;
         USVString                            rpId;
         sequence<PublicKeyCredentialDescriptor> allowCredentials = [];
+        AbortSignal                          signal;
```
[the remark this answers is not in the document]
jcjones: ... and it is now, so this parameter should be able to just be removed.

index.bs
equalsJeffH: AbortSignal was put in credman's CredentialCreationOptions via credman PR w3c/webappsec-credential-management#110

index.bs
```diff
@@ -735,6 +737,12 @@ When this method is invoked, the user agent MUST execute the following algorithm
         ::  [=set/For each=] |authenticator| in |issuedRequests| invoke the [=authenticatorCancel=] operation on |authenticator|
             and [=set/remove=] |authenticator| from |issuedRequests|.
 
+        :   If the <code>|options|.{{MakePublicKeyCredentialOptions/signal}}</code> is [=present=] and and the 
+            [=AbortSignal/aborted flag=] is set to false,
+        ::  [=set/For each=] |authenticator| in |issuedRequests| invoke the [=authenticatorCancel=] operation on |authenticator|
+            and [=set/remove=] |authenticator| from |issuedRequests|.</dd> Set the [=AbortSignal/aborted flag=] be true. Then 
+            reject the promise with an {{AbortError}} {{DOMException}}.
```
equalsJeffH: whoops, we can't reject promises here because we don't have any promises -- this is a synchronous alg that's called on a background thread and (now) returns a (callback) alg to the caller.  unfortunately, we've been working at a hectic pace and we have some loose ends in the spec, like mentioning promises where we now ought not.  See issue #671.

index.bs
equalsJeffH: plus we need to get PR #665 merged-to-master, then this branch merged-from-master in order to be working on the latest state of this alg.

index.bs
```diff
@@ -3754,6 +3781,43 @@ extension for transaction authorization.
...
+<pre class="example" highlight="js">
+    const authAbortController = new AbortController(); 
+    const authAbortSignal = authAbortController.signal;
+
+    authAbortSignal.onabort = function () {
+        // Migrate away from the page and other necessary clean-ups. 
+    }
+
+    var options = {
+        // A list of options.
```
equalsJeffH: should signal be in `options` like so:
```
   var options = {
       signal: authAbortSignal`,
        // additional options....
   }
```
..?  i.e., [per credman now](https://w3c.github.io/webappsec-credential-management/#credentialrequestoptions-dictionary)?

index.bs
```diff
@@ -3754,6 +3781,43 @@ extension for transaction authorization.
...
+        // Migrate away from the page and other necessary clean-ups. 
+    }
+
+    var options = {
+        // A list of options. 
+    }
+
+    navigator.credentials.create({
+        publicKey: options, 
+        signal: authAbortSignal})
```
equalsJeffH: `signal` is a part of options yes?
so this would be..
```
         publicKey: options
     })
```
..?

index.bs
equalsJeffH: agree w/@jyasskin that we may need to employ  https://dom.spec.whatwg.org/#abortsignal-add techniques similar to how [fetch](https://fetch.spec.whatwg.org/#fetch-method) handles aborting.  It also seems that fetch has a notion of "[An ongoing fetch can be terminated
with flag aborted](https://fetch.spec.whatwg.org/#concept-fetch-terminate)" which is local to the fetch spec, and not part of [the DOM's "aborting ongoing activities"](https://dom.spec.whatwg.org/#aborting-ongoing-activities). I wonder whether we might need to implement something similar.

index.bs
AngeloKai: the current design is that the signal can be a separate field (as described in the example and in my other comment).

AngeloKai:
@equalsJeffH @jyasskin @jcjones please review. This is now fully polished. Previously when I pushed commit, they were work in progress.

index.bs
AngeloKai: It seems like a bigger problem to fix here. If we do this here, it may be inconsistent. As jake mentioned in the other comment, our parallel calls here are problematic. I can work with you to submit a PR to clean this up.

index.bs
AngeloKai: @equalsJeffH I think authenticatorCancel should be sufficient to address the issue for now. We may need a more polished algorithm overall to describe the life cycle. But that is something we can do at PR time. I can collaborate with you and @jyasskin on that.

index.bs
```diff
@@ -547,6 +557,7 @@ the {{CredentialCreationOptions}} dictionary as follows:
 <pre class="idl">
     partial dictionary CredentialCreationOptions {
         MakePublicKeyCredentialOptions      publicKey;
+        AbortSignal                         signal;
```
equalsJeffH: hm, [webappsec-credential-management/#dictdef-credentialcreationoptions](https://w3c.github.io/webappsec-credential-management/#dictdef-credentialcreationoptions)  already has a s`signal` member of type `AbortSignal`

index.bs
```diff
@@ -558,6 +569,7 @@ To support obtaining assertions via {{CredentialsContainer/get()|navigator.crede
 <pre class="idl">
     partial dictionary CredentialRequestOptions {
         PublicKeyCredentialRequestOptions      publicKey;
+        AbortSignal                            signal;
```
equalsJeffH: ditto: [webappsec-credential-management/#dictdef-credentialrequestoptions](https://w3c.github.io/webappsec-credential-management/#dictdef-credentialrequestoptions)

index.bs
```diff
@@ -688,6 +702,10 @@ When this method is invoked, the user agent MUST execute the following algorithm
 
 1. Let |clientDataHash| be the [=hash of the serialized client data=] represented by |clientDataJSON|.
 
+1. If the <code>|options|.{{CredentialCreationOptions/signal}}</code> is [=present=] and its 
+    [=AbortSignal/aborted flag=] is set to true, return [=a promise rejected with=] an "{{AbortError}}"
```
equalsJeffH: we cannot return a promise here because we're just a synchronous algorithm; the Promise resolution/rejection is handled by `navigator.credentials.create()`.  see https://github.com/w3c/webappsec-credential-management/pull/100

[webappsec-credential-management/#algorithm-create](https://w3c.github.io/webappsec-credential-management/#algorithm-create) already is checking for `options.signal`'s aborted flag being set, but may need to also be doing it in the "in parallel" section.   maybe we need to have webauthn's `[[Create]]()` alg simply `return` as quickly as possible if an abort situation arises, and then have [webappsec-credential-management/#algorithm-create](https://w3c.github.io/webappsec-credential-management/#algorithm-create) check for abort right after webauthn's `[[Create]]()` alg returns to it ?

index.bs
```diff
@@ -735,6 +753,12 @@ When this method is invoked, the user agent MUST execute the following algorithm
         ::  [=set/For each=] |authenticator| in |issuedRequests| invoke the [=authenticatorCancel=] operation on |authenticator|
             and [=set/remove=] |authenticator| from |issuedRequests|.
 
+        :   If the <code>|options|.{{CredentialCreationOptions/signal}}</code> is [=present=] and its 
+            [=AbortSignal/aborted flag=] is set to true, 
+        ::  [=set/For each=] |authenticator| in |issuedRequests| invoke the [=authenticatorCancel=]
+            operation on |authenticator| and [=set/remove=] |authenticator| from |issuedRequests|. Then return [=a promise rejected with=]
```
equalsJeffH: ditto: can't reject promise here.  may need to use https://dom.spec.whatwg.org/#abortsignal-add magic here a la how fetch is doing it. e.g. see step 8 of https://fetch.spec.whatwg.org/#fetch-method

index.bs
```diff
@@ -906,6 +930,10 @@ method is invoked, the user agent MUST:
 
 1. Let |clientDataHash| be the [=hash of the serialized client data=] represented by |clientDataJSON|.
 
+1. If the <code>|options|.{{CredentialRequestOptions/signal}}</code> is [=present=] and its 
+    [=AbortSignal/aborted flag=] is set to true, return [=a promise rejected with=] an "{{AbortError}}"
```

[webappsec-credential-management/#algorithm-request](https://w3c.github.io/webappsec-credential-management/#algorithm-request) is checking `options.signal`'s aborted flag.  same thoughts as above, i.e., `[[DiscoverFromExternalSource]]()` needs to promptly return and [webappsec-credential-management/#algorithm-request](https://w3c.github.io/webappsec-credential-management/#algorithm-request) ought to check immediately for abort situation?

index.bs
```diff
@@ -985,6 +1013,12 @@ method is invoked, the user agent MUST:
             ::  [=set/For each=] |authenticator| in |issuedRequests| invoke the [=authenticatorCancel=] operation on
                 |authenticator| and [=set/remove=] |authenticator| from |issuedRequests|.
 
+            :   If the {{CredentialRequestOptions/signal}} member is [=present=] and the [=AbortSignal/aborted flag=] is set to 
+                true,
+            ::  [=set/For each=] |authenticator| in |issuedRequests| invoke the [=authenticatorCancel=] operation on |authenticator|
+                and [=set/remove=] |authenticator| from |issuedRequests|. Then 
+                return [=a promise rejected with=] an "{{AbortError}}" {{DOMException}}.
```
equalsJeffH: ditto wrt no promise to return and perhaps needing abortsignal-add magic

index.bs
equalsJeffH: yeah, doh, nevermind, sorry.

index.bs
```diff
@@ -1418,14 +1451,41 @@ an assertion. Its {{PublicKeyCredentialRequestOptions/challenge}} member must be
...
+The [=visibility states|visibility=] and [=focus=] state of the [=Window=] object determines whether the 
+{{PublicKeyCredential/[[Create]](options)}} and {{PublicKeyCredential/[[DiscoverFromExternalSource]](options)}} operations
+should continue. When the [=Window=] object associated with the [[=Document=] loses focus, 
+{{PublicKeyCredential/[[Create]](options)}} and {{PublicKeyCredential/[[DiscoverFromExternalSource]](options)}} operations
+SHOULD be aborted. 
+
+    Issue: The WHATWG HTML WG is discussing whether to provide a hook when a browsing context gains or
+    loses focuses. If a hook is provided, the above paragraph will be updated to include the hook.
+    See [WHATWG HTML WG Issue #2711](https://github.com/whatwg/html/issues/2711) for more details. 
+
```
equalsJeffH: above is helpful (nice!) but will need updating per however we end up polishing this abort functionality between webauthn & credman.

equalsJeffH:
further review: https://github.com/w3c/webauthn/pull/544#pullrequestreview-74229447  Thanks for tackling this @AngeloKai -- it's a gnarly one.

index.bs
AngeloKai: I added this to deal with a bikeshed linking error. I can take it out.

index.bs
AngeloKai: Does the note in the abort section help explain this? The reality is abort is a new concept to web platform APIs and our current spec languages don't handle this that well. This is why the section mentions we are checking the aborted flag three times to "fake" the effect of rejecting immediately.

AngeloKai:
s/jeffrey said he likely would get to the review in near term/jeffrey said he won't able to get the review in near term/

AngeloKai:
As a related note, Firefox has PRs in review to abort webauthn codes when the user switches tabs: https://bugzilla.mozilla.org/show_bug.cgi?id=1383799 and https://bugzilla.mozilla.org/show_bug.cgi?id=1409202. Edge kind of has an abort mechanism in place as well. This PR is just exposing the API to developers to let them abort.

jcjones:
I think this needs your re-review, @equalsJeffH, and then merge it if you're good with it.

index.bs
equalsJeffH: I'm not sure.  this will ostensibly get review from folks who are more familiar with specifying this new abort hotness and we will see.

equalsJeffH:
given @jcjones' approval, i suppose we can merge this -- though, there are IMV unresolved aspects of it such as @jyasskin's comment  https://github.com/w3c/webauthn/pull/544#discussion_r148402940  and these other ones:

https://github.com/w3c/webauthn/pull/544#discussion_r148877356
https://github.com/w3c/webauthn/pull/544#discussion_r148907952
https://github.com/w3c/webauthn/pull/544#discussion_r148908459

jcjones:
Thanks, I agree. It gets the functionality in here, and IMO the right wording for what should happen on aborts will be more clear when we actually try to do the thing. (Which is ... soon!)

Issue #139, opened 2016-07-06
imageURL privacy


AxelNennker:
Should the spec mention that retrieving the image from a site leaks information to that site?

Should the platform show the image only after user consent is given that a credential should be created if the imageURL is not a data URL?

Should the spec recommend that the image should be retrieved once and the image data be stored by the platform? What if the user updates the image at the RP?

Should the imageURL be constrained to secure-contexts?

rlin1:
In the original proposal, this URL needed to be converted to a data url before being sent to the authenticator.
In my opinion, updates (or "remote" URLs are ok when they refer to the same RP, i.e. ETLD+1)

AxelNennker:
I did not find a discussion here in the issues list. Was this discussed and consciously removed?
@rlin1 would you suggest text regarding this issue?

jcjones:
[TPAC 2016] Options: Restrict to same-origin as the credential; Specify the time of load so it won't correlate to user action; Limit to data-urls ...

leshi:
What if we just remove the icon altogether?  Save some space on the authenticator.  The selection UI will be okay -- there is still a username.

We'll gain privacy and lose complexity...

AxelNennker:
Please remind me, why isn't icon part of [PublicKeyCredentialUserEntity](https://w3c.github.io/webauthn/#sctn-user-credential-params)?

I think it is valuable to have icon images e.g. if the authenticator is implemented on a smart watch or smart phone.

My opinion:
The origins of the icon URL and the RP origin MUST match.
The icon URL must be secure.
Passing icon data to the Authenticator is optional.
The Conforming User Agent MAY re-scale the image data before passing it to the Authenticator.
The Conforming User Agent MUST NOT pass non-image data or unsupported image data to the Authenticator. If the Conforming User Agent passes any image data to the Authenticator then PNG MUST be supported.
If the icon data is passed to the Authenticator then the Conforming User Agent MUST download the icon data once at makeCredential-time and convert it to a "data:image/png;base64" before passing it to the authenticator. 
If the icon data is NOT passed to the Authenticator but the Conforming User Agent provides an UI that shows the image then the Conforming User Agent SHOULD download the icon data once at makeCredential-time and SHOULD not download it again to protect the privacy of the user.

AngeloKai:
@AxelNennker icon is part of PublicKeyCredentialUserEntity. This object inherits from the PublicKeyCredentialEntity.

AngeloKai:
[...]

PR #545, opened 2017-09-07
Address security and privacy issues witht the iconURL

AngeloKai:
Address issue https://github.com/w3c/webauthn/issues/139

***
[Preview](https://s3.amazonaws.com/pr-preview/AngeloKai/webauthn/iconURL.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/webauthn/14c2733...AngeloKai:5b4d94a.html)

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -91,6 +91,10 @@ spec: credential-management-1; urlPrefix: https://w3c.github.io/webappsec-creden
     for: CredentialsContainer
         type: method
             text: create(); url: dom-credentialscontainer-create
+
+spec: mixed-content; urlPrefix: https://w3c.github.io/webappsec-mixed-content/
+    type: dfn 
+        text: a priori authenticated
 </pre> <!-- class=anchors -->
 
 <pre class="link-defaults">
@@ -1189,7 +1193,7 @@ associated.
 
     :   <dfn>icon</dfn>
     ::  A [=URL serializer|serialized=] URL which resolves to an image associated with the entity. For example, this could be
-        a user's avatar or a [=[RP]=]'s logo.
+        a user's avatar or a [=[RP]=]'s logo. This URL MUST be an [=a priori authenticated URL=].
 </div>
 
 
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -91,6 +91,10 @@ spec: credential-management-1; urlPrefix: https://w3c.github.io/webappsec-creden
     for: CredentialsContainer
         type: method
             text: create(); url: dom-credentialscontainer-create
+
+spec: mixed-content; urlPrefix: https://w3c.github.io/webappsec-mixed-content/
```
equalsJeffH: mixed-content is recently at CR, should we refer to that?
https://www.w3.org/TR/mixed-content/

AngeloKai: I am not too familiar with the TPAC decisions. My read was that the three approaches are all options but there was no decision. Please correct me if I am wrong. The current spec doesn't have any restriction in this area. Adding the restriction of  "priori authenticated URL" is more or less an ad hoc approach anyway because otherwise the page should show mixed content warning. 

Can you please explain why there should be a corresponding CTAP change? I don't see a mention about converting URL before sending to CTAP in the spec today. 

I updated the PR to reference the CR draft.

nadalin:
@equalsJeffH are you ok to merge now ?

nadalin:
@jcjones @equalsJeffH  please review and this looks like it's ready to go

leshi:
See comments in #139

equalsJeffH:
this change is fine, in the context of having an image URL, which we do at this time.  HOwever, my comment https://github.com/w3c/webauthn/pull/545#pullrequestreview-62011105 remains relevant (perhaps it should be an issue), as well as issue #139.

index.bs
```diff
@@ -91,6 +91,10 @@ spec: credential-management-1; urlPrefix: https://w3c.github.io/webappsec-creden
     for: CredentialsContainer
         type: method
             text: create(); url: dom-credentialscontainer-create
+
+spec: mixed-content; urlPrefix: www.w3.org/TR/mixed-content/
```
jyasskin: Could you add the `https://` to the front of this `urlPrefix`?

jyasskin: Oh, I think you can remove this block entirely. You're using "a priori authenticated URL", which is exported from the spec.

Issue #456, opened 2017-05-08
An issue about setAttestationChallenge() in "android-key" attestation statement


caileb2017:
I'm testing about "android-key" attestation statement of FIDO2.0 and I got an issue.
The test code that I tried it is like the code below.
```
String challenge = "f9084f24ac8c33 ... 3c6ef635cdd75746516"; // **this data size is 103 bytes**
KeyGenParameterSpec.Builder builder = new KeyGenParameterSpec.Builder(name, KeyProperties.PURPOSE_SIGN )
   .setAttestationChallenge(challenge.getBytes())
   .setUserAuthenticationRequired(true)
   .setDigests(KeyProperties.DIGEST_SHA256, KeyProperties.DIGEST_SHA512)
   .setAlgorithmParameterSpec(new ECGenParameterSpec("secp256r1"));
mKeyGenerator.initialize(builder.build());
mKeyGenerator.generateKeyPair();
```

and it occures an error below.
```
java.security.ProviderException: Failed to generate attestation certificate chain
    at android.security.keystore.AndroidKeyStoreKeyPairGeneratorSpi.getAttestationChain(AndroidKeyStoreKeyPairGeneratorSpi.java:631)
    at android.security.keystore.AndroidKeyStoreKeyPairGeneratorSpi.createCertificateChain(AndroidKeyStoreKeyPairGeneratorSpi.java:519)
    at android.security.keystore.AndroidKeyStoreKeyPairGeneratorSpi.generateKeyPair(AndroidKeyStoreKeyPairGeneratorSpi.java:502)
    at java.security.KeyPairGenerator$Delegate.generateKeyPair(KeyPairGenerator.java:699)
    ...
Caused by: android.security.KeyStoreException: Invalid input length
    at android.security.KeyStore.getKeyStoreException(KeyStore.java:1097)
    at android.security.keystore.AndroidKeyStoreKeyPairGeneratorSpi.getAttestationChain(AndroidKeyStoreKeyPairGeneratorSpi.java:632)
    at android.security.keystore.AndroidKeyStoreKeyPairGeneratorSpi.createCertificateChain(AndroidKeyStoreKeyPairGeneratorSpi.java:519) 
    at android.security.keystore.AndroidKeyStoreKeyPairGeneratorSpi.generateKeyPair(AndroidKeyStoreKeyPairGeneratorSpi.java:502) 
    at java.security.KeyPairGenerator$Delegate.generateKeyPair(KeyPairGenerator.java:699) 
[...]

PR #546, opened 2017-09-08
Fix Android attestation

christiaanbrand:
Android attestation had a circular dependency on the public key: The authenticatorData has a public key that was originally intended to be stuck in the ChallengeData for generating a new keypair. When calling this function the public key isn't available to us yet. We have made a change to bring this in line with other attestation formats (ie. packed attestation).

***
[Preview](https://s3.amazonaws.com/pr-preview/christiaanbrand/webauthn/patch-1.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/webauthn/14c2733...christiaanbrand:201a816.html)

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -3193,7 +3193,7 @@ IANA "WebAuthn Attestation Statement Format Identifier" registry established by 
 - Specification Document: Section [[#tpm-attestation]] of this specification
     <br/><br/>
 - WebAuthn Attestation Statement Format Identifier: android-key
-- Description: Platform-provided authenticators based on s versions "N", and later, may provide this proprietary "hardware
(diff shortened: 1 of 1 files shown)


AngeloKai:
@christiaanbrand would this PR fix https://github.com/w3c/webauthn/issues/456?

christiaanbrand:
Yes

Kieun:
Since the parameter for setAttestationChallenge is changed from the concatenation of authenticatorData and clientDataHash to clientDataHash only, the verification procedure should be modified as well.
> The value of the attestationChallenge field is identical to the concatenation of authenticatorData and clientDataHash.

In stead,
> The value of the attestationChallenge field is identical to clientDataHash.

@christiaanbrand would you change the verification procedures?

PR #555, opened 2017-09-08
using descriptive names for authenticator selection criteria

balfanz:
Since we don't directly send the values of the AuthenticatorSelectionCriteria dictionary keys over the wire, it's ok to leave these values in a human-readable form.

***
[Preview](https://s3.amazonaws.com/pr-preview/w3c/webauthn/balfanz-patch-1.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/webauthn/14c2733...df905a3.html)


equalsJeffH:
Clarifying question: what you are implying here is that a client/platform CTAP2 implementation takes the passed-in boolean options and marshals them into place, in a platform-specific manner, for the subsequent authenticatorMakeCredential CTAP2 command, thus what they are named as dict members in webauthn does not matter at the platform+CTAP2 layers?

leshi:
yes

PR #558, opened 2017-09-08
Plumb User ID through

### LINKED ISSUE #556: user id should be returned in get()
When we use navigator.credentials.get() to obtain webauthn's authentication assertion, the caller should receive the user id associated with this credential. This is necessary if get() is called without an allowList, and optional if it is.

Proposal: add a field `userId` to the [`PublicKeyCredential`](https://github.com/w3c/webauthn/blob/14c2733ca6a4a9568e4c48fef1b870448818e811/index.bs#L460) dictionary.

christiaanbrand:
We need to plumb the custom user id that the RP gave the authenticator during MakeCredential back through to the RP when doing getAssertion.
Closes #556

***
[Preview](https://s3.amazonaws.com/pr-preview/christiaanbrand/webauthn/patch-2.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/webauthn/14c2733...christiaanbrand:e63537f.html)

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -459,7 +459,7 @@ that are returned to the caller when a new credential is created, or a new asser
     [SecureContext]
     interface PublicKeyCredential : Credential {
         [SameObject] readonly attribute ArrayBuffer              rawId;
-        [SameObject] readonly attribute AuthenticatorResponse    response;
+	[SameObject] readonly attribute AuthenticatorResponse    response;
         [SameObject] readonly attribute AuthenticationExtensions clientExtensionResults;
     };
 </xmp>
@@ -941,6 +941,9 @@ When this method is invoked, the user agent MUST execute the following algorithm
                 : {{AuthenticatorAssertionResponse/signature}}
                 :: A new {{ArrayBuffer}}, created using |global|'s [=%ArrayBuffer%=], containing the bytes of the returned
                     {{signature}}
+		: {{AuthenticatorAssertionResponse/userId}}
+                :: A new {{DOMString}} containing the user ID returned from the successful [=authenticatorGetAssertion=] operation, 
+			as defined in [[#op-get-assertion]].   
             :   {{PublicKeyCredential/clientExtensionResults}}
             ::  A new {{AuthenticationExtensions}} object containing the [=extension identifier=] → [=client extension output=]
                 entries created by running each extension's [=client extension processing=] algorithm to create the [=client
@@ -1058,7 +1061,8 @@ optionally evidence of [=user consent=] to a specific transaction.
     interface AuthenticatorAssertionResponse : AuthenticatorResponse {
         [SameObject] readonly attribute ArrayBuffer      authenticatorData;
         [SameObject] readonly attribute ArrayBuffer      signature;
-    };
+	[SameObject] readonly attribute DOMString	 userId;
+	};
 </pre>
 <div dfn-type="attribute" dfn-for="AuthenticatorAssertionResponse">
     :   {{AuthenticatorResponse/clientDataJSON}}
@@ -1721,6 +1725,7 @@ On successful completion, the authenticator returns to the user agent:
 - The identifier of the credential (credential ID) used to generate the [=assertion signature=].
 - The [=authenticator data=] used to generate the [=assertion signature=].
 - The [=assertion signature=].
+- The user id associated with the credential used to generate the [=assertion signature=].
 
 If the authenticator cannot find any credential corresponding to the specified [=[RP]=] that matches the specified criteria, it
 terminates the operation and returns an error.
--- a/index.bs
+++ b/index.bs
@@ -943,7 +943,7 @@ When this method is invoked, the user agent MUST execute the following algorithm
                     {{signature}}
                 : {{AuthenticatorAssertionResponse/userId}}
                 :: A new {{DOMString}} containing the user ID returned from the successful [=authenticatorGetAssertion=] operation, 
-                   as defined in [[#op-get-assertion]].   
+                    as defined in [[#op-get-assertion]].   
             :   {{PublicKeyCredential/clientExtensionResults}}
             ::  A new {{AuthenticationExtensions}} object containing the [=extension identifier=] → [=client extension output=]
                 entries created by running each extension's [=client extension processing=] algorithm to create the [=client
--- a/index.bs
+++ b/index.bs
@@ -459,7 +459,7 @@ that are returned to the caller when a new credential is created, or a new asser
     [SecureContext]
     interface PublicKeyCredential : Credential {
         [SameObject] readonly attribute ArrayBuffer              rawId;
-	[SameObject] readonly attribute AuthenticatorResponse    response;
+        [SameObject] readonly attribute AuthenticatorResponse    response;
         [SameObject] readonly attribute AuthenticationExtensions clientExtensionResults;
     };
 </xmp>
--- a/index.bs
+++ b/index.bs
@@ -1061,8 +1061,8 @@ optionally evidence of [=user consent=] to a specific transaction.
     interface AuthenticatorAssertionResponse : AuthenticatorResponse {
         [SameObject] readonly attribute ArrayBuffer      authenticatorData;
         [SameObject] readonly attribute ArrayBuffer      signature;
-        [SameObject] readonly attribute DOMString	 userId;
-        };
+        [SameObject] readonly attribute DOMString        userId;
+    };
 </pre>
 <div dfn-type="attribute" dfn-for="AuthenticatorAssertionResponse">
     :   {{AuthenticatorResponse/clientDataJSON}}
--- a/index.bs
+++ b/index.bs
@@ -943,7 +943,7 @@ When this method is invoked, the user agent MUST execute the following algorithm
                     {{signature}}
                 : {{AuthenticatorAssertionResponse/userId}}
                 :: A new {{DOMString}} containing the user ID returned from the successful [=authenticatorGetAssertion=] operation, 
-		   as defined in [[#op-get-assertion]].   
+	           as defined in [[#op-get-assertion]].   
             :   {{PublicKeyCredential/clientExtensionResults}}
             ::  A new {{AuthenticationExtensions}} object containing the [=extension identifier=] → [=client extension output=]
                 entries created by running each extension's [=client extension processing=] algorithm to create the [=client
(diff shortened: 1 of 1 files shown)


nadalin:
So the Privacy considerations section should be a statement about the issue of sharing PII with someone you were not expecting

jyasskin:
In response to the discussion today, I suggest adding a sentence to the Security/Privacy Considerations section along the lines of:
> The authenticator can be used to retrieve a [user ID](https://w3c.github.io/webauthn/#dom-makepublickeycredentialoptions-user) given a [credential ID](https://w3c.github.io/webauthn/#dom-publickeycredential-identifier-slot), so if the user ID is PII, the credential ID is also PII.

This might be wrong: it assumes that if I need to combine two pieces of information to get some PII, then each of those pieces are also PII individually.

wseltzer:
On 09/13/2017 01:35 PM, Anthony Nadalin wrote:
> So the Privacy considerations section should be a statement about the issue of sharing PII with someone you were not expecting

also the risk of sharing PII with someone who might not want to receive
it/expect it to be in that field.

jyasskin:
@wseltzer @nadalin, I suspect we shouldn't try to make this spec a tutorial on how to protect PII. We need to identify anything that might be PII by surprise, but then trust RPs to know what they need to know about protecting it.

leshi:
> Edit: The above was wrong: exporting the user ID with the assertionresponse allows someone with physical access to the authenticator to turn an RP ID into a user ID. That's the risk that needs to be highlighted in the privacy considerations section.

We only want the user id for the single factor use case (i.e., resident keys).  In such a use case, the attacker has to convince the authenticator that it's the legit user -- so know the right PIN or get the device to cough up the assertion.  Once the assertion is coughed up -- even if there is no user id -- the attacker can just give it to the RP and not only know the user id but actually see the user data...  So IMHO, this is not too much of a concern.  WDYT?

emlun:
Why is the user ID necessary for getAssertion? Even for the single factor use case, isn't it possible for the RP to identify the user from only the credential ID even with no allowCredentials? For example:

1. Setup: The RP has an internal table linking credential IDs to public keys and internal user IDs, and the user has previously registered a credential with the RP
2. The user initiates an authentication ritual (providing no additional info at this point)
3. The RP generates a challenge and sends a PublicKeyCredentialRequest (with no allowCredentials) to the client
4. The authenticator chooses a credential and generate an assertion
5. The RP receives the PublicKeyCredential with an AuthenticatorAssertionResponse containing a credential ID and a signature by that credential
6. The RP looks up the public key from its table using the credential ID and verifies the challenge signature
7. If (6) fails, the RP asks the user to try again with a different credential
8. If (6) succeeds, the RP looks up the user ID from its table using the credential ID and initiates an authenticated session for that user

Shouldn't that work?

christiaanbrand:
It's a bad idea to have some external system with limited context responsible for generating unique indices for a database. We really need to key off something we control, not the authenticator.

emlun:
That does sound like a very valid point.

jyasskin:
@leshi 
> We only want the user id for the single factor use case (i.e., resident keys). In such a use case, the attacker has to convince the authenticator that it's the legit user -- so know the right PIN or get the device to cough up the assertion. Once the assertion is coughed up -- even if there is no user id -- the attacker can just give it to the RP and not only know the user id but actually see the user data... So IMHO, this is not too much of a concern. WDYT?

I agree in the single-factor use case. I don't see anything in the patch that limits the privacy leak to the single-factor case. Am I missing something, or should we explicitly say that the authenticator shouldn't release the user ID when, e.g., it's been passed a credential ID?

christiaanbrand:
Sure. But only the actual RP will get it back, and if you have the credential ID (and can masquerade as the RP) you already have the corresponding account, right? Do we really need to limit this?

jyasskin:
@christiaanbrand Anyone with physical access to the authenticator can pretend to be any RP, and if they can also convince the authenticator that they're the user (trivial for second-factor authenticators), they can get the user ID for that RP out.

I'm not *certain* that's a leak we should worry about, but we should decide whether to worry about it, and we should probably mention it in the privacy considerations section anyway.

christiaanbrand:
Oh, you mean physical access. We have "displayName" that's even more challenging

jovasco:
Credential IDs are not guaranteed unique in any way. Unless I missed something in the specs, it is perfectly valid to store all data locally and return a single byte key index.

This has a lot of consequences that are not currently handled, especially in the RP assertion validation algorithm:
* a Credential ID is not guaranteed unique for a single user on a single RP. (Will happen if you buy two such authenticators and you register them as a pair on every account).
* You cannot refuse duplicate Credential IDs as this is both valid and, if such an authenticator becomes popular, likely to be relatively common (as all credential IDs will be in the same limited range). Refusing would put the user in a situation he cannot resolve, retrying the registration will just replace the current key entry (as specified) with likely the same index.
* You need to verify against the public key of every matching Credential ID for this user. If one matches, this is a valid assertion.

It also makes returning the User ID a necessity as you would have potentially thousands (millions) of accounts with this Credential ID for first factor scenarios.

In any case, the specs assume Credential ID collisions are at least unlikely. It might be easier to add a few requirements to promote that than try to handle all the issues.

I've raised an issue for this in the FIDO group. Should I raise one here too?

leshi:
Thanks, Johan!

I believe that on the last W3C call (https://www.w3.org/2017/09/20-webauthn-minutes.html), we have agreement to plumb userid through -- which will hopefully solve all the aforementioned problems.

Jeff is taking it on himself to review the PR meanwhile.

emlun:
@jovasco They are not _guaranteed_ unique, no, but they are required to be globally unique with a high probability (see #579). Of course this leaves the reader to judge exactly what qualifies as "high" probability. Your example with a single byte key index is technically valid, but only if the authenticator regards 0 as a "high probability". RPs will need to be prepared to handle some duplicate credential IDs, but not reasonably on the order of thousands or even hundreds.

However it should probably be mentioned in implementation considerations to try to keep credential ID lookup complexity low, to prevent DoS attacks by registering a billion accounts with the same credential ID. That vulnerability is neatly solved by also providing the user ID as suggested here.

[make-cred]: https://www.w3.org/TR/webauthn/#op-make-cred

akshayku:
CredentialID are supposed to be globally unique. Without that, these cannot be used as globally index in the database and that's how MS has implemented it being very privacy oriented where we are not using userid. You cannot have single byte key index. We should define a minimum length for credential ID length and its randomness if there is any confusion.

If RP's sees a duplicate during makeCredential time and are having a global index based on credentialID  they should reject the registration and start again.

emlun:
@akshayku Let's continue this discussion in #579.

akshayku:
@emlun Agreed. Thanks for opening the issue.

emlun:
On another note, wouldn't this make the millions of existing U2F devices permanently incompatible?

They are currently compatible via the fido-u2f attestation statement format, but they cannot store anything else in addition to the credential ID and thus cannot return the user ID back in `authenticatorGetAssertion`.

Unless I'm mistaken in the above, I think with the above in mind this is too much to ask of the authenticator - especially as @akshayku points out that the user ID is not necessary for identifying the user. If some RPs require the user ID to authenticate, they can either require it before initating the authentication ritual - most will probably do that anyway - or implement it as an extension.

akshayku:
@emlun: No, this will not make existing U2F devices incompatible. 

You are correct that existing U2F devices cannot store user ID so for U2F devices, `authenticatorGetAssertion` response will have this field set to null/zero bytes.

This is similar to scenario where AAGUID is not being present in `authenticatorMakeCredential` response for U2F devices. 

User ID is an hint for the RP's for lookup in their databases. User ID is not included in the signature itself. Its an extra information.

nadalin:
@akshayku so it looks like this is ready to merge so that we can also merge https://github.com/fido-alliance/fido-2-specs/pull/300

emlun:
@akshayku Ah yes, if it's optional then that's fine; I got the impression it would be a required output. But then again, if it's optional then RPs cannot in general rely on it being available. In that case I feel like it might be better to push this into an extension, and reserve the core for the parts that must always be present.

akshayku:
This will be required for FIDO 2 devices. For U2F, we have a compatibility story as we anyway has to convert U2F assertion response to FIDO 2 response. 

And this is not part of the signature so can't be part of the extension.

emlun:
Not in the authenticator data as an authenticator extension, indeed, but couldn't it be returned as a client extension?

Either way there's still the fact that it _won't_ be available from all authenticators, regardless of whether or not it's required for FIDO 2 devices. RPs will still need to handle it not being available, so making it "required" seems rather moot.

christiaanbrand:
Why can this not be *required* for all resident-credential FIDO2 authenticators?

Remember this whole deal is *only* about resident-credentials. If it's a U2F-type authenticator (even in FIDO2 land) none of this stuff is required. As Jeffrey mentioned yesterday, it should *explicitly* not be present if the keypair is being used as a second factor.

christiaanbrand:
Just for completeness. I approve this merge, with the understanding that Jeffrey will write up some language which says authenticators should never return userid (or any account info for that matter) when a signature was requested using a CredentialID [this means it's being used as a second factor].

nadalin:
@christiaanbrand or @jyasskin please merge

emlun:
@christiaanbrand

>Remember this whole deal is _only_ about resident-credentials. [...] it should explicitly not be present if the keypair is being used as a second factor.
>[...] authenticators should never return userid (or any account info for that matter) when a signature was requested using a CredentialID [this means it's being used as a second factor].

Ah, those pieces of context had gone right over my head. In that case it makes sense. Sorry for making a fuss.

index.bs
```diff
@@ -1721,6 +1725,7 @@ On successful completion, the authenticator returns to the user agent:
 - The identifier of the credential (credential ID) used to generate the [=assertion signature=].
 - The [=authenticator data=] used to generate the [=assertion signature=].
 - The [=assertion signature=].
+- The user id associated with the credential used to generate the [=assertion signature=].
```
selfissued: Whatever name is chosen, use it here too.

christiaanbrand:
TODO: Change names to not contain the word userid. It's confusing.

christiaanbrand:
How to folks feel about the term "user handle". I've already updated the PR.

akshayku:
Looks fine to me.

ve7jtb:
Fine with me.

PR #572, opened 2017-09-18
Clarify Safetynet attestation return value

### LINKED ISSUE #563: Safetynet attestation does not return byte array as a response.
Section 7.5 describes Android Safetynet attestation statement format. 
The "response" value is defined as a byte string (Major type2 ). However, Safetynet API returns String instead of byte array. 
See following getJwsResult() method.
[https://developers.google.com/android/reference/com/google/android/gms/safetynet/SafetyNetApi.AttestationResponse](url)
So, we need to change response type from a byte string (Major type 2) to a text string (Major type 3).

christiaanbrand:
Clarify the encoding of SafetyNet attestation as a UTF8 encoded string. Closes #563

***
[Preview](https://s3.amazonaws.com/pr-preview/christiaanbrand/webauthn/patch-4.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/webauthn/db1be80...christiaanbrand:ee912ee.html)

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2496,7 +2496,7 @@ and the identity of the calling application.
     :: The version number of Google Play Services responsible for providing the SafetyNet API.
 
     : response
-    :: The value returned by the above SafetyNet API. This value is a JWS [[RFC7515]] object (see
+    :: The UTF8 encoded result of the getJwsResult() call of the SafetyNet API. This value is a JWS [[RFC7515]] object (see
         [SafetyNet online documentation](https://developer.android.com/training/safetynet/index.html#compat-check-response))
         in Compact Serialization.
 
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -2496,7 +2496,7 @@ and the identity of the calling application.
     :: The version number of Google Play Services responsible for providing the SafetyNet API.
 
     : response
-    :: The value returned by the above SafetyNet API. This value is a JWS [[RFC7515]] object (see
+    :: The UTF8 encoded result of the getJwsResult() call of the SafetyNet API. This value is a JWS [[RFC7515]] object (see
```
jyasskin: Use `{{UTF-8 encoded}}` to link to https://encoding.spec.whatwg.org/#utf-8-encode.

PR #581, opened 2017-09-25
Fix #577 - CDDL for attStmtTemplate is ambiguous

### LINKED ISSUE #577: CDDL for attStmtType is confusing
When implementing the [CBOR-encoded Attestation Object](https://w3c.github.io/webauthn/#generating-an-attestation-object), the `attStmtTemplate` defines `attStmt` to be `bytes`:

```
attObj = {
            authData: bytes,
            $$attStmtType
         }

attStmtTemplate = (
                      fmt: text,
                      attStmt: bytes
                  )

; Every attestation statement format must have the above fields
attStmtTemplate .within $$attStmtType
```

However, all of the definitions of the `attStmtType` actually override the `attStmt` to be some other type, namely a **map**. E.g.:

```
  $$attStmtType //= (
                        fmt: "fido-u2f",
                        attStmt: u2fStmtFormat
                    )
  u2fStmtFormat = {
                      x5c: [ attestnCert: bytes, * (caCert: bytes) ],
                      sig: bytes
                  }
```

Defining it at the top-level as **bytes** is confusing, then. It's probably better to define the top-level as a blank map, so that it's clear what the type repeatedly is.

jcjones:
There are multiple definitions of `attStmtType`; the template defines it
to be `bytes`, while each concrete instance of the template defines it
as a map. This clarifies that it is always a map, since the ".within" control
operator for CDDL defines that the socket `$$attStmtType` to be the superset of
`attStmtTemplate`. [1]

[1] https://tools.ietf.org/html/draft-ietf-cbor-cddl-00#section-3.8.5

***
[Preview](https://s3.amazonaws.com/pr-preview/jcjones/webauthn/577-cddl_attStmt_type.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/webauthn/67e922c...jcjones:5630b47.html)

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1919,7 +1919,9 @@ attObj = {
 
 attStmtTemplate = (
                       fmt: text,
-                      attStmt: bytes
+                      attStmt: {
+                        ; Map is filled in by each concrete attStmtType
+                      }
                   )
 
 ; Every attestation statement format must have the above fields
--- a/index.bs
+++ b/index.bs
@@ -1919,9 +1919,7 @@ attObj = {
 
 attStmtTemplate = (
                       fmt: text,
-                      attStmt: {
-                        ; Map is filled in by each concrete attStmtType
-                      }
+                      attStmt: { * tstr => any } ; Map is filled in by each concrete attStmtType
                   )
 
 ; Every attestation statement format must have the above fields


index.bs
```diff
@@ -1919,7 +1919,9 @@ attObj = {
 
 attStmtTemplate = (
                       fmt: text,
-                      attStmt: bytes
+                      attStmt: {
```
jyasskin: I think [CDDL](https://tools.ietf.org/html/draft-ietf-cbor-cddl-00#page-18) says this should be 
```cddl
  attStmt: { * tstr => any }
```

It could be `* any=>any`, but all the formats I see use just strings as the keys, and using multiple types for keys makes canonicalization a bit more difficult, so I agree with restricting it to just `tstr`.

Issue #507, opened 2017-07-17
FIDO U2F Attestation Statement Format doesn't say what to do with Counter


jcjones:
Similar to #506 ... when creating a new credential, the U2F wire protocol doesn't provide back a Counter value. We should define how UAs should behave in this case.

The options we've discussed before are:

1) Encode this as all zeroes, or some other sentinel value to indicate that it is unset.

2) After generating a new credential, UAs must immediately trigger a U2F Sign operation with that credential using bogus data to obtain the Counter. This will require an additional Test of User Presence before the Create Credential flow completes.

(Firefox is picking Option 1 w/ all zeroes for now, and we'll update as this resolves)

agl:
(Could also be resolved by removing the signature count. See #453.)

AngeloKai:
This could be related to https://github.com/w3c/webauthn/issues/125

equalsJeffH:
having counter values is a given and are in fielded devices we need to support. we need to document RP counter usage in impl-cons.

agl:
I believe that the signature counter was a mistake because of its limited utility and privacy risks.

The FIDO spec hints that the signature counter may be used to identify duplicated keys, but provides no guidance about how they see that happening. If a token fails to write to flash and thus repeats a counter, should sites reject that token and require a reenrollment? How is the latter to be authenticated? It seems to me that the base-rate of false positives would completely swamp the _extremely_ rare true-positives and thus nobody would bother.

On the other hand, in order to only need a constant amount of storage, U2F tokens generally maintain a single, token-wide counter. Thus the counter is a privacy leak between sites and adds a requirement that tokens maintain state.

Current U2F devices emit this counter and that has to be taken into account, but in #453 I suggest repurposing it as a randomisation field. This gives tokens (that want to use it as such) a defense against side-channel attacks and would allow future tokens to fix this privacy leak.

nadalin:
@rlin1 to write up algorithm for RPs to use counters

rlin1:
see branch sign-counter-alg-507

rlin1:
See PR #539

rlin1:
has been addressed by merging PR #539 IMHO

nadalin:
@rlin1 Can we close this then with the reference to #539 ?

PR #597, opened 2017-09-28
Sign counter alg 507 alternative: optional sig counter

rlin1:
This is the one with MSB=1 flagging it is not supported. MSB=0 means it is suported and included in the other bits.

***
[Preview](https://s3.amazonaws.com/pr-preview/w3c/webauthn/sign-counter-alg-507-alternative.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/webauthn/089c10e...1a93e76.html)

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2221,6 +2221,9 @@ implementable by [=authenticators=] with limited resources (e.g., secure element
     and signing the result using the credential private key. It sets |alg| to the algorithm of the credential private key, and
     omits the other fields.
 
+    When connected through [[FIDO-CTAP]], the authenticator returns the attestation Statement as
+    response to authenticatorMakeCredential as required here.
+
 : Verification procedure
 :: Verify that the given attestation statement is valid CBOR conforming to the syntax defined above.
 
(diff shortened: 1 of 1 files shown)


selfissued:
This would replace #539

selfissued:
Fixes #507

equalsJeffH:
on 4-oct-2017 webauthn call, @rlin1 is asserting that in Sydney last week (fido meeting) folks want to go with this PR and we will close pr #539

equalsJeffH:
seems to need another merge-from-master

ve7jtb:
Rolf did two solutions.  One with the counter fixed at 0 for non support of the counter and one with the MSB set to 1 to show non support of the counter.  I thought people were wanting the 0 option in Sydney.   Both work for me.    However I think the 0 option one had more work done on it.  In this I don't see a verification step if the MSB is 1, so may not be complete.

jehrensvard:
I support Rolf's proposal.

It also gives us the option to use 0 as a "phantom" return value for MakeCredential calls with U2F devices. As we do not have a signature counter value for these, it seems sensible to return 0 then. Adding more potential use for the value 0 seems messy.

rlin1:
regarding https://github.com/w3c/webauthn/pull/597#issuecomment-334604391
see line 2179 in index.bs:
              <dt>greater or equal to 0x80000000 (i.e. has MSB set to 1).</dt>

PR #602, opened 2017-10-03
Name the parameters and variables in authenticatorMakeCredential.

jyasskin:
***
[Preview](https://s3.amazonaws.com/pr-preview/jyasskin/webauthn/name-authenticator-params.html#op-make-cred) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/webauthn/d89c503...jyasskin:29c00ca.html)

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1741,8 +1741,8 @@ When this operation is invoked, the authenticator must perform the following pro
         :  a global [=signature counter=]
         ::  Use the global [=signature counter=]'s actual value when generating 
               [=authenticator data=].
-        : a signature counter <i>per credential</i>
-        :: then initialize this counter with 0.
+        : a per credential [=signature counter=]
+        :: allocate the counter, associate it with the new credential, and initialize the counter value as zero.
     </dl>
 1. Let |attestationData| be the [=attestation data=] byte array including the |credentialId| and |publicKey|.
 1. Let |authenticatorData| be the byte array specified in [[#sec-authenticator-data]] including |attestationData| and any |processedExtensions|.
--- a/index.bs
+++ b/index.bs
@@ -1675,49 +1675,63 @@ session to exist at any particular time, or by providing more complicated sessio
 The following operations can be invoked by the client in an authenticator session.
 
 
-### The <dfn>authenticatorMakeCredential</dfn> operation ### {#op-make-cred}
+<h4 id="op-make-cred" algorithm>The <dfn>authenticatorMakeCredential</dfn> operation</h4>
 
 This operation must be invoked in an authenticator session which has no other operations in progress. It takes the following
 input parameters:
 
-- The caller's [=RP ID=], as <a href='#CreateCred-DetermineRpId'>determined</a> by the user agent and the client.
-- The [=hash of the serialized client data=], provided by the client.
-- The [=[RP]=]'s {{PublicKeyCredentialRpEntity}}.
-- The user account's {{PublicKeyCredentialUserEntity}}, containing the [=user handle=] given by the [=[RP]=].
-- A sequence of pairs of {{PublicKeyCredentialType}} and {{COSEAlgorithmIdentifier}} requested by the [=[RP]=].
-    This sequence is ordered from most preferred to least
-    preferred. The platform makes a best-effort to create the most preferred credential that it can.
-- An optional list of {{PublicKeyCredentialDescriptor}} objects provided by the [=[RP]=] with the intention that, if any of
+: |rpId|
+:: The caller's [=RP ID=], as <a href='#CreateCred-DetermineRpId'>determined</a> by the user agent and the client.
+: |hash|
+:: The [=hash of the serialized client data=], provided by the client.
+: |rpEntity|
+:: The [=[RP]=]'s {{PublicKeyCredentialRpEntity}}.
+: |userEntity|
+:: The user account's {{PublicKeyCredentialUserEntity}}, containing the [=user handle=] given by the [=[RP]=].
+: |credTypesAndPubKeyAlgs|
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -1649,54 +1649,69 @@ session to exist at any particular time, or by providing more complicated sessio
...
-    1. Generate an identifier for this credential, such that this identifier is globally unique with high probability across all
+    1. Let |publicKey|/|privateKey| be a new set of cryptographic keys using the combination of {{PublicKeyCredentialType}} and
+        cryptographic parameters represented by the first item in |credTypesAndPubKeyAlgs| that is supported by this authenticator.
+    1. Let |credentialId| be a new identifier for this credential that is globally unique with high probability across all
         credentials with the same type across all authenticators.
-    1. Associate the credential with the specified [=RP ID=] and the user's account identifier
-        {{MakePublicKeyCredentialOptions/user}}.{{PublicKeyCredentialUserEntity/id}}.
-    1. Delete any older credentials with the same [=RP ID=] and {{MakePublicKeyCredentialOptions/user}}.{{PublicKeyCredentialUserEntity/id}}
-        that are stored locally by the [=authenticator=].
+    1. Associate the |credentialId| and |privateKey| with |rpId| and the user's account identifier
```
jyasskin: We could model this association as a map, which would be clear about when a new credential has to overwrite an old one. Do we want to?

jcjones:
I really like this change. I want to re-review when 601 merges.

jyasskin:
@jcjones Ready. :)

index.bs
```diff
@@ -1653,54 +1653,69 @@ session to exist at any particular time, or by providing more complicated sessio
...
-    if it has its own output capability, or by the user agent otherwise. If the user denies consent, return an error code
+    if it has its own output capability, or by the user agent otherwise. This prompt can use information in |rpEntity| and
+    |userEntity|. If the user denies consent, return an error code
     equivalent to "{{NotAllowedError}}" and terminate the operation.
 1. Once user consent has been obtained, generate a new credential object:
-    1. Generate a set of cryptographic keys using the most preferred combination of {{PublicKeyCredentialType}} and cryptographic
-        parameters supported by this authenticator.
-    1. Generate an identifier for this credential, such that this identifier is globally unique with high probability across all
+    1. Let |publicKey|/|privateKey| be a new set of cryptographic keys using the combination of {{PublicKeyCredentialType}} and
+        cryptographic parameters represented by the first item in |credTypesAndPubKeyAlgs| that is supported by this authenticator.
```
equalsJeffH: s/item/[=item=]/ ?

index.bs
equalsJeffH: wrt modeling as map: i dunno, that might be interpreted as overly-prescriptive (by platform- & authnr-type folk).

wrt store more than the credential ID and private key:  is this not saying to store (at least) the 4-tuple:  ( |credentialID|, |privateKey|, |rpId|, |userEntity|.{{PublicKeyCredentialUserEntity/id}} ) ?

jyasskin: Yeah, which isn't enough to return the User Handle from https://w3c.github.io/webauthn/#op-get-assertion as demanded by #558.

jyasskin: 'k, picking one platform and one authenticator person arbitrarily: @leshi and @rlin1, am I going to break something terribly here if I say that the authenticator maintains a [map](https://infra.spec.whatwg.org/#maps) from (rpId, userEntity.id) to (credentialId, privateKey), which we'll then look things up from within getAssertion?

equalsJeffH: adding a person from yet another platform: @akshayku

equalsJeffH: seems to me the concern here ( as expressed by platform/browser folk in past) is "overly constrained/specified", rather than "break"  :)

jyasskin: After trying to model this as a map, it turns out that the map doesn't cover non-resident keys, and modeling those more precisely takes some work. So never mind about this question.

index.bs
```diff
@@ -1726,14 +1741,14 @@ When this operation is invoked, the authenticator must perform the following pro
         :  a global [=signature counter=]
         ::  Use the global [=signature counter=]'s actual value when generating 
               [=authenticator data=].
+        : a per credential [=signature counter=]
+        :: allocate the counter, associate it with the new credential, and initialize the counter value as zero.
```
jyasskin: @rlin1 This collided a bit with #539. Did I merge it reasonably?

index.bs
```diff
@@ -1675,49 +1675,64 @@ session to exist at any particular time, or by providing more complicated sessio
...
         credentials with the same type across all authenticators.
-    1. Associate the credential with the specified [=RP ID=] and [=user handle=].
-    1. Delete any older credentials with the same [=RP ID=] and [=user handle=] that are stored locally by the [=authenticator=].
+    1. Let |userHandle| be <code>|userEntity|.{{PublicKeyCredentialUserEntity/id}}</code>.
+    1. Associate the |credentialId| and |privateKey| with |rpId| and |userHandle|.
+    1. Delete any older credentials with the same |rpId| and |userHandle| that are stored locally by the [=authenticator=].
 1. If any error occurred while creating the new credential object, return an error code equivalent to "{{UnknownError}}" and
     terminate the operation.
-1. Process all the supported extensions requested by the client.
+1. Let |processedExtensions| be the result of [=authenticator extension processing=] for each supported extension identifier/input
```
jcjones: `extension identifier` should be `|extension identifier|` as it's a defined term.

index.bs
jcjones: Not to speak for Rolf, but it looks right to me.

equalsJeffH: this looks fine to me! we should merge this.

index.bs
equalsJeffH: is not the variable here `|extensions|` and "extension identifier" ought to be `[=extension identifier=]` ?

PR #611, opened 2017-10-04
Rewrite the "make an assertion" introduction to clarify how get() works.

### LINKED ISSUE #566: How should the browser handle CredentialMediationRequirement for public key credentials?
`CredentialRequestOptions` in webauthn extends `CredentialRequestOptions` from CredMan.  I'm reproducing it below for easy reference:

```
dictionary CredentialRequestOptions {
  CredentialMediationRequirement mediation = "optional";
};

enum CredentialMediationRequirement {
  "silent",
  "optional",
  "required"
};
```

I'm not sure how a WebAuthn implementation should handle the mediation parameter.  It is not referenced at all in the WebAuthn spec.  We should clarify this.

jyasskin:
Fixes #566.

I hope this makes things easier to understand.

***
[Preview](https://s3.amazonaws.com/pr-preview/jyasskin/webauthn/clarify-get-flow.html#getAssertion) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/webauthn/997e749...jyasskin:9ed6734.html)

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -98,6 +98,7 @@ spec: mixed-content; urlPrefix: www.w3.org/TR/mixed-content/
 </pre> <!-- class=anchors -->
 
 <pre class="link-defaults">
+spec:credential-management; type:dfn; text:credentials
 spec:html; type:dfn; for:environment settings object; text:global object
 spec:infra; type:dfn; text:list
 spec:url; type:dfn; text:domain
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -746,29 +747,28 @@ authorizing an authenticator.
...
+[[#op-get-assertion]] to sign a [RP]-provided challenge and other collected data into an assertion, which is used as a
+[=credential=]. The user may choose not to provide a credential even if one is present, for example to maintain privacy.
 
-Note: This algorithm is synchronous; the {{Promise}} resolution/rejection is handled by
-{{CredentialsContainer/get()|navigator.credentials.get()}}.
-
-This method accepts a single argument:
+The {{CredentialsContainer/get()}} implementation calls
+<code>PublicKeyCredential.{{PublicKeyCredential/[[CollectFromCredentialStore]]()}}</code> to collect any [=credentials=] that
+should be available without [=user mediation=] (roughly this specification's [=test of user presence=]), and if it doesn't find
```
equalsJeffH: s/roughly/roughly:/  or  /roughly,/

I would also mention user verif here:
...specification's [=test of user presence=] or [=user verification=]...

..but rather than that, I suggest just using [=authorization gesture=] instead of the more specific types of [=authorization gesture=].

index.bs
```diff
@@ -746,29 +747,28 @@ authorizing an authenticator.
...
+the user select a [=credential source=].
 
-<dl dfn-type="argument" dfn-for="PublicKeyCredential/[[DiscoverFromExternalSource]](options)">
-    :   <dfn>options</dfn>
-    ::  This argument is a {{CredentialRequestOptions}} object whose
-        <code>|options|.{{CredentialRequestOptions/publicKey}}</code> member contains a challenge and additional options as
-        described in [[#assertion-options]]. The selected authenticator signs the challenge along with other collected data in
-        order to produce an assertion. See [[#op-get-assertion]].
-</dl>
+Since this specification requires a [=test of user presence=] to create any [=credentials=], <code>PublicKeyCredential.<dfn
```
equalsJeffH: s/a [=test of user presence=]/an [=authorization gesture=]/

equalsJeffH: perhaps this parag ought to be a "Note:" ?

index.bs
```diff
@@ -746,29 +747,28 @@ authorizing an authenticator.
 </div>
 
 
-### Use an existing credential to make an assertion - PublicKeyCredential's `[[DiscoverFromExternalSource]](options)` method ### {#getAssertion}
+### Use an existing credential to make an assertion ### {#getAssertion}
```
equalsJeffH: I'm not sure about pulling the "PublicKeyCredential's `[[DiscoverFromExternalSource]](options)` method" text out of the section title. if we do that we need to update all the other section titles to adopt the same style (IMV).  I prefer having it in the title beause it yields hits in the ToC.

index.bs
```diff
@@ -746,29 +747,28 @@ authorizing an authenticator.
...
-existing [=public key credential=], with the user's consent. The script optionally specifies some criteria to indicate what
-credentials are acceptable to it. The user agent and/or platform locates credentials matching the specified criteria, and guides
-the user to pick one that the script will be allowed to use. The user may choose not to provide a credential even if one is
-present, for example to maintain privacy.
+[RPS] call <code><a idl for="CredentialsContainer" lt="get()">navigator.credentials.get({publicKey:..., ...})</a></code> to
+discover and use an existing [=public key credential=], with the user's consent. The script optionally specifies some criteria
+to indicate what [=credential sources=] are acceptable to it. The user agent and/or platform locates [=credential sources=]
+matching the specified criteria, and guides the user to pick one that the script will be allowed to use. It then uses
+[[#op-get-assertion]] to sign a [RP]-provided challenge and other collected data into an assertion, which is used as a
+[=credential=]. The user may choose not to provide a credential even if one is present, for example to maintain privacy.
```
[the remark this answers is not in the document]
jyasskin: Actually the user here.

index.bs
jyasskin: I do generally prefer short section titles, but I took the [[DiscoverFromExternalSource]] out of this section name for a different reason: this section discusses more than just that method. I'm inclined to make a sub-section if you really want the method name in the ToC?

index.bs
jyasskin: Formally, all three of these introductory paragraphs are non-normative. This one's the closest to normative, since it defines `PublicKeyCredential.[[CollectFromCredentialStore]]`, but that's also stated in file:///Users/jyasskin/src/webauthn/index.html#iface-pkcredential.

But we often have a couple introductory sentences or paragraphs at the top of a section to explain things before the normative algorithms, and we don't usually mark them. So, I'm inclined to not mark these either, but I will if you prefer it. I'd probably use a "<div class="note"> instead of the word "Note:".

index.bs
equalsJeffH: yeah, we have most all interface/dict/object/whathaveyou names in the toc (which I, at least, find useful) so i'd be inclined to have it in there one way or another. I spose this style of change would also apply to "[Create a new credential - PublicKeyCredential’s \[\[Create\]\](options) method](https://w3c.github.io/webauthn/#createCredential)"  since it also discusses "more than just that method" ([IIUC](https://en.wiktionary.org/wiki/IIUC)) ?

index.bs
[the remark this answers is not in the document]
equalsJeffH: hm. so what this last sentence..
>  The user may choose not to provide a credential even if one is present, for example to maintain privacy.

..is actually saying is that the user may decline the entire interaction, thus not "providing a credential", yes?

index.bs
equalsJeffH: nevermind I suppose.  it just seemed to me that this last parag is a detail-level nuance given it is wrt `[[CollectFromCredentialStore]]` which only especially-curious folks will care about.  i would not mark all three parags because then we'd have to go change the style of all intro parags... 

Also, I would s/[=test of user presence=]/[=authorization gesture=]/

index.bs
jyasskin: I've added the sub-section, but I can put it back in the "Use an existing credential" if you prefer. Note that "Create a new credential" starts with the definition of [[Create]], while ["Use an existing credential"](https://s3.amazonaws.com/pr-preview/jyasskin/webauthn/clarify-get-flow.html#getAssertion) starts with a description of `credentials.get()`.

index.bs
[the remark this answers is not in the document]
jyasskin: Yes, that's my understanding. Also note that this sentence was already present in the old version.

index.bs
```diff
@@ -746,29 +747,28 @@ authorizing an authenticator.
...
 
-Note: This algorithm is synchronous; the {{Promise}} resolution/rejection is handled by
-{{CredentialsContainer/get()|navigator.credentials.get()}}.
-
-This method accepts a single argument:
+The {{CredentialsContainer/get()}} implementation calls
```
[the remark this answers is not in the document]
jyasskin: Done.

index.bs
jyasskin: Sure.

index.bs
jyasskin: Done.

index.bs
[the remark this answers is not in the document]
equalsJeffH: understood -- I would clarify that last sentence that the user may decline the entire interaction.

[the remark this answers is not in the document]
jyasskin: Done.

PR #612, opened 2017-10-05
Tools: Add Vagrantfile for a VM with bikeshed installed

emlun:
These changes provide a definition of a reproducible VM with `bikeshed` installed. No changes are made to the spec.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
-bikeshed/
+/bikeshed/
+.vagrant/
 index.html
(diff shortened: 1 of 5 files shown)


emlun:
I've now connected my W3C and GitHub accounts, sorry about that!

jcjones:
Vagrant rather than Docker? Old school. 👍

emlun:
I just happen to have recently fiddled with Vagrant for other projects, so yeah, this kind of fell out as a byproduct of that. :)

Issue #578, opened 2017-09-20
Privacy Considerations should describe risks of storing userID/displayName in "second-factor" authenticators


jyasskin:
After discussions about #558, there's agreement that in cases where the authenticator doesn't itself authenticate the user (e.g. via PIN, fingerprint, etc), it's a bad idea to store PII in the authenticator. The PII appears to be all of [MakePublicKeyCredentialOptions.user](https://w3c.github.io/webauthn/#dom-makepublickeycredentialoptions-user) except perhaps for `id` if the RP is careful (see below).

The RP is required to pass all of these fields to `create()` operations, so we should advise authenticators to avoid returning the non-`id` ones unless the authenticator has evidence that the same person is retrieving the fields as stored them.

We should also give guidance to RPs about what to store in user.id to make it minimally identifying. @leshi suggested that the RP should aggregate the ID with a random nonce and encrypt that with a key secret to the RP, so that people with access to the authenticator get what looks like random data.

ve7jtb:
I recall Google did the encryption trick with their pairwise identifiers for openID 2, so it can work.

For privacy I agree that credentials without some sort of local pin (or other authenticated unlock) should not provide a display name.

equalsJeffH:
this seems related to issue #382

PR #616, opened 2017-10-06
Fix #609: Formally define User Handle

### LINKED ISSUE #609: technical & terminology issues with PR#558 Plumb User ID through
seems to me that PR#558 Plumb User ID through could have been reviewed more thoroughly prior to merging. I see a technical issue we arguably need to fix, and also feel that it further clouds our already fragmented/hazy/undisciplined terminology ( see also #358 and #462 ).

1.  technical issue with PR #558

please note here..
```
dictionary MakePublicKeyCredentialOptions {
    ...
    required PublicKeyCredentialUserEntity       user;
    ...
};

[...]

dictionary PublicKeyCredentialUserEntity : PublicKeyCredentialEntity {
    BufferSource   id;
    DOMString      displayName;
};
```
..that `user.id` is a BufferSource, here on the "input" side of things ie to [#createCrendential](https://w3c.github.io/webauthn/#createCredential), [authnrMakeCred()](https://w3c.github.io/webauthn/#op-make-cred) et al. 

However, on the "output" side of things (ie [#op-get-assertion](https://w3c.github.io/webauthn/#op-get-assertion) and [#getAssertion](https://w3c.github.io/webauthn/#getAssertion)), in the [AuthenticatorAssertionResponse interface](https://w3c.github.io/webauthn/#iface-authenticatorassertionresponse), it is declared as a `DOMString` -- it really should be an `ArrayBuffer` like the other returned values, no?
```
interface AuthenticatorAssertionResponse : AuthenticatorResponse {
    [...]
    [SameObject] readonly attribute DOMString        userId;
};
```

2.  overall confusing terminology

this is the terminology on the "input" side ie [#createCrendential](https://w3c.github.io/webauthn/#createCredential)l, [authnrMakeCred()](https://w3c.github.io/webauthn/#op-make-cred) et al:
```
dictionary MakePublicKeyCredentialOptions {
    ...
    required PublicKeyCredentialUserEntity       user;>     ...
};

[...]
dictionary PublicKeyCredentialUserEntity : PublicKeyCredentialEntity {
    BufferSource   id;
    DOMString      displayName;
};

id, of type BufferSource
  A unique identifier for the user account entity. This is a reference
  to an opaque byte array value specified by the Relying Party. The
  maximum size of this array is 64 bytes.
```
..so we have a "user.id" notion/term here, along with "user account".

In guts of [authnrMakeCred()](https://w3c.github.io/webauthn/#op-make-cred) we have..
```
   * Associate the credential with the specified RP ID and the user’s
     account identifier user.id.

   * Delete any older credentials with the same RP ID and user.id 
     that ...
```

..note use of "user's account identifier" and "user.id" here. 

the new term "user handle" only appears twice in the entire spec, both times on the "output side" of things, as noted below. 

On the "output" side, in [#getAssertion](https://w3c.github.io/webauthn/#getAssertion), there is:
```
    If any authenticator indicates success,

        Remove authenticator from issuedRequests.

        Let value be a new PublicKeyCredential associated with 
        global whose fields are:

        [...]>
        response

            A new AuthenticatorAssertionResponse object associated 
            with global whose fields are:

            [...]>
            userId

                A new DOMString containing the user handle returned
                from the successful authenticatorGetAssertion 
                operation, as defined in §5.2.2 The 
                authenticatorGetAssertion operation.
```

note the `userId` item that is part of `response`.  IMHO that contradicts the conclusion of the discussion in PR #558 where Christiaan agreed "[TODO: Change names to not contain the word userid. It's confusing](https://github.com/w3c/webauthn/pull/558#issuecomment-332736943)"

I would tend to want to name this "userHandle". (tho, then we have an impedance mismatch with the input side...)

And then the "user handle" prose term is used once more at the end of "[The authenticatorGetAssertion operation](https://w3c.github.io/webauthn/#op-get-assertion)"

emlun:
This makes the changes suggested in #609:

- Formally define User Handle
- Rename "user id" and similar terms to "user handle" everywhere
- Change name and type of `AuthenticatorAssertionResponse` field `DOMString userId` to `ArrayBuffer `BufferSource userHandle`
  - ~~Also change type of `PublicKeyCredentialUserEntity.id` from `BufferSource` to `ArrayBuffer` for consistency~~ _(reverted after review)_
- `PublicKeyCredentialUserEntity.id` is not renamed, but it is now referred to as the "user handle"

TODO:

- The term "user account" is not formally defined.
- [This part](https://github.com/w3c/webauthn/pull/558#issuecomment-331537953) from #558:
  >with the understanding that Jeffrey will write up some language which says authenticators should never return userid (or any account info for that matter) when a signature was requested using a CredentialID [this means it's being used as a second factor].

***
[Preview](https://s3.amazonaws.com/pr-preview/emlun/webauthn/issue-609.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/webauthn/089c10e...emlun:d38ad1b.html)

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -395,6 +395,12 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
 :: User consent means the user agrees with what they are being asked, i.e., it encompasses reading and understanding prompts.
     An [=authorization gesture=] is a [=ceremony=] component often employed to indicate [=user consent=].
 
+: <dfn>User Handle</dfn>
+:: The user handle is specified by a [=[RP]=] and is a unique identifier for a user account with that [=[RP]=]. A user handle is
+    an opaque byte array with a maximum size of 64 bytes.
+:: The user handle is not meant to be displayed to the user, but is used by the [=[RP]=] to control the number of credentials - an
+    authenticator will never contain more than one credential for a given [=[RP]=] under the same user handle.
+
 : <dfn>User Verification</dfn>
 :: The technical process by which an [=authenticator=] <em>locally authorizes</em> the invocation of the
     [=authenticatorMakeCredential=] and [=authenticatorGetAssertion=] operations. [=User verification=] may be instigated
@@ -945,9 +951,9 @@ When this method is invoked, the user agent MUST execute the following algorithm
                 : {{AuthenticatorAssertionResponse/signature}}
                 :: A new {{ArrayBuffer}}, created using |global|'s [=%ArrayBuffer%=], containing the bytes of the returned
                     {{signature}}
-                : {{AuthenticatorAssertionResponse/userId}}
-                :: A new {{DOMString}} containing the user handle returned from the successful [=authenticatorGetAssertion=] operation, 
-                    as defined in [[#op-get-assertion]].   
+                : {{AuthenticatorAssertionResponse/userHandle}}
+                :: A new {{ArrayBuffer}} containing the [=user handle=] returned from the successful [=authenticatorGetAssertion=]
+                    operation, as defined in [[#op-get-assertion]].
             :   {{PublicKeyCredential/clientExtensionResults}}
             ::  A new {{AuthenticationExtensions}} object containing the [=extension identifier=] → [=client extension output=]
                 entries created by running each extension's [=client extension processing=] algorithm to create the [=client
@@ -1065,7 +1071,7 @@ optionally evidence of [=user consent=] to a specific transaction.
     interface AuthenticatorAssertionResponse : AuthenticatorResponse {
         [SameObject] readonly attribute ArrayBuffer      authenticatorData;
         [SameObject] readonly attribute ArrayBuffer      signature;
-        [SameObject] readonly attribute DOMString        userId;
+        [SameObject] readonly attribute ArrayBuffer      userHandle;
     };
 </pre>
 <div dfn-type="attribute" dfn-for="AuthenticatorAssertionResponse">
@@ -1080,6 +1086,9 @@ optionally evidence of [=user consent=] to a specific transaction.
 
     :   <dfn>signature</dfn>
     ::  This attribute contains the raw signature returned from the authenticator. See [[#op-get-assertion]].
+
+    :   <dfn>userHandle</dfn>
+    ::  This attribute contains the [=user handle=] returned from the authenticator. See [[#op-get-assertion]].
 </div>
 
 ## Parameters for Credential Generation (dictionary <dfn dictionary>PublicKeyCredentialParameters</dfn>) ## {#credential-params}
@@ -1139,10 +1148,8 @@ optionally evidence of [=user consent=] to a specific transaction.
         Its value's {{PublicKeyCredentialUserEntity/displayName}} member is required, and contains a friendly name for the user
         account (e.g., "John P. Smith").
 
-        Its value's {{PublicKeyCredentialUserEntity/id}} member is required, and contains an identifier for the account, specified
-        by the [=[RP]=]. This is not meant to be displayed to the user, but is used by the [=[RP]=] to control the
-        number of credentials - an authenticator will never contain more than one credential for a given [=[RP]=] under the
-        same {{PublicKeyCredentialUserEntity/id}}.
+        Its value's {{PublicKeyCredentialUserEntity/id}} member is required and contains the [=user handle=] for the account,
+        specified by the [=[RP]=].
 
     :   <dfn>challenge</dfn>
     ::  This member contains a challenge intended to be used for generating the newly created credential's [=attestation
@@ -1219,16 +1226,14 @@ credential.
 
 <pre class="idl">
     dictionary PublicKeyCredentialUserEntity : PublicKeyCredentialEntity {
-        BufferSource   id;
+        ArrayBuffer    id;
         DOMString      displayName;
     };
 </pre>
 
 <div dfn-type="dict-member" dfn-for="PublicKeyCredentialUserEntity">
     :   <dfn>id</dfn>
-    ::  A unique identifier for the user account entity.
-        This is a reference to an opaque byte array value specified by the [=[RP]=].
-        The maximum size of this array is 64 bytes.
+    ::  The [=user handle=] of the user account entity.
 
     :   <dfn>displayName</dfn>
     ::  A friendly name for the user account (e.g., "John P. Smith").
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -395,6 +395,12 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
 :: User consent means the user agrees with what they are being asked, i.e., it encompasses reading and understanding prompts.
     An [=authorization gesture=] is a [=ceremony=] component often employed to indicate [=user consent=].
 
+: <dfn>User Handle</dfn>
+:: The user handle is specified by a [=[RP]=] and is a unique identifier for a user account with that [=[RP]=]. A user handle is
+    an opaque byte array with a maximum size of 64 bytes.
+:: The user handle is not meant to be displayed to the user, but is used by the [=[RP]=] to control the number of credentials - an
```
jyasskin: I think you're using this `::` to get a paragraph break, in which case use the following instead:

```
:: The user handle is specified by a [=[RP]=] and is a unique identifier for a user account with that [=[RP]=]. A user handle is
    an opaque byte array with a maximum size of 64 bytes.

    The user handle...
```

index.bs
```diff
@@ -945,9 +951,9 @@ When this method is invoked, the user agent MUST execute the following algorithm
                 : {{AuthenticatorAssertionResponse/signature}}
                 :: A new {{ArrayBuffer}}, created using |global|'s [=%ArrayBuffer%=], containing the bytes of the returned
                     {{signature}}
-                : {{AuthenticatorAssertionResponse/userId}}
-                :: A new {{DOMString}} containing the user handle returned from the successful [=authenticatorGetAssertion=] operation, 
-                    as defined in [[#op-get-assertion]].   
+                : {{AuthenticatorAssertionResponse/userHandle}}
+                :: A new {{ArrayBuffer}} containing the [=user handle=] returned from the successful [=authenticatorGetAssertion=]
```
jyasskin: Add `, created using |global|'s [=%ArrayBuffer%=],` like in the earlier items.

(I wish we had a more concise way to say this, but this is the state of the art at the moment.)

index.bs
```diff
@@ -1219,16 +1226,14 @@ credential.
 
 <pre class="idl">
     dictionary PublicKeyCredentialUserEntity : PublicKeyCredentialEntity {
-        BufferSource   id;
+        ArrayBuffer    id;
```
jyasskin: `BufferSource` is preferred here, according to the note in https://heycam.github.io/webidl/#idl-buffer-source-types:

> When designing APIs that take a buffer, it is recommended to use the BufferSource typedef rather than ArrayBuffer or any of the view types.

index.bs
```diff
@@ -1691,10 +1696,8 @@ When this operation is invoked, the authenticator must perform the following pro
         parameters supported by this authenticator.
     1. Generate an identifier for this credential, such that this identifier is globally unique with high probability across all
         credentials with the same type across all authenticators.
-    1. Associate the credential with the specified [=RP ID=] and the user's account identifier
-        {{MakePublicKeyCredentialOptions/user}}.{{PublicKeyCredentialUserEntity/id}}.
-    1. Delete any older credentials with the same [=RP ID=] and {{MakePublicKeyCredentialOptions/user}}.{{PublicKeyCredentialUserEntity/id}}
-        that are stored locally by the [=authenticator=].
+    1. Associate the credential with the specified [=RP ID=] and [=user handle=].
```
jyasskin: This will conflict with #602, but whoever merges second can just fix that.

index.bs
```diff
@@ -395,6 +395,12 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
 :: User consent means the user agrees with what they are being asked, i.e., it encompasses reading and understanding prompts.
     An [=authorization gesture=] is a [=ceremony=] component often employed to indicate [=user consent=].
 
+: <dfn>User Handle</dfn>
+:: The user handle is specified by a [=[RP]=] and is a unique identifier for a user account with that [=[RP]=]. A user handle is
+    an opaque byte array with a maximum size of 64 bytes.
```
jyasskin: One other thing: could you use [`[=byte sequence=]`](https://infra.spec.whatwg.org/#byte-sequence) instead of "byte array"?

index.bs
emlun: Good catch! Fixed.

index.bs
emlun: Fixed!

index.bs
emlun: Fixed!

index.bs
emlun: Absolutely, should we do that in [§5.1](https://w3c.github.io/webauthn/#sec-authenticator-data) as well?

jyasskin: Yes, but probably in a separate PR to avoid scope creep. :)

emlun: My thought as well. :)

index.bs
```diff
@@ -1139,10 +1149,8 @@ optionally evidence of [=user consent=] to a specific transaction.
         Its value's {{PublicKeyCredentialUserEntity/displayName}} member is required, and contains a friendly name for the user
         account (e.g., "John P. Smith").
 
-        Its value's {{PublicKeyCredentialUserEntity/id}} member is required, and contains an identifier for the account, specified
-        by the [=[RP]=]. This is not meant to be displayed to the user, but is used by the [=[RP]=] to control the
-        number of credentials - an authenticator will never contain more than one credential for a given [=[RP]=] under the
-        same {{PublicKeyCredentialUserEntity/id}}.
+        Its value's {{PublicKeyCredentialUserEntity/id}} member is required and contains the [=user handle=] for the account,
```
jcjones: What's the effect of this being just an empty `new Uint8Array()`?

jcjones:
Oops, should have asked out here:

What's the effect, now, of passing in just an empty BufferSource, like `id: new Uint8Array()` ? Is that still permitted, and browsers just echo back an empty buffer in the Response types?

jyasskin:
My reading of this change is that it doesn't affect behavior except for using array buffers instead of DOMStrings. So, we still don't prohibit 0-length handles.

jcjones:
Looking at the second TODO there, am I correct that  `AuthenticatorAssertionResponse.userHandle` is going to be an empty array for U2F, always, since U2F has no mechanism to pass this around?

emlun:
That's what the TODO is supposed to figure out:

>with the understanding that Jeffrey will write up some language which says authenticators should never return userid (or any account info for that matter) when a signature was requested using a CredentialID [this means it's being used as a second factor].

My gut says it should probably be `undefined` in that case.

jyasskin:
FWIW, I'd like to get this patch merged and then try for the text to fix #578. I didn't realize I'd signed up to write that text, and I don't expect to have time until at least Oct 23; maybe longer because we're starting to get back into C++ season.

nadalin:
@jyasskin go for it and merge it

PR #617, opened 2017-10-06
Remove [SameObject] from AuthenticatorAssertionResponse.userId

jcjones:
WebIDL does not allow `[SameObject]` to be used for non-interface types, like
`DOMString`.

The WebIDL spec describes its limitations [1] as:

> The `[SameObject]` extended attribute must not be used on anything other than a
> read only attribute whose type is an interface type or object.

`DOMString` is a primitive; it is neither an interface type nor an object [2],
so this patch removes the [SameObject] that was mistakenly annotated on it.

[1] https://heycam.github.io/webidl/#SameObject
[2] https://heycam.github.io/webidl/#idl-DOMString

***
[Preview](https://s3.amazonaws.com/pr-preview/jcjones/webauthn/sameobject-userid.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/webauthn/089c10e...jcjones:e11baf3.html)

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1065,7 +1065,7 @@ optionally evidence of [=user consent=] to a specific transaction.
     interface AuthenticatorAssertionResponse : AuthenticatorResponse {
         [SameObject] readonly attribute ArrayBuffer      authenticatorData;
         [SameObject] readonly attribute ArrayBuffer      signature;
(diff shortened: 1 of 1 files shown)


jcjones:
Note that #616 supercedes this.

jyasskin:
We should probably just take #616 instead.

PR #619, opened 2017-10-09
Fix #618 - Make PublicKeyCredential.isPlatformAuthenticatorAvailable static

### LINKED ISSUE #618: isPlatformAuthenticatorAvailable's examples imply it should be static
The example in the spec shows static usage of `isPlatformAuthenticatorAvailable`:

https://github.com/w3c/webauthn/blob/089c10e16bdcda0bd75db1d7f1d6a2be21304e08/index.bs#L3432-L3451

However, isPlatformAuthenticatorAvailable is defined [`[Unscopable]`](https://www.w3.org/2017/08/webidl-fpwd-draft.html#Unscopable), which is a [regular operation](https://www.w3.org/2017/08/webidl-fpwd-draft.html#dfn-regular-operation) which is defined as being **not** static.

I think the static method of using this is what we want, actually. Declaring this as a non-static method on `PublicKeyCredential` means users need to obtain a concrete `PublicKeyCredential` object on which to call `isPlatformAuthenticatorAvailable()`. This leads to a situation where you have to first complete a call to `navigator.credentials.create()` in order to call `isPlatformAuthenticatorAvailable()`.

Firefox's implementation is thus - until I hear otherwise - replacing `[Unscopable]` with `static` and I'll open a PR to do the same to the spec.

jcjones:
Please see issue #618 for details.

***
[Preview](https://s3.amazonaws.com/pr-preview/jcjones/webauthn/618-isPlatformAuthenticatorAvailable.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/webauthn/089c10e...jcjones:d0a010c.html)

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -998,7 +998,7 @@ but short enough that the dangling promise will still be resolved in a reasonabl
 <pre class="idl">
     [SecureContext]
     partial interface PublicKeyCredential {
-        [Unscopable] Promise < boolean > isPlatformAuthenticatorAvailable();
+        static Promise < boolean > isPlatformAuthenticatorAvailable();
     };
 </pre>
 


index.bs
```diff
@@ -998,7 +998,7 @@ but short enough that the dangling promise will still be resolved in a reasonabl
 <pre class="idl">
     [SecureContext]
     partial interface PublicKeyCredential {
-        [Unscopable] Promise < boolean > isPlatformAuthenticatorAvailable();
```
jyasskin: This [Unscopable] came from https://github.com/w3c/webauthn/pull/379#issuecomment-308037955, where I believe it was just a mistaken copy from a change to `Document`.

jcjones:
When we get a head-nod on this from @AngeloKai we've agreed to merge this.

PR #631, opened 2017-10-11
Fix #551 - Define a Store method for PublicKeyCredential.

### LINKED ISSUE #551: .store() is confusing
As defined, webauthn inherits the default implementation of .store() from CredMan.  Unfortunately, this means that .store() is a NOOP, but actually resolves.  It does not throw.  That means a caller will be confused in that they will call .store(), not see anything blow up, and not really realize that the .store() had no effect.

jcjones:
The Store operation isn't defined for PublicKeyCredential, even though it
is inherited from Credential Management. This defines that operation as
always resolving with an error.

This will resolve #551 .

***
[Preview](https://s3.amazonaws.com/pr-preview/jcjones/webauthn/551-store.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/webauthn/f6c9ed6...jcjones:b7613fa.html)

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -985,6 +985,26 @@ During the above process, the user agent SHOULD show some UI to the user to guid
 authorizing an authenticator with which to complete the operation.
 </div>
 
+### Store an existing credential - PublicKeyCredential's `[[Store]](credential)` method ### {#storeCredential}
+
+<div link-for-hint="PublicKeyCredential/[[Store]](credential)">
+
+The <dfn for="PublicKeyCredential" method>\[[Store]](credential)</dfn> method is not supported
+for Web Authentication's {{PublicKeyCredential}} type, so it always returns an error.
+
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -972,6 +972,26 @@ During the above process, the user agent SHOULD show some UI to the user to guid
...
 
+### Store an existing credential - PublicKeyCredential's `[[Store]](credential)` method ### {#storeCredential}
+
+<div link-for-hint="PublicKeyCredential/[[Store]](credential)">
+
+The <dfn for="PublicKeyCredential" method>\[[Store]](credential)</dfn> method is not supported
+for Web Authentication's {{PublicKeyCredential}} type, so it always returns an error.
+
+Note: This algorithm is synchronous; the {{Promise}} resolution/rejection is handled by
+{{CredentialsContainer/create()|navigator.credentials.create()}}.
```
AngeloKai: You need to point it to store instead of create

Issue #628, opened 2017-10-11
Allow RPs to choose between "required" and "optional" attestation in credentials.create()


balfanz:
TL;DR: When/if a client platform uses the Privacy CA model described in the spec, it would be beneficial for the RP to specify if they want an attestation from the Privacy CA. We should introduce an (optional) parameter in credentials.create() for this.

Details:

Some platforms/clients may opt into using Privacy CAs to address potential issues with poorly-implemented on-Authenticator attestation (e.g., small batch sizes). This does not change the syntax of the attestation, and is transparent to the RP.

What it *does* mean, however, is the following:

1. The platform may prompt the user about information being sent to the Privacy CA (since that information potentially reveals aspects about the Authenticator that the user is using)

2. Obtaining information from the Privacy CA introduces latency.

3. The Privacy CA learns certain aspects about the login to the RP, such as timing, and Authenticator used.

The RP might not wish to be exposed to these drawbacks, and not want to use a Privacy CA. In that case, a platform that normally uses a Privacy CA could instead use "self attestation" (i.e., replace the Authenticator's attestation data with client-generated self-attested attestation data). It's faster, and has no privacy concerns. 

We should give the RP an option to select between these two modes when calling credentials.create(). In one mode, the RP requests a "full" attestation, but acknowledges extra latency, potential user drop-off due to users not wanting to reveal information about their Authenticators to the Privacy CA and/or the RP, and potential leakage of some information to the Privacy CA. In the other mode, the RP signals to the client platform that they're not interested in attestation, and that the client platform should feel free to replace Authenticator attestations with a "self-attested" attestation data. (**TODO**: or should we just drop the attestation data in this case?)

Proposal:

Let's add an optional parameter `attestation` to `credentials.create`, with possible values `optional` and `required` (default being `optional`).

When the platform sees the `optional` value, it doesn't have to contact the Privacy CA, and can do whatever it sees fit to protect the user's privacy. When the value is set to `required`, then the client should ensure that the Authenticator's attestation is used to create a meaningful attestation for the RP, even if that means extra latency to contact the Privacy CA, obtaining extra consent from the user, etc.

ve7jtb:
We should stop calling this a privacy CA.  It is a attestation proxy not a privacy CA.  Yes it is using a certificate to sign but it returns a basic attestation signed by the proxy.

We also need a way to turn off this functionality so that the actual attestation can be delivered for some use cases.   I understand that for general social web this is probably fine, but not for all enterprise use cases.

[...]

PR #636, opened 2017-10-12
Adding a choice for RP to express preferences for attestation types

balfanz:
Adding a choice for RP to express preferences for attestation types. "low-cost" means don't bother the user to get the attestation, don't do a network roundtrip, etc., even if that means that we won't end up with a meaningful attestation (i.e., even if we only get self-attestation). "verifiable" means that the RP wants a verifiable attestation, even if that means that the user agent needs to connect to the Privacy CA, collect user consent, wait for ECDAA signatures to finish, etc.

***
[Preview](https://s3.amazonaws.com/pr-preview/w3c/webauthn/balfanz-patch-4.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/webauthn/e09e0c3...825cce7.html)


balfanz:
This addresses #628

nadalin:
@balfanz How does this help as there may be many attestation type that fit these categories and how do you know which one to choose ? I suggest that we don't get into this until next version when we add authenticator selection

index.bs
```diff
@@ -1307,6 +1313,29 @@ example of the latter, when the user is accessing the [=[RP]=] from a given clie
...
+    enum AttestationPreference {
+        "low-cost",
+        "verifiable" 
+    };
+</pre>
+
+<ul>
+    <li><dfn>low-cost</dfn> - indicates that the [=[RP]=] prefers a low-cost attestation type. For example, if a user agent
+      supports self attestion and Privacy CA, the user agent SHOULD use self attestation, avoiding calls to the Privacy
+      CA, user consent UI, etc. This is the default value.
```
emlun: I do not think this should be the default. I think we should encourage RPs to be secure by default and insecure in well thought out extraordinary situations.

akshayku:
Microsoft believes strongly that RPs MUST know the details of the device that holds a key including manufacturer, model and if relevant, a firmware version. 

Just recently a vulnerability in TPMs built by one of the main TPM manufacturers was uncovered (https://www.infineon.com/cms/en/product/promopages/rsa-update/). Cloud and internal databases have to be purged and x509 certificates revoked in response to this compromise. Attestation that specifies device details is a great data to identify weak or compromised keys. 

One could argue that a “privacy CA” could take on revocation responsibilities as well, but we know from experience that such revocation models don’t work. PKI support certificate revocation lists and OCSP responses, yet every modern browser has introduced an ability to revoke a specific TLS server certificate in a case of a compromise. This is not only because those revocation mechanism were not designed with Internet-speed security issues in mind, but because browser vendors don’t want to be at the mercy of CA operators that have reliability issues in their deployments, poor security response practices, etc. 

In conclusion, we believe that real mass security compromises are more important to address than perceived privacy problems such as mall batch device attestation. We can’t support a standard that doesn’t allow to know what type of device holds private keys we trust. Note that we do believe in privacy protection that prevent RPs from tracking specific devices so things like serial numbers or MAC addresses never need to be part of the attestation. 

Having said that, we understand other point of view and trade offs. We think, its RP's choice what behavior it wants. We need an ability for RP's to opt out of privacy CA and return the current format. The current wording of the PR does not reflect that. RP's `MUST` have an option to opt out explicitly of privacy CA if it chooses to.

agl:
The Infineon bug is worth thinking about in order to think about why RPs might want to look at attestation data retrospectively.

The overall goals, at least from my point of view, are to solve phishing, password reuse, and social engineering. If some vulnerability is found that compromises those goals, then RPs would want to address that.

#### Generation or wrapping problems

The Infineon bug, for one, is a case where attestation data is uninteresting. It's possible to identify a weak key from the public key and looking at attestation information would just introduce an extra step and false positive / negative risks.

The Debian weak-keys issue was the same.

In fact, if there's a problem of the type where it's feasible to work out a private key from the public key or key handle, RPs can always just try the attack against their key database, which gives more reliable results than asking every manufacturer for a breakdown of the impact on their devices. (Which may not be neatly segmented into different attestation certificates anyway.)

We can think about the case where the attack is cheap enough to be problematic, but not cheap enough to be attempted en-mass, and without any cheap distinguishing function. However, for ECC keys, I'm not sure that's actually possible. I think any such issues would have a sub-linear cost for checking _n_ keys. It's an interesting thought exercise if anyone can come up with one.

Perhaps there could be a problem where doing a single instance is still considered infeasible for the RP, but feasible for the attacker, but that seems to be the only niche left.

#### Local attacks

One can also think about vulnerabilities that require direct USB access, i.e. malware running on a machine with a token connected is able to extract keys. It seems plausible that some tokens could have such bugs.

But, while not ideal, it doesn't violate any of the initial goals. Users with malware installed already suffer from a wide range of issues as the malware can try to move horizontally and compromise the browser.

#### Remediation

While retrospective analysis of attestation information seems to be of minor utility, it's also unclear that RPs are the best place for dealing with the issue. Some RPs will care and, for those, it appears very likely that attestation information wouldn't be very useful in their response. But if we believe that some token is so hopeless that users need to retire and replace it, would we rely on users happening to interact with such an RP? It seems that we would probably need to update clients to alert users in order to alert a reasonable fraction of affected users.

If there is an argument that I've not thought of for why attestation would be especially useful in responses, it's not difficult to imagine an attestation proxy inserting data in a certificate that can be used for retrospective unblinding, should the proxy operator accept that was necessary. (Although perhaps that falls under the topic of &ldquo;revocation&rdquo;, which I agree has not worked well in the X.509 PKI, although I think the setting is quite different here.)

### On the other hand

There are more reasons for being cautious with attestation than just privacy: although a privacy scare could significantly hamper this effort overall.

Firstly, if many RPs all come up with their own policies for which tokens to accept and which to reject, we risk fragmenting the user experience. If it becomes common-place to need multiple tokens in order to reasonable coverage of RPs, users will have to start remembering which token works with which site. The reputation of U2F/webauthn tokens in general will suffer from the poor user experience.

Secondly, there is risk that some early token manufacturers will be positioned to extract rents by virtue of the distribution of their attestation roots. New entrants could be frustrated by the infeasibility of getting tens of thousands of RPs to trust their new devices. We have seen that happen in the WebPKI.

The FIDO MDS is hoped to solve a lot of the above issues, and perhaps it will. Although last time I checked it was significantly incomplete. However, if we let the ecosystem develop with direct attestation then fixing those issues later will be exceptionally difficult. In short: this has the hallmarks of an irreversible decision, which, for an ecosystem as an important as we all hope this one will be, should provoke significant caution.

Thus, while there are certainly cases around enterprise uses where direct attestation is sensible, it seems wise to move more slowly in the general case.

nadalin:
@balfanz I recommend that we close this with no action, let the RP deal how they want

akshayku:
In enterprise space, details about credential matters a lot including can the key be exported or not, therefore we need to know software vs hardware and react to issues like broken keys. If a customer needs to comply with some legal requirement to keep a key in hardware, we have to know details about authenticator. If a customer needs to have effective revocation story, we have to know details about it. Alerting users to throw away tokens has even less chance of working as asking them to keep their software up to date. FIDO metadata service is supposed to have information about FIDO tokens and manage them. If it is missing something, lets make it better. And FIDO should be THE ROOT that gives attestation certificate based on devices which passes certification process in FIDO. 

We also need to see full proposal of how privacy CA model works so that we can figure out how we can implement that in our platform if Google RP is being used in our Edge platform. Questions like who is standing up these privacy CA's? Does each platform supposed to stand up their own privacy CA or does FIDO will have a centralized privacy CA? Will platforms be able to call each other's CA's if each one is standing up their own. How about less powerful platforms in that case? Does RP has a say which privacy CA's they trust and can they influence that from the WebAuthN API level? What are the SLA for these CA's on revocation? What is the vulnerability disclosure process?  There are many questions which we don't have clear answer to. 

From the RP's perspective, we want to enable the same experience across all the platforms. 
From the platform's perspective, we want to enable same experience for all the RP's. 

All this gives me a feeling that we are lacking a solid proposal as of now and it needs much more thought. Anyway, if we believe we need this right now, I propose the following to enable the privacy CA path for the platforms as well as authenticators where we can discuss its characteristics at some later stage and also support both the deployment scenarios.

WebAuthN Credential Create needs to have the optional privacy enum option `PrivacyPreference` with following semantics: 

- `None`: Platform decides the default behavior

- `Require`: Platform must support privacy CA. If platform does not support privacy CA, makeCredential will fail. Platform needs to pass an additional option to CTAP authenticatorMakeCredential (lets call it as a Boolean option `enablePrivacy` with default behavior false) to enable privacy CA from the authenticator(for example, hide AAGUID in the attestation data by filling it with all zeros)

- `Disable`: Platform will not apply privacy CA semantics to this create credential call. Platform will not send `enablePrivacy` to the authenticator and authenticator will not hide any information like AAGUID for example. Platform will return authenticator Attestation information to RP.

agl:
At the webauthn level, we are requesting something with a slightly different direction than your `PrivacyPreference`: we feel that what would be helpful is if an RP can indicate whether it cares about attestation at all. Because, if not, clients may be able to avoid doing work, whether that involves a privacy CA, ECDAA signatures etc.

The experience with U2F, so far, has been that the vast majority of sites do not care about attestation (measured by whether it's possible to register a token with a dummy attestation certificate). For the current set of RPs, that may change because they're large entities. But, numerically, the majority of RPs (if we are successful) will be small Django, Wordpress, node.js, etc sites that are unlikely to want to maintain the infrastructure needed for attestation, nor to care. Thus it seems that not worrying about attestation should be the default, as setting a flag at registration time is trivial for sites that do care, but adding a flag as an optimisation is the sort of thing that smaller deployment may miss.

I hope that helps explain the thinking behind this pull request.

emlun:
I see, that makes sense. On the other hand, how many RPs are likely to directly implement WebAuthn themselves as opposed to using a library for generating and verifying requests? It would be equally trivial to add a flag to ignore attestation, and it would be hard to miss if the library by default has an empty trust store and therefore rejects all attestations. Compare with how `curl` by default rejects untrusted server certificates but can be configured to trust additional certificates or skip certificate validation.

emlun:
While we're at it, why not (optionally) provide the client with more precise information about attestation preferences? The RP could send a list naming the attestation types it's willing to accept, for example `attestation: ['self', 'privacy-ca', 'none']`. The client can then choose which type to return, depending on what the authenticator supports of course. If this list contains `none`, that's equivalent to the suggested `low-cost` value meaning the RP won't care if the response doesn't even have an attestation statement. `verifiable` corresponds to this list containing at least one element and not containing `self` or `none`.

I suppose one argument against it is that it would be easier to inadvertently add `self` or `none` to that list - because "hey, let's accept them all!" or something - than to inadvertently set `attestation: 'low-cost'`.

akshayku:
@agl Actually I am more confused now :)

The conversation started with Privacy CA and hiding user and device information from RP's and Privacy CA. Now it looks like it's about attestation itself. Or is it both with a open ended wording for `verifiable`? 

If it is only about attestation being passed or not, then it is a simpler conversation.
If it is open ended, where platform decides in future that it wants to hide that information, then we need a way out for our RP's as we need device information for our customers.
If it is not open ended and we are not doing privacy CA thing right now, we need to clarify that in the text. 

Privacy CA is a much bigger topic and I don't think we are ready yet with a solid proposal and we propose it to be addressed when we have a common understanding of what it means for all the parties (RP's/Platforms/Authenticators) involved in next version.

To remove confusion, I like @emlun suggestion with following caveats
- List should be ordered list of attestation types what RP prefers.
- The values can ONLY be one of the attestation format types defined in the spec.
- We introduce a new attestation type named `privacy-ca` to enable Privacy CA.
- We introduce a new attestation type named `none` which basically means no attestation.
- We introduce a new attestation type named `self` which basically means attestation by newly created credential.

So the whole options becomes based on today's attestation statement format type becomes
- `packed` : Meaning whole device information getting send to RP
- `tpm`
- `android-key`
- `android-safetynet`
- `fido-u2f`
- `privacy-ca`: NEW 
- `self`: NEW
- `none`: NEW

Now the default, we believe that giving more information by default is a good thing for the ecosystem. If someone does not care, it can pass `none` or he can ignore the attestation statement when it receives so there is no issue of RP forgetting to setup these flags.

emlun:
@akshayku It looks like you're confusing _attestation statement formats_ ("signing/verification procedures") and  _attestation types_ ("trust models"). The two are distinct: `packed`, `tpm`, `andoid-key`, `android-safetynet` and `fido-u2f` are the currently defined attestation statement formats and "Self Attestation", "Basic Attestation", "Privacy CA" and "ECDAA" are the currently defined attestation types. Each attestation statement format can convey some nonempty subset of the attestation types. I'm not suggesting any new attestation statement formats or attestation types, the only thing new I suggest is `none` as a pseudo-attestation type used only in this preference parameter to tell the client that any attestation statement will be completely ignored.

index.bs
```diff
@@ -1307,6 +1313,29 @@ example of the latter, when the user is accessing the [=[RP]=] from a given clie
...
 
+### Attestation Preference enumeration (enum <dfn enum>AttestationPreference</dfn>) ### {#attestation}
+
+[=[RPS]=] may use the {{AttestationPreference}} enum to specify their preference regarding attestation types used 
+during credential generation.
+
+<pre class="idl">
+    enum AttestationPreference {
+        "low-cost",
+        "verifiable"
```
equalsJeffH: needs another enum member of "direct"

index.bs
```diff
@@ -1307,6 +1313,29 @@ example of the latter, when the user is accessing the [=[RP]=] from a given clie
...
+</pre>
+
+<ul>
+    <li><dfn>low-cost</dfn> - indicates that the [=[RP]=] prefers a low-cost attestation type. For example, if a user agent
+      supports self attestion and Privacy CA, the user agent SHOULD use self attestation, avoiding calls to the Privacy
+      CA, user consent UI, etc. This is the default value.
+    <li><dfn>verifiable</dfn> - indicates that the [=[RP]=] prefers an attestation type that leads to verifiable attestation
+      statements, even if that incurs extra cost such as network roundtrips and/or collection of user consent. For example, 
+      if a user agent supports the self attestation and Privacy CA attestation types, the user agent SHOULD use the Privacy CA
+      attestation type.
```
equalsJeffH: need to describe "direct" attestation enum type as one wehre the client conveys the unadulterated attestation as generated by the authenticator to the RP.

index.bs
```diff
@@ -1307,6 +1313,29 @@ example of the latter, when the user is accessing the [=[RP]=] from a given clie
...
+<pre class="idl">
+    enum AttestationPreference {
+        "low-cost",
+        "verifiable" 
+    };
+</pre>
+
+<ul>
+    <li><dfn>low-cost</dfn> - indicates that the [=[RP]=] prefers a low-cost attestation type. For example, if a user agent
+      supports self attestion and Privacy CA, the user agent SHOULD use self attestation, avoiding calls to the Privacy
```
rlin1: a) it is not the user agent supporting attestation but the authenticator.
b) we should use more general language, i.e.:
if an authenticator supports self attestation and some ther attestation type, the authenticator SHOULD use self attestation.

rlin1:
Note we have 
a) different attestation formats and
b) different attestation types (e.g. self attestation, basic attestation, Privacy-CA, ECDAA)

rlin1:
Proposal: 3 different vaues:
low-cost (i.e. none)
Privacy-CA
verifiable attestation as generated by the authenticator

jcjones:
Straw-man proposal for the 3rd enum definition:

```
<dfn>direct</dfn> - indicates that the [=[RP]=] prefers to receive the attestation directly
from the authenticator; if that is not possible, the operation should not proceed.
```

equalsJeffH:
> indicates that the [=[RP]=] prefers to receive the attestation directly
from the authenticator; if that is not possible

Why would it not be possible?

jcjones:
Speaking just for Firefox ecosystem, it wouldn't be possible for Tor Browser, and potentially not in something like Private Browsing mode.

akshayku:
Your use cases are valid, but the proposed language allows a browser to reject direct attestation always and that is against the spirit of the compromise that we've discussed at TPAC2017. To find another compromise for the situations that you bring up, can we change the language as follows:

<dfn>direct</dfn> - indicates that the [=[RP]=] requires to receive the attestation directly
from the authenticator. In this case, the Platform MUST provide non-modified attestation statement from the authenticator. Platform MAY ask for user consent in privacy sensitive modes of operation such as Incognito or Private browsing,

nadalin:
We should not be selecting specific attestation types, but rather the class of attestation as attestation types will change, this also brings up the issue of the RP being able to distinguish the attestation type received and puts a burden on the authenticator to have to honor the requested type. I don't think this is fully thought out yet.

akshayku:
Corresponding CTAP PR: https://github.com/fido-alliance/fido-2-specs/pull/388

equalsJeffH:
@rlin1 wrote:
> Proposal: 3 different vaues:
low-cost (i.e. none)
Privacy-CA
verifiable attestation as generated by the authenticator.

@akshayku wrote:
> direct - indicates that the [=[RP]=] requires to receive the attestation directly
from the authenticator. In this case, the Platform MUST provide non-modified attestation statement from the authenticator. Platform MAY ask for user consent in privacy sensitive modes of operation such as Incognito or Private browsing,

combining the above and adding some polishing yields:

none:  [ some form of self-attestation ]
proxied: [ google privacy CA proposal ]
direct: the client platform MUST provide the [=attestation object=], as generated by the [=authenticator=], to the [=[RP]=], unless operating in a privacy sensitive operational mode such as Incognito or Private browsing. In the latter case, the client platform MAY prompt for [=user consent=], and substitute none or proxied if the user denies consent. Whether none or proxied may be established by local user configuration setting, or by enterprise policy. 

the above obviously needs further polishing and thought.

@rolf noted:
> We still have to decide whether low-cost means self attestation (generated by Authnr) or some public key generated by the user agent.

yes.

@nadalin wrote:
> I don't think this is fully thought out yet.

agreed.

emlun:
Perhaps "proxied" should be called "minimal" or some such, to further decouple it from the implementation?

index.bs
balfanz: Done.

index.bs
balfanz: Rewrote this. Can you take another look?

index.bs
balfanz: Users will be plenty secure with software-only authenticators (i.e., those that don't have meaningful attestations). Also, the other cases likely mean that the client will prompt the user with consent UI that the user may or may not consent to. Users aren't secure if they don't allow their authenticators to be used because of privacy concerns. With this being the default, we don't have to worry that users will opt out, and we give them great protections.

index.bs
```diff
@@ -1307,6 +1313,29 @@ example of the latter, when the user is accessing the [=[RP]=] from a given clie
...
+      CA, user consent UI, etc. This is the default value.
+    <li><dfn>verifiable</dfn> - indicates that the [=[RP]=] prefers an attestation type that leads to verifiable attestation
+      statements, even if that incurs extra cost such as network roundtrips and/or collection of user consent. For example, 
+      if a user agent supports the self attestation and Privacy CA attestation types, the user agent SHOULD use the Privacy CA
```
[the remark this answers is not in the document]
balfanz: Reworded this. Can you take another look?

index.bs
balfanz: Done.

index.bs
```diff
@@ -1527,6 +1546,48 @@ example of the latter, when the user is accessing the [=[RP]=] from a given clie
...
+      statements, but leaves it up to the client how to obtain such attestation statements. The client MAY replace the
+      authenticator-generated attestation statements with attestation statements generated by a Privacy CA, in order to protect
+      the privacy of the user, or to assist [=RPS=] with attestation verification in a heterogeneous ecosystem.
+      
+      Note that there is no guarantee that the [=RP=] will obtain a verifiable attestation statement in this case. The authenticator
+      may use self-attestation, or the user may have opted out of relaying any kind of (identifying) attestation information to
+      the RP
+    <li><dfn>direct</dfn> - indicates that the [=[RP]=] wants the attestation statement as generated by the authenticator. If
+      the client returns an attestation statement to the RP in this case, it MUST be the statement generated by the authenticator.
+      The client SHOULD obtain consent from the user in this case for relaying attestation information to the [=RP=].
```
akshayku: Why should platform penalize RP by adding user friction who ask for it legitimately and authenticators who are following all the rules set out for batch attestation?

I can understand consent situations when in incognito mode, but don't understand the normal case here.

index.bs
```diff
@@ -1527,6 +1546,48 @@ example of the latter, when the user is accessing the [=[RP]=] from a given clie
...
+      creation over obtaining a verifiable attestation. The client MAY replace the attestation statement generated
+      by the authenticator with a meaningless client-generated attestation statement in this case, for example in order to prevent
+      having to obtain user consent for relaying identifying information to the RP, or to save a roundtrip to a Privacy CA. 
+      The client SHOULD NOT prompt the user for consent to relay identifying information to the RP in this case, and MUST NOT 
+      fail the credential creation operation because of issues with the attestation. For example, if the client normally uses
+      a Privacy CA to anonymize attestations, and the Privacy CA is offline, the client must not fail the credential creation 
+      operation. Instead, it could replace the authenticator-generated attestation with a client-generated attestation.
+
+      This is the default value.
+    <li><dfn>verifiable</dfn> - indicates that the [=[RP]=] prefers an attestation type that leads to verifiable attestation
```
akshayku: Instead of verifiable, can we say Privacy-CA. Attestation produced by authenticators is also verifiable and adds confusion.

index.bs
```diff
@@ -839,6 +838,21 @@ When this method is invoked, the user agent MUST execute the following algorithm
...
+                    <dt>If the value of |attestationPreference| is "none" (or it has been left unspecified),</dt>
+                    <dd>replace potentially identifying information (such as AAGUID and attestation certificates) 
+                        in the attested credential data and attestation statement with non-identifying versions of the same data.</dd>
+
+                    <dt>If the value of |attestationPreference| is "verifiable",</dt>
+                    <dd>potentially replace the attestation statement with a more privacy-friendly and/or more easily verifiable
+                        version of the same data (for example, by employing a Privacy CA).</dd>
+ 
+                    <dt>If the value of |attestationPreference| is "direct",</dt>
+                    <dd>Obtain consent from the user to relay the authenticator's attestation statement to the RP. If the
```
akshayku: Consent should not be for normal case.

equalsJeffH:
Note that @akshayku made inline comments (about 5 hours ago upon my writing this) on the most recent proposed changes, but they are not highlighted here on this page because they are not a "review".  direct links to his inline comments: 

https://github.com/w3c/webauthn/pull/636/files#r150399658
https://github.com/w3c/webauthn/pull/636/files#r150399655
https://github.com/w3c/webauthn/pull/636/files#r150399643

index.bs
```diff
@@ -839,6 +838,21 @@ When this method is invoked, the user agent MUST execute the following algorithm
                 1.  Let |attestationObject| be a new {{ArrayBuffer}}, created using |global|'s [=%ArrayBuffer%=], containing the
                     bytes of <code>|credentialCreationData|.[=attestationObjectResult=]</code>'s value.
 
+                1.  Let |attestationPreference| be the value of |options|.{{MakePublicKeyCredentialOptions/attestation}}.
+                    <dl class="switch">
+                    <dt>If the value of |attestationPreference| is "none" (or it has been left unspecified),</dt>
+                    <dd>replace potentially identifying information (such as AAGUID and attestation certificates) 
+                        in the attested credential data and attestation statement with non-identifying versions of the same data.</dd>
```
equalsJeffH: should use [=Attested credential data=] and [=attestation statement=].

s/identifying/uniquely identifying/
s/non-identifying/blinded/ ?

index.bs
```diff
@@ -839,6 +838,21 @@ When this method is invoked, the user agent MUST execute the following algorithm
...
+
+                    <dt>If the value of |attestationPreference| is "verifiable",</dt>
+                    <dd>potentially replace the attestation statement with a more privacy-friendly and/or more easily verifiable
+                        version of the same data (for example, by employing a Privacy CA).</dd>
```
equalsJeffH: use [=Privacy CA=]

the distinction here from "none" is that in case the RP ends up needing it, there is possibly support for retroactive attestation unblinding, yes?

index.bs
equalsJeffH: I'd thought we were only talking about obtaining user consent in this case if the authenticator is known to be privacy un-friendly ?

index.bs
```diff
@@ -1388,6 +1403,10 @@ optionally evidence of [=user consent=] to a specific transaction.
     ::  This member is intended for use by [=[RPS]=] that wish to select the appropriate authenticators to participate in
         the {{CredentialsContainer/create()}} or {{CredentialsContainer/get()}} operation.
 
+    :   <dfn>attestation</dfn>
+    ::  This member is intended for use by [=[RPS]=] that wish to express their preference among different attestation types.
```
equalsJeffH: these are not "attestation types", as  noted above in

index.bs
```diff
@@ -1527,6 +1546,48 @@ example of the latter, when the user is accessing the [=[RP]=] from a given clie
 use a [=roaming authenticator=] which was originally registered with the [=[RP]=] using a different client.
 
 
+### Attestation Preference enumeration (enum <dfn enum>AttestationPreference</dfn>) ### {#attestation}
```
equalsJeffH: as noted above, am thinking "Attestation Presentation Preference" or "Attestation Conveyance Preference" is more accurately descriptive

equalsJeffH:
preliminary review: https://github.com/w3c/webauthn/pull/636#pullrequestreview-75952894

akshayku:
Can we pick one of these combinations which convey the meaning better.

`None`
`Proxy`
`Device`

OR

`None`
`Indirect`
`Direct`

index.bs
emlun: Yes, you're right. I've changed my mind since writing the previous comment. I agree that no attestation is a reasonable default.

index.bs
```diff
@@ -839,6 +838,21 @@ When this method is invoked, the user agent MUST execute the following algorithm
                 1.  Let |attestationObject| be a new {{ArrayBuffer}}, created using |global|'s [=%ArrayBuffer%=], containing the
                     bytes of <code>|credentialCreationData|.[=attestationObjectResult=]</code>'s value.
 
+                1.  Let |attestationPreference| be the value of |options|.{{MakePublicKeyCredentialOptions/attestation}}.
```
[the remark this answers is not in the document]
rlin1: I agree.

index.bs
rlin1: The "obtain user consent" handling should be identical for "verifiable" and "direct".  In both cases the authenticator model will be known to some other party (either the Privacy CA or the RP).
In my opinion Authenticator models satisfying the FIDO requirements (i.e. batch size >= 100000 and others) do not lead to a privacy risk and hence do not require yet another user consent (in addition to the general consent of using the authenticator a creating a credential).

index.bs
rlin1: The "obtain user consent" handling should be identical for "verifiable" and "direct".  In both cases the authenticator model will be known to some other party (either the Privacy CA or the RP).
In my opinion Authenticator models satisfying the FIDO requirements (i.e. batch size >= 100000 and others) do not lead to a privacy risk and hence do not require yet another user consent (in addition to the general consent of using the authenticator a creating a credential).

index.bs
rlin1: should use the term attestationPresentationPreference or attestationConveyancePreference here.

index.bs
rlin1: The "obtain user consent" handling should be identical for "verifiable" and "direct". In both cases the authenticator model will be known to some other party (either the Privacy CA or the RP).
In my opinion Authenticator models satisfying the FIDO requirements (i.e. batch size >= 100000 and others) do not lead to a privacy risk and hence do not require yet another user consent (in addition to the general consent of using the authenticator a creating a credential).

index.bs
rlin1: The "obtain user consent" handling should be identical for "verifiable" and "direct". In both cases the authenticator model will be known to some other party (either the Privacy CA or the RP).
In my opinion Authenticator models satisfying the FIDO requirements (i.e. batch size >= 100000 and others) do not lead to a privacy risk and hence do not require yet another user consent (in addition to the general consent of using the authenticator a creating a credential).

equalsJeffH:
@rlin1 has reviewed: https://github.com/w3c/webauthn/pull/636#pullrequestreview-76403786

christiaanbrand:
Telling an RP that you're using a Yubico token vs a Feitian token is a choice we think we need to involve the user in. That's the primary reason for showing additional consent.
We will probably also need to show consent before sending this info to a Privacy CA.

alradmsft:
@christiaanbrand said:

> Telling an RP that you're using a Yubico token vs a Feitian token is a choice we think we need to involve the user in. 

If this the principle you want to uphold, then you MUST show it for both privacy or direct attestation IMO. As a user, why should I feel more comfortable giving my private information to some thing called "privacy CA" run by a corporation and not comfortable giving it to an RP run by another or even same corporation? 

But if we go down that route, we would either be killing attestation because RPs would never want to deal with "scary" warnings that users say no to OR we are creating a warning that users ignore. I think we can agree that warnings that users ignore are not only useless, but are detrimental to other useful warnings. Do we think that attestation is also useless or bad?

index.bs
jcjones: I don't think we should mandate consent for this mode, either.

It's also semantically weird to require it at this point of the algorithm, as this is already broadly covered in Step 6 of https://w3c.github.io/webauthn/#op-make-cred and Step 5 of https://w3c.github.io/webauthn/#op-get-assertion, though admittedly the user agent only knows for sure how the authenticator responded here in the success branch.

As this is currently written, the user flow might be something like:

1. User clicks "register a strong authenticator"
1. Website renders a page explaining the authentication process and calls `create`
1. User Agent renders a dialog explaining the origin, RP, and the user account asking for the user to consent.  Probably something like, "To continue, touch your flashing token. Otherwise, you can click Cancel or Don't Show This Again."
1. User consents on the token
1. Because of this language, now User Agent says, "This token could be used by $origin to track you. Are you sure?"

I'd much rather roll the consent part together with the first dialog. We know [security indicators are difficult to get right](https://research.google.com/pubs/pub45366.html), and we know that warning fatigue is a real issue. Mandating that forevermore we display a dialog here after the user already interacted with their token seems like a premature UX choice.

If a browser wants to do this, I don't think we need to put a MUST NOT here -- I think we should have the freedom to innovate. Perhaps the first explanatory dialog I mentioned above could remain on screen  after the user expresses consent, and that dialog could be designed to say "Oh hey, since you haven't done this before: this thing you just did, it could be used to track you, so be careful and use another token for sensitive stuff if you need." 

Letting a UA do research and get that right seems great. Making us do it is going to be problematic. (Not even just because of my well-documented issues with mandatory UX in the spec)

rlin1:
Today, Web Browsers add the Mobile Device model name to the HTTP header - withoutr any user consent.
What exactly is the difference (from a privacy perspective)?

nadalin:
@christiaanbrand @balfanz So reading all the responses/reviews it does not look like we have consensus and there are a lot of unknowns in the design. I don't think it would be right just to add the enum (3 values) and not having definite actions/explanation/behavior fully defined.

nadalin:
@akshayku I like the labels/names for the enum NONE, PROXY, DIRECT but what are the semantics of these values as w/o agreement on the semantics this PR does not improve interoperability

akshayku:
@nadalin We are waiting for @balfanz / @christiaanbrand  to respond/address our concerns about scary UI and alternate name suggestions.

agl:
My apologies for being absent here: I was on vacation until yesterday.

> @nadalin We are waiting for @balfanz / @christiaanbrand to respond/address our concerns about scary UI and alternate name suggestions.

Addressing the names question. The current enum values in the PR are "none", "verifiable", and "direct". You [suggest](https://github.com/w3c/webauthn/pull/636#issuecomment-344012525) "none", "indirect" / "proxy", and "direct". I think we can easily change the middle value to "indirect".

R.e. UI, I agree [with J.C.](https://github.com/w3c/webauthn/pull/636#discussion_r151280967) that mandating consent is premature here. We determine security UI based on studies and metrics and I wouldn't want to prejudge that.

I will see whether Dirk agrees and, if so, work with him to update the PR.

agl:
p.s. while Dirk considers this, I've pushed my suggested changes to https://github.com/agl/webauthn/commit/2711064c7bb8f3ecbff340b96098e3ac1eef2ec2

akshayku:
@agl  Added some comments on agl/webauthn@2711064

agl:
@akshayku Thank you. Please see replies on that commit and updates in https://github.com/agl/webauthn/commit/c819a340d2ae2b97e95008a268347c3ef66a2a21

nadalin:
@jcjones @AngeloKai Please review before Wednesday meeting so we have all reviewers views on this one as we are looking to try to get to a consensus on or before Wednesday

balfanz:
Not sure where I'm supposed to comment, so I'll just add to the thread here:

- re/ "these are not attestation types": I think they are. The point of this parameter here is to help the client decide which of the attestation types defined in Section 6.3.3. ("Attestation Types") should be returned to the RP. If the RP insists on verifiable attestation coming directly from the Authenticator (i.e., Basic attestation or DAA), then they can say that here (using "direct"), and the client's job is to make sure the user understands what's going on, and then pass the Authenticator's attestation on to the RP. If the RP is ok with non-verifiable attestation (like self attestation), they can say that (using "none"), and the client knows that they can take certain shortcuts: not call the Privacy CA, not prompt the user, etc., and deliver *that* type of attestation to the RP.  I don't understand what an "AttestationPresentationPreference" is - a preference on how the attestation is presented? Presented to whom? To the user? We don't seem to explain in the spec how the attestation is presented to the user anywhere. I'm a bit confused about this naming choice.

- re/ "verifiable" vs. "indirect": I don't think that this value always means that a Privacy CA will be involved. The point of the original names was not to be prescriptive (i.e., prescribe to the client what they're supposed to do) but descriptive (describe what types of attestations the RP is willing to accept), and give the client more freedom to figure out what to do in each case.

Having said that, I don't want to waste time bikeshedding names this late in the game. Even though I liked the original names better, I'm ok with the PR as is.

index.bs
```diff
@@ -839,6 +838,20 @@ When this method is invoked, the user agent MUST execute the following algorithm
                 1.  Let |attestationObject| be a new {{ArrayBuffer}}, created using |global|'s [=%ArrayBuffer%=], containing the
                     bytes of <code>|credentialCreationData|.[=attestationObjectResult=]</code>'s value.
 
+                1.  Let |attestationPresentationPreference| be the value of |options|.{{MakePublicKeyCredentialOptions/attestation}}.
```
equalsJeffH: IIUC, we ought to rather add `attestationPresentationPreference` as an item of the `credentialCreationData` struct,  created in a prior step, and then utilize `credentialCreationData.attestationPresentationPreference`  in this step.

index.bs
```diff
@@ -839,6 +838,20 @@ When this method is invoked, the user agent MUST execute the following algorithm
...
+                    <dl class="switch">
```
equalsJeffH: for switches, we factor the "If the value of `foo`" portion up to the switch stmt

index.bs
```diff
@@ -839,6 +838,20 @@ When this method is invoked, the user agent MUST execute the following algorithm
...
+                    <dt>If the value of |attestationPresentationPreference| is "none" (or it has been left unspecified),</dt>
+                    <dd>replace potentially identifying information (such as AAGUID and attestation certificates) 
+                        in the [=attested credential data=] and [=attestation statement=] with blinded versions of the same data.</dd>
+
```
equalsJeffH: seems like we ought to have impl cons info somewhere in the spec that provides guidance wrt performaing such blinding.  I suppose that could be done at some point post-WD-07

index.bs
```diff
@@ -1527,6 +1545,40 @@ example of the latter, when the user is accessing the [=[RP]=] from a given clie
...
+    <li><dfn>indirect</dfn> - indicates that the [=[RP]=] prefers an attestation type that leads to verifiable attestation
+      statements, but leaves it up to the client how to obtain such attestation statements. The client MAY replace the
+      authenticator-generated attestation statements with attestation statements generated by a Privacy CA, in order to protect
+      the privacy of the user, or to assist [=RPS=] with attestation verification in a heterogeneous ecosystem.
+
+      Note that there is no guarantee that the [=RP=] will obtain a verifiable attestation statement in this case. (For
+      example, in the case that the authenticator uses self-attestation.)
+    <li><dfn>direct</dfn> - indicates that the [=[RP]=] wants the attestation statement as generated by the authenticator. If
+      the client returns an attestation statement to the RP in this case, it MUST be the statement generated by the authenticator.
+      If, for whatever reason, the client cannot pass on the authenticator-generated attestation statement, it MUST terminate
```
equalsJeffH: we would prefer in the direct case to simply be informed as to any issues the client may know of wrt the authnr, but still receive the attestation. Also, a processing requirment such as this needs to be in the relevant algorithm rather than here.

index.bs
```diff
@@ -1527,6 +1545,40 @@ example of the latter, when the user is accessing the [=[RP]=] from a given clie
 use a [=roaming authenticator=] which was originally registered with the [=[RP]=] using a different client.
 
 
+### Attestation Presentation Preference enumeration (enum <dfn enum>AttestationPresentationPreference</dfn>) ### {#attestation}
```
equalsJeffH: there's various editorial things in the below that ought to be cleaned up but it's easier for me to do a PR than to try to write it all in comments -- so will endeavor to do such on the plane tomorrow.  see also: https://github.com/w3c/webauthn/pull/636#issuecomment-345196396

index.bs
```diff
@@ -1527,6 +1545,40 @@ example of the latter, when the user is accessing the [=[RP]=] from a given clie
...
+        "none",
+        "indirect",
+        "direct"
+    };
+</pre>
+
+<ul>
+    <li><dfn>none</dfn> - indicates that the [=[RP]=] is not interested in attestation, and prefers fast and reliable credential
+      creation over obtaining a verifiable attestation. The client MAY replace the attestation statement generated
+      by the authenticator with a meaningless client-generated attestation statement in this case, for example in order to prevent
```
equalsJeffH: we need to correspondingly update the [Registering a new credential](https://w3c.github.io/webauthn/#registering-a-new-credential) relying party operation such that in the case of a blinded attestation that the RP not try to validate it, yes?

equalsJeffH:
@balfanz wrote:
> re/ "these are not attestation types": I think they are. The point of this parameter here is to help the client decide which of the attestation types defined in Section 6.3.3. ("Attestation Types") should be returned to the RP

Actually, in thinking about this further I would use the term `attestationConveyancePreference` rather then `attestationPresentationPreference` because it seems to me that that is what this is in regards to, rather than "attestation type" (after having gone back and re-read the several paragraphs beginning here: [attestation type](https://w3c.github.io/webauthn/#attestation-type)). 

* none - do not convey the attestation statement and AAGUID from the authenticator, and RP do not attempt to validate it the attestation statement. 
* indirect - client, and perhaps other 3d party, may alter the attestation statement. 
* direct - convey to the RP the authenticator's attestation statement and AAGUID.

I think that the blinding performed in the none case introduces an new [attestation type](https://w3c.github.io/webauthn/#attestation-type) which we ought to add to the [attestation types list](https://w3c.github.io/webauthn/#sctn-attestation-types).

Also, the present "privacy CA" attestation type ought to be name-changed to Attestation CA per TCG's specs and made clear that it is TCG-specific, and perhaps add another attestation type of "privacy CA" ( or someother name? attestation proxy? blinding proxy?)  i.e. the Chrome-proposed attestation type, which is not a TCG TPM attestation CA because (at least) it is the client who is employing it, not a TPM-based authenticator.

index.bs
emlun: I suggest something to the effect of:

>The client MAY replace the attestation statement generated by the authenticator with any value - including `null`, an empty object or a meaningless client-generated attestation statement - in this case, for example in order to...

I don't think we need to prescribe that the client has to generate a nonsense attestation statement when no attestation statement at all has equivalent security properties.

index.bs
akshayku: Responding in two places to make sure this is not lost:

Agree with @rlin1 concern. I'd like to suggest the following re-wording:

`If authenticator violates privacy requirements of the attestation scheme it is using, the client MUST terminate the credential generation operation with a AttestationNotPrivateError' 

In my opinion, this wording scopes the problem to privacy and at the same time extends it beyond batch attestation issues. For example, improper implementation of the TCG attestation scheme could also be covered.

christiaanbrand: Feel free to make a commit to this PR to reflect your changes.

akshayku: Done. https://github.com/w3c/webauthn/pull/636/commits/825cce701d532364a78c7ace2918d1a6a64a0641

@rlin1 / @equalsJeffH Is this description ok?

equalsJeffH:
@agl asks over in https://github.com/w3c/webauthn/pull/693#issuecomment-345463823 whether we ought to abandon this PR and land PR #693 directly to master instead (rather than land the latter on top of this PR which will require detailed conflict resolution) -- WDYT ?

rlin1:
Merging #693 into master would work for me.  But we'd have to "edit" the PR accordingly and check for conflicts.

christiaanbrand:
I agree with merging #693 and abandoning this one.

nadalin:
I would agree to close this PR and work on #693

akshayku:
I would also agree to close this PR and work on #693

nadalin:
@balfanz can you please close this and reference #693

balfanz:
We're going with #693 instead.

PR #641, opened 2017-10-12
clarified the ECDAA signing procedure. Close #591

### LINKED ISSUE #591: Description of attestation signature generation for ECDAA needs to be fixed.
§7.2 Packed Attestation Statement Format describes syntax and semantics of Packed Attestation Statement. 
The signing and verification attestation procedures are explained. 
In case of using ECDAA for the attestation, the signing procedure is somewhat weird.
Followings are the depicted signing procedure for ECDAA.
> If ECDAA is in use, the authenticator produces sig by concatenating authenticatorData and clientDataHash, and signing the result using ECDAA-Sign (see section 3.5 of [FIDOEcdaaAlgorithm]) with a ECDAA-Issuer public key selected through an authenticator-specific mechanism (see [FIDOEcdaaAlgorithm]). It sets alg to the algorithm of the ECDAA-Issuer public key and ecdaaKeyId to the identifier of the ECDAA-Issuer public key (see above).

In order to generated ECDAA signature, a signer (authenticator) generates signature with ECDAA credential and private key in stead of using ECDAA-Issuer public key.
So signing procedure for ECDAA should be fixed.

rlin1:
***
[Preview](https://s3.amazonaws.com/pr-preview/w3c/webauthn/clarify-ecdaa-sign-591.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/webauthn/641949f...f56c80e.html)

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2350,7 +2350,7 @@ implementable by [=authenticators=] with limited resources (e.g., secure element
         3. If [=ECDAA=] is in use, the authenticator produces |sig| by concatenating |authenticatorData| and |clientDataHash|, and
             signing the result using ECDAA-Sign (see section 3.5 of [[!FIDOEcdaaAlgorithm]]) after selecting an
             [=ECDAA-Issuer public key=] related to the ECDAA signature private key through an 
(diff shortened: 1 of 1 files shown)


equalsJeffH:
reviewed this. LGTM. let's merge.

Issue #650, opened 2017-10-16
authenticatorMakeCredential operation credential filtering


jovasco:
> 3. Check if a credential matching an item of excludeCredentialDescriptorList is present on this authenticator. If so, return an error code equivalent to "NotAllowedError" and terminate the operation.

This should explicitly mention that the the item must be checked to match the provided RP ID.

See PR #651

equalsJeffH:
This seems like a correct update to that step to me -- others?
If so, I'll fix this in PR #498 and we can close PR #651.

emlun:
Sounds good to me.

nadalin:
@jovasco This seems just like a editorial issue.

equalsJeffH:
@nadalin, this is technical.

equalsJeffH:
am assigning to @jovasco since he authored PR #651 intended to resolve this, and I ended up not addressing it in PR #498.

jcjones:
Closed in #651.

Issue #280, opened 2016-11-04
"might be present on this authenticator" could use a clearer definition


bzbarsky:
https://w3c.github.io/webauthn/#getAssertion step 7 says:

> execute a platform-specific procedure to determine which, if any, credentials listed in allowList might be present on this authenticator

It's not clear to me what this means.  Does the "id" of the ScopedCredentialDescriptor need to match something?  Does the type?  This seems impossible to implement interoperably as written.

equalsJeffH:
@bzbarsky does this https://github.com/w3c/webauthn/commit/b3dba2c356f231ab3c96608cfcbf96fe3ca1df7f address this issue?  the latter was merged to master in PR #371

bzbarsky:
It still seems unclear what an interoperable implementation here would look like, but at least now it's clearer than id and type are, or can be, involved....

equalsJeffH:
The present language is (step 16.2 of https://www.w3.org/TR/2017/WD-webauthn-20170811/#getAssertion (WD-06):
> If options.allowCredentials is not empty, execute a platform-specific procedure to determine which, if any, public key credentials described by options.allowCredentials are bound to this authenticator, by matching with rpId, options.allowCredentials.id, and options.allowCredentials.type. ...

Am closing this issue. if problems arise, eg the upcoming interop, due specifically to unclarity in this particular alg step, please re-open and provide details.

PR #651, opened 2017-10-16
Only exclude CredIDs matching the RPID

jovasco:
Only credentials in the exclude credentials list that match this RPID should result in a not allowed error.

***
[Preview](https://s3.amazonaws.com/pr-preview/jovasco/webauthn/patch-1.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/webauthn/ee174c2...jovasco:6e5f27f.html)

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1831,7 +1831,7 @@ When this method is invoked, the [=authenticator=] must perform the following pr
 1. If |allowCredentialDescriptorList| was not supplied, set it to a list of all credentials stored for |rpId| (as determined by
     an exact match of |rpId|).
 1. Remove any items from |allowCredentialDescriptorList| that are not present on this [=authenticator=] for
-    the supplied <code>|rpEntity|.{{PublicKeyCredentialRpEntity/id}}</code>
+    the supplied |rpId|.
 1. If |allowCredentialDescriptorList| is now empty, return an error code equivalent to "{{NotAllowedError}}" and terminate the
     operation.
 1. Prompt the user to select a [=public key credential|credential=] |selectedCredential| from |allowCredentialDescriptorList|.
--- a/index.bs
+++ b/index.bs
@@ -1763,7 +1763,8 @@ When this operation is invoked, the authenticator must perform the following pro
     |credTypesAndPubKeyAlgs| is supported.
     If not, return an error code equivalent to "{{NotSupportedError}}" and terminate the operation.
 1. Check if a credential matching an [=list/item=] of |excludeCredentialDescriptorList| is present on this authenticator for
-    the supplied |rpId|. If so, return an error code equivalent to "{{NotAllowedError}}" and terminate the operation.
+    the supplied <code>|rpEntity|.{{PublicKeyCredentialRpEntity/id}}</code>. If so, return an error code equivalent to 
+    "{{NotAllowedError}}" and terminate the operation.
 1. If |requireResidentKey| is |true| and the authenticator cannot store a [=Client-side-resident Credential
     Private Key=], return an error code equivalent to "{{ConstraintError}}" and terminate the operation.
 1. If |requireUserVerification| is |true| and the authenticator cannot perform user verification,
--- a/index.bs
+++ b/index.bs
@@ -1762,9 +1762,10 @@ When this operation is invoked, the authenticator must perform the following pro
 1. Check if at least one of the specified combinations of {{PublicKeyCredentialType}} and cryptographic parameters in
     |credTypesAndPubKeyAlgs| is supported.
     If not, return an error code equivalent to "{{NotSupportedError}}" and terminate the operation.
-1. Check if a credential matching an [=list/item=] of |excludeCredentialDescriptorList| is present on this authenticator for
-    the supplied <code>|rpEntity|.{{PublicKeyCredentialRpEntity/id}}</code>. If so, return an error code equivalent to 
-    "{{NotAllowedError}}" and terminate the operation.
+1. Check if any credential bound to this authenticator matches an [=list/item=] of |excludeCredentialDescriptorList|. A match 
+   occurs if a credential matches <code>|rpEntity|.{{PublicKeyCredentialRpEntity/id}}</code>, <code>|excludeCredentialDescriptorList|. 
+   {{PublicKeyCredentialDescriptor/id}}</code>, and <code>|excludeCredentialDescriptorList|.{{PublicKeyCredentialDescriptor/type}}</code>. 
+   If so, return an error code equivalent to "{{NotAllowedError}}" and terminate the operation.
 1. If |requireResidentKey| is |true| and the authenticator cannot store a [=Client-side-resident Credential
     Private Key=], return an error code equivalent to "{{ConstraintError}}" and terminate the operation.
 1. If |requireUserVerification| is |true| and the authenticator cannot perform user verification,
--- a/index.bs
+++ b/index.bs
@@ -1762,8 +1762,8 @@ When this operation is invoked, the authenticator must perform the following pro
 1. Check if at least one of the specified combinations of {{PublicKeyCredentialType}} and cryptographic parameters in
     |credTypesAndPubKeyAlgs| is supported.
     If not, return an error code equivalent to "{{NotSupportedError}}" and terminate the operation.
-1. Check if a credential matching an [=list/item=] of |excludeCredentialDescriptorList| is present on this authenticator. If
-    so, return an error code equivalent to "{{NotAllowedError}}" and terminate the operation.
+1. Check if a credential matching an [=list/item=] of |excludeCredentialDescriptorList| is present on this authenticator for
+    the supplied |rpId|. If so, return an error code equivalent to "{{NotAllowedError}}" and terminate the operation.
 1. If |requireResidentKey| is |true| and the authenticator cannot store a [=Client-side-resident Credential
     Private Key=], return an error code equivalent to "{{ConstraintError}}" and terminate the operation.
 1. If |requireUserVerification| is |true| and the authenticator cannot perform user verification,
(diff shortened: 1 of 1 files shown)


emlun:
Does anyone object to marking this as non-substantive to let @jovasco through the check?

emlun:
If not, I'll do so at 2017-10-17T16:00Z.

emlun:
Scratch that, I'll leave this to the requested reviewers instead. :)

index.bs
```diff
@@ -1762,8 +1762,8 @@ When this operation is invoked, the authenticator must perform the following pro
 1. Check if at least one of the specified combinations of {{PublicKeyCredentialType}} and cryptographic parameters in
     |credTypesAndPubKeyAlgs| is supported.
     If not, return an error code equivalent to "{{NotSupportedError}}" and terminate the operation.
-1. Check if a credential matching an [=list/item=] of |excludeCredentialDescriptorList| is present on this authenticator. If
-    so, return an error code equivalent to "{{NotAllowedError}}" and terminate the operation.
+1. Check if a credential matching an [=list/item=] of |excludeCredentialDescriptorList| is present on this authenticator for
+    the supplied |rpId|. If so, return an error code equivalent to "{{NotAllowedError}}" and terminate the operation.
```
equalsJeffH: actually, the discrete |rpId| parameter is disappearing in PR #498 because it is duplicated in `rpEntity.id`. That fix is in PR #498.  I'd also word this a bit differently.

jovasco: Wel, I am not a native English speaker :),

equalsJeffH: No worries -- is there a submitted issue corresponding to this PR ?  if not please submit one, and I can/will address in pr #498

equalsJeffH: Ok, I submitted issue #654 for this.

equalsJeffH:
I submitted issue #654 for this -- if another(s) can double-check that this change is correct, I'll address issue #654 in PR #498 and we can close this PR.

index.bs
equalsJeffH: sigh -- @jovasco had already submitted issue #650 which I'd missed. thx.

jcjones:
I disagree with adding this into the algorithm steps, as it implies checking the `rpId` of a credential before using it -- which is impossible for the implementing browser to do, as `PublicKeyCredentialDescriptor` doesn't provide a way to check that.

There should be a correspondence between a Credential ID and an RP ID, but it has to be managed by the authenticator. Some authenticators might use business logic here ("I made this CredID for X.Y, but this is for Z.Y, nope!"), some might use cryptography to mix the RP ID into the Credential ID's byte array. That's an implementation detail.

IMO this should go into the security considerations section instead. Maybe something to the effect of:

```
Authenticators, when presented a Credential ID for an RP ID, SHOULD 
ensure that the Credential ID was created for this RP ID before accepting
it for use.
```

jcjones:
Apologies, I misunderstood which section this was actually being placed into -- this does make sense for [the authenticator's makeCredential algorithm](https://w3c.github.io/webauthn/#op-make-cred). I'll properly review.

jcjones:
Marked as non substantive for IPR from ash-nazg.

equalsJeffH:
Ok, since this is not as clear-cut as I'd thought wrt the language to use, I will not address this  (and issue #650) in PR #498.  will comment further in this PR.

index.bs
akshayku: It has to be `rpEntity.id`. Apart from that it looks fine to me.

jovasco: @akshayku Fixed.

index.bs
```diff
@@ -1762,8 +1762,9 @@ When this operation is invoked, the authenticator must perform the following pro
 1. Check if at least one of the specified combinations of {{PublicKeyCredentialType}} and cryptographic parameters in
     |credTypesAndPubKeyAlgs| is supported.
     If not, return an error code equivalent to "{{NotSupportedError}}" and terminate the operation.
-1. Check if a credential matching an [=list/item=] of |excludeCredentialDescriptorList| is present on this authenticator. If
-    so, return an error code equivalent to "{{NotAllowedError}}" and terminate the operation.
+1. Check if a credential matching an [=list/item=] of |excludeCredentialDescriptorList| is present on this authenticator for
+    the supplied <code>|rpEntity|.{{PublicKeyCredentialRpEntity/id}}</code>. If so, return an error code equivalent to 
+    "{{NotAllowedError}}" and terminate the operation.
```
equalsJeffH: this step, where we are matching creds bound to an authnr to credDescriptors, is similar to step 16.2 in section [#discover-from-external-source](https://w3c.github.io/webauthn/#discover-from-external-source). Note that the latter step matches on rpId, credential ID, and cred type.  

Should not this authenticatorMakeCredential step similarly match on the same items?

e.g.: 
`1. Check if a credential matching an [=list/item=] of |excludeCredentialDescriptorList| is bound to this authenticator by matching with <code>|rpEntity|.{{PublicKeyCredentialRpEntity/id}}</code>, <code>|excludeCredentialDescriptorList|.{{PublicKeyCredentialDescriptor/id}}</code>, and <code>|excludeCredentialDescriptorList|.{{PublicKeyCredentialDescriptor/type}}</code>. If so, return an error code equivalent to "{{NotAllowedError}}" and terminate the operation.`

equalsJeffH:
review of this is here: https://github.com/w3c/webauthn/pull/651#pullrequestreview-71259617

equalsJeffH:
Do we similarly need to refine step 3 of the authenticatorGetAssertion operation ([#op-get-assertion](https://w3c.github.io/webauthn/#op-get-assertion))?

akshayku:
@equalsJeffH Yes, we do.

index.bs
jovasco: I believe that is inherent in checking <code>[=list/item]</code> but the check for the rpID is not. It perfectly possible that an item *is* present on the authenticator but for a different rpID while if ID and type are invalid, this cannot be a credential on this authenticator.

jovasco:
@equalsJeffH Yes.

jovasco:
Fixed.

nadalin:
@akshayku Have you reviewed the changes yet as we would like to merge this

index.bs
equalsJeffH: It seems you @jovasco are saying that we do not need to be explicit regarding the definition of "matching" in this context.  However, the present language in step 16.2 in section [#discover-from-external-source](https://w3c.github.io/webauthn/#discover-from-external-source) is the result of addressing issue #280.  I would not be surprised to have a similar issue raised here down the road if we are not explicit.  Is there any harm/errors with the proposed language above?

equalsJeffH:
i still have a question here: https://github.com/w3c/webauthn/pull/651#discussion_r146402013

index.bs
jovasco: @equalsJeffH Alright, if we do this, we better do it right:
1. Check if a credential matching an <code>[=list/item=]</code> of <code>|excludeCredentialDescriptorList|</code> is bound to this authenticator with the following procedure:
   1. verify that the <code>|excludeCredentialDescriptorList|.{{PublicKeyCredentialDescriptor/id}}</code> was generated by this Authenticator.
   1.  verify that the <code>|excludeCredentialDescriptorList|.{{PublicKeyCredentialDescriptor/type}}</code> matches the <code>|excludeCredentialDescriptorList|.{{PublicKeyCredentialDescriptor/id}}</code> 
   1. verify that <code>|rpEntity|.{{PublicKeyCredentialRpEntity/id}}</code> is identical to the one used to create the <code>|excludeCredentialDescriptorList|.{{PublicKeyCredentialDescriptor/id}}</code>.

If all of these check succeed, return an error code equivalent to "{{NotAllowedError}}" and terminate the operation.

If you agree I will update the pull request for both make crednetial and getassertion.

equalsJeffH: @jovasco proposed:
> 1. Check if a credential matching an <code>[=list/item=]</code> of <code>|excludeCredentialDescriptorList|</code> is bound to this authenticator with the following procedure:

hm, I think i see whay you are getting at -- i.e., being even more explicit -- though, this does not seem correct to me...in the above, do you mean to also say "...for each item of |excludeCredentialDescriptorList| ?

>          i. verify that the <code>|excludeCredentialDescriptorList|.
>              {{PublicKeyCredentialDescriptor/id}}</code> 
>              was generated by this Authenticator.

in the above, "generated by this authnr" is not what we are looking for. Rather, we are trying to ascertain whether there exists a cred bound to this authnr having a credential ID matching `|excludeCredentialDescriptorList|.{{PublicKeyCredentialDescriptor/id}}`.

>         ii.  verify that the <code>|excludeCredentialDescriptorList|.
>              {{PublicKeyCredentialDescriptor/type}}</code> matches the
>             <code>|excludeCredentialDescriptorList|.{{PublicKeyCredentialDescriptor/id}}</code> 

in the above, do you mean to say something like "if there exists a cred bound to this authnr whose `{{PublicKeyCredentialDescriptor/type}}` value matches `|excludeCredentialDescriptorList|.{{PublicKeyCredentialDescriptor/type}}` ?

>         iii. verify that <code>|rpEntity|.{{PublicKeyCredentialRpEntity/id}}</code> is 
>              identical to the one used to create the <code>|excludeCredentialDescriptorList|.
>              {{PublicKeyCredentialDescriptor/id}}</code>.

in the above, do you mean to say something like "if there exists a cred bound to this authnr whose RP ID matches `|excludeCredentialDescriptorList|.{{PublicKeyCredentialDescriptor/id}}`" ?


Perhaps the below works ?

`3. Check if any credential, bound to this authenticator and matching an 
[=list/item=] of |excludeCredentialDescriptorList|, exists.  
    A match occurs if a credential matches all of an 
|excludeCredentialDescriptorList| item's <code>|rpEntity|.
{{PublicKeyCredentialRpEntity/id}}</code>, <code>|excludeCredentialDescriptorList|.
{{PublicKeyCredentialDescriptor/id}}</code>, and 
<code>|excludeCredentialDescriptorList|.{{PublicKeyCredentialDescriptor/type}}</code>.
     If so, return an error code equivalent to "{{NotAllowedError}}" and 
terminate the operation.`

equalsJeffH:
reply to @jovasco here: https://github.com/w3c/webauthn/pull/651#discussion_r148131877

jcjones:
I'm willing to take over this PR to resolve @equalsJeffH 's comments if @jovasco does not have time to do it this week. I can do it on Friday, 3 Nov.

index.bs
jovasco: @equalsJeffH  I modified it slightly. rpId is not part of the exclude list item and I found the language a bit tortured.

equalsJeffH:
thanks, this looks good to me.

Issue #613, opened 2017-10-05
Allow hot-plugged authenticators?


emlun:
In discussing the Firefox implementation with @jcjones we discovered a wrinkle in the `create()` and `get()` ceremonies: Hot-plugged authenticators - yes or no?

Currently, [§4.1.3. Create a new credential][make-cred] and [§4.1.4. Use an existing credential to make an assertion][assert] both specify their procedure on the form

 1. Prepare authenticator arguments
 2. Let `currentlyAvailableAuthenticators` be a new ordered set consisting of all authenticators currently available on this platform.
 3. Send command to each element of `currentlyAvailableAuthenticators`
 4. Wait for success or timeout

This seems to imply that hot-plugging is not supported - that authenticators that become available after (2) but before (4) must be ignored. However there are multiple benefits to allowing hot-plugging:

- This is how the U2F implementations in Chrome and Firefox work, so users are accustomed to it.
- Relaxing the order of events makes for a much smoother user experience.
- It simplifies use of the JS API by making it less stateful. Without it the only way for the caller to ensure that the authenticator is ready, before calling `create()` or `get()`, is to prompt the user to ready the authenticator (i.e. plug it in) and then manually confirm this (i.e. press the "Next" button).
- The language "currently available" becomes very vague in the context of NFC, as the authenticator is necessarily "plugged in" _after_ the call to `create()` or `get()`.

The main caveat is that a call to `create()` or `get()` whose `excludeCredentials` or `allowCredentials` exclude all currently available authenticators would also wait for the timeout, which may not always be wanted. But on the other hand, the vast majority of users will probably only have one authenticator anyway.

I can draft a suggestion for this if it sounds like a good idea.

[make-cred]: https://w3c.github.io/webauthn/#createCredential
[assert]: https://w3c.github.io/webauthn/#getAssertion

jcjones:
A very good point. Right now our WebAuthn implementation allows hotplugging... while the algorithm language is pretty clear on not allowing hotplugging. 

https://github.com/w3c/webauthn/blob/0cae38154c59cf4760e7dfed00f2c44e298f8c9f/index.bs#L661-L664

Obviously we paid very close attention to the specific algorithm language! 😀

jovasco:
For BLE authenticators, it is almost certain that the authenticators will appear somewhere during the process, whenever the user turns it on.

equalsJeffH:
yeah, this seems to be a good point.  

@jcjones 
[...]

PR #655, opened 2017-10-18
Fix #574 - Change language for Create and Get to support hotplugging

### LINKED ISSUE #574: Not clear what to do with cross platform authenticators during make-an-assertion step
The current text in section 4.1.4 step 16 says:

```
16. For each authenticator currently available on this platform, perform the following steps:

  1. Let allowCredentialDescriptorList be a new list.
  2. If options.allowCredentials is not empty, execute a platform-specific procedure to determine 
      which, if any, public key credentials described by options.allowCredentials are bound to this 
      authenticator, by matching with rpId, options.allowCredentials.id, and 
       options.allowCredentials.type. Set allowCredentialDescriptorList to this filtered list.
```

If an authenticator is `xplat` (or cross platform), say an NFC token that only appears once the user taps it, how does that jive with step 16 which says "... currently available on this platform..."?

Also, "...execute a platform-specific procedure..." works for platform authenticators, but not for cross-platform authenticators.

jcjones:
This is an incomplete fix; a full fix is intended to be handled [by the resolution to] Issue #613.

This reorders [a portion of the guts of] the Create and Get operations to 
indicate that the algorithms for
interacting with devices should be applied as devices are hotplugged / arrive.
It does not specify what happens when devices are removed, nor does it use
precise language. I'm not sure what language would be appropriate in this world,
so this patch is just to make things "better" not "correct".

***
[Preview](https://s3.amazonaws.com/pr-preview/jcjones/webauthn/574-hotplugging.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/webauthn/01aa320...jcjones:83da7ff.html)

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -682,30 +682,25 @@ When this method is invoked, the user agent MUST execute the following algorithm
 
 1. Let |clientDataHash| be the [=hash of the serialized client data=] represented by |clientDataJSON|.
 
-1. Let |currentlyAvailableAuthenticators| be a new [=ordered set=] consisting of all [=authenticators=]
-    currently available on this platform.
+1. Let |lifetimeTimer| be a timer for |adjustedTimeout| milliseconds.
 
-1. Let |selectedAuthenticators| be a new [=ordered set=].
+1. Let |issuedRequests| be a new [=ordered set=].
 
-1. If |currentlyAvailableAuthenticators| [=list/is empty=], return a {{DOMException}} whose name is
-    "{{NotFoundError}}", and terminate this algorithm.
+1. If <code>|options|.{{MakePublicKeyCredentialOptions/authenticatorSelection}}</code> is
+    [=present|present=], [=set/for each=] |authenticator| that becomes available on this
+    platform during the lifetime of |lifetimeTimer| and do the following:
 
-1. If <code>|options|.{{MakePublicKeyCredentialOptions/authenticatorSelection}}</code> is [=present|present=], iterate through
-        |currentlyAvailableAuthenticators| and do the following [=set/for each=] |authenticator|:
+    Issue: The definitions of "lifetime of" and "becomes available" are intended to represent how
+    devices are hotplugged into browsers, and are under-specified. Resolving this with good definitions
+    or some other means will be addressed by resolving [Issue #613](https://github.com/w3c/webauthn/issues/613).
+
     1. If {{AuthenticatorSelectionCriteria/authenticatorAttachment}} is [=present|present=] and its value is not equal
         to |authenticator|'s attachment modality, [=iteration/continue=].
     1. If {{AuthenticatorSelectionCriteria/requireResidentKey}} is set to |true| and the |authenticator|
         is not capable of storing a [=Client-Side-Resident Credential Private Key=], [=iteration/continue=].
     1. If {{AuthenticatorSelectionCriteria/requireUserVerification}} is set to |true| and the
         |authenticator| is not capable of performing [=user verification=], [=iteration/continue=].
-    1. [=set/Append=] |authenticator| to |selectedAuthenticators|.
 
-1. If |selectedAuthenticators| [=list/is empty=], return a {{DOMException}} whose name is
-    "{{ConstraintError}}", and terminate this algoritm.
-
-1. Let |issuedRequests| be a new [=ordered set=].
-
-1. [=set/For each=] |authenticator| in |currentlyAvailableAuthenticators|:
     1. Let |excludeCredentialDescriptorList| be a new [=list=].
 
     1. [=list/For each=] credential descriptor |C| in <code>|options|.{{MakePublicKeyCredentialOptions/excludeCredentials}}</code>:
@@ -725,13 +720,11 @@ When this method is invoked, the user agent MUST execute the following algorithm
 
     1. [=set/Append=] |authenticator| to |issuedRequests|.
 
-1. Start a timer for |adjustedTimeout| milliseconds.
-
-1. [=While=] |issuedRequests| [=list/is not empty=], perform the following actions depending upon the |adjustedTimeout| timer
-    and responses from the authenticators:
+1. [=While=] |issuedRequests| [=list/is not empty=], perform the following actions depending upon the
+    |lifetimeTimer| timer and responses from the authenticators:
     <dl class="switch">
 
-        :   If the |adjustedTimeout| timer expires,
+        :   If the |lifetimeTimer| timer expires,
         ::  [=set/For each=] |authenticator| in |issuedRequests| invoke the [=authenticatorCancel=] operation on |authenticator|
             and [=set/remove=] |authenticator| from |issuedRequests|.
 
@@ -908,12 +901,16 @@ method is invoked, the user agent MUST:
 
 1. Let |issuedRequests| be a new [=ordered set=].
 
-1. If there are no [=authenticators=] currently available on this platform, return a {{DOMException}} whose name is
-    "{{NotFoundError}}", and terminate this algorithm.
-
 1. Let |authenticator| be a platform-specific handle whose value identifies an [=authenticator=].
 
-1. For each |authenticator| currently available on this platform, perform the following steps:
+1. Let |lifetimeTimer| be a timer for |adjustedTimeout| milliseconds.
+
+1. [=set/For each=] |authenticator| that becomes available on this platform during the lifetime of
+    |lifetimeTimer|, perform the following steps:
+
+    Issue: The definitions of "lifetime of" and "becomes available" are intended to represent how
+    devices are hotplugged into browsers, and are under-specified. Resolving this with good definitions
+    or some other means will be addressed by resolving [Issue #613](https://github.com/w3c/webauthn/issues/613).
 
     1. Let |allowCredentialDescriptorList| be a new [=list=].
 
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -866,12 +859,14 @@ method is invoked, the user agent MUST:
...
-
 1. Let |authenticator| be a platform-specific handle whose value identifies an [=authenticator=].
 
-1. For each |authenticator| currently available on this platform, perform the following steps:
+1. For each |authenticator| that becomes available on this platform within the overall lifetime of
+    this operation, perform the following steps:
+
+    Note: The definitions of "overall lifetime" and "becomes available" are intended to represent how
+    devices are hotplugged into browsers, and are under-specified. Resolving this with good definitions
+    or some other means will be handled in [Issue #613](https://github.com/w3c/webauthn/issues/613).
```
equalsJeffH: pedantic nit:  s/handled in/addressed by resolving/

index.bs
```diff
@@ -676,30 +676,23 @@ When this method is invoked, the user agent MUST execute the following algorithm
...
-        |authenticator| is not capable of performing [=user verification=], [=iteration/continue=].
-    1. [=set/Append=] |authenticator| to |selectedAuthenticators|.
+1. [=set/For each=] |authenticator| that becomes available on this platform within the overall
+    lifetime of this operation, perform the following steps:
 
-1. If |selectedAuthenticators| [=list/is empty=], return a {{DOMException}} whose name is
-    "{{ConstraintError}}", and terminate this algoritm.
+    Note: The definitions of "overall lifetime" and "becomes available" are intended to represent how
+    devices are hotplugged into browsers, and are under-specified. Resolving this with good definitions
+    or some other means will be handled in [Issue #613](https://github.com/w3c/webauthn/issues/613).
```
equalsJeffH: pedantic nit:  s/handled in/addressed by resolving/

equalsJeffH: also, rather than a "Note:", perhaps should use "Issue:" here ?

index.bs
equalsJeffH: ditto wrt "Issue:"

index.bs
```diff
@@ -676,30 +676,23 @@ When this method is invoked, the user agent MUST execute the following algorithm
...
-1. If |options|.{{MakePublicKeyCredentialOptions/authenticatorSelection}} is [=present|present=], iterate through
-        |currentlyAvailableAuthenticators| and do the following [=set/for each=] |authenticator|:
-    1. If {{AuthenticatorSelectionCriteria/authenticatorAttachment}} is [=present|present=] and its value is not equal
-        to |authenticator|'s attachment modality, [=iteration/continue=].
-    1. If {{AuthenticatorSelectionCriteria/requireResidentKey}} is set to |true| and the |authenticator|
-        is not capable of storing a [=Client-Side-Resident Credential Private Key=], [=iteration/continue=].
-    1. If {{AuthenticatorSelectionCriteria/requireUserVerification}} is set to |true| and the
-        |authenticator| is not capable of performing [=user verification=], [=iteration/continue=].
-    1. [=set/Append=] |authenticator| to |selectedAuthenticators|.
+1. [=set/For each=] |authenticator| that becomes available on this platform within the overall
```
equalsJeffH: Offhand, I'm thinking that the "Start a timer for adjustedTimeout milliseconds", that is set down below in this alg, would come up to around this here step, which would then define "overall lifetime".   but perhaps that's something to wait to incorp in the resolution of issue #613.

index.bs
```diff
@@ -866,12 +859,14 @@ method is invoked, the user agent MUST:
 
 1. Let |issuedRequests| be a new [=ordered set=].
 
-1. If there are no [=authenticators=] currently available on this platform, return a {{DOMException}} whose name is
-    "{{NotFoundError}}", and terminate this algorithm.
-
 1. Let |authenticator| be a platform-specific handle whose value identifies an [=authenticator=].
 
-1. For each |authenticator| currently available on this platform, perform the following steps:
+1. For each |authenticator| that becomes available on this platform within the overall lifetime of
```
equalsJeffH: ditto here wrt "Start a timer for adjustedTimeout milliseconds" being done here.

equalsJeffH:
This "improves" issue #613.

equalsJeffH:
would be good to get at least one other review on this by someone(s) other than @jcjones and myself before we merge!  
cc: @akshayku @AngeloKai @jyasskin

index.bs
```diff
@@ -682,30 +682,25 @@ When this method is invoked, the user agent MUST execute the following algorithm
...
+1. Let |issuedRequests| be a new [=ordered set=].
+
+1. If <code>|options|.{{MakePublicKeyCredentialOptions/authenticatorSelection}}</code> is
+    [=present|present=], [=set/for each=] |authenticator| that becomes available on this
+    platform during the lifetime of |lifetimeTimer| and do the following:
 
-1. If |currentlyAvailableAuthenticators| [=list/is empty=], return a {{DOMException}} whose name is
-    "{{NotFoundError}}", and terminate this algorithm.
+    Issue: The definitions of "lifetime of" and "becomes available" are intended to represent how
+    devices are hotplugged into browsers, and are under-specified. Resolving this with good definitions
```
AngeloKai: If we intend to address what hotplug is after CR, I'd recommend giving an example such as NFC which people relate to more.

index.bs
```diff
@@ -682,30 +682,25 @@ When this method is invoked, the user agent MUST execute the following algorithm
 
 1. Let |clientDataHash| be the [=hash of the serialized client data=] represented by |clientDataJSON|.
 
-1. Let |currentlyAvailableAuthenticators| be a new [=ordered set=] consisting of all [=authenticators=]
-    currently available on this platform.
+1. Let |lifetimeTimer| be a timer for |adjustedTimeout| milliseconds.
```
AngeloKai: Do we need to add a new name? We can just change adjustedTimeout to lifetimeTimer. 

Note: this is more of a nitpick. Feel free to merge the PR without changing the adjustedTimeout naming.

AngeloKai:
The PR looks nominally ok given it is  one step toward the right direction. I am the same question as @equalsJeffH about whether we think the PR is good enough to address hot-plugged authenticator before CR. If so, I encourage us to think about what error should be thrown if there is only one authenticator connected via NFC and the user takes the authenticator move in the middle. AbortError seems like a reasonable candidate here. Developers can act based on the error to ask the user to put the authenticator back on again.

jcjones:
Thanks for the review, @AngeloKai. I've made both sets of corrections. Assuming it passes bikeshed, I think it's ready to merge - maybe with one last look by @equalsJeffH ?

index.bs
```diff
@@ -725,13 +721,11 @@ When this method is invoked, the user agent MUST execute the following algorithm
...
-1. Start a timer for |adjustedTimeout| milliseconds.
-
-1. [=While=] |issuedRequests| [=list/is not empty=], perform the following actions depending upon the |adjustedTimeout| timer
-    and responses from the authenticators:
+1. [=While=] |issuedRequests| [=list/is not empty=], perform the following actions depending upon the
+    |lifetimeTimer| timer and responses from the authenticators:
     <dl class="switch">
 
-        :   If the |adjustedTimeout| timer expires,
+        :   If the |lifetimeTimer| timer expires,
```
equalsJeffH: s/If the |lifetimeTimer| timer/If |lifetimeTimer|/

index.bs
```diff
@@ -725,13 +721,11 @@ When this method is invoked, the user agent MUST execute the following algorithm
 
     1. [=set/Append=] |authenticator| to |issuedRequests|.
 
-1. Start a timer for |adjustedTimeout| milliseconds.
-
-1. [=While=] |issuedRequests| [=list/is not empty=], perform the following actions depending upon the |adjustedTimeout| timer
-    and responses from the authenticators:
+1. [=While=] |issuedRequests| [=list/is not empty=], perform the following actions depending upon the
+    |lifetimeTimer| timer and responses from the authenticators:
```
equalsJeffH: s/the |lifetimeTimer| timer/|lifetimeTimer|/

index.bs
```diff
@@ -969,16 +968,14 @@ method is invoked, the user agent MUST:
 
     1. [=set/Append=] |authenticator| to |issuedRequests|.
 
+1. Execute the following steps [=in parallel=]. The [=task source=] for these [=tasks=] is the [=dom manipulation task source=].
 
-1. Start a timer for |adjustedTimeout| milliseconds. Then execute the following steps [=in parallel=]. The [=task source=] for
-    these [=tasks=] is the [=dom manipulation task source=].
-
-    1. While |issuedRequests| [=list/is not empty=], perform the following actions depending upon the |adjustedTimeout| timer
+    1. While |issuedRequests| [=list/is not empty=], perform the following actions depending upon the |lifetimeTimer| timer
```
equalsJeffH: s/the |lifetimeTimer| timer/|lifetimeTimer|/

index.bs
```diff
@@ -969,16 +968,14 @@ method is invoked, the user agent MUST:
...
         and responses from the authenticators:
 
         <dl class="switch">
 
-            :   If the |adjustedTimeout| timer expires,
+            :   If the |lifetimeTimer| timer expires,
```
equalsJeffH: s/If the |lifetimeTimer| timer/If |lifetimeTimer|/

equalsJeffH:
re-reviewed, thanks!

PR #662, opened 2017-10-25
fix #657: where does return value of PublicKeyCredential.getClientExtensionResults() come from?

### LINKED ISSUE #657: where does return value of PublicKeyCredential.getClientExtensionResults() come from?
[PublicKeyCredential Interface used to look like this](https://www.w3.org/TR/webauthn/#iface-pkcredential):
```
[SecureContext]
interface PublicKeyCredential : Credential {
  [SameObject] readonly attribute ArrayBuffer              rawId;
  [SameObject] readonly attribute AuthenticatorResponse    response;
  [SameObject] readonly attribute AuthenticationExtensions clientExtensionResults;
};
```
..and contained the `clientExtensionResults`. but we can't do that due to webidl reasons, [so we have changed it to be](https://w3c.github.io/webauthn/#iface-pkcredential) (via PR #633)..
```
[SecureContext, Exposed=Window]
interface PublicKeyCredential : Credential {
    [SameObject] readonly attribute ArrayBuffer              rawId;
    [SameObject] readonly attribute AuthenticatorResponse    response;
    AuthenticationExtensions getClientExtensionResults();
};
```
However, in trying to re-craft the generation of a `PublicKeyCredential` object in the context of PR #498, we cannot invoke `getClientExtensionResults()` to get the client extension results because we're building the `PublicKeyCredential` object and `getClientExtensionResults()` is part of said object. 

I presume we need to obtain the the [client extension output value](https://w3c.github.io/webauthn/#client-extension-output) and stash it somewhere such that it can be returned to RP script when said RP script invokes `getClientExtensionResults()` ?

To do that, would we need to define an internal slot in `PublicKeyCredential` which contains ref to an `arraybuffer` associated with the |global| which contains the bytes of the result of invoking any requested client extensions, and then invocation of `getClientExtensionResults()` returns (a reference to?) the `arraybuffer`?

equalsJeffH:
This fixes #657: where does return value of PublicKeyCredential.getClientExtensionResults() come from?

***
[Preview](https://s3.amazonaws.com/pr-preview/w3c/webauthn/jeffh-fixup-algs-contd-5.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/webauthn/e9a9c33...5ea8a28.html)

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -506,8 +506,9 @@ that are returned to the caller when a new credential is created, or a new asser
         will be an {{AuthenticatorAssertionResponse}}.
 
     :   {{PublicKeyCredential/getClientExtensionResults()}}
-    ::  This operation returns a [=map=] containing [=extension identifier=] → [=client extension output=] entries
-        produced by the extension's [=client extension processing=].
+    ::  This operation returns the value of a {{PublicKeyCredential/[[clientExtensionsResults]]}}, which is a [=map=] containing
+        [=extension identifier=] → [=client extension output=] entries produced by the extension's
+        [=client extension processing=].
 
     :   <dfn>\[[type]]</dfn>
     ::  The {{PublicKeyCredential}} [=interface object=]'s {{Credential/[[type]]}} [=internal slot=]'s value is the string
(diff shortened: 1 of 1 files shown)


jcjones:
Approved to merge on call. 25 Oct.

PR #664, opened 2017-10-26
Update the RP operation section to be "not normative"

AngeloKai:
Unfortunately no matter how much we'd like web developers to write their codes in exactly the way we want, we aren't gonna get that. As a web developer myself, I can ignore the section however I want. It's not up to the WG to define normative action for the RP. Keeping this in by CR would just make us look foolish.

***
[Preview](https://s3.amazonaws.com/pr-preview/AngeloKai/webauthn/informative.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/webauthn/bcb5c11...AngeloKai:230e3da.html)


AngeloKai:
Setting this to CR milestone because I don't want to waste people's time discussing this before WD-07.

jcjones:
Seems reasonable to me.

index.bs
```diff
@@ -199,8 +199,9 @@ A variety of additional use cases and configurations are also possible, includin
 
 # Conformance # {#conformance}
 
-This specification defines three conformance classes. Each of these classes is specified so that conforming members of the class
-are secure against non-conforming or hostile members of the other classes.
+This specification defines two conformance classes. Each of these classes is specified so that conforming members of the class
+are secure against non-conforming or hostile members of the other classes. The specification recommends a set of 
+[[#rp-operations]] that [=[RP]=] SHOULD perform to get the security benefits offered by this specification.
```
jyasskin: I don't think we should use "SHOULD" about a group that's not a conformance class.

jyasskin: I don't think we should use "SHOULD" about a group that's not a conformance class.

jyasskin: "[=[RPS]=]", right? See https://s3.amazonaws.com/pr-preview/AngeloKai/webauthn/informative.html#conformance to check the final text.

rlin1: what about "should"?

rlin1:
We already have defined RP as one of the 3 conformance classes.
What else should we do in order to addres this issue?

equalsJeffH:
Although I think I understand the @AngeloKai's motivation for proposing these changes, I agree with @jyasskin -- this spec defines a cryptographic protocol (as well as a Web API (and more!)) and the RP operations are simply defining what the RP must do in order to faithfully participate in the protocol, thus gaining the security benefits of doing so. It is common practice in _protocol_ specs to stipulate the behavior of both endpoints. I'm sorry, but I do not think we should make the changes proposed by this PR.

nadalin:
@AngeloKai can you please look at this

nadalin:
@AngeloKai can we just close this no action ?

agl:
Closing this based on discussion during the call today. We can't actually do anything about sites which don't implement what we say, but we can be clear about what we expect them to do, at least.

Issue #660, opened 2017-10-25
Define sensible limits for User and RP Entity to be stored on Authenticator as part of create credential


akshayku:
Currently, we have not defined limits to individual fields for User and RP entity to be stored on Authenticator apart from `User.id` which can be maximum of 64 bytes. We should define a scheme which can work for both constrained and non-constrained authenticators for a predictable ecosystem.

Here is one proposal:

**User Entity:**
- Id: Authenticators MUST support maximum of 64 bytes.
- Name: Authenticators must support minimum 64 UTF8 characters. If RP provides more than 64 UTF8 characters, authenticator optionally can **truncate** how much it wants to store. 
- DisplayName: Authenticators MUST support minimum of 64 UTF8 characters. If RP provides more than 64 UTF8 characters, authenticator optionally can **truncate** how much it wants to store.
- Icon: Authenticators MUST support minimum of 128 UTF8 characters. If RP provides more than 128 UTF8 characters, authenticator optionally can **drop** this field all together. Here truncation does not makes sense.

**RP Entity:**
- Id: Authenticators can use HASH of this field as a primary index inside their database. So limit is not really needed. For completeness, we can say maximum 256 UTF characters.
- Name: Authenticators must support minimum 64 UTF8 characters. If RP provides more than 64 UTF8 characters, authenticator optionally can **truncate** how much it wants to store. 
- Icon: Authenticators MUST support minimum of 128 UTF8 characters. If RP provides more than 128 UTF8 characters, authenticator optionally can **drop** this field all together. Here truncation does not makes sense.

jovasco:
@akshayku I have a problem with the minimum of 128 bytes for icon. Especially for cases where there is no local storage, this means at least 129 bytes extra in the key handle. I believe the authenticator should be allowed to drop this field.

As a side issue for id, this becomes "MUST support 64 bytes.".

jyasskin:
A "minimum of 64 UTF8 characters" is a little odd. Do you mean 64 bytes (64 code units) or 64 unicode characters (roughly code points), which means up to 256 utf-8 bytes? (I don't have an opinion here, just want the spec to be precise.)

akshayku:
In CTAP spec, Authenticators only support UTF8 characters for these fields and that is why it is written as UTF8 characters. Defining Unicode characters limits is not needed here as platform will do the translation. The limits are for authenticators which will store these fields.

jyasskin:
[...]

PR #623, opened 2017-10-10
Describe how authenticators unique and find credential sources.

jyasskin:
This PR builds on #620. Just look at the last commit (50d8d1b) to see what's new here. @leshi, @rlin1, @akshayku, you were cc'ed on https://github.com/w3c/webauthn/pull/602#discussion_r142500439, which this patch implements.

This happens to fix a maybe-bug where the authenticator didn't check that a
decrypted credential ID came from the right RP.

It's also much more precise about the distinction between a credential
descriptor and a credential or credential source.

This still only improves #578, but it makes it clearer where we'll put that fix.

***
<a href="https://pr-preview.s3.amazonaws.com/jyasskin/webauthn/pull/623.html" title="Last updated on Feb 2, 2018, 3:30 PM GMT (14335b1)">Preview</a> (<a href="https://pr-preview.s3.amazonaws.com/jyasskin/webauthn/pull/623.html#authenticator-credentials-map" title="#authenticator-credentials-map">#authenticato…</a>) (<a href="https://pr-preview.s3.amazonaws.com/jyasskin/webauthn/pull/623.html#op-make-cred" title="#op-make-cred">#op-make-cred</a>) (<a href="https://pr-preview.s3.amazonaws.com/jyasskin/webauthn/pull/623.html#op-get-assertion" title="#op-get-assertion">#op-get-asser…</a>) | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/623/1fc8906...jyasskin:14335b1.html" title="Last updated on Feb 2, 2018, 3:30 PM GMT (14335b1)">Diff</a>

PR #667, opened 2017-11-01
Sensible limits for RP and User Entity fields.

akshayku:
***
[Preview](https://s3.amazonaws.com/pr-preview/w3c/webauthn/akshayku-user-rp-entity-limits.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/webauthn/bcb5c11...7f1068f.html)

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1274,15 +1274,15 @@ associated.
 <div dfn-type="dict-member" dfn-for="PublicKeyCredentialEntity">
     :   <dfn>name</dfn>
     ::  A human-friendly identifier for the entity. For example, this could be a company name for a [=[RP]=], or a
-        user's name. This identifier is intended for display. [=Authenticators=] MUST support a 64 byte minimum length
-        for a [=PublicKeyCredentialEntity/name=] members's value. Authenticators MAY truncate a
-        [=PublicKeyCredentialEntity/name=] member's value to a length equal to or greater than 64 bytes.
+        user's name. This identifier is intended for display. [=Authenticators=] MUST accept and store a 64 byte minimum length
+        for a name members's value. Authenticators MAY truncate a
+        name member's value to a length equal to or greater than 64 bytes.
 
     :   <dfn>icon</dfn>
     ::  A [=URL serializer|serialized=] URL which resolves to an image associated with the entity. For example, this could be
         a user's avatar or a [=[RP]=]'s logo. This URL MUST be an [=a priori authenticated URL=]. [=Authenticators=] MUST
-        support a 128 byte minimum length for a [=PublicKeyCredentialEntity/icon=] members's value.
-        Authenticators MAY ignore a [=PublicKeyCredentialEntity/icon=] members's value if its length is greater than 128 byes.
+        accept and store a 128 byte minimum length for a icon members's value.
+        Authenticators MAY ignore a icon members's value if its length is greater than 128 byes.
 </div>
 
 
@@ -1319,9 +1319,9 @@ credential.
     ::  The [=user handle=] of the user account entity.
 
     :   <dfn>displayName</dfn>
-    ::  A friendly name for the user account (e.g., "John P. Smith"). [=Authenticators=] MUST support a 64 byte minimum
-    length for a [=PublicKeyCredentialEntity/displayName=] members's value. Authenticators MAY truncate a
-    [=PublicKeyCredentialEntity/displayName=] member's value to a length equal to or greater than 64 bytes.
+    ::  A friendly name for the user account (e.g., "John P. Smith"). [=Authenticators=] MUST accept and store a 64 byte minimum
+        length for a displayName members's value. Authenticators MAY truncate a
+        displayName member's value to a length equal to or greater than 64 bytes.
 </div>
 
 
--- a/index.bs
+++ b/index.bs
@@ -1274,14 +1274,15 @@ associated.
 <div dfn-type="dict-member" dfn-for="PublicKeyCredentialEntity">
     :   <dfn>name</dfn>
     ::  A human-friendly identifier for the entity. For example, this could be a company name for a [=[RP]=], or a
-        user's name. This identifier is intended for display. Authenticators while creating a credential MUST support
-        minimum of 64 bytes for this field and optionally can truncate how much it wants to store if this field is more than 64 bytes.
+        user's name. This identifier is intended for display. [=Authenticators=] MUST support a 64 byte minimum length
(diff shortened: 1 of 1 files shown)


akshayku:
PR for #660

emlun:
Is this meant for resident credentials only?

equalsJeffH:
@emlun:
> resident credentials

I presume you are referring to [Client-side-resident Credential Private Keys](https://w3c.github.io/webauthn/#client-side-resident-credential-private-key) ?

index.bs
```diff
@@ -1274,11 +1274,14 @@ associated.
 <div dfn-type="dict-member" dfn-for="PublicKeyCredentialEntity">
     :   <dfn>name</dfn>
     ::  A human-friendly identifier for the entity. For example, this could be a company name for a [=[RP]=], or a
-        user's name. This identifier is intended for display.
+        user's name. This identifier is intended for display. Authenticators while creating a credential MUST support
+        minimum of 64 bytes for this field and optionally can truncate how much it wants to store if this field is more than 64 bytes.
```
equalsJeffH: suggested re-write:
[=Authenticators=] MUST support a 64 byte minimum length for a [=PublicKeyCredentialEntity/name=] members's value. Authenticators MAY truncate a [=PublicKeyCredentialEntity/name=] member's value to a length equal to or greater than 64 bytes.

index.bs
```diff
@@ -1274,11 +1274,14 @@ associated.
...
 
     :   <dfn>icon</dfn>
     ::  A [=URL serializer|serialized=] URL which resolves to an image associated with the entity. For example, this could be
-        a user's avatar or a [=[RP]=]'s logo. This URL MUST be an [=a priori authenticated URL=].
+        a user's avatar or a [=[RP]=]'s logo. This URL MUST be an [=a priori authenticated URL=]. Authenticators while
+        creating a credential MUST support minimum of 128 bytes for this field and optionally can drop this field if
+        this field is more than 128 bytes.
```
equalsJeffH: suggested re-write:
[=Authenticators=] MUST support a 128 byte minimum length for a [=PublicKeyCredentialEntity/icon=] members's value. Authenticators MAY ignore a [=PublicKeyCredentialEntity/icon=] members's value if its length is greater than 128 byes.

index.bs
```diff
@@ -1315,7 +1318,8 @@ credential.
     ::  The [=user handle=] of the user account entity.
 
     :   <dfn>displayName</dfn>
-    ::  A friendly name for the user account (e.g., "John P. Smith").
+    ::  A friendly name for the user account (e.g., "John P. Smith"). Authenticators while creating a credential MUST support
+        minimum of 64 bytes for this field and optionally can truncate how much it wants to store if this field is more than 64 bytes.
```
equalsJeffH: suggested re-write:
[=Authenticators=] MUST support a 64 byte minimum length for a [=PublicKeyCredentialEntity/displayName=] members's value. Authenticators MAY truncate a [=PublicKeyCredentialEntity/displayName=] member's value to a length equal to or greater than 64 bytes.

emlun:
@equalsJeffH Yes, that's what I meant. The presumption that the values will be stored make sense only in that context, right?

Actually there's no mention at all in #op-make-cred about storing any of the `userEntity` members except `id`. Is that a problem?

equalsJeffH:
review submitted, changes requested: https://github.com/w3c/webauthn/pull/667#pullrequestreview-73653135

emlun:
Sorry, I didn't mean to imply that a length limit is nonsensical for RP-stored credentials. My thinking was that the MUST could seem to clash with if the authenticator won't store the values at all - like U2F devices, do they "support" 64 bytes by not storing the value?

Anyway, #623 looks like it addresses most of my above thoughts.

akshayku:
@equalsJeffH : Can you review again? 
@emlun: U2F is different and they don't get these fields to be stored in U2F_REGISTER. IMO, We are fine with this wording

index.bs
```diff
@@ -1274,11 +1274,15 @@ associated.
 <div dfn-type="dict-member" dfn-for="PublicKeyCredentialEntity">
     :   <dfn>name</dfn>
     ::  A human-friendly identifier for the entity. For example, this could be a company name for a [=[RP]=], or a
-        user's name. This identifier is intended for display.
+        user's name. This identifier is intended for display. [=Authenticators=] MUST support a 64 byte minimum length
+        for a [=PublicKeyCredentialEntity/name=] members's value. Authenticators MAY truncate a
+        [=PublicKeyCredentialEntity/name=] member's value to a length equal to or greater than 64 bytes.
```
jovasco: "support"  is open to interpretation. Are you requiring for example that the authenticator must be able to display 64 bytes worth of characters?

akshayku: Displaying is not applicable for UI less authenticators on authenticator side but needed for multiple accounts scenario where Platform shows the UI. So Authenticators MUST "Store". I think we don't need to be explain every type of authenticators now or come in future? I believe "support" is better word. What do you suggest?

jovasco: If you mean accept and store, then write accept and store? It is what you need while "support"  covers every possible use of the string, including displaying.

equalsJeffH:
I approved these revised changes from my non-authnr-vendor-perspective, as they appear to be a step in the right direction. 

However, I offer the following observations/caveats:
1.  @jovasco has a good point over in https://github.com/w3c/webauthn/issues/660#issuecomment-341472667 that arbitrary truncation of these byte strings may result in mal-formed unicode encodings. We may want to Note: this.
1.  nominally, [DOMStrings contain UTF-16 code units, though that it not required](https://heycam.github.io/webidl/#idl-DOMString).  We should probably declare which strings we are expecting to be UTF-8 encoded to explicitly be UTF-8 encoded.
1.  These constraints the PR adds are authenticator-specific and begs the question of having a specific authenticator implementation considerations section rather than scattering such things throughout the spec (e.g., in the portion of the spec aimed at the browser developer audience.  
1.  An authenticator implementation considerations section would be the appropriate place to provide a more full definition of "support" ([which @jovasco is asking about above](https://github.com/w3c/webauthn/pull/667#pullrequestreview-73851978))

These items perhaps should be turned into yet more issues :)

equalsJeffH:
btw, I observed several bikeshed errors when compiling index.bs -- we oughta clean those up before any merging.

index.bs
akshayku: Alright..  I changed it to "accept and store".

akshayku:
Incorporated feedback and merging.

Issue #645, opened 2017-10-13
clarify "authenticator model": RPs may perform feature-based authenticator selection


equalsJeffH:
clarify "authenticator model": RPs may perform feature-based authenticator selection with both 
1. [`navigator.credentials.create()`](https://w3c.github.io/webappsec-credential-management/#dom-credentialscontainer-create) (which invokes [webauthn's #createCredential](https://w3c.github.io/webauthn/#createCredential)), and 
2. [`navigator.credentials.get()`](https://w3c.github.io/webappsec-credential-management/#dom-credentialscontainer-get) (which invokes [webauthn's #getAssertion](https://w3c.github.io/webauthn/#getAssertion)).

In (1), options stipulating "require resident key" and "require user verification" are expected to be passed (see also #536, #524), e.g., ultimately to [CTAP's authenticatorMakeCredential command](https://fidoalliance.org/specs/fido-v2.0-rd-20170927/fido-client-to-authenticator-protocol-v2.0-rd-20170927.html#authenticatorMakeCredential).

In (2), options stipulating "require user verification", "require user presence test"  are expected to be passed (see also #629, #524), e.g., ultimately to [CTAP's authenticatorGetAssertion command](https://fidoalliance.org/specs/fido-v2.0-rd-20170927/fido-client-to-authenticator-protocol-v2.0-rd-20170927.html#authenticatorGetAssertion).

emlun:
>In (2), options stipulating "require user verification", "require user presence test" are expected to be passed

Current language specifies to always perform a user presence test (unless user verification is performed), so the latter option does not currently exist (as discussed in https://github.com/w3c/webauthn/issues/629#issuecomment-336574658).

selfissued:
This is not actionable without a PR.  @equalsJeffH - do you plan to create one soon - ideally during this week's FIDO plenary?

equalsJeffH:
yes i plan to "create one soon" but meeting(s) I'm participating in have priority.

equalsJeffH:
fixed by PR #645

Issue #629, opened 2017-10-11
credentials.get() should have optional parameters for userVerification and userPresence


balfanz:
When requesting an assertion from an Authenticator in `credentials.get()`, the RP should be able to specify whether a test of user presence is required, whether user verification is required, or whether neither is required (a "silent signature").

Note: [`credentials.get()`](https://w3c.github.io/webappsec-credential-management/#dom-credentialscontainer-get), in the webauthn case, is implemented by the algorithm given by the [#getAssertion](https://w3c.github.io/webauthn/#getAssertion) section in the webauthn spec.

By default (i.e., when not mentioning either user verification or test of user presence), it should be assumed that test of user presence is required.

Proposal 1:
have a parameter `userPresence` with values `test`, `verify`, `none`, with default value `test`.

Proposal 2:
have two parameters:
- boolean parameter `requireUserVerification`, default value `false`
- boolean parameter `requireUserPresence`, default value `true`

Note that in this case, the combination `requireUserVerification = true` and `requireUserPresence = false` is invalid.

akshayku:
Proposal 2 looks better to me.

akshayku:
Thinking more about this, doesn't silent signatures problematic as they leak information without user knowing? 

How does one make sure that, user has given consent the first time and then it can be used at later stage without user presence but no silent signatures without user consent at some earlier time?

emlun:
The authenticator signs over the user presence and user verification flags, so it won't be possible to assert later that a silent signature was made with user consent. But the RP does need to take care to actually verify those flags (or store the signature), so that should probably be added to [§6.2. Verifying an authentication assertion][verify].

What would be the use case for silent signatures? I suppose they can't do much harm assuming the RP is trusted, but could XSS/XSRF attacks make it an issue?

[verify]: https://w3c.github.io/webauthn/#verifying-assertion

idamlaj:
Enabling silent authentication might be problematic from a privacy standpoint.

Ad trackers will be able to silently create and use "super cookies" that track the machine across apps, and can't be easily deleted by the user.

As for making user verificatIon optional, can you provide a use case for this? The user agent can already do this at the CTAP level which enables some use cases relevant to the web (e.g. "I forgot my CTAP device PIN" button when the user agent asks for PIN entry)

equalsJeffH:
fyi, silent authenticators aka "touchless" are the topic of this existing issue: https://github.com/w3c/webauthn/issues/199

[...]

Issue #524, opened 2017-08-08
Not necessary to pass AuthenticatorSelectionCriteria members to authenticatorMakeCredential()


equalsJeffH:
In reviewing the present state of [webauthn](http://w3c.github.io/webauthn/) in the context of reviewing some CTAP changes, I (finally, sigh) noticed that we are passing `options.authenticatorSelection.rk` and `options.authenticatorSelection.uv` to   authenticatorMakeCredential(). 

This is unnecessary [IIUC](http://onlineslangdictionary.com/meaning-definition-of/iiuc).  Note that `options.authenticatorSelection` members are used in step 21 of  [#createCredential](http://w3c.github.io/webauthn/#createCredential) in order to select authenticators having the desired characteristics.  It is subsequently not necessary to inform these authnrs to behave in a certain way, because they only behave in whatever their stated behavior fashion is (and as is reported in some platform-specific manner to the webauthn client). For example, this is how UAF authenticators behave (and essentially how authnr selection in UAF works). 

_Unless_ the authnr model postulated by webauthn is different from U2F and UAF in that authnrs may alter their key storage and user verification behavior on a per-requested credential basis?  (I am hoping not).

If the answer to the above question is "authnr model not different from U2F & UAF in these aspects", then the resoluition of this issue is to craft a PR which obviates passing  `options.authenticatorSelection.rk` and `options.authenticatorSelection.uv` to   authenticatorMakeCredential(). 

Note that a similar PR will need to be crafted for the CTAP spec.

equalsJeffH:
fwiw, passing options.authenticatorSelection.rk and options.authenticatorSelection.uv to authenticatorMakeCredential() was added in these two recently-merged PRs:

https://github.com/w3c/webauthn/pull/502 add requireResidentKey param to the invocation step of authenticatorMakeCredential 

https://github.com/w3c/webauthn/pull/460 Adds requireUserVerification option in AuthenticatorSelectionCriteria

selfissued:
They are passed dynamically to allow authenticators, for instance, to support both resident keys and non-resident keys.  Likewise, it's fine for authenticators supporting user verification to use this functionality to only provide user presence, when UV isn't needed.

I understand your point, but particularly for software authenticators, I suspect that many will be able to tailor their functionality to that requested.  Let's experiment with it this way for now, and if the authenticator vendors say that you're right, I'm fine pulling this later.

nadalin:
@equalsJeffH to review and commnet

[...]

Issue #199, opened 2016-09-14
WebAuthn available to Workers?  aka "silent authentication"


equalsJeffH:
As I recall, we've verbally discussed the question of whether the WebAuthn API ought to be available to {service, web}workers (aka "Workers") but we do not have an issue tracking it as yet AFAICT, so here it is. 

The default argument, as I recall, against worker use of WebAuthn is a UX one: authenticators, upon receiving a makeCredential() or getAssertion() call will signal to the user in some fashion, and since workers are asynchronous tasks, this may seem to the user like coming out of the blue and be confusing. 

Though, there are a couple of authenticator-behavior notions that do not always involve annunciation to the user (see also [1]):
1. for an already-registered authenticator, and an already-authenticated session, the RP may wish to periodically/asynchronously make a getAssertion() request in order to ascertain the continued presence of the same device+authenticator+privateKey as was present at the beginning of the session, and to do so, using subsequent getAssertion() requests without user interaction (because that can become tiresome and distracting to the user).   There may be a timeout associated with the session which upon expiry causes appropriate user interaction upon a subsequent getAssertion() request.  Let's term this "silent authentication".
2. for a not-yet-registered so-called "silent authenticator" (the authnr has no provision for user interaction itself), and a session already authenticated  (either via webauthn or legacy means), the RP can send a makeCredential (aka register) request, the user is prompted (by the RP or platform) for consent and informed of that the nature of this authnr is to not (itself) prompt the user for registration or authentication events. If the user gives consent, the authnr is registered, and subsequently responds to getAssertion() requests from the RP without user annunciation. 

So, if there are use cases where an RP's webapp wishes to employ workers, say, to do some sort of background processing, and that RP wishes to employ [periodic|asynchronous] getAssertion() requests without user interaction (during some allowed time interval perhaps) to aid assurance of overall session context continuity, being able to employ (1) above may be useful.  

[1] See also the "silent authenticator" notion:
from https://fidoalliance.org/specs/fido-uaf-v1.0-ps-20141208/fido-uaf-asm-api-v1.0-ps-20141208.html#security-and-privacy-guidelines:

> - ASMs SHOULD ensure that applications cannot use silent authenticators for tracking purposes. ASMs implementing support for a silent authenticator MUST show, during every registration, a user interface which explains what a silent authenticator is, asking for the users consent for the registration. Also, it is RECOMMENDED that ASMs designed to support roaming silent authenticators either
>   - Run with a special permission/privilege on the system, or
[...]

PR #672, opened 2017-11-06
Fix #644: Add UV parameter to getAssertion

### LINKED ISSUE #644: #getAssertion alg needs to pass authenticator selection requirements to authenticatorGetAssertion operation
Ostensibly, RPs want/need to pass authenticatorSelection.requireUserVerification (aka "uv" in [CTAP](https://fidoalliance.org/specs/fido-v2.0-rd-20170927/fido-client-to-authenticator-protocol-v2.0-rd-20170927.html#authenticatorGetAssertion) parlance) in the [#getAssertion](https://w3c.github.io/webauthn/#getAssertion) call, per issue #629.

However, [authenticatorSelectionCriteria](https://w3c.github.io/webauthn/#authenticatorSelection) is part of [MakePublicKeyCredentialOptions](https://w3c.github.io/webauthn/#dictionary-makecredentialoptions), **_which is not passed at all_** to [#getAssertion](https://w3c.github.io/webauthn/#getAssertion), and thus is not subsequently passed to [webauthn's authenticatorGetAssertion operation](https://w3c.github.io/webauthn/#op-get-assertion), which subsequently needs to pass both "uv" and "up" booleans as options to, for example, [CTAP's authenticatorGetAssertion command](https://fidoalliance.org/specs/fido-v2.0-rd-20170927/fido-client-to-authenticator-protocol-v2.0-rd-20170927.html#authenticatorGetAssertion).

See also issue #629

emlun:
This resolves #644, and possibly #524 and #645.

Summary:

- ~~The definition of "user present" now states that if the user is "verified" they are also "present". Background: #629~~ *Reverted after review*
- A UV parameter has been added to `PublicKeyCredentialRequestOptions`.
- A Boolean UV parameter is now passed to `authenticatorGetAssertion()`.
- The type of the UV parameter of the client operations has changed from Boolean to `UserVerificationRequirement`.
  - Background: #644, #629.
  - The enum type `UserVerificationRequirement` has been added, with values `required`, ~~`wanted`~~ `preferred` and ~~`not-wanted`~~ `discouraged`.
  - The type of the member `requireUserVerification` of `AuthenticatorSelectionCriteria` has changed from Boolean to `UserVerificationRequirement`.
  - A member `requireUserVerification` of type `UserVerificationRequirement` has been added to `PublicKeyCredentialRequestOptions`.
  - This value is used by the client operations to compute a Boolean value to pass to the authenticator operations.
- ~~Authenticator operations are now specified to always require UP or UV.~~ This was already stated in prose in [§5.1.4 Use an existing credential to make an assertion][get-assertion], but not in the algorithms:

  >Since this specification requires an authorization gesture to create any credentials, [...]

  Silent authenticator operations has also raised objections (#629, #644), so we should state this requirement clearly until it's decided to support silent operations. This probably breaks compatibility with UAF Silent Authenticators (see https://github.com/w3c/webauthn/issues/199#issuecomment-341576776).

  *EDIT: This logic has moved to the client level instead.*
- ~~`authenticatorMakeCredential` now always uses UV if the authenticator supports it. If `requireUserVerification` is `true` the client will only send the command to capable authenticators.~~ *Reverted after review*
- Authenticator operations' UP/UV behaviour has been aligned to agree with CTAP2.
  - A Boolean UP parameter is now passed to `authenticatorMakeCredential()` and `authenticatorGetAssertion()`. The client always sets this to the inverse of the UV parameter.
- CTAP compatibility should not be affected by any of the above. If it is, that's a bug.

[get-assertion]: https://w3c.github.io/webauthn/#getAssertion

***
[Preview](https://s3.amazonaws.com/pr-preview/w3c/webauthn/issue-644.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/webauthn/e09e0c3...05f4b23.html)

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -478,7 +478,7 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
 : <dfn id=concept-user-present>User Present</dfn>
 : <dfn>UP</dfn>
 :: Upon successful completion of a [=test of user presence|user presence test=], the user is said to be
-    "[=user present|present=]". If the user is [=user verified|verified=], the user is also said to be [=user present|present=].
+    "[=user present|present=]".
 
 : <dfn id=concept-user-verified>User Verified</dfn>
 : <dfn>UV</dfn>
@@ -1921,18 +1921,14 @@ When this operation is invoked, the authenticator must perform the following pro
     Private Key=], return an error code equivalent to "{{ConstraintError}}" and terminate the operation.
 1. If |requireUserVerification| is true and the authenticator cannot perform [=user verification=], return an error code
     equivalent to "{{ConstraintError}}" and terminate the operation.
+1. If |requireUserVerification| is true, perform [=user verification=]. If [=user verification=] fails, return an
+    error code equivalent to "{{NotAllowedError}}" and terminate the operation.
 1. Obtain [=user consent=] for creating a new credential. The prompt for obtaining this [=user consent|consent=] is shown by the
-    authenticator if it has its own output capability, or by the user agent otherwise. The prompt SHOULD display the contents of
+    authenticator if it has its own output capability, or by the user agent otherwise.  The prompt SHOULD display the contents of
     |rpEntity| and |userEntity|, if possible.
 
-    The method of obtaining [=user consent=] MUST include [=user verification=] or a [=test of user presence=] as follows:
-
-    - If |requireUserVerification| is true, [=user verification=] MUST be used.
-    - If |requireUserVerification| is false, [=user verification=] SHOULD NOT be used.
-    - If [=user verification=] is not used, a [=test of user presence=] MUST be used.
-
-    If the user denies [=user consent|consent=] or if [=user verification=] fails, return an error code equivalent to
-    "{{NotAllowedError}}" and terminate the operation.
+    If the user denies [=user consent|consent=], return an error code equivalent to "{{NotAllowedError}}" and terminate the
+    operation.
 
 1. Once [=user consent=] has been obtained, generate a new credential object:
     1. Let (|publicKey|,|privateKey|) be a new set of cryptographic keys using the combination of {{PublicKeyCredentialType}}
@@ -2000,6 +1996,8 @@ When this method is invoked, the [=authenticator=] must perform the following pr
     operation.
 1. If |requireUserVerification| is true and the authenticator cannot perform [=user verification=], return an error code
     equivalent to "{{ConstraintError}}" and terminate the operation.
+1. If |requireUserVerification| is true, perform [=user verification=]. If [=user verification=] fails, return an error code
+    equivalent to "{{NotAllowedError}}" and terminate the operation.
 
 1. Prompt the user to select a [=public key credential|credential=] |selectedCredential| from |allowCredentialDescriptorList|.
 
@@ -2007,14 +2005,8 @@ When this method is invoked, the [=authenticator=] must perform the following pr
     the [=authenticator=] if it has its own output capability, or by the user agent otherwise. The prompt SHOULD display the
     |rpId| and any additional data associated with |selectedCredential|, if possible.
 
-    The method of obtaining [=user consent=] MUST include [=user verification=] or a [=test of user presence=] as follows:
-
-    - If |requireUserVerification| is true, [=user verification=] MUST be used.
-    - If |requireUserVerification| is false, [=user verification=] SHOULD NOT be used.
-    - If [=user verification=] is not used, a [=test of user presence=] MUST be used.
-
-    If the user denies [=user consent|consent=] or if [=user verification=] fails, return an error code equivalent to
-    "{{NotAllowedError}}" and terminate the operation.
+    If the user denies [=user consent|consent=], return an error code equivalent to "{{NotAllowedError}}" and terminate the
+    operation.
 
 1. Let |processedExtensions| be the result of [=authenticator extension processing=] for each supported [=extension
     identifier=]/input pair in |extensions|.
--- a/index.bs
+++ b/index.bs
@@ -1466,30 +1466,6 @@ example of the latter, when the user is accessing the [=[RP]=] from a given clie
 use a [=roaming authenticator=] which was originally registered with the [=[RP]=] using a different client.
 
 
-### User Verification Requirement enumeration (enum <dfn enum>UserVerificationRequirement</dfn>) ### {#userVerificationRequirement}
-
-<pre class="idl">
-    enum UserVerificationRequirement {
-        "required",
-        "wanted",
-        "not-wanted"
-    };
-</pre>
-
-A [=[RP]=] may require [=user verified|user verification=] for some of its operations but not for others, and may use this type to
-express this level of requirement.
-
-The value {{UserVerificationRequirement/required}} indicates that the [=[RP]=] requires [=user verification=] for this operation
-and will fail the operation if the response does not have the [=UV=] [=flag=] set.
-
-The value {{UserVerificationRequirement/wanted}} indicates that the [=[RP]=] would like [=user verification=] for this operation
-is possible, but will not fail the operation if the response does not have the [=UV=] [=flag=] set.
-
-The value {{UserVerificationRequirement/not-wanted}} indicates that the [=[RP]=] does not require [=user verification=], and does
-not want the [=client=] or [=authenticator=] to ask for it, but will not fail the operation if the response has the [=UV=]
-[=flag=] set.
-
-
 ## Options for Assertion Generation (dictionary <dfn dictionary>PublicKeyCredentialRequestOptions</dfn>) ## {#assertion-options}
 
 The {{PublicKeyCredentialRequestOptions}} dictionary supplies {{CredentialsContainer/get()}} with the data it needs to generate
@@ -1554,6 +1530,30 @@ context.
 
 The [=public key credential=] type uses certain data structures that are specified in supporting specifications. These are as
 follows.
+
+
+### User Verification Requirement enumeration (enum <dfn enum>UserVerificationRequirement</dfn>) ### {#userVerificationRequirement}
+
+<pre class="idl">
+    enum UserVerificationRequirement {
+        "required",
+        "wanted",
+        "not-wanted"
+    };
+</pre>
+
+A [=[RP]=] may require [=user verified|user verification=] for some of its operations but not for others, and may use this type to
+express this level of requirement.
+
+The value {{UserVerificationRequirement/required}} indicates that the [=[RP]=] requires [=user verification=] for this operation
+and will fail the operation if the response does not have the [=UV=] [=flag=] set.
+
+The value {{UserVerificationRequirement/wanted}} indicates that the [=[RP]=] would like [=user verification=] for this operation
+is possible, but will not fail the operation if the response does not have the [=UV=] [=flag=] set.
+
+The value {{UserVerificationRequirement/not-wanted}} indicates that the [=[RP]=] does not require [=user verification=], and does
+not want the [=client=] or [=authenticator=] to ask for it, but will not fail the operation if the response has the [=UV=]
+[=flag=] set.
 
 
 ### Client data used in WebAuthn signatures (dictionary <dfn dictionary>CollectedClientData</dfn>) ### {#sec-client-data}
--- a/index.bs
+++ b/index.bs
@@ -2043,9 +2043,9 @@ When this method is invoked, the [=authenticator=] must perform the following pr
     equivalent to "{{UnknownError}}" and terminate the operation.
 1. If |allowCredentialDescriptorList| was not supplied, set it to a list of all credentials stored for |rpId| (as determined by
     an exact match of |rpId|).
-1. Remove any items from |allowCredentialDescriptorList| that do not match a credential bound to this authenticator. A match 
-    occurs if a credential matches <code>|rpId|</code> and an |allowCredentialDescriptorList| item's <code>|allowCredentialDescriptorList| 
-    {{PublicKeyCredentialDescriptor/id}}</code> and <code>|allowCredentialDescriptorList|{{PublicKeyCredentialDescriptor/type}}</code>. 
+1. Remove any items from |allowCredentialDescriptorList| that do not match a credential bound to this authenticator. A match
+    occurs if a credential matches <code>|rpId|</code> and an |allowCredentialDescriptorList| item's
+    {{PublicKeyCredentialDescriptor/id}} and {{PublicKeyCredentialDescriptor/type}} members.
 1. If |allowCredentialDescriptorList| is now empty, return an error code equivalent to "{{NotAllowedError}}" and terminate the
     operation.
 1. If |requireUserVerification| is `true` and the authenticator cannot perform [=user verification=], return an error code
--- a/index.bs
+++ b/index.bs
@@ -1931,11 +1931,14 @@ When this operation is invoked, the authenticator must perform the following pro
     authenticator if it has its own output capability, or by the user agent otherwise. The prompt SHOULD display the contents of
     |rpEntity| and |userEntity|, if possible.
 
-    - If |requireUserVerification| is true, obtain [=user consent=] by performing [=user verification=].
-    - If |requireUserVerification| is false, obtain [=user consent=] by performing a [=test of user presence=].
+    The method of obtaining [=user consent=] MUST include [=user verification=] or a [=test of user presence=] as follows:
 
-    If the user denies [=user consent|consent=], return an error code equivalent to "{{NotAllowedError}}" and terminate the
-    operation.
+    - If |requireUserVerification| is true, [=user verification=] MUST be used.
+    - If |requireUserVerification| is false, [=user verification=] SHOULD NOT be used.
+    - If [=user verification=] is not used, a [=test of user presence=] MUST be used.
+
+    If the user denies [=user consent|consent=] or if [=user verification=] fails, return an error code equivalent to
+    "{{NotAllowedError}}" and terminate the operation.
 
 1. Once [=user consent=] has been obtained, generate a new credential object:
     1. Let (|publicKey|,|privateKey|) be a new set of cryptographic keys using the combination of {{PublicKeyCredentialType}}
@@ -2010,11 +2013,14 @@ When this method is invoked, the [=authenticator=] must perform the following pr
     the [=authenticator=] if it has its own output capability, or by the user agent otherwise. The prompt SHOULD display the
     contents of |rpEntity| and |userEntity|, if possible.
 
-    - If |requireUserVerification| is true, obtain [=user consent=] by performing [=user verification=].
-    - If |requireUserVerification| is false, obtain [=user consent=] by performing a [=test of user presence=].
+    The method of obtaining [=user consent=] MUST include [=user verification=] or a [=test of user presence=] as follows:
 
-    If the user denies [=user consent|consent=], return an error code equivalent to "{{NotAllowedError}}" and terminate the
-    operation.
+    - If |requireUserVerification| is true, [=user verification=] MUST be used.
+    - If |requireUserVerification| is false, [=user verification=] SHOULD NOT be used.
+    - If [=user verification=] is not used, a [=test of user presence=] MUST be used.
+
+    If the user denies [=user consent|consent=] or if [=user verification=] fails, return an error code equivalent to
+    "{{NotAllowedError}}" and terminate the operation.
 
 1. Let |processedExtensions| be the result of [=authenticator extension processing=] for each supported [=extension
     identifier=]/input pair in |extensions|.
--- a/index.bs
+++ b/index.bs
@@ -752,7 +752,7 @@ When this method is invoked, the user agent MUST execute the following algorithm
 
     1. If {{AuthenticatorSelectionCriteria/authenticatorAttachment}} is [=present|present=] and its value is not equal
         to |authenticator|'s attachment modality, [=iteration/continue=].
-    1. If {{AuthenticatorSelectionCriteria/requireResidentKey}} is set to |true| and the |authenticator|
+    1. If {{AuthenticatorSelectionCriteria/requireResidentKey}} is set to true and the |authenticator|
         is not capable of storing a [=Client-Side-Resident Credential Private Key=], [=iteration/continue=].
     1. If {{AuthenticatorSelectionCriteria/requireUserVerification}} is set to |true| and the
         |authenticator| is not capable of performing [=user verification=], [=iteration/continue=].
@@ -969,6 +969,10 @@ method is invoked, the user agent MUST:
     Resolving this with good definitions or some other means will be addressed by resolving
     [Issue #613](https://github.com/w3c/webauthn/issues/613).
 
+    1. If |options|.{{PublicKeyCredentialRequestOptions/authenticatorSelection}} is [=present=], and
+        |options|.{{PublicKeyCredentialRequestOptions/authenticatorSelection}}.{{AuthenticatorSelectionCriteria/requireUserVerification}}
+        is set to true, and the |authenticator| is not capable of performing [=user verification=], [=iteration/continue=].
+
     1. Let |allowCredentialDescriptorList| be a new [=list=].
 
     1. If <code>|options|.{{PublicKeyCredentialRequestOptions/allowCredentials}}</code> [=list/is not empty=], execute a
@@ -1002,19 +1006,23 @@ method is invoked, the user agent MUST:
                             selection.
 
                             Then, using |transport|, invoke [=in parallel=] the [=authenticatorGetAssertion=] operation on
-                            |authenticator|, with |rpId|, |clientDataHash|, |allowCredentialDescriptorList|, and
-                            |authenticatorExtensions| as parameters.
+                            |authenticator|, with |rpId|, |clientDataHash|, |allowCredentialDescriptorList|,
+                            |options|.{{PublicKeyCredentialRequestOptions/authenticatorSelection}}.{{requireUserVerification}},
+                            and |authenticatorExtensions| as parameters.
 
                         :   [=list/is empty=]
                         ::  Using local configuration knowledge of the appropriate transport to use with |authenticator|,
                             invoke [=in parallel=] the [=authenticatorGetAssertion=] operation on |authenticator| with |rpId|,
-                            |clientDataHash|, |allowCredentialDescriptorList|, and |clientExtensions| as parameters.
+                            |clientDataHash|, |allowCredentialDescriptorList|,
+                            |options|.{{PublicKeyCredentialRequestOptions/authenticatorSelection}}.{{requireUserVerification}},
+                            and |clientExtensions| as parameters.
                     </dl>
 
             :   [=list/is empty=]
             ::  Using local configuration knowledge of the appropriate transport to use with |authenticator|, invoke
                 [=in parallel=] the [=authenticatorGetAssertion=] operation on |authenticator| with |rpId|, |clientDataHash|,
-                and |clientExtensions| as parameters.
+                |options|.{{PublicKeyCredentialRequestOptions/authenticatorSelection}}.{{requireUserVerification}}, and
+                |clientExtensions| as parameters.
 
                 Note: In this case, the [=[RP]=] did not supply a list of acceptable credential descriptors. Thus the
                     authenticator is being asked to exercise any credential it may possess that is bound to
@@ -1299,7 +1307,7 @@ optionally evidence of [=user consent=] to a specific transaction.
 
     :   <dfn>authenticatorSelection</dfn>
     ::  This member is intended for use by [=[RPS]=] that wish to select the appropriate authenticators to participate in
-        the {{CredentialsContainer/create()}} or {{CredentialsContainer/get()}} operation.
+        the {{CredentialsContainer/create()}} operation.
 
     :   <dfn>extensions</dfn>
     ::  This member contains additional parameters requesting additional processing by the client and authenticator. For
@@ -1398,10 +1406,9 @@ attributes.
         [=Client-side-resident Credential Private Key=] when creating a [=public key credential=].
 
     :   <dfn>requireUserVerification</dfn>
-    ::  This member describes the [=[RPS]=]' requirements regarding the authenticator being capable of performing user verification.
-        If the parameter is set to true, the authenticator MUST perform user verification when performing
-        the {{CredentialsContainer/create()}} operation and future [[#getAssertion]] operations when it is
-        requested to verify the credential.
+    ::  This member describes the [=[RPS]=]' requirements regarding the authenticator being capable of performing user
+        verification. If the parameter is set to true, the authenticator MUST perform [=user verification=] when performing the
+        {{CredentialsContainer/create()}} or {{CredentialsContainer/get()}} operation.
 </div>
 
 
@@ -1451,6 +1458,7 @@ an assertion. Its {{PublicKeyCredentialRequestOptions/challenge}} member must be
         unsigned long                        timeout;
         USVString                            rpId;
         sequence<PublicKeyCredentialDescriptor> allowCredentials = [];
+        AuthenticatorSelectionCriteria       authenticatorSelection;
         AuthenticationExtensions             extensions;
     };
 </xmp>
@@ -1473,6 +1481,13 @@ an assertion. Its {{PublicKeyCredentialRequestOptions/challenge}} member must be
     ::  This optional member contains a list of {{PublicKeyCredentialDescriptor}} object representing [=public key credentials=]
         acceptable to the caller, in decending order of the caller's preference (the first item in the list is the most
         preferred credential, and so on down the list).
+
+    :   <dfn>authenticatorSelection</dfn>
+    ::  This member is intended for use by [=[RPS]=] that wish to select the appropriate authenticators to participate in
+        the {{CredentialsContainer/get()}} operation.
+
+        Note: Only the {{AuthenticatorSelectionCriteria/requireUserVerification}} member of this member currently has any effect
+        when used in {{PublicKeyCredentialRequestOptions}}; any other members are ignored.
 
     :   <dfn>extensions</dfn>
     ::  This optional member contains additional parameters requesting additional processing by the client and authenticator.
@@ -1866,9 +1881,9 @@ When this operation is invoked, the authenticator must perform the following pro
     If so, return an error code equivalent to "{{NotAllowedError}}" and terminate the operation.
 1. If |requireResidentKey| is |true| and the authenticator cannot store a [=Client-side-resident Credential
     Private Key=], return an error code equivalent to "{{ConstraintError}}" and terminate the operation.
-1. If |requireUserVerification| is |true| and the authenticator cannot perform user verification,
-    return an error code equivalent to "{{ConstraintError}}" and terminate the operation.
-1. Prompt the user for consent to create a new credential. The prompt for obtaining this consent is shown by the authenticator
+1. If |requireUserVerification| is true and the authenticator cannot perform [=user verification=], return an error code
+    equivalent to "{{ConstraintError}}" and terminate the operation.
+  1. Prompt the user for consent to create a new credential. The prompt for obtaining this consent is shown by the authenticator
     if it has its own output capability, or by the user agent otherwise. If the user denies consent, return an error code
     equivalent to "{{NotAllowedError}}" and terminate the operation. The Authenticator and user agent MAY skip this prompt
     if the Authenticator is a [=platform authenticator=] and |excludeCredentialDescriptorList| is empty.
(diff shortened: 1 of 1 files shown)


emlun:
@kieun may have something to say about how this affects UAF.

emlun:
I would also like to rename the `requireUserVerification` member to `userVerification`, but left it out for now to avoid unnecessary breakage.

Kieun:
This modification indeed affects UAF. But I'm not sure that how many number of UAF silent authenticators are deployed in the market. I would also think that at least user presence is needed during authentication in web space for privacy reason.
But for native application, it would be worth to provide silent authenticators for authentication since a lot of applications already tend to use automated login(authenticate) approach.

emlun:
By "native application", do you mean platform authenticators (i.e. TPMs integrated in laptops/phones)? There [was an exception for them][exception], but it was in `authenticatorMakeCredential` which has been agreed should for now not be allowed to be silent. Do you request a similar exception for platform authenticators in `authenticatorGetAssertion`?

[exception]: https://github.com/w3c/webauthn/pull/672/files#diff-ec9cfa5f3f35ec1f84feb2e59686c34dL1873

Kieun:
I just mean that "native application" is a platform specific RP application such as Android or iOS app that can leverage FIDO silent authenticators for automated login use case.

emlun:
Ok. That sounds like it falls outside the scope of the WebAuthn spec.

index.bs
```diff
@@ -1383,7 +1407,7 @@ attributes.
     dictionary AuthenticatorSelectionCriteria {
         AuthenticatorAttachment      authenticatorAttachment;
         boolean                      requireResidentKey = false;
-        boolean                      requireUserVerification = false;
+        UserVerificationRequirement  requireUserVerification = "not-wanted";
```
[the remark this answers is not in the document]
emlun: Oh right, my bad.

index.bs
```diff
@@ -1498,6 +1527,30 @@ The [=public key credential=] type uses certain data structures that are specifi
 follows.
 
 
+### User Verification Requirement enumeration (enum <dfn enum>UserVerificationRequirement</dfn>) ### {#userVerificationRequirement}
+
+<pre class="idl">
+    enum UserVerificationRequirement {
+        "required",
+        "wanted",
```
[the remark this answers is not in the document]
emlun: I like "preferred". How about "discouraged" for the last one?

I feel there's a value in having the options in order of the level of requirement. Ascending or descending doesn't matter, but I think "preferred" belongs in the middle since it's the middle level of requirement.

index.bs
```diff
@@ -478,7 +478,7 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
 : <dfn id=concept-user-present>User Present</dfn>
 : <dfn>UP</dfn>
 :: Upon successful completion of a [=test of user presence|user presence test=], the user is said to be
-    "[=user present|present=]".
+    "[=user present|present=]". If the user is [=user verified|verified=], the user is also said to be [=user present|present=].
```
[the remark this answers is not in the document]
emlun: Good point, thanks. I was thinking that a hardware authenticator that supports PIN verification would have its own input controls, but you're right - CTAP specifies a protocol for entering the PIN in the client, so this assumption does not hold.

index.bs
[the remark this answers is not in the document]
akshayku: I still like "Not required" better as discouraged is something prescriptive and looks like Platform has to choice while asking. "Not required" says that platform MUST not ask for it even if authenticator supports it. 

I will leave it up to you.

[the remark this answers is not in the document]
emlun: I think "discouraged" better communicates that the RP has an *active* opinion that the client should not ask for UV, while "not required" instead communicates that the RP has no specific opinion other than that UV is passively allowed but not required. "not preferred" also feels too passive to me, and "forbidden" is too strict since some authenticators may always require UV.

emlun:
@akshayku Thank you for reviewing!

equalsJeffH:
I did a merge-from-master locally in order to more easily review -- shall I push it to this branch -- it merged automagically, with no manual conflict resolution.

emlun:
@equalsJeffH Go ahead!

index.bs
```diff
@@ -1383,7 +1407,7 @@ attributes.
     dictionary AuthenticatorSelectionCriteria {
         AuthenticatorAttachment      authenticatorAttachment;
         boolean                      requireResidentKey = false;
-        boolean                      requireUserVerification = false;
+        UserVerificationRequirement  requireUserVerification = "preferred";
```
equalsJeffH: we may have an impedance mis-match here with the default value of "preferred".  CTAP has UV default to `false` on its `authenticatorMakeCredential` method.

index.bs
```diff
@@ -1451,6 +1474,7 @@ an assertion. Its {{PublicKeyCredentialRequestOptions/challenge}} member must be
         unsigned long                        timeout;
         USVString                            rpId;
         sequence<PublicKeyCredentialDescriptor> allowCredentials = [];
+        UserVerificationRequirement          requireUserVerification = "preferred";
```
equalsJeffH: we may have an impedance mis-match here with the default value of "preferred".  CTAP has UP default to `true` and UV default to `false` on its `authenticatorGetAssertion` method.

index.bs
```diff
@@ -1474,6 +1498,11 @@ an assertion. Its {{PublicKeyCredentialRequestOptions/challenge}} member must be
         acceptable to the caller, in decending order of the caller's preference (the first item in the list is the most
         preferred credential, and so on down the list).
 
+    :   <dfn>requireUserVerification</dfn>
+    ::  This member describes the [=[RP]=]'s requirements regarding [=user verification=] for this
```
equalsJeffH: s/for this/for the/

index.bs
```diff
@@ -1398,10 +1422,9 @@ attributes.
         [=Client-side-resident Credential Private Key=] when creating a [=public key credential=].
 
     :   <dfn>requireUserVerification</dfn>
-    ::  This member describes the [=[RPS]=]' requirements regarding the authenticator being capable of performing user verification.
-        If the parameter is set to true, the authenticator MUST perform user verification when performing
-        the {{CredentialsContainer/create()}} operation and future [[#getAssertion]] operations when it is
-        requested to verify the credential.
+    ::  This member describes the [=[RP]=]'s requirements regarding [=user verification=] for this
```
equalsJeffH: s/for this/for the/

index.bs
```diff
@@ -1498,6 +1527,30 @@ The [=public key credential=] type uses certain data structures that are specifi
...
+
+<pre class="idl">
+    enum UserVerificationRequirement {
+        "required",
+        "preferred",
+        "discouraged"
+    };
+</pre>
+
+A [=[RP]=] may require [=user verified|user verification=] for some of its operations but not for others, and may use this type to
```
equalsJeffH: s/ [=user verified|user verification=] / [=user verification=] /

index.bs
```diff
@@ -1498,6 +1527,30 @@ The [=public key credential=] type uses certain data structures that are specifi
...
+express this level of requirement.
+
+The value {{UserVerificationRequirement/required}} indicates that the [=[RP]=] requires [=user verification=] for this operation
+and will fail the operation if the response does not have the [=UV=] [=flag=] set.
+
+The value {{UserVerificationRequirement/preferred}} indicates that the [=[RP]=] would like [=user verification=] for this
+operation if possible, but will not fail the operation if the response does not have the [=UV=] [=flag=] set.
+
+The value {{UserVerificationRequirement/discouraged}} indicates that the [=[RP]=] does not require [=user verification=], and does
+not want the [=client=] or [=authenticator=] to ask for it, but will not fail the operation if the response has the [=UV=]
```
equalsJeffH: s/ask for/request/

nadalin:
@jcjones Can you please review

index.bs
```diff
@@ -1864,15 +1917,20 @@ When this operation is invoked, the authenticator must perform the following pro
...
+    equivalent to "{{ConstraintError}}" and terminate the operation.
+1. If |requireUserVerification| is true, perform [=user verification=]. If [=user verification=] fails, return an
+    error code equivalent to "{{NotAllowedError}}" and terminate the operation.
+1. Obtain [=user consent=] for creating a new credential. The prompt for obtaining this [=user consent|consent=] is shown by the
+    authenticator if it has its own output capability, or by the user agent otherwise.  The prompt SHOULD display the contents of
+    |rpEntity| and |userEntity|, if possible.
+
+    If the user denies [=user consent|consent=], return an error code equivalent to "{{NotAllowedError}}" and terminate the
+    operation.
+
```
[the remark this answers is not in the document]
emlun: So the previous discussion https://github.com/w3c/webauthn/pull/672#discussion_r149431060 is incorrect?

index.bs
emlun: So should we change this default back to `"discouraged"`?

index.bs
```diff
@@ -1934,10 +1994,20 @@ When this method is invoked, the [=authenticator=] must perform the following pr
...
-    Obtain [=user consent=] for using |selectedCredential|. The prompt for obtaining this [=user consent|consent=] may be shown
-    by the
-    [=authenticator=] if it has its own output capability, or by the user agent otherwise.
+
+1. Obtain [=user consent=] for using |selectedCredential|. The prompt for obtaining this [=user consent|consent=] may be shown by
+    the [=authenticator=] if it has its own output capability, or by the user agent otherwise. The prompt SHOULD display the
+    |rpId| and any additional data associated with |selectedCredential|, if possible.
+
+    If the user denies [=user consent|consent=], return an error code equivalent to "{{NotAllowedError}}" and terminate the
+    operation.
```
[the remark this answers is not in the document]
emlun: There's a subtle wrinkle here: In the case where `authenticatorGetAssertion()` is called with no `allowCredentialDescriptorList`, the authenticator must perform UV before giving out the list of contained resident credentials. Therefore the "Prompt the user to select a credential *selectedCredential* from *allowCredentialDescriptorList* must come after the step performing UV, while the "obtain consent" step currently comes after the selection prompt.

index.bs
akshayku: These two are different things. CTAP `UV` is defining what to interpret if that parameter is not present in the API. Here is what the behavior should be via the web and how to handle 3 different use case for getAssertion. Its `UserVerificationRequirement` and not `UV`. 

We don't have to match these things.

Default is authenticator behavior (`preferred`)

index.bs
```diff
@@ -963,6 +986,35 @@ method is invoked, the user agent MUST:
...
     [Issue #613](https://github.com/w3c/webauthn/issues/613).
 
+    1. If |options|.{{PublicKeyCredentialRequestOptions/requireUserVerification}} is set to
+        {{UserVerificationRequirement/required}} and the |authenticator| is not capable of performing [=user verification=],
+        [=iteration/continue=].
+
+    1. Let |userVerification| be the <dfn>effective user verification requirement for assertion</dfn>, a Boolean value, as
+        follows. If <code>|options|.{{AuthenticatorSelectionCriteria/requireUserVerification}}</code>
+
+            <dl class="switch">
```
akshayku: Below procedure is not correct. 

There are two ways user verification can be requested for getAssertion. One is via `UV` flag to the authenticator if authenticator says that it supports `UV` in authenticatorGetInfo. Second is via ClientPin (also indicated via authenticatorGetInfo ) semantics where platform does the ClientPin flow and then in the authenticatorGetAssertion request passes `pinAuth` parameter. That `pinAuth` parameter is equivalent of asking "user verification" from such devices.

index.bs
```diff
@@ -746,12 +746,36 @@ When this method is invoked, the user agent MUST execute the following algorithm
...
         is not capable of storing a [=Client-Side-Resident Credential Private Key=], [=iteration/continue=].
-    1. If {{AuthenticatorSelectionCriteria/requireUserVerification}} is set to |true| and the
+    1. If {{AuthenticatorSelectionCriteria/requireUserVerification}} is set to {{UserVerificationRequirement/required}} and the
         |authenticator| is not capable of performing [=user verification=], [=iteration/continue=].
 
-    1. Let |excludeCredentialDescriptorList| be a new [=list=].
+    1. Let |userVerification| be the <dfn>effective user verification requirement for credential creation</dfn>, a Boolean value,
+        as follows. If {{AuthenticatorSelectionCriteria/requireUserVerification}}
+
+            <dl class="switch">
```
akshayku: Actually, this procedure is not correct for creating a credential. Once a device has been set with UV option or ClientPin semantics, authenticator always asks for user verification while creating a credential. So there is no option to skip it. Skipping is available for authenticatorGetAssertion. 

So for creating a credential, from the web, it has to be a Boolean and not an enum.  Then if it `true` (meaning RP wants user verification), Platform can reject the call if authenticator does not support it via any of the authenticator's mechanism. In case of ClientPin semantics, if authenticator supports a PIN but it has not been set up yet, platform can first provision or set up a PIN and then try for authenticatorMakeCredential. 

If it is set to `false`, and authenticator is configured with ClientPin or in built `UV` mechanism, platform can reject the call as authenticator always requires user verification while creating a credential. For example, if ClientPin is setup for the authenticator, you have to provide PIN for creating a credential.

emlun: Good points. I don't agree that the platform should necessarily reject authenticators that require UV when the RP doesn't, but I've made the other changes. I left the PIN setup and entry unspecified as before.

index.bs
emlun: I don't think we should explicitly couple these steps to CTAP. I think it's probably better to specify that "UV should be used" but leave unspecified how it should be done as that will be different for each backend protocol. CTAP may have two ways to trigger UV, but they're both conceptually the same thing in the WebAuthn abstraction. As long as we're compatible with the CTAP procedures we should be good, right?

index.bs
emlun: So is this a wontfix, then?

index.bs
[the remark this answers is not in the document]
emlun: I think I managed to think something up for this, please review again!

index.bs
akshayku: Yes, IMO

index.bs
```diff
@@ -1498,14 +1526,43 @@ an assertion. Its {{PublicKeyCredentialRequestOptions/challenge}} member must be
...
+The value {{UserVerificationRequirement/required}} indicates that the [=[RP]=] requires [=user verification=] for this operation
+and will fail the operation if the response does not have the [=UV=] [=flag=] set.
+
+The value {{UserVerificationRequirement/preferred}} indicates that the [=[RP]=] would like [=user verification=] for this
+operation if possible, but will not fail the operation if the response does not have the [=UV=] [=flag=] set.
+
+The value {{UserVerificationRequirement/discouraged}} indicates that the [=[RP]=] does not require [=user verification=], and does
+not want the [=client=] or [=authenticator=] to request it, but will not fail the operation if the response has the [=UV=]
+[=flag=] set.
+
```
balfanz: Why is it important to signal to the Authenticator whether or not the RP will fail the operation?

It seems to me that we can get the same semantics with a boolean field:

- An RP who intents to fail the operation if the UV bit is not set in the response will set the userVerificationRequirement to `true`, and then fail the operation if the UV bit is missing in the response.

- An RP who prefers user verification, but won't fail the operation if user verification is not performed, will set the userVerificationRequirement to `true`, and then not fail the operation if the UV bit is missing in the response.

- An RP who prefers to *not* do user verification, but won't fail the operation if user verification performed, will set the userVerificationRequirement to `false`, and then not fail the operation if the UV bit is present in the response.

So I would recommend to make this a boolean. It covers all the use cases described here, uses fewer values, and makes it much easier to map these values to the boolean option we have for this in CTAP. 

I would also recommend to name this `userVerification`.

emlun: I agree with renaming to `userVerification`.

It's not important to signal to the authenticator, but it's useful to signal this to the client so it can select an appropriate authenticator if multiple ones are available.

- If `userVerification` is `"required"`, the client will send the command only to UV capable authenticators
- If `userVerification` is `"preferred"`, the client will send the command only to UV capable authenticators if any are available, and to any eligible authenticator otherwise
- If `userVerification` is `"discouraged"`, the client could refrain from sending the command to authenticators known to always require UV (i.e. first-factor-only authenticators).

If this were a Boolean parameter and `true` could mean either `"required"` or `"preferred"`, the client could fail the operation unnecessarily because it thinks no eligible authenticator is available.

balfanz: What's the benefit of doing it this way? Contrast this with the alternate approach:

- if `userVerification` is set to `true`, client will prioritize UV-capable Authenticators (send them the requests first, or hold responses from other Authenticators until it hears from an UV-capable Authenticator, etc.), but will accept responses from non-UV-capable Authenticators.

- if 'userVerification` is set to `false` (or missing), then the client will not prioritize according to UV capabilities.

As far as I can tell, the difference between this approach and the one you're describing is *when* the error occurs if

- the RP requires user verification, *and*
- there are non-UV capable Authenticators attached, *and*
- those non-UV-capable Authenticators manage credentials that are on the `allowCredentials` list (or there is no `allowCredentials` list), and would therefore process the request.

If all these conditions are met, then with the tri-state value, the client can fail the operation early, whereas in the boolean case, the RP will have to fail the operation. This will likely result in the user seeing an error a few hundred milliseconds later than they would in the tri-state case. 

Is that worth the added complexity?

emlun: It becomes more relevant now that we are moving to waiting for hot-plugged authenticators; it's more likely a question of tens of seconds rather than hundreds of milliseconds.

In the hot-plugging context it's likely that no authenticators will be available at the start, and sometime (delay on the order of seconds) within the timeout window an authenticator will appear. If the RP has set `requireUserVerification: true` and this first authenticator doesn't support UV, should the client wait X more seconds to see if a UV capable authenticator appears? How should you design the UX around this so users aren't confused and start over because "it hangs"?

With the tri-state approach there is no ambiguity: If the RP sets `userVerification: "required"` and the first authenticator doesn't support UV, the client can immediately tell the user to use another authenticator. If the RP sets `userVerification: "preferred"` the client knows it doesn't have to wait for a UV capable authenticator and can immediately go with whatever the user provides.

I also find it cognitively jarring that "required" wouldn't actually mean *required*, but I guess that's more of a subjective aesthetic issue.

All that said, you're right that it might not be worth the complexity if we assume that most users will only have one authenticator and likely won't ever find themselves in this situation. On the other hand this would save RPs the complexity of informing the user of this failure mode, since that would be handled by the browser instead. I think it's reasonable to make implementing an RP easier by pushing this to the browser, but then again I'm not the one who would be implementing it. :)

@AngeloKai @jcjones thoughts on this?

emlun: Also, addressing this:

>[a Boolean parameter instead] makes it much easier to map these values to the boolean option we have for this in CTAP.

The mapping is already suggested in the spec of `create()` and `getAssertion()`; does it need to be made clearer?

emlun:
Decided on TPAC face-to-face meeting to fix conflicts and merge this.

equalsJeffH:
I'm still reviewing...

equalsJeffH:
I have edits to propose, dunno if can do 'em today. will do tmrw morning PT at latest.

equalsJeffH:
thanks for working on this Emil -- it's another one of those gnarly ones it seems. this branch could use a merge-from-master which would aid review.   Ill enter my comments in any case.

emlun:
@equalsJeffH Done! I'll be awaiting your review.

index.bs
equalsJeffH: I think we need consensus from amongst implementors as to requireUserVerification's default value.  cc: @balfanz  @battre @jcjones  @AngeloKai

index.bs
```diff
@@ -1538,6 +1591,7 @@ an assertion. Its {{PublicKeyCredentialRequestOptions/challenge}} member must be
         unsigned long                        timeout;
         USVString                            rpId;
         sequence<PublicKeyCredentialDescriptor> allowCredentials = [];
+        UserVerificationRequirement          userVerification = "preferred";
```

index.bs
```diff
@@ -1738,6 +1798,31 @@ the {{CredentialsContainer/create()}} or {{CredentialsContainer/get()}} methods.
...
+
+<pre class="idl">
+    enum UserVerificationRequirement {
+        "required",
+        "preferred",
+        "discouraged"
+    };
+</pre>
+
+A [=[RP]=] may require [=user verification=] for some of its operations but not for others, and may use this type to express this
```
equalsJeffH: s/this level of requirement/its needs/

index.bs
```diff
@@ -1738,6 +1798,31 @@ the {{CredentialsContainer/create()}} or {{CredentialsContainer/get()}} methods.
...
+level of requirement.
+
+The value {{UserVerificationRequirement/required}} indicates that the [=[RP]=] requires [=user verification=] for this operation
```
equalsJeffH: nit: s/this/the/

index.bs
```diff
@@ -1738,6 +1798,31 @@ the {{CredentialsContainer/create()}} or {{CredentialsContainer/get()}} methods.
...
+and will fail the operation if the response does not have the [=UV=] [=flag=] set.
+
+The value {{UserVerificationRequirement/preferred}} indicates that the [=[RP]=] would like [=user verification=] for this
```
equalsJeffH: s/would like/prefers/
nit: s/this/the/

index.bs
```diff
@@ -1738,6 +1798,31 @@ the {{CredentialsContainer/create()}} or {{CredentialsContainer/get()}} methods.
...
+operation if possible, but will not fail the operation if the response does not have the [=UV=] [=flag=] set.
+
+The value {{UserVerificationRequirement/discouraged}} indicates that the [=[RP]=] does not require [=user verification=], and does
+not want the [=client=] or [=authenticator=] to request it, but will not fail the operation if the response has the [=UV=]
+[=flag=] set.
+
```
equalsJeffH: > The value {{UserVerificationRequirement/discouraged}} indicates that the [=[RP]=] does not require [=user verification=], and does
not want the [=client=] or [=authenticator=] to request it, 

suggest:
The value {{UserVerificationRequirement/discouraged}} indicates that the [=[RP]=] does not want [=user verification=] employed during the operation (e.g., in the interest of minimizing disruption to the user interaction flow),

index.bs
```diff
@@ -1983,15 +2068,30 @@ When this operation is invoked, the authenticator must perform the following pro
...
-1. Prompt the user for consent to create a new credential. The prompt for obtaining this consent is shown by the authenticator
-    if it has its own output capability, or by the user agent otherwise. If the user denies consent, return an error code
-    equivalent to "{{NotAllowedError}}" and terminate the operation. The Authenticator and user agent MAY skip this prompt
-    if the Authenticator is a [=platform authenticator=] and |excludeCredentialDescriptorList| is empty.
-1. Once user consent has been obtained, generate a new credential object:
+1. If |requireUserVerification| is `true` and the authenticator cannot perform [=user verification=], return an error code
+    equivalent to "{{ConstraintError}}" and terminate the operation.
+1. Obtain [=user consent=] for creating a new credential. The prompt for obtaining this [=user consent|consent=] is shown by the
+    authenticator if it has its own output capability, or by the user agent otherwise.  The prompt SHOULD display the contents of
+    |rpEntity| and |userEntity|, if possible.
```
equalsJeffH: we need to be more explicitregarding what is displayed here from amongst |rpEntity|'s and |userEntity|'s components, because e.g. userEntity.id aka userHandle is explicitly not human-meaningful nor palatable.

index.bs
```diff
@@ -2039,24 +2139,53 @@ input parameters:
...
 1. If |allowCredentialDescriptorList| is now empty, return an error code equivalent to "{{NotAllowedError}}" and terminate the
     operation.
+
+1. If |isFirstFactorOperation| is `true`, perform [=user verification=]. If [=user verification=] fails, return an error code
+    equivalent to "{{NotAllowedError}}" and terminate the operation.
+
+    Remove any items from |allowCredentialDescriptorList| that match a credential that was created by a [=user verified|verified
+    user=] other than the current [=user verified|verified user=].
+
 1. Prompt the user to select a [=public key credential|credential=] |selectedCredential| from |allowCredentialDescriptorList|.
```
equalsJeffH: prompting the user to select ought to occur prior to performing user verification, as currently specified, yes?

index.bs
```diff
@@ -2039,24 +2139,53 @@ input parameters:
...
-    Obtain [=user consent=] for using |selectedCredential|. The prompt for obtaining this [=user consent|consent=] may be shown
-    by the
-    [=authenticator=] if it has its own output capability, or by the user agent otherwise.
+
+1. Obtain [=user consent=] for using |selectedCredential|. The prompt for obtaining this [=user consent|consent=] may be shown by
```
equalsJeffH: The spec's original intent is that the act of user verification is the act of obtaining user consent, so splitting them apart here seems incorrect to me.

akshayku: In Client PIN case, user verification is not same as user presence check. User still has to do perform touch or equivalent. This looks correct to me.

index.bs
akshayku: IMO, Default should be to allow both kind of authenticators (one with UV and one without).

index.bs
emlun: Not for first factor authentications, i.e. when `allowCredentialDescriptorList` is not given. That would mean that an attacker could trivially list the contents of a stolen authenticator.

This does not apply for when `allowCredentialDescriptorList` is given, though, since in that case the attacker already knows the credential ID. The only new information the attacker could gain in that case is the user handle, which is intended to be unlikely to be duplicated between RPs, and it's likely the attacker already knows it if they know the credential ID.

index.bs
emlun: That was my intent as well, I've tried to formulate that UV and UP are interchangeable as user consent unless UV is specifically required.

My thinking in the case of `authenticatorGetAssertion` was also that there are two scenarios:

- This is a second factor operation (`allowCredentials` given); the user has not been verified before selecting a credential. In this case UV or UP can be done interchangeably in one step to confirm the selection.
- This is a first factor operation (`allowCredentials` not given); the user has already been verified before selecting a credential. However it's probably wise to collect user consent to confirm the selection anyway - for example, the authenticator may have skipped the selection prompt if there was only one eligible credential.

index.bs
equalsJeffH: I do not understand your comment @akshayku -- please explain?

akshayku: Out of `required` (RP requires UV), `preferred` (RP supports both kind of authenticators, one who sends UV bit set and one who don't ) and `discouraged` (RP don't want friction of user verification via ClientPin entry or otherwise but OK if he gets user verification anyway), I like `preferred`.

index.bs
equalsJeffH: I still suggest the above update to the language in lines 1821 & 1822.

emlun: Isn't this addressed in 2969b37 ?

index.bs
emlun: Done in 2969b37.

index.bs
emlun: Done in 2969b37.

index.bs
emlun: Done in 2969b37.

index.bs
emlun: Agreed; addressed in 2969b37.

index.bs
equalsJeffH: Hm, I'm not sure I agree with the above comments -- for several reasons...

1. @akshayku wrote:
    > In Client PIN case, user verification is not same as user presence check. 

    I do not see that stated in the [CTAP spec wrt Client PIN](https://fidoalliance.org/specs/fido-v2.0-rd-20170927/fido-client-to-authenticator-protocol-v2.0-rd-20170927.html#authenticatorClientPIN). 

    If I recall correctly, you argued that a user's entering a Client PIN on an authnr supporting Client PIN ought to interpreted as an act of [User Verification](https://w3c.github.io/webauthn/#user-verification).  I do not understand how that can not also be interpreted as a [user presence test](https://w3c.github.io/webauthn/#test-of-user-presence). 

2. @emlun wrote: 
    > This is a first factor operation (allowCredentials not given); the user has already been verified before selecting a credential.

    I am not sure that the absence of `allowCredentials` necessarily denotes a "first factor" operation (aka "passwordless").  What do you mean by "the user has already been verified before selecting a credential" ?  Do you mean that the user was "verified" by presence of a cookie, say?

    If I understand correctly (IIUC), in the case where the browser has valid ambient authentication credentials, e.g. cookies, for the RP, then the RP will be able (and ought) to supply `allowCredentials` because the RP will know who the user is.  I.e., this authn aka getAssertion operation is being a form a "step-up" in that the RP knows who the nominally-authn'd user is, and is employing webauthn to ensure that.  

    And IIUC, this is regardless of whether the authnr is performing as 1st multi-factor or 2nd factor.  @rlin1, what do you think ?

3. Does an authnr operating over CTAP perform the same detailed steps as are now specified in [WebAuthn's Authenticator model's operations](https://w3c.github.io/webauthn/#authenticator-model) ?

4. Overall, these autenticatorMakeCredential  and  authenticatorGetAssertion operations are _supposed_ to be "fuzzy" and leave room for platform implementors to figure out how implement this layer. CTAP is only one concrete instantiation of these operations.  Also, one can argue that there is an unspecified-in-webauthn layer between the [WebAuthn Authenticator model](https://w3c.github.io/webauthn/#authenticator-model) (of which the above operations are only a portion). For example, CTAP's client pin seems to occupy this unspecified-in-webauthn layer.  Are the algorithm descriptions becoming too detailed?  Do they match up with the [CTAP spec](https://fidoalliance.org/specs/fido-v2.0-rd-20170927/fido-client-to-authenticator-protocol-v2.0-rd-20170927.html)?

index.bs
equalsJeffH: ought to also get consensus with @jcjones, @balfanz & @leshi, @rlin1

index.bs
equalsJeffH: yes, thanks.

equalsJeffH:
added some further comments to this:

https://github.com/w3c/webauthn/pull/672#discussion_r150812527
https://github.com/w3c/webauthn/pull/672#discussion_r150811822

index.bs
emlun: (2): What I mean by "first factor operation" is exactly that `allowCredentials` has not been supplied to `credentials.get()`. Unfortunately this currently doesn't exactly correspond to when no `allowCredentialDescriptorList` is supplied to [authenticatorGetAssertion][authnr-get] (see #685), but I'm hoping that will get addressed.

What I mean by "the user has already been verified" is that with these proposed changes, the user has already been verified before step 9 in [authenticatorGetAssertion][authnr-get]. I meant my previous comment to be in the context of step 9 only.

(3-4): Hm, you're right that CTAP doesn't explicitly specify that the user being verified for the assertion must be the same user that created the credential. It looks like CTAP's PIN mechanism is implicitly single-user only, but the same cannot be said for the unspecified other kinds of UV gestures. CTAP also doesn't automatically require UV if `allowList` is not specified.

So I guess I'll have to revert those two things, then.

I agree that the authenticator operations in WebAuthn should be kept as loosely specified as possible, but I thought it seemed like the above were agreed-upon limitations necessary to ensure the user's privacy.

[authnr-get]: https://s3.amazonaws.com/pr-preview/w3c/webauthn/issue-644.html#op-get-assertion

emlun:
@equalsJeffH I made some changes to agree better with CTAP2. What do you think?

equalsJeffH:
thanks @emlun -- this now looks nominally OK to me, but I think it really needs explicit buy-in from Chrome, FF, and @rlin1:  in that my understanding of the nominal agreement had been that the default interaction case would be 2nd factor -- this seems to be changing it to 1st factor.  this is fine by me, however we ought to explcitly get buy-in from other working group participants on it.  thx, HTH.    

cc: @jcjones @rlin1 @balfanz @leshi @AngeloKai

emlun:
@equalsJeffH Thanks!

Hm, I did not intend to change the default interaction case. What parts of the changes are suggesting that?

index.bs
```diff
@@ -2039,24 +2140,42 @@ input parameters:
...
-1. Remove any items from |allowCredentialDescriptorList| that do not match a credential bound to this authenticator. A match 
-    occurs if a credential matches <code>|rpId|</code> and an |allowCredentialDescriptorList| item's <code>|allowCredentialDescriptorList| 
-    {{PublicKeyCredentialDescriptor/id}}</code> and <code>|allowCredentialDescriptorList|{{PublicKeyCredentialDescriptor/type}}</code>. 
+1. Remove any items from |allowCredentialDescriptorList| that do not match a credential bound to this authenticator. A match
+    occurs if a credential matches <code>|rpId|</code> and an |allowCredentialDescriptorList| item's
+    {{PublicKeyCredentialDescriptor/id}} and {{PublicKeyCredentialDescriptor/type}} members.
 1. If |allowCredentialDescriptorList| is now empty, return an error code equivalent to "{{NotAllowedError}}" and terminate the
     operation.
+
 1. Prompt the user to select a [=public key credential|credential=] |selectedCredential| from |allowCredentialDescriptorList|.
```
rlin1: this is only required if the list has more than 1 entry.

rlin1: so we should say: If the |allowCredentialDescriptorList| contains multiple entries, then ...

index.bs
rlin1: We SHOULD be clear that user verification, e.g. a fingerprint swipe IS a method for obtaining user consent - we don't want additional dialog boxes!

emlun: >We SHOULD NOT require additional user consent gathering by the platform for using an authenticator that is already registered!

It seems like you're saying that `getAssertion` operations should never ask for user consent, and should always be silent. Is that what you mean?

>We SHOULD be clear that user verification, e.g. a fingerprint swipe IS a method for obtaining user consent - we don't want additional dialog boxes!

Does the proposed language below need to be clearer on this point? It mirrors the corresponding part of the `getAssertion` operation in the CTAP2 spec:

>Collect user consent if required. [...]
>- If the "uv" option was specified and set to true:
>   - If device doesn't support user-identifiable gestures, return the CTAP2_ERR_OPTION_NOT_SUPPORTED error.
>   - Collect a user-identifiable gesture. If gesture validation fails, return the CTAP2_ERR_OPERATION_DENIED error.
>- If the "up" option was specified and set to true, collect the user's consent.
>   - If no consent is obtained and a timeout occurs, return the CTAP2_ERR_OPERATION_DENIED error.

index.bs
```diff
@@ -1560,7 +1618,12 @@ an assertion. Its {{PublicKeyCredentialRequestOptions/challenge}} member must be
     ::  This optional member contains a list of {{PublicKeyCredentialDescriptor}} object representing [=public key credentials=]
```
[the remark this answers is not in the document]
emlun: Good catch, thanks.

index.bs
emlun: Done!

index.bs
jcjones: The advantage we see in having it default to `"discouraged"` is that implementations don't default to a mechanism that will provide more information than they need. It also might keep the common case fast, for a hypothetical authenticator that requires different actions depending on `UV`'s state.

That said, I think there are enough U2F devices in the field that implementors realistically will have to deal with no-UV information for a long time to come, so I won't object to keeping this `"preferred"`.

jcjones:
@emlun said:
> Hm, I did not intend to change the default interaction case. What parts of the changes are suggesting that?

I think using `"preferred"` as the default option is what @equalsJeffH is referring to, and I agree that it does change the nature of the default interaction. 

I added a comment where I was pinged (https://github.com/w3c/webauthn/pull/672/files#r151271242). I won't oppose the will of the working group if there's consensus on this, but my preference would be to have this default to `"discouraged"` to keep the common case simple.

emlun:
@jcjones Ah, I see. Those are good points. I don't have a strong opinion either way. As far as I can tell it seems like most are more in favor of `"discouraged"`, but only @akshayku has voiced a strong opinion. I'm fine with changing the default to `"discouraged"` if noone objects.

jcjones:
@emlun: It looks to me like there's broad agreement on that. I don't see anything problematic with this change otherwise, and I think if you swap those defaults then this should be free to merge (after we get confirmation from the other 3 reviewers)

jcjones:
Oops, I spoke too soon. I missed the embedded threads. Disregard.

Still, I think adjusting that default would help. But the points about presence vs. user verification are better articulated by Rolf, Akshay, Jeff, et. al than I can do.

akshayku:
@jcjones : What is the common case in your opinion? Does not it maps to RP supporting all kind of authenticators?? and Isn't that the direction we should be going to?

emlun:
`"preferred"` indicates that the RP supports all authenticators, yes, but so does `"discouraged"`. I think it's also fair to say `"discouraged"` maps closer to the previous default `requireUserVerification: false` (in makeCredential).

akshayku:
@emlun 

No. In `discouraged` case, even if authenticator supports user verification, RP probably will not get it and RP has no way to know/take advantage of that feature. Similarly, Authenticator who has done extra work to make user verification possible in their product, will unfairly suffer because by default that feature is not used. `preferred` maps to default behavior of authenticator and RP has a better chance of supporting all kind of authenticators.

emlun:
@akshayku Ok. I guess you and I use "support" to mean different things - neither the client nor the RP will _refuse_ authenticators with UV in the `"discouraged"` case, so they'll still support them in that sense.

Anyway, with that it looks like consensus is possible for `"preferred"` but not for `"discouraged"`. Still waiting for responses from @balfanz and @leshi.

jcjones:
I think @akshayku and I are on different pages, and maybe I'm on the same page as @emlun, but to be honest: I'm fine with `"preferred"`, and arguing that letting the common case _promote_ UV authenticators is fine and in keeping with the rest of the model IMO.

christiaanbrand:
I don't really care that much. As long as we can override it.

akshayku:
@emlun : I think we have an agreement for `preferred`. If you agree, lets check this in so that we can move on.

emlun:
Great! I'll go ahead and merge if noone objects by Monday morning CET.

nadalin:
@jcjones @balfanz @AngeloKai  Please review asap so try to get closure before meeting this week

equalsJeffH:
@nadalin -- this was merged 7.5 hr before your comment above :)

PR #681, opened 2017-11-09
Fix #404 - Add a Security Consideration for Cryptographic Challenges

### LINKED ISSUE #404: explain challenge's security importance and use in both registration and authentication operations
need to expand the explanation of "challenge" in the spec -- challenge is used in both regstn and authn protocol runs and has security implications.   It MUST NOT be created on the client-side -- rather, it MUST be randomly created on the server-side with very low probability of collisions, and verified upon receipt back from the client and end of protocol run in order to provide protocol-run round-trip integrity. this needs to be explained in the spec and the code in the examples altered to reflect this -- especially since developers will simply cut'n'paste our examples!  :-P

jcjones:
***
[Preview](https://s3.amazonaws.com/pr-preview/jcjones/webauthn/404-challanges.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/webauthn/5f4f3e6...jcjones:f496efa.html)


index.bs
```diff
@@ -3989,6 +3995,16 @@ handled on the server side and do not need support from the API specified here.
     * Sometime later, the server deregisters this credential due to inactivity.
 
 
+# Security Considerations # {#security-considerations}
+
+## Cryptographic Challenges ## {#cryptographic-challenges}
+As a cryptographic protocol, Web Authentication is dependent upon randomized challenges
+to avoid replay attacks. Therefore, the [=challenge=] fields MUST be randomly generated
```
kpaulh: nit: would making 'field' singular be more consistent with the new wording?

jcjones:
This is not contentious, and I think it needs to be in WD-07. Merging it.

equalsJeffH:
this improved issue #322

Issue #184, opened 2016-08-24
Privacy concerns with blacklist/whitelist


hillbrad:
Do the blacklist/whitelist features allow re-identification of the user without informed consent?

If I call makeCredential() with a blacklist, is it clear to the user that the blacklist may reveal other identities they have registered with the site previously?

If I call getAssertion() with a whitelist, can't the timing of an immediate return if no credentials in that whitelist are present vs. a delay for user approval, also effectively re-identify a user without consent?

equalsJeffH:
see also issue #204

equalsJeffH:
see also issue #140

kpaulh:
Regarding getAssertion() with a whitelist, the Chrome implementation circumvents this by not returning immediately but by continuing to request user consent until timeout. In either case of user consent or timeout, Chrome will return a NotAllowedError.

emlun:
@kpaulh So essentially Chrome ignores this part of [§5.1.3][create].19 and [§5.1.4.1][get].17?

>**If any *authenticator* returns a status indicating that the user cancelled the operation,**
>
> 1. Remove *authenticator* from *issuedRequests*.
> 2. For each remaining *authenticator* in *issuedRequests* invoke the `authenticatorCancel` operation on *authenticator* and remove it from *issuedRequests*.

That should indeed make it impossible for the caller to see a difference between (1) timeout and (2) the credential is available, but the user denied consent to use it. It doesn't seem like this instruction would add much to the user experience either. So should we just delete it from the spec?

[create]: https://w3c.github.io/webauthn/#createCredential
[get]: https://w3c.github.io/webauthn/#discover-from-external-source

kpaulh:
Oh, those steps are still useful. We perform them if the user denies consent. The distinction I meant is that Chrome will not immediately return if there are no credentials in the whitelist present.

emlun:
Ah, I see. I also realized that those steps indeed won't have to go as long as the timeout isn't terminated early. What do you think of the additions proposed in #687?

kpaulh:
Yeah, that sums it up pretty well!

equalsJeffH:
see also #743

emlun:
I believe this can be closed now that #687 is merged.

nadalin:
@emlun please close

emlun:
Decided on 2018-01-17 call to close this. With #655 and #687 merged, and #613 in progress, this should no longer be an issue.

[...]

Issue #204, opened 2016-09-17
Privacy across Account IDs


yaronf:
4.5: `excludeList` allows an RP to tie different identities, i.e. to check if Alice and Bob are both used as identities on the same authenticator. This is because each of the `CredentialDescription` structures can contain a different id value, whereas if we only wanted to prevent multiple credentials for the same account, we would simply use the id value of the Account structure. Is this an attack we are willing to live with? Why not require (or allow) user consent for this step, e.g. "RP X wants to see other identities you have with it, do you allow that?"

@vijaybh: One issue is that there are authenticators which have no local storage, but encode the entire credential and all its metadata into the credential ID. So for these authenticators, a credential ID is required. However, **the authenticator could ignore any excludeList entries that are not for the same account ID**.

vijaybh:
Doesn't that just create another attack where the bad guy can find out which credential IDs map to which account IDs?

I think you express a valid concern BTW, just trying to figure out if there is any way to solve it without just displacing the problem.

yaronf:
I agree this is a problem, I just think it's less important, because it only applies to users who use multiple authenticators (multiple mobile devices?) to authenticate to the same account. And as far as I know, this is less common than having multiple accounts for the same user with the same device.

equalsJeffH:
see also #184

equalsJeffH:
see also issue 
#140

equalsJeffH:
see also #743

equalsJeffH:
A portion of this issue was fairly recently addressed by PR #651 which formalized the language for matching the items in the [`excludeCredentialDescriptorList`](https://w3c.github.io/webauthn/#op-make-cred) such that an RP can only probe for existing credentials that are mapped to itself. 

This seems to me to address a portion of @yaronf's suggestion of:
> the authenticator could ignore any excludeList entries that are not for the same account ID.

..in that we no longer have a notion of "account ID" and mapping to RP ID (and cred type) narrows what the RP can probe for, and is the best the spec presently can accomplish (i.e., without further changes).

Yes?


That said,  it does seem there is presently an issue of sorts at credential creation time, [as noted by](https://github.com/w3c/webauthn/issues/204#issue-177561011) @yaronf, in that a malicious RP can populate the  [`excludeCredentials`](https://w3c.github.io/webauthn/#dom-makepublickeycredentialoptions-excludecredentials) list with credential IDs it either knows or guesses in order to see if they exist on the user's presently-available authenticator(s). 

[...]

Issue #140, opened 2016-07-06
credential id privacy


AxelNennker:
the current editor's draft does not limit credential id in any way while it promises that the privacy of the user is protected cross RP - unlinkability.
https://w3c.github.io/webauthn/#credential-id

The Privacy subsection https://w3c.github.io/webauthn/#sec-attestation-privacy mentions Attestation Keys but not credential id. A credential id could leak PII if the platform or the authenticator is not careful about this.

If https://github.com/w3c/webauthn/issues/6 introduces structure to credential id then information leakage should be considered too.

equalsJeffH:
@jcjones notes in webauthn f2f tpac 2017-11-09 :   we will want to address this, but do not need to do for CR milestone -- punt to PR

rlin1:
TODO: address in privacy considerations section

nadalin:
@selfissued can you add a sentence about reuse of credential ids in the privacy sense, see @jcjones for deltails

rlin1:
This is NOT only related to the credential ID.  Authenticators can (and are intended) to be used to authenticate the user.  This might lead to unexpected consequences when using "anonymous" browsing mode.

rlin1:
Need to address whether/why not it could be possible to detect whether a CredentialID belongs to the current user *without* active user interaction(e.g. UP, UV).

emlun:
It is currently possible to detect without user consent in at least some circumstances, although at least the following method cannot be done silently. You make a `getAssertion()` call naming the credential ID you want to probe for; if it times out you gain no knowledge (courtesy of hotplugging), but if it fails before timeout you know the credential is available and that the user has declined the operation.

equalsJeffH:
Punting to PR milestone per discussion at webauthn f2f tpac 2017-11-09, see: https://github.com/w3c/webauthn/issues/140#issuecomment-343223907

equalsJeffH:
I believe this issue is addressed by the recent PR #899, cf. https://w3c.github.io/webauthn/#sctn-privacy-attacks and https://w3c.github.io/webauthn/#sec-non-correlatable-credentials

emlun:
Even with #899 I don't think there's any normative requirement to the effect of making sure credential IDs are not correlatable/identifying, but I agree that what we have is probably good enough.

equalsJeffH:
I dunno if it'd be possible to have such a normative requirement. 

Shall we close this as fixed?

AxelNennker:
Thank you

PR #687, opened 2017-11-13
Add privacy consideration about terminating getAssertion early

emlun:
Related issues: #184, #204 . Loosely related: #140 .

This adds a privacy consideration note to [§5.1.4. Use an existing credential to make an assertion][get] about the privacy implications of terminating before the timer expires.

The PR also includes some language changes to the waiting condition in both [§5.1.3. Create a new credential][create] and [§5.1.4. Use an existing credential to make an assertion][get] which are needed to make the privacy consideration text meaningful, and which were probably already intended.

[create]: https://w3c.github.io/webauthn/#createCredential
[get]: https://w3c.github.io/webauthn/#getAssertion

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/pull/687.html" title="Last updated on Jan 12, 2018, 6:06 PM GMT (f716b7f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/687/ade8321...f716b7f.html" title="Last updated on Jan 12, 2018, 6:06 PM GMT (f716b7f)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -791,8 +791,8 @@ When this method is invoked, the user agent MUST execute the following algorithm
 
     1. [=set/Append=] |authenticator| to |issuedRequests|.
 
-1. [=While=] |issuedRequests| [=list/is not empty=], perform the following actions depending upon
-    |lifetimeTimer| and responses from the authenticators:
+1. [=While=] |lifetimeTimer| has not expired, perform the following actions depending upon |lifetimeTimer| and responses from the
+    authenticators:
     <dl class="switch">
 
         :   If |lifetimeTimer| expires,
@@ -1069,7 +1069,7 @@ When this method is invoked, the user agent MUST execute the following algorithm
 
     1. [=set/Append=] |authenticator| to |issuedRequests|.
 
-1. While |issuedRequests| [=list/is not empty=], perform the following actions depending upon |lifetimeTimer|
+1. [=While=] |lifetimeTimer| has not expired, perform the following actions depending upon |lifetimeTimer|
     and responses from the authenticators:
 
     <dl class="switch">
--- a/index.bs
+++ b/index.bs
@@ -1089,6 +1089,9 @@ When this method is invoked, the user agent MUST execute the following algorithm
             1. [=set/For each=] remaining |authenticator| in |issuedRequests| invoke the [=authenticatorCancel=] operation
                 on |authenticator| and [=set/remove=] it from |issuedRequests|.
 
+            Note: This does not mean that the algorithm should be terminated, as that could leak identifying information about the
+            user. See [[#getAssertion-privacy-considerations]].
+
         :   If any |authenticator| returns an error status,
         ::  [=set/Remove=] |authenticator| from |issuedRequests|.
 
@@ -1166,6 +1169,20 @@ When this method is invoked, the user agent MUST execute the following algorithm
 During the above process, the user agent SHOULD show some UI to the user to guide them in the process of selecting and
 authorizing an authenticator with which to complete the operation.
 </div>
+
+
+#### Privacy considerations #### {#getAssertion-privacy-considerations}
+
+In order to protect the user from being identified without [=user consent|consent=], the client MUST make it impossible for the
+[=[RP]=] to distinguish between the cases:
+
+- A named [=public key credential|credential=] is not available.
+- A named [=public key credential|credential=] is available, but the user denies [=user consent|consent=] to use it.
+
+If the above cases are distinguishable, that leaks information by which a malicious [=[RP]=] could identify the user by probing
+for which [=public key credential|credentials=] are available. In particular, the client MUST NOT return a "{{NotAllowedError}}"
+before the |lifetimeTimer| has expired, as that would constitute such an information leak.
+
 
 ### Store an existing credential - PublicKeyCredential's `[[Store]](credential)` method ### {#storeCredential}
 
--- a/index.bs
+++ b/index.bs
@@ -4348,7 +4348,7 @@ RogueList when performing ECDAA-Verify (see section 3.6 in [[!FIDOEcdaaAlgorithm
 
 
 
-# Privacy Considertations # {#sctn-privacy-considerations}
+# Privacy Considerations # {#sctn-privacy-considerations}
 
 ## Attestation Privacy ## {#sec-attestation-privacy}
 
--- a/index.bs
+++ b/index.bs
@@ -955,7 +955,9 @@ When this method is invoked, the user agent MUST execute the following algorithm
 
     </dl>
 
-1. Return a {{DOMException}} whose name is "{{NotAllowedError}}".
+1. Return a {{DOMException}} whose name is "{{NotAllowedError}}". In order to prevent information leak that could identify the
+    user without [=user consent|consent=], this step MUST NOT be executed before |lifetimeTimer| has expired. See
+    [[#sec-assertion-privacy]] for details.
 
 During the above process, the user agent SHOULD show some UI to the user to guide them in the process of selecting and
 authorizing an authenticator.
@@ -1222,9 +1224,6 @@ When this method is invoked, the user agent MUST execute the following algorithm
             1. [=set/For each=] remaining |authenticator| in |issuedRequests| invoke the [=authenticatorCancel=] operation
                 on |authenticator| and [=set/remove=] it from |issuedRequests|.
 
-            Note: This does not mean that the algorithm should be terminated, as that could leak identifying information about the
-            user. See [[#sec-assertion-privacy]].
-
         :   If any |authenticator| returns an error status,
         ::  [=set/Remove=] |authenticator| from |issuedRequests|.
 
@@ -1297,7 +1296,9 @@ When this method is invoked, the user agent MUST execute the following algorithm
             1.  Return |constructAssertionAlg| and terminate this algorithm.
     </dl>
 
-1. Return a {{DOMException}} whose name is "{{NotAllowedError}}".
+1. Return a {{DOMException}} whose name is "{{NotAllowedError}}". In order to prevent information leak that could identify the
+    user without [=user consent|consent=], this step MUST NOT be executed before |lifetimeTimer| has expired. See
+    [[#sec-assertion-privacy]] for details.
 
 During the above process, the user agent SHOULD show some UI to the user to guide them in the process of selecting and
 authorizing an authenticator with which to complete the operation.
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -1168,6 +1170,20 @@ During the above process, the user agent SHOULD show some UI to the user to guid
...
+
+#### Privacy considerations #### {#getAssertion-privacy-considerations}
+
+In order to protect the user from being identified without [=user consent|consent=], the client MUST make it impossible for the
+[=[RP]=] to distinguish between the cases:
+
+- A named [=public key credential|credential=] is not available.
+- A named [=public key credential|credential=] is available, but the user denies [=user consent|consent=] to use it.
+
+If the above cases are distinguishable, that leaks information by which a malicious [=[RP]=] could identify the user by probing
```
kpaulh: "that leaks information" -> "information is leaked"?

emlun: That sounds better, thanks!

index.bs
```diff
@@ -1168,6 +1170,20 @@ During the above process, the user agent SHOULD show some UI to the user to guid
...
+In order to protect the user from being identified without [=user consent|consent=], the client MUST make it impossible for the
+[=[RP]=] to distinguish between the cases:
+
+- A named [=public key credential|credential=] is not available.
+- A named [=public key credential|credential=] is available, but the user denies [=user consent|consent=] to use it.
+
+If the above cases are distinguishable, information is leaked by which a malicious [=[RP]=] could identify the user by probing
+for which [=public key credential|credentials=] are available. In particular, the client MUST NOT return a "{{NotAllowedError}}"
+before the |lifetimeTimer| has expired, as that difference in timing would constitute such an information leak.
+
```
equalsJeffH: IMV, the latter final normative sentence needs to be stated in the algorithm itself, rather than buried down here in Priv Cons.

equalsJeffH:
Reviewed, modest changes requested: https://github.com/w3c/webauthn/pull/687#pullrequestreview-81645077

equalsJeffH:
so this fixes issues: #184, #204 ?

emlun:
Agreed on all three points. I'll finish this up after #705 is merged.

Yes, this fixes #184 and #204, assuming I'm right in interpreting #204 as raising the same concern as #184.

emlun:
@equalsJeffH Please re-review!

index.bs
```diff
@@ -2589,6 +2594,23 @@ in several ways, including:
     signature using the [=ECDAA-Issuer public key=], but the attestation signature does not serve as a global correlation handle.
 
 
+#### Privacy and Authentication Ceremonies #### {#sec-assertion-privacy}
+
+In order to protect the user from being identified without [=user consent|consent=], implementations of the
+{{PublicKeyCredential/[[DiscoverFromExternalSource]](origin, options, sameOriginWithAncestors)}} method need to take care to
+not leak information that could enable the [=[RP]=] to distinguish between the cases:
```
equalsJeffH: nit: s/the cases/these cases/

index.bs
```diff
@@ -2589,6 +2594,23 @@ in several ways, including:
...
+
+- A named [=public key credential|credential=] is not available.
+- A named [=public key credential|credential=] is available, but the user denies [=user consent|consent=] to use it.
```
equalsJeffH: nit: s/user denies/user does not/

index.bs
```diff
@@ -2589,6 +2594,23 @@ in several ways, including:
...
+
+If the above cases are distinguishable, information is leaked by which a malicious [=[RP]=] could identify the user by probing for
+which [=public key credential|credentials=] are available. For example, one such information leak is if the client returns a
+failure response as soon as the user denies [=user consent|consent=] to proceed with the operation. In this case the [=[RP]=]
```
equalsJeffH: nit: s/user denies/user does not/

emlun:
@equalsJeffH Reading this again, I'm inclined to revert this one change https://github.com/w3c/webauthn/pull/687#discussion_r159073322  to "the user denies consent", as the containing sentence is meant to express "if the client returns a failure in response to the user's action of refusing a prompt for consent". What do you think?

nadalin:
@selfissued please review so we can merge

index.bs
```diff
@@ -885,8 +885,8 @@ When this method is invoked, the user agent MUST execute the following algorithm
 
     1. [=set/Append=] |authenticator| to |issuedRequests|.
 
-1. [=While=] |issuedRequests| [=list/is not empty=], perform the following actions depending upon
-    |lifetimeTimer| and responses from the authenticators:
+1. [=While=] |lifetimeTimer| has not expired, perform the following actions depending upon |lifetimeTimer| and responses from the
```
selfissued: Why was the clause "issuedRequests not empty" deleted?

index.bs
```diff
@@ -1233,7 +1235,7 @@ When this method is invoked, the user agent MUST execute the following algorithm
 
     1. [=set/Append=] |authenticator| to |issuedRequests|.
 
-1. While |issuedRequests| [=list/is not empty=], perform the following actions depending upon |lifetimeTimer|
+1. [=While=] |lifetimeTimer| has not expired, perform the following actions depending upon |lifetimeTimer|
```
selfissued: Why was the clause "issuedRequests not empty" deleted?

index.bs
emlun: Because with the previous language the operation would terminate immediately if no authenticator is present at the moment when the timer is started, but step 19 is intended to be run asynchronously as authenticators appear within the timeout. This change is mostly orthogonal to the addition of the privacy considerations section, but it's necessary for the privacy consideration text to make any sense. I also believe we need to do this change for the algorithms themselves to make sense, even disregarding the new privacy consideration.

index.bs
emlun: See https://github.com/w3c/webauthn/pull/687#discussion_r161974515

emlun:
@selfissued Thank you for reviewing. The clause "issuedRequests is not empty" was changed to "lifetimeTimer has not expired" because with the previous language the operation would terminate immediately if no authenticator is present at the moment when the timer is started, but makeCredential step 19 and getAssertion step 17 are intended to be run asynchronously as authenticators appear within the timeout. This change is mostly orthogonal to the addition of the privacy considerations section, but it's necessary for the privacy consideration text to make any sense. I also believe we need to make this change for the algorithms themselves to make sense, even without the new privacy consideration.

index.bs
emlun: See https://github.com/w3c/webauthn/pull/687#issuecomment-358223418

index.bs
emlun: See https://github.com/w3c/webauthn/pull/687#issuecomment-358223418

selfissued:
Thanks for the explanation, @emlun .  Please proceed with the merge.

equalsJeffH:
@emlun wrote in https://github.com/w3c/webauthn/pull/687#issuecomment-350059267:
> Yes, this fixes #184 and #204, assuming I'm right in interpreting #204 as raising the same concern as #184.

Actually, upon further review, I _do not believe_ this PR addressed issue #204, because the latter is in regards to [#createCredential](https://w3c.github.io/webauthn/#createCredential) aka `[[Create]]`, and this PR addressed privacy concerns at authentication time i.e. [#getAssertion](https://w3c.github.io/webauthn/#getAssertion) aka `[[DiscoverFromExternalSource]]` aka "[authentication ceremonies](https://w3c.github.io/webauthn/#sec-assertion-privacy)".   See also https://github.com/w3c/webauthn/issues/184#issuecomment-359051104

So we still have #204 to address.

emlun:
I believe with hot-plugging `makeCredential` doesn't have the same information leak anymore, and either way this PR made the same normative changes to both the `makeCredential` and the `getAssertion` algorithms. But you're right that the Privacy Considerations subsection only mentions `getAssertion`, which should be fixed.

My reasoning is the following. The attack would previously be:

1. The attacker calls `makeCredential` with some credentials in the `excludeCredentials` parameter.
2. The browser notices that one of the excluded parameters is available, and terminates the operation immediately.
3. The attacker concludes that the user possesses one of the named credentials.

Whereas now, if an excluded credential is available, the browser will simply ignore that authenticator and instead wait for a new authenticator without an excluded credential to appear. If the browser were to violate the spec and return a `NotAllowedError` as soon as the user denies the prompt (which step 21 now expressly forbids), the attack would be:

1. The attacker calls `makeCredential` with some credentials in the `excludeCredentials` parameter.
2. The browser notices that one of the excluded parameters is available, ignores that authenticator and waits for another to appear.
3. The user plugs in a new authenticator.
4. The browser asks the user if they want to create a credential on that authenticator.
5. The user denies the prompt.
6. The browser returns a `NotAllowedError` to the attacker.
7. The attacker notices that a `NotAllowedError` was returned before the timer expired.
    - The attacker **can** conclude that the user has an authenticator that does not contain one of the named credentials.
    - The attacker **cannot** conclude that the user has an authenticator that contains one of the named credentials.
    - The attacker **cannot** conclude that the user does not have an authenticator that contains one of the named credentials.

I suppose this conclusion could perhaps be combined with some external assumptions (like an upper limit on how many authenticators a user is likely to own) to narrow the field to some extent, possibly enough to identify a single user. But again, this should also not be possible if the browser never returns the `NotAllowedError` before the timer expires.

Please let me know if you spot any errors in the above reasoning. :)

equalsJeffH:
**_Thanks_** @emlun for your [detailed analysis above](https://github.com/w3c/webauthn/pull/687#issuecomment-359087435). I had not reviewed [the diff](https://github.com/w3c/webauthn/pull/687/files) closely enough and had focused on the added [Authentication Ceremonies privacy considerations](https://w3c.github.io/webauthn/#sec-assertion-privacy) subsection rather than the revised lifetime timer stipulations in _**both**_  [#createCredential](https://w3c.github.io/webauthn/#createCredential) aka `[[Create]]` **_and_** [#getAssertion](https://w3c.github.io/webauthn/#getAssertion) aka `[[DiscoverFromExternalSource]]`.  For [#createCredential](https://w3c.github.io/webauthn/#createCredential) aka `[[Create]]` these are steps 20 and 21. mea culpa!

I agree with the overall conclusion of the [detailed analysis above](https://github.com/w3c/webauthn/pull/687#issuecomment-359087435): a malicious RP (aka "the attacker") will not learn anything through seeding of the [`excludeCredentials` list](https://w3c.github.io/webauthn/#dom-makepublickeycredentialoptions-excludecredentials) because any returned `NotAllowedError`s are returned after the `lifetimeTimer` has expired, regardless of the user's actions (see step 21). 

It seems, IIUC, there are a couple details-level issues with the steps in the [detailed analysis above](https://github.com/w3c/webauthn/pull/687#issuecomment-359087435), but they do not AFAICT affect this conclusion.

Here they are for completeness' sake:
> 2. The browser notices that one of the excluded parameters is available, ignores that authenticator and waits for another to appear.

AFAICT, the browser/client-platform does not do this (see [#createCredential](https://w3c.github.io/webauthn/#createCredential) steps 19.5 and 19.6), rather the authnr does this (see [authenticatorMakeCredential](https://w3c.github.io/webauthn/#op-make-cred) step 3), and returns `NotAllowedError`, which is handled by [#createCredential](https://w3c.github.io/webauthn/#createCredential) step 20 switch branch "If any authenticator returns an error status".

> 4. The browser asks the user if they want to create a credential on that authenticator.

nit: AFAICT the spec does not stipulate whether the browser, client platform, or authenticator prompts the user.

> 6. The browser returns a `NotAllowedError` to the attacker.
> 7. The attacker notices that a `NotAllowedError` was returned before the timer expired.

Ah, but the imperative "this step **MUST NOT** be executed before lifetimeTimer has expired" language this PR inserted into [#createCredential](https://w3c.github.io/webauthn/#createCredential) step 21 means that the error is not returned before the timer expires, yes?  (which makes the above analysis stronger). 

thanks again.

emlun:
Thanks for your detailed response!

>>2. The browser notices that one of the excluded parameters is available, ignores that authenticator and waits for another to appear.
>
>AFAICT, the browser/client-platform does not do this [...] rather the authnr does this [...] and returns `NotAllowedError`

Ah yes, that is correct. Either way, the end result is that the client ignores the authenticator in question and keeps waiting for another candidate.

>nit: AFAICT the spec does not stipulate whether the browser, client platform, or authenticator prompts the user.

Indeed, although [authenticatorMakeCredential][amc] step 6 softly specifies that the authenticator does it if capable, and otherwise the client. I chose to keep it simple because I don't know of existing authenticator hardware that allows active denial of consent (as opposed to passive denial by timeout).

>Ah, but the imperative [...] this PR inserted into [#createCredential](https://w3c.github.io/webauthn/#createCredential) step 21 means that the error is not returned before the timer expires, yes? [...]

Assuming implementers read and implement all the steps, yes. :)

[amc]: https://w3c.github.io/webauthn/#op-make-cred

PR #693, opened 2017-11-17
build on Adding a choice for RP to express preferences for attestation types

### LINKED ISSUE #628: Allow RPs to choose between "required" and "optional" attestation in credentials.create()

equalsJeffH:
here is a build on the present state of PR #636 that addresses my comments and does a bunch of polishing it would have been too hard to describe in comments over there on #636.

fix #628

***
[Preview](https://s3.amazonaws.com/pr-preview/w3c/webauthn/jeffh-balfanz-patch-4.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/webauthn/cc94364...cbd4fad.html)


index.bs
```diff
@@ -839,6 +841,27 @@ When this method is invoked, the user agent MUST execute the following algorithm
...
                     bytes of <code>|credentialCreationData|.[=attestationObjectResult=]</code>'s value.
 
+                1.  If <code>|credentialCreationData|.[=AttestationConveyancePreferenceOption=]</code>'s value is
+                    <dl class="switch">
+                        :   "none"
+                        ::  replace potentially uniquely identifying information (such as [=AAGUID=] and 
+                            [=attestation certificates=]) in the [=attested credential data=] and [=attestation statement=],
+                            respectively, with blinded versions of the same data.
+                            
+                        Issue: need to specify how to perform such blinding?
```
jcjones: I think we need to define "blinding" rather than necessarily proscribe what to do, but let's do it at CR.

If we get agreement on this PR, let's follow up and add an issue number into this line.

index.bs
```diff
@@ -839,6 +841,27 @@ When this method is invoked, the user agent MUST execute the following algorithm
...
+                        :   "indirect"
+                        ::  The client MAY replace the [=AAGUID=] and [=attestation statement=] with a more privacy-friendly 
+                            and/or more easily verifiable version of the same data (for example, by employing a [=Privacy CA=]).
+    
+                        :   "direct"
+                        ::  convey the authenticator's [=AAGUID=] and [=attestation statement=], unaltered, to the RP.
+  
+                        Issue: @balfanz wishes to add to the "direct" case: 
+                            If, for whatever reason, the client cannot pass on the authenticator-generated attestation
+                            statement, it MUST terminate the credential generation operation with a "NotAllowedError".
```
jcjones: This needs to be resolved for interop, but ... I think I am OK with resolving it for CR, too.

If we get agreement on this PR, let's follow up and add an issue number into this line.

index.bs
```diff
@@ -1527,6 +1555,42 @@ example of the latter, when the user is accessing the [=[RP]=] from a given clie
...
+
+<div dfn-type="enum-value" dfn-for="AttestationConveyancePreference">
+    *   <dfn>none</dfn> - indicates that the [=[RP]=] is not interested in [=authenticator=] [=attestation=].
+        The client may replace the [=AAGUID=] and [=attestation statement=] generated
+        by the authenticator with meaningless client-generated values. For example, in order to avoid having to obtain 
+        [=user consent=] to relay uniquely identifying information to the [=[RP]=], or to save a roundtrip to a Privacy CA. 
+
+        This is the default value.
+
+    *   <dfn>indirect</dfn> - indicates that the [=[RP]=] prefers an [=attestation=] conveyance yielding to verifiable
```
jcjones: s/yielding to verifiable/yielding verifiable/

index.bs
rlin1: I am concerned about the phrase "for whatever reason, the client cannot pass", since it is pretty vague.
Would not installing+activating some "enterprise policy" be such a "reason"?
I would be fine with 
a) leaving the statement as is, i.e. "convey the authenticator's AAGUID and attestation statement unaltered to the RP." or
b) defining such reasons being "if the authenticator's attestation statement violates the privacy principles, ...", where privacy principles refer to the minimum number of 100k Authenticator of the same model or not publishing Metadata Statement regarding this authenticator.

agl:
This looks good to me.

index.bs
akshayku: Responding in two places to make sure this is not lost:

Agree with @rlin1 concern. I'd like to suggest the following re-wording:

`If authenticator violates privacy requirements of the attestation scheme it is using, the client MUST terminate the credential generation operation with a AttestationNotPrivateError' 

In my opinion, this wording scopes the problem to privacy and at the same time extends it beyond batch attestation issues. For example, improper implementation of the TCG attestation scheme could also be covered.

equalsJeffH:
thanks for your reviews.  though, in my haste I messed up and created this PR against the master branch rather than against the balfanz-patch-4 branch, sigh. 

so i spose there's a couple of options (assuming we decide to merge this upon revisions):
1. close this PR with no action, and I create a new PR against PR #636.
2. merge this PR to master rather than merging PR #636 to master (and close PR #636 with no action)

WDYT?

akshayku:
#636 looks in better shape, have more changes than is reflected in this PR and ready to be checked in. 
How about, you check this in and after that update #636 with merge against master??

akshayku:
and then we can check in #636 if everyone agrees.

equalsJeffH:
thanks @agl, but hm, not sure need to do that because I did originally create jeffh-balfanz-patch-4 from balfanz-patch-4.  hm, it seems balfanz-patch-4 has changed out from underneath me as @akshayku points out (thx).... I have to head to the airport now, but will look into this further once airborne, thanks.

index.bs
equalsJeffH: done  in 802883f

index.bs
equalsJeffH: doh, thx!

index.bs
equalsJeffH: Note that @balfanz has updated this language in PR #636 with this commit: 825cce7

equalsJeffH:
OK, am back online.  turns out [one can "edit" a PR and change the base branch](https://github.com/blog/2224-change-the-base-branch-of-a-pull-request), so this PR is now properly targeting `balfanz-patch-4`.

I have also made the couple of editorial fixes proposed by @jcjones, found something I had not appropriately edited, and submitted issue #694 for defining "blinding".   

what do folks think about merging this into balfanz-patch-4 ?   I also have to figure out how to resolve the conflicts.  i tried to merge balfanz-patch-4 into this branch but that just says "alredy up to date".

agl:
gitk shows that the relationship between this and balfanz-patch-4 is complex. Does this not include & subsume the latter? I.e. why not land this and abandon balfanz-patch-4?

equalsJeffH:
@agl wrote:
> gitk shows that the relationship between this and balfanz-patch-4 is complex. 

yep.

> Does this not include & subsume the latter? I.e. why not land this and abandon balfanz-patch-4?

yep, that's an option. what do folks think? @balfanz ?   (I've adapted the language from the latest commit 825cce7 to PR #636)

nadalin:
@akshayku @jyasskin @balfanz  Please review asap

index.bs
```diff
@@ -289,7 +296,9 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
     In the WebAuthn context, [=attestation=] is employed to <em>attest</em> to the <em>provenance</em> of an [=authenticator=]
     and the data it emits; including, for example: [=credential IDs=], [=credential key pairs=], signature counters, etc. An
     [=attestation statement=] is conveyed in an [=attestation object=] during [=registration=]. See also [[#sctn-attestation]]
-    and [Figure 3](#fig-attStructs).
+    and [Figure 3](#fig-attStructs). Whether or how the client platform conveys the [=attestation statement=] and [=AAGUID=] 
+    portions of the [=attestation object=] to the [=[RP]=] is described tby <dfn>attestation conveyance</dfn>, see
```
jyasskin: Typo: tby

jyasskin: If you `<dfn>` "attestation conveyance" in #attestation-convey, you can remove the "see" clause.

index.bs
```diff
@@ -1545,38 +1565,40 @@ example of the latter, when the user is accessing the [=[RP]=] from a given clie
 use a [=roaming authenticator=] which was originally registered with the [=[RP]=] using a different client.
 
 
-### Attestation Presentation Preference enumeration (enum <dfn enum>AttestationPresentationPreference</dfn>) ### {#attestation}
+### Attestation Conveyance Preference enumeration (enum <dfn enum>AttestationConveyancePreference</dfn>) ### {#attestation-convey}
 
-[=[RPS]=] may use the {{AttestationPresentationPreference}} enum to specify their preference regarding attestation types used
+[=[RPS]=] may use {{AttestationConveyancePreference}} to specify their preference regarding [=attestation conveyance=]
```
jyasskin: I'd `<dfn>` "attestation conveyance" here or in the header of this section, since this is where it's defined.

index.bs
equalsJeffH: fixed.

equalsJeffH:
@jyasskin said:
> If this is the PR we're thinking about merging, could you point it at master?

was waiting to hear from @balfanz, but ok.

index.bs
```diff
@@ -827,6 +835,9 @@ When this method is invoked, the user agent MUST execute the following algorithm
                 :   <code><dfn for="credentialCreationData">clientDataJSONResult</dfn></code>
                 ::  whose value is the bytes of |clientDataJSON|.
 
+                :   <code><dfn for="credentialCreationData">AttestationConveyancePreferenceOption</dfn></code>
```
jyasskin: It's in a struct, so maybe it doesn't matter, but we usually have field names start with lower-case.

index.bs
```diff
@@ -839,6 +850,28 @@ When this method is invoked, the user agent MUST execute the following algorithm
...
+                        :   "indirect"
+                        ::  The client MAY replace the [=AAGUID=] and [=attestation statement=] with a more privacy-friendly
+                            and/or more easily verifiable version of the same data (for example, by employing a [=Privacy CA=]).
+
+                        :   "direct"
+                        ::  convey the [=authenticator=]'s [=AAGUID=] and [=attestation statement=], unaltered, to the RP.
+
+                        Issue: @balfanz wishes to add to the "direct" case:
+                            If the [=authenticator=] violates the privacy requirements of the [=attestation type=] it is using, 
+                            terminate this algorithm with a "{{AttestationNotPrivateError}}".
```
jyasskin: I think the most we can say is that "the UA SHOULD terminate ...", since it takes work to determine that the authentigator is violating the privacy requirements, so UAs won't catch it immediately.

index.bs
```diff
@@ -839,6 +850,28 @@ When this method is invoked, the user agent MUST execute the following algorithm
...
                     bytes of <code>|credentialCreationData|.[=attestationObjectResult=]</code>'s value.
 
+                1.  If <code>|credentialCreationData|.[=AttestationConveyancePreferenceOption=]</code>'s value is
+                    <dl class="switch">
+                        :   "none"
+                        ::  replace potentially uniquely identifying information (such as [=AAGUID=] and
+                            [=attestation certificates=]) in the [=attested credential data=] and [=attestation statement=],
+                            respectively, with blinded versions of the same data.
+
+                        Issue: need to define "blinding".  See issues [#694](https://github.com/w3c/webauthn/issues/694) and
```
jyasskin: Bikeshed lets you write `Issue(694):` to link to the issue: https://tabatkins.github.io/bikeshed/#remote-issues

index.bs
```diff
@@ -866,6 +876,28 @@ When this method is invoked, the user agent MUST execute the following algorithm
                 1.  Let |attestationObject| be a new {{ArrayBuffer}}, created using |global|'s [=%ArrayBuffer%=], containing the
                     bytes of <code>|credentialCreationData|.[=attestationObjectResult=]</code>'s value.
 
+                1.  If <code>|credentialCreationData|.[=AttestationConveyancePreferenceOption=]</code>'s value is
+                    <dl class="switch">
+                        :   "none"
+                        ::  replace potentially uniquely identifying information (such as [=AAGUID=] and
```
jyasskin: Capitalize "Replace", I think.

index.bs
```diff
@@ -866,6 +876,28 @@ When this method is invoked, the user agent MUST execute the following algorithm
...
+
+                        Issue: need to define "blinding".  See issues [#694](https://github.com/w3c/webauthn/issues/694) and
+                            [#462](https://github.com/w3c/webauthn/issues/462).
+
+                        :   "indirect"
+                        ::  The client MAY replace the [=AAGUID=] and [=attestation statement=] with a more privacy-friendly
+                            and/or more easily verifiable version of the same data (for example, by employing a [=Privacy CA=]).
+
+                        :   "direct"
+                        ::  convey the [=authenticator=]'s [=AAGUID=] and [=attestation statement=], unaltered, to the RP.
```
jyasskin: Capitalize "Convey", I think.

index.bs

index.bs

index.bs
equalsJeffH: its actually of the form "Issue(w3c/webauthn#694): ...issue text..."

index.bs

index.bs

christiaanbrand:
This looks good for merging.

index.bs
```diff
@@ -866,6 +876,28 @@ When this method is invoked, the user agent MUST execute the following algorithm
...
+
+                        :   "indirect"
+                        ::  The client MAY replace the [=AAGUID=] and [=attestation statement=] with a more privacy-friendly
+                            and/or more easily verifiable version of the same data (for example, by employing a [=Privacy CA=]).
+
+                        :   "direct"
+                        ::  Convey the [=authenticator=]'s [=AAGUID=] and [=attestation statement=], unaltered, to the RP.
+
+                        Issue: @balfanz wishes to add to the "direct" case:
+                            If the [=authenticator=] violates the privacy requirements of the [=attestation type=] it is using,
```
akshayku: Thinking more about this, going forward, we have to decide who makes the decision whether an authenticator is violating privacy requirements, how they are reaching that conclusion and what are the procedures/requirements. 

IMO, this fits nicely into FIDO metadata service and Certification group who are giving the certification to these authenticators in the first place. There are unknowns here that MDS/Certification group has to figure out and this discussion here looks little premature to have right now given the state of the ecosystem.

nadalin:
@balfanz Please review so we can get this merged

nadalin:
@gmandyam Can you review as you had some concerns expressed

gmandyam:
a) Re:  'none'.  It seems to be potentially harmful to allow the client to provide "meaningless client-generated values" for attestation/AAGUID since what constitutes 'meaningless' is ambiguous.  Poorly-designed clients can provide legitimate AAGUID's or replayed attestations while claiming they are meaningless.  My recommendation:  assign default values for the AAGUID and the attestation for this option (e.g. all 1's for AAGUID and all 0's for attestation).

b) Re: 'indirect'.  "There is no guarantee that the  Relying Party  will obtain a verifiable  attestation statement  in this case."  I take this to mean that it is possible for the RP to receive a verifiable attestation when "indirect" is selected.  What does a verifiable attestation statement look like in this case?  Would it conform to one of the pre-registered attestation formats?  Or it is yet to be defined?

emlun:
>a) Re: 'none'. It seems to be potentially harmful to allow the client to provide "meaningless client-generated values" [...] My recommendation: assign default values for the AAGUID and the attestation for this option (e.g. all 1's for AAGUID and all 0's for attestation).

I agree, but instead suggest leaving the attestation statement out completely. For example, the attestation object could look like `{ "authData": [bytes], "fmt": "packed", "attStmt": {} }`, or `{ "authData": [bytes], "fmt": "none" }` if we add `"none"` as an attestation statement format.

rlin1:
>b) Re: 'indirect'. "There is no guarantee that the Relying Party will obtain a verifiable attestation statement in this case." I take this to mean that it is possible for the RP to receive a verifiable attestation when "indirect" is selected. What does a verifiable attestation statement look like in this case? Would it conform to one of the pre-registered attestation formats? Or it is yet to be defined?

It needs to be one of the already defined attestation formats since the specification of AuthenticatorAttestationResponse and all related details haven't changed.

gmandyam:
@rlin1 

Is your view that the verification and signing procedures outlined in Sec. 8 do not have to change depending on whether 'indirect' is selected?  In addition, is your view that any of the attestation formats are sufficient?

For instance, the signing procedure for packed states "If Basic or Privacy CA attestation is in use, the authenticator produces the sig by concatenating authenticatorData and clientDataHash" - the procedure makes no mention about whether 'indirect' is selected.  

However, authenticatorData includes attestedCredentialData, which in turn includes "The AAGUID of the authenticator".  It is not clear whether this text should be interpreted as the actual AAGUID of the authenticator, or the client-generated dummy AAGUID when 'indirect' is selected.

equalsJeffH:
thanks @balfanz for merging-in the recent changes from master branch.

agl:
Adding a "none" attestation format might be interesting in the future but we are pondering providing a retrospective unblinding token in the no-attestation case, which we would need to be carried in the attestation data(*).

> However, authenticatorData includes attestedCredentialData, which in turn includes "The AAGUID of the authenticator". Since AAGUID can be used to fingerprint the device, should it be left out if 'indirect' is selected and the aim is to provide a verifiable attestation statement to the RP?

Eliminating, or not, the AAGUID is an underlying CTAP issue. Because a Privacy CA cannot learn the site that a user is registering with, it can only learn the hash of the signed data. That means that the signed data cannot be altered by a client without invalidating the signature by the Privacy CA and thus the AAGUID cannot be eliminated without co-operation from the token.

Thus I think specifying how to remove the AAGUID would be premature at this point. 

(*) Without a Privacy CA the unblinding scheme needs to be asymmetric, which makes it considerably more complicated. Thus, as an aside, if any RPs who might consider the "none" scheme would be influenced by the ability to find, in the event that a security problem is disclosed for certain tokens, which of their users are using those tokens, please let me know.

equalsJeffH:
> @gmandyam assiduously wrote:
> >a) Re: 'none'. It seems to be potentially harmful to allow the client to provide "meaningless client-generated values" [...] My recommendation: assign default values for the AAGUID and the attestation for this option (e.g. all 1's for AAGUID and all 0's for attestation).

@emlun replied creatively:
> I agree, but instead suggest leaving the attestation statement out completely. For example, the attestation object could look like `{ "authData": [bytes], "fmt": "packed", "attStmt": {} }`, 

that sounds relatively easily workable.

> or `{ "authData": [bytes], "fmt": "none" }` if we add `"none"` as an attestation statement format.

this approach seems like it'd require more spec surgery, ie, it looks to me like the [`attStmtTemplate` ](https://w3c.github.io/webauthn/#generating-an-attestation-object)would have to change to make the `attStmt` member optional. Plus defining the "none" attstn stmt format. Though, a side-effect of this would be one could have authnrs that just plain do not provide attestation.

emlun:
@agl 

>[...] we are pondering providing a retrospective unblinding token in the no-attestation case, which we would need to be carried in the attestation data(*).

For that to be of any value, aren't you assuming the RP will store the attestation statement? What's to say that an RP will store it if they don't care about authenticator attestation (since they sent `attestation: "none"`, or left it at that default)?

It is definitely a good idea for the Privacy CA case, where the RP explicitly _does_ care and can reasonably be expected to store the attestation statement.

>[...] the signed data cannot be altered by a client without invalidating the signature by the Privacy CA and thus the AAGUID cannot be eliminated without co-operation from the token.

However in the Privacy CA case, the authenticator's signature isn't passed on to the RP, correct? If so, the Privacy CA could alter the authenticator data - e.g., zero out the AAGUID - before signing it, without cooperation from the token.

agl:
> For that to be of any value, aren't you assuming the RP will store the attestation statement? What's to say that an RP will store it if they don't care about authenticator attestation (since they sent attestation: "none", or left it at that default)?

I believe the dominant practice at the moment, for sites using the U2F API in Chrome, is not to discriminate on attestation at registration time, but to store the certificate in case it's useful later for messaging users who might be affected by specific security issues.

> However in the Privacy CA case, the authenticator's signature isn't passed on to the RP, correct? If so, the Privacy CA could alter the authenticator data - e.g., zero out the AAGUID - before signing it, without cooperation from the token.

A Privacy CA is essentially signing statements to the effect: &ldquo;I have seen a signature for the following hash made by a valid token&rdquo;. If the client passes only the hash of the signed data, the Privacy CA can check the token's attestation signature and sign that same hash. However, that means that the AAGUID cannot be altered. The client could send the signed data itself to the Privacy CA, but then the Privacy CA learns the RP ID that the user is registering with, which is likely unacceptable.

emlun:
@equalsJeffH

>Though, a side-effect of this would be one could have authnrs that just plain do not provide attestation.

Yeah. It looks like an implicit assumption up until now has been that attestation statements are created exclusively by authenticators, but with the `"indirect"` feature we're opening up for the client and/or a proxy to also create their own attestation statements. We could perhaps forbid authenticators from using a `none` attestation statement format, but it could definitely be confusing.

On the other hand I wouldn't want to complicate any of the existing formats with additional optional behaviour, like making members optional in the `packed` format. Would it be possible to do something like this (pseudo-spec)?

>If the RP sends `attestation: "none"`, the client and/or authenticator MAY ignore any requirements on the `attStmt` member of the attestation object.

Overriding behaviour elsewhere like that could of course be quite confusing in its own way instead.

equalsJeffH:
@emlun wrote:
> We could perhaps forbid authenticators from using a none attestation statement format, but it could definitely be confusing.

@agl had said:
> Adding a "none" attestation format might be interesting in the future...

Yeah. I'm thinking that at this time we not define a "none" attstn stmt format.

emlun:
@agl Ah, I see, that makes sense. I'm happy as long as we don't urge clients to send _completely_ useless attestation statements for blinding. I think no attestation statement is better than nonsense that could be confused for something meaningful.

PR #705, opened 2017-12-04
fix #322: flesh out Security Considerations (for now)

### LINKED ISSUE #322: overall security considerations section or document

equalsJeffH:
per https://github.com/w3c/webauthn/issues/322#issuecomment-347936349 and https://github.com/w3c/webauthn/issues/322#issuecomment-347953417

Note: this moves the Attestation Sec Cons to be a subsection of the overall Sec Cons, and creates a separate Privacy Cons section.

fix #322

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/pull/705.html" title="Last updated on Jan 10, 2018, 9:49 PM GMT (67b41d5)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/705/3cfaeba...67b41d5.html" title="Last updated on Jan 10, 2018, 9:49 PM GMT (67b41d5)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -4211,6 +4211,8 @@ handled on the server side and do not need support from the API specified here.
 
 # Security Considerations # {#security-considerations}
 
+
+
 ## Cryptographic Challenges ## {#cryptographic-challenges}
 
 As a cryptographic protocol, Web Authentication is dependent upon randomized challenges
@@ -4385,7 +4387,7 @@ Axel Nennker, Kimberly Paulhamus, Adam Powers, Yaron Sheffer, Mike West, Jeffrey
   "FIDOSecRef": {
     "authors": ["R. Lindemann", "D. Baghdasaryan", "B. Hill"],
     "title": "FIDO Security Reference",
-    "href": "https://fidoalliance.org/specs/fido-uaf-v1.0-ps-20141208/fido-security-ref-v1.0-ps-20141208.html",
+    "href": "https://fidoalliance.org/specs/fido-u2f-v1.2-ps-20170411/fido-security-ref-v1.2-ps-20170411.html",
     "status": "FIDO Alliance Proposed Standard"
   },
 
--- a/index.bs
+++ b/index.bs
@@ -347,7 +347,7 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
     those [=ceremonies=].
 
 : <dfn>Client</dfn>
-:: See [=Conforming User Agent=].
+:: See [=WebAuthn Client=], [=Conforming User Agent=].
 
 : <dfn>Client-Side</dfn>
 :: This refers in general to the combination of the user's platform device, user agent, authenticators, and everything gluing
@@ -509,7 +509,7 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
 :: Upon successful completion of a [=user verification=] process, the user is said to be "[=user verified|verified=]".
 
 : <dfn>[WAC]</dfn>
-:: Also referred to herein as simply a [=client=]. See also [=Conforming User Agent=].
+:: Also referred to herein as simply a [=client=]. See also [=Conforming User Agent=]. A [=[WAC]=] is an intermediary entity typically implemented in the user agent (in whole, or in part). Conceptually, it underlies the [=Web Authentication API=] and embodies the implementation of the {{PublicKeyCredential/[[Create]](origin, options, sameOriginWithAncestors)}} and {{PublicKeyCredential/[[DiscoverFromExternalSource]](origin, options, sameOriginWithAncestors)}} [=internal methods=]. It is responsible for both marshalling the inputs for the underlying [=authenticator operations=], and for returning the results of the latter operations to the [=Web Authentication API=]'s callers.
 
 
 # <dfn>Web Authentication API</dfn> # {#api}
@@ -2144,9 +2144,9 @@ Authenticators:
 - should ensure that the [=signature counter=] value does not 
     accidentally decrease  (e.g., due to hardware failures).
 
-## Authenticator operations ## {#authenticator-ops}
+## <dfn>Authenticator operations</dfn> ## {#authenticator-ops}
 
-A client must connect to an authenticator in order to invoke any of the operations of that authenticator. This connection
+A [=[WAC]=] must connect to an authenticator in order to invoke any of the operations of that authenticator. This connection
 defines an <dfn>authenticator session</dfn>. An authenticator must maintain isolation between sessions. It may do this by only allowing one
 session to exist at any particular time, or by providing more complicated session management.
 
@@ -4211,8 +4211,18 @@ handled on the server side and do not need support from the API specified here.
 
 # Security Considerations # {#security-considerations}
 
+This specification defines a Web API and a cryptographic peer-entity authentication protocol.
+The [=Web Authentication API=] allows Web developers (i.e., "authors") to utilize the Web Authentication protocol in their
+[=registration=] and [=authentication=] [=ceremonies=].
+The entities comprising the Web Authentication protocol endpoints are user-controlled [=authenticators=] and a [=[RP]=]'s
+computing environment hosting the [=[RP]=]'s web application.
+In this model, the user agent, together with the [=[WAC]=], comprise an intermediary between [=authenticators=] and [=[RPS]=].
+Additionally, [=authenticators=] can [=attestation|attest=] to [=[RPS]=] as to their provenance.
 
+At this time, this specification does not feature detailed security considerations. However, the [[FIDOSecRef]] document provides a security analysis which is overall applicable to this specfication.
+Also, the [[FIDOAuthnrSecReqs]] document suite defines [=authenticator=] security characteristics which are overall applicable for WebAuthn [=authenticators=].
 
+The below subsections comprise the current Web Authentication-specific security considerations.
 
 
 ## Cryptographic Challenges ## {#cryptographic-challenges}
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -4269,16 +4210,94 @@ handled on the server side and do not need support from the API specified here.
...
+Also, the [[FIDOAuthnrSecReqs]] document suite defines [=authenticator=] security characteristics which are overall applicable for WebAuthn [=authenticators=].
+
+The below subsections comprise the current Web Authentication-specific security considerations.
+
+
 ## Cryptographic Challenges ## {#cryptographic-challenges}
+
 As a cryptographic protocol, Web Authentication is dependent upon randomized challenges
-to avoid replay attacks. Therefore, both {MakePublicKeyCredentialOptions/challenge}}'s 
+to avoid replay attacks. Therefore, both {MakePublicKeyCredentialOptions/challenge}}'s
```
emlun: Missing `{`

index.bs
```diff
@@ -4269,16 +4210,94 @@ handled on the server side and do not need support from the API specified here.
...
 and {{PublicKeyCredentialRequestOptions/challenge}}'s value, MUST be randomly generated
```
emlun: Suggest removing the comma in "value, MUST".

index.bs
```diff
@@ -4269,16 +4210,94 @@ handled on the server side and do not need support from the API specified here.
...
-by the [=Relying Party=] in an environment they trust (e.g., on the server-side), and the challenge in the client's
+by [=[RPS]=] in an environment they trust (e.g., on the server-side), and the
+returned challenge value in the client's
 response must match what was generated. This should be done in a fashion that does not rely
 upon a client's behavior; e.g.: the Relying Party should store the challenge temporarily
```
emlun: `s/must/MUST/`, `s/should/SHOULD/g` maybe?

index.bs
```diff
@@ -4269,16 +4210,94 @@ handled on the server side and do not need support from the API specified here.
...
 until the operation is complete. Tolerating a mismatch will compromise the security
 of the protocol.
 
+## Attestation Security Considerations ## {#sec-attestation-security-considerations}
+
+### Attestation Certificate Hierarchy ### {#cert-hierarchy}
```
emlun: Ditto wrt RFC2119 terms in this section: "must", "recommended", "should"

index.bs
equalsJeffH: thx

index.bs
equalsJeffH: thx

index.bs
equalsJeffH: OK

index.bs
equalsJeffH: OK

equalsJeffH:
this ready to merge?

index.bs
```diff
@@ -4269,16 +4210,94 @@ handled on the server side and do not need support from the API specified here.
...
+
+A 3-tier hierarchy for attestation certificates is RECOMMENDED (i.e., Attestation Root, Attestation Issuing CA, Attestation
+Certificate). It is also RECOMMENDED that for each WebAuthn Authenticator device line (i.e., model), a separate issuing CA is
+used to help facilitate isolating problems with a specific version of a device.
+
+If the attestation root certificate is not dedicated to a single WebAuthn Authenticator device line (i.e., AAGUID), the AAGUID
```
agl: The previous paragraph recommends a 3-level certificate chain with a new intermediate for each model. This paragraph then implicitly recommends a 2-level chain with a separate _root_ for each model. These seem to be contradictory.

I would tend to recommend a 2-level chain (and would suggest putting the AAGUID in the certificate in all cases): I expect the FIDO MDS to be the metadata source for attestation information and, if a whole model (i.e. AAGUID) needs to be "revoked", then the status of that AAGUID as a whole can be changed in the MDS. (I'm presuming that AAGUID will become a key in the MDS in the future since CTAP2 is a FIDO spec.)

index.bs
```diff
@@ -4269,16 +4210,94 @@ handled on the server side and do not need support from the API specified here.
...
+SHOULD be specified in the attestation certificate itself, so that it can be verified against the [=authenticator data=].
+
+
+### Attestation Certificate and Attestation Certificate CA Compromise ### {#ca-compromise}
+
+When an intermediate CA or a root CA used for issuing attestation certificates is compromised, WebAuthn [=authenticator=]
+attestation keys are still safe although their certificates can no longer be trusted. A WebAuthn Authenticator manufacturer that
```
agl: Perhaps be more precise here?: attestations produced before the time of CA compromise can still be trusted.

index.bs
```diff
@@ -4269,16 +4210,94 @@ handled on the server side and do not need support from the API specified here.
...
+has recorded the public attestation keys for their devices can issue new attestation certificates for these keys from a new
+intermediate CA or from a new root CA. If the root CA changes, the [=[RPS]=]  MUST update their trusted root certificates
```
agl: > If the root CA changes, the [=[RPS]=]  MUST update their trusted root certificates accordingly

Does that make sense as a MUST? Most RPs probably won't be checking attestation at all. I think all we can say is that RPs which wish to validate the update attestation statements will obviously need to trust the new root.

index.bs
```diff
@@ -4269,16 +4210,94 @@ handled on the server side and do not need support from the API specified here.
...
+accordingly.
+
+A WebAuthn Authenticator attestation certificate MUST be revoked by the issuing CA if its key has been compromised. A WebAuthn
```
agl: What does revoked mean here? There are two X.509 revocation mechanisms but I don't believe that either are required to be supported (and, in practice, are not supported). Also, the X.509 revocation mechanisms can't revoke a root.

index.bs
```diff
@@ -4269,16 +4210,94 @@ handled on the server side and do not need support from the API specified here.
...
+# Privacy Considertations # {#sctn-privacy-considerations}
+
+## Attestation Privacy ## {#sec-attestation-privacy}
+
+Attestation keys may be used to track users or link various online identities of the same user together. This may be mitigated
+in several ways, including:
+
+- A WebAuthn [=authenticator=] manufacturer may choose to ship all of their devices with the same (or a fixed number of)
+    attestation key(s) (called [=Basic Attestation=]). This will anonymize the user at the risk of not being able to revoke a
+    particular attestation key should its WebAuthn Authenticator be compromised.
```
agl: Completely your call, but maybe s/WebAuthn Authenticator/private key/

index.bs
equalsJeffH: Ok, so perhaps this ought to be a separate issue, which we submit and address separately from this PR. This is because I only moved the above text to this new section, and @rlin1 needs to weigh-in since AFAIK he wrote the above and has been the primary editor of the Metadata Statement and Metadata Service specs. 

WDYT?

index.bs
equalsJeffH: ditto - this is separate issue, and @rlin1 ought to weigh-in.

index.bs
equalsJeffH: ditto - this is separate issue, and @rlin1 ought to weigh-in.

index.bs
equalsJeffH: ditto - this is separate issue, and @rlin1 ought to weigh-in.

index.bs
equalsJeffH: good point, perhaps:
...should its WebAuthn Authenticator be compromised, i.e., its private key is revealed. 

..?

equalsJeffH:
thanks @agl for the detailed review of these sec cons.  However, the subsections you've commented on are only being moved to this new sec cons section by this PR.  Perhaps the issues you are identifying ought to be distinct ones, which we submit and address separately from this PR?   I.e., if there are not issues with the new text I've written (the intro to the new Sec Cons section, definition of webauthn client), then we go ahead and merge this and address the detail-level issues you identify separately. 

Also, @rlin1 ought to weigh-in since AFAIK he wrote those subsections and has been the primary editor of the Metadata Statement and Metadata Service specs.

WDYT?

agl:
I agree that, if this is just moving code, we don't need to fix these issues in this change.

index.bs
agl: (Whatever you feel best here is good with me.)

index.bs
```diff
@@ -4269,16 +4210,94 @@ handled on the server side and do not need support from the API specified here.
...
 ## Cryptographic Challenges ## {#cryptographic-challenges}
+
 As a cryptographic protocol, Web Authentication is dependent upon randomized challenges
-to avoid replay attacks. Therefore, both {MakePublicKeyCredentialOptions/challenge}}'s 
-and {{PublicKeyCredentialRequestOptions/challenge}}'s value, MUST be randomly generated
-by the [=Relying Party=] in an environment they trust (e.g., on the server-side), and the challenge in the client's
-response must match what was generated. This should be done in a fashion that does not rely
-upon a client's behavior; e.g.: the Relying Party should store the challenge temporarily
+to avoid replay attacks. Therefore, both {{MakePublicKeyCredentialOptions/challenge}}'s
+and {{PublicKeyCredentialRequestOptions/challenge}}'s value MUST be randomly generated
```
leshi: I don't think "randomly generated" is the right term here.  For example, it might be the encrypted server state of a session.  It's not random, but serves as a good challenge because the server state includes a random nonce.

equalsJeffH: @leshi: pls submit issue regarding the above after this is merged? thx.

Issue #420, opened 2017-04-21
attachment is only explicitly used in create()


equalsJeffH:
The description of `MakeCredentialOptions.attachment` (of type `Attachment`) is:
> This member contains authenticator attachment descriptions, which are used as an additional constraint on which authenticators are eligible to participate in a create() or get() operation.

Also, in [the description of the `Attachment` type](https://w3c.github.io/webauthn/#attachment), there is this first brief example in the final parag:
> ...there are use-cases where only platform authenticators are acceptable to a Relying Party...As a concrete example of [this], a credential on a platform authenticator may be used by Relying Parties to quickly and conveniently reauthenticate the user with a minimum of friction, e.g., the user will not have to dig around in their pocket for their key fob or phone.

Finally, there is no option/parameter of type `Attachment` provided to `ScopedCredential::[[DiscoverFromExternalSource]](options)` (see: https://w3c.github.io/webauthn/#getAssertion)

equalsJeffH:
[ oops, did not mean to hit [Comment] yet, sorry, so continuing with the issue submission here, I've also edited the initial description, so please do not rely on the emailed copy ]

Is the example (above) expected to work (a) due to the RP knowing, from credential creation, that it has a cred for the user backed by a platform-attached authnr and thus is able to place just that one cred  in the `options.allowList` in the call to  `ScopedCredential::[[DiscoverFromExternalSource]](options)` (aka get()) ?

I.e., rather than, (b) being able to state "use any cred registered with this RP that has an attachment of `platform`" ?

I do not think we necessarily need to have (b), but if (a) is the expected way for the use case to work we may want to clarify that in the spec.

nadalin:
@equalsJeffH please verify

equalsJeffH:
Yes, this remains a valid issue -- need feedback from @jyasskin @christiaanbrand @emlun @AngeloKai @akshayku wrt is the answer (a) or (b) ?

emlun:
For what it's worth **(a)** is not possible in the first factor use case (no `allowCredentials` given to `credentials.get()`), so we need this if we want to support attachment selection there.

[...]

PR #708, opened 2017-12-06
#420: Add note on why authenticator attachment is used only in create()

emlun:
See #420.


equalsJeffH:
Reviewed: https://github.com/w3c/webauthn/pull/708#pullrequestreview-81634828
LGTM. Thanks @emlun!

agl:
(I was asked to review this during last week's call. Looks good to me.)

akshayku:
Can someone explain to me how an RP who only wants to talk to, lets say, platform authenticators can achieve this without attachment option in assertion?

equalsJeffH:
discussed @akshayku's comment on the 20-Dec-2017 call and obtained consensus to merge this.

PR #709, opened 2017-12-06
move the credentialId uniqueness handling to the formal alg steps. 

### LINKED ISSUE #579: Credential ID uniqueness expectations are inconsistent/vague
As pointed out in #558, the requirements on the uniqueness of the credential ID are not completely clear.

[§4.1. PublicKeyCredential Interface][att-resp] reads:

>[[identifier]]
>[...] This identifier is used to look up credentials for use, and is therefore **expected to be globally unique with high probability** across all credentials of the same type, across all authenticators. This API does not constrain the format or length of this identifier, except that it **must be sufficient for the platform to uniquely select a key**. [...]

[§4.2.1 Information about Public Key Credential (interface AuthenticatorAttestationResponse)][att-resp], point `attestationObject` reads:

>[...] [the authenticator data] contains the AAGUID, **a unique credential ID**, and the credential public key. [...]

[§5. WebAuthn Authenticator model][authenticator-model] reads:

>[...] Each public key credential has an identifier which is **unique (or extremely unlikely to be duplicated) among all public key credentials**. [...]

[§5.2.1. The authenticatorMakeCredential operation][auth-make-cred] reads:

>When this operation is invoked, the authenticator must perform the following procedure:
>[...]
>Generate an identifier for this credential, such that this identifier is **globally unique with high probability** across all credentials with the same type across all authenticators.

(All emphasis added)

In summary, the uniqueness of the credential ID is specified as

- "expected to be" probably globally unique for all public key credentials
- informally required to be unique with undefined scope
- informally required to be probably globally unique
- formally required to be probably globally unique
- formally required only to be unique among the public key credentials available at a given time when one is needed

I suggest that

 1. the requirements on credential ID uniqueness be specified only once - for example in §4.1 where it is defined or in §5.2.1 which defines how it should be generated.
 2. ~~a concrete requirement is given instead of "with a high probability" - for example "with a probability greater than 1 - 2^150" (~~this value~~ (edit: 150 bits of entropy) would mean a collision chance < 1E-9 at 1E18 credential IDs generated).~~

[att-resp]: https://www.w3.org/TR/webauthn/#iface-authenticatorattestationresponse
[auth-make-cred]: https://www.w3.org/TR/webauthn/#op-make-cred
[authenticator-model]: https://www.w3.org/TR/webauthn/#authenticator-model
[id-attr]: https://www.w3.org/TR/webauthn/#dom-publickeycredential-identifier-slot

rlin1:
Close #579

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/pull/709.html" title="Last updated on Dec 14, 2017, 8:47 AM GMT (f577291)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/709/986d627...f577291.html" title="Last updated on Dec 14, 2017, 8:47 AM GMT (f577291)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2674,7 +2674,7 @@ When registering a new credential, represented by a {{AuthenticatorAttestationRe
 
 1. Check that the <code>[=credentialId=]</code> is not yet registered to any other user. If registration 
     is requested for a credential that is already registered to a different user, the [=[RP]=] SHOULD 
-    fail this ceremony, or it MAY decide to accept the registration, e.g. while deleting the older registration.
+    fail this [=registration=] ceremony, or it MAY decide to accept the registration, e.g. while deleting the older registration.
 
 1. If the attestation statement |attStmt| verified successfully and is found to be trustworthy, then register the new
     credential with the account that was denoted in the
--- a/index.bs
+++ b/index.bs
@@ -2639,61 +2639,65 @@ When registering a new credential, represented by a {{AuthenticatorAttestationRe
     {{AuthenticatorAttestationResponse}} object to extract the [=client data=] |C| claimed as collected during the credential
     creation.
 
-2. Verify that the {{CollectedClientData/type}} in |C| is the string `webauthn.create`.
+1. Verify that the {{CollectedClientData/type}} in |C| is the string `webauthn.create`.
 
-2. Verify that the {{CollectedClientData/challenge}} in |C| matches the challenge that was sent to the authenticator in the
+1. Verify that the {{CollectedClientData/challenge}} in |C| matches the challenge that was sent to the authenticator in the
     {{CredentialsContainer/create()}} call.
 
-3. Verify that the {{CollectedClientData/origin}} in |C| matches the [=[RP]=]'s [=origin=].
+1. Verify that the {{CollectedClientData/origin}} in |C| matches the [=[RP]=]'s [=origin=].
 
-4. Verify that the {{CollectedClientData/tokenBindingId}} in |C| matches the [=Token Binding ID=] for the TLS connection over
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -2639,61 +2639,65 @@ When registering a new credential, represented by a {{AuthenticatorAttestationRe
...
         - If [=self attestation=] was used, check if [=self attestation=] is acceptable under [=[RP]=] policy.
         - If [=ECDAA=] was used, verify that the [=identifier of the ECDAA-Issuer public key=] used is included in the set of
             acceptable trust anchors obtained in step 11.
         - Otherwise, use the X.509 certificates returned by the verification procedure to verify that the attestation public key
             correctly chains up to an acceptable root certificate.
 
-13. If the attestation statement |attStmt| verified successfully and is found to be trustworthy, then register the new
+1. Check that the <code>[=credentialId=]</code> is not yet registered to any other user. If registration 
+    is requested for a credential that is already registered to a different user, the [=[RP]=] SHOULD 
+    fail this ceremony, or it MAY decide to accept the registration, e.g. while deleting the older registration.
```
equalsJeffH: s/this ceremony /this [=registration=] ceremony /

equalsJeffH:
reviewed: https://github.com/w3c/webauthn/pull/709#pullrequestreview-81694373
LGTM
assigned to CR milestone.

rlin1:
Just waiting for feedback from @AngeloKai

index.bs
rlin1: Done.

rlin1:
Waiting for feedback from @akshayku instead.

PR #710, opened 2017-12-06
Specify that SHA-256 is used for hashing the client data.

### LINKED ISSUE #362: Crypto algorithm agility: e.g., Specify the set of hash algorithms UAs can select between.
Simply requiring a ["recognized algorithm name"](https://www.w3.org/TR/WebCryptoAPI/#recognized-algorithm-name) isn't enough because arbitrary other specifications can [define an algorithm](https://www.w3.org/TR/WebCryptoAPI/#concept-define-an-algorithm), while relying parties need to know which hash algorithms they need to be able to use to verify signatures.

@equalsJeffH suggested limiting to {SHA-256, SHA-384, SHA-512} in https://github.com/w3c/webauthn/pull/347#discussion_r103040671. There's been mention of [SM3](http://www.oscca.gov.cn/UpFile/20101222141857786.pdf) in the spec, but it's not currently defined in [WebCrypto](https://www.w3.org/TR/WebCryptoAPI/). It could be defined somewhere appropriate and re-added to the set of allowed algorithms in this spec later, if folks want to do so.

agl:
Previously, the client could choose any "recognized algorithm name" for
hashing, but RPs need to know which hash function(s) to support.

Rather than choose a set of hash functions and add implementation burden
for no clear reason, this change specifies that SHA-256 will be used.

Should we need to revisit this, we can spin a new version of the spec:
the RP can signal support for other algorithms via
`PublicKeyCredentialType` and the hashAlgorithm member can be readded to
the JSON to indicate when a new hash function was added.

Fixes #362.

***
<a href="https://pr-preview.s3.amazonaws.com/agl/webauthn/pull/710.html" title="Last updated on Dec 8, 2017, 4:28 AM GMT (eeac8d1)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/710/12f2d09...agl:eeac8d1.html" title="Last updated on Dec 8, 2017, 4:28 AM GMT (eeac8d1)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -3237,7 +3237,7 @@ This attestation statement format is used with FIDO U2F authenticators using the
             confirm its size to be of 32 bytes and concatenate it with |publicKeyU2F|.
             If size differs or "-3" key is not found, terminate this algorithm and return an appropriate error. 
     5. Let |verificationData| be the concatenation of (0x00 || |rpIdHash| || 
-        |tbsHash| || |credentialId| || |publicKeyU2F|) (see Section 4.3 of [[!FIDO-U2F-Message-Formats]]).
+        |clientDataHash| || |credentialId| || |publicKeyU2F|) (see Section 4.3 of [[!FIDO-U2F-Message-Formats]]).
     6. Verify the |sig| using |verificationData| and |certificate public key| per [[!SEC1]].
     7. If successful, return attestation type Basic with the [=attestation trust path=] set to |x5c|.
 
--- a/index.bs
+++ b/index.bs
@@ -3204,27 +3204,26 @@ This attestation statement format is used with FIDO U2F authenticators using the
 : Signing procedure
 :: If the credential public key of the given credential is not of algorithm -7 ("ES256"), stop and return an error.
     Otherwise, let |authenticatorData| denote the [=authenticator data for the attestation=],
-    and let |tbsHash| denote the [=hash of the serialized client data=]. (Since SHA-256 is used to hash the
-    serialized client data, |tbsHash| will be 32 bytes long.)
+    and let |clientDataHash| denote the [=hash of the serialized client data=]. (Since SHA-256 is used to hash the
+    serialized client data, |clientDataHash| will be 32 bytes long.)
 
     Generate a Registration Response Message as specified in [[FIDO-U2F-Message-Formats]] section 4.3, with the application parameter set to the
-    SHA-256 hash of the [=RP ID=] associated with the given credential, the challenge parameter set to |tbsHash|, and the key handle
+    SHA-256 hash of the [=RP ID=] associated with the given credential, the challenge parameter set to |clientDataHash|, and the key handle
     parameter set to the [=credential ID=] of the given credential. Set the raw signature part of this Registration Response Message (i.e., without the user public key,
     key handle, and attestation certificates) as |sig| and set the attestation certificates of
     the attestation public key as |x5c|.
 
 : Verification procedure
-:: Given the [=verification procedure inputs=] |attStmt|, |authenticatorData| and
-    [=hash of the serialized client data=], the verification procedure is as follows:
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -3233,17 +3214,16 @@ This attestation statement format is used with FIDO U2F authenticators using the
     the attestation public key as |x5c|.
 
 : Verification procedure
-:: Given the [=verification procedure inputs=] |attStmt|, |authenticatorData| and |clientDataHash|, the verification procedure is
-    as follows:
+:: Given the [=verification procedure inputs=] |attStmt|, |authenticatorData| and
+    [=hash of the serialized client data=], the verification procedure is as follows:
```
emlun: I don't think this needs to change, does it? On the other hand, the `|tbsHash|` intermediate step is now unnecessary. I therefore suggest reverting this change, eliminating step 4, and changing `|tbsHash|` to `|clientDataHash|` in step 6. You're welcome to ignore this if you disagree.

index.bs
```diff
@@ -3233,17 +3214,16 @@ This attestation statement format is used with FIDO U2F authenticators using the
...
     1. Verify that |attStmt| is valid CBOR conforming to the syntax defined above, and perform CBOR decoding on it to extract the
         contained fields.
     2. Let |attCert| be value of the first element of |x5c|. Let |certificate public key| be the public key
         conveyed by |attCert|. If |certificate public key| is not an Elliptic Curve (EC) public 
         key over the P-256 curve, terminate this algorithm and return an appropriate error. 
     3. Extract the claimed |rpIdHash| from |authenticatorData|, and the claimed |credentialId| and |credentialPublicKey| from
         |authenticatorData|.<code>[=attestedCredentialData=]</code>.
-    4. If |clientDataHash| is 256 bits long, set |tbsHash| to this value. Otherwise set |tbsHash| to the SHA-256
-        hash of |clientDataHash|.
+    4. Set |tbsHash| to the [=hash of the serialized client data=].
```
emlun: See previous comment.

index.bs
```diff
@@ -3220,11 +3203,9 @@ This attestation statement format is used with FIDO U2F authenticators using the
...
 : Signing procedure
 :: If the credential public key of the given credential is not of algorithm -7 ("ES256"), stop and return an error.
-    Otherwise, let |authenticatorData| denote the [=authenticator data for the attestation=], 
-    and let |clientDataHash| denote the [=hash of the serialized client data=].
-
-    If |clientDataHash| is 256 bits long, set |tbsHash| to this value. Otherwise set |tbsHash| to the SHA-256 hash of
-    |clientDataHash|.
+    Otherwise, let |authenticatorData| denote the [=authenticator data for the attestation=],
+    and let |tbsHash| denote the [=hash of the serialized client data=]. (Since SHA-256 is used to hash the
+    serialized client data, |tbsHash| will be 32 bytes long.)
```
emlun: If you do rename `|tbsHash|` to `|clientDataHash|` as suggested in the next comment, it's probably good to do that here too.

agl:
> Looks good to me, just a couple of minor style nitpicks.

Thanks. I think your comments make sense and I've hopefully aligned this with what you had in mind.

index.bs
```diff
@@ -3237,10 +3236,10 @@ This attestation statement format is used with FIDO U2F authenticators using the
             (representing y coordinate) from |credentialPublicKey|, 
             confirm its size to be of 32 bytes and concatenate it with |publicKeyU2F|.
             If size differs or "-3" key is not found, terminate this algorithm and return an appropriate error. 
-    6. Let |verificationData| be the concatenation of (0x00 || |rpIdHash| || 
+    5. Let |verificationData| be the concatenation of (0x00 || |rpIdHash| || 
         |tbsHash| || |credentialId| || |publicKeyU2F|) (see Section 4.3 of [[!FIDO-U2F-Message-Formats]]).
```
emlun: `|tbsHash|` here needs to be updated to `|clientDataHash|`.

PR #723, opened 2017-12-08
Rework the FIDO AppID extension.

### LINKED ISSUE #491: various issues with AppId extension
There are various issues with the AppId extension as presently written:

1.  why is [the appId extension](https://w3c.github.io/webauthn/#sctn-appid-extension) used only during [get()](https://w3c.github.io/webappsec-credential-management/#dom-credentialscontainer-get) (which in this case is impl'd by [PublicKeyCredential’s  \[\[DiscoverFromExternalSource\]\](options) method)](https://w3c.github.io/webauthn/#getAssertion)?  If browsers will initially be using U2F authenticators (authnrs) underneath WebAuthn, will they not need to pass an appId down to the authnr during [create()](https://w3c.github.io/webappsec-credential-management/#dom-credentialscontainer-create)  (which in this case is impl'd by [PublicKeyCredential’s \[\[Create\]\](options) method](https://w3c.github.io/webauthn/#createCredential)) ?  

    The description of the appid extension says:
    > This authentication extension allows Relying Parties that have previously registered a credential using the legacy FIDO JavaScript APIs to request an assertion.

     But if the initial authnrs are U2F, will they not also need to perform [registration](https://fidoalliance.org/specs/fido-u2f-v1.1-id-20160915/fido-u2f-raw-message-formats-v1.1-id-20160915.html#registration-messages)? 

2.  the rpId is not syntactically the same (at this time) as the appId. The RP ID is "a [valid domain string](https://url.spec.whatwg.org/#valid-domain-string)", while the appid is a [serialized origin](https://html.spec.whatwg.org/#ascii-serialisation-of-an-origin).  So one cannot simply overwrite the RP ID (at this time) with an appid.

3.  "client extension processing" references step 3 of [{#getAssertion}](https://w3c.github.io/webauthn/#getAssertion)   though I believe that is now step 7 of the latter alg. 

4.  "client extension processing" says in part:
    > Replace the calculation of rpId in [Step 7 of {#getAssertion}](https://w3c.github.io/webauthn/#GetAssn-DetermineRpId) with the following procedure: The client uses the value of appid to perform the AppId validation procedure (as defined by [FIDO-APPID]). If valid, the value of rpId for all client processing should be replaced by the value of appid. 

    But we cannot do simply that because of the syntactic differences between rpId and appid.

(off-the-cuff unvalidated) solution options:

S1.  alter RP ID syntax to be serialized origin

S2.  have appid extn add additional "appid" dictionary item to the options passed into both [[Create]](options) and [[DiscoverFromExternalSource]](options)

agl:
This change clarifies the the behaviour of the `appid` client extension
and removes the client extension output.

Fixes #491.

***
<a href="https://pr-preview.s3.amazonaws.com/agl/webauthn/pull/723.html" title="Last updated on Jan 4, 2018, 11:39 PM GMT (6c9d529)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/723/35b730b...agl:6c9d529.html" title="Last updated on Jan 4, 2018, 11:39 PM GMT (6c9d529)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -110,6 +110,11 @@ spec: page-visibility; urlPrefix: https://www.w3.org/TR/page-visibility/
 spec: WHATWG HTML; urlPrefix: https://html.spec.whatwg.org/
     type: dfn 
         text: focus
+
+spec: FIDO-APPID; urlPrefix: https://fidoalliance.org/specs/fido-u2f-v1.2-ps-20170411/fido-appid-and-facets-v1.2-ps-20170411.html
+    type: dfn
+        text: determining the FacetID of a calling application; url: determining-the-facetid-of-a-calling-application
+        text: determining if a caller's FacetID is authorized for an AppID; url: determining-if-a-caller-s-facetid-is-authorized-for-an-appid
 </pre> <!-- class=anchors -->
 
 <!-- L128 spec:webappsec-credential-management-1; type:dictionary; for:/; text:CredentialRequestOptions -->
@@ -284,6 +289,11 @@ below and in [[#index-defined-elsewhere]].
 :: Many of the interface definitions and all of the IDL in this specification depend on [[!WebIDL-1]]. This updated version of
     the Web IDL standard adds support for {{Promise}}s, which are now the preferred mechanism for asynchronous
     interaction in all new web APIs.
+
+: FIDO AppID
+:: The algorithms for [=determining the FacetID of a calling application=] and
+    [=determining if a caller's FacetID is authorized for an AppID=] (used only in
+    the `appid` extension) are defined by [[!FIDO-APPID]].
 
 The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED",  "MAY", and
 "OPTIONAL" in this document are to be interpreted as described in [[!RFC2119]].
@@ -3497,33 +3507,46 @@ IANA "WebAuthn Extension Identifier" registry established by [[!WebAuthn-Registr
 These are recommended for implementation by user agents targeting broad interoperability.
 
 
-## FIDO AppId Extension (appid) ## {#sctn-appid-extension}
+## FIDO AppID Extension (appid) ## {#sctn-appid-extension}
 
-This [=authentication extension=] allows [=[RPS]=] that have previously registered a
-credential using the legacy FIDO JavaScript APIs to request an assertion.
-Specifically, this extension allows [=[RPS]=] to specify an |appId| [[FIDO-APPID]]
-to overwrite the otherwise computed |rpId|.  This extension is only valid if
-used during the {{CredentialsContainer/get()}} call; other usage will result in client
-error.
+This [=client extension=] allows [=[RPS]=] that have previously registered a
+credential using the legacy FIDO JavaScript APIs to request an assertion. The
+FIDO APIs use an alternative identifier for [=relying parties=] called an |AppID|
+[[FIDO-APPID]], and any credentials created using those APIs will be bound to
+that identifier. Without this extension they would need to be re-registered in
+order to be bound to an [=RP ID=].
+
+This extension does not allow FIDO-compatible credentials to be created. Thus
+credentials created with WebAuthn cannot be backwards compatible with the FIDO
+JavaScript APIs.
 
 : Extension identifier
 :: `appid`
 
 : Client extension input
-:: A single JSON string specifying a FIDO |appId|.
+:: A single JSON string specifying a FIDO |AppID|.
 
 : Client extension processing
-:: If {{PublicKeyCredentialRequestOptions/rpId}} is present, return a DOMException
-    whose name is "{{NotAllowedError}}", and terminate this algorithm ([[#discover-from-external-source]]).
-
-    Otherwise, replace the calculation of |rpId| in Step 6 of [[#discover-from-external-source]] with the
-    following procedure:  The client uses the value of |appid| to perform
-    the AppId validation procedure (as defined by [[FIDO-APPID]]).  If valid,
-    the value of |rpId| for all client processing should be replaced by the
-    value of |appid|.
+::  1. If present in a {{CredentialsContainer/create()}} call, return a
+        "{{NotSupportedError}}" {{DOMException}}—this extension is only valid when
+        requesting an assertion.
+    1. Let |facetId| be the result of passing the caller's [=origin=] to the
+        FIDO algorithm for [=determining the FacetID of a calling application=].
+    1. Let |appId| be the extension input.
+    1. Pass |facetId| and |appId| to the FIDO algorithm for [=determining if a
+        caller's FacetID is authorized for an AppID=]. If that algorithm rejects
+        |appId| then return a "{{SecurityError}}" {{DOMException}}.
+    1. When building |allowCredentialDescriptorList|, if a U2F authenticator
+        indicates that a credential is inapplicable (i.e. by returning `SW_WRONG_DATA`)
+        then the client may retry with the U2F application parameter set to the SHA-256
+        hash of |appId|. If this results in an applicable credential, the client may
+        include the credential in |allowCredentialDescriptorList|. The value of |appId|
+        then becomes an additional input parameter for [=authenticatorGetAssertion=] to
+        allow the credential to be used by replacing `rp.id` when calcuating
+        `rpIdHash`. (See section 7.2 of [[!FIDO-CTAP]].)
 
 : Client extension output
-:: Returns the JSON value `true` to indicate to the RP that the extension was acted upon
+:: None.
 
 : Authenticator extension input
 :: None.
--- a/index.bs
+++ b/index.bs
@@ -1173,7 +1173,7 @@ When this method is invoked, the user agent MUST execute the following algorithm
     1. Let |userPresence| be a Boolean value set to the inverse of |userVerification|.
 
     1. <span id="allowCredentialDescriptorListCreation"></span>
-       If <code>|options|.{{PublicKeyCredentialRequestOptions/allowCredentials}}</code>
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -3497,33 +3507,46 @@ IANA "WebAuthn Extension Identifier" registry established by [[!WebAuthn-Registr
...
+order to be bound to an [=RP ID=].
 
-This [=authentication extension=] allows [=[RPS]=] that have previously registered a
-credential using the legacy FIDO JavaScript APIs to request an assertion.
-Specifically, this extension allows [=[RPS]=] to specify an |appId| [[FIDO-APPID]]
-to overwrite the otherwise computed |rpId|.  This extension is only valid if
-used during the {{CredentialsContainer/get()}} call; other usage will result in client
-error.
+This extension does not allow FIDO-compatible credentials to be created. Thus
+credentials created with WebAuthn cannot be backwards compatible with the FIDO
```
equalsJeffH: s/cannot be/are not/  ..?

index.bs
```diff
@@ -3497,33 +3507,46 @@ IANA "WebAuthn Extension Identifier" registry established by [[!WebAuthn-Registr
...
+::  1. If present in a {{CredentialsContainer/create()}} call, return a
+        "{{NotSupportedError}}" {{DOMException}}—this extension is only valid when
+        requesting an assertion.
+    1. Let |facetId| be the result of passing the caller's [=origin=] to the
+        FIDO algorithm for [=determining the FacetID of a calling application=].
+    1. Let |appId| be the extension input.
+    1. Pass |facetId| and |appId| to the FIDO algorithm for [=determining if a
+        caller's FacetID is authorized for an AppID=]. If that algorithm rejects
+        |appId| then return a "{{SecurityError}}" {{DOMException}}.
+    1. When building |allowCredentialDescriptorList|, if a U2F authenticator
```
equalsJeffH: does "When building |allowCredentialDescriptorList|" refer to [\[\[DiscoverFromExternalSource\]\]](https://w3c.github.io/webauthn/#discover-from-external-source) step 17.4.isNotEmpty.2 ?  If so, i suggest adding an anchor there and explicitly linking from here to that alg step.

index.bs
```diff
@@ -3497,33 +3507,46 @@ IANA "WebAuthn Extension Identifier" registry established by [[!WebAuthn-Registr
...
+        indicates that a credential is inapplicable (i.e. by returning `SW_WRONG_DATA`)
+        then the client may retry with the U2F application parameter set to the SHA-256
+        hash of |appId|. If this results in an applicable credential, the client may
```
equalsJeffH: s/may/MAY/

index.bs
```diff
@@ -3497,33 +3507,46 @@ IANA "WebAuthn Extension Identifier" registry established by [[!WebAuthn-Registr
...
+        include the credential in |allowCredentialDescriptorList|. The value of |appId|
+        then becomes an additional input parameter for [=authenticatorGetAssertion=] to
+        allow the credential to be used by replacing `rp.id` when calcuating
+        `rpIdHash`. (See section 7.2 of [[!FIDO-CTAP]].)
 
 : Client extension output
-:: Returns the JSON value `true` to indicate to the RP that the extension was acted upon
+:: None.
```
equalsJeffH: this returning of `true` was explicitly added to some of the extensions in the semi-distant past, so I'm thinking we ought to leave this as-is for now and submit  a separate issue against it if we believe that we do not need client extensions to return `true` in order to signal that they carried out their duties.

index.bs
```diff
@@ -3497,33 +3507,46 @@ IANA "WebAuthn Extension Identifier" registry established by [[!WebAuthn-Registr
...
+    1. Let |appId| be the extension input.
+    1. Pass |facetId| and |appId| to the FIDO algorithm for [=determining if a
+        caller's FacetID is authorized for an AppID=]. If that algorithm rejects
+        |appId| then return a "{{SecurityError}}" {{DOMException}}.
+    1. When building |allowCredentialDescriptorList|, if a U2F authenticator
+        indicates that a credential is inapplicable (i.e. by returning `SW_WRONG_DATA`)
+        then the client may retry with the U2F application parameter set to the SHA-256
+        hash of |appId|. If this results in an applicable credential, the client may
+        include the credential in |allowCredentialDescriptorList|. The value of |appId|
+        then becomes an additional input parameter for [=authenticatorGetAssertion=] to
```
equalsJeffH: hm. so perhaps we ought to add an optional `appId` param to [authenticatorGetAssertion](https://w3c.github.io/webauthn/#op-get-assertion) ?

index.bs
```diff
@@ -3497,33 +3507,46 @@ IANA "WebAuthn Extension Identifier" registry established by [[!WebAuthn-Registr
...
+        allow the credential to be used by replacing `rp.id` when calcuating
+        `rpIdHash`. (See section 7.2 of [[!FIDO-CTAP]].)
```
equalsJeffH: hm, so we perhaps ought to similarly add an optional `appId` param to [section 7.2 of \[\[!FIDO-CTAP\]\] ](https://fidoalliance.org/specs/fido-v2.0-rd-20170927/fido-client-to-authenticator-protocol-v2.0-rd-20170927.html#using-the-ctap2-authenticatorgetassertion-command-with-ctap1-u2f-authenticators)?

index.bs
```diff
@@ -110,6 +110,11 @@ spec: page-visibility; urlPrefix: https://www.w3.org/TR/page-visibility/
 spec: WHATWG HTML; urlPrefix: https://html.spec.whatwg.org/
     type: dfn 
         text: focus
+
+spec: FIDO-APPID; urlPrefix: https://fidoalliance.org/specs/fido-u2f-v1.2-ps-20170411/fido-appid-and-facets-v1.2-ps-20170411.html
```
equalsJeffH: shall we leverage https://fidoalliance.org/specs/fido-v2.0-rd-20170927/fido-client-to-authenticator-protocol-v2.0-rd-20170927.html instead? it is what's ref'd in the biblio.

equalsJeffH:
reviewed https://github.com/w3c/webauthn/pull/723#pullrequestreview-83941948

index.bs
emlun: There are no references to `rp.id` in [authenticatorGetAssertion][aga], and U2F devices use the AppID hash in place of the RP ID for everything, not only for computing the hash. [authenticatorGetAssertion][aga] uses its `rpId` parameter only for:

- Identifying which of the credentials in `allowCredentialDescriptorList` are eligible
- Prompting the user to confirm the operation

I therefore suggest we simply send `appId` as the `rpId` argument to [authenticatorGetAssertion][aga] instead of adding an additional parameter.

[aga]: https://pr-preview.s3.amazonaws.com/agl/webauthn/pull/723.html#authenticatorgetassertion

index.bs
emlun: I actually think this should be s/may/MUST/ in both instances.

index.bs
agl: That's CTAP2, which doesn't define an AppID nor FacetID. However, I do see that there's a `FIDO-APPID` reference already in the biblio and I'm not sure what is the right thing to do here. Based on other cases, I've deleted the current biblio entry because it'll get populated automatically based on this chunk.

index.bs

index.bs
agl: Not sure how to anchor a <li> when it's generated by Markdown. But I've added an empty `span`, which appears to work.

index.bs
agl: Done. (And below.)

index.bs
agl: Based on Emil's comment, I've just specified that `appId` replaces `rpId`. I think this is always going to be a little awkward since we're stitching together two different specs and don't want to burden webauthn with baggage. But implementations have a clear ground-truth here: FIDO credentials should continue to work, so I hope that eliminates any misunderstandings.

index.bs
agl: Elim suggests just replacing rpId with appId. Although that's a bit of a type error, it avoids polluting webauthn with too much U2F. Also, I'm not that that we can still update CTAP2.

index.bs

index.bs
agl: That seems simple, albeit a type error. So I've done that for now and I guess people can discuss whether that'll do on the call.

emlun: Oh, I didn't realize that would be a type error. What makes it so?

index.bs
equalsJeffH: I corrected my mistake in the prior comment. apologies.

equalsJeffH:
on 20-Dec-2017 call, @agl indicates that he is reticent to pollute the spec with legacy explicit appID stuff when we "can" overload the RP ID as an AppID. they are both strings, but syntactically distinct. something to think about.

emlun:
I think the format difference between the two won't matter in practice - except perhaps in how the client stores them internally, as @agl pointed out. But for interop purposes it should make no difference: The client will have both the RP ID and the AppID; non-U2F authenticators will get the RP ID as usual (and will never see a FIDO AppID); and U2F authenticators will get the SHA256 hash of the AppID and won't actually see the underlying AppID.

index.bs
```diff
@@ -3497,33 +3508,45 @@ IANA "WebAuthn Extension Identifier" registry established by [[!WebAuthn-Registr
...
+## FIDO AppID Extension (appid) ## {#sctn-appid-extension}
 
-This [=authentication extension=] allows [=[RPS]=] that have previously registered a
-credential using the legacy FIDO JavaScript APIs to request an assertion.
-Specifically, this extension allows [=[RPS]=] to specify an |appId| [[FIDO-APPID]]
-to overwrite the otherwise computed |rpId|.  This extension is only valid if
-used during the {{CredentialsContainer/get()}} call; other usage will result in client
-error.
+This [=client extension=] allows [=[RPS]=] that have previously registered a
+credential using the legacy FIDO JavaScript APIs to request an assertion. The
```
equalsJeffH: suggest: s/assertion/[=assertion=]/

PR #695, opened 2017-11-19
CDDL description of location extension

gmandyam:
Partially addresses https://github.com/w3c/webauthn/issues/679

***
<a href="https://pr-preview.s3.amazonaws.com/gmandyam/webauthn/pull/695.html" title="Last updated on Dec 6, 2017, 10:56 PM GMT (a4fe0fa)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/695/12f2d09...gmandyam:a4fe0fa.html" title="Last updated on Dec 6, 2017, 10:56 PM GMT (a4fe0fa)">Diff</a>

Issue #679, opened 2017-11-09
Add CDDL to every extension


jcjones:
The extensions currently describe their CBOR as example octets; these can be more precisely defined using CDDL, and we should add normative CDDL for each extension.

ttaubert:
None of the example octets for the currently listed extensions use indefinite length encoding. When we define them using CDDL, can we add a note that allows definite length encoding only? That should simplify CBOR parsers quite a bit.

jcjones:
Avoiding indefinite length encoding is a good security goal in general, whatever the encoding format. Per @ttaubert, I'd recommend we add that to the implementation considerations as part of this change.

kpaulh:
I agree. CTAP is already requiring their CBOR to be canonical, which means no indefinite length encoding.

selfissued:
There was agreement on the 28-Feb-18 call that CDDL isn't needed for most extensions, since the transformations between the JavaScript data structures and the CBOR data structures are now obvious.

PR #724, opened 2017-12-12
Adding CDDL to txAuthSimple

gmandyam:
***
<a href="https://pr-preview.s3.amazonaws.com/gmandyam/webauthn/pull/724.html" title="Last updated on Feb 6, 2018, 12:39 AM GMT (d6bba25)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/724/0d52835...gmandyam:d6bba25.html" title="Last updated on Feb 6, 2018, 12:39 AM GMT (d6bba25)">Diff</a>


index.bs
```diff
@@ -3555,13 +3558,27 @@ This [=registration extension=] and [=authentication extension=] allows for a si
 : Authenticator extension input
 :: The client extension input encoded as a CBOR text string (major type 3).
 
+<pre>
+CDDL:
+(
+txAuthSimple: tstr,
+)
+</pre>
+
```
equalsJeffH: tho headed in the correct direction IMO (see https://github.com/w3c/webauthn/issues/725#issuecomment-352139120).  

i am thinking this ought to be spec'd more similarly to how [#sctn-generic-txauth-extension](https://w3c.github.io/webauthn/#sctn-generic-txauth-extension) does it:

```
txAuthSimpleArg = ( tstr )  
```

index.bs
```diff
@@ -3555,13 +3558,27 @@ This [=registration extension=] and [=authentication extension=] allows for a si
...
 : Authenticator extension output
 :: A single CBOR string, representing the prompt as displayed (including any eventual line breaks).
 
+<pre>
+CDDL:
+(
+txAuthSimple: tstr,
+)
+</pre>
+
```
equalsJeffH: I'm not sure we need this CDDL representation of the extension output. If we were to use something like it, we'd want to go thru and apply the same treatment to all extension definitions IMO.

also I think the CDDL is not quite what we'd want, it'd be more like..
```
txAuthSimpleRetVal = ( tstr )  
```
...I think (?)

equalsJeffH:
reviewed: https://github.com/w3c/webauthn/pull/724#pullrequestreview-83953626

gmandyam:
@equalsJeffH 

Thanks for the review.  I have tried to address your comments, and I still am not sure we are there yet.

a) extidArg and extidRetVal conventions are fine, in my opinion.  They should be specified in https://www.w3.org/TR/webauthn/#sctn-extension-specification, but as far as I can tell they are not.

b) I didn't leverage your suggested CDDL notation exactly, because I thought it would be clearer to represent the values as CBOR maps.

c) Although it was not the original intent of this PR, should we be more explicit about what is expected in the client extension input?  "A single JSON string prompt" seems like it can be interpreted in multiple ways.  If you have a suggestion, I can incorporate it into the PR and take care of that as well.

nadalin:
@equalsJeffH Have you reviewed the changes ?

gmandyam:
@selfissued Requested change has been made.

nadalin:
@equalsJeffH @selfissued is this one that we just close

gmandyam:
@selfissued, @equalsJeffH, @nadalin 

As per https://github.com/w3c/webauthn/issues/679, I was assigned to provide CDDL for each extension.  My approach has been to do this one at a time.  It seems that there is an issue with my approach to this rather than the end goal.

selfissued:
It may be a goal to add CDDL to each extension, but if so, let's create a PR that does that.  Adding to some but not others, which is what this PR does, only makes the spec less parallel and more confusing.  As such, I think this one should never be merged.  It should therefore be closed.

gmandyam:
@selfissued 

Actually this is exactly the way I did it with loc:  see https://github.com/w3c/webauthn/pull/695.  The spec in this regard is already 'imbalanced'.

jcjones:
This can merge now that #765 is merged. We agree this is imbalanced, but the consensus of the WG is that we can just be sure to write CDDL for all the other extensions before publishing the next document.

jcjones:
(Note: When merging this, let's file issues for the remaining CDDL.)

equalsJeffH:
@gmandyam 
> I thought it would be clearer to represent the values as CBOR maps.

if using maps, is not the key "txAuthSimpleArg" also part of the CBOR-encoded object?  If so, do we really need to transport those bytes down the stack (do they add any value)?  The input arg for txAuthSimple is presently stated as:
> The client extension input encoded as a CBOR text string (major type 3).

..which is just a `tstr` with no accompanying "key" such as "txAuthSimpleArg", which would seem to be needlessly consuming bytes on-the-wire (?).

gmandyam:
@equalsJeffH 

>if using maps, is not the key "txAuthSimpleArg" also part of the CBOR-encoded object? If so, do we really need to transport those bytes down the stack (do they add any value)?

As per https://w3c.github.io/webauthn/#sctn-authenticator-extension-processing, "The CBOR authenticator extension input value of each processed authenticator extension is included in the extensions data part of the authenticator request. **_This part is a CBOR map_**, with CBOR extension identifier values as keys, and the CBOR authenticator extension input value of each extension as the value."

I interpreted the above to mean that the authenticator extension input is always presented to the authenticator as a CBOR map.  I agree that this may be inefficient in the case of txAuthSimple, but I don't know how else to interpret the above requirement.

equalsJeffH:
@gmandyam 
doh!  carry on then...

emlun:
@gmandyam @equalsJeffH 

Sorry for being late to the party, but I had read [§9.5][extproc] to mean that the _whole_ of the extensions authenticator input - i.e., the inputs for _all_ extensions - is a CBOR map, but the individual extension inputs are not necessarily CBOR maps. The structure of this CBOR map is then analogous to [`AuthenticationExtensionsClientInputs`][aeci], which seems more correct to me and also saves some bandwidth as @equalsJeffH [points out][comment]. [CTAP][ctap] also refers to [a previous WebAuthn draft][prev] with an example with such a structure. I think @equalsJeffH's previously suggested CDDL `txAuthSimpleRetVal = ( tstr )` is better in that case.

[extproc]: https://w3c.github.io/webauthn/#sctn-authenticator-extension-processing
[aeci]: https://w3c.github.io/webauthn/#dictdef-authenticationextensionsclientinputs
[comment]: https://github.com/w3c/webauthn/pull/724#issuecomment-362669638
[prev]: https://www.w3.org/TR/webauthn/#authenticator-extension-input
[ctap]: https://fidoalliance.org/specs/fido-v2.0-rd-20170927/fido-client-to-authenticator-protocol-v2.0-rd-20170927.html#authenticatorMakeCredential

gmandyam:
@emlun 

Please see the text I cited again, this time with a different emphasis:

As per https://w3c.github.io/webauthn/#sctn-authenticator-extension-processing, "The CBOR authenticator extension input value **_of each processed authenticator extension_** is included in the extensions data part of the authenticator request. **_This part is a CBOR map_**, with CBOR extension identifier values as keys, and the CBOR authenticator extension input value of each extension as the value."

I don't agree with your interpretation of the above text, because the CBOR map corresponds to "each processed authenticator extension", not to all collective extensions.

If this was not the intent of the cited text (which is something that the editors need to address), then they should strongly consider a revision of the above.

emlun:
This is how I read the grammar in that section - read the emphasized parts as one sentence:

>The CBOR authenticator extension input value of each processed authenticator extension is included in **the extensions data part** of the authenticator request. This part **is a CBOR map**, with CBOR extension identifier values as keys, and the CBOR authenticator extension input value of each extension as the value.

I agree this section needs to be revised regardless of which interpretation is deemed to be correct. I can take a stab at that once we reach consensus here.

equalsJeffH:
upon review I think @emlun is correct [here](https://github.com/w3c/webauthn/pull/724#issuecomment-363105368) & [here](https://github.com/w3c/webauthn/pull/724#issuecomment-363118688) (good catch, thx) -- getting @selfissued's input would be good since he did much work on the extensions mechanism.

index.bs
```diff
@@ -3555,13 +3555,27 @@ This [=registration extension=] and [=authentication extension=] allows for a si
 : Authenticator extension input
 :: The client extension input encoded as a CBOR text string (major type 3).
 
+<pre>
+CDDL:
+{
+txAuthSimpleArg: tstr,
```
selfissued: Please rename txAuthSimpleArg to txAuthSimpleInput

index.bs
```diff
@@ -3555,13 +3555,27 @@ This [=registration extension=] and [=authentication extension=] allows for a si
...
 :: The authenticator MUST display the prompt to the user before performing either [=user verification=] or [=test of user
     presence=]. The authenticator may insert line breaks if needed.
 
 : Authenticator extension output
 :: A single CBOR string, representing the prompt as displayed (including any eventual line breaks).
 
+<pre>
+CDDL:
+{
+txAuthSimpleRetVal: tstr,
```
selfissued: Please rename txAuthSimpleRetVal to txAuthSimpleOutput.

nadalin:
@gmandyam Where do we stand on this ?

gmandyam:
@nadalin 

I am fine with the suggestion by @emlun  and @equalsJeffH and @selfissued.  The latest changes take these into account.

I hope the text in https://w3c.github.io/webauthn/#sctn-authenticator-extension-processing will also be revised to better reflect the conclusions reached on this PR.

selfissued:
@emlun I'm not sure which section you're referring to with your comment "I agree this section needs to be revised regardless of which interpretation is deemed to be correct. I can take a stab at that once we reach consensus here."  But if you want to create a PR to make something clearer, I'll gladly review it.

nadalin:
@gmandyam So looks like a go to merge

PR #730, opened 2017-12-20
Fix #720: Align user handle management with CTAP

### LINKED ISSUE #720: Contradiction in whether user handle is required
The definition of [Public Key Credential Source][pkcs] includes:

>[...] A public key credential source has:
>  - [...]
> - An optional _user handle_ for the person who created this credential source.

However the user handle is _required_ in [PublicKeyCredentialUserEntity][pkcue]. We should drop the "optional" from the above bullet point.

[pkcue]: https://w3c.github.io/webauthn/#sctn-user-credential-params
[pkcs]: https://w3c.github.io/webauthn/#public-key-credential-source

emlun:
As stated in https://github.com/w3c/webauthn/pull/558#issuecomment-331537953. This resolves #720.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/pull/730.html" title="Last updated on Jan 3, 2018, 5:18 PM GMT (f3e8afb)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/730/33ac796...f3e8afb.html" title="Last updated on Jan 3, 2018, 5:18 PM GMT (f3e8afb)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -392,7 +392,7 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
     * A [=Credential ID=].
     * A [=credential private key=].
     * The [=Relying Party Identifier=] for the [=[RP]=] that created this credential source.
-    * An optional [=user handle=] for the person who created this credential source.
+    * A [=user handle=] for the person who created this credential source.
     * Optional other information used by the authenticator to inform its UI. For example, this might include the user's
         {{displayName}}.
 
--- a/index.bs
+++ b/index.bs
@@ -392,7 +392,7 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
     * A [=Credential ID=].
     * A [=credential private key=].
     * The [=Relying Party Identifier=] for the [=[RP]=] that created this credential source.
-    * A [=user handle=] for the person who created this credential source.
+    * An optional [=user handle=] for the person who created this credential source.
     * Optional other information used by the authenticator to inform its UI. For example, this might include the user's
         {{displayName}}.
 
--- a/index.bs
+++ b/index.bs
@@ -1245,8 +1245,8 @@ When this method is invoked, the user agent MUST execute the following algorithm
                 ::  whose value is the bytes of the signature value returned by the [=authenticator=].
 
                 :   <code><dfn for="assertionCreationData">userHandleResult</dfn></code>
-                ::  If |savedCredentialId| exists, set the value of [=userHandleResult=] to be the bytes of the [=user handle=]
-                    returned by the [=authenticator=]. Otherwise, set the value of [=userHandleResult=] to null.
+                ::  If the authenticator returned a [=user handle=], set the value of [=userHandleResult=] to be the bytes of that
+                    [=user handle=] returned by the [=authenticator=]. Otherwise, set the value of [=userHandleResult=] to null.
 
                 :   <code><dfn for="assertionCreationData">clientExtensionResults</dfn></code>
                 ::  whose value is an {{AuthenticationExtensions}} object containing [=extension identifier=] →
@@ -2340,16 +2340,14 @@ When this method is invoked, the [=authenticator=] must perform the following pr
     Return to the user agent:
         - |selectedCredential|'s [=credential ID=], if either a [=list=] of credentials of [=list/size=] 2 or greater was supplied
             by the client, or no such [=list=] was supplied. Otherwise, return only the below values.
+
+            Note: If the client supplies a [=list=] of exactly one credential and it was successfully employed, then its
+                [=credential ID=] is not returned since the client already knows it. This saves transmitting these bytes over
+                what may be a constrained connection in what is likely a common case.
+
         - |authenticatorData|
         - |signature|
-        - The [=user handle=] associated with |selectedCredential|, if no [=list=] of credentials was supplied by the client, or
-            no such [=list=] was supplied. Otherwise, do not return this value.
-
-      Note: If the client supplies a [=list=] of exactly one credential and it was successfully employed, then its
-          [=credential ID=] and [=user handle=] are not returned since the [=[RP]=], having provided the [=credential ID=] to the
-          client, already knows the [=credential ID=] and the identity of the user. This saves transmitting these bytes over what
-          may be a constrained connection in what is likely a common case, and enables backwards compatibility with legacy
-          [=authenticators=] that cannot store the [=user handle=].
+        - The [=user handle=] associated with |selectedCredential|, if available.
     </li>
 
 If the authenticator cannot find any credential corresponding to the specified [=[RP]=] that matches the specified criteria, it
--- a/index.bs
+++ b/index.bs
@@ -2225,7 +2225,9 @@ When this operation is invoked, the [=authenticator=] must perform the following
     1. Let |credentialId| be a new identifier for this credential that is globally unique with high probability across all
         credentials with the same type across all authenticators.
     1. Let |userHandle| be <code>|userEntity|.{{PublicKeyCredentialUserEntity/id}}</code>.
-    1. Associate the |credentialId| and |privateKey| with <code>|rpEntity|.{{PublicKeyCredentialRpEntity/id}}</code> and |userHandle|.
+    1. Associate the |credentialId| and |privateKey| with <code>|rpEntity|.{{PublicKeyCredentialRpEntity/id}}</code> and
+        |userHandle|. The authenticator MAY discard |userHandle| from this association if |privateKey| is not a
+        [=Client-side-resident Credential Private Key=].
     1. Delete any older credentials with the same <code>|rpEntity|.{{PublicKeyCredentialRpEntity/id}}</code> and |userHandle| that are stored locally by the [=authenticator=].
 1. If any error occurred while creating the new credential object, return an error code equivalent to "{{UnknownError}}" and
     terminate the operation.
--- a/index.bs
+++ b/index.bs
@@ -1245,8 +1245,8 @@ When this method is invoked, the user agent MUST execute the following algorithm
                 ::  whose value is the bytes of the signature value returned by the [=authenticator=].
 
                 :   <code><dfn for="assertionCreationData">userHandleResult</dfn></code>
-                ::  If the authenticator returned a [=user handle=], set the value of [=userHandleResult=] to be the bytes of that
-                    [=user handle=] returned by the [=authenticator=]. Otherwise, set the value of [=userHandleResult=] to null.
+                ::  If the [=authenticator=] returned a [=user handle=], set the value of [=userHandleResult=] to be the bytes of
+                    the returned [=user handle=]. Otherwise, set the value of [=userHandleResult=] to null.
 
                 :   <code><dfn for="assertionCreationData">clientExtensionResults</dfn></code>
                 ::  whose value is an {{AuthenticationExtensions}} object containing [=extension identifier=] →
--- a/index.bs
+++ b/index.bs
@@ -1245,7 +1245,8 @@ When this method is invoked, the user agent MUST execute the following algorithm
                 ::  whose value is the bytes of the signature value returned by the [=authenticator=].
 
                 :   <code><dfn for="assertionCreationData">userHandleResult</dfn></code>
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -392,7 +392,7 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
     * A [=Credential ID=].
     * A [=credential private key=].
     * The [=Relying Party Identifier=] for the [=[RP]=] that created this credential source.
-    * An optional [=user handle=] for the person who created this credential source.
+    * A [=user handle=] for the person who created this credential source.
```
emlun: This change is already merged into master via #721, but GitHub doesn't seem to detect it because I based this PR off the `issue-720-user-handle-optional` branch whose commits were rebased onto master instead of merging.

emlun:
This resolves #720.

emlun:
Assigned this to the CR milestone since the associated issue is assigned to CR.

emlun:
@akshayku pointed out on the 2017-12-20 WG call that the issues this was meant to solve are probably non-issues. We agreed but decided to ask @christiaanbrand to take one last look at this.

index.bs
```diff
@@ -2336,14 +2340,16 @@ When this method is invoked, the [=authenticator=] must perform the following pr
...
-
-            Note: If the client supplies a [=list=] of exactly one credential and it was successfully employed, then its
-                [=credential ID=] is not returned since the client already knows it. This saves transmitting these bytes over
-                what may be a constrained connection in what is likely a common case.
-
         - |authenticatorData|
         - |signature|
-        - The [=user handle=] associated with |selectedCredential|.
+        - The [=user handle=] associated with |selectedCredential|, if no [=list=] of credentials was supplied by the client, or
+            no such [=list=] was supplied. Otherwise, do not return this value.
```
akshayku: I don't see any reason on why not. 

Its not a private information and RP may need it even if RP is supplying a set of credentialIDs as userHandle is helping RP for the lookup even in second factor case.

index.bs
```diff
@@ -2336,14 +2340,16 @@ When this method is invoked, the [=authenticator=] must perform the following pr
...
+
+      Note: If the client supplies a [=list=] of exactly one credential and it was successfully employed, then its
+          [=credential ID=] and [=user handle=] are not returned since the [=[RP]=], having provided the [=credential ID=] to the
+          client, already knows the [=credential ID=] and the identity of the user. This saves transmitting these bytes over what
```
akshayku: That's not true. 

CredentialID and userHandle are providing different information and RP knowing CredentialID to information that is stored in userHandle is not accurate.

emlun:
I now agree that the user handle is not private information, but I don't see how the RP could not know the user's identity if the RP has already looked up a list of credential IDs for that user.

Either way, I see now that [CTAP's getAssertion method](https://fidoalliance.org/specs/fido-v2.0-rd-20170927/fido-client-to-authenticator-protocol-v2.0-rd-20170927.html#authenticatorGetAssertion) always returns the `user.id` even in the 2nd factor case, so I support closing this. Sorry for wasting everyone's time on yesterday's call.

akshayku:
Let me give you our example.

We have a distributed server environment and is doing credentialID based lookup. We use user id as a helper for lookup in this distributed environment for scalability purposes. The response server is not the same as request server. And that is why, userId is a helpful to us even in a second factor case. 

As this is not optional in CTAP spec, I see no reason for WebAuthN to hide this information. 

I am closing this. Please reopen if someone does not agree.

akshayku:
I am reopening this PR. 

The confusion is coming because 2nd factor != server credentials in all the cases. 

CTAP spec needs clarification as there is no user information returned in authenticatorGetAssertion for server credentials and U2F devices as it is not available to the authenticator. For device resident keys, userID MUST be returned (irrespective of whether credentialID list is provided or not). I will open the clarification PR for CTAP spec. 

Regarding this PR, it needs more work. Irrespective of allow credential ID list is present or not, if authenticator is giving userID back, it should be returned back to the RP. 

So the sections in this PR which talks about whether user ID returned from authenticator is null or not is correct. Sections which talks about removing this information when credential ID list is present in not correct.

index.bs
```diff
@@ -1245,7 +1245,8 @@ When this method is invoked, the user agent MUST execute the following algorithm
                 ::  whose value is the bytes of the signature value returned by the [=authenticator=].
 
                 :   <code><dfn for="assertionCreationData">userHandleResult</dfn></code>
-                ::  whose value is the bytes of the [=user handle=] returned by the [=authenticator=].
+                ::  If |savedCredentialId| exists, set the value of [=userHandleResult=] to be the bytes of the [=user handle=]
```
akshayku: Even if savedCredentialID exists or a list of credentialID exists, if authenticator is returning userID, it should be returned back to the RP.

index.bs
```diff
@@ -1278,7 +1279,9 @@ When this method is invoked, the user agent MUST execute the following algorithm
                             <code>|assertionCreationData|.[=assertionCreationData/signatureResult=]</code>.
 
                         :   {{AuthenticatorAssertionResponse/userHandle}}
-                        ::  A new {{ArrayBuffer}}, created using |global|'s [=%ArrayBuffer%=], containing the bytes of
+                        ::  If <code>|assertionCreationData|.[=assertionCreationData/userHandleResult=]</code> is null, set this
```
akshayku: This part is correct.

index.bs
```diff
@@ -2336,14 +2340,16 @@ When this method is invoked, the [=authenticator=] must perform the following pr
...
-                [=credential ID=] is not returned since the client already knows it. This saves transmitting these bytes over
-                what may be a constrained connection in what is likely a common case.
-
         - |authenticatorData|
         - |signature|
-        - The [=user handle=] associated with |selectedCredential|.
+        - The [=user handle=] associated with |selectedCredential|, if no [=list=] of credentials was supplied by the client, or
+            no such [=list=] was supplied. Otherwise, do not return this value.
+
+      Note: If the client supplies a [=list=] of exactly one credential and it was successfully employed, then its
```
akshayku: I would suggest this note section to be deleted. If user handle is returned from the authenticator, it should be returned back to the RP. 

Regarding only one CredentialID being sent, not returning it back does not work for distributed server scenarios like ours where request processing server is not the same as response processing server.

emlun:
What do you mean by "server credentials"?

Ah yes, I see now that CTAP's authenticatorMakeCredential method stores the user data only for resident keys. However, even with that limitation I think your use case can be solved by embedding a user ID in the `challenge` parameter. Since `challenge` is an opaque byte array of unspecified length, and is returned in the assertion response, you could set `challenge` to the UTF-8 bytes of for example `{"userId":"ff642b","random":"1/BjTenZNLAw9l03J2J2BcpgXP5Ic7gEuoVfVVrf7Bg="}`. This should allow you to embed the user ID while maintaining `challenge` as a valid nonce, correct?

So I don't think changing CTAP is strictly necessary, but I'll update this PR either way just in case we need it.

emlun:
Wait - actually, the response processing server (ResPS) needs to be able to verify that the returned `challenge` equals that sent to the client, so there needs to be some trusted communication path between the request processing server (ReqPS) and the ResPS so the ResPS can obtain the `challenge` from the ReqPS. Could that message then not also contain the user ID (if known, i.e. in 2nd factor mode)?

index.bs
emlun: Agreed.

index.bs
emlun: Agreed.

emlun:
Summary of currently proposed changes:

- `AuthenticatorAssertionResponse.userHandle` is now nullable.
- The authenticator MAY now skip storing the user handle for credentials that do not have a client-side-resident credential private key. (CTAP [currently does this][ctap])
- The authenticator now always returns the user handle if it is available.
- The client now returns `userHandle: null` if the authenticator did not return the user handle.

However it looks like CTAP is internally inconsistent: the `user` argument to authenticatorMakeCredential is stored only for resident keys, but the `user` attribute is required in the authenticatorGetAssertion response...

[ctap]: https://fidoalliance.org/specs/fido-v2.0-rd-20170927/fido-client-to-authenticator-protocol-v2.0-rd-20170927.html#h3_authenticatorMakeCredential

akshayku:
@emlun: Thanks for taking this on. 

You are correct, even if response server is different request server (rare case), user information is still available in addition to credentialID as it is a second factor case and user has already entered user information in first step. I was incorrect earlier and when I relooked at the whole scenario, The confusion came because earlier we were equating second factor case with server credentials case which is not always correct. 

For server credentials, it is of the same semantics as of U2F meaning authenticators MAY drop all user information while returning the credentialID. Some authenticators MAY also store this information in encrypted form in future however there is no need and I am not aware of any such authenticators as of now. In any case, this is authenticator specific. CTAP spec needs a clarification as it says user ID is required in authenticatorGetAssertion response which is inconsistent for U2F devices or for server credentials where there is no such information available. I will make that clarification.

Platform logic is simple, if it returned from authenticator (for resident keys, irrespective of whether credential list if present or not), return it to the RP, otherwise set it to NULL. 

Thanks

emlun:
@akshayku Ok, thanks!

equalsJeffH:
@akshayku please clarify what you mean by "server credentials"? thx.

akshayku:
@equalsJeffH : By "server credentials", I meant credentials which are resident on the server instead of FIDO device. Or in other words, credentials which are created via authenticatorMakeCredential API where "rk" is not set to TRUE and credentialID returned by the API contains encrypted private credential. 

These credentials created by FIDO 2 device has the same characteristics as credentials which would have been created via U2F devices.

ve7jtb:
all

the concern seems more around how resident credentials are protected.  It shouldn't matter if the resident credential is being used as first or second factor.   Both need to be protected.  
I think any sort of resident credential needs to be protected with at least a device pin.

I am not sure that a resident credential with no userID is sufficient protection, without some device unlock.  I think from a CTAP point of view a resident credential is resident credential.

John B.

index.bs
```diff
@@ -1245,7 +1245,8 @@ When this method is invoked, the user agent MUST execute the following algorithm
                 ::  whose value is the bytes of the signature value returned by the [=authenticator=].
 
                 :   <code><dfn for="assertionCreationData">userHandleResult</dfn></code>
-                ::  whose value is the bytes of the [=user handle=] returned by the [=authenticator=].
+                ::  If the authenticator returned a [=user handle=], set the value of [=userHandleResult=] to be the bytes of that
+                    [=user handle=] returned by the [=authenticator=]. Otherwise, set the value of [=userHandleResult=] to null.
```
If the [=authenticator=] returned a [=user handle=], set the value of [=userHandleResult=] to be the bytes of the returned [=user handle=]. Otherwise, ...

index.bs
```diff
@@ -1278,7 +1279,9 @@ When this method is invoked, the user agent MUST execute the following algorithm
                             <code>|assertionCreationData|.[=assertionCreationData/signatureResult=]</code>.
 
                         :   {{AuthenticatorAssertionResponse/userHandle}}
-                        ::  A new {{ArrayBuffer}}, created using |global|'s [=%ArrayBuffer%=], containing the bytes of
+                        ::  If <code>|assertionCreationData|.[=assertionCreationData/userHandleResult=]</code> is null, set this
+                            field to null. Otherwise, set this field to a new {{ArrayBuffer}}, created using |global|'s
```
equalsJeffH: nit: I'm not sure these two sentences are technically correct given that the dfn of [ArrayBuffer](https://heycam.github.io/webidl/#idl-ArrayBuffer) is: "An _object_ that holds a pointer (which may be null) to a buffer of a fixed number of bytes". I.e., may need to only instantiate the arraybuffer and assign the bytes of userHandleResult to it, and if userHandleResult is null, then the buffer pointer is automagically set to null.

However, the present language may be close enough to pass muster with reviewers.

index.bs

index.bs
emlun: As far as I could tell when I looked into it, it seemed like `ArrayBuffer` is a nullable type, if that's what you mean. Did I get that wrong?

emlun: Or, rather, it looked to me making a field of type `ArrayBuffer` nullable is legal. In that case I suppose the value could be null in two ways, so you may be right in that it's more correct to keep it not nullable, but I feel like making the field itself explicitly nullable better communicates the intent. What do you think?

PR #731, opened 2017-12-20
fix #455: we are using CTAP canonical CBOR encoding form everywhere

### LINKED ISSUE #455: Consider requiring canonical CBOR throughout
[Canonical CBOR](https://tools.ietf.org/html/rfc7049#section-3.9) is a distinguished subset of CBOR encodings, and is specified in the original RFC. Most CBOR implementations will produce something close to canonical CBOR most of the time but, by requiring it, implementations can avoid decoder complexity. This might be especially valuable for authenticators, as they are constrained and need to parse CBOR if they are to handle extensions.

Without this, implementations are very likely to implement an ill-defined subset of CBOR anyway (e.g. by rejecting lengths with superfluous zeros and so on) and this will result in arguments whenever something produces valid (but non-canonical) CBOR that is rejected by some existing implementations. If this discovery happens late in the development of an implementation, it can be quite costly.

Conversely, if canonical CBOR is required, a single common implementation that is strict will ensure that the whole ecosystem remains healthy and interoperable on this point.

Lastly, the ordering requirements of canonical CBOR make it easy to reject maps with duplicate keys. Although that doesn't appear to be pertinent to the current webauthn spec, it's a classic way to cause confusion between two parsers (and thus security issues) and could arise in future extensions.

equalsJeffH:
see issue #455 consider requiring canonical CBOR thoughout.

see also: https://github.com/fido-alliance/fido-2-specs/pull/401

fix #455

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/pull/731.html" title="Last updated on Dec 22, 2017, 12:57 AM GMT (c7149f7)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/731/986d627...c7149f7.html" title="Last updated on Dec 22, 2017, 12:57 AM GMT (c7149f7)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -250,8 +250,10 @@ A [=[RP]=] MUST behave as described in [[#rp-operations]] to obtain the security
 
 ## All Conformance Classes ## {#conforming-all-classes}
 
-All [=CBOR=] encoding and decoding performed by the members of the above conformance classes MUST be done using the
-[=CTAP canonical CBOR encoding form=].
+All [=CBOR=] encoding performed by the members of the above conformance classes MUST be done using the
+[=CTAP canonical CBOR encoding form=]. 
+All decoders of the above conformance classes SHOULD reject CBOR that is not validly encoded
+in the [=CTAP canonical CBOR encoding form=] and SHOULD reject messages with duplicate map keys.
 
 
 # Dependencies # {#dependencies}
--- a/index.bs
+++ b/index.bs
@@ -106,6 +106,11 @@ spec: page-visibility; urlPrefix: https://www.w3.org/TR/page-visibility/
 spec: WHATWG HTML; urlPrefix: https://html.spec.whatwg.org/
     type: dfn 
         text: focus
+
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -241,7 +246,13 @@ external hardware, or a combination of both.
 
 ## [RPS] ## {#conforming-relying-parties}
 
-A [=[RP]=] MUST behave as described in [[#rp-operations]] to get the security benefits offered by this specification.
+A [=[RP]=] MUST behave as described in [[#rp-operations]] to obtain the security benefits offered by this specification.
+
+## All Conformance Classes ## {#conforming-all-classes}
+
+All [=CBOR=] encoding and decoding performed by the members of the above conformance classes MUST be done using the
```
akshayku: Encoding is MUST. Decoding is SHOULD as per CTAP spec.

equalsJeffH: thx, fixed to match language in fido-alliance/fido-2-specs#401

PR #732, opened 2017-12-22
fix 543: improve COSE_Key spec language and add COSE_Key examples

### LINKED ISSUE #543: musings wrt webauthn's profile of COSE_Key
presently, in [WD-06 section 5.3.1. Attestation data](https://www.w3.org/TR/2017/WD-webauthn-20170811/#sec-attestation-data), we say (in the last row of the table) in regards to the [credential public key](https://www.w3.org/TR/2017/WD-webauthn-20170811/#credential-public-key) (aka "user public key" see issue #540):
> The credential public key encoded in COSE_Key format, as defined in Section 7 of [RFC8152]. The encoded credential public key MUST contain the "alg" parameter and MUST NOT contain any other optional parameters. The "alg" parameter MUST contain a COSEAlgorithmIdentifier value. 

Given the stipulations in [Section 7 of [RFC8152]](https://tools.ietf.org/html/rfc8152#section-7) -- note that the "kty" param is required, and "alg" is one of the optional params, plus the MUST stipulations in [section 13.1.1](https://tools.ietf.org/html/rfc8152#section-13.1.1) for "x", "y", and "crv". This means we'd actually have instantiated COSE_Key structures, for EC public keys, like so:
```
 {
    1: <kty>     ; required by S 7 RFC8152, example value '2', ie "EC2"
    3: <alg>     ; optional per RFC8152, required per webauthn, exp. value '-7', ie "ES256"
   -2: <x coord> ; required by S 13.1.1 RFC8152
   -3: <y coord> ; required by S 13.1.1 RFC8152
   -1: <crv>     ; required by S 13.1.1 RFC8152, exp. value '1', ie "P-256"
 }
```

Yes?

equalsJeffH:
this builds on PR #731 to improve COSE_Key spec language and add COSE_Key examples

fix #543

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/pull/732.html" title="Last updated on Jan 11, 2018, 11:44 PM GMT (0bac86e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/732/ade8321...0bac86e.html" title="Last updated on Jan 11, 2018, 11:44 PM GMT (0bac86e)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2561,11 +2561,11 @@ to be used with the PS256 signature algorithm
 
 <pre class="example" highlight="json">
   {
-     1:  3,  ; kty: RSA key type
-     3:-37,  ; alg: PS256
-    -1:  n,  ; n:   RSA modulus n byte string 256 bytes in length
+    1:   3,  ; kty: RSA key type
+    3: -37,  ; alg: PS256
+   -1:   n,  ; n:   RSA modulus n byte string 256 bytes in length
                     e.g., in hex (middle bytes elided for brevity): DB5F651550...6DC6548ACC3
-    -2:  e   ; e:   RSA public exponent e byte string 3 bytes in length
+   -2:   e   ; e:   RSA public exponent e byte string 3 bytes in length
                     e.g., in hex: 010001
   }
 </pre>
@@ -2575,11 +2575,11 @@ to be used with the RS256 signature algorithm (RSASSA-PKCS1-v1_5 with SHA-256, s
 
 <pre class="example" highlight="json">
   {
-     1:   3,  ; kty: RSA key type
-     3:-257,  ; alg: RS256
-    -1:  n,  ; n:   RSA modulus n byte string 256 bytes in length
+    1:   3,  ; kty: RSA key type
+    3:-257,  ; alg: RS256
+   -1:   n,  ; n:   RSA modulus n byte string 256 bytes in length
                     e.g., in hex (middle bytes elided for brevity): DB5F651550...6DC6548ACC3
-    -2:  e   ; e:   RSA public exponent e byte string 3 bytes in length
+   -2:   e   ; e:   RSA public exponent e byte string 3 bytes in length
                     e.g., in hex: 010001
   }
 </pre>
--- a/index.bs
+++ b/index.bs
@@ -2555,9 +2555,9 @@ are included here for clarity and to match the [[CDDL]] presentation above:
 </pre>
 
 
-Below is an example of a COSE_Key-encoded 2048-bit RSA public key (see [[RFC8230]] [Section 4](https://tools.ietf.org/html/rfc8230#section-4))
+Below is an example of a COSE_Key-encoded 2048-bit RSA public key (see [[RFC8230]] [Section 4](https://tools.ietf.org/html/rfc8230#section-4)),
 to be used with the PS256 signature algorithm
-(RSASSA-PSS w/ SHA-256, see [[RFC8230]] [Section 2](https://tools.ietf.org/html/rfc8230#section-2)):
+(RSASSA-PSS with SHA-256, see [[RFC8230]] [Section 2](https://tools.ietf.org/html/rfc8230#section-2)):
 
 <pre class="example" highlight="json">
   {
@@ -2570,9 +2570,8 @@ to be used with the PS256 signature algorithm
   }
 </pre>
 
-Below is an example of a COSE_Key-encoded 2048-bit RSA public key (see [[RFC8230]]
-[Section 4](https://tools.ietf.org/html/rfc8230#section-4))
-to be used with the RS256 signature algorithm (RSASSA-PKCS1-v1_5 w/ SHA-256, see [[#sctn-cose-alg-reg]]):
+Below is an example of the same COSE_Key-encoded RSA public key as above, 
+to be used with the RS256 signature algorithm (RSASSA-PKCS1-v1_5 with SHA-256, see [[#sctn-cose-alg-reg]]):
 
 <pre class="example" highlight="json">
   {
--- a/index.bs
+++ b/index.bs
@@ -2443,7 +2443,7 @@ object=] for a given credential. It has the following format:
     <tr>
         <td><dfn>credentialIdLength</dfn></td>
         <td>2</td>
-        <td>Byte length L of Credential ID</td>
+        <td>Byte length <strong>L</strong> of Credential ID</td>
     </tr>
     <tr>
         <td><dfn>credentialId</dfn></td>
@@ -2454,7 +2454,8 @@ object=] for a given credential. It has the following format:
         <td><dfn>credentialPublicKey</dfn></td>
         <td>variable</td>
         <td>
-            The [=credential public key=] encoded in COSE_Key format, as defined in Section 7 of [[RFC8152]].
+            The [=credential public key=] encoded in COSE_Key format,
+            as defined in Section 7 of [[RFC8152]], using the [=CTAP canonical CBOR encoding form=].
             The COSE_Key-encoded [=credential public key=] MUST contain the optional "alg" parameter and MUST NOT
             contain any other optional parameters. The "alg" parameter MUST contain a {{COSEAlgorithmIdentifier}} value.
             The encoded [=credential public key=] MUST also contain any additional required parameters stipulated by the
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -2454,16 +2454,93 @@ object=] for a given credential. It has the following format:
...
+<pre class="example" highlight="json">
+  {
+     1:   3,  ; kty: RSA key type
+     3:-257,  ; alg: RS256
+    -1:  n,  ; n:   RSA modulus n byte string 257 bytes in length
+                    e.g., in hex (middle bytes elided for brevity): DB5F651550...6DC6548ACC3
+    -2:  e   ; e:   RSA public exponent e byte string 3 bytes in length
+                    e.g., in hex: 010001
+  }
+</pre>
```
akshayku: One example:

{1: 3, 3: -257, -1: h'ea9fbf0b9766e8e7412f86adeab7af8e11dd01308c1b71ac7c037f98a796b98f54253cf8b83f72b656340895608752af185e560695da06a77a655760ab28df3d44bfa417fe28978bfe34f1c93ad44508f4e6c56a4840f09aaf89481faa22098b41bb49082faad0ba94ffb529cd95cec0a56f3eebee9ff0f9cd95b22c48ef4854ce0f719cb06e76ba746e4129fbcfcb3f80b44bb2edd3908f468991a5249a1b8d9234316f245b1289f95d778a461733543612e3705c5ec2b78a9674305ca9fa464c3e95b0b67f058ff0cc6a84b0890551c413f1557a2a198c6d2629ade0f16ef9aec06ae89162f15a98ae49ce8b0ce8e07bafe1bd3b728bd1fc18f07e87a425f1', -2: h'010001'}

equalsJeffH:
the build errors just now were both "disk quota exceeded":
`[0Koci runtime error: exec failed: container_linux.go:265: starting container process caused "could not create session key: disk quota exceeded"`
[I built the spec successfully locally using cloud bikeshed]

@jcjones -- help?

index.bs
```diff
@@ -2467,6 +2468,79 @@ object=] for a given credential. It has the following format:
...
+
+Below is an example of a COSE_Key-encoded 2048-bit RSA public key (see [[RFC8230]] [Section 4](https://tools.ietf.org/html/rfc8230#section-4))
+to be used with the RSASSA-PSS w/ SHA-256 signature algorithm 
+(see [[RFC8230]] [Section 2](https://tools.ietf.org/html/rfc8230#section-2)):
+
+<pre class="example" highlight="json">
+  {
+     1:  3,  ; kty: RSA key type
+     3:-37,  ; alg: RS256
+    -1:  n,  ; n:   RSA modulus n byte string 257 bytes in length
```
agl: Should this value (and below) be a 25<em>6</em> byte value? (There's no zero padding to avoid negative numbers in COSE.)

index.bs
```diff
@@ -2467,6 +2468,79 @@ object=] for a given credential. It has the following format:
...
+
+
+Below is an example of a COSE_Key-encoded 2048-bit RSA public key (see [[RFC8230]] [Section 4](https://tools.ietf.org/html/rfc8230#section-4))
+to be used with the RSASSA-PSS w/ SHA-256 signature algorithm 
+(see [[RFC8230]] [Section 2](https://tools.ietf.org/html/rfc8230#section-2)):
+
+<pre class="example" highlight="json">
+  {
+     1:  3,  ; kty: RSA key type
+     3:-37,  ; alg: RS256
```
agl: nit: PS256, rather than <em>R</em>S256, I believe.

index.bs
```diff
@@ -2467,6 +2468,79 @@ object=] for a given credential. It has the following format:
...
+</pre>
+
+Below is an example of a COSE_Key-encoded 2048-bit RSA public key (see [[RFC8230]]
+[Section 4](https://tools.ietf.org/html/rfc8230#section-4))
+to be used with the RSASSA-PKCS1-v1_5 w/ SHA-256 signature algorithm (see [[#sctn-cose-alg-reg]]):
+
+<pre class="example" highlight="json">
+  {
+     1:   3,  ; kty: RSA key type
+     3:-257,  ; alg: RS256
```
agl: Is algorithm -257 defined for COSE? I don't see it in the [IANA registry](https://www.iana.org/assignments/cose/cose.xhtml#algorithms). (In fact, I don't see PKCS <span>#</span>1 v1.5 at all.)

agl:
(I see that I commented on the wrong subchange, but I think my point about s/257/256/ is still valid even if GitHub thinks it's outdated.)

index.bs
jcjones: It's registered in WebAuthn, actually: https://w3c.github.io/webauthn/#sctn-cose-alg-reg

index.bs
equalsJeffH: yes, fixed, thx.

index.bs
equalsJeffH: thx hm. I dunno, you guys tell me. Here's how I generated the example RSA key pair, and openssl sez the modulus is 257 bytes:
```
 bash-3.2$ openssl genrsa -des3 -out private.pem 2048
Generating RSA private key, 2048 bit long modulus
.......................+++
..........................................................................................................+++
e is 65537 (0x10001)
Enter pass phrase for private.pem:
Verifying - Enter pass phrase for private.pem:
bash-3.2$ 
bash-3.2$ 
bash-3.2$ openssl rsa -in private.pem -outform PEM -pubout -out public.pem
Enter pass phrase for private.pem:
writing RSA key
bash-3.2$ 
bash-3.2$ 
bash-3.2$ 
bash-3.2$ PUBKEY=`grep -v -- ----- public.pem | tr -d '\n'`
bash-3.2$ 
bash-3.2$ 
bash-3.2$ echo $PUBKEY | base64 -d | openssl asn1parse -inform DER -i
base64: invalid option -- d
Usage:	base64 [-hvD] [-b num] [-i in_file] [-o out_file]
  -h, --help     display this message
  -D, --decode   decodes input
  -b, --break    break encoded string into num character lines
  -i, --input    input file (default: "-" for stdin)
  -o, --output   output file (default: "-" for stdout)
Error: offset too large
bash-3.2$ 
bash-3.2$ 
bash-3.2$ echo $PUBKEY | base64 -D | openssl asn1parse -inform DER -i
    0:d=0  hl=4 l= 290 cons: SEQUENCE          
    4:d=1  hl=2 l=  13 cons:  SEQUENCE          
    6:d=2  hl=2 l=   9 prim:   OBJECT            :rsaEncryption
   17:d=2  hl=2 l=   0 prim:   NULL              
   19:d=1  hl=4 l= 271 prim:  BIT STRING        
bash-3.2$ 
bash-3.2$ 
bash-3.2$ 
bash-3.2$ echo $PUBKEY | base64 -D | openssl asn1parse -inform DER -i -strparse 19
    0:d=0  hl=4 l= 266 cons: SEQUENCE          
    4:d=1  hl=4 l= 257 prim:  INTEGER           :DB5F651550E3E99916CEE05C15D9322AD5AA6FF068FC63943A598F860635F413E84DC33CF72DCCE21FAE9FF9ED9DCA0EAD57DFF01921EF4DD1ADE1C22232D94BDA40DB66EF35855DC241D64496C70503FB4FE4CCAC05395DD9040E5C8168186FBFB61B3158BBBC8C7CD0F4309D8944C620522F5648AD0EDF544D318DD882BDE80CB1FE5F0C3924CD5F95616E66CA2B79D79C28B1192A01AA0A6D5614BF754D3085CEBE50F684AC2EDFD52B37BBB3740650EC4A736DD90F1455526AF3D8B3FA535795DB1A40175EF23E559DE7B0357CB050889C554BE149AF254DF147352276BD5F0453FAFCFEFD5AE019F69DFEB53AB2DD1063C4C4E4A6D8AD5B16DC6548ACC3
  265:d=1  hl=2 l=   3 prim:  INTEGER           :010001
```
is there any zero-padding in the above modulus value?  Ah, is the leading `:` padding (that COSE_Key encoding omits)?

equalsJeffH:
@agl 

ok, you are referring to https://github.com/w3c/webauthn/pull/732#discussion_r159558639, i believe. 

I think I understand, tho I remain unsure whether the correct modulus length is 256 or 257 bytes, please see: https://github.com/w3c/webauthn/pull/732#discussion_r160811814, thx.

index.bs
agl: 2048 bits / 8 = 256, so we would expect a 256-byte value.

ASN.1 DER requires that an INTEGER with the most-significant bit set be padded with a leading zero to avoid it appearing to be negative. Since the leading byte is 0xdb, that will have happened in this case. In fact, it'll happen for all RSA keys since the MSB is required to be 1 (otherwise it would be a 2047-bit key).

COSE, however, doesn't have this leading padding byte. That is made explicit [here](https://tools.ietf.org/html/rfc8230#section-4):

  >  The octet sequence MUST utilize the minimum number of octets needed to represent the value.

equalsJeffH: Ok, thx, so is openssl just not displaying a leading zero padding byte?

agl: Correct, as it's not considered part of the value. If you base64-decode the public key and look for the modulus in a hex dump, it should be preceded by 02820101<b>00</b>.

equalsJeffH:
Ready to merge?

index.bs
```diff
@@ -2528,7 +2528,7 @@ to be used with the RSASSA-PSS w/ SHA-256 signature algorithm
 
 Below is an example of a COSE_Key-encoded 2048-bit RSA public key (see [[RFC8230]]
 [Section 4](https://tools.ietf.org/html/rfc8230#section-4))
-to be used with the RSASSA-PKCS1-v1_5 w/ SHA-256 signature algorithm (see [[#sctn-cose-alg-reg]]):
+to be used with the RS256 signature algorithm (RSASSA-PKCS1-v1_5 w/ SHA-256, see [[#sctn-cose-alg-reg]]):
```
agl: (Maybe replace "w/" with "with". Your call.)

index.bs
```diff
@@ -2467,6 +2468,79 @@ object=] for a given credential. It has the following format:
...
+
+Below is an example of a COSE_Key-encoded 2048-bit RSA public key (see [[RFC8230]]
+[Section 4](https://tools.ietf.org/html/rfc8230#section-4))
+to be used with the RSASSA-PKCS1-v1_5 w/ SHA-256 signature algorithm (see [[#sctn-cose-alg-reg]]):
+
+<pre class="example" highlight="json">
+  {
+     1:   3,  ; kty: RSA key type
+     3:-257,  ; alg: RS256
+    -1:  n,  ; n:   RSA modulus n byte string 257 bytes in length
```
agl: I think one extra space is needed after the colon here and below. (And then the continued comment lines would need an extra space too.)

index.bs
```diff
@@ -2467,6 +2468,79 @@ object=] for a given credential. It has the following format:
...
+     1:  3,  ; kty: RSA key type
+     3:-37,  ; alg: RS256
+    -1:  n,  ; n:   RSA modulus n byte string 257 bytes in length
+                    e.g., in hex (middle bytes elided for brevity): DB5F651550...6DC6548ACC3
+    -2:  e   ; e:   RSA public exponent e byte string 3 bytes in length
+                    e.g., in hex: 010001
+  }
+</pre>
+
+Below is an example of a COSE_Key-encoded 2048-bit RSA public key (see [[RFC8230]]
```
agl: (Could also just say that it's the same key because it's also RSA. Your call.)

index.bs
selfissued: Once we reach CR, I plan to submit the registration request to IANA.

index.bs
equalsJeffH: done

index.bs
equalsJeffH: done

index.bs
equalsJeffH: done

PR #739, opened 2018-01-10
Changed uses of JSON string to USVString

### LINKED ISSUE #417: Correct uses of "JSON string" versus "DOMString" and other string terminology usage
@jyasskin pointed out in his comment https://github.com/w3c/webauthn/pull/389#pullrequestreview-33915025 that some uses of "JSON string" should be changed to "DOMString".  We should review uses of these terms, as well as "CBOR string", uses of "UTF-8", and other string encoding/decoding language to make it consistent and correct.

selfissued:
Fixes #417

***
<a href="https://pr-preview.s3.amazonaws.com/selfissued/webauthn/pull/739.html" title="Last updated on Jan 10, 2018, 4:55 PM GMT (e9808a7)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/739/0ed6257...selfissued:e9808a7.html" title="Last updated on Jan 10, 2018, 4:55 PM GMT (e9808a7)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -3558,7 +3558,7 @@ JavaScript APIs.
 :: `appid`
 
 : Client extension input
-:: A single JSON string specifying a FIDO |AppID|.
+:: A single USVString specifying a FIDO |AppID|.
 
 : Client extension processing
 ::  1. If present in a {{CredentialsContainer/create()}} call, return a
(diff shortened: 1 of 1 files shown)


rlin1:
Is it this USVString: http://docs.w3cub.com/dom/usvstring/

jcjones:
@bzbarsky - can you confirm for us that `USVString` is what we should be using to pass user-supplied textual strings around?

bzbarsky:
Does "user" mean "actual person sitting at a keyboard"?

If so, then I'd think USVString is a reasonable option.  @annevk may have better guidance somewhere, apart from whatever is at https://heycam.github.io/webidl/#idl-USVString

If you plan to pass the string to someone who expects "Unicode text", then USVString is a good idea.

annevk:
I think that's it. We also have https://infra.spec.whatwg.org/#strings but it doesn't mention IDL types (at some point IDL will be rewritten in terms of Infra at which point it'll reference all of that).

PR #760, opened 2018-01-24
fix #658: add user cancelled operation Note

### LINKED ISSUE #658: CTAP/U2F doesn't status indicating the user cancelled the operation
In the authenticator parallel processing section (https://github.com/w3c/webauthn/blob/master/index.bs#L738), it says: 

`        :   If any |authenticator| returns a status indicating that the user cancelled the operation,
        ::  1. [=set/Remove=] |authenticator| from |issuedRequests|.
            2. [=set/For each=] remaining |authenticator| in |issuedRequests| invoke the [=authenticatorCancel=] operation on
`

Looking at both CTAP and U2F specs, I haven't found an error matching this indication. Should we remove this from the algorithm?

equalsJeffH:
fix #658

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/pull/760.html" title="Last updated on Feb 6, 2018, 5:00 PM GMT (914399a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/760/1fc8906...914399a.html" title="Last updated on Feb 6, 2018, 5:00 PM GMT (914399a)">Diff</a>


emlun:
This might be slightly unrelated to this particular PR, but should these steps also make the client stop accepting new authenticators (but not terminate the timer) in step 19/17?

What I mean is, say an operation is initiated with two authenticators connected. The user cancels the operation on one of them, and the client then cancels the operation on the other authenticator as well. Say the user then plugs in a third authenticator, perhaps for a completely unrelated purpose. As currently specified (likely a relic from before hot plugging), the client would initiate the operation anew with this authenticator, which might surprise the user since they just canceled it. It could in some cases lead to the user inadvertently accepting the Webauthn operation instead of the unrelated task they plugged in the authenticator for.

What do you think?

equalsJeffH:
wrt https://github.com/w3c/webauthn/pull/760#issuecomment-360095734, WDYT @jcjones @akshayku @leshi ?

kpaulh:
The situation does sound plausible, and I agree with @emlun that this could be surprising to the user. Since the intent of the spec seems to be that canceling an operation on one authenticator cancels the entire operation across all authenticators, is there a reason we can't just terminate the timer?

emlun:
>is there a reason we can't just terminate the timer?

Yes, there is: See [getAssertion step 19][get] and [§14 Privacy Considerations 2. Authentication Ceremonies][pc].

[get]: https://w3c.github.io/webauthn/#discover-from-external-source
[pc]: https://w3c.github.io/webauthn/#sec-assertion-privacy

kpaulh:
Right-o, thanks for reminding me. In that case, denying new authenticators seems like a reasonable solution.

index.bs
```diff
@@ -908,6 +908,9 @@ When this method is invoked, the user agent MUST execute the following algorithm
             1. [=set/For each=] remaining |authenticator| in |issuedRequests| invoke the [=authenticatorCancel=] operation on
                 |authenticator| and [=set/remove=] it from |issuedRequests|.
 
+                Note: An [=authenticator=] may return some indication of "the user cancelled the entire operation".
+                Conveyance of such an indication is client platform-specific.
```
selfissued: I think you mean transport-specific - not platform-specific.

index.bs
```diff
@@ -1260,6 +1263,9 @@ When this method is invoked, the user agent MUST execute the following algorithm
             1. [=set/For each=] remaining |authenticator| in |issuedRequests| invoke the [=authenticatorCancel=] operation
                 on |authenticator| and [=set/remove=] it from |issuedRequests|.
 
+                Note: An [=authenticator=] may return some indication of "the user cancelled the entire operation".
+                Conveyance of such an indication is client platform-specific.
```
selfissued: Ditto

index.bs
jcjones: Hrmm; I think Jeff means it's up to the user agent?

index.bs
jcjones: Yeah... maybe it could just be "It is unspecified how a user agent should convey this state to the user" ?

index.bs
equalsJeffH: whether such a "user cancelled entire operation" indication is conveyed successfully to the RP client-side JS, is specific, as I understand it, to the entire userAgent+OSplatform+transport+authenticator stack. That's what "client platform-specific" is attempting to convey.  

fyi/fwiw, "client platform" (and synonyms thereof) are among our currently undefined terms.  So making this stuff more clear may involved getting around to defining at least that term, seems to me.

index.bs
equalsJeffH: yeah, how this state is annunciated is another aspect of this.

equalsJeffH:
@selfissued @jcjones -- pls re-review, thx for the wording suggesting @jcjones.

nadalin:
@selfissued Please re-review the changes

PR #765, opened 2018-01-26
Strongly type client extension inputs and outputs

### LINKED ISSUE #346: Client arguments should be specified with "partial dictionary AuthenticationExtensions"
https://w3c.github.io/webauthn/#dictdef-authenticationextensions says
> This is a dictionary containing zero or more extensions as defined in [§8 WebAuthn Extensions](https://w3c.github.io/webauthn/#extensions).

However, section 8 doesn't define any extensions to this dictionary. I suspect that the [client argument](https://w3c.github.io/webauthn/#client-argument) for each extension should actually be the type of a new dictionary member on `AuthenticationExtensions`. e.g.:

> 9.1. FIDO AppId
> ...
> ```idl
> partial dictionary AuthenticationExtensions {
>   BufferSource fido_appid;
> };
> ```

(In this case the client processing would check that the buffer contained valid UTF-8.)

Alternately, you could specify AuthenticationExtensions as a

```idl
typedef record<DOMString, any> AuthenticationExtensions
```

And have the client processing do any type checking manually. This would work better if any extension names need to include `.`s.

### LINKED ISSUE #626: Extensions need to define how their parameters convert to/from CBOR
Extension definitions need to be more thorough, as @bzbarsky [points out in the discussion for issue #624](https://github.com/w3c/webauthn/issues/624#issuecomment-335853416). In particular, for each extension -- including the example -- we need two sections:

1. How to encode the `AuthenticationExtensions` data structure  into CBOR for the wire protocol
2. How to decode the resulting extension CBOR into the `AuthenticationExtensions` data structure

Without these sections, it's under-specified how to actually implement these extensions.

### LINKED ISSUE #713: JSON deserialization is not defined
The prose would read much clearer if this was an operation that returned something.

### LINKED ISSUE #738: UAs can't pass unknown extensions to authenticators
https://w3c.github.io/webauthn/#extensions claims that 
> Clients wishing to support the widest possible range of extensions MAY choose to pass through any extensions that they do not recognize to authenticators, generating the authenticator extension input by simply encoding the client extension input in CBOR.

However,
1. "simply encoding the client extension input in CBOR" is underspecified: it could refer to https://tools.ietf.org/html/rfc7049#section-4.2, but like with canonical CBOR, the RFC's suggestion is probably not quite right for our use. For example, should the value `1.9` be sent as-is, rejected, or converted to an integer first? If it's sent as-is, as suggested by the RFC, we'll need to extend CTAP's definition of canonical CBOR, which doesn't support floating point values.
2. As long as extensions define their own conversions as requested by #626, these conversions are likely to be different from the generic conversion, forcing authenticators to deal with two different input formats, which they're likely to do wrong.

To keep things simple, I suggest requiring that UAs drop unknown extensions instead of forwarding them. If someone disagrees, I'd like them to write a PR defining the generic conversion from JS objects to CBOR that UAs are supposed to use.

selfissued:
Client extension inputs and outputs are now strongly typed.  Several individual extensions were also cleaned up so that their JavaScript and CBOR representations directly correspond to one another.  The example extension was so irregular and ad-hoc as to actually be a terrible example.  It was therefore deleted.  The existing extensions now should serve as good examples.

Fixes #346 
Fixes #626 
Partially fixes #713 
Partially fixes #738

***
<a href="https://pr-preview.s3.amazonaws.com/selfissued/webauthn/pull/765.html" title="Last updated on Jan 31, 2018, 11:51 AM GMT (96fe109)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/765/1e77b42...selfissued:96fe109.html" title="Last updated on Jan 31, 2018, 11:51 AM GMT (96fe109)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1869,15 +1869,6 @@ This is a dictionary containing the [=client extension output=] values for zero 
 This is a dictionary containing the [=authenticator extension input=] values for zero or more WebAuthn extensions, as defined in [[#extensions]].
 
 
-## Authentication Extensions Authenticator Outputs (typedef <dfn>AuthenticationExtensionsAuthenticatorOutputs</dfn>) ## {#iface-authentication-extensions-authenticator-outputs}
-
-<xmp class="idl">
-    typedef record<DOMString, DOMString> AuthenticationExtensionsAuthenticatorOutputs;
-</xmp>
-
-This is a dictionary containing the [=authenticator extension output=] values for zero or more WebAuthn extensions, as defined in [[#extensions]].
-
-
 ## Supporting Data Structures ## {#supporting-data-structures}
 
 The [=public key credential=] type uses certain data structures that are specified in supporting specifications. These are as
--- a/index.bs
+++ b/index.bs
@@ -598,7 +598,7 @@ that are returned to the caller when a new credential is created, or a new asser
     interface PublicKeyCredential : Credential {
         [SameObject] readonly attribute ArrayBuffer              rawId;
         [SameObject] readonly attribute AuthenticatorResponse    response;
-        AuthenticationExtensions getClientExtensionResults();
+        AuthenticationExtensionsClientOutputs getClientExtensionResults();
     };
 </xmp>
 <dl dfn-type="attribute" dfn-for="PublicKeyCredential">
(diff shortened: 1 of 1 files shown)


agl:
The type `AuthenticationExtensionsAuthenticatorOutputs` seems to be unused. Otherwise LGTM.

selfissued:
The `AuthenticatorExtensionsAuthenticationOutputs` type is there to be parallel to the other three types.  But you're right looking at it @agl  - it's not used.  Rather, these outputs appear in the extensions field of https://w3c.github.io/webauthn/#authenticator-data.  I'll remove it.

selfissued:
@agl can you please approve this PR now that the unused typedef has been removed?  Thanks.

index.bs
```diff
@@ -3879,13 +3878,26 @@ This [=registration extension=] and [=authentication extension=] enables use of
...
 
 : Client extension processing
 :: None, except creating the authenticator extension input from the client extension input.
 
 : Client extension output
-:: Returns a JSON array of 3-element arrays of numbers that encodes the factors in the authenticator extension output
+:: Returns a JSON array of 3-element arrays of numbers that encodes the factors in the authenticator extension output.
+    <xmp class="idl">
+    typedef sequence<unsigned long> uvmEntry;
+    typedef sequence<uvmEntry> uvmEntries;
```
emlun: Nit: Should these typedefs' names also have initial uppercase like the others?

selfissued: Done.  Thanks for the careful read @emlun .

kpaulh:
I concur with agl, this looks good to me as well.

PR #766, opened 2018-01-26
Security Considerations for Unsigned Credential ID

### LINKED ISSUE #394: "credential ID" not signed over by authenticatorGetAssertion operation
I note that in [authenticatorGetAssertion operation](https://w3c.github.io/webauthn/#op-get-assertion), the "credential ID" not signed over -- i.e., it is not included in `authenticator data` because no `attObj` is contained in the `authenticator data` returned by this operation.  As I (vaguely) recall, we discussed this long ago and determined that the worst downside of a buggy authenticator returning an incorrect credential ID is that the RP will not look up the correct cred public key with which to verify the returned signed `authenticator data` (aka assertion) and the overall ceremony would thus end in error. 

we should probably document this rationale and consequence somewhere in the spec.

selfissued:
Fixes #394

***
<a href="https://pr-preview.s3.amazonaws.com/selfissued/webauthn/pull/766.html" title="Last updated on Jan 26, 2018, 4:21 PM GMT (86f8f7c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/766/1e77b42...selfissued:86f8f7c.html" title="Last updated on Jan 26, 2018, 4:21 PM GMT (86f8f7c)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -3654,13 +3654,13 @@ as well. This allows authenticators without a font rendering engine to be used a
 :: `txAuthGeneric`
 
 : Client extension input
-:: A CBOR map defined as follows:
-    ```
-    txAuthGenericArg = {
-                           contentType: text,   ; MIME-Type of the content, e.g. "image/png"
-                           content: bytes
-                       }
-    ```
+:: A JavaScript object defined as follows:
+    <pre class="idl">
+    dictionary txAuthGenericArg {
+        required USVString contentType;    // MIME-Type of the content, e.g., "image/png"
+        required ArrayBuffer content;
+    };
+    </pre>
 
 : Client extension processing
 :: None, except creating the authenticator extension input from the client extension input.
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -4427,7 +4427,13 @@ RogueList when performing ECDAA-Verify (see section 3.6 in [[!FIDOEcdaaAlgorithm
 [[FIDOMetadataService]] provides one way to access such information.
 
 
+## credentialId Unsigned ## {#credentialIdSecurity}
 
+The credential ID is not signed.
+This is not a problem because all that would happen if an authticator returns
```
emlun: Typo: authticator

index.bs
```diff
@@ -4427,7 +4427,13 @@ RogueList when performing ECDAA-Verify (see section 3.6 in [[!FIDOEcdaaAlgorithm
...
+the wrong crednetial ID is that the RP will not look up the correct
```
emlun: Typo: crednetial
Suggest: RP -> [=[RP] =]

emlun: I think a more interesting case to point out is if a MitM were to manipulate the credential ID (the conclusion is the same), so I would probably write this as something like "[...] if an authenticator returns the wrong credential ID, or if an attacker intercepts and manipulates the credential ID, is that the RP [...]"

emlun:
Also, perhaps link some of the terms throughout? I can make a PR of that into this PR if you can't run bikeshed.

selfissued:
@emlun if you can create a PR into my PR addressing the issues you raise that would great.  Then I could apply it in my morning and merge the combination into master.

PR #772, opened 2018-02-01
hmac-secret extension

akshayku:
Description for hmac-secret extension.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/pull/772.html" title="Last updated on Feb 1, 2018, 11:36 PM GMT (60ed25f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/772/1fc8906...60ed25f.html" title="Last updated on Feb 1, 2018, 11:36 PM GMT (60ed25f)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -3991,6 +3991,72 @@ This [=registration extension=] and [=authentication extension=] enables use of 
                     01           -- Subitem 3: CBOR short for Matcher Protection Type Software
     </pre>
 
+## HMAC Secret Extension (hmac-secret) ## {#sctn-hmac-secret-extension}
+
+This [=registration extension=] and [=authentication extension=] allows for a getting a symmetric secret from the authenticator.
+
+: Extension identifier
+:: `hmac-secret`
+
+: Client extension input
+:: {{CredentialsContainer/create()}} : A boolean value to indicate that this extension is requested by the Relying Party.
+      <pre class="idl">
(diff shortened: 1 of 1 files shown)


agl:
At the meeting I raised some concerns about the cryptography. That's a CTAP2 issue, however, and I have filed [this PR](https://github.com/fido-alliance/fido-2-specs/pull/445) to try and address that in a non-breaking way.

I had thought that the `hmac-secret` extension was going to be externally defined and thus did not expect to see it in webauthn. Also, this PR references the CTAP2 spec for details of the authenticator inputs / outputs, but I didn't think that was going in CTAP2.

Assuming that it is going to be specified in webauthn & CTAP2, the naming of some of the variables seems to be a little specific to a single use-case. The salts are called `salt1` and `salt2`, reasonably enough, but then I would expect the outputs to be `output1` and `output2` rather than `offlineCurrentHash` and `offlineNewHash`.

akshayku:
I will change outputs to `output1` and `output2`. Also, I met with Mike and we decided that defining everything in CTAP2 spec is better as extension refers to some sections present in CTAP2 spec. I will define everything in CTAP2 spec and IANA registry will point to CTAP2 spec. I am closing this PR. Regarding cryptography, thanks for the feedback for future extensions.

PR #776, opened 2018-02-06
Revise §9.5. "Authenticator extension processing"

### LINKED ISSUE #775: §9.5. "Authenticator extension processing" is ambiguous
As discussed in #724, [§9.5. Authenticator extension processing][sec] could use a revision to make it less ambiguous.

>The CBOR authenticator extension input value of each processed authenticator extension is included in the extensions data part of the authenticator request. This part is a CBOR map, with CBOR extension identifier values as keys, and the CBOR authenticator extension input value of each extension as the value.

This section can be interpreted as both of the following:

- The authenticator extension inputs as a whole is a CBOR map, where the values are not necessarily maps.
- Each value in the authenticator extension inputs is a CBOR map.

[sec]: https://w3c.github.io/webauthn/#sctn-authenticator-extension-processing

emlun:
This aims to fix #775.

I'm not entirely happy with it, but I think it's at least an improvement of section §9.5. What bugs me most is that it feels like "the result of authenticator extension processing" could be taken to mean processing the entire map or just one of the key/value pairs. It feels like it should be possible to rephrase the

>Let _processedExtensions_ be the result of **authenticator extension processing** for each supported _extensionId_ → _authenticatorExtensionInput_ in _extensions_.

steps of the authenticator operations so they don't require iterating over the map, but "authenticator extension processing" currently refers specifically to the processing for a single extension.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/pull/776.html" title="Last updated on Feb 6, 2018, 9:29 PM GMT (894c2f5)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/776/572446d...894c2f5.html" title="Last updated on Feb 6, 2018, 9:29 PM GMT (894c2f5)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2290,8 +2290,8 @@ It takes the following input parameters:
     these are known to the authenticator, it should not create a new credential. |excludeCredentialDescriptorList| contains a 
     list of known credentials.
 : |extensions|
-:: A [=map=] from [=extension identifiers=] to their [=authenticator extension inputs=], created by the client based on the
-    extensions requested by the [=[RP]=], if any.
+:: A [=CBOR=] [=map=] from [=extension identifiers=] to their [=authenticator extension inputs=], created by the client based on
+    the extensions requested by the [=[RP]=], if any.
 
 Note: Before performing this operation, all other operations in progress in the [=authenticator session=] MUST be aborted by 
 running the [=authenticatorCancel=] operation.
@@ -2399,8 +2399,8 @@ It takes the following input parameters:
 : |requireUserVerification|
 :: The [=effective user verification requirement for assertion=], a Boolean value provided by the client.
 : |extensions|
-:: A [=map=] from [=extension identifiers=] to their [=authenticator extension inputs=], created by the client based on the
-    extensions requested by the [=[RP]=], if any.
+:: A [=CBOR=] [=map=] from [=extension identifiers=] to their [=authenticator extension inputs=], created by the client based on
+    the extensions requested by the [=[RP]=], if any.
 
 Note: Before performing this operation, all other operations in progress in the [=authenticator session=] must be aborted by running the [=authenticatorCancel=] operation.
 
--- a/index.bs
+++ b/index.bs
@@ -3613,16 +3613,18 @@ the process by which the [=CBOR=] [=authenticator extension output=] can be used
 
 ## <dfn>Authenticator extension processing</dfn> ## {#sctn-authenticator-extension-processing}
 
-The [=CBOR=] [=authenticator extension input=] value of each processed
-[=authenticator extension=] is included in the extensions data part of the authenticator request. This part is a CBOR map, with
-[=CBOR=] [=extension identifier=] values as keys, and the [=CBOR=] [=authenticator extension input=] value of each extension as
-the value.
+The [=CBOR=] [=authenticator extension input=] value of each processed [=authenticator extension=] is included in the |extensions|
+parameter of the [=authenticatorMakeCredential=] and [=authenticatorGetAssertion=] operations. The |extensions| parameter is a
+[=CBOR=] map where each key is an [=extension identifier=] and the corresponding value is the [=authenticator extension input=]
+for that extension.
 
-Likewise, the extension output is represented in the [=authenticator data=] as a CBOR map with [=CBOR=] [=extension
-identifiers=] as keys, and the [=CBOR=] <dfn>authenticator extension output</dfn> value of each extension as the value.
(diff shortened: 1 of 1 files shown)


nadalin:
@selfissued @equalsJeffH Can you all please review as soon as possible

index.bs
```diff
@@ -2431,8 +2431,8 @@ When this method is invoked, the [=authenticator=] MUST perform the following pr
     If the user does not [=user consent|consent=], return an error code equivalent to
     "{{NotAllowedError}}" and terminate the operation.
 
-1. Let |processedExtensions| be the result of [=authenticator extension processing=] for each supported [=extension
-    identifier=]/input pair in |extensions|.
+1. Let |processedExtensions| be the result of [=authenticator extension processing=] [=map/for each=] supported <var
+    ignore>extensionId</var> → <var ignore>authenticatorExtensionInput</var> in |extensions|.
```
selfissued: Why did you remove the indexed term [=extension identifier=]?  If anything, I think this should say "... supported [=extension identifier=] -> [=authenticator extension input=] pair".  Please keep the indexed term and add the second one.

Also, why were you using "var ignore"?  I don't understand your goals here.

index.bs
```diff
@@ -3562,7 +3562,7 @@ extension=] that does not otherwise require any result values MUST return a valu
 An extension defines one or two request arguments. The <dfn>client extension input</dfn>,
 which is a value that can be encoded in JSON, is passed from the [=[RP]=] to the client
 in the {{CredentialsContainer/get()}} or {{CredentialsContainer/create()}} call,
-while the [=CBOR=] <dfn>authenticator extension input</dfn> is
+while the [=CBOR=] [=authenticator extension input=] is
```
selfissued: I disagree with removing the dfn here and moving it to the processing steps.  It makes more sense here.  Please revert this change.

index.bs
```diff
@@ -3613,16 +3613,18 @@ the process by which the [=CBOR=] [=authenticator extension output=] can be used
 
 ## <dfn>Authenticator extension processing</dfn> ## {#sctn-authenticator-extension-processing}
 
-The [=CBOR=] [=authenticator extension input=] value of each processed
-[=authenticator extension=] is included in the extensions data part of the authenticator request. This part is a CBOR map, with
-[=CBOR=] [=extension identifier=] values as keys, and the [=CBOR=] [=authenticator extension input=] value of each extension as
-the value.
+The [=CBOR=] <dfn>authenticator extension input</dfn> value of each processed [=authenticator extension=] is included in the
```
selfissued: Please remove the "dfn" here.

index.bs
```diff
@@ -3613,16 +3613,18 @@ the process by which the [=CBOR=] [=authenticator extension output=] can be used
...
+|extensions| parameter of the [=authenticatorMakeCredential=] and [=authenticatorGetAssertion=] operations. The |extensions|
+parameter is a [=CBOR=] map where each key is an [=extension identifier=] and the corresponding value is the [=authenticator
+extension input=] for that extension.
 
-Likewise, the extension output is represented in the [=authenticator data=] as a CBOR map with [=CBOR=] [=extension
-identifiers=] as keys, and the [=CBOR=] <dfn>authenticator extension output</dfn> value of each extension as the value.
+Likewise, the <dfn>authenticator extension output</dfn> value of each processed [=authenticator extension=] is represented by one
```
selfissued: This is just wrong.  The "authenticator extension output" is the value of the key/value pair.  You're using the term in a way that makes it sound like it's both the key and the value.  Please revert this change.

index.bs
emlun: I made this change to closer reflect the usage description of `map/for each` in Infra. The `ignore` on the `<var>`s was then necessary to make Bikeshed not complain about that the variables were only used once.

I'll change that back to using the indexed terms.

index.bs
emlun: Ok.

emlun: Ok. The reason I did this was so that the `<dfn>` would be in the same section as the corresponding one for the authenticator outputs.

index.bs
emlun: Ah yes, sorry.

index.bs
emlun: Ok.

emlun:
@selfissued Thanks for reviewing. Do you mind if I squeeze #778 in as well?

selfissued:
I will merge this on Friday unless @gmandyam objects by then.

gmandyam:
I am fine with the proposed text.  Please merge now.

PR #777, opened 2018-02-06
fix #204: factor makeCred into reg & authn priv considerations

### LINKED ISSUE #204: Privacy across Account IDs

equalsJeffH:
fixes #204 
see https://github.com/w3c/webauthn/issues/204#issuecomment-359126232
and https://github.com/w3c/webauthn/pull/687#issuecomment-359087435
and https://github.com/w3c/webauthn/pull/687#issuecomment-359114271

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/pull/777.html" title="Last updated on Feb 7, 2018, 4:14 PM GMT (381bbf0)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/777/f13e030...381bbf0.html" title="Last updated on Feb 7, 2018, 4:14 PM GMT (381bbf0)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -4607,12 +4607,12 @@ or {{PublicKeyCredentialRequestOptions/allowCredentials}}, as applicable:
 - A named [=public key credential|credential=] is not available.
 - A named [=public key credential|credential=] is available, but the user does not [=user consent|consent=] to use it.
 
-If the above cases are distinguishable, information is leaked by which a malicious [=[RP]=] could identify the user by probing for
-which [=public key credential|credentials=] are available. For example, one such information leak is if the client returns a
-failure response as soon as the user denies [=user consent|consent=] to proceed with the operation. In this case the [=[RP]=]
-could detect that the operation was canceled by the user and not the timeout, and thus conclude that at least one of the
-[=public key credential|credentials=] listed in the {{PublicKeyCredentialRequestOptions/allowCredentials}} parameter is
-available to the user.
+If the above cases are distinguishable, information is leaked by which a malicious [=[RP]=] could identify the user by probing
+for which [=public key credential|credentials=] are available. For example, one such information leak is if the client returns a
+failure response as soon as the user denies [=user consent|consent=] to proceed with a [=registration=] or [=authentication=]
+[=ceremony=]. In this case the [=[RP]=] could detect that the [=ceremony=] was canceled by the user and not the timeout, and
+thus conclude that at least one of the [=public key credential|credentials=] listed in the
+{{PublicKeyCredentialRequestOptions/allowCredentials}} parameter is available to the user.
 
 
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -4595,21 +4595,24 @@ in several ways, including:
...
+not leak information that could enable a malicious [=[RP]=] to distinguish between these cases, where "named" means that the
+[=public key credential|credential=] is listed by the [=[RP]=] in either {{MakePublicKeyCredentialOptions/excludeCredentials}}
+or {{PublicKeyCredentialRequestOptions/allowCredentials}}, as applicable:
 
 - A named [=public key credential|credential=] is not available.
 - A named [=public key credential|credential=] is available, but the user does not [=user consent|consent=] to use it.
 
 If the above cases are distinguishable, information is leaked by which a malicious [=[RP]=] could identify the user by probing for
 which [=public key credential|credentials=] are available. For example, one such information leak is if the client returns a
 failure response as soon as the user denies [=user consent|consent=] to proceed with the operation. In this case the [=[RP]=]
```
emlun: Perhaps change

>as soon as the user denies [=user consent|consent=] to proceed with the operation

to something like this?

>as soon as the user denies [=user consent|consent=] to proceed with an [=authentication=] [=ceremony=]

Seeing as the following sentence refers specifically to `allowCredentials`, I mean.

nadalin:
@equalsJeffH can you finish the merge here ?

index.bs
equalsJeffH: incorp'd a build on this suggestion, pls review

equalsJeffH:
incorp'd a build on suggestion https://github.com/w3c/webauthn/pull/777#discussion_r166447515, pls review

index.bs
emlun: What I was trying to say is that the conclusion "at least one of the credentials listed in the allowCredentials parameter is available" is applicable only to the case of authentication ceremonies, not registration ceremonies. For registration ceremonies the conclusion is instead that the user has an authenticator that is not registered, which probably isn't as big a privacy concern.

PR #789, opened 2018-02-07
Add instructions for passing unknown extensions to authenticators

### LINKED ISSUE #738: UAs can't pass unknown extensions to authenticators

selfissued:
Fixes #738

***
<a href="https://pr-preview.s3.amazonaws.com/selfissued/webauthn/pull/789.html" title="Last updated on Feb 12, 2018, 8:21 PM GMT (86f9f5b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/789/ca4cf0f...selfissued:86f9f5b.html" title="Last updated on Feb 12, 2018, 8:21 PM GMT (86f9f5b)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -3531,24 +3531,24 @@ understand by encoding the [=authenticator extension output=] value into JSON, p
 present in JSON.
 
 When clients choose to pass through extensions they do not recognize,
-the JavaScript values in the [=client extension input=] values
-are converted to [=CBOR=] values in the [=authenticator extension input=] values.
-When the JavaScript input value is an [=%ArrayBuffer%=], it is converted to a [=CBOR=] byte array.
-When the JavaScript input value is a non-integer number, it is converted to a 64-bit CBOR floating point number.
-Otherwise, when the JavaScript input type corresponds to a JSON type, the conversion is done
+the JavaScript values in the [=client extension inputs=]
+are converted to [=CBOR=] values in the [=authenticator extension inputs=].
+When the JavaScript value is an [=%ArrayBuffer%=], it is converted to a [=CBOR=] byte array.
+When the JavaScript value is a non-integer number, it is converted to a 64-bit CBOR floating point number.
+Otherwise, when the JavaScript type corresponds to a JSON type, the conversion is done
 using the rules defined in Section 4.2 of [[RFC7049]] (Converting from JSON to CBOR),
-but operating on JavaScript inputs rather than JSON inputs.
+but operating on inputs of JavaScript type values rather than inputs of JSON type values.
 Once these conversions are done,
 canonicalization of the resulting [=CBOR=] MUST be performed using the [=CTAP2 canonical CBOR encoding form=].
 
 Likewise, when clients receive outputs from extensions they have passed through that they do not recognize,
-the [=CBOR=] values in the [=authenticator extension output=] values
-are converted to JavaScript values in the [=client extension output=] values.
-When the CBOR input value is a byte string, it is converted to a JavaScript [=%ArrayBuffer%=]
+the [=CBOR=] values in the [=authenticator extension outputs=]
+are converted to JavaScript values in the [=client extension outputs=].
+When the CBOR value is a byte string, it is converted to a JavaScript [=%ArrayBuffer%=]
 (rather than a base64url-encoded string).
-Otherwise, when the CBOR input type corresponds to a JSON type, the conversion is done
+Otherwise, when the CBOR type corresponds to a JSON type, the conversion is done
 using the rules defined in Section 4.1 of [[RFC7049]] (Converting from CBOR to JSON),
-but producing JavaScript outputs rather than JSON outputs.
+but producing outputs of JavaScript type values rather than outputs of JSON type values.
 
 The IANA "WebAuthn Extension Identifier" registry established by [[!WebAuthn-Registries]] can be consulted
 for an up-to-date list of registered WebAuthn Extensions.
--- a/index.bs
+++ b/index.bs
@@ -3530,6 +3530,26 @@ operation. Likewise, clients can choose to produce a [=client extension output=]
 understand by encoding the [=authenticator extension output=] value into JSON, provided that the CBOR output uses only types
 present in JSON.
 
+When clients choose to pass through extensions they do not recognize,
+the JavaScript values in the [=client extension input=] values
+are converted to [=CBOR=] values in the [=authenticator extension input=] values.
+When the JavaScript input value is an [=%ArrayBuffer%=], it is converted to a [=CBOR=] byte array.
+When the JavaScript input value is a non-integer number, it is converted to a 64-bit CBOR floating point number.
+Otherwise, when the JavaScript input type corresponds to a JSON type, the conversion is done
+using the rules defined in Section 4.2 of [[RFC7049]] (Converting from JSON to CBOR),
+but operating on JavaScript inputs rather than JSON inputs.
+Once these conversions are done,
+canonicalization of the resulting [=CBOR=] MUST be performed using the [=CTAP2 canonical CBOR encoding form=].
+
+Likewise, when clients receive outputs from extensions they have passed through that they do not recognize,
+the [=CBOR=] values in the [=authenticator extension output=] values
+are converted to JavaScript values in the [=client extension output=] values.
+When the CBOR input value is a byte string, it is converted to a JavaScript [=%ArrayBuffer%=]
+(rather than a base64url-encoded string).
+Otherwise, when the CBOR input type corresponds to a JSON type, the conversion is done
+using the rules defined in Section 4.1 of [[RFC7049]] (Converting from CBOR to JSON),
+but producing JavaScript outputs rather than JSON outputs.
+
 The IANA "WebAuthn Extension Identifier" registry established by [[!WebAuthn-Registries]] can be consulted
 for an up-to-date list of registered WebAuthn Extensions.
 
--- a/index.bs
+++ b/index.bs
@@ -3550,6 +3550,10 @@ Otherwise, when the CBOR type corresponds to a JSON type, the conversion is done
 using the rules defined in Section 4.1 of [[RFC7049]] (Converting from CBOR to JSON),
 but producing outputs of JavaScript type values rather than outputs of JSON type values.
 
+Note that some clients may choose to implement this pass-through capability under a feature flag.
+Supporting this capability can facilitate innovation, allowing authenticators to experiment with new extensions
+and [=[RPS]=] to use them before there is explicit support for them in clients.
+
 The IANA "WebAuthn Extension Identifier" registry established by [[!WebAuthn-Registries]] can be consulted
 for an up-to-date list of registered WebAuthn Extensions.
 


index.bs
```diff
@@ -3530,6 +3530,26 @@ operation. Likewise, clients can choose to produce a [=client extension output=]
 understand by encoding the [=authenticator extension output=] value into JSON, provided that the CBOR output uses only types
 present in JSON.
 
+When clients choose to pass through extensions they do not recognize,
+the JavaScript values in the [=client extension input=] values
```
agl: (language nit: could go either way:) I think there are too many "values" here and would drop the instances that follow &ldquo;input=] &rdquo;. (Same below.)

index.bs
```diff
@@ -3530,6 +3530,26 @@ operation. Likewise, clients can choose to produce a [=client extension output=]
...
+are converted to [=CBOR=] values in the [=authenticator extension input=] values.
+When the JavaScript input value is an [=%ArrayBuffer%=], it is converted to a [=CBOR=] byte array.
+When the JavaScript input value is a non-integer number, it is converted to a 64-bit CBOR floating point number.
```
agl: I thought that floats were banned in CTAP2 CBOR? Even if not, I would tend to ban them here:

Consider something like the geolocation extension. If it wished to send floats then it would need to ensure that the values in Javascript were never exact integers. Otherwise, as one moves across a line of latitude / longitude, the type signature of the CBOR would change. That seems fragile.

Rather, I would suggest that values are always rounded and that extensions use a fixed-point representation for fractional values.

index.bs
selfissued: Thanks - I'll work on the exposition.

index.bs
selfissued: Issue https://github.com/fido-alliance/fido-2-specs/issues/419 is exactly about making sure that floating point values can be present in CTAP2 canonical CBOR - motivated in part by the geolocation extension.

The geolocation is a test case that we want to work.  Rather than banning floats, we should define that they are doubles unless integers - as the text already does.  It will be infinitesimally rare that longitude and latitude values are integers, so it will nearly always work.  Rounding longitude/latitude values to integers introduces errors of more than 60 nautical miles - a non-starter!

selfissued: https://github.com/fido-alliance/fido-2-specs/pull/456 defines that floats are allowed by CTAP2 canonicalization (and are not changed by it).

index.bs
```diff
@@ -3530,6 +3530,26 @@ operation. Likewise, clients can choose to produce a [=client extension output=]
...
+Otherwise, when the JavaScript input type corresponds to a JSON type, the conversion is done
+using the rules defined in Section 4.2 of [[RFC7049]] (Converting from JSON to CBOR),
+but operating on JavaScript inputs rather than JSON inputs.
+Once these conversions are done,
+canonicalization of the resulting [=CBOR=] MUST be performed using the [=CTAP2 canonical CBOR encoding form=].
+
+Likewise, when clients receive outputs from extensions they have passed through that they do not recognize,
+the [=CBOR=] values in the [=authenticator extension output=] values
+are converted to JavaScript values in the [=client extension output=] values.
+When the CBOR input value is a byte string, it is converted to a JavaScript [=%ArrayBuffer%=]
```
equalsJeffH: by "CBOR input value", are you referring to a value amongst the client extension output?  I'm finding this confusing.

index.bs
```diff
@@ -3530,6 +3530,26 @@ operation. Likewise, clients can choose to produce a [=client extension output=]
...
 present in JSON.
 
+When clients choose to pass through extensions they do not recognize,
+the JavaScript values in the [=client extension input=] values
+are converted to [=CBOR=] values in the [=authenticator extension input=] values.
+When the JavaScript input value is an [=%ArrayBuffer%=], it is converted to a [=CBOR=] byte array.
+When the JavaScript input value is a non-integer number, it is converted to a 64-bit CBOR floating point number.
+Otherwise, when the JavaScript input type corresponds to a JSON type, the conversion is done
+using the rules defined in Section 4.2 of [[RFC7049]] (Converting from JSON to CBOR),
+but operating on JavaScript inputs rather than JSON inputs.
```
equalsJeffH: s/ JavaScript inputs rather than JSON inputs / inputs of JavaScript type values rather than JSON-encoded inputs /   ...?

should perhaps generally reference https://tc39.github.io/ecma262/#sec-ecmascript-data-types-and-values ?

index.bs
```diff
@@ -3530,6 +3530,26 @@ operation. Likewise, clients can choose to produce a [=client extension output=]
...
+canonicalization of the resulting [=CBOR=] MUST be performed using the [=CTAP2 canonical CBOR encoding form=].
+
+Likewise, when clients receive outputs from extensions they have passed through that they do not recognize,
+the [=CBOR=] values in the [=authenticator extension output=] values
+are converted to JavaScript values in the [=client extension output=] values.
+When the CBOR input value is a byte string, it is converted to a JavaScript [=%ArrayBuffer%=]
+(rather than a base64url-encoded string).
+Otherwise, when the CBOR input type corresponds to a JSON type, the conversion is done
+using the rules defined in Section 4.1 of [[RFC7049]] (Converting from CBOR to JSON),
+but producing JavaScript outputs rather than JSON outputs.
```
equalsJeffH: s/ JavaScript outputs rather than JSON outputs / outputs of JavaScript type values rather than JSON-encoded outputs/   ...?

nadalin:
@agl are you ok with Mike's repsonse ? @equalsJeffH can you please review

selfissued:
Thanks for the wording suggestions, Jeff.  I plan to update the PR to incorporate them later today.

index.bs
agl: If we go this route, the type-signature of extensions will fluctuate depending on the values involved, i.e. from float to int when crossing a line of latitude / longitude. This seems like a pool of bugs waiting to happen and it would be more stable to translate Javascript numbers as CBOR floats in all cases if we wish to have a generic transformation defined.

Also, the geolocation extension doesn't need floats. As noted above, they could just use fixed-point integers. (I.e. express a latitude as the integer round(x * 1000000), for whatever scale-factor works in their context.)

selfissued: The geolocation extension uses the Coordinates type defined at https://www.w3.org/TR/geolocation-API/#coordinates_interface , which does use floats (doubles).  It would be silly/unhelpful to define a different equivalent type in a W3C spec.

The generic conversion will always involve some tradeoffs.  Given that most numbers used in these interfaces are integers, representing them as such is certainly the right thing to do.  Prohibiting floats would be wrong/unhelpful too.

So yes, the generic conversion has tradeoffs, but I think we've written down the right ones for WebAuthn and CTAP2.

agl: It would not be silly to define the geolocation extension in terms of a scaled version of that structure in order to eliminate all the complexity of floats throughout this spec, I feel.

In fact, I think this has convinced me more strongly that the generic transformation is a mistake in general. I'll write a more general note about that below.

equalsJeffH:
@agl's [comment](https://github.com/w3c/webauthn/pull/789#pullrequestreview-95256815) is certainly food for thought.

selfissued:
Actually, authenticators don't have to deal with two versions of an extension.  If the generic conversion doesn't result in a syntactically legal request, they will drop it, which is fine.  Also, the spec already says that extensions must be defined in such a way that the generic transform won't result in security or privacy problems.  For most extensions, the generic transform result will be identical to the correct result - by design.  That's what makes the generic transform useful and worth specifying.

If experience shows that no browser chooses to do this between CR and the final specification, then we can drop it for final.  But I think it should be defined for CR so that the option is present for browser vendors to evaluate.

emlun:
>@agl: Taking the geolocation extension as an example: browsers would not want that to bypass their geolocations permissions framework.

>@selfissued: Also, the spec already says that extensions must be defined in such a way that the generic transform won't result in security or privacy problems.

This, then, looks to me like the spec is internally inconsistent. A generic transform of the [`loc`][loc] extension _will_ be passed through as a valid authenticator extension input, the authenticator (if it supports it, of course) will happily include location data in its response, and the client will pass that on to the RP in the authenticator data. The browser would have to actively **opt out** of generically transforming the `loc` extension to prevent it from circumventing the browser's location access controls (because the location data source is the authenticator, not the browser).

On the other hand, the authenticator is always assumed to be trusted, right? In some sense, then, the user _does_ consent to sharing location information if they consent to proceeding with the operation, even if this happens to circumvent the browser's access controls - but I guess the issue arises if it's not obvious (or even visible) to the user that the RP requests location data. It doesn't seem unreasonable to expect that someone might market an authenticator with a GPS chip but no display capability.

I think the specific case of the `loc` extension could be solved by either (1) specifying that the authenticator MUST obtain clear consent to share the location, or by (2) having the client provide the location data instead of the Boolean `true` as the authenticator input. Perhaps some strategy like that can be generalized to cover future extensions too?

[loc]: https://pr-preview.s3.amazonaws.com/selfissued/webauthn/pull/789.html#sctn-location-extension

agl:
> Actually, authenticators don't have to deal with two versions of an extension. If the generic conversion doesn't result in a syntactically legal request, they will drop it, which is fine.

So if a specific transformation is defined for an extension (that alters the structure) then the "pass through" properties of the generic transform (i.e. no need to update browsers) wouldn't work, but extensions won't want that so will either have authenticators accept both, or will avoid defining specific transforms even when it would make sense.

It also gives malicious sites a larger attack surface: they can send arbitrary valid CBOR to authenticators. (I have two tokens that try to implement the much simpler CTAP1 protocol and crash when processing certain, valid commands!) 

> If experience shows that no browser chooses to do this between CR and the final specification, then we can drop it for final.

Looking at it from the other direction, does any browser intend to implement this? As noted, I don't strongly object if there's demand, I would just go the other path.

nadalin:
@selfissued Where do we stand on this and the comments that @agl and @equalsJeffH have stated

selfissued:
I need to apply the editorial improvements that @equalsJeffH suggested.

In thinking about @agl 's comments some more, one of my concerns is that if browsers don't implement the ability for RPs to call new extensions that the browsers don't yet understand, it will stifle innovation, as authenticator vendors won't be able to implement new extensions and have RPs have any way to use them.  The browsers would become the gatekeepers, preventing legitimate communication between RPs and the authenticators that they use.

At the very least, I believe that browsers should support passing through unknown extensions under a feature flag, so new extensions can be developed, tested, and used.  In that way, the browsers will "get out of the way", rather than being an impediment to innovation.

nadalin:
@agl please review

equalsJeffH:
Agree with @selfissued's feature-flag suggestion: https://github.com/w3c/webauthn/pull/789#issuecomment-364536969

Perhaps a Note is warranted?

Otherwise text seems OK, tho as @agl notes, RPs can send arbitrary (CBOR-encoded) data to authenticators, if this functionality is enabled.  FWIW, this threat _is_ addressed in the FIDO Security Reference: T-1.4.9, T-1.4.30.

agl:
Looks good to me editorially. My other reservations stand. If other browsers are desire to implement this, however, then sure.

agl:
> In thinking about @agl 's comments some more, one of my concerns is that if browsers don't implement the ability for RPs to call new extensions that the browsers don't yet understand, it will stifle innovation, as authenticator vendors won't be able to implement new extensions and have RPs have any way to use them. 

On this topic (and thinking out aloud: this is not a commitment on Chrome's part) we could overload the current enterprise policy that we have for allowing attestation without a prompt to also enable extension pass-through on a per-RP basis if groups wish to develop extensions. However, I'm not sure that it's crazy that groups doing extension development build a custom Chromium or Firefox build to add support for there experimental extension.

selfissued:
Thanks @equalsJeffH and @agl .  I plan to add a note about the possibility of doing pass-through under a feature flag, as Jeff suggested, and then merge.

Issue #803, opened 2018-02-15
agl doesn't understand extensions


agl:
(This is just a request for clarification to @selfissued, but I'm writing it as an issue in case it ends up being helpful for others to see it.)

In the course of implementing the `appid` extension, I hit a few points where my understanding is deficient:

Why are the client inputs contained in the [CollectedClientData](https://w3c.github.io/webauthn/#sec-client-data)? I see that the RP could learn from this which extensions the client supports but, if so, a) why also [echo a boolean](https://w3c.github.io/webauthn/#sctn-appid-extension) to indicate this in `appid` and b) why echo the full input when a boolean would do? (Echoing the full input value is a pain because we're unmarshaling them from Javascript but then have to marshal internal structures back to JSON for this.)

Is the RP expected to verify that its inputs were faithfully reflected? The [instructions say](https://w3c.github.io/webauthn/#registering-a-new-credential) (step 7) to do a subset check, which could imply checking the full value, but it says the same thing about the `authenticatorExtensions`, and the RP can't check the value of those. (Unless it's supposed to verify the client's translation to CBOR? If so, why?)

Even if the RP does verify the extension fields in the CollectedClientData, that doesn't mean that the client passed those exact values to the authenticator [via CTAP2](https://fidoalliance.org/specs/fido-v2.0-rd-20170927/fido-client-to-authenticator-protocol-v2.0-rd-20170927.html#authenticatorGetAssertion) so I'm unclear why the RP should bother.

In a similar vein, why are the authenticator inputs included in the [CollectedClientData](https://w3c.github.io/webauthn/#sec-client-data) too?

On the way back out, step 18.3 of [these instructions](https://w3c.github.io/webauthn/#discover-from-external-source) says that [[clientExtensionsResults]] is an &ldquo;ArrayBuffer &hellip; containing the bytes of assertionCreationData.clientExtensionResults&rdquo;. But the latter is a [AuthenticationExtensionsClientOutputs](https://w3c.github.io/webauthn/#dictdef-authenticationextensionsclientoutputs) so how is that turned into a bytestring?

Lastly, why is `getClientExtensionResults` [a function](https://w3c.github.io/webauthn/#iface-pkcredential) and not simply a member of the `PublicKeyCredential` with type `AuthenticationExtensionsClientOutputs`? That seems uniquely out of place in this API.

Thank you for any clarity you can provide.

emlun:
[...]

PR #810, opened 2018-02-20
Change from getClientExtensionResults function to clientExtensionResults attribute

selfissued:
Fixes part of #803

***
<a href="https://pr-preview.s3.amazonaws.com/selfissued/webauthn/pull/810.html" title="Last updated on Feb 21, 2018, 10:31 PM GMT (c6bb3fe)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/810/05335d4...selfissued:c6bb3fe.html" title="Last updated on Feb 21, 2018, 10:31 PM GMT (c6bb3fe)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -620,7 +620,7 @@ that are returned to the caller when a new credential is created, or a new asser
     interface PublicKeyCredential : Credential {
         [SameObject] readonly attribute ArrayBuffer              rawId;
         [SameObject] readonly attribute AuthenticatorResponse    response;
-        AuthenticationExtensionsClientOutputs getClientExtensionResults();
+        [SameObject] readonly attribute AuthenticationExtensionsClientOutputs clientExtensionResults;
     };
 </xmp>
 <dl dfn-type="attribute" dfn-for="PublicKeyCredential">
@@ -639,8 +639,8 @@ that are returned to the caller when a new credential is created, or a new asser
         the {{PublicKeyCredential}} was created in response to {{CredentialsContainer/get()}}, and this attribute's value
         will be an {{AuthenticatorAssertionResponse}}.
 
-    :   {{PublicKeyCredential/getClientExtensionResults()}}
-    ::  This operation returns the value of {{PublicKeyCredential/[[clientExtensionsResults]]}}, which is a [=map=] containing
+    :   <dfn>clientExtensionResults</dfn>
+    ::  The {{PublicKeyCredential/[[clientExtensionsResults]]}} value is a [=map=] containing
         [=extension identifier=] → [=client extension output=] entries produced by the extension's
         [=client extension processing=].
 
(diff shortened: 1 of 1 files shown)


nadalin:
@selfissued Please review @emlun comment

selfissued:
@emlun - you wrote "It looks like the [[clientExtensionsResults]] internal slot (added in 5c8dc49) is no longer needed, correct?".  I suspect you're right but note that the new attribute is calledclientExtensionsResults and so that has to stay.  Could you create a PR against my PR to show us what you think should change?

emlun:
There's both the attribute [`clientExtensionResults`][attr] and the internal slot [`[[clientExtensionsResults]]`][internal] now. That seems a bit redundant to me, and I believe the latter was introduced only because the former was converted into an operation. I can write a PR but I'm not entirely sure how to merge the two descriptions.

[attr]: https://pr-preview.s3.amazonaws.com/selfissued/webauthn/pull/810.html#dom-publickeycredential-clientextensionresults
[internal]: https://pr-preview.s3.amazonaws.com/selfissued/webauthn/pull/810.html#dom-publickeycredential-clientextensionsresults-slot

emlun:
PR submitted: https://github.com/selfissued/webauthn/pull/4

selfissued:
Fixes #803

ttaubert:
Sorry for the late comment, but I just noticed that per [1]:

"The type of the attribute, after resolving typedefs, must not be a nullable or non-nullable version of any of the following types: ... a dictionary type"

[1] https://heycam.github.io/webidl/#idl-attributes

emlun:
Oh. I think that's exactly the same rule that made us change the previous record-type attribute to an operation.

So I guess we'll have to revert this, then? @equalsJeffH @selfissued

selfissued:
@agl and @jcjones , do you believe that Tim's comment means that we have to revert this?

agl:
I'm not immediately convinced that we need to revert this, although some more spec lawyering may be needed.

We have the `response` member of `PublicKeyCredential` already which is a structure with fields and such. That's all we want for `clientExtensionResults` too. So maybe we need to call `AuthenticationExtensionsClientOutputs` an interface rather than a dictionary, but it certainly looks like we can get the desired shape without a weird callback to do it.

I have to get to a meeting now, but I'll try to find someone who claims to be a WebIDL expert to weigh in.

selfissued:
Thanks @agl

ttaubert:
I'm certainly not a WebIDL expert, but IIUIC interfaces have only required members. And those can't have any default values either. If that's the case an interface probably wouldn't be a good fit for `AuthenticationExtensionsClientOutputs`.

selfissued:
Adding @nadalin so he's aware of this discussion

agl:
I've asked @jyasskin to weigh in. However, I'm not sure that the lack of optional members is a problem here. This is the browser's extension output, thus the structure only contains fields for extensions that the browser supports. It's typical for browsers that don't support something not to have certain fields (I think). For example, we suggest testing for `window.PublicKeyCredential` to tell whether webauthn is supported or not.

agl:
To summarise what I've managed to get from emailing people who do know WebIDL:

- Indeed, we cannot have an interface attribute with a dictionary value for unknown reasons.
- We can define `AuthenticationExtensionsClientOutputs` as a series of `partial interface` chunks, one for each extension, and with the expectation that browsers will only implement a subset of the attributes. That, however, means that there'll probably be null-valued attributes for all the extensions that a client supports, even if those extensions weren't used in a given request.
- We can also have a function returning a dictionary (i.e. what we had).

I don't have a strong preference either way. I think the Javascript is a little cleaner if it's an interface, but not having null-valued attributes hanging around might swing it for some people.

selfissued:
Having the extraneous null-valued attributes for extensions that were not requested by the RP sounds pretty sub-optimal.  I hate to say it, but it seems like things will be simpler if we revert this change and go back to the function.  I'll create a new PR to do so and submit it to the commenters on this thread to review.

nadalin:
@selfissued Should this be closed now ?

selfissued:
It's merged, which is the closed state for PRs.

nadalin:
@selfissued It wasn't when I wrote the comment

PR #813, opened 2018-02-21
Fix #712: Refer to the JSON object as %JSON%

### LINKED ISSUE #712: JSON-serialized client data is wrong
You basically specify the correct thing by only allowing UTF-8 and such, but https://infra.spec.whatwg.org/#parse-json-from-bytes is a little more formally correct.

emlun:
This fixes #712.

***
<a href="https://pr-preview.s3.amazonaws.com/emlun/webauthn/pull/813.html" title="Last updated on Feb 21, 2018, 1:19 PM GMT (9a091f2)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/813/3a3700a...emlun:9a091f2.html" title="Last updated on Feb 21, 2018, 1:19 PM GMT (9a091f2)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -69,6 +69,7 @@ spec: ECMAScript; urlPrefix: https://tc39.github.io/ecma262/#
         for: JSON; text: stringify; url: sec-json.stringify
     type: dfn
         text: %ArrayBuffer%; url: sec-arraybuffer-constructor
+        text: %JSON%; url: sec-json-object
         url: sec-object-internal-methods-and-internal-slots
             text: internal method
             text: internal slot
@@ -309,7 +310,7 @@ below and in [[#index-defined-elsewhere]].
 :: {{DOMException}} and the DOMException values used in this specification are defined in [[!DOM4]].
 
 : ECMAScript
-:: [=%ArrayBuffer%=] is defined in [[!ECMAScript]].
+:: [=%ArrayBuffer%=] and [=%JSON%=] are defined in [[!ECMAScript]].
 
 : HTML
 :: The concepts of [=relevant settings object=], [=origin=],
@@ -1962,7 +1963,7 @@ following Web IDL.
     This structure is used by the client to compute the following quantities:
 
     : <dfn dfn>JSON-serialized client data</dfn>
-    :: This is the [=UTF-8 encoding=] of the result of calling the initial value of {{JSON/stringify|JSON.stringify}} on a
+    :: This is the [=UTF-8 encoding=] of the result of calling [=%JSON%=].{{JSON/stringify|stringify}} on a
         {{CollectedClientData}} dictionary.
 
     : <dfn dfn>Hash of the serialized client data</dfn>


index.bs
```diff
@@ -1962,7 +1963,7 @@ following Web IDL.
     This structure is used by the client to compute the following quantities:
 
     : <dfn dfn>JSON-serialized client data</dfn>
-    :: This is the [=UTF-8 encoding=] of the result of calling the initial value of {{JSON/stringify|JSON.stringify}} on a
+    :: This is the [=UTF-8 encoding=] of the result of calling [=%JSON%=].{{JSON/stringify|stringify}} on a
```
equalsJeffH: WebIDL/browser-guts experts will need to review this. -- "the initial value of" is a magic incantation, i am not sure [=%JSON%=] is equivalent/better.

emlun: I agree this needs expert review. As I understood @annevk's [comment](https://github.com/w3c/webauthn/issues/712#issuecomment-366196187) I think %JSON% should unambiguously identify "the initial value", which is why I dropped that, but I could very well be wrong.

equalsJeffH:
pls review @bzbarsky @jcjones @annevk

index.bs
annevk: @domenic can probably review this. I suspect this is wrong given what we ended up with for the parse situation in Infra.

domenic: This seems like a step backward because now, while you're using the initial value of the JSON object, you're no longer using the initial value for the stringify function. So the new version would be robust to `window.JSON = {}`, but would not be robust to `JSON.stringify = () => {};`.

I think what's there for now is fine-ish. A fully formalized version of this would require adding a %JSONStringify% intrinsic to the JS spec, then writing something like `! Call(%JSONStringify%, undefined, « a CollectedClientData dictionary, converted to a JavaScript object »)`.

(Side note: this spec seems to imply you're specifically calling with no third argument to JSON.stringify, so no spaces will be produced in the output. That's worth testing.)

emlun: Ok. Would "the initial value of %JSON%.stringify" fix that?

I _think_ it ultimately shouldn't matter whether the client produces pretty or compact JSON (or whether different browsers decide differently on that), as the JSON result is returned verbatim to the RP for verification anyway.

domenic: I think it's better to stick with the original text, as it's at least clear that it's not going for full-precision. "Initial value of %JSON%.stringify" is in an uncanny valley.

emlun: Ok, in that case I suggest we close this and #712. Thanks for the reviews!

nadalin:
@emlun Please close

emlun:
Given @domenic's [comment][comment], I believe we can close this and #712 with no further action.

[comment]: https://github.com/w3c/webauthn/pull/813#discussion_r170031615

annevk:
Not sure about no further action, it'd be good to test the spaces bit.

emlun:
I think that would fall on RPs to test that they support JSON both with and without spaces, because all that matters is that the RP can parse it. Different browsers (or even different invocations in the same browser) do not need to produce this JSON in the same format.

domenic:
Er, per the spec, different browsers MUST produce exactly the same format... the one produced by JSON.stringify with no third argument.

emlun:
Sorry, yes, you are correct. What I mean is that if a browser were to violate that and output `clientDataJSON` in a different valid JSON format, it would still be interoperable with compliant authenticators and RPs.

index.bs
equalsJeffH: agreed, thx for reviews :)

PR #821, opened 2018-02-26
UVI modification with CDDL

gmandyam:
In partial fulfillment of https://github.com/w3c/webauthn/issues/679

***
<a href="https://pr-preview.s3.amazonaws.com/gmandyam/webauthn/pull/821.html" title="Last updated on Apr 11, 2018, 8:13 PM GMT (3bae3c0)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/821/6d5aa3e...gmandyam:3bae3c0.html" title="Last updated on Apr 11, 2018, 8:13 PM GMT (3bae3c0)">Diff</a>


gmandyam:
As per discussion in in https://github.com/w3c/webauthn/issues/823, it would be helpful for @equalsJeffH and @rlin1 to clarify whether uviOutput = (bstr), or uviOutput=bstr.size (1..32).

rlin1:
the output size depends on the hash algorithm used.  That is SHA256 today --> 256/8=32 bytes.  But it might change when new hash algorithms are used (crypto agility).
So not sure how strict we should limit the size.

gmandyam:
@rlin1 

We can relax size constraint to 64 bytes to allow for going up to SHA-512, if the desire ts to allow for stronger hashes.  Or we can keep the byte size indeterminate.

However, the current text allows the RP to ignore UVI sizes beyond 32 ("Servers supporting UVI extensions MUST support a length of up to 32 bytes for the UVI value.")  It seems that this requirement should be removed or reworded.

index.bs
```diff
@@ -4005,8 +4005,11 @@ This [=registration extension=] and [=authentication extension=] enables use of
 
     Servers supporting UVI extensions MUST support a length of up to 32 bytes for the UVI value.
 
-    Example for [=authenticator data=] containing one UVI extension
+    Example for [=authenticator data=] containing one UVI extension along with CDDL
```
equalsJeffH: well, I spose I'd write it this way:
Examples, in [=CDDL=] and of the CBOR-encoded output, of [=authenticator data=] containing one UVI extension:

gmandyam: @equalsJeffH 

>question: given the discussion on the webauthn call a couple days ago (see: #679 (comment)), do we still wish to do this?

I think we need to first determine: what the constraints are on UVI size.  The current text seems to imply 32 bytes, but as @rlin1 pointed out, that may restrict stronger hashes.

Based on whatever is decided, then we can determine whether a  CDDL description required for this extension.  If the size is unconstrained, I would say a CDDL description adds no value and the current text related to CBOR usage is sufficient ("The UVI is encoded as CBOR byte string (type 0x58)").  

If the size is constrained, then a CDDL description could add value from a canonical perspective.

gmandyam:
Latest changes:
a) Editorial changes as suggest by @equalsJeffH 
b) Limiting UVI size to 64 bytes
c) Changing server requirements accordingly

nadalin:
@gmandyam to update

rlin1:
LGTM

equalsJeffH:
@gmandyam -- shall I push from my local clone of gmandyam/webauthn to catch it up with upstream/master?

gmandyam:
@equalsJeffH 

Yes, please.  Sorry that my repo was not synched properly.

equalsJeffH:
done, hope it is correct :)

gmandyam:
@nadalin 

Current PR looks OK (thanks, @equalsJeffH).  Can I merge?

equalsJeffH:
LGTM, merge ok.

PR #827, opened 2018-03-02
Emphasise that the CollectedClientData can be extended.

agl:
There is a risk that RPs will implement overly simplistic parsing of the `CollectedClientData` and end up intolerant of any future additions. This change emphasises the need to parse it properly and will be coupled with a behaviour in Chrome that inserts a key (`new_keys_may_be_added_here`) in a random 20% of cases.

***
<a href="https://pr-preview.s3.amazonaws.com/agl/webauthn/pull/827.html" title="Last updated on Mar 13, 2018, 6:40 PM GMT (200f12a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/827/716a169...agl:200f12a.html" title="Last updated on Mar 13, 2018, 6:40 PM GMT (200f12a)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1926,7 +1926,7 @@ follows.
 
 The <dfn>client data</dfn> represents the contextual bindings of both the [=[RP]=] and the client platform. It is a key-value
 mapping with string-valued keys. Values can be any type that has a valid encoding in JSON. Its structure is defined by the
-following Web IDL.
+following Web IDL but it's important to note that it may be extended in the future. Therefore it's critical when parsing to be tolerant of unknown keys and of any reordering of the keys.
 
 <pre class="idl">
     dictionary CollectedClientData {
--- a/index.bs
+++ b/index.bs
@@ -1926,7 +1926,9 @@ follows.
 
 The <dfn>client data</dfn> represents the contextual bindings of both the [=[RP]=] and the client platform. It is a key-value
 mapping with string-valued keys. Values can be any type that has a valid encoding in JSON. Its structure is defined by the
-following Web IDL but it's important to note that it may be extended in the future. Therefore it's critical when parsing to be tolerant of unknown keys and of any reordering of the keys.
+following Web IDL.
+
+Note: The {{CollectedClientData}} may be extended in the future. Therefore it's critical when parsing to be tolerant of unknown keys and of any reordering of the keys.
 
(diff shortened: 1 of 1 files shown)


emlun:
Good idea. Should we add the same note to the other interfaces sent back to the RP (`PublicKeyCredential`, `AuthenticatorAttestationResponse` and `AuthenticatorAssertionResponse`)? Yubico's prototype server has so far gone with rejecting any response structure containing unknown keys anywhere (and yes, Chrome's `new_keys_may_be_added_here` caught that).

index.bs
```diff
@@ -1926,7 +1926,7 @@ follows.
 
 The <dfn>client data</dfn> represents the contextual bindings of both the [=[RP]=] and the client platform. It is a key-value
 mapping with string-valued keys. Values can be any type that has a valid encoding in JSON. Its structure is defined by the
-following Web IDL.
+following Web IDL but it's important to note that it may be extended in the future. Therefore it's critical when parsing to be tolerant of unknown keys and of any reordering of the keys.
```
equalsJeffH: suggest making the new text a "Note:"

IIUC, we may want to add that the [=CTAP2 canonical CBOR encoding form=] stipulates sorted keys thus adding other keys will likely alter ordering one might otherwise be expecting.

emlun: The client data doesn't contain any CBOR data, though. :)

equalsJeffH: doh  0_o

agl:
> Should we add the same note to the other interfaces sent back to the RP (PublicKeyCredential, AuthenticatorAttestationResponse and AuthenticatorAssertionResponse)? 

I don't think that we need to in those cases. Those values are not JSON that needs to be parsed, but Javascript objects. It is possible for Javascript to enumerate the set of attributes in them using the `for (attr in obj)` syntax, but one would have to go out of the way to do it. On the other hand, the JSON must be parsed to be usable and checking it against a template rather than parsing is something that seems quite plausible. Thus I think this only really applies to JSON (and similar cases).

The CBOR structures, however, are candidates. However, they don't seem quite so worrying. (Famous last words.) The extensions should only reflect extensions that the RP has requested and the credential public key is likewise negotiated with the RP via the `pubKeyCredParams`.

PR #829, opened 2018-03-07
Add RP conformance section on ignoring attestation

### LINKED ISSUE #576: RP guidelines should allow RP to not check attestation
A Relying Party who merely wants to use public-key credentials without caring how well the private key is protected could get by without including code to parse the attestations, especially after #557 is fixed. However, as @emlun pointed out there, http://w3c.github.io/webauthn/#registering-a-new-credential currently says the RP MUST validate the attestation statement they receive. We should probably provide an explicit path in that algorithm to avoid parsing the attestation statement.

emlun:
This fixes #576.

***
<a href="https://pr-preview.s3.amazonaws.com/emlun/webauthn/pull/829.html" title="Last updated on May 9, 2018, 11:41 AM GMT (f80ea1a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/829/b3aa419...emlun:f80ea1a.html" title="Last updated on May 9, 2018, 11:41 AM GMT (f80ea1a)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -277,7 +277,7 @@ external hardware, or a combination of both.
 ## [RPS] ## {#conforming-relying-parties}
 
 A [=[RP]=] MUST behave as described in [[#rp-operations]] to obtain all the security benefits offered by this specification. See
-[[#sctn-rp-benefits]] for a detailed discussion of this.
+[[#sctn-rp-benefits]] for further discussion of this.
 
 
 ## All Conformance Classes ## {#conforming-all-classes}
@@ -4758,28 +4758,30 @@ The main benefits offered to [=[RPS]=] by this specification include:
 1. The [=[RP]=] can automatically support multiple types of [=user verification=] - for example PIN, biometrics and/or future
     methods - with little or no code change, and can let each user decide which they prefer to use via their choice of
     [=authenticator=].
-1. The [=[RP]=] does not need to store any secrets or sensitive information in order to gain the above benefits.
+1. The [=[RP]=] does not need to store additional secrets in order to gain the above benefits.
 
 As stated in the [[#conforming-relying-parties|Conformance]] section, the [=[RP]=] MUST behave as described in [[#rp-operations]]
 to obtain all of the above security benefits. However, one notable use case that departs slightly from this is described in the
 next section.
 
 
-### Self Attestation and Ignoring Attestation ### {#sctn-no-attestation-security-attestation}
+### Considerations for Self and None Attestation Types and Ignoring Attestation ### {#sctn-no-attestation-security-attestation}
 
 When [[#registering-a-new-credential|registering a new credential]], the [=[RP]=] MAY choose to accept an [=attestation
-statement=] with [=self attestation=] or [=no attestation statement|no attestation=], or to not verify the [=attestation
-statement=] at all. In all of these cases the [=[RP]=] loses much of benefit (3) listed above, but retains the other benefits.
+statement=] of [[#sctn-attestation-types|type]] [=self attestation|Self=] or [=no attestation statement|None=], or to not verify
+the [=attestation statement=]. In all of these cases the [=[RP]=] loses much of benefit (3) listed above, but retains the other
+benefits.
 
 In these cases it is possible for a [=man-in-the-middle attack|man-in-the-middle attacker=] - for example, a malicious [=client=]
-or script - to replace the [=credential public key=] to be registered, and subsequently tamper with any future [=assertion=]
-[=ceremony=] that passes through the same attacker. [=Authentication=] [=ceremonies=] are still highly resistant to
-[=man-in-the-middle attacks=], but only against attackers that were not present at the time of [=registration=]. Accepting these
-kinds of [=attestation statements=] therefore constitutes a [=leap of faith=]. Note, however, that such an attack would be easy to
-detect and very difficult to maintain, since any [=assertion=] [=ceremony=] that the same attacker does not or cannot tamper with
-would always fail.
+or script - to replace the [=credential public key=] to be registered, and subsequently tamper with future [=authentication
+assertions=] bound for the same [=[RP]=] and passing through the same attacker. Accepting these [[#sctn-attestation-types|types]]
+of [=attestation statements=] therefore constitutes a [=leap of faith=]. In cases where [=registration=] was accomplished
+securely, subsequent [=authentication=] [=ceremonies=] remain resistant to [=man-in-the-middle attacks=], i.e., benefit (3) is
+retained. Note, however, that such an attack would be easy to detect and very difficult to maintain, since any [=authentication=]
+[=ceremony=] that the same attacker does not or cannot tamper with would always fail.
 
-The [=[RP]=] SHOULD consider the above in its threat model when deciding its policy on what [=attestation statements=] to accept.
+The [=[RP]=] SHOULD consider the above in its threat model when deciding its policy on what [[#sctn-attestation-types|attestation
+types]] to accept or whether to ignore [=attestation=].
 
 
 ## credentialId Unsigned ## {#credentialIdSecurity}
--- a/index.bs
+++ b/index.bs
@@ -4751,7 +4751,9 @@ The main benefits offered to [=[RPS]=] by this specification include:
 
 1. Users and accounts can be secured using widely compatible, easy-to-use multi-factor authentication.
 1. The [=[RP]=] does not need to provision [=authenticator=] hardware to its users. Instead, each user can independently obtain
-    any conforming [=authenticator=] and use that same [=authenticator=] with any number of [=[RPS]=].
+    any conforming [=authenticator=] and use that same [=authenticator=] with any number of [=[RPS]=]. The [=[RP]=] can optionally
+    enforce requirements on [=authenticators=]' security properties by inspecting the [=attestation statements=] returned from the
+    [=authenticators=].
 1. [=Registration=] and [=authentication=] [=ceremonies=] are highly resistant to [=man-in-the-middle attacks=].
 1. The [=[RP]=] can automatically support multiple types of [=user verification=] - for example PIN, biometrics and/or future
     methods - with little or no code change, and can let each user decide which they prefer to use via their choice of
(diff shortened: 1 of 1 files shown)


equalsJeffH:
Thanks for writing this up @emlun. This appears to be useful info that we do not already address. Though, it seems to me to be more security considerations material than conformance. 

In looking through it I notice that we probably ought to explicitly address the case of "none" attestation in the [Registering a new credential](https://www.w3.org/TR/2018/WD-webauthn-20180306/#registering-a-new-credential) RP ops subsection, at least in step 14 (since the "none" attStmt format is the only one without a sig to verify).

There's various polishing and terminological fixes I'd offer but will wait on doing so since I'm betting we'll address this issue & PR for PR-milestone and we're presently still trying to attain  CR.  thanks again.

emlun:
Yeah, those thoughts had crossed my mind as well. How much of it do you think should go in security considerations instead: both the list of benefits and the subsection, or just the subsection?

...Actually, I'll go ahead and pre-emptively draft a move of all of it to security considerations, as that seems more appropriate.

And yes, I too imagine this as a PR issue rather than CR.

emlun:
On a related note: with this, the Security Considerations section starts to get a little messy. Perhaps we should split it up into one subsection of concerns for each of the three audiences: authenticator implementers, client implementers and RP implementers.

index.bs
```diff
@@ -4738,6 +4745,41 @@ RogueList when performing ECDAA-Verify (see section 3.6 in [[!FIDOEcdaaAlgorithm
 [[FIDOMetadataService]] provides one way to access such information.
 
 
+## Security Benefits for [RPS] ## {#sctn-rp-benefits}
+
+The main benefits offered to [=[RPS]=] by this specification include:
+
+1. Users and accounts can be secured using widely compatible, easy-to-use multi-factor authentication.
+1. The [=[RP]=] does not need to provision [=authenticator=] hardware to its users. Instead, each user can independently obtain
```
rlin1: Please add benefit of using attestation to understand the authenticator security characteristic.

emlun: Please elaborate what you mean by that.

emlun:
Thanks @equalsJeffH for the review! I'm on vacation this week; I'll address your comments early next week.

index.bs
emlun: Fixed.

index.bs
```diff
@@ -4738,6 +4745,41 @@ RogueList when performing ECDAA-Verify (see section 3.6 in [[!FIDOEcdaaAlgorithm
...
+    any conforming [=authenticator=] and use that same [=authenticator=] with any number of [=[RPS]=].
```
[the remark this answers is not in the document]
emlun: Fixed.

index.bs
```diff
@@ -271,7 +276,9 @@ external hardware, or a combination of both.
 
 ## [RPS] ## {#conforming-relying-parties}
 
-A [=[RP]=] MUST behave as described in [[#rp-operations]] to obtain the security benefits offered by this specification.
+A [=[RP]=] MUST behave as described in [[#rp-operations]] to obtain all the security benefits offered by this specification. See
+[[#sctn-rp-benefits]] for a detailed discussion of this.
```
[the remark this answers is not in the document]
emlun: Fixed.

index.bs
```diff
@@ -4738,6 +4745,41 @@ RogueList when performing ECDAA-Verify (see section 3.6 in [[!FIDOEcdaaAlgorithm
...
+The main benefits offered to [=[RPS]=] by this specification include:
+
+1. Users and accounts can be secured using widely compatible, easy-to-use multi-factor authentication.
+1. The [=[RP]=] does not need to provision [=authenticator=] hardware to its users. Instead, each user can independently obtain
+    any conforming [=authenticator=] and use that same [=authenticator=] with any number of [=[RPS]=].
+1. [=Registration=] and [=authentication=] [=ceremonies=] are highly resistant to [=man-in-the-middle attacks=].
+1. The [=[RP]=] can automatically support multiple types of [=user verification=] - for example PIN, biometrics and/or future
+    methods - with little or no code change, and can let each user decide which they prefer to use via their choice of
+    [=authenticator=].
+1. The [=[RP]=] does not need to store any secrets or sensitive information in order to gain the above benefits.
```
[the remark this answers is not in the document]
emlun: Fixed.

index.bs
```diff
@@ -4738,6 +4745,41 @@ RogueList when performing ECDAA-Verify (see section 3.6 in [[!FIDOEcdaaAlgorithm
...
+
+As stated in the [[#conforming-relying-parties|Conformance]] section, the [=[RP]=] MUST behave as described in [[#rp-operations]]
+to obtain all of the above security benefits. However, one notable use case that departs slightly from this is described in the
+next section.
+
+
+### Self Attestation and Ignoring Attestation ### {#sctn-no-attestation-security-attestation}
+
+When [[#registering-a-new-credential|registering a new credential]], the [=[RP]=] MAY choose to accept an [=attestation
+statement=] with [=self attestation=] or [=no attestation statement|no attestation=], or to not verify the [=attestation
```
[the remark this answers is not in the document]
emlun: Fixed.

index.bs
```diff
@@ -4738,6 +4745,41 @@ RogueList when performing ECDAA-Verify (see section 3.6 in [[!FIDOEcdaaAlgorithm
...
+statement=] at all. In all of these cases the [=[RP]=] loses much of benefit (3) listed above, but retains the other benefits.
+
+In these cases it is possible for a [=man-in-the-middle attack|man-in-the-middle attacker=] - for example, a malicious [=client=]
+or script - to replace the [=credential public key=] to be registered, and subsequently tamper with any future [=assertion=]
```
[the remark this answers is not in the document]
emlun: 1. Fixed.
2. Yes, that's what I'm imagining.

I think for assertion, the attack code would only need to implement signing `authData || clientDataHash` (where `authData` can be a constant per RP) with the private key. For registration, if the RP accepts None attestation, it's enough to generate a fixed-length random credential ID and construct the attestation object as `prefix || credentialId || suffix`, where `prefix` is constant per RP and `suffix` is globally constant. For self attestation it would also be necessary to sign the `authData || clientDataHash` and return `prefix || credentialId || infix || signature || suffix`, where `infix` is also globally constant.

index.bs
```diff
@@ -4738,6 +4745,41 @@ RogueList when performing ECDAA-Verify (see section 3.6 in [[!FIDOEcdaaAlgorithm
...
+[=ceremony=] that passes through the same attacker. [=Authentication=] [=ceremonies=] are still highly resistant to
+[=man-in-the-middle attacks=], but only against attackers that were not present at the time of [=registration=]. Accepting these
```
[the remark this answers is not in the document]
emlun: Fixed.

index.bs
```diff
@@ -4738,6 +4745,41 @@ RogueList when performing ECDAA-Verify (see section 3.6 in [[!FIDOEcdaaAlgorithm
...
+kinds of [=attestation statements=] therefore constitutes a [=leap of faith=]. Note, however, that such an attack would be easy to
+detect and very difficult to maintain, since any [=assertion=] [=ceremony=] that the same attacker does not or cannot tamper with
+would always fail.
+
+The [=[RP]=] SHOULD consider the above in its threat model when deciding its policy on what [=attestation statements=] to accept.
```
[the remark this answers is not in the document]
emlun: 1. Fixed.
2. Good idea, I drafted a note mentioning that.

index.bs
```diff
@@ -4776,6 +4782,49 @@ RogueList when performing ECDAA-Verify (see section 3.6 in [[!FIDOEcdaaAlgorithm
...
+## Security Benefits for [RPS] ## {#sctn-rp-benefits}
+
+The main benefits offered to [=[RPS]=] by this specification include:
+
+1. Users and accounts can be secured using widely compatible, easy-to-use multi-factor authentication.
+1. The [=[RP]=] does not need to provision [=authenticator=] hardware to its users. Instead, each user can independently obtain
+    any conforming [=authenticator=] and use that same [=authenticator=] with any number of [=[RPS]=]. The [=[RP]=] can optionally
+    enforce requirements on [=authenticators=]' security properties by inspecting the [=attestation statements=] returned from the
+    [=authenticators=].
+1. [=Registration=] and [=authentication=] [=ceremonies=] are highly resistant to [=man-in-the-middle attacks=].
```
equalsJeffH: I cant help feeling uneasy about throwing down the "highly resistant" gauntlet...  I'd be inclined to just say "resistant", or maybe "strongly resistant", or "more resistant than username/password"...

index.bs

index.bs
```diff
@@ -4776,6 +4782,49 @@ RogueList when performing ECDAA-Verify (see section 3.6 in [[!FIDOEcdaaAlgorithm
...
+1. The [=[RP]=] can automatically support multiple types of [=user verification=] - for example PIN, biometrics and/or future
+    methods - with little or no code change, and can let each user decide which they prefer to use via their choice of
+    [=authenticator=].
```
equalsJeffH: yeah, we can make the above statement but it leaves out the hard work a deployer has to (or ought to, anyway) do in order to provide for account recovery modulo users' use of a multiplicity of authenticators (some 1st-multi-factor, some 2nd-factor-only) and also train customer support personnel wrt the multiplicity of user verification modalities.  Maybe the latter will end up not being that bad, but I have heard some deployers thinking that for these reasons they would prefer to shepherd all their users to use (primarily) one user verif modality (eg fingerprint).  Again, maybe as the authnr matchers improve and become ubiquitous, and users get used to using them to unlock their devices in the first place, this wont be as much of a concern.... but I wonder if we ought to mention some of these deployment facets in the spec (deploy-cons).

equalsJeffH:
again merging from the moving-target master would be good...  re-reviewed: https://github.com/w3c/webauthn/pull/829#pullrequestreview-117090132

index.bs
emlun: Ok, let's go with just "resistant" then. :)

index.bs
emlun: Good points. Would you prefer we just strike this point off the list?

emlun:
@equalsJeffH I'm happy to entertain your comments, as they make the result better. :) Sorry for the late response, I've had to prioritize other work for the past week.

nadalin:
@emlun Please merge

PR #830, opened 2018-03-08
Add User Public Key as alias of Credential Public Key

### LINKED ISSUE #540: note that "credential public key" is the same as "user's public key"
One of the things that jumped out to me in [Denis Pinkas' webauthn review](https://lists.w3.org/Archives/Public/public-webauthn/2017Jul/0052.html) was his noting that we have two very similar terms in "[public key credential](https://w3c.github.io/webauthn/#public-key-credential)" and "[credential public key](https://w3c.github.io/webauthn/#credential-public-key)" and that there lurks potential confusion for readers other than us'ns. 
	
In both UAF and [U2F](https://fidoalliance.org/specs/fido-u2f-v1.2-ps-20170411/fido-u2f-raw-message-formats-v1.2-ps-20170411.html#registration-response-message-success),  this is referred to as the "user public key".  

I suggest we change "credential public key" to "user public key"  sooner rather than later. 

AFAICT this will not alter the API surface at all (because the user public key is not directly defined in the WebIDL) and is only an editorial change.  Also, we will have to update the [attestation object figure 3](https://w3c.github.io/webauthn/#fig-attStructs).

emlun:
This should fix #540.

Does this need to be mentioned in prose as well?

***
<a href="https://pr-preview.s3.amazonaws.com/emlun/webauthn/pull/830.html" title="Last updated on Mar 28, 2018, 5:21 PM GMT (ececdf7)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/830/716a169...emlun:ececdf7.html" title="Last updated on Mar 28, 2018, 5:21 PM GMT (ececdf7)">Diff</a>


equalsJeffH:
I suppose I'd be inclined to insert a new 2nd sentence of "It is also referred to as the [=user public key=]". Otherwise, LGTM, thx.

index.bs
```diff
@@ -423,12 +423,16 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
...
 
 : <dfn>Credential Public Key</dfn>
+: <dfn>User Public Key</dfn>
 :: The public key portion of a [=[RP]=]-specific <dfn>credential key pair</dfn>, generated by an [=authenticator=] and
     returned to a [=[RP]=] at [=registration=] time (see also [=public key credential=]). The private key portion of the
     [=credential key pair=] is known as the <dfn>credential private key</dfn>. Note that in the case of [=self
     attestation=], the [=credential key pair=] is also used as the [=attestation key pair=], see [=self attestation=]
     for details.
 
+    The [=credential public key=] is referred to as the [=user public key=] in FIDO U2F [[FIDO-U2F-Message-Formats]] and in some
```
equalsJeffH: "user public key" is also used in FIDO UAF [[UAFProtocol]]

Issue #831, opened 2018-03-09
Review of WD08


Kieun:
Since the lack of time for reviewing full documents, I added some suggestions after reading the part of the specs.

1. In Abstract section, there is a high level description of this spec. Since attestation also involves signature generation, we may use _attestation_ in stead of _signature_ from the following sentence.
   > This specification also describes the functional model for WebAuthn conformant authenticators, including their **signature** and attestation functionality.

2. Chapter 2.4 describes CBOR encoding by referring FIDO2 CTAP Spec. The link for the spec is broken. Also chapter 3 has same reference which is broken.
Here's valid link for the spec
https://fidoalliance.org/specs/fido-v2.0-ps-20170927/fido-client-to-authenticator-protocol-v2.0-ps-20170927.html#message-encoding

3. Regarding FIDO AppID definition in Chapter 3, we may add link to FIDO AppID extension (https://www.w3.org/TR/2018/WD-webauthn-20180306/#sctn-appid-extension).

4. In Chapter 4, there is a terminology regarding an Authenticator. Since some authenticator dose not have user verification features, we need to add **user presence test** to following sentence.
   > A cryptographic entity used by a WebAuthn Client to (i) generate a public key credential and register it with a Relying Party, and (ii) authenticate by potentially verifying the user, and then cryptographically signing and returning, in the form of an Authentication Assertion, a challenge and other data presented by a Relying Party (in concert with the WebAuthn Client).

5. In Chapter 4, Client-side-resident Credential Private Key is described. There is a term **roaming authenticator**. We may refer following link.
https://www.w3.org/TR/2018/WD-webauthn-20180306/#roaming-authenticators
Also we may change Authenticator to authenticator in this description.

emlun:
Thanks for the review!

1. Do you suggest rewriting the quoted text to this?

    >This specification also describes the functional model for WebAuthn conformant authenticators, including their attestation and attestation functionality.

2. Good catch, thanks!

    For the record: What is broken is the fragment part `#ctap2-canonical-cbor-encoding-form`, which should be `#message-encoding`. The document URL is fine.

3. Good catch, thanks!

4. I don't think that's strictly necessary, because "**potentially** verifying the user" points out that authenticators do not need to perform user verification.

5. Good catch, thanks!

emlun:
Opened PR #832 which addresses (2), (3) and (5). Leaving (1) and (4) open for discussion.

Kieun:
[...]

PR #832, opened 2018-03-09
Address review comments by @kieun

emlun:
Review comments by @kieun in #831

***
<a href="https://pr-preview.s3.amazonaws.com/emlun/webauthn/pull/832.html" title="Last updated on Mar 12, 2018, 12:11 PM GMT (4d42146)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/832/716a169...emlun:4d42146.html" title="Last updated on Mar 12, 2018, 12:11 PM GMT (4d42146)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -123,7 +123,7 @@ spec: WHATWG HTML; urlPrefix: https://html.spec.whatwg.org/
 
 spec: FIDO-CTAP; urlPrefix: https://fidoalliance.org/specs/fido-v2.0-ps-20170927/fido-client-to-authenticator-protocol-v2.0-ps-20170927.html
     type: dfn
-        text: CTAP2 canonical CBOR encoding form; url: message-encoding
+        text: CTAP2 canonical CBOR encoding form; url: ctap2-canonical-cbor-encoding-form
 
 spec: FIDO-APPID; urlPrefix: https://fidoalliance.org/specs/fido-u2f-v1.2-ps-20170411/fido-appid-and-facets-v1.2-ps-20170411.html
     type: dfn
--- a/index.bs
+++ b/index.bs
@@ -323,7 +323,7 @@ below and in [[#index-defined-elsewhere]].
 : FIDO AppID
 :: The algorithms for [=determining the FacetID of a calling application=] and
     [=determining if a caller's FacetID is authorized for an AppID=] (used only in
-    the <code>[[#sctn-appid-extension|appid]]</code> extension) are defined by [[!FIDO-APPID]].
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -123,7 +123,7 @@ spec: WHATWG HTML; urlPrefix: https://html.spec.whatwg.org/
 
 spec: FIDO-CTAP; urlPrefix: https://fidoalliance.org/specs/fido-v2.0-ps-20170927/fido-client-to-authenticator-protocol-v2.0-ps-20170927.html
     type: dfn
-        text: CTAP2 canonical CBOR encoding form; url: ctap2-canonical-cbor-encoding-form
+        text: CTAP2 canonical CBOR encoding form; url: message-encoding
```
equalsJeffH: the "ctap2-canonical-cbor-encoding-form" URL frag will be correct once the latest CTAP spec is published, whereupon the urlPrefix will be incorrect :-/

the fido-client-to-authenticator-protocol-v2.0-ps-20170927.html#message-encoding section does not correctly define "CTAP2 canonical CBOR encoding form" :(

It seems there is not good solution here until there the revised CTAP spec containing the \<dfn> for "CTAP2 canonical CBOR encoding form" is published.

index.bs
```diff
@@ -323,7 +323,7 @@ below and in [[#index-defined-elsewhere]].
 : FIDO AppID
 :: The algorithms for [=determining the FacetID of a calling application=] and
     [=determining if a caller's FacetID is authorized for an AppID=] (used only in
-    the `appid` extension) are defined by [[!FIDO-APPID]].
+    the <code>[[#sctn-appid-extension|appid]]</code> extension) are defined by [[!FIDO-APPID]].
```
equalsJeffH: I would \<dfn> "AppID" in the title for #sctn-appid-extension, and then just use [=AppID=] here.

index.bs
emlun: Ah, I see. I'll revert this one, then. Good catch!

emlun:
Re: https://github.com/w3c/webauthn/pull/832#discussion_r173599568
Hm. I did this, and it looks a bit weird to me, but if you're happy with the result I can be too. But perhaps the `<dfn>` should be in place of the `|AppID|` in the extension description instead of in the secion heading?

equalsJeffH:
@emlun 
> I did this, and it looks a bit weird to me
well, doing it this way the page ends up with the section title at top of frame -- i think there's at least one other section header we've done this with but I'd have to nose thru the spec source....

PR #811, opened 2018-02-20
Remove clientExtensions and authenticatorExtensions from CollectedClientData

selfissued:
Fixes part of #803 
Fixes #804

***
<a href="https://pr-preview.s3.amazonaws.com/selfissued/webauthn/pull/811.html" title="Last updated on Feb 21, 2018, 10:25 PM GMT (8a3ea00)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/811/05335d4...selfissued:8a3ea00.html" title="Last updated on Feb 21, 2018, 10:25 PM GMT (8a3ea00)">Diff</a>

PR #849, opened 2018-03-21
Fix #848: Weirdness in RP UP verification

### LINKED ISSUE #848: RP UP verification instruction is weird
[§7.1. Registering a new credential][reg] currently reads (and §7.2. Verifying an authentication assertion is analogous):

>10. If user verification is required for this registration, verify that the User Verified bit of the `flags` in _authData_ is set.
>11. If user verification is not required for this registration, verify that the User Present bit of the `flags` in _authData_ is set.

This results in the following truth table:
```
UV req.     | UP req? | UV res. | UP res. | Accept?
------------+---------+---------+---------+--------
discouraged | Yes     | 0       | 0       | No
discouraged | Yes     | 0       | 1       | Yes
discouraged | Yes     | 1       | 0       | No
discouraged | Yes     | 1       | 1       | Yes
preferred   | Yes     | 0       | 0       | No
preferred   | Yes     | 0       | 1       | Yes
preferred   | Yes     | 1       | 0       | No
preferred   | Yes     | 1       | 1       | Yes
required    | No      | 0       | 0       | No
required    | No      | 0       | 1       | No
required    | No      | 1       | 0       | Yes
required    | No      | 1       | 1       | Yes
```

Note, for example, how UV `preferred` means that a response with `(UV=1, UP=0)` should be rejected.

It would make more sense to let the UP requirement be defined by the UV _response_ instead of the UV _requirement_:

```
UV req.     | UV res. | UP req? | UP res. | Accept?
------------+---------+---------+---------+--------
discouraged | 0       | Yes     | 0       | No
discouraged | 0       | Yes     | 1       | Yes
discouraged | 1       | No      | 0       | Yes
discouraged | 1       | No      | 1       | Yes
preferred   | 0       | Yes     | 0       | No
preferred   | 0       | Yes     | 1       | Yes
preferred   | 1       | No      | 0       | Yes
preferred   | 1       | No      | 1       | Yes
required    | 0       | Yes     | 0       | No
required    | 0       | Yes     | 1       | No
required    | 1       | No      | 0       | Yes
required    | 1       | No      | 1       | Yes
```

[reg]: https://www.w3.org/TR/2018/CR-webauthn-20180320/#registering-a-new-credential

emlun:
This fixes #848.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2883,8 +2883,8 @@ When registering a new credential, represented by an {{AuthenticatorAttestationR
 1. If [=user verification=] is required for this registration, verify that the [=User Verified=] bit of the <code>[=flags=]</code>
     in |authData| is set.
 
-1. If [=user verification=] is not required for this registration, verify that the [=User Present=] bit of the
-    <code>[=flags=]</code> in |authData| is set.
+1. If the [=User Verified=] bit of the <code>[=flags=]</code> in |authData| is not set, verify that the [=User Present=] bit of
+    the <code>[=flags=]</code> in |authData| is set.
 
 1. Verify that the values of the [=client extension outputs=] in |clientExtensionResults| and the [=authenticator extension
     outputs=] in the <code>[=authdataextensions|extensions=]</code> in |authData| are as expected, considering the [=client
@@ -2994,8 +2994,8 @@ When verifying a given {{PublicKeyCredential}} structure (|credential|) and an {
 1. If [=user verification=] is required for this assertion, verify that the [=User Verified=] bit of the <code>[=flags=]</code> in
     |aData| is set.
 
-1. If [=user verification=] is not required for this assertion, verify that the [=User Present=] bit of the <code>[=flags=]</code>
-    in |aData| is set.
+1. If the [=User Verified=] bit of the <code>[=flags=]</code> in |authData| is not set, verify that the [=User Present=] bit of
+    the <code>[=flags=]</code> in |authData| is set.
 
 1. Verify that the values of the [=client extension outputs=] in |clientExtensionResults| and the [=authenticator extension
     outputs=] in the <code>[=authdataextensions|extensions=]</code> in |authData| are as expected, considering the [=client


equalsJeffH:
on 2018-03-28 webauthn call @emlun noted that this constitutes more of a "RP policy" thing than a normative change to the protocol.

emlun:
What I mean by that is that it's ultimately up to the RP to decide whether to even care about this - it's not a technical requirement for the protocol to work.

index.bs
```diff
@@ -2883,8 +2883,8 @@ When registering a new credential, represented by an {{AuthenticatorAttestationR
 1. If [=user verification=] is required for this registration, verify that the [=User Verified=] bit of the <code>[=flags=]</code>
     in |authData| is set.
 
-1. If [=user verification=] is not required for this registration, verify that the [=User Present=] bit of the
-    <code>[=flags=]</code> in |authData| is set.
+1. If the [=User Verified=] bit of the <code>[=flags=]</code> in |authData| is not set, verify that the [=User Present=] bit of
```
akshayku: When will user presence bit not set? As of now, silent signatures are not allowed in WebAuthN. I would remove "If the [=User Verified=] bit of the <code>[=flags=]</code> in |authData| is not set" all together.

emlun: The User _Present_ bit will always be set, but the UV bit might not.

emlun: I would like to do that, but the CTAP drafts I'd read at the time didn't make it clear that UP would always be required. It looks like it is in the latest drafts, though. Thanks!

emlun:
This conversation https://github.com/w3c/webauthn/pull/849#discussion_r179211553 also drew my attention to how recent CTAP drafts no longer have the `requireUserPresence` option in the [`authenticatorMakeCredential`][amc] operation, and have clearly specified when to set UP in [`authenticatorGetAssertion`][aga]. Our versions of the operations could use updating to agree with CTAP here, which also simplifies the logic in them. I'll open a separate issue for that.

[amc]: https://w3c.github.io/webauthn/#op-make-cred
[aga]: https://w3c.github.io/webauthn/#op-get-assertion

emlun:
Opened #862.

akshayku:
Let me clarify. As of now, WebAuthN does not support silent signatures in getAssertion  but CTAP does. We decided some time back that we are not allowing silent signatures from the web as of now. It may change in future to allow silent signatures in WebAuthN but that has security implications and really need to be thought through.

emlun:
Thanks, I completely agree.

index.bs
```diff
@@ -2999,12 +2998,11 @@ When verifying a given {{PublicKeyCredential}} structure (|credential|) and an {
 
 1. Verify that the <code>[=rpIdHash=]</code> in |aData| is the SHA-256 hash of the [=RP ID=] expected by the [=[RP]=].
 
+1. Verify that the [=User Present=] bit of the <code>[=flags=]</code> in |authData| is set.
```
equalsJeffH: I think the proposed changes per se are good, but here in [7.2. Verifying an authentication assertion](https://w3c.github.io/webauthn/#verifying-assertion) we seem to have an issue with using both |aData| and |authData| interchangeably, yes? 

I _think_ all instances of |aData| and |authData| in this section ought to be changed to be either one or the other. If I am not correct, then I'm thinking we ought to have a Note explaining the difference between them. 

Otherwise this seems fine.

emlun: Thanks, good catch. This looks like two unrelated copy-and paste errors (this PR and #811). I agree we'd be better off using the same variable name in both algorithms, I suggest we choose `|authData|`.

equalsJeffH: using `|authData|` is fine by me.

emlun: I'll change to `|aData|` in this PR and submit a separate PR to change from `|aData|` and `|adata|` to `|authData|`.

emlun: Opened issue #879.

equalsJeffH: OK, thanks. I note the new PR to "change from `|aData|` and `|adata|` to `|authData|`" has not yet appeared but we now have issue #879 for that, thus this plan sounds fine so approving this PR (and plan). thx!

PR #850, opened 2018-03-22
Add note about verifying U2F assertion signatures

### LINKED ISSUE #845: Assertion signature verification procedure defined in Section 7.2 doesn't fit for the signature generated by FIDO-U2F token.
WebAuthn signature generation procedure is not equal to that of FIDO-U2F. [1]
Assertion signature verification procedure defined in Section 7.2 doesn't fit for the signature generated by FIDO-U2F token.
The spec should denote that the assertion generated by FIDO-U2F token should be verified in the manner defined in the FIDO-U2F spec.

[1]
The web authentication spec defines that the signature is generated
over the concatenation of the authenticator data and the hash of the
serialized client data
(See https://w3c.github.io/webauthn/#op-get-assertion), but
FIDO U2F defines that the signature is generated
over the following source:
- The application parameter [32 bytes] from the authentication request message.
- The user presence byte [1 byte].
- The counter [4 bytes].
- The challenge parameter [32 bytes] from the authentication request message.
(See https://fidoalliance.org/specs/fido-u2f-v1.1-id-20160915/fido-u2f-raw-message-formats-v1.1-id-20160915.html#authentication-response-message-success)

emlun:
This fixes #845.


emlun:
Added discussion point to OP:

>Should we move this to a small subsection of §6.1 Authenticator Data and refer the note to there?

index.bs
```diff
@@ -3016,6 +3016,10 @@ When verifying a given {{PublicKeyCredential}} structure (|credential|) and an {
 1. Using the credential public key looked up in step 3, verify that |sig| is a valid signature over the binary concatenation of
     |aData| and |hash|.
 
+    Note: This verification step is automatically compatible with signatures generated by FIDO U2F authenticators, because the
```
equalsJeffH: re (from https://github.com/w3c/webauthn/pull/850#issue-176761510):
> Should we move this to a small subsection of §6.1 Authenticator Data and refer the note to there?

that is fine by me.  So I spect this note would become something like:

`Note: This verification step is automatically compatible with signatures generated by FIDO U2F authenticators, see [[#sctn-fido-u2f-sig-format-compat]]`

emlun: Done, please re-review!

nadalin:
@emlun Please fix and merge

PR #858, opened 2018-04-04
Add recommendation of challenge length

### LINKED ISSUE #85: assertionChallenge recommendations
Should we include some recommendations on the size / format of the assertion challenge? I had arbitrarily selected 256 random bytes in my server implementation, but that seems like a burden for NFC / BLE authenticators -- something that I hadn't thought about until now.

I just figure that a hint or two might prevent people who are unfamiliar with the system as a whole from creating unintentional problems.

emlun:
This fixes #85.


selfissued:
@equalsJeffH to merge if it looks good to him.

PR #859, opened 2018-04-04
Allow client to refuse too short challenges

emlun:
As suggested in https://github.com/w3c/webauthn/issues/85#issuecomment-372309459 . This would merge into #858.

This would allow clients to refuse challenges that are clearly suspicious in their soundness, although it wouldn't help enforce any requirements on the actual randomness of the contents.

This is potentially breaking, depending on perspective. Clients will not need to change, but they may if they want to. RPs will need to change if at least one client does.

I will not object to closing this if it doesn't seem like a good idea.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -758,6 +758,9 @@ When this method is invoked, the user agent MUST execute the following algorithm
 
 1. Let |options| be the value of <code>|options|.{{CredentialCreationOptions/publicKey}}</code>.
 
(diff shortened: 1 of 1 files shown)


jcjones:
Mmm... I can't think of many reasons why anyone would ever want to decrease the length of a challenge. It does seem like it might be short-sighted to declare 16 good, 15 bad in a long-lived spec. However for interop, I'd hate if Firefox required 32 bytes but Chrome was okay with only 4.... This is not straightforward.

emlun:
Is that two votes for closing, then?

PR #878, opened 2018-04-23
Fix #593 - Refer to RFC 8266 for RP-controlled UI strings

### LINKED ISSUE #593: Display name content rules?
[This review was done against https://www.w3.org/TR/2016/WD-webauthn-20160531/#iface-account, and the comment was raised by Addison Phillips. The i18n WG agreed to forward as a WG comment.]

There are no rules (other than that they are DOMStrings) associated with display names. Should there be any content restrictions?

jcjones:
The RP provides 'PublicKeyCredentialUserEntity/displayName' and 'PublicKeyCredentialEntity/name',
both of which are intended for display by User Agent. As DOMString objects, these could be
manipulated by a malicious RP to try and confuse the user about what is being displayed, so
User Agents should be careful in how they display these fields.

This PR points to RFC 8266 for its guidance on showing those fields. This is guidance that
browser vendors already follow for other specifications, so it's nothing new -- it merely
codifies what should be.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1656,6 +1656,10 @@ associated.
             For example, a [=[RP]=] might choose to map [=human palatability|human-palatable=] [=username=] account identifiers to
             the {{PublicKeyCredentialEntity/name}} member of {{PublicKeyCredentialUserEntity}}.
 
+        Wherever [=Conforming User Agents=] display the {{PublicKeyCredentialEntity/name}}, they
+        SHOULD abide by the [[RFC8266]] Nickname Profile to protect users against potentially confusing
+        strings.
+
         [=Authenticators=] MUST accept and store a 64-byte minimum length for a {{PublicKeyCredentialEntity/name}} member's
         value. Authenticators MAY truncate a {{PublicKeyCredentialEntity/name}} member's value to a length equal to or greater
         than 64 bytes.
@@ -1703,6 +1707,10 @@ credential.
     :   <dfn>displayName</dfn>
     ::  A human-friendly name for the user account, intended only for display. For example, "Alex P. Müller" or "田中 倫". The
         [=[RP]=] SHOULD let the user choose this, and SHOULD NOT restrict the choice more than necessary.
+
+        Wherever [=Conforming User Agents=] display the {{PublicKeyCredentialUserEntity/displayName}}, they
+        SHOULD abide by the [[RFC8266]] Nickname Profile to protect users against potentially confusing
+        strings.
 
         [=Authenticators=] MUST accept and store a 64-byte minimum length for a {{PublicKeyCredentialUserEntity/displayName}}
         member's value. Authenticators MAY truncate a {{PublicKeyCredentialUserEntity/displayName}} member's value to a length


equalsJeffH:
I agree that the spec needs to refer to the PRECIS spec suite [1] regarding its use and treatment of internationalized (unicode-encoded) strings. I am thinking ([IIUC](https://en.wiktionary.org/wiki/IIUC)) the proposed two single-line additions in this PR are a start, but to properly utilize PRECIS (and user-visible internat'd strings) we ought to specify things more thoroughly.  

I.e., in RFC8266 [S 4 Use in Application Protocols](https://tools.ietf.org/html/rfc8266#section-4), there's these directions:
> It is the responsibility of an application
   protocol (e.g., MSRP, XCON, or XMPP) or application definition to
   specify **(1)** the protocol slots in which nickname strings can appear, **(2) the
   entities that are expected to enforce the rules governing nickname
   strings,** and **(3) the point during protocol processing or interface
   handling when the rules need to be enforced**.
   [numbers and emphasis added]

This PR presently seems to be a start on (1), but (2) and (3) do not seem to be presently address? 

E.g., as a suggestion we should perhaps specify in [5.4.1. Public Key Entity Description](https://w3c.github.io/webauthn/#dictionary-pkcredentialentity) that the RP is responsible for [the preparation and the enforcement](https://tools.ietf.org/html/rfc8264#section-3) of [PublicKeyCredentialEntity.name](https://w3c.github.io/webauthn/#dictdef-publickeycredentialentity)'s value, via the PRECIS [nickname profile](https://tools.ietf.org/html/rfc8266#section-2)'s [preparation](https://tools.ietf.org/html/rfc8266#section-2.2) and [enforcement](https://tools.ietf.org/html/rfc8266#section-2.3) rules. 

Similarly in [5.4.3. User Account Parameters for Credential Generation](https://w3c.github.io/webauthn/#sctn-rp-credential-params) specify the same for  [PublicKeyCredentialUserEntity.displayName](https://w3c.github.io/webauthn/#dictdef-publickeycredentialuserentity)'s value.  

Then [IIUC](https://en.wiktionary.org/wiki/IIUC) we need to find the places in the spec where these strings are compared and indicate that PRECIS [nickname profile comparison rules](https://tools.ietf.org/html/rfc8266#section-2.4) are enforced. 

WDYT?

cc @stpeter

-----
[1] PRECIS Framework: Preparation, Enforcement, and Comparison of Internationalized Strings in Application Protocols, RFCs [8264](https://tools.ietf.org/html/rfc8264), [8265](https://tools.ietf.org/html/rfc8265), [8266](https://tools.ietf.org/html/rfc8266).

emlun:
>[...] as a suggestion we should perhaps specify in [5.4.1. Public Key Entity Description](https://w3c.github.io/webauthn/#dictionary-pkcredentialentity) that the RP is responsible for [the preparation and the enforcement](https://tools.ietf.org/html/rfc8264#section-3) of [PublicKeyCredentialEntity.name](https://w3c.github.io/webauthn/#dictdef-publickeycredentialentity)'s value, [...]
>
>Similarly in [5.4.3. User Account Parameters for Credential Generation](https://w3c.github.io/webauthn/#sctn-rp-credential-params) specify the same for  [PublicKeyCredentialUserEntity.displayName](https://w3c.github.io/webauthn/#dictdef-publickeycredentialuserentity)'s value.

Shouldn't that be the client's responsibility? As @jcjones points out, this is primarily meant to defend against malicious RPs, so it seems odd to put this responsibility on the RP.

akshayku:
This recommendation is also applicable for RPs when they pass this information?

jcjones:
Being not sure what to write for the RPs, I'm going to tug on @stpeter's sleeve for advice.

stpeter:
Hey all, sorry I missed the initial ping from @equalsJeffH. I'm happy to help out here. It would be great if someone could add the `i18n-tracking` label to this thread in accordance with [i18n WG practice](https://www.w3.org/International/review-request#labels). I will provide a longer reply soon.

stpeter:
First, let's try to understand who the actors are, why they might want to use internationalized strings, and what could possibly go wrong with such a usage (which in i18n is usually "a lot").

The intent of the "nickname" profile in RFC 8266 is to provide a way for people to create memorable or expressive handles for their own stuff (a nickname in a chatroom, a personal tag for a device, etc.). The nickname profile allows a nearly limitless range of Unicode characters and therefore should be employed with extreme caution. For instance, it should not be employed when handles will be entered in multiple places by multiple actors, when handles might be the basis for authentication and authorization decisions, etc. Please consider very carefully whether we want a relying party or a user to generate a PublicKeyCredentialEntity name or a PublicKeyCredentialUserEntity displayName containing characters like ∞ (INFINITY, U+221E) or ♚  (BLACK CHESS KING, U+265A). I suspect that we we do not want this, and that RPs and users do not need this. 

What RPs and users do need is a way to associate human-friendly names (not necessarily expressive handles) with themselves as RPs or users, where the names can include Unicode characters outside the ASCII range and where "human-friendly" means friendly to humans whose native language cannot be represented in the Latin script.

Thus I would strongly suggest that the "nickname" profile of the PRECIS FreeformClass from RFC 8266 is not the right construct here, and that we actually want to create an application-layer construct consisting of a space-separated set of userparts that themselves conform to the [UsernameCaseMapped](https://tools.ietf.org/html/rfc8265#section-3.3) profile of the PRECIS IdentifierClass from RFC 8265. This is a much safer approach because IMHO we do not need the expressiveness that RFC 8266 allows. See [Section 12.3 of RFC 8264](https://tools.ietf.org/html/rfc8264#section-12.3) for relevant warnings.

Second, assuming that we want to do what I suggest above (or something even simpler like a [character string](https://www.w3.org/TR/charmod/#def-character-string) if folks like @r12a and @aphillips can provide suggestions), we need to look at the mechanics of the spec text. I'll pause for breath here so we can discuss the higher-order bit and hopefully come to agreement.

stpeter:
We discussed this issue on the i18n WG call this morning and concluded that we would like to understand the context and attack surface more fully. In particular, what prevents an RP from (say) [asserting it is Stripe](https://www.comodoca.com/en-us/about/blog/on-comodo-ca%e2%80%99s-recent-revocation-of-an-ssl-certifi/) and does an attacker really need to go through all the headaches of an internationalized spoofing attack if something much simpler is possible?

aphillips:
@emlun But I think our point is that it doesn't just have to "look like" `GitHub`. There's nothing to prevent it actually *being* the string `GitHub`. There is no Unicode magic involved here. Choosing a baroque or bizarre string would just make your `name` hard to read--but have no other deleterious side effects.

emlun:
Huh... you're absolutely right. That hadn't even occurred to me. :smile:

stpeter:
Taking a cue from RFC 8266 (Section 6.1), we might want to say something like this:

OLD

> Its value’s name member is required.
>
> Its value’s id member specifies the relying party identifier with which the credential should be associated. If omitted, its value will be the CredentialsContainer object’s relevant settings object's origin's effective domain.

NEW

> Its value's name member, which is required, is a user-friendly handle for the RelyingParty.
>
> Its value’s id member specifies the relying party identifier with which the credential should be associated. If omitted, its value will be the CredentialsContainer object’s relevant settings object's origin's effective domain.
>
> To ensure secure operation, authentication and authorization decisions MUST be made on the basis of the id member, not on the name member.

Issue #867, opened 2018-04-12
What is the point of `allowCredentials`?


subyraman:
My understanding, reading the spec, was that `allowCredentials` provides a means for the RP to filter registered authenticators, eg request the browser to "wink" a single selected authenticator (like a Yubikey). As an RP, this would be useful (essential?) for providing a good UX, so that the user is not confused by several authentication notifications, only one of which is valid.

I think there are differences of opinion on this as noted on this discussion in the [Chromium bug board; ](https://bugs.chromium.org/p/chromium/issues/detail?id=828567) there is a stance that that `allowedCredentials` is a list used to fail authentications only after a ceremony takes place (if I understand it correctly).

If that is the case, I don't understand why it is useful at all; the RP themselves can reject the authentication on our end given that we receive the `credentialId `.

Can we get some clarification on this, and perhaps emphasize it in the spec if it isn't there?

herrjemand:
@subyraman allowCredentials is used to provide authenticator with exact credential you want assertion from, and it is essential as FIDO2 supports 2FA-only mode, as it requires exact credID to be presented. Same applies for U2F.

emlun:
@subyraman Your understanding of the spec is correct, in addition to what @herrjemand points out.

What the Chrome team is pointing out is that they want to allow the RP to detect if the user attempts to authenticate with an authenticator that does not have one of the allowed credentials, so the RP can inform the user that they need to use a different authenticator. In order to do that, they need the user to first confirm the attempt to authenticate - otherwise there's no way to know that the user won't plug in a different authenticator a few seconds into the future. I'm guessing this could be solved better by instad showing a browser popup when the user plugs in an authenticator without any of the `allowCredentials`, rather than returning an error to the RP. I added a comment on that to the Chrome thread too.

This issue stands in opposition to #863.

equalsJeffH:
shall we map this issue to the same milestone as issue #863?

emlun:
Maybe, but I feel like this deserves some discussion on a WG call now that we've seen confusion between RPs and browsers in CR implementations. I think #863 might need to be moved to PR since it seems to be something Chrome is already testing implementation of. @equalsJeffH WDYT?

emlun:
@subyraman has clearly made a different interpretation of the spec than what's currently implemented in Chrome.

kpaulh:
[...]

Issue #863, opened 2018-04-06
Authenticators that do not recognize any handles shouldn't just be dropped on the floor


kpaulh:
Context from the sister issue raised in the CTAP2 spec: [fido-2-specs/issues/509](https://github.com/fido-alliance/fido-2-specs/issues/509):
"When a privacy-conscious client collects assertions, it does not reveal immediately to an RP that no authenticator recognized any of the identifiers in the allow list. Doing so would allow malicious RPs to test for the presence of authenticators using only a credential id.

In Chrome, both in U2F and WebAuthn, we handle this by making plugged in USB authenticators blink until the user selects one of them (and thereby gives consent to reveal its presence). At that point we return an error indicating that no key handle was recognized."

We had been doing this in the cryptoken extension as well. However, the WebAuthN spec, in 6.2.3.2, states that if no credentials are present on that authenticator, 
"return an error code equivalent to "NotAllowedError" and terminate the operation."
(Effectively, drop the device on the floor and waiting for other hotplugged devices.)

To be consistent with CTAP2, we think we should send the authenticator one of the keyhandles so that it blinks, and if the user selects that authenticator by giving it a tap, then return "InvalidStateError" to indicate the user tried to use a device that wasn't actually registered.

(Similar to how the spec handles user consent for already-registered authenticators in 6.2.2.3)

emlun:
This issue stands in opposition to #867.

equalsJeffH:
@emlun wrote: 
> If the browser has this knowledge, why not present that to the user directly?

I wonder whether the browser folk are reticent to directly provide UX of this sort because of the desires for RPs to have fine-grained control over such UX (as has been argued as the reason the UA-provided HTTP authentication dialogs are hardly ever employed) ?

emlun:
>Hm, also, if not the UA handling the UX, the RP's client-side JS could handle all this authnr-selection interaction, without necessarily incurring the latency of comms with the server, yes?

Yes, but what I'm more worried about is friction in the user interaction - having to start the ceremony all over after it seems like you're done, instead of being nudged to plug in a different authenticator.

kpaulh:
Sorry for going AWOL, wanted to double-check some facts with folks first before making a bunch of assertions :-)

Jeff largely has it right with regards to not wanting to do all the UX in the browser. It's also difficult to build functioning RPs if we don't respond to certain calls. 

[...]

PR #882, opened 2018-04-24
Include an AuthenticatorTransport when creating a new credential.

agl:
FIDO U2F found it neccessary to wedge the authenticator transport in an
X.509 extension of the attestation certificate in order to communicate
this information to the RP.

In Webauthn, we currently note that it's possible that an RP might learn
this information from the attestation, but now have several kinds of
attesattion and it seems dumb to define ways to wedge this information
in each.

Instead, have the client include the transport in the
AuthenticatorAttestationResponse. Also, define another transport type
for cases where a non-standard protocol is used to communicate with a
platform authenticator.

Since interface attributes cannot be optional, this technically defines a
breaking change, although I don't believe that it will break anything in
practice. Still, this could also be punted to an extension if we wished.
However, given that U2F found it to be central, I've gone this route.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1021,6 +1021,9 @@ When this method is invoked, the user agent MUST execute the following algorithm
                         :   {{AuthenticatorAttestationResponse/attestationObject}}
                         ::  |attestationObject|
 
+                        :   {{AuthenticatorAttestationResponse/transport}}
+                        ::  A value of type {{AuthenticatorTransport}} specifying how |authenticator| was contacted.
+
                     :   {{PublicKeyCredential/[[clientExtensionsResults]]}}
                     ::  A new {{ArrayBuffer}}, created using |global|'s [=%ArrayBuffer%=], containing the bytes of
                         <code>|credentialCreationData|.[=credentialCreationData/clientExtensionResults=]</code>.
@@ -1490,7 +1493,8 @@ during registration.
 <pre class="idl">
     [SecureContext, Exposed=Window]
     interface AuthenticatorAttestationResponse : AuthenticatorResponse {
-        [SameObject] readonly attribute ArrayBuffer      attestationObject;
+        [SameObject] readonly attribute ArrayBuffer            attestationObject;
+        [SameObject] readonly attribute AuthenticatorTransport transport;
     };
 </pre>
 <div dfn-type="attribute" dfn-for="AuthenticatorAttestationResponse">
@@ -1509,6 +1513,9 @@ during registration.
         [=attestation statement=], as well as to decode and validate the [=authenticator data=] along with the
         [=JSON-serialized client data=]. For more details, see [[#sctn-attestation]], [[#generating-an-attestation-object]],
         and [Figure 3](#fig-attStructs).
+
+    :   <dfn>transport</dfn>
+    ::  This attribute indicates which [transport](#transport) the client used to communicate with the authenticator.
 </div>
 
 ### Web Authentication Assertion (interface <dfn interface>AuthenticatorAssertionResponse</dfn>) ### {#iface-authenticatorassertionresponse}
@@ -2024,21 +2031,30 @@ parameter to the {{CredentialsContainer/create()}} or {{CredentialsContainer/get
     enum AuthenticatorTransport {
         "usb",
         "nfc",
-        "ble"
+        "ble",
+        "internal"
     };
 </pre>
 
 <div dfn-type="enum-value" dfn-for="AuthenticatorTransport">
-    Authenticators may communicate with clients using a variety of transports. This enumeration defines a hint as to how clients
-    might communicate with a particular authenticator in order to obtain an assertion for a specific credential. Note that these
-    hints represent the [=[RP]=]'s best belief as to how an authenticator may be reached. A [=[RP]=] may obtain a list of
-    transports hints from some [=attestation statement formats=] or via some out-of-band mechanism; it is outside the scope of
-    this specification to define that mechanism.
+    Authenticators may communicate with clients using a variety of transports. This enumeration defines a hint as to how
+    clients might communicate with a particular authenticator. It is used in two contexts: from client to RP when
+    [creating a new credential](#createCredential), and from RP to client when when [requesting an
+    assertion](#getAssertion).
+
+    In the first context, the RP might store this information in order to present appropriate instructions to the user
+    in the future. For example, instructing the user to press an NFC authenticator to the reader. (A [=[RP]=] may also
+    obtain a list of transports hints from some [=attestation statement formats=] or via some other out-of-band
+    mechanism; it is outside the scope of this specification to define such mechanisms.)
+
+    In the second context, this is a hint representing the [=[RP]=]'s best belief as to how an authenticator may be
+    reached.
 
     <ul>
         <li><dfn>usb</dfn> - the respective authenticator can be contacted over USB.
         <li><dfn>nfc</dfn> - the respective authenticator can be contacted over Near Field Communication (NFC).
         <li><dfn>ble</dfn> - the respective authenticator can be contacted over Bluetooth Smart (Bluetooth Low Energy / BLE).
+        <li><dfn>internal</dfn> - the respective authenticator is contacted using a platform-specific transport. These authenticators are usually not removable from the platform.
     </ul>
 </div>
 


equalsJeffH:
Is there an associated issue for this?

equalsJeffH:
see issues #863 and #867

equalsJeffH:
@gmandyam notes on 2018-Apr-25 call that there is an extension that nominally addresses this: https://www.w3.org/TR/webauthn/#authenticatorSelection

agl:
(Sorry, missed the call due to the access code not working.)

[AuthenticatorSelectionCriteria](https://www.w3.org/TR/webauthn/#authenticatorSelection) may be used [when creating a credential](https://www.w3.org/TR/webauthn/#dictionary-makecredentialoptions) to control the attachment (i.e. platform or roaming). But this PR is about the RP learning (not controlling) the transport (i.e. USB / BLE / NFC) after creating a credential.

agl:
(Also, I've specified this as a single value, but Alexei reminds me that it should be a sequence of them. I'll amend that in the coming days.

christiaanbrand:
I just want to echo the importance of this PR to implementing real-world use cases. Since it *technically* doesn't break anything, I wonder if we can't slip it into the current v1 spec...

christiaanbrand:
@agl, let me know if you disagree with my addition of the word "removable" (I made another commit on your PR).

equalsJeffH:
@christiaanbrand wrote:
> let me know if you disagree with my addition of the word "removable"

hm, I think I understand what you're trying to differentiate there, but I would not do it that way. we have two orthogonal authenticator characteristics: [attachment modality](https://w3c.github.io/webauthn/#attachment), and [transport](https://w3c.github.io/webauthn/#transport).  

attachment modality of `cross-platform` matches your attempted use of the term "removable" I am guessing. 

AFAIK, a [platform authnr](https://w3c.github.io/webauthn/#platform-authenticators) (i.e., non-removable) might actually employ one of the transports, e.g. USB, rather than something else that the proposed "internal" platform-specific value is intended to address (i suppose a [platform authnr](https://w3c.github.io/webauthn/#platform-authenticators) could use _any_ of the transports, but having a one using BLE or NFC offhand seems silly).

I think the Note in section [5.4.5 Authenticator Attachment enumeration](https://w3c.github.io/webauthn/#attachment) already largely addresses what you are attempting to hint at with the use of "removable".

equalsJeffH:
@agl wrote: 
> this PR is about the RP learning (not controlling) the transport (i.e. USB / BLE / NFC) after creating a credential.

ACK, thx for clarification :)

index.bs
```diff
@@ -1021,6 +1021,9 @@ When this method is invoked, the user agent MUST execute the following algorithm
                         :   {{AuthenticatorAttestationResponse/attestationObject}}
                         ::  |attestationObject|
 
+                        :   {{AuthenticatorAttestationResponse/transport}}
+                        ::  A value of type {{AuthenticatorTransport}} specifying how |authenticator| was contacted.
```
s/ specifying / indicating /

index.bs
```diff
@@ -1509,6 +1513,9 @@ during registration.
         [=attestation statement=], as well as to decode and validate the [=authenticator data=] along with the
         [=JSON-serialized client data=]. For more details, see [[#sctn-attestation]], [[#generating-an-attestation-object]],
         and [Figure 3](#fig-attStructs).
+
+    :   <dfn>transport</dfn>
+    ::  This attribute indicates which [transport](#transport) the client used to communicate with the authenticator.
```
equalsJeffH: suggest using [=client=] and [=authenticator=] here.

index.bs
```diff
@@ -2024,21 +2031,30 @@ parameter to the {{CredentialsContainer/create()}} or {{CredentialsContainer/get
...
 <div dfn-type="enum-value" dfn-for="AuthenticatorTransport">
-    Authenticators may communicate with clients using a variety of transports. This enumeration defines a hint as to how clients
-    might communicate with a particular authenticator in order to obtain an assertion for a specific credential. Note that these
-    hints represent the [=[RP]=]'s best belief as to how an authenticator may be reached. A [=[RP]=] may obtain a list of
-    transports hints from some [=attestation statement formats=] or via some out-of-band mechanism; it is outside the scope of
-    this specification to define that mechanism.
+    Authenticators may communicate with clients using a variety of transports. This enumeration defines a hint as to how
+    clients might communicate with a particular authenticator. It is used in two contexts: from client to RP when
+    [creating a new credential](#createCredential), and from RP to client when when [requesting an
+    assertion](#getAssertion).
```
equalsJeffH: Suggest a modest re-write such that terms are aligned with other portions of the spec:
[=Authenticators=] may implement various [[#transport|transports]] for communicating with [=clients=]. This enumeration defines hints as to how clients might communicate with a particular authenticator. It is used during both [[#createCredential|credential creation]] and [[#getAssertion|assertion generation]].

index.bs
```diff
@@ -2024,21 +2031,30 @@ parameter to the {{CredentialsContainer/create()}} or {{CredentialsContainer/get
...
+
+    In the first context, the RP might store this information in order to present appropriate instructions to the user
+    in the future. For example, instructing the user to press an NFC authenticator to the reader. (A [=[RP]=] may also
+    obtain a list of transports hints from some [=attestation statement formats=] or via some other out-of-band
+    mechanism; it is outside the scope of this specification to define such mechanisms.)
+
+    In the second context, this is a hint representing the [=[RP]=]'s best belief as to how an authenticator may be
+    reached.
```
equalsJeffH: Suggest a modest re-write of the abovce such that terms are aligned with other portions of the spec:
[=Authenticators=] may implement various [[#transport|transports]] for communicating with [=clients=]. This enumeration defines hints as to how clients might communicate with a particular authenticator. It is used during both [[#createCredential|credential creation]] and [[#getAssertion|assertion generation]].

During [[#createCredential|credential creation]], the RP can store the hint in order to present appropriate instructions to the user in the future. For example, instructing the user to press an NFC authenticator to the reader. 

Note: A [=[RP]=] may also obtain a list of transports hints from some [=attestation statement formats=] or via some other out-of-band mechanism (e.g., [[FIDOMetadataService]]; it is outside the scope of this specification to define such mechanisms.

Later, during [[#getAssertion|assertion generation]], the [=[RP]=] can supply the hint to the [=[client]=] to facilitate contacting the [=authenticator=]. The [=[RP]=] can also use the hint to appropriately tailor the user experience.

index.bs
```diff
@@ -2024,21 +2031,30 @@ parameter to the {{CredentialsContainer/create()}} or {{CredentialsContainer/get
...
 
     <ul>
-        <li><dfn>usb</dfn> - the respective authenticator can be contacted over USB.
-        <li><dfn>nfc</dfn> - the respective authenticator can be contacted over Near Field Communication (NFC).
-        <li><dfn>ble</dfn> - the respective authenticator can be contacted over Bluetooth Smart (Bluetooth Low Energy / BLE).
+        <li><dfn>usb</dfn> - the respective authenticator can be contacted over removable USB.
+        <li><dfn>nfc</dfn> - the respective authenticator can be contacted over removable Near Field Communication (NFC).
+        <li><dfn>ble</dfn> - the respective authenticator can be contacted over removable Bluetooth Smart (Bluetooth Low Energy / BLE).
+        <li><dfn>internal</dfn> - the respective authenticator is contacted using a platform-specific transport. These authenticators are usually not removable from the platform.
```
equalsJeffH: per https://github.com/w3c/webauthn/pull/882#issuecomment-385553226, suggest removing the three instances of "removable" in the above. 

Perhaps we should add a cross-ref to [authnr attachment](https://github.com/w3c/webauthn/pull/882#issuecomment-385553226) clarify that attachement modality and employed transports are (essentially?) orthogonal?

index.bs
emlun: s/During [[#createCredential|credential creation]]/After completing [[#createCredential|credential creation]]/
?

index.bs

index.bs

index.bs

index.bs

equalsJeffH:
fyi, merging from master ought to fix the breaking build

agl:
Have included the suggestions above, save for deleting &ldquo;removable&rdquo; from the `usb` transport.

There are two things going on here:

Firstly, we want RPs to be able to show UI based on the transport and know about the transport generally. This is (I believe) the reason that FIDO wedged this information in the attestation certificate for U2F. As part of this, we need an `internal` value to handle (real world) cases like &ldquo;it's attached via an I²C bus to the application processor&rdquo;&mdash;things that we can't handle today.

Secondly, we want RPs to echo this information back down to the client so that we don't, for example, prompt the user to insert an external device when the RP knows that it's a platform authenticator. We are conflating the transport and the attachment here and another design for this would be to plumb an attachment down when getting an assertion. The reason that we're not is that we suspect that many will not get subtle points, like setting the attachment mode, correct, but they're more likely to be able to copy a value from the registration into each request.

(Maybe we could plumb an explicit attachment up in the registration too for them to copy? I should let Christiaan take this point since he speaks to a lot of potential RPs.)

Anyway, for this second motivation, we want to make `internal` mean that it's a platform authenticator.

equalsJeffH:
@agl thanks for the [rationale](https://github.com/w3c/webauthn/pull/882#issuecomment-385743344) 
wrt the new "internal" [AuthenticatorTransport](https://w3c.github.io/webauthn/#enumdef-authenticatortransport) value -- the explanatory language is:
> the respective authenticator is contacted using a platform-specific transport. These authenticators are not removable from the platform.

...meaning that the transport in this case _could_ be, e.g., USB, yes?

If so, perhaps we ought to provide guidance in sections [#attachment](https://w3c.github.io/webauthn/#attachment ) and [#transport](https://w3c.github.io/webauthn/#transport) that if a [platform authnr](https://w3c.github.io/webauthn/#platform-authenticators) is employed, the returned [AuthenticatorTransport](https://w3c.github.io/webauthn/#enumdef-authenticatortransport) value should be "internal" ?  And perhaps also in the [authenticatorMakeCredential](https://w3c.github.io/webauthn/#op-make-cred) operation?

I also wonder if some of the [rationale](https://github.com/w3c/webauthn/pull/882#issuecomment-385743344)  details wrt RP's use of this info ought to be further explained?

akshayku:
We have authenticators which support multiple transports. Relying on user to always use the same transport may not be the best choice. Directing the experience on what was used in Make Credential is not something that can be relied upon. For example, lets say user provisions the key on NFC reader on one machine. Then it roams around to a different machine where there is no NFC reader and it breaks down. But user wants/can use USB. 

Relying on transport information during MakeCredential and making decisions based on that is not a good design. Also, I would like RP's not to be burdened with this new caveat. Rp will not understand this and can make wrong choices.

akshayku:
For the second thing about prompting user when RP knows that it only created platform authenticator in first place is one of the gaps that we missed and boils down to not having attachment property in getAssertion. This becomes very relevant when user moves to a brand new machine and tries to login or the same machine gets reset. 

We have to decide how much important are these issues and whether spec needs to be changed at this time of the game. I think we have two options, either break the spec and add attachment property in getAssertion similar to MakeCredential OR go via an extension route. 

I would not put "internal" enum in transports to fix the real issue.

akshayku:
Also, this transports design to solve second issue does not help in username less flows.

agl:
@akshayku 

> We have authenticators which support multiple transports.

As noted [above](https://github.com/w3c/webauthn/pull/882#issuecomment-385107649), this was intended to be a sequence but I've been slow in getting around to updating it. I'm hoping to do so before today's call.

> Relying on transport information during MakeCredential and making decisions based on that is not a good design.

This does not introduce any decisions based on the transport during MakeCredential. It's reporting the information that's currently in the FIDO attestation statement in a less weird place.

> I would not put "internal" enum in transports to fix the real issue.

We need a value for this irrespective because the current enumeration is insufficient. (See [here](https://github.com/w3c/webauthn/pull/882#issuecomment-385743344).)

> and boils down to not having attachment property in getAssertion.

This is a valid point of discussion. Christiaan believes that it'll be easier for RPs to copy the transports from the credential into the assertion response and he talks to a lot of them. I believe that reporting an [`AuthenticatorAttachment`](https://www.w3.org/TR/webauthn/#enumdef-authenticatorattachment) in addition to the transport and plumbing that into [`PublicKeyCredentialDescriptor`](https://www.w3.org/TR/webauthn/#dictdef-publickeycredentialdescriptor) is also a plausible answer, but I'll let Christiaan argue between them.

@equalsJeffH 

> &hellip;meaning that the transport in this case could be, e.g., USB, yes?

Depending on how the second desire pans out, a platform authenticator that happens to be attached via a USB bus could be recorded as either `usb` or `internal` (if we decide to have `AuthenticatorAttachment` hold the attachment), or must be `internal` (if we decide to use the transport to indicate attachment here).

Although, even in the former case, if this information is intended to guide RPs in their messaging and UI, an internal USB bus isn't &ldquo;USB&rdquo; as far as any user goes, so I'd still say that `internal` would be a more accurate answer.

> If so, perhaps we ought to provide guidance&hellip;

Yea, this change is too skinny at the moment, but I'll see how the discussion about using `internal` to indicate a platform authenticator pans out first. Thanks.

akshayku:
But authenticators don't tell what transports it support in authenticatorGetInfo. So platform has no idea how to pass that information to the RP.

agl:
> But authenticators don't tell what transports it support in authenticatorGetInfo. So platform has no idea how to pass that information to the RP.

For now we can extract it from the attestation certificate, since it's wedged in there for RPs already. Since one of the goals is to guide the interface of the RP, if the client has no information, its best guess would be to include the transport that it actually used on the basis that the user will probably use that same mechanism again.

christiaanbrand:
I'm happy to present how this proposal fits into the larger plans in Amsterdam. I don't believe what I have in mind is satisfied by simply stating the attachment preference during Sign.

equalsJeffH:
issue #334's use cases are perhaps relevant to this discussion...

emlun:
I'm also curious about what UX would be expected if the RP passes an `allowCredentials` with one credential with `platform` attachment and one with `cross-platform` attachment...

agl:
> I'm also curious about what UX would be expected if the RP passes an allowCredentials with one credential with platform attachment and one with cross-platform attachment&hellip;

In this case, since the request could be satisfied by a cross-platform authenticator, I think the client would want to prompt the user to insert one. Either way, if the client has the information, it can hopefully do something smarter than when it doesn't.

christiaanbrand:
There is no such thing as passing "platform" and "cross platform" during signature. It will be passed as part of the transport as "internal" and "caBLE" for example. I guess this is what you mean?

If that's the case, the platform will by default, try to solve the request with the built-in authenticator first (since that allows for the best UX), if there's no credential there, it will then show the dialog to plug in an external authenticator.
If there *is* an internal credential, it's up to the platform to decide if it wants to allow the user to cancel out of using that one, and rather insert an external token, although I struggle to really see the use case there.

However, I think I now also understand what Akshay was alluding to on the call today:
What if we send down a Sign request to the user-agent with an empty list of credentialIds. It's still conceivable that the RP would want to express some preference over whether to allow this request to be satisfied by a local authenticator vs a roaming authenticator. I think this is a corner case though (if I already have a platform authenticator the RP should *usually* be able to have a hint to it from a cookie or something), and I would expect an RP who supports the "typeless flow" to be equally happy with a built-in (chances of getting a hit there is slim) OR an external authenticator (this is mostly where the credential will be during bootstrapping, so maybe this use case isn't that important.

nadalin:
@christiaanbrand The typical use case is that the platform key represents an identity to the RP and the external key represents another identity to RP

equalsJeffH:
this PR could (please) use a merge-from-master to enable easier review...

emlun:
Re: https://github.com/w3c/webauthn/pull/882#issuecomment-386441090

>Might some vendors of "webauthn/FIDO2 authenticators" weigh-in regarding whether the latter authnrs' attestation certs do or do not include the [FIDO U2F Authenticator Transports Extension](https://fidoalliance.org/specs/fido-u2f-v1.2-ps-20170411/fido-u2f-authenticator-transports-extension-v1.2-ps-20170411.html) ?

Yubico's FIDO2 authenticators do include this extension in their attestation certs.

akshayku:
@emlun 
>I'm also curious about what UX would be expected if the RP passes an allowCredentials with one credential with platform attachment and one with cross-platform attachment...

We show option to use external authenticator when we cannot find all the credentials in our local platform. I can give a demo at plenary. 

@christiaanbrand 
Username less flows are very important to us. That's the magical experience we want to achieve in web also as we are doing in our windows login. Now consider a scenario where RPs has expressed interest in platform authenticators for whatever reason by selecting platform during MakeCredential. All their credentials are platform ones. User has created multiple platform credentials on different machines. Now if it goes to a new machine, and tries to do username less flow, nothing will match and we are currently falling back to external authenticators which RP never intended. This is one issue. Another issue is somehow keystore gets reset and even if RP has cookies, we are falling back to external devices. We cannot rely on cookies. User clears cookies all the time.

Regarding what to show when there is a match in platform authenticator and there is an ambiguity, we provide an option to the user to select external device also if it chooses to. We have streamlined the experience to directly use the platform authenticator with an option to go to external authenticator as part of that dialog box. This may be the scenario where a user has created platform authenticator for convenience and external authenticator for backup or roaming or something. 

Regarding getting transports from authenticator certificate, not every vendor is putting this information in the attestation certificate. If we have to support multiple transports for a given authenticator, as of now the best place to get this information IMO is authenticatorGetInfo. I also want to understand transports better at plenary.

Lets discuss this in more detail and go through the different user types journey.

emlun:
>[...] If we have to support multiple transports for a given authenticator, as of now the best place to get this information IMO is authenticatorGetInfo [...]

`authenticatorGetInfo` doesn't exist in WebAuthn, though, only in CTAP, so the RP cannot use it.

akshayku:
Platform will add this information to the credential for the RP. I was talking about how to get this information reliably which works 100% of the time by all authenticators.

equalsJeffH:
@akshayku stated:
> Platform will add this information [results of getInfo(), which includes transport hints] to the credential for the RP.

Add "this information" to "the credential"? how, where?  

Does that hint at an alternative proposal to this pull request (#882) ?  If so, perhaps y'all ought to codify it in a pull request prior to the plenary?

> I was talking about how to get this information reliably which works 100% of the time by all authenticators.

I do not understand how you are going to concoct something that "works 100% of the time by all authenticators" when not all authnrs will be connected via CTAP2, e.g., CTAP1/U2F authnrs which AFAICT do not have the equivalent of `authenticatorGetInfo()` (rather, they apparently have (only) a [GetVersion operation](https://fidoalliance.org/specs/fido-u2f-v1.2-ps-20170411/fido-u2f-raw-message-formats-v1.2-ps-20170411.html#getversion-request-and-response---u2f_version)) . Splain?

akshayku:
@equalsJeffH I am thinking as I am reading and trying to understand the use case of this PR. I am not clear on that and I don't have a alternative right now. What I meant from "this information" was "transports supported by authenticator" and "to the credential" meaning that this PR does to add "transports" to credential descriptor. 

I see your point of U2F authenticators but not all in market U2F authenticators also don't put this information in certificates. What I meant by 100% was for newer FIDO2 V2 authenticator devices and I would like an easy and reliable way to figure out this information going forward instead of looking somewhere in certificate which is optional and may not even come back depending upon attestation type.

selfissued:
Per the 9-May-18 call, as this PR currently does multiple things, it was agreed to split it into separate PRs that can be considered independently.  In particular, adding a value to an enumeration is not a breaking change whereas adding a value to a data structure is.

index.bs
```diff
@@ -1509,6 +1513,10 @@ during registration.
         [=attestation statement=], as well as to decode and validate the [=authenticator data=] along with the
         [=JSON-serialized client data=]. For more details, see [[#sctn-attestation]], [[#generating-an-attestation-object]],
         and [Figure 3](#fig-attStructs).
+
+    :   <dfn>transports</dfn>
+    ::  This attribute indicates which [transports](#transport) the [=client=] believes can be used to communicate with
```
jcjones: If the UA gets its belief wrong, what happens? 

For example: A UA using a Feiten U2F token that supports BLE and USB might know that it used USB, but may not know for sure that it has other transports available. It practically mandates decoding `TRANSPORT_EXTENSION_OID`, I think, and maybe should reference it.

index.bs
```diff
@@ -1509,6 +1513,10 @@ during registration.
...
+        this [=authenticator=].
```
jcjones: How does this interact with anonymization, like the `none` attestation? If the user somehow requests maximum privacy, is it acceptable for `transports` to be the empty set? (I'm thinking Tor Browser here, not mainline Firefox.)

agl:
#903 has been split off from this to just add the &ldquo;internal&rdquo; transport type, as requested.

PR #884, opened 2018-04-25
Add example of new device enrollment via roaming authenticator

### LINKED ISSUE #151: Clarify how a user can authenticate from multiple devices
The current draft specification makes no attempts to address the question of authentication by a user from different devices.

Consider the following common use cases:
1. A user wishes to access their account from multiple devices that they trust; e.g., a desktop computer, a laptop computer, and a mobile device.
2. A user wishes to access their account from an untrusted device; e.g., a friend's laptop.

The specification doesn't include recommendations to implementers on how to address these common scenarios. Once a private key has been created, how does a user authorize other devices to gain access to the account?

Passwords are portable.  As long as one recalls the password, an account can be accessed from any device.  If Web Authentication has the goal to replace passwords, then it must address the issue of portability.

emlun:
This fixes #151.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/pull/884.html" title="Last updated on Jun 13, 2018, 4:42 PM GMT (b609a29)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/884/204da23...b609a29.html" title="Last updated on Jun 13, 2018, 4:42 PM GMT (b609a29)">Diff</a>


equalsJeffH:
wrt the failing build here, here's the error:
```
FATAL ERROR: Unknown inline tag 'non-' found:
  non-`null`
ERR Did not generate, due to fatal errors
```
..which one also gets when building the current master branch. In looking at how the WebIDL spec is formatted, it uses the term "non-null" as regular text without any special formatting, so perhaps we ought to just do the same thing and s/ non-`null` / non-null / ?

Once that is done and merged-to-master then other branches that are failing on the same bug can be fixed by merging-from-master...

index.bs
```diff
@@ -231,6 +231,43 @@ scenarios. Additional scenarios, including sample code, are given later in [[#sa
...
 
+### New device enrollment ### {#usecase-new-device-enrollment}
+
+This use case illustrates how a [=roaming authenticator=] - e.g., a USB security key fob - can be used for initial sign-in on a
+new device, and a [=platform authenticator=] - typically built into a laptop or smartphone - can be employed for easy day-to-day
+use.
+
+- First, on a desktop computer:
+    * User navigates to example.com in a browser and signs in to an existing account using whatever method they have been using
+        (possibly a legacy method such as a password), or creates a new account.
```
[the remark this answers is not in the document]
emlun: Are you sure? I think that parenthesis is closed earlier in the sentence.

index.bs
```diff
@@ -231,6 +231,43 @@ scenarios. Additional scenarios, including sample code, are given later in [[#sa
     * Web page shows that the selected user is signed in, and navigates to the signed-in page.
 
 
+### New device enrollment ### {#usecase-new-device-enrollment}
+
+This use case illustrates how a [=roaming authenticator=] - e.g., a USB security key fob - can be used for initial sign-in on a
+new device, and a [=platform authenticator=] - typically built into a laptop or smartphone - can be employed for easy day-to-day
+use.
```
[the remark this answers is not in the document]
emlun: Added paragraph:

>Note that the various prompts and user-entered texts in this section are examples, and not normative.

index.bs
```diff
@@ -231,6 +231,43 @@ scenarios. Additional scenarios, including sample code, are given later in [[#sa
...
+
+- First, on a desktop computer:
+    * User navigates to example.com in a browser and signs in to an existing account using whatever method they have been using
```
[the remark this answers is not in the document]

index.bs
```diff
@@ -231,6 +231,43 @@ scenarios. Additional scenarios, including sample code, are given later in [[#sa
...
+    * The USB key blinks to indicate the user should press the button on it.
+    * User presses the button on the USB key.
+    * Website shows message, "Registration complete."
+
+- Later, on their laptop:
+    * User navigates to example.com in a browser and initiates signing in.
+    * Website prompts the user to plug in the "Keychain key" USB key.
+    * User plugs in the previously registered USB key and presses the button.
+    * Website shows that the user is signed in, and navigates to the signed-in page.
+    * Website prompts, "Do you want to register this device with example.com?"
```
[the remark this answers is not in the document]
emlun: Good point, thanks! I changed "device" to "computer" - let me know if you think it's still ambiguous.

index.bs
```diff
@@ -231,6 +231,43 @@ scenarios. Additional scenarios, including sample code, are given later in [[#sa
...
+    * User agrees.
+    * Laptop prompts the user for a previously configured [=authorization gesture=] (PIN, biometric, etc.); the user
+        provides this.
+    * Website prompts the user for a device nickname; user enters "Work laptop".
+    * Website shows message, "Registration complete."
+    * User signs out.
+
+- Later, again on their laptop:
+    * User navigates to example.com in a browser and initiates signing in.
+    * Website shows message, "If this device is registered, please follow the instructions in your browser. If not, please plug in
```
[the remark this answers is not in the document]
emlun: Most likely, yeah, but since cookies and such may be cleared from time to time it's not entirely reliable. But I agree that assuming the common case here makes the use case example more straightforward.

index.bs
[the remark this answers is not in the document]
emlun: Good idea, thanks!

index.bs
```diff
@@ -231,6 +231,43 @@ scenarios. Additional scenarios, including sample code, are given later in [[#sa
     * Web page shows that the selected user is signed in, and navigates to the signed-in page.
 
 
+### New device enrollment ### {#usecase-new-device-enrollment}
+
+This use case illustrates how a [=roaming authenticator=] - e.g., a USB security key fob - can be used for initial sign-in on a
+new device, and a [=platform authenticator=] - typically built into a laptop or smartphone - can be employed for easy day-to-day
+use.
+
+- First, on a desktop computer:
```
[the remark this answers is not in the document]
emlun: Sure!

index.bs
```diff
@@ -231,6 +231,43 @@ scenarios. Additional scenarios, including sample code, are given later in [[#sa
...
+    * User navigates to example.com in a browser and signs in to an existing account using whatever method they have been using
+        (possibly a legacy method such as a password), or creates a new account.
+    * User navigates to account security settings and selects "Register security key".
+    * Website prompts the user to plug in a USB security key fob and enter a nickname for the key.
+    * User enters "Keychain key" and plugs in the USB key.
+    * The USB key blinks to indicate the user should press the button on it.
+    * User presses the button on the USB key.
+    * Website shows message, "Registration complete."
```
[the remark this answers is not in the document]
emlun: Thanks! I replaced [=[RP]=] with "website" to make it blend better with the rest of the section.

index.bs
```diff
@@ -231,6 +231,43 @@ scenarios. Additional scenarios, including sample code, are given later in [[#sa
...
+
+- Later, on their laptop:
```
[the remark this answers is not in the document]

index.bs
```diff
@@ -231,6 +231,43 @@ scenarios. Additional scenarios, including sample code, are given later in [[#sa
...
+    * User navigates to example.com in a browser and initiates signing in.
+    * Website prompts the user to plug in the "Keychain key" USB key.
```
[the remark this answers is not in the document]

index.bs
```diff
@@ -231,6 +231,43 @@ scenarios. Additional scenarios, including sample code, are given later in [[#sa
...
+    * User plugs in the previously registered USB key and presses the button.
```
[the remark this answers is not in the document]

emlun:
@equalsJeffH Thanks for the review! I agree on the "getting into this level of detail" issue of what other uses cases to include descriptions of. Perhaps most of #334 can be resolved together with #422?

index.bs
```diff
@@ -243,6 +243,50 @@ scenarios. Additional scenarios, including sample code, are given later in [[#sa
...
+    devices that lack a [=platform authenticator=], and
+  - a low-friction means to strongly re-authenticate on devices having [=platform authenticators=].
+
+Note: This approach of registering multiple [=authenticators=] for an account is also useful in account recovery use cases.
+
+Note: The various prompts and user-entered texts in this section are examples, and not normative.
+
+- First, on a desktop computer (lacking a [=platform authenticator=]):
+    * User navigates to `example.com` in a browser and signs in to an existing account using whatever method they have been using
+        (possibly a legacy method such as a password), or creates a new account.
```
akshayku: One more flow:

During new account creation, RP can choose to directly associate a FIDO key with the account without the need to associate an password/other methods with the account. This way, this account can only be unlocked with these FIDO keys and RP's can get the benefit of not having passwords in the first place.

index.bs
```diff
@@ -243,6 +243,50 @@ scenarios. Additional scenarios, including sample code, are given later in [[#sa
...
+    * Website prompts, "Do you want to register this computer with example.com?"
+    * User agrees.
+    * Laptop prompts the user for a previously enrolled [=authorization gesture=] (PIN, biometric, etc.); the user provides this.
+    * Website prompts the user for a device nickname; user enters "Work laptop".
+    * Website shows message, "Registration complete."
+    * User signs out.
+
+- Later, again on their laptop:
+    * User navigates to example.com in a browser and initiates signing in.
+    * Website shows message, "Please follow the instructions in your browser to sign in."
```
akshayku: In this username less flow, even if platform finds a key for that RP, Platform needs to provide a way to support external FIDO devices in the UI as platform does not know whether user has multiple keys associated with that RP and wants to use external keys for whatever reason.

akshayku: In with username flow, where platform gets credential ID allow list, if all the credential ID passed are found on the authenticator, platform can skip provide an option for the user to user external devices. If not all the credentialID are found on the platform, but some are found, again, platform needs to provide an option to the user to use external keys.

index.bs
emlun: I agree this makes sense to point out, but I think putting it into this use case might make it too many things all at once. How about mentioning that in [§1.1.4. Other use cases and configurations][other]?

[other]: https://pr-preview.s3.amazonaws.com/w3c/webauthn/pull/884.html#other-configurations

index.bs
emlun: Yes, this is true, but I don't think the example needs to exhaustively detail all the possibilities. I think it'll be clearer to just illustrate one of the possible ways it could play out. What do you think?

index.bs
emlun: @akshayku Bump

emlun:
@equalsJeffH Please re-review!

index.bs
```diff
@@ -243,6 +243,50 @@ scenarios. Additional scenarios, including sample code, are given later in [[#sa
     * Web page shows that the selected user is signed in, and navigates to the signed-in page.
 
 
+### New device enrollment ### {#usecase-new-device-enrollment}
+
+This use case scenario illustrates how a [=[RP]=] can leverage a combination of a [=roaming authenticator=] (e.g., a USB security
+key fob) and a [=platform authenticator=] (e.g., a built-in fingerprint sensor) such that the user has:
+
+  - a "primary" [=authenticator=] that they use to authenticate on new-to-them devices (e.g., laptops, desktops) or on such
```
s/ "primary" [=authenticator=] / "primary" [=roaming authenticator=] /

index.bs
```diff
@@ -243,6 +243,50 @@ scenarios. Additional scenarios, including sample code, are given later in [[#sa
...
+
+Note: This approach of registering multiple [=authenticators=] for an account is also useful in account recovery use cases.
+
+Note: The various prompts and user-entered texts in this section are examples, and not normative.
+
+- First, on a desktop computer (lacking a [=platform authenticator=]):
+    * User navigates to `example.com` in a browser and signs in to an existing account using whatever method they have been using
+        (possibly a legacy method such as a password), or creates a new account.
+    * User navigates to account security settings and selects "Register security key".
+    * Website prompts the user to plug in a USB security key fob and enter a nickname for the key.
```
equalsJeffH: the "entering a nickname for the key" stuff is not part of the protocol and is webapp-deployment-specific, IIUC?  I'm inclined to not include it, and only say:
> * Website prompts the user to plug in a USB security key.

index.bs
```diff
@@ -243,6 +243,50 @@ scenarios. Additional scenarios, including sample code, are given later in [[#sa
...
+    * User enters "Keychain key" and plugs in the USB security key.
```
equalsJeffH: ... and say only "User plugs in their USB security key." here.  another related comment is below.

index.bs
```diff
@@ -243,6 +243,50 @@ scenarios. Additional scenarios, including sample code, are given later in [[#sa
...
+    * The USB security key blinks to indicate the user should press the button on it; the user does.
+    * Website shows message, "Registration complete."
+
+    Note: Since this computer lacks a [=platform authenticator=], the website may require the user to present their USB security
+    key from time to time or each time the user interacts with the website. This is at the website's discretion.
+
+- Later, on their laptop (which features a [=platform authenticator=]):
+    * User navigates to example.com in a browser and initiates signing in.
+    * Website prompts the user to plug in the "Keychain key" USB security key.
```
equalsJeffH: I would say just "Website prompts the user to plug in their USB security key."

we can add a Note below to the effect that allowing the user to "name" which security key they are using is a perhaps useful webapp feature.....tho webapps will do this differently or not at all, and the security key name is specific to the webapp, so I'm wondering if we want to get into this here in this section in this spec....

index.bs
```diff
@@ -243,6 +243,50 @@ scenarios. Additional scenarios, including sample code, are given later in [[#sa
     * Web page shows that the selected user is signed in, and navigates to the signed-in page.
 
 
+### New device enrollment ### {#usecase-new-device-enrollment}
```
equalsJeffH: I'm thinking we ought to use the term "registration" rather than enrollment here because "enrollment" is what one does for their biometric authz gestures.

index.bs
```diff
@@ -243,6 +243,50 @@ scenarios. Additional scenarios, including sample code, are given later in [[#sa
...
+    key from time to time or each time the user interacts with the website. This is at the website's discretion.
+
+- Later, on their laptop (which features a [=platform authenticator=]):
+    * User navigates to example.com in a browser and initiates signing in.
+    * Website prompts the user to plug in the "Keychain key" USB security key.
+    * User plugs in the previously registered USB security key and presses the button.
+    * Website shows that the user is signed in, and navigates to the signed-in page.
+    * Website prompts, "Do you want to register this computer with example.com?"
+    * User agrees.
+    * Laptop prompts the user for a previously enrolled [=authorization gesture=] (PIN, biometric, etc.); the user provides this.
```
equalsJeffH: perhaps we ought to explain/define "enrollment" wrt authorization gestures, eg in  a Note or in the terminology section?

index.bs
```diff
@@ -243,6 +243,50 @@ scenarios. Additional scenarios, including sample code, are given later in [[#sa
...
+    * Website prompts the user for a device nickname; user enters "Work laptop".
```
equalsJeffH: as above, I dunno if we ought to include the above step.

index.bs
equalsJeffH: re "Please follow the instructions in your browser to sign in.", the next step is "Laptop prompts the user for an authorization gesture...", so perhaps the former ought to be:
> Please follow your computer's prompts to complete sign in

..?

equalsJeffH: perhaps add a Note indicating that there are variations to the above if a user has registered a mix of platform and roaming authnrs with a given webapp?

index.bs
```diff
@@ -243,6 +243,50 @@ scenarios. Additional scenarios, including sample code, are given later in [[#sa
...
+This use case scenario illustrates how a [=[RP]=] can leverage a combination of a [=roaming authenticator=] (e.g., a USB security
+key fob) and a [=platform authenticator=] (e.g., a built-in fingerprint sensor) such that the user has:
+
+  - a "primary" [=authenticator=] that they use to authenticate on new-to-them devices (e.g., laptops, desktops) or on such
+    devices that lack a [=platform authenticator=], and
+  - a low-friction means to strongly re-authenticate on devices having [=platform authenticators=].
+
+Note: This approach of registering multiple [=authenticators=] for an account is also useful in account recovery use cases.
+
+Note: The various prompts and user-entered texts in this section are examples, and not normative.
```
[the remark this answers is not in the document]
emlun: Good point - I'll delete the second Note:, then.

index.bs
emlun: Nice!

index.bs
emlun: Yeah, good point.

index.bs
emlun: Yeah, good points.

index.bs
emlun: Ok!

index.bs
emlun: Yeah, looks good.

index.bs
emlun: Yep.

index.bs
emlun: I included the nickname thing as a way to make it clear which credential is used at which time, but in retrospect I guess it's clear enough without it.

index.bs
emlun: Or perhaps change "enrolled" to "configured"?

index.bs
emlun: Deleted.

index.bs
equalsJeffH: offhand "configured" is fine by me.

equalsJeffH:
merge?

Issue #462, opened 2017-05-15
undefined terms and terms we really ought to define


equalsJeffH:
The below terms are formally undefined and we should consider defining them (and linking their occurrences to their dfn.   Be sure to see also issue #358 -- there is overlap between this issue and that one. 

Add to, or remove from, this list by updating this original post (OP):
-------------------------------------------------------------------

- [ ] attesting authority (aka authenticator vendor (which could be a (client) platform vendor))

- [ ] attestation trust model (presently we discuss "trust model" in terms of attestation types, but do not define the latter term)

- [x] AAGUID

- [x] assertion

- [ ] authenticator-related terms:
```
          authenticator characteristics     // are discussed in #sctn-authenticator-taxonomy
          authenticator session
   cloned authenticator
          authenticator protection measures
```
- [x] external authenticator  (to be defined in conjunction with roaming authnr)

- [ ] CREDENTIAL:
```
(a credential)   bound to a authenticator
(a credential is)   bound to   an/this authenticator
                    managed by    "         "
                    controlled by    "         "
                    present on    "         "
                    stored on     "         "
                    owning             authenticator

                     credential ID
                     credential object 
Client-side-resident Public Key Credential Source  //  is presently defined 
                                                   // synonymously with 
                                                   // 'resident credential'
server-side resident credential       // presently undefined 
```
Note: residentKey is currently used in WebIDL (so we're likely stuck with it) and as a variable name in algorithms. It is synonymous with Client-side-resident Public Key Credential Source / resident credential.

- [ ] credential metadata -- everything in the [Public Key Credential Source](https://w3c.github.io/webauthn/#public-key-credential-source) other than the credential private key.
```
CLIENT:

client-side   // see also issue #80

```
- [ ] cross-platform transport protocols

- [ ] extension data

- [ ] first-factor
  - [x] as in "first-factor authenticator" aka one that is actually multi-factor because it is user verification-capable (1st factor, something you are), and wields the private key (2nd factor, a secret you possess).
  - [ ] also may want to clarify/define/use terms such as "multi-factor authn", "first multi-factor", etc.

- [x] identifier of the credential
```
supported by this implementation
```
- [ ] LDH Labels (perhaps just make that single-occurrance term a link to https://tools.ietf.org/html/rfc5890#section-2.3.1)

- [ ] local configuration knowledge
```
PLATFORM:

       Android "N" or later platform
                    Android platforms

currently available on this platform
   supported by this client   "
          user agent and/or   "
[...]

PR #899, opened 2018-05-09
Document prevention of attacks on privacy

### LINKED ISSUE #382: Describe attacks on privacy that are allowed/prevented
Several issues and proposed changes are motivated by privacy concerns, but I don't see a list of what private information we want to expose vs keep hidden. For example, attestation implies that we want to expose the brand of authenticator the user owns, and parts of #379 imply that we want to be able to hide that the user's hardware supports authentication until they consent.

### LINKED ISSUE #743: document preventions of RP-driven de-anonymization attempts in privacy-cons

emlun:
Fixes #743. Fixes #382.

#140 is related, but requires a normative requirement that authenticators make credential IDs look like opaque random data.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/pull/899.html" title="Last updated on Jun 20, 2018, 5:06 PM GMT (48d6579)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/899/b3aa419...48d6579.html" title="Last updated on Jun 20, 2018, 5:06 PM GMT (48d6579)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -4794,8 +4794,7 @@ use or provide any form of global identity, the following kinds of potentially c
 - The user's [=credential IDs=] and [=credential public keys=].
 
     These are registered by the [=[RP]=] and subsequently used by the user to prove possession of the corresponding [=credential
-    private key=], and thus ownership of the [=public key credential=] identified by the [=credential ID=]. They are also visible
-    to the [=client=] in the communication with the [=authenticator=].
+    private key=]. They are also visible to the [=client=] in the communication with the [=authenticator=].
 
 - The user's identities specific to each [=[RP]=], e.g., usernames and [=user handles=].
 
@@ -4856,9 +4855,10 @@ without a traditional username, further improving non-correlatability between [=
 
 [=Biometric authenticators=] perform the [=biometric recognition=] internally in the [=authenticator=] - though for [=platform
 authenticators=] the biometric data might also be visible to the [=client=], depending on the implementation. Biometric data is
-not revealed to the [=[RP]=]; it is only used as an additional layer of security for unlocking use of the [=authenticator=]. A
-malicious [=[RP]=] therefore cannot discover the user's personal identity via biometric data, and a security breach at a [=[RP]=]
-cannot expose biometric data for an attacker to use for forging logins at other [=[RPS]=].
+not revealed to the [=[RP]=]; it is used only locally to perform [=user verification=] authorizing the creation and
+[=registration=] of, or [=authentication=] using, a [=public key credential=]. A malicious [=[RP]=] therefore cannot discover the
+user's personal identity via biometric data, and a security breach at a [=[RP]=] cannot expose biometric data for an attacker to
+use for forging logins at other [=[RPS]=].
 
 In the case where a [=[RP]=] requires [=biometric recognition=], this is achieved by the [=biometric authenticator=] setting the
 [=UV=] [=flag=] in the signed [=assertion=] response, instead of revealing the biometric data itself to the [=[RP]=]. The [=[RP]=]
--- a/index.bs
+++ b/index.bs
@@ -4839,6 +4839,10 @@ shared between [=[RPS]=].
     not correlatable as having the same [=managing authenticator=]. A pair of malicious [=[RPS]=] thus cannot correlate users
     between their systems without additional information, e.g., a willfully reused username or e-mail address.
 
+- [=Authenticators=] ensure that their [=attestation certificates=] are not unique enough to identify a single [=authenticator=]
+    or a small group of [=authenticators=]. This is detailed further in [[#sec-attestation-privacy]]. A pair of malicious
+    [=[RPS]=] thus cannot correlate users between their systems by tracking individual [=authenticators=].
+
 Additionally, a [=public key credential=] with a [=client-side-resident credential private key=] can optionally include a [=user
 handle=] specified by the [=[RP]=]. The [=public key credential|credential=] can then be used to both identify and
 [=authentication|authenticate=] the user. This means that a privacy-conscious [=[RP]=] can allow the user to create an account
--- a/index.bs
+++ b/index.bs
@@ -147,10 +147,6 @@ spec: FIDO-Registry; urlPrefix: https://fidoalliance.org/specs/fido-v2.0-ps-2017
     type: dfn
         text: Section 3.6.2; url: public-key-representation-formats
 
-spec: Wikipedia-NaturalPerson; urlPrefix: https://en.wikipedia.org/wiki/Natural_person
-    type: dfn
-        text: natural person
-
 
 </pre> <!-- class=anchors -->
 
@@ -4787,9 +4783,9 @@ The privacy principles in [[!FIDO-Privacy-Principles]] also apply to this specif
 [INFORMATIVE]
 
 Many aspects of the design of the [=Web Authentication API=] are motivated by privacy concerns. The main concern considered in
-this specification is the protection of the user's personal identity, i.e., the identification of a [=natural person=] or a
-correlation of separate identities as belonging to the same [=natural person=]. Although the [=Web Authentication API=] does not
-use or provide any form of global identity, the following kinds of potentially correlatable identifiers are used:
+this specification is the protection of the user's personal identity, i.e., the identification of a human being or a correlation
+of separate identities as belonging to the same human being. Although the [=Web Authentication API=] does not use or provide any
+form of global identity, the following kinds of potentially correlatable identifiers are used:
 
 - The user's [=credential IDs=] and [=credential public keys=].
 
--- a/index.bs
+++ b/index.bs
@@ -4862,9 +4862,7 @@ use for forging logins at other [=[RPS]=].
 
 In the case where a [=[RP]=] requires [=biometric recognition=], this is performed locally by the [=biometric authenticator=]
 perfoming [=user verification=] and then signaling the result by setting the [=UV=] [=flag=] in the signed [=assertion=] response,
-instead of revealing the biometric data itself to the [=[RP]=]. The [=[RP]=] can trust the authenticity of the [=UV=] [=flag=] as
-long as it trusts the security guarantees of the [=attestation certificate=] presented when the [=biometric authenticator=] was
-[=registration|registered=].
+instead of revealing the biometric data itself to the [=[RP]=].
 
 
 ## Attestation Privacy ## {#sec-attestation-privacy}
--- a/index.bs
+++ b/index.bs
@@ -4785,7 +4785,7 @@ The privacy principles in [[!FIDO-Privacy-Principles]] also apply to this specif
 Many aspects of the design of the [=Web Authentication API=] are motivated by privacy concerns. The main concern considered in
 this specification is the protection of the user's personal identity, i.e., the identification of a human being or a correlation
 of separate identities as belonging to the same human being. Although the [=Web Authentication API=] does not use or provide any
-form of global identity, the following kinds of potentially correlatable identities are used:
+form of global identity, the following kinds of potentially correlatable identifiers are used:
 
 - The user's [=credential IDs=] and [=credential public keys=].
 
@@ -4798,7 +4798,7 @@ form of global identity, the following kinds of potentially correlatable identit
     These identities are obviously used by each [=[RP]=] to identify a user in their system. They are also visible to the
     [=client=] in the communication with the [=authenticator=].
 
-- The user's biometric information, e.g., fingerprints or facial recognition data.
+- The user's biometric characteristic(s), e.g., fingerprints or facial recognition data [[ISOBiometricVocabulary]].
 
     This is optionally used by the [=authenticator=] to perform [=user verification=]. It is not revealed to the [=[RP]=], but in
     the case of [=platform authenticators=], it might be visible to the [=client=] depending on the implementation.
@@ -4822,14 +4822,13 @@ prevent malicious [=[RPS]=] from using it to discover a user's personal identity
 [INFORMATIVE]
 
 Although [=Credential IDs=] and [=credential public keys=] are necessarily shared with the [=[RP]=] to enable strong
-authentication, the [=public key credentials=] employed in this specification are designed to be minimally identifying and never
-shared between [=[RPS]=].
+authentication, they are designed to be minimally identifying and not shared between [=[RPS]=].
 
 - [=Credential IDs=] and [=credential public keys=] are meaningless in isolation, as they only identify [=credential key pairs=]
     and not users directly.
 
-- Each [=public key credential=] is strictly bound to a specific [=[RP]=], and the [=client=] ensures that its existence is never
-    revealed to other [=[RPS]=]. A malicious [=[RP]=] thus cannot ask the [=client=] to reveal other identities owned by the user.
+- Each [=public key credential=] is strictly bound to a specific [=[RP]=], and the [=client=] ensures that its existence is not
+    revealed to other [=[RPS]=]. A malicious [=[RP]=] thus cannot ask the [=client=] to reveal a user's other identities.
 
 - The [=client=] also ensures that the existence of a [=public key credential=] is not revealed to the [=[RP]=] without [=user
     consent=]. This is detailed further in [[#sec-make-credential-privacy]] and [[#sec-assertion-privacy]]. A malicious [=[RP]=]
@@ -4855,12 +4854,12 @@ without a traditional username, further improving non-correlatability between [=
 authenticators=] the biometric data might also be visible to the [=client=], depending on the implementation. Biometric data is
 not revealed to the [=[RP]=]; it is only used as an additional layer of security for unlocking use of the [=authenticator=]. A
 malicious [=[RP]=] therefore cannot discover the user's personal identity via biometric data, and a security breach at a [=[RP]=]
-cannot expose biometric data for an attacker to use for forging logins at a different [=[RP]=].
+cannot expose biometric data for an attacker to use for forging logins at other [=[RPS]=].
 
-In the case where a [=[RP]=] requires [=biometric recognition=], this is instead achieved by the [=biometric authenticator=]
-setting the [=UV=] [=flag=] in the signed [=assertion=] response. The [=[RP]=] can trust the authenticity of this bit as long as
-it trusts the security guarantees of the [=attestation certificate=] presented when the [=biometric authenticator=] was
-[=registration|registered=].
+In the case where a [=[RP]=] requires [=biometric recognition=], this is achieved by the [=biometric authenticator=] setting the
+[=UV=] [=flag=] in the signed [=assertion=] response, instead of revealing the biometric data itself to the [=[RP]=]. The [=[RP]=]
+can trust the authenticity of the [=UV=] [=flag=] as long as it trusts the security guarantees of the [=attestation certificate=]
+presented when the [=biometric authenticator=] was [=registration|registered=].
 
 
 ## Attestation Privacy ## {#sec-attestation-privacy}
--- a/index.bs
+++ b/index.bs
@@ -4777,6 +4777,88 @@ which should make use of the existing browser permissions framework for the Geol
 
 The privacy principles in [[!FIDO-Privacy-Principles]] also apply to this specification.
 
+
+## De-anonymization prevention measures ## {#sctn-privacy-attacks}
+
+[INFORMATIVE]
+
+Many aspects of the design of the [=Web Authentication API=] are motivated by privacy concerns. The main concern considered in
+this specification is the protection of the user's personal identity, i.e., the identification of a physical person or a
+correlation of separate identities as belonging to the same physical person. Although the [=Web Authentication API=] does not use
+or provide and form of global identity, the following kinds of potentially correlatable identities are used:
+
+- The user's [=credential IDs=] and [=credential public keys=].
+
+    These are registered by the [=[RP]=] and subsequently used by the user to prove possession of the corresponding [=credential
+    private key=], and thus ownership of the [=public key credential=] identified by the [=credential ID=]. They are also visible
+    to the [=client=] in the communication with the [=authenticator=].
+
+- The user's identities specific to each [=[RP]=], e.g. usernames, [=user handles=] and similar aliases.
+
+    These identities are obviously used by each [=[RP]=] to identify a user in their system. They are also visible to the
+    [=client=] in the communication with the [=authenticator=].
+
+- The user's biometric information, e.g., fingerprints or facial recognition data.
+
+    This is optionally used by the [=authenticator=] to perform [=user verification=]. It is not revealed to the [=[RP]=], but in
+    the case of [=platform authenticators=] it might be visible to the [=client=] depending on the implementation.
+
+- The models of the user's [=authenticators=], e.g. product names.
+
+    This is exposed in the [=attestation statement=] provided to the [=[RP]=] during [=registration=]. It is also visible to the
+    [=client=] in the communication with the [=authenticator=].
+
+- The identities of the user's [=authenticators=], e.g. serial numbers.
+
+    This is possibly used by the [=client=] to enable communication with the [=authenticator=], but is not exposed to the
+    [=[RP]=].
+
+Some of the above information is necessarily shared with the [=[RP]=]. The following sections describe the measures taken to
+prevent malicious [=[RPS]=] from using it to discover a user's personal identity.
+
+
+## Anonymous, scoped, non-correlatable [=public key credentials=] ## {#sec-non-correlatable-credentials}
+
+[INFORMATIVE]
+
+Although [=Credential IDs=] and [=credential public keys=] are necessarily shared with the [=[RP]=] to enable strong
+authentication, the [=public key credentials=] employed in this specification are designed to be minimally identifying and never
+shared between [=[RPS]=].
+
+- [=Credential IDs=] and [=credential public keys=] are meaningless in isolation, as they only identify [=credential key pairs=]
+    and not users directly.
+
+- Each [=public key credential=] is strictly bound to a specific [=[RP]=], and the [=client=] ensures that its existence is never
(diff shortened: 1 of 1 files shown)


emlun:
I'm starting to feel like many of these more "explainer" type sections might belong better in a separate appendix. The point of them is to make the spec more accessible and easier to understand - but the spec is already gigantic, and having them inline in the spec might just make it even more intimidating. What do others think?

index.bs
```diff
@@ -4777,6 +4777,92 @@ which should make use of the existing browser permissions framework for the Geol
 
 The privacy principles in [[!FIDO-Privacy-Principles]] also apply to this specification.
 
+
+## De-anonymization prevention measures ## {#sctn-privacy-attacks}
+
+[INFORMATIVE]
+
+Many aspects of the design of the [=Web Authentication API=] are motivated by privacy concerns. The main concern considered in
+this specification is the protection of the user's personal identity, i.e., the identification of a physical person or a
```
selfissued: physical person -> human being (replace all occurrences)

index.bs
```diff
@@ -4777,6 +4777,92 @@ which should make use of the existing browser permissions framework for the Geol
...
+correlation of separate identities as belonging to the same physical person. Although the [=Web Authentication API=] does not use
+or provide and form of global identity, the following kinds of potentially correlatable identities are used:
```
selfissued: and -> any

index.bs
```diff
@@ -4777,6 +4777,92 @@ which should make use of the existing browser permissions framework for the Geol
...
+
+- The user's [=credential IDs=] and [=credential public keys=].
+
+    These are registered by the [=[RP]=] and subsequently used by the user to prove possession of the corresponding [=credential
+    private key=], and thus ownership of the [=public key credential=] identified by the [=credential ID=]. They are also visible
+    to the [=client=] in the communication with the [=authenticator=].
+
+- The user's identities specific to each [=[RP]=], e.g. usernames, [=user handles=] and similar aliases.
```
selfissued: I have no idea what "similar aliases" refers to.  Maybe just say "e.g., usernames and [=user handles=].  (Also note the comma after "e.g.".

index.bs
```diff
@@ -4777,6 +4777,92 @@ which should make use of the existing browser permissions framework for the Geol
...
+
+    These identities are obviously used by each [=[RP]=] to identify a user in their system. They are also visible to the
+    [=client=] in the communication with the [=authenticator=].
+
+- The user's biometric information, e.g., fingerprints or facial recognition data.
+
+    This is optionally used by the [=authenticator=] to perform [=user verification=]. It is not revealed to the [=[RP]=], but in
+    the case of [=platform authenticators=] it might be visible to the [=client=] depending on the implementation.
```
selfissued: Add the missing comma after "[=platform authenticators=].

index.bs
```diff
@@ -4777,6 +4777,92 @@ which should make use of the existing browser permissions framework for the Geol
...
+
+- The models of the user's [=authenticators=], e.g. product names.
```
selfissued: A comma is always required after "e.g.".  Please add it to all uses.

index.bs
```diff
@@ -4777,6 +4777,92 @@ which should make use of the existing browser permissions framework for the Geol
...
+
+- [=Credential IDs=] and [=credential public keys=] are meaningless in isolation, as they only identify [=credential key pairs=]
+    and not users directly.
+
+- Each [=public key credential=] is strictly bound to a specific [=[RP]=], and the [=client=] ensures that its existence is never
+    revealed to other [=[RPS]=]. A malicious [=[RP]=] thus cannot ask the [=client=] to reveal other identities owned by the user.
+
+- The [=client=] also ensures that the existence of a [=public key credential=] is not revealed to the [=[RP]=] without [=user
+    consent=]. This is detailed further in [[#sec-make-credential-privacy]] and [[#sec-assertion-privacy]]. A malicious [=[RP]=]
+    thus cannot silently identify a user even if the user has a [=public key credential=] registered and available.
```
selfissued: Add a comma between "user" and "even".

index.bs
```diff
@@ -4777,6 +4777,92 @@ which should make use of the existing browser permissions framework for the Geol
...
+
+- [=Authenticators=] ensure that the [=credential IDs=] and [=credential public keys=] of different [=public key credentials=] are
+    not correlatable as having the same [=managing authenticator=]. A pair of malicious [=[RPS]=] thus cannot correlate users
```
selfissued: The phrase "are not correlatable as having the same [=managing authenticator=]" is awkward and hard to understand.  Do you mean "are not correlatable as belonging to the same [=authenticator=]"?  Or perhaps "are not correlatable as belonging to the same user"?

index.bs
```diff
@@ -4777,6 +4777,92 @@ which should make use of the existing browser permissions framework for the Geol
...
+    between their systems without additional information, e.g., a willfully reused username or e-mail address.
+
+- [=Authenticators=] ensure that their [=attestation certificates=] are not unique enough to identify a single [=authenticator=]
+    or a small group of [=authenticators=]. This is detailed further in [[#sec-attestation-privacy]]. A pair of malicious
+    [=[RPS]=] thus cannot correlate users between their systems by tracking individual [=authenticators=].
+
+Additionally, a [=public key credential=] with a [=client-side-resident credential private key=] can optionally include a [=user
```
selfissued: I don't see any target in the spec that "[=client-side-resident credential private key=]" refers to.  Please review the uses of "[= ... =]" in this PR and ensure that all of them refer to locations with matching <dfn> locations.

index.bs

index.bs
emlun: Fixed.

index.bs
emlun: Both done.

index.bs
emlun: Fixed.

index.bs
emlun: Fixed.

index.bs
emlun: Done; I went with "correlatable as belonging to the same user".

emlun:
@selfissued Please re-review!

nadalin:
@equalsJeffH Please review

index.bs
```diff
@@ -4777,6 +4777,92 @@ which should make use of the existing browser permissions framework for the Geol
...
+handle=] specified by the [=[RP]=]. The [=public key credential|credential=] can then be used to both identify and
+[=authentication|authenticate=] the user. This means that a privacy-conscious [=[RP]=] can allow the user to create an account
+without a traditional username, further improving non-correlatability between [=[RPS]=].
+
+
+## Authenticator-local [=biometric recognition=] ## {#sec-biometric-privacy}
+
+[=Biometric authenticators=] perform the [=biometric recognition=] internally in the [=authenticator=] - though for [=platform
+authenticators=] the biometric data might also be visible to the [=client=], depending on the implementation. Biometric data is
+not revealed to the [=[RP]=]; it is only used as an additional layer of security for unlocking use of the [=authenticator=]. A
```
equalsJeffH: hm, for "it is only used as an additional layer of security for unlocking use of the [=authenticator=]", suggest:

it is used only locally to obtain [=user consent=] to create and [=registration|register=], or [=authentication|authenticate=] using, a [=public key credential=].

index.bs
```diff
@@ -4777,6 +4777,92 @@ which should make use of the existing browser permissions framework for the Geol
...
+malicious [=[RP]=] therefore cannot discover the user's personal identity via biometric data, and a security breach at a [=[RP]=]
+cannot expose biometric data for an attacker to use for forging logins at a different [=[RP]=].
```
/ at a different [=[RP]=] / at other [=[RPS]=] /

index.bs
```diff
@@ -4777,6 +4777,92 @@ which should make use of the existing browser permissions framework for the Geol
...
+
+In the case where a [=[RP]=] requires [=biometric recognition=], this is instead achieved by the [=biometric authenticator=]
```
equalsJeffH: hm, i'm not sure why you're using the word "instead" here?  IIUC, the [=UV=] [=flag=] is set by an authnr whenever it successfully performs user verification?

index.bs
```diff
@@ -4777,6 +4777,92 @@ which should make use of the existing browser permissions framework for the Geol
...
+setting the [=UV=] [=flag=] in the signed [=assertion=] response. The [=[RP]=] can trust the authenticity of this bit as long as
```
equalsJeffH: yes, however, if an RP is not checking attstn, then it is "trusting" the entire contraption here, and certainly "can" do that.  Am not sure how to rephrase this. Perhaps it does not really need to be said?

index.bs
```diff
@@ -4777,6 +4777,92 @@ which should make use of the existing browser permissions framework for the Geol
 
 The privacy principles in [[!FIDO-Privacy-Principles]] also apply to this specification.
 
+
+## De-anonymization prevention measures ## {#sctn-privacy-attacks}
+
+[INFORMATIVE]
+
+Many aspects of the design of the [=Web Authentication API=] are motivated by privacy concerns. The main concern considered in
+this specification is the protection of the user's personal identity, i.e., the identification of a human being or a correlation
```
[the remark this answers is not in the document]
emlun: Shall I also add that as a non-normative definition reference?

index.bs
```diff
@@ -4777,6 +4777,92 @@ which should make use of the existing browser permissions framework for the Geol
...
+of separate identities as belonging to the same human being. Although the [=Web Authentication API=] does not use or provide any
+form of global identity, the following kinds of potentially correlatable identities are used:
```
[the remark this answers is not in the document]
emlun: Ah yes, good point.

index.bs
```diff
@@ -4777,6 +4777,92 @@ which should make use of the existing browser permissions framework for the Geol
...
+
+- The user's [=credential IDs=] and [=credential public keys=].
+
+    These are registered by the [=[RP]=] and subsequently used by the user to prove possession of the corresponding [=credential
+    private key=], and thus ownership of the [=public key credential=] identified by the [=credential ID=]. They are also visible
```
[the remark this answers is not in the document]
emlun: It's not clear to me how that would improve this sentence - care to explain what you mean?

index.bs
```diff
@@ -4777,6 +4777,92 @@ which should make use of the existing browser permissions framework for the Geol
...
+    to the [=client=] in the communication with the [=authenticator=].
+
+- The user's identities specific to each [=[RP]=], e.g., usernames and [=user handles=].
+
+    These identities are obviously used by each [=[RP]=] to identify a user in their system. They are also visible to the
+    [=client=] in the communication with the [=authenticator=].
+
+- The user's biometric information, e.g., fingerprints or facial recognition data.
```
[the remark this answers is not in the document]

index.bs
```diff
@@ -4777,6 +4777,92 @@ which should make use of the existing browser permissions framework for the Geol
...
+Some of the above information is necessarily shared with the [=[RP]=]. The following sections describe the measures taken to
+prevent malicious [=[RPS]=] from using it to discover a user's personal identity.
+
+
+## Anonymous, scoped, non-correlatable [=public key credentials=] ## {#sec-non-correlatable-credentials}
+
+[INFORMATIVE]
+
+Although [=Credential IDs=] and [=credential public keys=] are necessarily shared with the [=[RP]=] to enable strong
+authentication, the [=public key credentials=] employed in this specification are designed to be minimally identifying and never
```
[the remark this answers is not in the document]

index.bs
```diff
@@ -4777,6 +4777,92 @@ which should make use of the existing browser permissions framework for the Geol
...
+shared between [=[RPS]=].
+
+- [=Credential IDs=] and [=credential public keys=] are meaningless in isolation, as they only identify [=credential key pairs=]
+    and not users directly.
+
+- Each [=public key credential=] is strictly bound to a specific [=[RP]=], and the [=client=] ensures that its existence is never
```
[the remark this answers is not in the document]

index.bs
```diff
@@ -4777,6 +4777,92 @@ which should make use of the existing browser permissions framework for the Geol
...
+    revealed to other [=[RPS]=]. A malicious [=[RP]=] thus cannot ask the [=client=] to reveal other identities owned by the user.
```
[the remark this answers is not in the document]

index.bs
emlun: Hm, I think that formulation might be inaccurate, though - biometrics are not only used to obtain consent, but also as an additional authentication factor. That may not be relevant in this section, but I wouldn't like to suggest that they're used only as a way to obtain user consent (which could be interpreted to just mean "user presence").

But I like the initial rewording "it is used only localy", at least.

index.bs

index.bs
emlun: Hm, good point. What I meant was that the authenticator sets the (verifiably authentic) UV flag instead of sending the biometric data to the server, but you're right that the right hand side of that "instead" isn't clear enough.

index.bs
emlun: I think it's worthwhile to point out how this is dfferent from the usual kind of (completely worthless) client-side authentication. And if the RP is not checking attestation, then the RP _cannot_ trust that the UV flag is authentic, whether or not the RP chooses to act upon its value anyway. If the RP actually _requires_ UV, then the RP MUST verify the attestation statement - otherwise that requirement is nothing more than wishful thinking.

index.bs
emlun: I reformulated this a bit, does that look better?

emlun: I'm also considering perhaps mentioning the `uvi` and `uvm` extensions here, what do you think about that?

nadalin:
@emlun Please merge

emlun:
There are still a few more unresolved discussion points in https://github.com/w3c/webauthn/pull/899#pullrequestreview-128168178.

index.bs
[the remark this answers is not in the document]
equalsJeffH: sure :)

index.bs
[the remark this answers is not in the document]
equalsJeffH: well...  "ownership" has connotations of physical possession. the public key is handed to the RP who can do with it what they will.  It correlates to the cred source and the cred private key therein. such correlation is proved by verifying signatures produced by the cred priv key, using the cred pub key held by the verifier.

index.bs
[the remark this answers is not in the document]
emlun: Done; I'm not sure it belongs in the bibliography, though, so I didn't add it to there.

index.bs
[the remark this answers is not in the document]
emlun: Ok. I think the word "correlation" makes it feel like this has something to do with the non-correlatability we talk about elsewhere, which might be confusing.

Would it perhaps be better to just remove the whole ", and thus ownership [...]" subordinate clause?

index.bs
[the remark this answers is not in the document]
jyasskin: 1) I'm nervous about using legal terms unless a lawyer has reviewed them.
2) "Historically, a human being was not necessarily a natural person in some jurisdictions where slavery existed (subject of a property right) rather than a person." makes me think we'd prefer "human being" here anyway.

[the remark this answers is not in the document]
emlun: Valid points. @equalsJeffH What do you think, do we revert this?

index.bs
equalsJeffH: well, yeah, that formulation is dancing around the "first multi-factor" aka "first factor" (of a multi-factor authn) notions (which themselves are something else to clean up and clarify in the spec, cf. issue #462). our dfn for [user consent](https://w3c.github.io/webauthn/#user-consent) does link off to further terms and leads to [user verification](https://w3c.github.io/webauthn/#user-verification).

So here's a reformulation using [user verification](https://w3c.github.io/webauthn/#user-verification) instead:

... it is used only locally to perform [=user verification=] authorizing the creation and [=registration=] of, or [=authentication=] using, a [=public key credential=]. 

WDYT?

index.bs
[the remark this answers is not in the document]
equalsJeffH: ok, yeah wrt "correlation".

> Would it perhaps be better to just remove the whole ", and thus ownership [...]" subordinate clause?

works for me, thx.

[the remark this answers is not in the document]
emlun: Ok, will do.

index.bs
emlun: Ok, that looks good to me.

index.bs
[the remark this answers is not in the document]
equalsJeffH: oh geeze....  I'd still use natural person but just not link it to wikipedia, or sure, use human being....

index.bs
```diff
@@ -4777,6 +4781,91 @@ which should make use of the existing browser permissions framework for the Geol
...
+## Authenticator-local [=biometric recognition=] ## {#sec-biometric-privacy}
+
+[=Biometric authenticators=] perform the [=biometric recognition=] internally in the [=authenticator=] - though for [=platform
+authenticators=] the biometric data might also be visible to the [=client=], depending on the implementation. Biometric data is
+not revealed to the [=[RP]=]; it is used only locally to perform [=user verification=] authorizing the creation and
+[=registration=] of, or [=authentication=] using, a [=public key credential=]. A malicious [=[RP]=] therefore cannot discover the
+user's personal identity via biometric data, and a security breach at a [=[RP]=] cannot expose biometric data for an attacker to
+use for forging logins at other [=[RPS]=].
+
+In the case where a [=[RP]=] requires [=biometric recognition=], this is achieved by the [=biometric authenticator=] setting the
```
equalsJeffH: perhaps rephrase as: 

... , this is performed locally by the [=biometric authenticator=] performing [=user verification=] and then signaling the result by setting the [=UV=] [=flag=] in the signed [=assertion=] response, instead of revealing the biometric data itself to the [=[RP]=]. 

I'm not sure about this part because there's going to be lots of RPs who "trust" webauthn assertions even though they do  not verify attestations:

The [=[RP]=]
+can trust the authenticity of the [=UV=] [=flag=] as long as it trusts the security guarantees of the [=attestation certificate=]
+presented when the [=biometric authenticator=] was [=registration|registered=].

emlun: Thanks, formulation looks very good!

>I'm not sure about this part because there's going to be lots of RPs who "trust" webauthn assertions even though they do not verify attestations:

That is true, and the assertion as a whole can indeed be reasonably trusted without verifying the attestation statement (since a MitM attack on it, while theoretically possible, would be extremely difficult to maintain). The RP can't trust that the authenticator provides any given level of protection for the private key, but it can trust that something on the client side has possession of a particular private key. However, this part is specifically concerned with the authenticity of the UV bit, which can _only_ be trusted if there's a trusted attestation signature.

equalsJeffH: agreed on most of the above, but...

> However, this part is specifically concerned with the authenticity of the UV bit, which can only be trusted if there's a trusted attestation signature.

well, i disagree with that -- trust is in the eye of the RP.  If an RP leaves [AttestationConveyancePreference](https://w3c.github.io/webauthn/#dom-publickeycredentialcreationoptions-attestation) defaulting to none, it's still totally their decision whether to "trust" the UV bit flag, and "how much" to trust it.  Amongst RPs performing "risk based authn", "trust" is a continuum. Amongst less sophisticated long-tail RPs, even with ignoring attestation, webauthn/fido-based authn is WAY better than shared secret username/password.

emlun: Ok, true. The point I'm trying to make is that this statement is in the context of "RP **requires** biometric recognition", which I would expect means a hard requirement that absolutely must be satisfied, and I would say risk-based authn falls outside that.

But ok, perhaps the word "trust" is the issue here. Maybe something like one of these?

- The RP can verify that the UV flag is authentic if the RP trusts the biometric authenticator's attestation certificate, presented when the credential was registered.
- The UV flag is guaranteed to be accurate if the RP trusts the biometric authenticator's attestation certificate, presented when the credential was registered.
- The biometric authenticator's attestation certificate, presented when the credential was registered, can attest to the accuracy of the UV flag.

equalsJeffH: hm, it seems to me that "guaranteeing" that the value of the UV is "accurate" is multi-faceted, beyond simply the presence of verifiable attestation.  Even if one has verified the attestation, retrieved the corresponding authnr metadata, verified metadata sigs, verified that the authnr metadata asserts that the authnr performs user verification, it is possible for the authnr to have non-negligible (for some definition of negligible) False Acceptance Rate (i.e., is a lousy UV impl).  What the ecosystem has to (ostensibly) protect against the latter scenario are the FIDO [authnr](https://fidoalliance.org/certification/authenticator-certification-levels/) and [biometric](https://drafts.fidoalliance.org/biometrics/requirements/latest/) certification programs, so add "verifying that the authnr is certified (perhaps to some particular "level")" to the list of things for the RP to verify, but that still is not a 100% guarantee. Note also not all authnrs will be certified. 

> The point I'm trying to make is that this statement is in the context of "RP requires biometric recognition", which I would expect means a hard requirement that absolutely must be satisfied...

well, in practical terms there is some wiggle room even if one has "a hard requirement that absolutely must be satisfied" because no biometric recognition system will be perfect. See: https://drafts.fidoalliance.org/biometrics/requirements/latest/Biometrics-Requirements-v1.0-wd-20180524.html#FARReq

I suspect that for simplistic low-value RPs (and possibly large RPs where user is involved in low-value interactions at that time), simply checking the assertion sig is valid and relying on UV flag value is Good Enough, whether or not the RP paid any attention to attestation. 

For high-value RPs (or more generally, RP where user is involved in a high-value interaction), they can perform all the above checks and assign a higher score to the value of the UV flag in their overall assessment of the state of the particular interaction. 

I'm not sure where we ought to document such detailed deployment considerations. And because "trust" is complex & subtle & RP-specific, I'm hesitant to make simplistic statements about it.

emlun: Oh boy, this turned out to be a gnarly one, didn't it... :) Thanks, you are indeed right on all points. I feel like what you suggested in https://github.com/w3c/webauthn/pull/899#discussion_r194915469 might be the way to go in the end, and dodge the issue instead:

>Perhaps it does not really need to be said?

I would have liked to clearly point out how this is different from usual client side authentication, but that doesn't seem important enough to hold this up over if you still agree that deleting the last sentence is a reasonable way to go. I also notice that the first sentence does mention that the flag is signed over, so there's at least a hint of it still there.

index.bs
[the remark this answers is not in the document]
equalsJeffH: after grovelling in the spec on another PR, i note that we use "human" in other terms/phrases, such as [human-palatable](https://w3c.github.io/webauthn/#human-palatability), so I'm thinking "human being" is fine. silly me.

index.bs
```diff
@@ -4777,6 +4781,92 @@ which should make use of the existing browser permissions framework for the Geol
...
+
+- The models of the user's [=authenticators=], e.g., product names.
+
+    This is exposed in the [=attestation statement=] provided to the [=[RP]=] during [=registration=]. It is also visible to the
+    [=client=] in the communication with the [=authenticator=].
+
+- The identities of the user's [=authenticators=], e.g., serial numbers.
+
+    This is possibly used by the [=client=] to enable communication with the [=authenticator=], but is not exposed to the
+    [=[RP]=].
```
equalsJeffH: wrt "The identities of the user's [=authenticators=]..." -- this confused me because it is separate from the immediately above models of authnrs revealed thru attestation, but this one does not mention attestation -- so is this referring to info that is revealed at he OS platform and transport layers which the [=client=] may see?  if so, perhaps clarify?

index.bs
equalsJeffH: I hesitate to mention those extensions here, mostly due to being unsure of whether they will be implemented (?).

index.bs
```diff
@@ -4777,6 +4781,92 @@ which should make use of the existing browser permissions framework for the Geol
...
+[=Biometric authenticators=] perform the [=biometric recognition=] internally in the [=authenticator=] - though for [=platform
+authenticators=] the biometric data might also be visible to the [=client=], depending on the implementation. Biometric data is
+not revealed to the [=[RP]=]; it is used only locally to perform [=user verification=] authorizing the creation and
+[=registration=] of, or [=authentication=] using, a [=public key credential=]. A malicious [=[RP]=] therefore cannot discover the
+user's personal identity via biometric data, and a security breach at a [=[RP]=] cannot expose biometric data for an attacker to
+use for forging logins at other [=[RPS]=].
+
+In the case where a [=[RP]=] requires [=biometric recognition=], this is performed locally by the [=biometric authenticator=]
+perfoming [=user verification=] and then signaling the result by setting the [=UV=] [=flag=] in the signed [=assertion=] response,
+instead of revealing the biometric data itself to the [=[RP]=]. The [=[RP]=] can trust the authenticity of the [=UV=] [=flag=] as
```
equalsJeffH: so we'll delete this final sentence?

index.bs
equalsJeffH: see thread below https://github.com/w3c/webauthn/pull/899#pullrequestreview-128485441 for resolution

index.bs
```diff
@@ -4777,6 +4781,92 @@ which should make use of the existing browser permissions framework for the Geol
 
 The privacy principles in [[!FIDO-Privacy-Principles]] also apply to this specification.
 
+
+## De-anonymization prevention measures ## {#sctn-privacy-attacks}
+
+[INFORMATIVE]
+
+Many aspects of the design of the [=Web Authentication API=] are motivated by privacy concerns. The main concern considered in
+this specification is the protection of the user's personal identity, i.e., the identification of a [=natural person=] or a
```
equalsJeffH: see https://github.com/w3c/webauthn/pull/899#discussion_r195522026 -- human being is fine, silly me. sorry/thx.

index.bs
emlun: Maybe this is just a brain fart on my part - the WebAuthn spec says absolutely nothing about this, so I agree it's probably confusing to even bring it up. The OS may be able to see serial numbers and such on the device level (YubiKeys provide APIs for querying for it, for example), but it should never be exposed to the RP.

So I'm thinking I'll just delete this bullet. That sound good?

index.bs
emlun: Yep, done.

index.bs
emlun: Done!

PR #900, opened 2018-05-09
Finish up hot-plugging algorithm language

### LINKED ISSUE #613: Allow hot-plugged authenticators?

emlun:
This fixes #613.

This makes changes to the client algorithms, so it could technically be construed as breaking, but in practice this should be more of an editorial re-wording that still agrees with what implementations actually do. At least, that is the intent.

- This merges the previous step 19 of [makeCredential][mc] in as a switch case of step 20, and the previous step 18 of [getAssertion][ga] in as a switch case of step 19. This way there is only one step in each algorithm that tries to express things to do asynchronously for the duration of the timer.

- The inline `Issue:`s mentioning underspecified behaviour are replaced with descriptions of an abstract "set of presently available authenticators" and `Note:`s indicating that this is intentionally underspecified and meant to represent different connection and discovery mechanisms all in one.

- This also swaps the order of the previous steps 17 ("Start _lifetimeTimer_.") and 18 ("Let _issuedRequests_ be a new ordered set") of [makeCredential][mc], for consistency with [getAssertion][ga] and so that the step "Start _lifetimeTimer_." immediately precedes the step "While _lifetimeTimer_ has not expired [...]".

- This also fixes the `Issue:` about the incorrect scope of _savedCredentialId_ in [getAssertion][ga]. This variable is replaced by a top-level _savedCredentialIds_ variable containing a map of <i>authenticator</i>s to credential IDs, to preserve the previous scoping of the _savedCredential_ variable to a specific _authenticator_.

[mc]: https://w3c.github.io/webauthn/#createCredential
[ga]: https://w3c.github.io/webauthn/#getAssertion

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/pull/900.html" title="Last updated on May 30, 2018, 6:23 PM GMT (c8f110d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/900/2c6faa8...c8f110d.html" title="Last updated on May 30, 2018, 6:23 PM GMT (c8f110d)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -877,7 +877,7 @@ When this method is invoked, the user agent MUST execute the following algorithm
 
     Note: What qualifies an [=authenticator=] as "available" is intentionally unspecified; this is meant to represent how
     [=authenticators=] can be hot-plugged into (e.g., via USB) or discovered (e.g., via NFC or Bluetooth) by the [=client=] by
-    various mechanisms.
+    various mechanisms, or permanently built into the [=client=].
 
 1. Start |lifetimeTimer|.
 
@@ -895,7 +895,9 @@ When this method is invoked, the user agent MUST execute the following algorithm
             whose name is "{{AbortError}}" and terminate this algorithm.
 
         :   If an |authenticator| becomes available on this platform,
-        ::  1. If <code>|options|.{{PublicKeyCredentialCreationOptions/authenticatorSelection}}</code> is [=present=]:
+        ::  Note: This includes if the |authenticator| was initially available when |lifetimeTimer| started.
+
+            1. If <code>|options|.{{PublicKeyCredentialCreationOptions/authenticatorSelection}}</code> is [=present=]:
 
                   1. If <code>|options|.{{PublicKeyCredentialCreationOptions/authenticatorSelection}}.{{authenticatorAttachment}}</code> is
                     [=present|present=] and its value is not equal to |authenticator|'s attachment modality, [=iteration/continue=].
@@ -1210,7 +1212,7 @@ When this method is invoked, the user agent MUST execute the following algorithm
 
     Note: What qualifies an [=authenticator=] as "available" is intentionally unspecified; this is meant to represent how
     [=authenticators=] can be hot-plugged into (e.g., via USB) or discovered (e.g., via NFC or Bluetooth) by the [=client=] by
-    various mechanisms.
+    various mechanisms, or permanently built into the [=client=].
 
 1. Start |lifetimeTimer|.
 
@@ -1230,7 +1232,9 @@ When this method is invoked, the user agent MUST execute the following algorithm
             return a {{DOMException}} whose name is "{{AbortError}}" and terminate this algorithm.
 
         :   If an |authenticator| becomes available on this platform,
-        ::  1. If <code>|options|.{{PublicKeyCredentialRequestOptions/userVerification}}</code> is set to
+        ::  Note: This includes if the |authenticator| was initially available when |lifetimeTimer| started.
+
+            1. If <code>|options|.{{PublicKeyCredentialRequestOptions/userVerification}}</code> is set to
                 {{UserVerificationRequirement/required}} and the |authenticator| is not capable of performing [=user verification=],
                 [=iteration/continue=].
 
--- a/index.bs
+++ b/index.bs
@@ -1202,6 +1202,8 @@ When this method is invoked, the user agent MUST execute the following algorithm
 
 1. Let |authenticator| be a platform-specific handle whose value identifies an [=authenticator=].
 
+1. Let |savedCredentialIds| be a new [=map=].
+
 1. Start |lifetimeTimer|.
 
 1. [=set/For each=] |authenticator| that becomes available on this platform during the lifetime of
@@ -1260,13 +1262,9 @@ When this method is invoked, the user agent MUST execute the following algorithm
 
                 1. Let |distinctTransports| be a new [=ordered set=].
 
-                1. If |allowCredentialDescriptorList| has exactly one value, let |savedCredentialId| be a new 
-                    {{PublicKeyCredentialDescriptor}}.{{PublicKeyCredentialDescriptor/id}} and set its value to <code>|allowCredentialDescriptorList|[0].id</code>'s
+                1. If |allowCredentialDescriptorList| has exactly one value, set
+                    <code>|savedCredentialIds|[|authenticator|]</code> to <code>|allowCredentialDescriptorList|[0].id</code>'s
                     value (see [here](#authenticatorGetAssertion-return-values) in [[#op-get-assertion]] for more information).
-
-                Issue: The foregoing step _may_ be incorrect, in that we are attempting to create |savedCredentialId|
-                    here and use it later below, and we do not have a global in which to allocate a place for it. Perhaps this
-                    is good enough?  addendum: [@jcjones feels the above step is likely good enough](https://github.com/w3c/webauthn/pull/665#discussion_r148130187).
 
                 1. [=list/For each=] credential descriptor |C| in |allowCredentialDescriptorList|,
                     [=set/append=] each value, if any, of <code>|C|.{{transports}}</code> to |distinctTransports|.
@@ -1336,10 +1334,10 @@ When this method is invoked, the user agent MUST execute the following algorithm
             1.  Let <var ignore>assertionCreationData</var> be a [=struct=] whose [=items=] are:
 
                 :   <code><dfn for="assertionCreationData">credentialIdResult</code>
-                ::  If |savedCredentialId| exists, set the value of [=credentialIdResult=] to be the bytes of
-                    |savedCredentialId|. Otherwise, set the value of [=credentialIdResult=] to be the bytes of the
-                    [=credential ID=] returned from the successful [=authenticatorGetAssertion=] operation, as defined in
-                    [[#op-get-assertion]].
+                ::  If <code>|savedCredentialIds|[|authenticator|]</code> exists, set the value of [=credentialIdResult=] to be
+                    the bytes of <code>|savedCredentialIds|[|authenticator|]</code>. Otherwise, set the value of
+                    [=credentialIdResult=] to be the bytes of the [=credential ID=] returned from the successful
+                    [=authenticatorGetAssertion=] operation, as defined in [[#op-get-assertion]].
 
                 :   <code><dfn for="assertionCreationData">clientDataJSONResult</dfn></code>
                 ::  whose value is the bytes of |clientDataJSON|.
--- a/index.bs
+++ b/index.bs
@@ -874,11 +874,11 @@ When this method is invoked, the user agent MUST execute the following algorithm
 
 1. Let |issuedRequests| be a new [=ordered set=].
 
-1. [=set/For each=] |authenticator| that becomes available on this platform during the lifetime of |lifetimeTimer|, do the
-    following:
+1. [=set/For each=] |authenticator| that becomes available on this platform until |lifetimeTimer| expires, perform the following
+    steps:
 
-    Issue: The definitions of "lifetime of" and "becomes available" are intended to represent how
-    devices are hot-plugged into (USB) or discovered by (NFC) browsers, and are underspecified.
+    Issue: The definition of "becomes available" is intended to represent how
+    devices are hot-plugged into (USB) or discovered by (NFC) browsers, and is underspecified.
     Resolving this with good definitions or some other means will be addressed by resolving
     [Issue #613](https://github.com/w3c/webauthn/issues/613).
 
@@ -1206,11 +1206,11 @@ When this method is invoked, the user agent MUST execute the following algorithm
 
 1. Start |lifetimeTimer|.
 
-1. [=set/For each=] |authenticator| that becomes available on this platform during the lifetime of
-    |lifetimeTimer|, perform the following steps:
+1. [=set/For each=] |authenticator| that becomes available on this platform until |lifetimeTimer| expires, perform the following
+    steps:
 
-    Issue: The definitions of "lifetime of" and "becomes available" are intended to represent how
-    devices are hot-plugged into (USB) or discovered by (NFC) browsers, and are underspecified.
+    Issue: The definition of "becomes available" is intended to represent how
+    devices are hot-plugged into (USB) or discovered by (NFC) browsers, and is underspecified.
     Resolving this with good definitions or some other means will be addressed by resolving
     [Issue #613](https://github.com/w3c/webauthn/issues/613).
 
--- a/index.bs
+++ b/index.bs
@@ -870,88 +870,91 @@ When this method is invoked, the user agent MUST execute the following algorithm
     [=AbortSignal/aborted flag=] is set to true, return a {{DOMException}} whose name is "{{AbortError}}" 
     and terminate this algorithm.
 
-1. Start |lifetimeTimer|.
-
 1. Let |issuedRequests| be a new [=ordered set=].
 
-1. [=set/For each=] |authenticator| that becomes available on this platform until |lifetimeTimer| expires, perform the following
-    steps:
+1. Let |authenticators| represent a [=set=] of platform-specific handles, where each value identifies an [=authenticator=]
+    presently available on this platform at a given instant.
 
-    Issue: The definition of "becomes available" is intended to represent how
-    devices are hot-plugged into (USB) or discovered by (NFC) browsers, and is underspecified.
-    Resolving this with good definitions or some other means will be addressed by resolving
-    [Issue #613](https://github.com/w3c/webauthn/issues/613).
+    Note: What qualifies an [=authenticator=] as "available" is intentionally unspecified; this is meant to represent how
+    [=authenticators=] can be hot-plugged into (e.g., via USB) or discovered (e.g., via NFC or Bluetooth) by the [=client=] by
+    various mechanisms.
 
-    1. If <code>|options|.{{PublicKeyCredentialCreationOptions/authenticatorSelection}}</code> is [=present=]:
+1. Start |lifetimeTimer|.
 
-          1. If <code>|options|.{{PublicKeyCredentialCreationOptions/authenticatorSelection}}.{{authenticatorAttachment}}</code> is
-            [=present|present=] and its value is not equal to |authenticator|'s attachment modality, [=iteration/continue=].
-          1. If <code>|options|.{{PublicKeyCredentialCreationOptions/authenticatorSelection}}.{{requireResidentKey}}</code> is set to
-            `true` and the |authenticator| is not capable of storing a [=Client-Side-Resident Credential Private Key=],
-            [=iteration/continue=].
-          1. If <code>|options|.{{PublicKeyCredentialCreationOptions/authenticatorSelection}}.{{AuthenticatorSelectionCriteria/userVerification}}</code> is
-            set to {{UserVerificationRequirement/required}} and the |authenticator| is not capable of performing [=user
-            verification=], [=iteration/continue=].
-
-    1. Let |userVerification| be the <dfn>effective user verification requirement for credential creation</dfn>, a Boolean value,
-        as follows. If
-        <code>|options|.{{PublicKeyCredentialCreationOptions/authenticatorSelection}}.{{AuthenticatorSelectionCriteria/userVerification}}</code>
-
-            <dl class="switch">
-
-                :   is set to {{UserVerificationRequirement/required}}
-                ::  Let |userVerification| be `true`.
-
-                :   is set to {{UserVerificationRequirement/preferred}}
-                ::  If the |authenticator|
-
-                    <dl class="switch">
-                        :   is capable of [=user verification=]
-                        ::  Let |userVerification| be `true`.
-
-                        :   is not capable of [=user verification=]
-                        ::  Let |userVerification| be `false`.
-                    </dl>
-
-                :   is set to {{UserVerificationRequirement/discouraged}}
-                ::  Let |userVerification| be `false`.
-
-            </dl>
-
-    1. Let |userPresence| be a Boolean value set to the inverse of |userVerification|.
-
-    1. Let |excludeCredentialDescriptorList| be a new [=list=].
-
-    1. [=list/For each=] credential descriptor |C| in <code>|options|.{{PublicKeyCredentialCreationOptions/excludeCredentials}}</code>:
-        1. If <code>|C|.{{transports}}</code> [=list/is not empty=], and |authenticator| is connected over a transport not
-            mentioned in <code>|C|.{{transports}}</code>, the client MAY [=continue=].
-        1. Otherwise, [=list/Append=] |C| to |excludeCredentialDescriptorList|.
-
-<!-- @@EDITOR-ANCHOR-01A: KEEP THIS LIST SYNC'D WITH THE LIST UP AT @@EDITOR-ANCHOR-01B -->
-    1. Invoke the [=authenticatorMakeCredential=] operation on |authenticator| with
-        |clientDataHash|,
-        <code>|options|.{{PublicKeyCredentialCreationOptions/rp}}</code>, <code>|options|.{{PublicKeyCredentialCreationOptions/user}}</code>,
-        <code>|options|.{{PublicKeyCredentialCreationOptions/authenticatorSelection}}.{{AuthenticatorSelectionCriteria/requireResidentKey}}</code>,
-        |userPresence|,
-        |userVerification|,
-        |credTypesAndPubKeyAlgs|,
-        |excludeCredentialDescriptorList|,
-        and |authenticatorExtensions| as parameters.
-
-    1. [=set/Append=] |authenticator| to |issuedRequests|.
-
-1. [=While=] |lifetimeTimer| has not expired, perform the following actions depending upon |lifetimeTimer| and responses from the
-    authenticators:
+1. [=While=] |lifetimeTimer| has not expired, perform the following actions depending upon |lifetimeTimer|
+    and the state and response [=set/for each=] |authenticator| in |authenticators|:
     <dl class="switch">
         :   If |lifetimeTimer| expires,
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -870,88 +870,91 @@ When this method is invoked, the user agent MUST execute the following algorithm
...
-        as follows. If
-        <code>|options|.{{PublicKeyCredentialCreationOptions/authenticatorSelection}}.{{AuthenticatorSelectionCriteria/userVerification}}</code>
+        :   If the <code>|options|.{{CredentialCreationOptions/signal}}</code> is [=present=] and its
+            [=AbortSignal/aborted flag=] is set to true,
+        ::  [=set/For each=] |authenticator| in |issuedRequests| invoke the [=authenticatorCancel=]
+            operation on |authenticator| and [=set/remove=] |authenticator| from |issuedRequests|. Then return a {{DOMException}}
+            whose name is "{{AbortError}}" and terminate this algorithm.
 
-            <dl class="switch">
+        :   If an |authenticator| becomes available on this platform,
```
agl: Does this ignore authenticators that were already available when the process started? I don't see that existing authenticators aren't ignored by this. Perhaps mention all initially available authenticators and all that become available?

index.bs
```diff
@@ -1200,124 +1203,123 @@ When this method is invoked, the user agent MUST execute the following algorithm
...
-1. [=set/For each=] |authenticator| that becomes available on this platform during the lifetime of
-    |lifetimeTimer|, perform the following steps:
+1. Start |lifetimeTimer|.
 
-    Issue: The definitions of "lifetime of" and "becomes available" are intended to represent how
-    devices are hot-plugged into (USB) or discovered by (NFC) browsers, and are underspecified.
-    Resolving this with good definitions or some other means will be addressed by resolving
-    [Issue #613](https://github.com/w3c/webauthn/issues/613).
+1. [=While=] |lifetimeTimer| has not expired, perform the following actions depending upon |lifetimeTimer|
+    and the state and response [=set/for each=] |authenticator| in |authenticators|:
```
agl: (ditto.)

index.bs
```diff
@@ -1200,124 +1203,123 @@ When this method is invoked, the user agent MUST execute the following algorithm
...
-            <dl class="switch">
+        :   If the {{CredentialRequestOptions/signal}} member is [=present=] and the [=AbortSignal/aborted flag=] is set to
+            true,
+        ::  [=set/For each=] |authenticator| in |issuedRequests| invoke the [=authenticatorCancel=] operation on |authenticator|
+            and [=set/remove=] |authenticator| from |issuedRequests|. Then
+            return a {{DOMException}} whose name is "{{AbortError}}" and terminate this algorithm.
 
-                :   is set to {{UserVerificationRequirement/required}}
-                ::  Let |userVerification| be `true`.
+        :   If an |authenticator| becomes available on this platform,
```
agl: (ditto.)

index.bs
```diff
@@ -870,88 +870,91 @@ When this method is invoked, the user agent MUST execute the following algorithm
...
-          1. If <code>|options|.{{PublicKeyCredentialCreationOptions/authenticatorSelection}}.{{authenticatorAttachment}}</code> is
-            [=present|present=] and its value is not equal to |authenticator|'s attachment modality, [=iteration/continue=].
-          1. If <code>|options|.{{PublicKeyCredentialCreationOptions/authenticatorSelection}}.{{requireResidentKey}}</code> is set to
-            `true` and the |authenticator| is not capable of storing a [=Client-Side-Resident Credential Private Key=],
-            [=iteration/continue=].
-          1. If <code>|options|.{{PublicKeyCredentialCreationOptions/authenticatorSelection}}.{{AuthenticatorSelectionCriteria/userVerification}}</code> is
-            set to {{UserVerificationRequirement/required}} and the |authenticator| is not capable of performing [=user
-            verification=], [=iteration/continue=].
+1. [=While=] |lifetimeTimer| has not expired, perform the following actions depending upon |lifetimeTimer|
+    and the state and response [=set/for each=] |authenticator| in |authenticators|:
```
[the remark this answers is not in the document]
emlun: Yep, thanks!

index.bs
emlun: It's not meant to ignore them, but I agree a clarification like that is probably needed.

emlun: Does https://github.com/w3c/webauthn/pull/900/commits/e5696ed6bee6f13d9eb1b2062ab1cce6d2298df7 suffice to clarify this?

index.bs
```diff
@@ -870,88 +870,93 @@ When this method is invoked, the user agent MUST execute the following algorithm
...
+            [=AbortSignal/aborted flag=] is set to true,
+        ::  [=set/For each=] |authenticator| in |issuedRequests| invoke the [=authenticatorCancel=]
+            operation on |authenticator| and [=set/remove=] |authenticator| from |issuedRequests|. Then return a {{DOMException}}
+            whose name is "{{AbortError}}" and terminate this algorithm.
 
-    1. Let |userVerification| be the <dfn>effective user verification requirement for credential creation</dfn>, a Boolean value,
-        as follows. If
-        <code>|options|.{{PublicKeyCredentialCreationOptions/authenticatorSelection}}.{{AuthenticatorSelectionCriteria/userVerification}}</code>
+        :   If an |authenticator| becomes available on this platform,
+        ::  Note: This includes if the |authenticator| was initially available when |lifetimeTimer| started.
```
jcjones: I think this works. I tried writing it a few different ways, particularly to be rid of the Note, but I haven't discovered something I like better.

emlun:
Yes, most of the diff is whitespace changes due to how one large algorithm step has been merged in as a switch case of the following step. I'll post a whitespace-ignoring diff here in case it's useful.

emlun:
`$ git diff --ignore-all-space $(git merge-base master issue-613-hot-plugging-finish) issue-613-hot-plugging-finish`:

```diff --git a/index.bs b/index.bs
index 9cc2c53..5af0577 100644
--- a/index.bs
+++ b/index.bs
@@ -870,17 +870,32 @@ When this method is invoked, the user agent MUST execute the following algorithm
     [=AbortSignal/aborted flag=] is set to true, return a {{DOMException}} whose name is "{{AbortError}}" 
     and terminate this algorithm.
 
+1. Let |issuedRequests| be a new [=ordered set=].
+
+1. Let |authenticators| represent a [=set=] of platform-specific handles, where each value identifies an [=authenticator=]
+    presently available on this platform at a given instant.
+
+    Note: What qualifies an [=authenticator=] as "available" is intentionally unspecified; this is meant to represent how
+    [=authenticators=] can be hot-plugged into (e.g., via USB) or discovered (e.g., via NFC or Bluetooth) by the [=client=] by
+    various mechanisms, or permanently built into the [=client=].
+
 1. Start |lifetimeTimer|.
 
-1. Let |issuedRequests| be a new [=ordered set=].
+1. [=While=] |lifetimeTimer| has not expired, perform the following actions depending upon |lifetimeTimer|,
+    and the state and response [=set/for each=] |authenticator| in |authenticators|:
+    <dl class="switch">
+        :   If |lifetimeTimer| expires,
+        ::  [=set/For each=] |authenticator| in |issuedRequests| invoke the [=authenticatorCancel=] operation on |authenticator|
+            and [=set/remove=] |authenticator| from |issuedRequests|.
 
-1. [=set/For each=] |authenticator| that becomes available on this platform during the lifetime of |lifetimeTimer|, do the
-    following:
+        :   If the <code>|options|.{{CredentialCreationOptions/signal}}</code> is [=present=] and its
+            [=AbortSignal/aborted flag=] is set to true,
+        ::  [=set/For each=] |authenticator| in |issuedRequests| invoke the [=authenticatorCancel=]
+            operation on |authenticator| and [=set/remove=] |authenticator| from |issuedRequests|. Then return a {{DOMException}}
+            whose name is "{{AbortError}}" and terminate this algorithm.
 
-    Issue: The definitions of "lifetime of" and "becomes available" are intended to represent how
-    devices are hot-plugged into (USB) or discovered by (NFC) browsers, and are underspecified.
-    Resolving this with good definitions or some other means will be addressed by resolving
-    [Issue #613](https://github.com/w3c/webauthn/issues/613).
+        :   If an |authenticator| becomes available on this platform,
+        ::  Note: This includes if the |authenticator| was initially available when |lifetimeTimer| started.
 
             1. If <code>|options|.{{PublicKeyCredentialCreationOptions/authenticatorSelection}}</code> is [=present=]:
 
@@ -940,18 +955,8 @@ When this method is invoked, the user agent MUST execute the following algorithm
 
             1. [=set/Append=] |authenticator| to |issuedRequests|.
 
-1. [=While=] |lifetimeTimer| has not expired, perform the following actions depending upon |lifetimeTimer| and responses from the
-    authenticators:
-    <dl class="switch">
-        :   If |lifetimeTimer| expires,
-        ::  [=set/For each=] |authenticator| in |issuedRequests| invoke the [=authenticatorCancel=] operation on |authenticator|
-            and [=set/remove=] |authenticator| from |issuedRequests|.
-
-        :   If the <code>|options|.{{CredentialCreationOptions/signal}}</code> is [=present=] and its 
-            [=AbortSignal/aborted flag=] is set to true, 
-        ::  [=set/For each=] |authenticator| in |issuedRequests| invoke the [=authenticatorCancel=]
-            operation on |authenticator| and [=set/remove=] |authenticator| from |issuedRequests|. Then return a {{DOMException}} 
-            whose name is "{{AbortError}}" and terminate this algorithm.
+        :   If an |authenticator| ceases to be available on this platform,
+        ::  [=set/Remove=] |authenticator| from |issuedRequests|.
 
         :   If any |authenticator| returns a status indicating that the user cancelled the operation,
         ::  1. [=set/Remove=] |authenticator| from |issuedRequests|.
@@ -1200,17 +1205,34 @@ When this method is invoked, the user agent MUST execute the following algorithm
 
 1. Let |issuedRequests| be a new [=ordered set=].
 
-1. Let |authenticator| be a platform-specific handle whose value identifies an [=authenticator=].
+1. Let |savedCredentialIds| be a new [=map=].
+
+1. Let |authenticators| represent a [=set=] of platform-specific handles, where each value identifies an [=authenticator=]
+    presently available on this platform at a given instant.
+
+    Note: What qualifies an [=authenticator=] as "available" is intentionally unspecified; this is meant to represent how
+    [=authenticators=] can be hot-plugged into (e.g., via USB) or discovered (e.g., via NFC or Bluetooth) by the [=client=] by
+    various mechanisms, or permanently built into the [=client=].
 
 1. Start |lifetimeTimer|.
 
-1. [=set/For each=] |authenticator| that becomes available on this platform during the lifetime of
-    |lifetimeTimer|, perform the following steps:
+1. [=While=] |lifetimeTimer| has not expired, perform the following actions depending upon |lifetimeTimer|,
+    and the state and response [=set/for each=] |authenticator| in |authenticators|:
 
-    Issue: The definitions of "lifetime of" and "becomes available" are intended to represent how
-    devices are hot-plugged into (USB) or discovered by (NFC) browsers, and are underspecified.
-    Resolving this with good definitions or some other means will be addressed by resolving
-    [Issue #613](https://github.com/w3c/webauthn/issues/613).
+    <dl class="switch">
+
+        :   If |lifetimeTimer| expires,
+        ::  [=set/For each=] |authenticator| in |issuedRequests| invoke the [=authenticatorCancel=] operation on
+            |authenticator| and [=set/remove=] |authenticator| from |issuedRequests|.
+
+        :   If the {{CredentialRequestOptions/signal}} member is [=present=] and the [=AbortSignal/aborted flag=] is set to
+            true,
+        ::  [=set/For each=] |authenticator| in |issuedRequests| invoke the [=authenticatorCancel=] operation on |authenticator|
+            and [=set/remove=] |authenticator| from |issuedRequests|. Then
+            return a {{DOMException}} whose name is "{{AbortError}}" and terminate this algorithm.
+
+        :   If an |authenticator| becomes available on this platform,
+        ::  Note: This includes if the |authenticator| was initially available when |lifetimeTimer| started.
 
             1. If <code>|options|.{{PublicKeyCredentialRequestOptions/userVerification}}</code> is set to
                 {{UserVerificationRequirement/required}} and the |authenticator| is not capable of performing [=user verification=],
@@ -1260,14 +1282,10 @@ When this method is invoked, the user agent MUST execute the following algorithm
 
                         1. Let |distinctTransports| be a new [=ordered set=].
 
-                1. If |allowCredentialDescriptorList| has exactly one value, let |savedCredentialId| be a new 
-                    {{PublicKeyCredentialDescriptor}}.{{PublicKeyCredentialDescriptor/id}} and set its value to <code>|allowCredentialDescriptorList|[0].id</code>'s
+                        1. If |allowCredentialDescriptorList| has exactly one value, set
+                            <code>|savedCredentialIds|[|authenticator|]</code> to <code>|allowCredentialDescriptorList|[0].id</code>'s
                             value (see [here](#authenticatorGetAssertion-return-values) in [[#op-get-assertion]] for more information).
 
-                Issue: The foregoing step _may_ be incorrect, in that we are attempting to create |savedCredentialId|
-                    here and use it later below, and we do not have a global in which to allocate a place for it. Perhaps this
-                    is good enough?  addendum: [@jcjones feels the above step is likely good enough](https://github.com/w3c/webauthn/pull/665#discussion_r148130187).
-
                         1. [=list/For each=] credential descriptor |C| in |allowCredentialDescriptorList|,
                             [=set/append=] each value, if any, of <code>|C|.{{transports}}</code> to |distinctTransports|.
 
@@ -1304,20 +1322,8 @@ When this method is invoked, the user agent MUST execute the following algorithm
 
             1. [=set/Append=] |authenticator| to |issuedRequests|.
 
-1. [=While=] |lifetimeTimer| has not expired, perform the following actions depending upon |lifetimeTimer|
-    and responses from the authenticators:
-
-    <dl class="switch">
-
-        :   If |lifetimeTimer| expires,
-        ::  [=set/For each=] |authenticator| in |issuedRequests| invoke the [=authenticatorCancel=] operation on
-            |authenticator| and [=set/remove=] |authenticator| from |issuedRequests|.
-
-        :   If the {{CredentialRequestOptions/signal}} member is [=present=] and the [=AbortSignal/aborted flag=] is set to 
-            true,
-        ::  [=set/For each=] |authenticator| in |issuedRequests| invoke the [=authenticatorCancel=] operation on |authenticator|
-            and [=set/remove=] |authenticator| from |issuedRequests|. Then 
-            return a {{DOMException}} whose name is "{{AbortError}}" and terminate this algorithm.
+        :   If an |authenticator| ceases to be available on this platform,
+        ::  [=set/Remove=] |authenticator| from |issuedRequests|.
 
         :   If any |authenticator| returns a status indicating that the user cancelled the operation,
         ::  1. [=set/Remove=] |authenticator| from |issuedRequests|.
@@ -1336,10 +1342,10 @@ When this method is invoked, the user agent MUST execute the following algorithm
             1.  Let <var ignore>assertionCreationData</var> be a [=struct=] whose [=items=] are:
 
                 :   <code><dfn for="assertionCreationData">credentialIdResult</code>
-                ::  If |savedCredentialId| exists, set the value of [=credentialIdResult=] to be the bytes of
-                    |savedCredentialId|. Otherwise, set the value of [=credentialIdResult=] to be the bytes of the
-                    [=credential ID=] returned from the successful [=authenticatorGetAssertion=] operation, as defined in
-                    [[#op-get-assertion]].
+                ::  If <code>|savedCredentialIds|[|authenticator|]</code> exists, set the value of [=credentialIdResult=] to be
+                    the bytes of <code>|savedCredentialIds|[|authenticator|]</code>. Otherwise, set the value of
+                    [=credentialIdResult=] to be the bytes of the [=credential ID=] returned from the successful
+                    [=authenticatorGetAssertion=] operation, as defined in [[#op-get-assertion]].
 
                 :   <code><dfn for="assertionCreationData">clientDataJSONResult</dfn></code>
                 ::  whose value is the bytes of |clientDataJSON|.
```

index.bs
```diff
@@ -1202,124 +1207,125 @@ When this method is invoked, the user agent MUST execute the following algorithm
...
-    1. Let |userVerification| be the <dfn>effective user verification requirement for assertion</dfn>, a Boolean value, as
-        follows. If <code>|options|.{{PublicKeyCredentialRequestOptions/userVerification}}</code>
+        :   If the {{CredentialRequestOptions/signal}} member is [=present=] and the [=AbortSignal/aborted flag=] is set to
+            [TRUE],
+        ::  [=set/For each=] |authenticator| in |issuedRequests| invoke the [=authenticatorCancel=] operation on |authenticator|
+            and [=set/remove=] |authenticator| from |issuedRequests|. Then
+            return a {{DOMException}} whose name is "{{AbortError}}" and terminate this algorithm.
 
-            <dl class="switch">
+        :   If an |authenticator| becomes available on this platform,
```
equalsJeffH: s/ becomes / is, or becomes, /  ?

index.bs
```diff
@@ -1202,124 +1207,125 @@ When this method is invoked, the user agent MUST execute the following algorithm
...
+        ::  Note: This includes if the |authenticator| was initially available when |lifetimeTimer| started.
```
equalsJeffH: suggest for the Note: 
This includes the case where an |authenticator| was available upon |lifetimeTimer| initiation.

index.bs
```diff
@@ -1202,124 +1207,125 @@ When this method is invoked, the user agent MUST execute the following algorithm
...
-                Issue: The foregoing step _may_ be incorrect, in that we are attempting to create |savedCredentialId|
-                    here and use it later below, and we do not have a global in which to allocate a place for it. Perhaps this
-                    is good enough?  addendum: [@jcjones feels the above step is likely good enough](https://github.com/w3c/webauthn/pull/665#discussion_r148130187).
+                        1. Let |distinctTransports| be a new [=ordered set=].
 
-                1. [=list/For each=] credential descriptor |C| in |allowCredentialDescriptorList|,
-                    [=set/append=] each value, if any, of <code>|C|.{{transports}}</code> to |distinctTransports|.
+                        1. If |allowCredentialDescriptorList| has exactly one value, set
+                            <code>|savedCredentialIds|[|authenticator|]</code> to <code>|allowCredentialDescriptorList|[0].id</code>'s
+                            value (see [here](#authenticatorGetAssertion-return-values) in [[#op-get-assertion]] for more information).
```
equalsJeffH: nicely done! :)

index.bs
```diff
@@ -1338,10 +1344,10 @@ When this method is invoked, the user agent MUST execute the following algorithm
...
 
                 :   <code><dfn for="assertionCreationData">credentialIdResult</code>
-                ::  If |savedCredentialId| exists, set the value of [=credentialIdResult=] to be the bytes of
-                    |savedCredentialId|. Otherwise, set the value of [=credentialIdResult=] to be the bytes of the
-                    [=credential ID=] returned from the successful [=authenticatorGetAssertion=] operation, as defined in
-                    [[#op-get-assertion]].
+                ::  If <code>|savedCredentialIds|[|authenticator|]</code> exists, set the value of [=credentialIdResult=] to be
+                    the bytes of <code>|savedCredentialIds|[|authenticator|]</code>. Otherwise, set the value of
+                    [=credentialIdResult=] to be the bytes of the [=credential ID=] returned from the successful
+                    [=authenticatorGetAssertion=] operation, as defined in [[#op-get-assertion]].
```
equalsJeffH: nice.

index.bs
```diff
@@ -872,88 +872,93 @@ When this method is invoked, the user agent MUST execute the following algorithm
...
+        :   If the <code>|options|.{{CredentialCreationOptions/signal}}</code> is [=present=] and its
+            [=AbortSignal/aborted flag=] is set to [TRUE],
+        ::  [=set/For each=] |authenticator| in |issuedRequests| invoke the [=authenticatorCancel=]
+            operation on |authenticator| and [=set/remove=] |authenticator| from |issuedRequests|. Then return a {{DOMException}}
+            whose name is "{{AbortError}}" and terminate this algorithm.
 
-    1. Let |userVerification| be the <dfn>effective user verification requirement for credential creation</dfn>, a Boolean value,
-        as follows. If
-        <code>|options|.{{PublicKeyCredentialCreationOptions/authenticatorSelection}}.{{AuthenticatorSelectionCriteria/userVerification}}</code>
+        :   If an |authenticator| becomes available on this platform,
```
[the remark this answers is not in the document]
emlun: I was considering something like that, to emphasize authenticators that are already available, but I thought it might be confusing since this is all wrapped in a "while" loop. That would technically require something like "is, or becomes, available on this platform and is not present in |issuedRequests|" to prevent issuing an unlimited number of requests to it, but then it would still trigger again after another of the switch cases causes it to be removed from |issuedRequests|.

I guess it's probably obvious enough what we would mean by that, so I'm ready to agree if you're positive it would be an improvement. It just makes my inner mathematician cringe a little bit. :)

index.bs
```diff
@@ -872,88 +872,93 @@ When this method is invoked, the user agent MUST execute the following algorithm
     [=AbortSignal/aborted flag=] is set to [TRUE], return a {{DOMException}} whose name is "{{AbortError}}"
     and terminate this algorithm.
 
-1. Start |lifetimeTimer|.
-
 1. Let |issuedRequests| be a new [=ordered set=].
 
-1. [=set/For each=] |authenticator| that becomes available on this platform during the lifetime of |lifetimeTimer|, do the
-    following:
+1. Let |authenticators| represent a [=set=] of platform-specific handles, where each value identifies an [=authenticator=]
```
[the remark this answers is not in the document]
emlun: The reason I chose the word "represent" here is to hint at how this differs from a "normal" set in that it's an abstract reference to something that may change over time, as opposed to a normal set which only changes when methods are called on it.

Come to think of it, the only purpose of |authenticators| is to be able to then say "[=set/for each=] |authenticator| in |authenticators|". It's not used for anything else, so perhaps it might be better to try to merge its definition into the "for each authenticator" step. That would make the latter pretty unwieldy, though.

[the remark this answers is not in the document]
emlun: That said, I like your suggestion, but I think it's necessary to call out the non-constant nature of this value. How about this?

18. Let |authenticators| represent a value which at any given instant is a [=set=] of platform-specific handles, where each
    [=set/item=] identifies an [=authenticator=] presently available on this platform at that instant.

index.bs
```diff
@@ -872,88 +872,93 @@ When this method is invoked, the user agent MUST execute the following algorithm
...
+            [=AbortSignal/aborted flag=] is set to [TRUE],
+        ::  [=set/For each=] |authenticator| in |issuedRequests| invoke the [=authenticatorCancel=]
+            operation on |authenticator| and [=set/remove=] |authenticator| from |issuedRequests|. Then return a {{DOMException}}
+            whose name is "{{AbortError}}" and terminate this algorithm.
 
-    1. Let |userVerification| be the <dfn>effective user verification requirement for credential creation</dfn>, a Boolean value,
-        as follows. If
-        <code>|options|.{{PublicKeyCredentialCreationOptions/authenticatorSelection}}.{{AuthenticatorSelectionCriteria/userVerification}}</code>
+        :   If an |authenticator| becomes available on this platform,
+        ::  Note: This includes if the |authenticator| was initially available when |lifetimeTimer| started.
```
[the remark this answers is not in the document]
emlun: Great! Thanks!

index.bs
```diff
@@ -872,88 +872,93 @@ When this method is invoked, the user agent MUST execute the following algorithm
...
-    following:
+1. Let |authenticators| represent a [=set=] of platform-specific handles, where each value identifies an [=authenticator=]
+    presently available on this platform at a given instant.
 
-    Issue: The definitions of "lifetime of" and "becomes available" are intended to represent how
-    devices are hot-plugged into (USB) or discovered by (NFC) browsers, and are underspecified.
-    Resolving this with good definitions or some other means will be addressed by resolving
-    [Issue #613](https://github.com/w3c/webauthn/issues/613).
+    Note: What qualifies an [=authenticator=] as "available" is intentionally unspecified; this is meant to represent how
+    [=authenticators=] can be hot-plugged into (e.g., via USB) or discovered (e.g., via NFC or Bluetooth) by the [=client=] by
```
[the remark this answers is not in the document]
emlun: Sure, thanks!

index.bs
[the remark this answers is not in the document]
equalsJeffH: I like the above re-phrased step 18, it seems clearer wrt what's intended, thx.

index.bs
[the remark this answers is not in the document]
equalsJeffH: yeah, hm, let's leave it as-is, thx for cogent explanation.

index.bs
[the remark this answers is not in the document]
equalsJeffH: I think the "traditional hyperlink" is fine, we'll see if any downstream reviewers object, thx.

PR #903, opened 2018-05-10
Add an “internal” AuthenticatorTransport.

agl:
The motivating example is a built-in fingerprint reader. It might be
connected via an I²C bus or the like, but the current
AuthenticatorTransport enumeration cannot express anything like that.

This change adds a catch-all for these internal transports because, from
the point of view of the client, they're all the same: there's nothing
for the user to do if they're not there so no point prompting them.

It also clarifies that the “usb” type means a removable USB device. Some
built-in hardware (esp in laptops) is connected via an internal USB bus,
but a user would not know that and would not want to be prompted like it
was a removable device in that case.

***
<a href="https://pr-preview.s3.amazonaws.com/agl/webauthn/pull/903.html" title="Last updated on May 10, 2018, 11:51 PM GMT (6e4480e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/903/1c3dd46...agl:6e4480e.html" title="Last updated on May 10, 2018, 11:51 PM GMT (6e4480e)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2052,21 +2052,23 @@ parameter to the {{CredentialsContainer/create()}} or {{CredentialsContainer/get
     enum AuthenticatorTransport {
         "usb",
         "nfc",
-        "ble"
+        "ble",
+        "internal"
     };
 </pre>
 
 <div dfn-type="enum-value" dfn-for="AuthenticatorTransport">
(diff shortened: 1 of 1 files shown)


agl:
This is split off from #882, as requested.

equalsJeffH:
LGTM

PR #904, opened 2018-05-11
Remove mention of a timeout for isUserVerifyingPlatformAuthenticatorAvailable

### LINKED ISSUE #575: isUserVerifyingPlatformAuthenticatorAvailable() timeout really 10 minutes?
Is [isUserVerifyingPlatformAuthenticatorAvailable()](https://w3c.github.io/webauthn/#isUserVerifyingPlatformAuthenticatorAvailable) really set as 10 minutes? Why not on the order of a couple of seconds? I don't really see a mechanism for fingerprinting if we wait for a random time within that couple of seconds before returning.

agl:
As discussed on the issue, implementations appear to be converging on
implementing this call without prompting the user and returning
immediately. The wording in this section is loose enough that
implementations that wish to continue using a timeout can find enough
slack to do so, but this change removes the firm suggestion to do so.

Also, align the spacing of “Promise<T>” to match the style used
elsewhere in the W3C specs.

Fixes #575

***
<a href="https://pr-preview.s3.amazonaws.com/agl/webauthn/pull/904.html" title="Last updated on Jun 19, 2018, 4:25 PM GMT (1678bbc)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/904/4fd5dd5...agl:1678bbc.html" title="Last updated on Jun 19, 2018, 4:25 PM GMT (1678bbc)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1470,19 +1470,9 @@ Based on the result, the [=[RP]=] can take further actions to guide the user to 
 
 This method has no arguments and returns a boolean value.
 
-If the promise will return <i>False</i>,
-the [=client=] SHOULD wait a fixed period of time from the invocation of the method before returning <i>False</i>.
-This is done so that callers cannot distinguish between
-the case where the user was unwilling to create a credential using one of the available [=user verification|user-verifying=] [=platform authenticators=] and
-the case where no [=user verification|user-verifying=] [=platform authenticator=] exists.
-Trying to make these cases indistinguishable is done in an attempt to not provide additional information that could be used for fingerprinting.
-A timeout value on the order of 10 minutes is recommended;
-this is enough time for successful user interactions to be performed
-but short enough that the dangling promise will still be resolved in a reasonably timely fashion.
-
 <pre class="idl">
     partial interface PublicKeyCredential {
-        static Promise < boolean > isUserVerifyingPlatformAuthenticatorAvailable();
+        static Promise&lt;boolean&gt; isUserVerifyingPlatformAuthenticatorAvailable();
     };
 </pre>
 


agl:
Labeling as &ldquo;PropRec&rdquo; due to #575 having that label.

emlun:
I opened https://github.com/agl/webauthn/pull/1 which would merge into this PR.

nadalin:
@emlun fix and merge

equalsJeffH:
LGTM per https://github.com/w3c/webauthn/pull/904#issuecomment-393249422

index.bs
```diff
@@ -1461,32 +1461,15 @@ This [=internal method=] accepts no arguments.
...
-the case where no [=user verification|user-verifying=] [=platform authenticator=] exists.
-Trying to make these cases indistinguishable is done in an attempt to not provide additional information that could be used for fingerprinting.
-A timeout value on the order of 10 minutes is recommended;
-this is enough time for successful user interactions to be performed
-but short enough that the dangling promise will still be resolved in a reasonably timely fashion.
-
 <pre class="idl">
     partial interface PublicKeyCredential {
-        static Promise < boolean > isUserVerifyingPlatformAuthenticatorAvailable();
+        static Promise&lt;boolean&gt; isUserVerifyingPlatformAuthenticatorAvailable();
```
equalsJeffH: using literal angle brackets "< ... >"  in `<pre>` does not seem to be an issue in bikeshed, eg our use of `sequence<...>` all over the place, so perhaps the `&lt` and `&gt` should be literal angle brackets?  I'm ok with removing the gratuitous whitespace.

equalsJeffH: ah, perhaps an issue here is the use of `<pre>` block rather than `<xmp>`, see https://tabatkins.github.io/bikeshed/#xmp

see also #943

nadalin:
@agl Please merge

index.bs
agl: `pre` got changed to `xmp` throughout so I've switched to `<` and `>` here. Thanks.

Issue #905, opened 2018-05-11
Allow clients to stop the `get` flow when certain conditions are met


agl:
Christiaan requests the following:

If an RP sends down a list of credentialIds to the platform, and:

1. The RP specifically asked only for built-in credentials (we're still debating how an RP would do this), and none matches any built-in credential, _or_
1. The user explicitly taps the &ldquo;wrong&rdquo; external authenticator which doesn't house any of the credential IDs that were sent down by the RP, _or_
1. The user explicitly cancels out of the flow that tells them to insert an external authenticator since the credential IDs couldn't be found on built-in authenticators

Then the client should render a dialog that says &ldquo;no credential found&rdquo; and, upon the user dismissing the dialog box, the RP should immediately be notified that no matching credentialIds could be found.

agl:
emlun is knee-deep in this part of the spec in #900 so I'll try building upon that.

emlun:
(2) is what's discussed in #863 , right?

emlun:
As noted in https://github.com/w3c/webauthn/pull/906#pullrequestreview-119810884, I have some reservations regarding the specific implementations of (1) and (3). We need to be very careful so we don't accidentally make them open up leaks of identifying information (#184, #204).

nadalin:
@agl Will discuss @ FIDO Plenary

equalsJeffH:
@agl wrote:
> The RP specifically asked only for built-in credentials (we're still debating how an RP would do this) ....

by "built-in credentials" you mean "[client-side resident private key](https://w3c.github.io/webauthn/#client-side-resident-credential-private-key)" (which ought to define a "client-side resident credential" but does not as yet), yes?

Perhaps that ought to be a specific issue, e.g.: "allow RPs to specifically ask for only client-side resident credentials" -- and the

equalsJeffH:
ah, hm, we seem to have some terminology vagueness here.

[...]

PR #906, opened 2018-05-11
Add two abort paths for getting an assertion

agl:
This change handles cases (1) and (3) of issue #905. Specifically it augments the process for getting an assertion to allow a client to abort based on (optional) UI, and to abort (after user notification) when it becomes aware that the process cannot be satisfied.

Updates #905

***
<a href="https://pr-preview.s3.amazonaws.com/agl/webauthn/pull/906.html" title="Last updated on Jun 19, 2018, 5:55 PM GMT (1e44ebd)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/906/4fd5dd5...agl:1e44ebd.html" title="Last updated on Jun 19, 2018, 5:55 PM GMT (1e44ebd)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1223,11 +1223,20 @@ When this method is invoked, the user agent MUST execute the following algorithm
         ::  [=set/For each=] |authenticator| in |issuedRequests| invoke the [=authenticatorCancel=] operation on
             |authenticator| and [=set/remove=] |authenticator| from |issuedRequests|.
 
+        :   If the user exercises a user-interface option to cancel the process,
+        ::  [=set/For each=] |authenticator| in |issuedRequests| invoke the [=authenticatorCancel=] operation on |authenticator|
+            and [=set/remove=] |authenticator| from |issuedRequests|. Return a {{DOMException}} whose name is "{{NotAllowedError}}".
+
         :   If the {{CredentialRequestOptions/signal}} member is [=present=] and the [=AbortSignal/aborted flag=] is set to
             true,
         ::  [=set/For each=] |authenticator| in |issuedRequests| invoke the [=authenticatorCancel=] operation on |authenticator|
             and [=set/remove=] |authenticator| from |issuedRequests|. Then
             return a {{DOMException}} whose name is "{{AbortError}}" and terminate this algorithm.
+
+        :   If |issuedRequests| is empty, <code>|options|.{{PublicKeyCredentialRequestOptions/allowCredentials}}</code> is not empty, and no |authenticator| will become available for any [=public key credentials=] therein.
+        ::  Indicate to the user that the credential could not be found. When the user acknowledges the dialog, or once |lifetimeTimer| expires, return a {{DOMException}} whose name is "{{NotAllowedError}}".
+
+            Note: One way in which the platform may determine that no |authenticator| will become available is by using the <code>{{transports}}</code> members of <code>|options|.{{PublicKeyCredentialRequestOptions/allowCredentials}}</code>. For example, if all credentials only list {{AuthenticatorTransport/internal}}, but all internal |authenticator|s have been tried, then there is no possibility of satisfying the request. Alternatively, all credentials may require transports that the platform does not support.
 
         :   If an |authenticator| becomes available on this platform,
         ::  1. If <code>|options|.{{PublicKeyCredentialRequestOptions/userVerification}}</code> is set to
--- a/index.bs
+++ b/index.bs
@@ -1274,9 +1274,9 @@ When this method is invoked, the user agent MUST execute the following algorithm
             return a {{DOMException}} whose name is "{{AbortError}}" and terminate this algorithm.
 
         :   If |issuedRequests| is empty, <code>|options|.{{PublicKeyCredentialRequestOptions/allowCredentials}}</code> is not empty, and no |authenticator| will become available for any [=public key credentials=] therein,
-        ::  Indicate to the user that the credential could not be found. When the user acknowledges the dialog, or once |lifetimeTimer| expires, return a {{DOMException}} whose name is "{{NotAllowedError}}".
+        ::  Indicate to the user that no eligible credential could be found. When the user acknowledges the dialog, return a {{DOMException}} whose name is "{{NotAllowedError}}".
 
-            Note: One way in which the platform may determine that no |authenticator| will become available is by using the <code>{{transports}}</code> members of <code>|options|.{{PublicKeyCredentialRequestOptions/allowCredentials}}</code>. For example, if all credentials only list {{AuthenticatorTransport/internal}}, but all internal |authenticator|s have been tried, then there is no possibility of satisfying the request. Alternatively, all credentials may require transports that the platform does not support.
+            Note: One way a platform can determine that no |authenticator| will become available is by examining the <code>{{transports}}</code> members of the present <code>{{PublicKeyCredentialDescriptor}}</code> [=list/items=] of <code>|options|.{{PublicKeyCredentialRequestOptions/allowCredentials}}</code>, if any. For example, if all <code>{{PublicKeyCredentialDescriptor}}</code> [=list/items=] list only <code>{{AuthenticatorTransport/internal}}</code>, but all internal |authenticator|s have been tried, then there is no possibility of satisfying the request. Alternatively, all <code>{{PublicKeyCredentialDescriptor}}</code> [=list/items=] may list <code>{{transports}}</code> that the platform does not support.
 
         :   If an |authenticator| becomes available on this platform,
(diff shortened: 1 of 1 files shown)


nadalin:
@agl discuss at FIDO Plenary

agl:
> Can't this be covered by the existing AbortError path?

Do you mean that it could be merged with the subsequent case because the actions are almost identical? I agree and the only reason that I included a separate (and very similar) case was because the disjunction of the two conditions seemed excessively complex when written. Totally don't care about which way to structure this, however.

(Although I do thing there is value in returning a `NotAllowedError` in this case, see below:)

> I think it must be made much clearer that this requires active and informed consent to reveal credential existence to the RP.

Why does this disclose credential existence? I think all the following situations should look the same to the RP:
1. The user activates some UI to abort the operation. (For example, Firefox currently has a &ldquo;Cancel&rdquo; button in a bubble that does this.)
1. The user presses the wrong authenticator. (The spec doesn't allow this, but both Chrome and Firefox flash all authenticators and abort the operation if the wrong one is pressed.)
1. An internal authenticator matches, but the user activates some system UI to abort using it. (For example, triggering Touch ID on iOS shows _system_ UI that allows the operation to be canceled.)
1. (_This case_) The client concludes that the operation is doomed and lets the user press the same sort of cancel button as in case one.

Now, in practice Firefox and Chrome are returning different errors for many of these cases and I think that should probably be fixed. But, assuming that we manage that, I don't see that we're leaking information about the presence of the credential here.

emlun:
This PR branches from #900, and currently adds only one commit on top of it. The diff for the changes unique to this PR can be seen here: https://github.com/w3c/webauthn/pull/906/commits/ed3abec313375cb38a11e964be917306c95308d8

emlun:
@agl 

Sorry, you're right - credential existence is only revealed if some algorithm path results in immediate response to the RP. If all paths lead to a user prompt, there should be no issue. I retract my objection.

>Do you mean that it could be merged with the subsequent case because the actions are almost identical?

Yes, I mean to either merge them or somehow say that "If the user exercises a user-interface option to cancel the process" is just another way to set the abort signal. I was imagining that as the "obvious" way for the browser would implement it, but I'll admit I don't really know how those signals work or are usually used. But it might be worthwhile to mention that explicitly, and I see how doing both in the subsequent case would probably be clumsy. If there's a good way to cover both with the same case then that would be preferable, but I'm fine either way.

agl:
> Yes, I mean to either merge them or somehow say that "If the user exercises a user-interface option to cancel the process" is just another way to set the abort signal.

I have left it as a separate item, purely because it returns a `NotAllowedError` rather than an `AbortedError`. I like that because I don't see why the RP should know that the user decided to abort the operation via UI, vs one of the other fatal paths, and worry that there might be a small information leak otherwise.

equalsJeffH:
I find side-by-side diffs helpful for gnarly alg changes such as this, here's the one I just gen'd for this: http://jeffhodges.org/doc/diff/diff-index-agl-issue905-0244f7c--from---master-3c5e383.pdf

nadalin:
@agl Please merge

agl:
(Will request review from @akshayku once I've merged this with master.)

agl:
Have merged with master.

index.bs
```diff
@@ -1223,12 +1223,21 @@ When this method is invoked, the user agent MUST execute the following algorithm
...
+            and [=set/remove=] |authenticator| from |issuedRequests|. Return a {{DOMException}} whose name is "{{NotAllowedError}}".
+
         :   If the {{CredentialRequestOptions/signal}} member is [=present=] and the [=AbortSignal/aborted flag=] is set to
             [TRUE],
         ::  [=set/For each=] |authenticator| in |issuedRequests| invoke the [=authenticatorCancel=] operation on |authenticator|
             and [=set/remove=] |authenticator| from |issuedRequests|. Then
             return a {{DOMException}} whose name is "{{AbortError}}" and terminate this algorithm.
 
+        :   If |issuedRequests| is empty, <code>|options|.{{PublicKeyCredentialRequestOptions/allowCredentials}}</code> is not empty, and no |authenticator| will become available for any [=public key credentials=] therein,
+        ::  Indicate to the user that the credential could not be found. When the user acknowledges the dialog, or once |lifetimeTimer| expires, return a {{DOMException}} whose name is "{{NotAllowedError}}".
```
s/ the credential / appropriate credential(s) /

Is not the |lifetimeTimer| expiry handled by the above **If lifetimeTimer expires** switch branch?

index.bs
```diff
@@ -1223,12 +1223,21 @@ When this method is invoked, the user agent MUST execute the following algorithm
...
+
+            Note: One way in which the platform may determine that no |authenticator| will become available is by using the <code>{{transports}}</code> members of <code>|options|.{{PublicKeyCredentialRequestOptions/allowCredentials}}</code>. For example, if all credentials only list {{AuthenticatorTransport/internal}}, but all internal |authenticator|s have been tried, then there is no possibility of satisfying the request. Alternatively, all credentials may require transports that the platform does not support.
```
Note: One way a platform can determine that no |authenticator| will become available is by examining the <code>{{transports}}</code>  members of the present {{PublicKeyCredentialDescriptor}} [=list/items=] of <code>|options|.{{PublicKeyCredentialRequestOptions/allowCredentials}}</code>, if any. For example, if all {{PublicKeyCredentialDescriptor}} [=list/items=] list only  {{AuthenticatorTransport/internal}}, but all internal |authenticator|s have been tried, then there is no possibility of satisfying the request. Alternatively, all {{PublicKeyCredentialDescriptor}} [=list/items=] may list <code>{{transports}}</code> that the platform does not support.

index.bs
emlun: Or perhaps:
s/the credential could not be found/no eligible credential could be found/
?

I'm inclined to agree that the |lifetimeTimer| expiry is handled by that other switch case.

index.bs
```diff
@@ -1223,12 +1223,21 @@ When this method is invoked, the user agent MUST execute the following algorithm
         ::  [=set/For each=] |authenticator| in |issuedRequests| invoke the [=authenticatorCancel=] operation on
             |authenticator| and [=set/remove=] |authenticator| from |issuedRequests|.
 
+        :   If the user exercises a user-interface option to cancel the process,
+        ::  [=set/For each=] |authenticator| in |issuedRequests| invoke the [=authenticatorCancel=] operation on |authenticator|
+            and [=set/remove=] |authenticator| from |issuedRequests|. Return a {{DOMException}} whose name is "{{NotAllowedError}}".
+
```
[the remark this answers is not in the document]
agl: I think one could say that a user-interface element is wired up to the abort signal, but this makes it more explicit. (Although [this section](https://www.w3.org/TR/webauthn/#abortoperation) clearly suggests that browser UI actions can trigger the abort signal.) So I could go either way: like this, or somewhere else say that other UI actions (like a button press) can also trigger the abort.

As for creating a credential: I don't disagree, but the bug was focused on the get flow.

index.bs
```diff
@@ -1223,12 +1223,21 @@ When this method is invoked, the user agent MUST execute the following algorithm
...
         :   If the {{CredentialRequestOptions/signal}} member is [=present=] and the [=AbortSignal/aborted flag=] is set to
             [TRUE],
         ::  [=set/For each=] |authenticator| in |issuedRequests| invoke the [=authenticatorCancel=] operation on |authenticator|
             and [=set/remove=] |authenticator| from |issuedRequests|. Then
             return a {{DOMException}} whose name is "{{AbortError}}" and terminate this algorithm.
 
+        :   If |issuedRequests| is empty, <code>|options|.{{PublicKeyCredentialRequestOptions/allowCredentials}}</code> is not empty, and no |authenticator| will become available for any [=public key credentials=] therein,
```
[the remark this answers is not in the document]
agl: I think that makes ambiguous what should happen when `allowCredentials` is empty.

index.bs

index.bs

index.bs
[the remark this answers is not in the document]

index.bs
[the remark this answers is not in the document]
equalsJeffH: if y'all think this is good enough as-is, then OK.  wrt applying it to creating a credential, if it ought to be fixed and is not done as part of this pr (fixing it as part of this PR is fine by me), then I spose an issue regarding it ought to get filed.

nadalin:
@agl If all ok, please merge

equalsJeffH:
@agl [does not disagree](https://github.com/w3c/webauthn/pull/906#discussion_r196517535) that the new switch branch in new lines 1266-1269 ought to also be applied to the [#createCredential](https://w3c.github.io/webauthn/#createCredential) alg  -- I submitted issue #985.

PR #909, opened 2018-05-15
Add registration/authentication extensions for cloud-assisted BLE

kpaulh:
This is a counterpart to https://github.com/fido-alliance/fido-2-specs/pull/529/, the FIDO2 cloud-assisted BLE PR. We're not quite sure where the extension should live - FIDO2 or WebAuthN. We're leaning towards WebAuthN, since it is relevant to RPs, but hoping others have some thoughts.

Also, this PR currently has references to sections in the FIDO2 spec that I can either replace with complete explanations or update links to once they're available in the FIDO spec. We figured we would get eyes on it now and clean up along the way.

Note that since this is the first registration extension, this PR also adds RegistrationExtensionsClientInputs/Outputs.

***
<a href="https://pr-preview.s3.amazonaws.com/kpaulh/webauthn/pull/909.html" title="Last updated on May 1, 2019, 11:44 PM UTC (63be77d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/909/5d11e62...kpaulh:63be77d.html" title="Last updated on May 1, 2019, 11:44 PM UTC (63be77d)">Diff</a>

PR #910, opened 2018-05-15
Trim unnecessary step from appId extension

kpaulh:
It's not actually necessary to specify that the extension is only valid for getAssertion because it's already restricted by being part of AuthenticatorExtensionClientInputs.

***
<a href="https://pr-preview.s3.amazonaws.com/kpaulh/webauthn/pull/910.html" title="Last updated on May 30, 2018, 5:41 PM GMT (8e76351)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/910/1c3dd46...kpaulh:8e76351.html" title="Last updated on May 30, 2018, 5:41 PM GMT (8e76351)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -3852,10 +3852,7 @@ JavaScript APIs.
     </xmp>
 
 : Client extension processing
-::  1. If present in a {{CredentialsContainer/create()}} call, return a
-        "{{NotSupportedError}}" {{DOMException}}—this extension is only valid when
-        requesting an assertion.
-    1. Let |facetId| be the result of passing the caller's [=origin=] to the
+::  1. Let |facetId| be the result of passing the caller's [=origin=] to the
         FIDO algorithm for [=determining the FacetID of a calling application=].
     1. Let |appId| be the extension input.
     1. Pass |facetId| and |appId| to the FIDO algorithm for [=determining if a
--- a/index.bs
+++ b/index.bs
@@ -4007,10 +4007,9 @@ creation.
     </xmp>
 
 : Client extension processing
-:: This extension can only be used during {{CredentialsContainer/create()}}. If the client supports the Authenticator Selection
-    Extension, it MUST use the first available authenticator whose AAGUID is present in the {{AuthenticatorSelectionList}}. If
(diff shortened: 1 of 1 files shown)


nadalin:
@kpaulh Will review @ FIDO Plenary

nadalin:
See #909

kpaulh:
Talked over this more with agl@ and uncovered some confusion.

The wording under https://w3c.github.io/webauthn/#extensions refers to 'registration extensions' and 'authentication extensions'. It's unclear, then, whether 'AuthenticationClientExtensionInputs' refers to 'authentication extensions' as distinct from 'registration extensions', or authentication as in "web authentication".

1) if the former, this implies we should also have 'RegistrationClientExtensionInputs'. In which case, this PR is still valid, the addition of RegistrationClientExtensionInputs in #909 is valid, and other registration extensions should be changed to use it.

2) if the latter, it might be better to call the input "ClientExtensionInputs". In which case, this PR is unnecessary, I'll adjust #909 to use 'AuthenticationClientExtensionInputs', and we should consider if there's any purpose in having wording that distinguishes 'registration extensions' from 'authentication extensions'.

emlun:
Yeah, that's what I was thinking too - [`AuthenticationClientExtensionInputs`][aceip] is used in both [create()][ref-create] and [get()][ref-get]. I think you're right this is intended to be "authentication" as in "Web Authentication".

I agree it would be nice to rename the dictionaries, but would that be too big a breaking change?

[aceip]: https://w3c.github.io/webauthn/#dictdef-authenticationextensionsclientinputs
[ref-get]: https://w3c.github.io/webauthn/#ref-for-dictdef-authenticationextensionsclientinputs%E2%91%A1
[ref-create]: https://w3c.github.io/webauthn/#ref-for-dictdef-authenticationextensionsclientinputs

emlun:
>I found the processing you mentioned - Step 12 of makeCredential says [...]

Great! Then I let's do this.

kpaulh:
Done. And confirmed with @leshi that it's fine to ignore the appID extension and continue when the appID doesn't validate.

nadalin:
@kpaulh Please merge

PR #918, opened 2018-05-29
Create draft-jones-webauthn-secp256k1 to register secp256k1 curve and algorithm identifiers

selfissued:


selfissued:
This is in the optional algorithms set in the FIDO Server document - per https://github.com/fido-alliance/fido-2-specs/pull/537.  I'd taken on the action item earlier to register all the algorithms needed for WebAuthn and FIDO.  draft-jones-webauthn-cose-algorithms is also part of that effort.  This is another step in fulfilling that action item.

This is in a separate draft because if we make it an RFC, we can get one of the small COSE algorithm identifiers for it - versus the larger ones that non-RFCs can get.  Once checked in, I will attempt to obtain Area Director sponsorship for it.

There are other communities also wanting this algorithm registered, including Blockchain folks.  I'm doing all this under the WebAuthn umbrella as a matter of convenience - something that was discussed on at least one of the calls.

selfissued:
Part of the reason this is relevant to the W3C is that, for better or for worse, some of the projects of the W3C Verifiable Credentials interest group use the secp256k1 curve, but currently doing so using non-standard crypto formats and identifiers.  During the recent Internet Identity Workshop, I promised some of the interest group members that I'd register the identifiers necessary for them to use standard JOSE crypto representations.  The COSE registrations are being requested by FIDO.

PR #935, opened 2018-06-06
Per Credential Signature Counters

### LINKED ISSUE #871: Delete per RP ID Signature counters
Section 6.1.1 mentions:  
  
> (The authenticator) should implement per RP-ID signature counters. This prevents the signature counter value from being shared between Relying Parties and being possibly employed as a correlation handle for the user. Authenticators may implement a global signature counter, i.e., on a per-authenticator basis, but this is less privacy-friendly for users.
  
A global signature counter may hypothetically allow several websites to identify a given user even if said user uses different IP addresses and other means of covering his tracks.  
The per RP-ID suggestion is therefore a great improvement. It however does (IMHO) only offers a "half way" compromise as it would allow a given RP to identify the different aliases of a given user for a given RP through his per RP-ID counter.  

Section 6.2.2, step 10 then mentions the possibility of a "per credential signature counter" which does offer the best solution regarding privacy-concerned users.  

The question I therefore would like to ask is : why offer the possibility to authenticators to implement per RP-ID credential counters when the privacy friendly and non-privacy friendly alternatives are available?  
I would then suggest adding a counter to the Public Key Credential Source structure, especially given the fact that "an authenticator will never contain more than one credential for a given Relying Party under the same user handle." (from 4. Terminology), which (if I'm not mistaken) also means that a given Public Key Credential Source will only be used for a single credential.

I'm still digging into the spec so I apologize if I may have missed some important info.

akshayku:
This fixes #871

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/pull/935.html" title="Last updated on Jun 8, 2018, 6:30 PM GMT (7958ffc)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/935/0f5b3a8...7958ffc.html" title="Last updated on Jun 8, 2018, 6:30 PM GMT (7958ffc)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2347,12 +2347,12 @@ An [=[RP]=] stores the [=signature counter=] of the most recent [=authenticatorG
 Detecting a [=signature counter=]  mismatch does not indicate whether the current operation was performed by a cloned authenticator or the original authenticator.  [=[RPS]=] should address this situation appropriately relative to their individual situations, i.e., their risk tolerance. 
 
 Authenticators:
-- should implement per credential [=signature counters=].  This prevents the 
+- SHOULD implement per credential [=signature counters=].  This prevents the 
     [=signature counter=] value from being shared between [=[RPS]=] and being possibly employed 
     as a correlation handle for the user. Authenticators may implement a global [=signature counter=], 
     i.e., on a per-authenticator basis, but this is less privacy-friendly for users. 
 
-- should ensure that the [=signature counter=] value does not 
+- SHOULD ensure that the [=signature counter=] value does not 
     accidentally decrease  (e.g., due to hardware failures).
 
 
--- a/index.bs
+++ b/index.bs
@@ -2335,7 +2335,7 @@ follows.
 
 ### <dfn>Signature Counter</dfn> Considerations ### {#sign-counter}
 
-Authenticators MUST implement a [=signature counter=] feature. The [=signature counter=] is incremented for each successful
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -2335,7 +2335,7 @@ follows.
 
 ### <dfn>Signature Counter</dfn> Considerations ### {#sign-counter}
 
-Authenticators MUST implement a [=signature counter=] feature. The [=signature counter=] is incremented for each successful
+Authenticators SHOULD implement a [=signature counter=] feature. The [=signature counter=] is incremented for each successful
```
akshayku: Changing this to SHOULD as authenticators which are purely server based (non-resident), can't store anything. We support such kind of authenticators in FIDO2.

index.bs
```diff
@@ -2347,7 +2347,7 @@ An [=[RP]=] stores the [=signature counter=] of the most recent [=authenticatorG
 Detecting a [=signature counter=]  mismatch does not indicate whether the current operation was performed by a cloned authenticator or the original authenticator.  [=[RPS]=] should address this situation appropriately relative to their individual situations, i.e., their risk tolerance. 
 
 Authenticators:
-- should implement per-[=RP ID=] [=signature counters=].  This prevents the 
+- should implement per credential [=signature counters=].  This prevents the
```
emlun: Please change "should" to "SHOULD" here, and in the subsequent list item.

PR #940, opened 2018-06-07
PR #900 post-merge review changes

### LINKED ISSUE #613: Allow hot-plugged authenticators?

emlun:
This PR is for after-the-fact review changes to PR #900 requested by @equalsJeffH in https://github.com/w3c/webauthn/pull/900#pullrequestreview-126523183 . This fixes #613.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/pull/940.html" title="Last updated on Jun 7, 2018, 3:36 PM GMT (8b6b7e1)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/940/b0ca15f...8b6b7e1.html" title="Last updated on Jun 7, 2018, 3:36 PM GMT (8b6b7e1)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -874,8 +874,8 @@ When this method is invoked, the user agent MUST execute the following algorithm
 
 1. Let |issuedRequests| be a new [=ordered set=].
 
-1. Let |authenticators| represent a [=set=] of platform-specific handles, where each value identifies an [=authenticator=]
-    presently available on this platform at a given instant.
+1. Let |authenticators| represent a value which at any given instant is a [=set=] of platform-specific handles, where each
+    [=set/item=] identifies an [=authenticator=] presently available on this platform at that instant.
 
     Note: What qualifies an [=authenticator=] as "available" is intentionally unspecified; this is meant to represent how
     [=authenticators=] can be <a href="https://en.wikipedia.org/w/index.php?title=Hot_plug">hot-plugged</a> into (e.g., via USB)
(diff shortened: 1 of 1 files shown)


emlun:
Thanks @equalsJeffH for the review! I also added the change rephrasing the |authenticators| definition steps; that commit was not included in your https://github.com/w3c/webauthn/pull/940#pullrequestreview-126831731 but it shows up earlier in the event log because I wrote the commit ahead of time but didn't push it.

Since these changes would have belonged in #900, I'll regard this as ready to merge once @equalsJeffH signs off on commit 8b6b7e1ded0f64806da8a739f4879bc5edfc7c6f . I'll go ahead and perform the merge if nobody objects before 2018-06-11T08:00Z (and if nobody else does it before then :) ).

PR #941, opened 2018-06-07
Specify extension types in listing instead of prose

emlun:
I think this makes it easier to identify the type(s) of the extensions. Since the extension types are referenced in [create()][create] and [get()][get], it seems preferable to specify them clearly and formally instead of in the prose description as is currently done.

The term "extension type" is currently left without a formal definition, though. I'm thinking one probably isn't necessary, but I can add one if it is.

This PR currently has no accompanying issue.

[create]: https://w3c.github.io/webauthn/#ref-for-registration-extension
[get]: https://w3c.github.io/webauthn/#ref-for-authentication-extension

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/pull/941.html" title="Last updated on Jun 20, 2018, 4:42 PM GMT (b971c66)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/941/b0ca15f...b971c66.html" title="Last updated on Jun 20, 2018, 4:42 PM GMT (b971c66)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -3882,7 +3882,7 @@ These are RECOMMENDED for implementation by user agents targeting broad interope
 
 ## FIDO <dfn>AppID</dfn> Extension (appid) ## {#sctn-appid-extension}
 
-This [=client extension=] allows [=[RPS]=] that have previously registered a
+This extension allows [=[RPS]=] that have previously registered a
 credential using the legacy FIDO JavaScript APIs to request an [=assertion=]. The
 FIDO APIs use an alternative identifier for [=relying parties=] called an |AppID|
 [[FIDO-APPID]], and any credentials created using those APIs will be bound to
@@ -3895,6 +3895,9 @@ JavaScript APIs.
 
 : Extension identifier
 :: `appid`
+
+: Extension type
+:: [=authentication extension|Authentication=]
 
 : Client extension input
 :: A single USVString specifying a FIDO |AppID|.
@@ -3939,11 +3942,14 @@ JavaScript APIs.
 
 ## Simple Transaction Authorization Extension (txAuthSimple) ## {#sctn-simple-txauth-extension}
 
-This [=registration extension=] and [=authentication extension=] allows for a simple form of transaction authorization. A
+This extension allows for a simple form of transaction authorization. A
 [=[RP]=] can specify a prompt string, intended for display on a trusted device on the authenticator.
 
 : Extension identifier
 :: `txAuthSimple`
+
+: Extension type
+:: [=registration extension|Registration=] and [=authentication extension|authentication=]
 
 : Client extension input
 :: A single USVString prompt.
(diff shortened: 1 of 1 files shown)


emlun:
In reviewing the rendered diff myself, I realize there are two orthogonal "type" concepts for extensions: the registration/authentication type (which this PR is about) and the client/authenticator type. Perhaps a different term should be used to pre-empt confusion between the two, although it seems they're currently never referred to as "types" in the spec.

equalsJeffH:
>  realize there are two orthogonal "type" concepts for extensions .... although it seems they're currently never referred to as "types" in the spec.

yes, perhaps we should refer to them as "extension characteristics", where "targeted component" and "related operation" are such characteristics.

targeted component characteristics are client or authenticator.
related operation characteristics are registration or authentication.

derive authnr characteristic-based naming from them...here's perhaps one way to think of it:
```
---------------------------------------------------------------------------
webauthn extension characteristics and naming taxonomy
---------------------------------------------------------------------------
      \            |                            |
       \  targeted |                            |
related \ component|       client               |       authenticator
operation\         |                            |
-------------------+-------------------------------------------------------
                   |                            |
  registration     |  reg-op client extension   |  reg-op authnr extension
                   |                            |
-------------------|----------------------------+--------------------------
                   |                            |
  authentication   | authn-op client extension  | authn-op authnr extension
                   |                            |
                   |                            |
---------------------------------------------------------------------------

```

emlun:
I like that idea, but on the other hand it might be better to just leave this as is instead of introducing new concepts for this relatively small gain. What do you think?

Or perhaps we could just change the title in the listings from "Extension type" to "Operation applicability" or so (note that an extension can be both a registration extension and an authentication extension), without specifically defining the term.

equalsJeffH:
hm, [S 9. WebAuthn Extensions](https://w3c.github.io/webauthn/#extensions), defines "every extension" as being a [client extension](file:///Users/jehodges/documents/work/standards/W3C/webauthn/index-extension-types-9006267.html#client-extension), so the taxonomy above is inaccurate. [oops](http://sephora.i.lithium.com/t5/image/serverpage/image-id/278062i078E938EE74BEE55/image-size/original?v=1.0&px=-1).

> perhaps we could just change the title in the listings from "Extension type" to "Operation applicability" or so (note that an extension can be both a registration extension and an authentication extension), without specifically defining the term.

that's fine by me.

nadalin:
@selfissued  Please review

nadalin:
@selfissued please review

index.bs
```diff
@@ -3939,12 +3942,15 @@ JavaScript APIs.
...
 
-This [=registration extension=] and [=authentication extension=] allows for a simple form of transaction authorization. A
+This extension allows for a simple form of transaction authorization. A
 [=[RP]=] can specify a prompt string, intended for display on a trusted device on the authenticator.
 
 : Extension identifier
 :: `txAuthSimple`
 
+: Operation applicability
+:: [=registration extension|Registration=] and [=authentication extension|authentication=]
```
selfissued: I would always capitalize both Registration and Authentication when used in this context.  So please change [=authentication extension|authentication=] to [=authentication extension|Authentication=] - both here and in all other occurrences.

emlun: Done, thanks!

PR #951, opened 2018-06-14
Fix #593: employ PRECIS RFC8264 et al for 'name'-ish domstring values

### LINKED ISSUE #593: Display name content rules?

equalsJeffH:
I suggest this PR as an alternative to PR #878.  In my review of the latter, and of PRECIS RFCs 8264, 8265, 8266, it seems that we ought to provide more fine-grained guidance regarding employing PRECIS to "corral" RP- and user-supplied "name"-ish strings.  See discussion in both issue #593 and pr #878.

cc @stpeter

fixes #593

This does not (as yet) address #593 because the latter issue touches on two aspects to the overall issue:
1. corralling the unicode content of the "name"-ish domstrings
2. providing implementer guidance regarding how to display/present these string values in order to mitigate effects of possibly malicious string content.

Originally, this PR addressed only item 1, above. 
Now (11-Jul-2018) it also addresses item 2, see commit 4ddc3c0 and https://github.com/w3c/webauthn/pull/951#issuecomment-404334287.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/pull/951.html" title="Last updated on Jul 11, 2018, 10:51 PM GMT (4ddc3c0)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/951/ca80875...4ddc3c0.html" title="Last updated on Jul 11, 2018, 10:51 PM GMT (4ddc3c0)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1793,8 +1793,8 @@ credential.
 <div dfn-type="dict-member" dfn-for="PublicKeyCredentialUserEntity">
     :   <dfn>id</dfn>
     ::  The [=user handle=] of the user account entity. To ensure secure operation, authentication and authorization
-        decisions MUST be made on the basis of this {{PublicKeyCredentialEntity/id}} member,  not the
-        {{PublicKeyCredentialEntity/displayName}} nor {{PublicKeyCredentialEntity/name}} members. See [[!RFC8266]]
+        decisions MUST be made on the basis of this {{PublicKeyCredentialUserEntity/id}} member,  not the
+        {{PublicKeyCredentialUserEntity/displayName}} nor {{PublicKeyCredentialEntity/name}} members. See [[!RFC8266]]
         Section 6.1.
 
     :   <dfn>displayName</dfn>
--- a/index.bs
+++ b/index.bs
@@ -1724,12 +1724,12 @@ associated.
           - When inherited by {{PublicKeyCredentialRpEntity}} it is a [=human palatability|human-palatable=] identifier for the [=[RP]=], intended only
             for display. For example, "ACME Corporation", "Wonderful Widgets, Inc." or "ОАО Примертех".
 
-              - [=[RPS]=] SHOULD perform enforcement, as prescribed in [[!RFC8266]]
-                Section 2.3 for the Nickname Profile of the PRECIS FreeformClass [[!RFC8264]],
+              - [=[RPS]=] SHOULD perform enforcement, as prescribed in Section 2.3 of
+                [[!RFC8266]] for the Nickname Profile of the PRECIS FreeformClass [[!RFC8264]],
                 when setting {{PublicKeyCredentialEntity/name}}'s value, or displaying the value to the user.
 
-              - [=Clients=] MUST perform enforcement, as prescribed in [[!RFC8266]]
-                Section 2.3 for the Nickname Profile of the PRECIS FreeformClass [[!RFC8264]],
+              - [=Clients=] MUST perform enforcement, as prescribed in Section 2.3 of
+                [[!RFC8266]] for the Nickname Profile of the PRECIS FreeformClass [[!RFC8264]],
                 on {{PublicKeyCredentialEntity/name}}'s value prior to displaying the value to the user or
                 including the value as a parameter of the [=authenticatorMakeCredential=] operation.
 
@@ -1739,12 +1739,12 @@ associated.
             or "+14255551234".
 
               - The [=[RP]=] MAY let the user choose this value. The [=[RP]=] SHOULD perform enforcement,
-                as prescribed in [[!RFC8265]] Section 3.4.3 for the UsernameCasePreserved Profile of the PRECIS
+                as prescribed in Section 3.4.3 of [[!RFC8265]] for the UsernameCasePreserved Profile of the PRECIS
                 IdentifierClass [[!RFC8264]], when setting {{PublicKeyCredentialEntity/name}}'s value, or displaying the value
                 to the user.
 
-              - [=Clients=] MUST perform enforcement, as prescribed in [[!RFC8265]]
-                Section 3.4.3 for the UsernameCasePreserved Profile of the PRECIS IdentifierClass [[!RFC8264]],
+              - [=Clients=] MUST perform enforcement, as prescribed in Section 3.4.3 of [[!RFC8265]]
+                for the UsernameCasePreserved Profile of the PRECIS IdentifierClass [[!RFC8264]],
                 on {{PublicKeyCredentialEntity/name}}'s value prior to displaying the value to the user or
                 including the value as a parameter of the [=authenticatorMakeCredential=] operation.
 
@@ -1794,19 +1794,18 @@ credential.
     :   <dfn>id</dfn>
     ::  The [=user handle=] of the user account entity. To ensure secure operation, authentication and authorization
         decisions MUST be made on the basis of this {{PublicKeyCredentialUserEntity/id}} member,  not the
-        {{PublicKeyCredentialUserEntity/displayName}} nor {{PublicKeyCredentialEntity/name}} members. See [[!RFC8266]]
-        Section 6.1.
+        {{PublicKeyCredentialUserEntity/displayName}} nor {{PublicKeyCredentialEntity/name}} members. See Section 6.1 of [[!RFC8266]].
 
     :   <dfn>displayName</dfn>
     ::  A [=human palatability|human-palatable=] name for the user account, intended only for display. For example, "Alex P. Müller" or "田中 倫". The
         [=[RP]=] SHOULD let the user choose this, and SHOULD NOT restrict the choice more than necessary.
 
-        - [=[RPS]=] SHOULD perform enforcement, as prescribed in [[!RFC8266]]
-            Section 2.3 for the Nickname Profile of the PRECIS FreeformClass [[!RFC8264]],
+        - [=[RPS]=] SHOULD perform enforcement, as prescribed in Section 2.3 of
+            [[!RFC8266]] for the Nickname Profile of the PRECIS FreeformClass [[!RFC8264]],
             when setting {{PublicKeyCredentialUserEntity/displayName}}'s value, or displaying the value to the user.
 
-        - [=Clients=] MUST perform enforcement, as prescribed in [[!RFC8266]]
-            Section 2.3 for the Nickname Profile of the PRECIS FreeformClass [[!RFC8264]],
+        - [=Clients=] MUST perform enforcement, as prescribed in Section 2.3 of
+            [[!RFC8266]] for the Nickname Profile of the PRECIS FreeformClass [[!RFC8264]],
             on {{PublicKeyCredentialUserEntity/displayName}}'s value prior to displaying the value to the user or
             including the value as a parameter of the [=authenticatorMakeCredential=] operation.
 
--- a/index.bs
+++ b/index.bs
@@ -1728,7 +1728,7 @@ associated.
                 [[!RFC8266]] for the Nickname Profile of the PRECIS FreeformClass [[!RFC8264]],
                 when setting {{PublicKeyCredentialEntity/name}}'s value, or displaying the value to the user.
 
-              - [=Clients=] MUST perform enforcement, as prescribed in Section 2.3 of
+              - [=Clients=] SHOULD perform enforcement, as prescribed in Section 2.3 of
                 [[!RFC8266]] for the Nickname Profile of the PRECIS FreeformClass [[!RFC8264]],
                 on {{PublicKeyCredentialEntity/name}}'s value prior to displaying the value to the user or
                 including the value as a parameter of the [=authenticatorMakeCredential=] operation.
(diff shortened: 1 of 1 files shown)


emlun:
Just noting that "not (as yet) fix #593" in fact triggers GitHub to close #593 when this gets merged, so look out for that.

index.bs
```diff
@@ -1716,16 +1719,35 @@ associated.
...
             for display. For example, "ACME Corporation", "Wonderful Widgets, Inc." or "ОАО Примертех".
+
+              - [=[RPS]=] SHOULD perform preparation and enforcement, as prescribed in [[!RFC8266]]
+                Sections 2.2 and 2.3 for the Nickname Profile of the PRECIS FreeformClass [[!RFC8264]],
+                when setting {{PublicKeyCredentialEntity/name}}'s value, or displaying the value to the user.
+
+              - [=Clients=] MUST perform preparation and enforcement, as prescribed in [[!RFC8266]]
+                Sections 2.2 and 2.3 for the Nickname Profile of the PRECIS FreeformClass [[!RFC8264]],
+                on {{PublicKeyCredentialEntity/name}}'s value prior to displaying the value to the user or
+                including the value as a parameter of the [=authenticatorMakeCredential=] operation.
```
emlun: Is this a breaking normative change?

index.bs
```diff
@@ -1716,16 +1719,35 @@ associated.
...
+
           - When inherited by {{PublicKeyCredentialUserEntity}}, it is a [=human palatability|human-palatable=] identifier for a
-            user account. It is intended only for display, and SHOULD allow the user to easily tell the difference between user
+            user account. It is intended only for display, i.e., aiding the user in determining the difference between user
```
emlun: Nice!

index.bs
```diff
@@ -1770,12 +1792,25 @@ credential.
 
 <div dfn-type="dict-member" dfn-for="PublicKeyCredentialUserEntity">
     :   <dfn>id</dfn>
-    ::  The [=user handle=] of the user account entity.
+    ::  The [=user handle=] of the user account entity. To ensure secure operation, authentication and authorization
+        decisions MUST be made on the basis of this {{PublicKeyCredentialEntity/id}} member,  not the
+        {{PublicKeyCredentialEntity/displayName}} nor {{PublicKeyCredentialEntity/name}} members. See [[!RFC8266]]
```
emlun: The class name for `{{PublicKeyCredentialEntity/id}}` and `{{PublicKeyCredentialEntity/displayName}}` should be `PublicKeyCredentialUserEntity`.

equalsJeffH:
@emlun noted:
> Just noting that "not (as yet) fix #593" in fact triggers GitHub to close #593

oops, good catch, thx!  s/fix/address/

index.bs
stpeter: Looks like it to me...

stpeter: BTW, I would just say "enforcement" (not also "preparation") because it's enforcement that really matters here.

index.bs
equalsJeffH: oops, good catch, thx! fixed.

index.bs
emlun: I'm not against a MUST in and of itself, but I expect @nadalin and/or @akshayku will object to a breaking change...

equalsJeffH: @stpeter 
> I would just say "enforcement" (not also "preparation")

thx, yeah, also it turns out "enforcement" invokes "preparation".  I corrected this in https://github.com/w3c/webauthn/pull/951/commits/3ec508d155a3e84a4f8253961282dd1fa361299c

index.bs
```diff
@@ -1716,16 +1719,35 @@ associated.
...
     :   <dfn>name</dfn>
-    ::  A human-readable name for the entity. Its function depends on what the {{PublicKeyCredentialEntity}} represents:
+    ::  A [=human palatability|human-palatable=] name for the entity. Its function depends on what the {{PublicKeyCredentialEntity}} represents:
 
-          - When inherited by {{PublicKeyCredentialRpEntity}} it is a human-friendly identifier for the [=[RP]=], intended only
+          - When inherited by {{PublicKeyCredentialRpEntity}} it is a [=human palatability|human-palatable=] identifier for the [=[RP]=], intended only
             for display. For example, "ACME Corporation", "Wonderful Widgets, Inc." or "ОАО Примертех".
+
+              - [=[RPS]=] SHOULD perform enforcement, as prescribed in [[!RFC8266]]
+                Section 2.3 for the Nickname Profile of the PRECIS FreeformClass [[!RFC8264]],
```
selfissued: Change this and similar constructions from `[[!RFC8264] Section 2.3` to `Section 2.3 of [[!RFC8264]`.

equalsJeffH: sounds fine to me. thx for review.

equalsJeffH:
@stpeter 
> do we really need to go down this path? What attacks are we trying to prevent?

it seems you are referring to [your question here](https://github.com/w3c/webauthn/pull/878#issuecomment-388111270).

As @emlun and @aphillips noted [here](https://github.com/w3c/webauthn/pull/878#issuecomment-388122773) and [here](https://github.com/w3c/webauthn/pull/878#issuecomment-388126103) (respectively), it does not seem that appying PRECIS  to the name-ish DOMStrings aids in preventing any known attacks per se.  

However, do we wish to suggest to the RP that they may allow the user to supply arbitrary unicode values for [`PublicKeyCredentialUserEntity/displayName`](https://w3c.github.io/webauthn/#dom-publickeycredentialuserentity-displayname) and [`PublicKeyCredentialEntity/name`](https://w3c.github.io/webauthn/#dom-publickeycredentialentity-name) without doing some form of PRECIS preparation and enforcement on them, even if only for "hygine" reasons?

@selfissued 
> in the name of simplicity, I believe we should close this PR and #593 with no action

Well, there's more going on in this PR than strictly applying PRECIS to those name-ish DOMStrings.  For example, it incorporates @stpeter's [suggestion](https://github.com/w3c/webauthn/pull/878#issuecomment-388162488) wrt warning developers wrt using these name-ish DOMStrings as authz identifiers. 

If we decide we do not wish to impose PRECIS on the unicode values of [`PublicKeyCredentialUserEntity/displayName`](https://w3c.github.io/webauthn/#dom-publickeycredentialuserentity-displayname) and [`PublicKeyCredentialEntity/name`](https://w3c.github.io/webauthn/#dom-publickeycredentialentity-name), retaining the above warning is perhaps worthwhile.

equalsJeffH:
@emlun, yeah, if we incorp the PRECIS stipulations, mebbe all should be SHOULDs.

stpeter:
Thanks @equalsJeffH for your consideration. Retaining the warning text and changing the normativity to SHOULD seems like a good approach. Better safe than sorry...

equalsJeffH:
thx, I changed client-side normativity to SHOULD.

Please note also that this PR only improves #593 -- the other, original, portion of the issue is in regards to providing implementer guidance regarding how to display/present these string values in order to mitigate effects of possibly malicious string content, per @zbraniecki, [as quoted by](https://github.com/w3c/webauthn/issues/593#issuecomment-369402225) @jcjones:
> @zbraniecki suggests that if we want to let browsers use this information for prompts, that we should write / refer to some guidance on how to present the strings, given that the attacker can do three major things with a DOMString:

> > play with position and directionality (so the string may appear on the far right of the field rather than left, or in the middle), make things invisible within the string or confusing (emojis), or may the string contain characters that look like other characters.

> My poor synopsis of some sample guidance would be advice to always use UI elements to provide a clear boundary around these strings, and not allow overflow into other elements, etc.

[as I noted in](https://github.com/w3c/webauthn/issues/593#issuecomment-397378715) #593, I have done some modest searching around WHATWG and W3C specs regarding how one might specify guidance per @jcjones's suggestion above, but have been unable to find anything useful -- does anyone have any clues?

stpeter:
Good point. I looked in the [Character Model for the Web](https://w3c.github.io/charmod-norm/) but it doesn't talk about user interfaces. Perhaps @aphillips or @r12a knows of a helpful pointer.

aphillips:
Display issues are complicated at best. I don't know that there is a good pointer per-se. As far as I know, no one has solved the problem of *both* general readability of arbitrary Unicode text *and* distinct display of strings that are encoded distinctly. That's certainly true across a myriad of platforms, display technologies, fonts, etc.

Bidi isolation is one thing that is probably good to cite, since that helps solve the problem of using bidi characters/controls to cause text to "[swap places](https://www.w3.org/International/articles/inline-bidi-markup/uba-basics#isolation)" in an abusive way.

WRT overall avoidance of spoofing, it boils down to: some characters/character sequences look like other characters/character sequences. Some applications try to make some of these differences visible (notably changing the background color or other decoration in the address bar when an IDNA address is shown), but none of these completely remove the possibility of confusables being used. As noted previously, PRECIS doesn't fix this.

On some level, the byte- or code-point-level comparison of strings boils down to the code units used for interchange. PRECIS and Charmod are both about the need for namespaces to have reasonable, reliable, consistent rules for doing this--mainly to assist users with the problem of inputting the values in separate places at separate times and getting the match one expects. 

One last observation: W3C I18N used to promote NFC and the idea of really strict rules for namespaces and such. In practice, over 25 years or so, this hasn't turned out to be the problem we thought it would be. Permissive name spaces seem to work the majority of the time: they perform better, are simpler, don't disadvantage scripts/languages that need unusual encoding features, and don't try to solve problems no one is asking about--the main caveat being that you can't always see that the strings do/don't match (which, as noted, they didn't solve anyway). In namespaces, it often seems that it works better to put the health warning of "since you can't see the underlying bytes, be sure that they do match" and do little to "assist" users by mutating their input than to try to create complex rules that force the issue (or arcane display hoops to jump through). YMMV.

equalsJeffH:
thx @stpeter & @aphillips 

Ok, so if I interpret the above comments correctly, having the guidance in the spec whereas servers and clients SHOULD apply PRECIS to those name-ish strings, is nominally OK?

WRT the other portion of issue #593 ..
> 2. providing implementer guidance regarding how to display/present these string values in order to mitigate effects of possibly malicious string content.

..we could as @jcjones suggests concoct some modest _guidance/advice_ to "...always use UI elements to provide a clear boundary around these strings, and **not allow overflow into other elements**, etc."  ?    (the latter is [the crux of the feedback](https://github.com/w3c/webauthn/issues/593#issuecomment-369402225) that @jcjones got from @zbraniecki)

stpeter:
Agreed, the overflow protections are important.

aphillips:
@equalsJeffH Actually, I think:

> guidance in the spec whereas servers and clients SHOULD apply PRECIS to those name-ish strings

 is not nominally okay :-). Either webauth needs to require the normalization or it should have a health warning that similar looking strings that are not encoded identically do not match. In effect this is what [Charmod](http://w3c.github.io/charmod-norm/#normalizationChoice) is about (link goes to the specific section talking about spec choices related to normalization. If you choose to specify PRECIS, you should require it so that matches are consistent across implementations. Otherwise, you can recommend to content authors that they use the PRECIS rules and recommend to implementations that they warn users when `name` type values are not in PRECIS form--these are good things. But `SHOULD` is insufficient for good interop if the names are expected to match later.

I agree about the presentation warning.

equalsJeffH:
we decided on the 11-Jul webauthn call to go ahead and merge this PR as-is -- if one feels there are problems with it, please submit specific new issues.

equalsJeffH:
wrt..
> Agreed, the overflow protections are important.  (@stpeter)
> I agree about the presentation warning.  (@aphillips)

I have done some modest searching thru WhatWG and W3C specs and am unable to find appropriate example language to plagiarize for such a "presentation warning".

In any case, here's what I presently came up with:

When [=clients=], client platform, or [=authenticators=] display a [`PublicKeyCredentialUserEntity/displayName`](https://w3c.github.io/webauthn/#dom-publickeycredentialuserentity-displayname)'s or [`PublicKeyCredentialEntity/name`](https://w3c.github.io/webauthn/#dom-publickeycredentialentity-name)'s value, they should always use UI elements to provide a clear boundary around the displayed value, and not allow overflow into other elements [[css-overflow-3]].

thoughts?

stpeter:
Seems like a good start, thanks @equalsJeffH!

equalsJeffH:
ok, I've added a presentation warning in 4ddc3c0 as proposed in https://github.com/w3c/webauthn/pull/951#issuecomment-404295659 -- would be good to get someones to review that addition.  thx.

equalsJeffH:
given https://github.com/w3c/webauthn/pull/951#issuecomment-404278183 where we decided on the 11-Jul webauthn call to go ahead and merge this PR as-is, and @stpeter's nominal ok of the presentation warning https://github.com/w3c/webauthn/pull/951#issuecomment-404297635, I'm merging this.  if anyone feels there are problems with it, please submit specific new issues.

PR #962, opened 2018-06-20
Allow client to return NotAllowedError early if success is impossible

### LINKED ISSUE #953: Clarify WebAuthn spec to allow us to return an error to RP when it makes sense
I believe that right now, if a list of credentialIDs (with corresponding transports) gets sent to the client, and the client's isn't able to satisfy the request (let's say, all the credentialIDs referenced BLE as the only transport, but the platform doesn't support BLE), the client should be allowed to show an informative error to the user, and return an error to the RP. I think right now the spec says the client has to time the request out, but that seems subpar from a UX point of view.

Note that this is similar to the case where an RP is only looking for built-in credentials, and no credentialIDs that match is found on the platform. In this case we already said we're going to show an error to the user, and report an error back to the RP. This is just an extension of that use case.

emlun:
This aims to fix #953.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/pull/962.html" title="Last updated on Jun 26, 2018, 3:51 PM GMT (7791942)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/962/2d669de...7791942.html" title="Last updated on Jun 26, 2018, 3:51 PM GMT (7791942)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1101,9 +1101,9 @@ When this method is invoked, the user agent MUST execute the following algorithm
             1. Return |constructCredentialAlg| and terminate this algorithm.
 
         :   If the [=client=] concludes that it is impossible for any eligible [=authenticator=] to appear,
-        ::  The client MAY prompt the user for [=user consent|consent=] to create a new credential. If the user confirms [=user
-            consent|consent=], the [=client=] MAY return a {{DOMException}} whose name is "{{NotAllowedError}}" and terminate this
-            algorithm.
+        ::  The client MAY prompt the user for [=user consent|consent=] to create a new credential. If the user [=user
+            consent|consents=], the [=client=] MAY return a {{DOMException}} whose name is "{{NotAllowedError}}" and terminate
+            this algorithm.
 
             Note: For example, the [=client=] might reach this conclusion if
             <code>|options|.{{PublicKeyCredentialCreationOptions/authenticatorSelection}}.{{authenticatorAttachment}}</code> is
@@ -1457,8 +1457,8 @@ When this method is invoked, the user agent MUST execute the following algorithm
 
         :   If the [=client=] concludes that it is impossible for any eligible [=authenticator=] to appear,
         ::  The client MAY prompt the user for [=user consent|consent=] to [=authentication|authenticate=] to the [=[RP]=]. If
-            the user confirms [=user consent|consent=], the [=client=] MAY return a {{DOMException}} whose name is
-            "{{NotAllowedError}}" and terminate this algorithm.
+            the user [=user consent|consents=], the [=client=] MAY return a {{DOMException}} whose name is "{{NotAllowedError}}"
+            and terminate this algorithm.
 
             Note: For example, the [=client=] might reach this conclusion if
             <code>|options|.{{PublicKeyCredentialRequestOptions/allowCredentials}}</code> is not [=list/empty=] and all its
--- a/index.bs
+++ b/index.bs
@@ -1101,9 +1101,9 @@ When this method is invoked, the user agent MUST execute the following algorithm
             1. Return |constructCredentialAlg| and terminate this algorithm.
 
         :   If the [=client=] concludes that it is impossible for any eligible [=authenticator=] to appear,
-        ::  The client MAY prompt the user for [=user consent|consent=] to [=authentication|authenticate=] to the [=[RP]=]. If
-            the user does [=user consent|consent=], the [=client=] MAY return a {{DOMException}} whose name is
-            "{{NotAllowedError}}" and terminate this algorithm.
+        ::  The client MAY prompt the user for [=user consent|consent=] to create a new credential. If the user does [=user
+            consent|consent=], the [=client=] MAY return a {{DOMException}} whose name is "{{NotAllowedError}}" and terminate this
+            algorithm.
 
             Note: For example, the [=client=] might reach this conclusion if
             <code>|options|.{{PublicKeyCredentialCreationOptions/authenticatorSelection}}.{{authenticatorAttachment}}</code> is
@@ -1456,9 +1456,9 @@ When this method is invoked, the user agent MUST execute the following algorithm
             1.  Return |constructAssertionAlg| and terminate this algorithm.
 
         :   If the [=client=] concludes that it is impossible for any eligible [=authenticator=] to appear,
-        ::  The client MAY prompt the user for [=user consent|consent=] to create a new credential. If the user does [=user
-            consent|consent=], the [=client=] MAY return a {{DOMException}} whose name is "{{NotAllowedError}}" and terminate this
-            algorithm.
+        ::  The client MAY prompt the user for [=user consent|consent=] to [=authentication|authenticate=] to the [=[RP]=]. If
+            the user does [=user consent|consent=], the [=client=] MAY return a {{DOMException}} whose name is
+            "{{NotAllowedError}}" and terminate this algorithm.
 
             Note: For example, the [=client=] might reach this conclusion if
             <code>|options|.{{PublicKeyCredentialRequestOptions/allowCredentials}}</code> is not [=list/empty=] and all its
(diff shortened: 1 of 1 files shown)


emlun:
This is currently written as MAYs so as to be minimally breaking.

An alternative solution to the case where `authenticatorSelection` or `allowCredentials.transports` are incompatible with what the platform supports would be to add a step directly after the origin check where the client MAY return a `ConstraintError` or similar without user consent. That might be a "more" breaking change, though. There might also be implications for privacy considerations, but I think they would be very minor.

If anyone would prefer this alternative solution, please voice your support explicitly.

index.bs
```diff
@@ -1100,6 +1100,16 @@ When this method is invoked, the user agent MUST execute the following algorithm
 
             1. Return |constructCredentialAlg| and terminate this algorithm.
 
+        :   If the [=client=] concludes that it is impossible for any eligible [=authenticator=] to appear,
+        ::  The client MAY prompt the user for [=user consent|consent=] to [=authentication|authenticate=] to the [=[RP]=]. If
```
selfissued: I believe that the first "The" should be "the".

index.bs
```diff
@@ -1445,6 +1455,15 @@ When this method is invoked, the user agent MUST execute the following algorithm
                 on |authenticator| and [=set/remove=] it from |issuedRequests|.
 
             1.  Return |constructAssertionAlg| and terminate this algorithm.
+
+        :   If the [=client=] concludes that it is impossible for any eligible [=authenticator=] to appear,
+        ::  The client MAY prompt the user for [=user consent|consent=] to create a new credential. If the user does [=user
```
selfissued: I believe that the first "The" should be "the".

index.bs
emlun: At first I thought so too - but all the other cases in this switch statement use this style, so I changed this one to do that too for consistency. Let me know if you want me to revert it to lowercase.

equalsJeffH: offhand, IMHO, we should use ":" at the end of the switch stmt title rather than "," -- fwiw, using comma may be an artifact of my having just blindly following the style established by other more w3c-experienced editors and not thinking about it...

equalsJeffH: hm, here we're adding this switch clause here to the [#createCredential](https://w3c.github.io/webauthn/#createCredential) alg, yes?

However, @christiaanbrand's issue #953 appears to be in regards to the [#getAssertion](https://w3c.github.io/webauthn/#discover-from-external-source) alg, even though he's not explicit about it, yes?

emlun: Yes, I added it to both operations. The situation is relevant to both, but caused by different circumstances.

emlun: Oh, but I had apparently swapped the Note:s between the two operations. That might not be what you asked about, but still, thanks!

index.bs
```diff
@@ -1100,6 +1100,15 @@ When this method is invoked, the user agent MUST execute the following algorithm
 
             1. Return |constructCredentialAlg| and terminate this algorithm.
 
+        :   If the [=client=] concludes that it is impossible for any eligible [=authenticator=] to appear,
+        ::  The client MAY prompt the user for [=user consent|consent=] to [=authentication|authenticate=] to the [=[RP]=]. If
+            the user does [=user consent|consent=], the [=client=] MAY return a {{DOMException}} whose name is
```
equalsJeffH: should "user does" actually be "user does not" ?

index.bs
```diff
@@ -1445,6 +1454,16 @@ When this method is invoked, the user agent MUST execute the following algorithm
                 on |authenticator| and [=set/remove=] it from |issuedRequests|.
 
             1.  Return |constructAssertionAlg| and terminate this algorithm.
+
+        :   If the [=client=] concludes that it is impossible for any eligible [=authenticator=] to appear,
+        ::  The client MAY prompt the user for [=user consent|consent=] to create a new credential. If the user does [=user
```
equalsJeffH: We are not asking for consent to create a new cred here, rather to generate an assertion ie to authenticate, yes?

equalsJeffH: should "user does" be "user does not" ?

index.bs
equalsJeffH: for some reason I saw the "If the client concludes that it is impossible for any eligible authenticator to appear" switch clause only in the [[Create]] alg, and not in [[DiscoverFrom...Source]].  cockpit error on my part.

equalsJeffH:
I re-reviewed and added come other comments. thx.

index.bs
emlun: No, this is intended to mean "if the user indeed consents"; so perhaps that needs to be reworded. My thinking is that if the user does not consent, then the client must wait for the timeout instead of returning early.

index.bs
emlun: Good catch, thanks! See https://github.com/w3c/webauthn/pull/962#discussion_r197862736 for my answer to the second question.

emlun: Heh, I had swapped not only the Note:s between algorithms, but in fact the whole algorithm steps...

index.bs
```diff
@@ -1100,6 +1100,15 @@ When this method is invoked, the user agent MUST execute the following algorithm
 
             1. Return |constructCredentialAlg| and terminate this algorithm.
 
+        :   If the [=client=] concludes that it is impossible for any eligible [=authenticator=] to appear,
+        ::  The client MAY prompt the user for [=user consent|consent=] to create a new credential. If the user confirms [=user
+            consent|consent=], the [=client=] MAY return a {{DOMException}} whose name is "{{NotAllowedError}}" and terminate this
```
equalsJeffH: I would write:
If the user confirms [=user consent|consent=]  ->  If the user [=user consent|consents=]

index.bs
```diff
@@ -1445,6 +1454,16 @@ When this method is invoked, the user agent MUST execute the following algorithm
                 on |authenticator| and [=set/remove=] it from |issuedRequests|.
 
             1.  Return |constructAssertionAlg| and terminate this algorithm.
+
+        :   If the [=client=] concludes that it is impossible for any eligible [=authenticator=] to appear,
+        ::  The client MAY prompt the user for [=user consent|consent=] to [=authentication|authenticate=] to the [=[RP]=]. If
+            the user confirms [=user consent|consent=], the [=client=] MAY return a {{DOMException}} whose name is
```

emlun:
On second thought, this should probably be a different error than `NotAllowedError`, right? In [create()][create], the current possible errors and their reasons are:

- `NotAllowedError`
  - _sameOriginWithAncestors_ is false
  - _callerOrigin_ is an opaque origin
  - Timeout
- `SecurityError`
  - Effective domain is not a valid domain
  - `options.rpId` is not a registrable domain suffix of and is not equal to _effectiveDomain_
- `NotSupportedError`
  - _credTypesAndPubKeyAlgs_ is empty and `options.pubKeyCredParams` is not empty
- `AbortError`
  - Aborted by user
- `InvalidStateError`
  - User tried to register an already registered authenticator

And in [get()][get]:

- `NotAllowedError`
  - _sameOriginWithAncestors_ is false
  - _callerOrigin_ is an opaque origin
  - Timeout
- `SecurityError`
  - Effective domain is not a valid domain
  - `options.rpId` is not a registrable domain suffix of and is not equal to _effectiveDomain_
- `AbortError`
  - Aborted by user

If we want this error condition to be clearly discernible from the others, it should probably be something more like `ConstraintError` in create() and `NotFoundError` in get().

[create]: https://w3c.github.io/webauthn/#createCredential
[get]: https://w3c.github.io/webauthn/#discover-from-external-source

emlun:
Noted on 2018-06-27 WG call that this doesn't need to be a discernible error condition, so `NotAllowedError` is fine.

equalsJeffH:
so do we need to wait for goog folk to review further before merging?

emlun:
@christiaanbrand requested on the call that the requirement to obtain consent be removed, and that just informing the user before returning the error should be permitted. @akshayku said in response that he wanted to look at/think about this some more before we make that change, and said he'd add a comment here in a day or so.

emlun:
Hm, I just noticed the getAssertion part of this is already fixed in #906, so if the makeCredential part of this isn't important we can just close this.

emlun:
Closing this since #906 already covers the motivating issue #953. If anyone wants the makeCredential part of this, you're welcome to re-open it.

PR #974, opened 2018-06-28
fix #180: do not totally lose the term "WebAuthn Relying Party"

### LINKED ISSUE #180: do not totally lose the term "WebAuthn Relying Party"
this issue is derived from, and focuses only on the "WebAuthn Relying Party" portion of..

> Re: [webauthn] Text aesthetic updates
> https://lists.w3.org/Archives/Public/public-webauthn/2016Jul/0021.html

Regarding this set of commits..

> [webauthn] new commits pushed by jcjones
> https://lists.w3.org/Archives/Public/public-webauthn/2016Jul/0016.html

...

> - Change "WebAuthn Relying Party" to just "Relying Party"  

..although this change reduces wordiness, I do have some concerns..

I can live with using the unqualified term "Relying Party" in
many places in the text, however, we have been hearing literally for
years, especially from the Fed'd Identity Mgmt community, how confusing
they find the use of the unqualified "Relying Party" term.  Over the years
I've observed many instances of unnecessary misunderstanding and
getting-bolloxed-up due to use of imprecise/ambiguous terminology.

Thus, if others object to using the fully-qualified WebAuthn Relying Party" term 
everywhere, then I suggest editing the new RP definition to be like so..

```
WebAuthn Relying Party
Relying Party

  The entity whose web application utilizes the Web
  Authentication API to register and authenticate users.
  See Registration and Authentication, respectively.

  Note: While the unqualified term Relying Party is also used in other
  contexts (e.g., OAuth, SAML, X.509, etc.), an entity acting as a
  Relying Party in one context is not necessarily a Relying Party
  in others. For example, in an identity management deployment,
  WebAuthn may be used as a user authentication mechanism at the
  Identity Provider, who thus is also acting in the WebAuthn
  Relying Party role.
```

Additionally, within each major section I'd use the "WebAuthn Relying
Party" term on first use and "Relying Party" otherwise.

I think it's really important we make this extra effort to disambiguate
the terms. 

Also, we /could/ use the acronym "WRP", say, rather than the spelled-out
unqualified "Relying Party" term.

equalsJeffH:
fixes #180

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/pull/974.html" title="Last updated on Jun 28, 2018, 1:22 AM GMT (c0aefc4)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/974/a583650...c0aefc4.html" title="Last updated on Jun 28, 2018, 1:22 AM GMT (c0aefc4)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -52,6 +52,7 @@ Text Macro: RPS Relying Parties
 Text Macro: INFORMATIVE <em>This section is not normative.</em>
 Text Macro: TRUE <code>true</code>
 Text Macro: WAC WebAuthn Client
+Text Macro: WRP WebAuthn Relying Party
 Ignored Vars: op, alg, type, algorithm
 Abstract: This specification defines an API enabling the creation and use of strong, attested, [=scoped=], public key-based
  credentials by web applications, for the purpose of strongly authenticating users. Conceptually, one or more [=public key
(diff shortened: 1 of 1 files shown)


emlun:
All right, that makes a lot of sense. :)

nadalin:
@selfissued  Please review

PR #976, opened 2018-06-28
fix #493: be explicit about "same user" is verified at get() time as was verified at create() time

### LINKED ISSUE #493: be explict about "same user" is verified at get() time as was verified at create() time
this may involve some changes to the two main algorithms, and definitely should be mentioned in the user verification definition.  as @jyasskin notes in PR #460:
> Even if the authenticator can "distinguish individual users", it needs to record which user authorized create() and only sign credentials when the same user authorizes a get() call.

equalsJeffH:
fixes #493

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/pull/976.html" title="Last updated on Jul 6, 2018, 10:19 PM GMT (9bd0894)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/976/a96110e...9bd0894.html" title="Last updated on Jul 6, 2018, 10:19 PM GMT (9bd0894)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2601,8 +2601,10 @@ When this method is invoked, the [=authenticator=] MUST perform the following pr
     by the [=authenticator=] if it has its own output capability, or by the user agent otherwise.
 
     If |requireUserVerification| is [TRUE], the method of obtaining [=user consent=] MUST include [=user verification=].
-    If successful, the verified user MUST be the same user as was verified when this [=authenticator=] created this
-    [=public key credential=] in <a href='#op-makecred-step-user-consent'>Step 6</a> of [[#op-make-cred]].
+
+        Note: For the overal WebAuthn security properties to hold, the verified user must be the same user that was verified
+        when this [=authenticator=] created this [=public key credential=] in <a href='#op-makecred-step-user-consent'>Step
+        6</a> of [[#op-make-cred]].
 
     If |requireUserPresence| is [TRUE], the method of obtaining [=user consent=] MUST include a
         [=test of user presence=].
--- a/index.bs
+++ b/index.bs
@@ -846,7 +846,8 @@ When this method is invoked, the user agent MUST execute the following algorithm
         such as [=domain=], [=ipv4 address=], [=ipv6 address=], [=opaque host=], or [=empty host=].
         Only the [=domain=] format of [=host=] is allowed here.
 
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -2488,18 +2489,24 @@ When this operation is invoked, the [=authenticator=] MUST perform the following
...
-    If |requireUserPresence| is [TRUE], the method of obtaining [=user consent=] MUST include a [=test of user presence=].
+        If |requireUserVerification| is [TRUE], the method of obtaining [=user consent=] MUST include [=user verification=].
 
-    If the user does not [=user consent|consent=] or if [=user verification=] fails, return an error code equivalent to
-    "{{NotAllowedError}}" and terminate the operation.
+        If |requireUserPresence| is [TRUE], the method of obtaining [=user consent=] MUST include a [=test of user presence=].
+
+        If the user does not [=user consent|consent=] or if [=user verification=] fails, return an error code equivalent to
+        "{{NotAllowedError}}" and terminate the operation.
+    </li>
```
emlun: The above changes seem to have had no effect on the rendered result, correct?

index.bs
```diff
@@ -2595,6 +2602,10 @@ When this method is invoked, the [=authenticator=] MUST perform the following pr
 
     If |requireUserVerification| is [TRUE], the method of obtaining [=user consent=] MUST include [=user verification=].
 
+        Note: For the overall WebAuthn security properties to hold, the user verified here must be the same user that was
+        verified when this [=authenticator=] created this [=public key credential=] in
+        <a href='#op-makecred-step-user-consent'>Step 6</a> of [[#op-make-cred]].
+
```
emlun: The "must" here looks normative to me. Should this be in a Note: and not in the algorithm proper?

equalsJeffH: Well, yeah, one could argue that. I made it a note with lowercase 'must' at this point to sidestep the whole breaking change thing.  

That the topic of the Note must be true for webauthn to "work" is tacitly understood by everyone (one hopes) and this is just pedantically stating it. 

in any case I'm happy to make it a normative statement if that's what folks want to do. one way or the other we ought to explicitly state it and get it into the spec.

emlun: I agree. Let's hear what people think on the 2018-07-09 WG call.

PR #994, opened 2018-07-11
Add table numbers and captions

### LINKED ISSUE #990: add captions for tables and id attrs so we can link to them
we are likely adding yet another table (PR #979). we have captions for figures, and id attrs so we can link to them, should do the same for tables.  the two current tables are:

https://w3c.github.io/webauthn/#authenticator-data

https://w3c.github.io/webauthn/#attested-credential-data

emlun:
Fixes #990.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/pull/994.html" title="Last updated on Jul 12, 2018, 10:01 AM GMT (bc15894)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/994/ca80875...bc15894.html" title="Last updated on Jul 12, 2018, 10:01 AM GMT (bc15894)">Diff</a>


emlun:
Ok, I'll do this for the new table too. Thanks!

Yes, the style is needed. [Here][bikeshed] is how Bikeshed's builtin figure numbering works; without the styles added here, these tables would be captioned and numbered as figures.

Oh, but maybe `<table><caption/>` is more suitable for this than `<figure><table/><figcaption/>`; I didn't find that element when I looked around earlier. I'll try changing to that and seeing if that works with Bikeshed; in that case the styles can probably be pruned a bit.

[bikeshed]: https://github.com/tabatkins/bikeshed/blob/master/bikeshed/Spec.py#L556-L561

emlun:
Nope, `<table><caption/>` requires the caption be placed before the table body.

The `max-width` style is not necessary for the table numbers, but it prevents the tables from getting oddly offset from the text if viewed on a very wide screen.

emlun:
See also tabatkins/bikeshed#1314

PR #1003, opened 2018-07-22
Removed old optional counter step to remove confusions

herrjemand:
***
<a href="https://pr-preview.s3.amazonaws.com/herrjemand/webauthn/pull/1003.html" title="Last updated on Jul 22, 2018, 11:52 AM GMT (abe4374)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/1003/741cef6...herrjemand:abe4374.html" title="Last updated on Jul 22, 2018, 11:52 AM GMT (abe4374)">Diff</a>


herrjemand:
@emlun So my confusion comes from the fact that we previously said that if authenticator does not implement Counter, then there would be either MSB set to true, or the counter value is zero. Currently there is none of this mentions in Webauthn or CTAP2 specs. So we need to clarify this behaviour.

emlun:
Yes, sorry - I agree this should be clarified, but I think we should add the missing steps to the authenticator operations instead. It doesn't necessarily need to align with CTAP; the specs will still be compatible as long as CTAP doesn't specify a different behavior for the "not implemented" case (but it would of course be more useful if it's optional in both specs).

herrjemand:
Okay, will reopen in FIDO2 repo then *)

PR #1009, opened 2018-07-25
Clarify behaviour for authnrs not implementing signature counter

### LINKED ISSUE #1008: Clarify authenticator behaviour when not implementing signature counter
@herrjemand notes in PR #1003:

>[...] we previously said that if authenticator does not implement Counter, then there would be either MSB set to true, or the counter value is zero. Currently there is none of this mentions in Webauthn or CTAP2 specs. So we need to clarify this behaviour.

I agree we should clarify this: if the authenticator does not implement a signature counter, it should always set the signature count to zero. This will be a small change to the authenticator algorithms, but I argue that this is an editorial change since the RP operations already imply that this is the only possible implementation that RPs would accept.

emlun:
Fixes #1008

I argue that this change is editorial since this was already implied by the [RP operations][rp-ops] as the only acceptable implementation. This merely makes that explicit in the authenticator operations.

[rp-ops]: https://w3c.github.io/webauthn/#ref-for-signature-counter%E2%91%A0%E2%91%A6

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/pull/1009.html" title="Last updated on Jul 25, 2018, 1:06 PM GMT (62cdb51)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/1009/8b0eb71...62cdb51.html" title="Last updated on Jul 25, 2018, 1:06 PM GMT (62cdb51)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2796,6 +2796,9 @@ a numbered step. If outdented, it (today) is rendered as a bullet in the midst o
               [=authenticator data=].
         : a per credential [=signature counter=]
         :: allocate the counter, associate it with the new credential, and initialize the counter value as zero.
+
+        : no [=signature counter=]
+        :: let the [=signature counter=] value for the new credential be constant at zero.
     </dl>
 
 1. Let |attestedCredentialData| be the [=attested credential data=] byte array including the |credentialId| and |publicKey|.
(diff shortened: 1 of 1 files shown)


nadalin:
@emlun Please merge

PR #1023, opened 2018-08-02
fix #1012 determine Attested Credential Data length

### LINKED ISSUE #1012: Determining length of `attestedCredentialData` when authenticator extensions present.
Pardon the confusion, but I'm having trouble figuring out how, as a Relying Party, to reliably determine the length of the `attestedCredentialData` in [authenticator data][1] when authenticator extensions are present.

According to the spec, `attestedCredentialData`'s "length depends on the length of the credential ID and credential public key being attested." It also states the following regarding the length of both the `attestedCredentialData` and `extensions`:

"Note that the authenticator data describes its own length: If the AT and ED flags are not set, it is always 37 bytes long. The attested credential data (which is only present if the AT flag is set) describes its own length. If the ED flag is set, then the total length is 37 bytes plus the length of the attested credential data, plus the length of the CBOR map that follows."

The `credentialPublicKey` within `attestedCredentialData` is also of variable length. In the past, before handling extensions, I just took the rest of the `authenticatorData` after `credentialId`, and assumed that was the `credentialPublicKey`. I've noticed a handful of other implementations making this same assumption. However, if the ED flag is set, how does the Relying Party know the length of the `credentialPublicKey`? Is the solution just to figure out where the `credentialPublicKey` CBOR data structure ends? If so, maybe this should be made clear in the spec?

Thanks!

[1]: https://w3c.github.io/webauthn/#authenticator-data

equalsJeffH:
fixes #1012

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/pull/1023.html" title="Last updated on Aug 2, 2018, 8:13 PM GMT (2d1bdc8)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/1023/24aed4e...2d1bdc8.html" title="Last updated on Aug 2, 2018, 8:13 PM GMT (2d1bdc8)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2501,7 +2501,7 @@ The [figure below](#fig-authData) shows a visual representation of the [=authent
     The [=attested credential data=] (which is only present if the AT [=flag=] is set) describes its own length. If the ED [=flag=] is set, then the total length is 37 bytes plus the length of the [=attested credential data=] (if the AT [=flag=] is set), plus the length of the [=authDataExtensions|extensions=] output (a [=CBOR=] map) that
     follows.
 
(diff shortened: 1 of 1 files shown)


nadalin:
@equalsJeffH Please merge

PR #1024, opened 2018-08-02
Use final name for "serialize JSON to bytes"

domenic:
This is a follow-up to #1017 to use the final name arrived at in https://github.com/whatwg/infra/pull/207 and merged into the Infra Standard.

This will cause Bikeshed linking warnings now (due to the removed link block) but sometime within the next 24 hours it should start working automatically, so as long as you merge then, it'll be fine.

***
<a href="https://pr-preview.s3.amazonaws.com/domenic/webauthn/pull/1024.html" title="Last updated on Aug 2, 2018, 9:05 PM GMT (89bfd90)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/1024/24aed4e...domenic:89bfd90.html" title="Last updated on Aug 2, 2018, 9:05 PM GMT (89bfd90)">Diff</a>

PR #1026, opened 2018-08-08
Add bikeshed update to bikeshed Dockerfile

emlun:
As suggested by @domenic in https://github.com/w3c/webauthn/pull/1024#issuecomment-411116681


emlun:
This PR is mostly for visibility and has no effect on the spec; merging.

PR #1032, opened 2018-08-09
Add note about actual implementations of appid.

### LINKED ISSUE #972: AppID extension: protocol version number?
According to the [AppID and Facets spec](https://fidoalliance.org/specs/fido-v2.0-id-20180227/fido-appid-and-facets-v2.0-id-20180227.html), after downloading the `TrustedFacetList` the client should:

> From among the objects in the trustedFacet array, select the one with the version matching that of the protocol message version. With "matching" we mean: the highest version that appears in the TrustedFacetList that is smaller or equal to the actual protocol version being used. 

UAF has protocol versions, but I'm not sure what to use for WebAuthn. A clarification in the appid extension would be helpful.

agl:
This change adds a note to the `appid` extension remarking that, in
practice, I don't think anyone actually implements the FIDO FacetID spec
fully and instead checks whether the claimed AppID is same host with the
origin.

Fixes #972.

***
<a href="https://pr-preview.s3.amazonaws.com/agl/webauthn/pull/1032.html" title="Last updated on Aug 9, 2018, 8:16 PM GMT (b738a04)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/1032/d74f56b...agl:b738a04.html" title="Last updated on Aug 9, 2018, 8:16 PM GMT (b738a04)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -154,6 +154,10 @@ spec: WHATWG HTML; urlPrefix: https://html.spec.whatwg.org/
         text: focus
         text: username; url: attr-fe-autocomplete-username
 
+spec: WHATWG URL; urlPrefix: https://url.spec.whatwg.org/
+    type: dfn
+        text: same site; url: host-same-site
+
 spec: FIDO-CTAP; urlPrefix: https://fidoalliance.org/specs/fido-v2.0-id-20180227/fido-client-to-authenticator-protocol-v2.0-id-20180227.html
     type: dfn
         text: CTAP2 canonical CBOR encoding form; url: ctap2-canonical-cbor-encoding-form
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -4257,6 +4264,11 @@ JavaScript APIs.
         <var ignore>allowCredentialDescriptorList</var> and set |output| to [TRUE]. The value of |appId| then replaces the `rpId`
         parameter of [=authenticatorGetAssertion=].
 
+Note: In practice, several implementations do not implement steps four and onward of the
+algorithm for [=determining if a caller's FacetID is authorized for an AppID=].
+Instead, in step three, the comparison on the host is relaxed to accept hosts on the
```
selfissued: What does "hosts on the same site" mean?  Do you mean subdomains, such as "onmicrosoft.com" also matching "mytenant.onmicrosoft.com"?

agl: The words `same site` should be linked in the generated HTML to [the WHATWG specification](https://url.spec.whatwg.org/#host-same-site) which defines this.

nadalin:
@aql Please merge

PR #1033, opened 2018-08-09
Note that the appid extension changes RP ID hash

### LINKED ISSUE #980: Clarification of valid rpIdHash value requested in section 7 when using AppID extension
There are a couple of usability issues with following the RP operations guidance in section 7.2 for verifying an authentication assertion when using the AppID extension.

Even though I can validly send just the web server's hostname as the RP-ID in the call to navigator.credentials.get, if I include the AppID extension, the rpIdHash returned in aData (see step 11 of section 7.2) is the hash of the U2F APPID. 

What adds to this confusion is that verification of client extension outputs (where you can detect if "appid": true is returned) is not mentioned until step 14 of section 7.2. In fact in order to validate the rpIdHash in step 11 you must FIRST figure out if appid was used.

My suggestion is to add clarification to step 11 indicating that appid in the client extension outputs may need to be checked first to ensure the correct rpIdHash is compared.

agl:
This change adds a note to clarify that successful use of the `appid`
extension will cause the RP ID hash in the returned assertion to be the
hash of the AppID rather than the hash of the RP ID.

Fixes #980.

***
<a href="https://pr-preview.s3.amazonaws.com/agl/webauthn/pull/1033.html" title="Last updated on Aug 9, 2018, 8:32 PM GMT (b6a48d7)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/1033/d74f56b...agl:b6a48d7.html" title="Last updated on Aug 9, 2018, 8:32 PM GMT (b6a48d7)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -4258,7 +4258,7 @@ JavaScript APIs.
         parameter of [=authenticatorGetAssertion=].
 
 : Client extension output
-:: Returns the value of |output|.
+:: Returns the value of |output|. If true, the |AppID| was used and thus, when [verifying an assertion](#verifying-assertion), the [=[RP]=] MUST expect the <code>[=rpIdHash=]</code> to be the hash of the |AppID|, not the [=RP ID=].
 
     <xmp class="idl">
     partial dictionary AuthenticationExtensionsClientOutputs {


nadalin:
@agl Please merge

PR #1031, opened 2018-08-09
Define "bound" for credentials and use the term "scoped" where appropriate

emlun:
This fixes some of #462:

```
CREDENTIAL:

(a credential)   bound to a authenticator
(a credential is)   bound to   an/this authenticator
                    managed by    "         "
                    present on    "         "
                    stored on     "         "
                    owning             authenticator
```

This PR leaves a few occurrences of "managed by authenticator" and "stored on authenticator" in introductory sections. The constructions "present on authenticator" and "owning authenticator" do not seem to appear anywhere in the spec at this time.

These terms in #462 seem to have already been fixed:

>scope, as in:
>- Public key credential's scope
>- strong, attested, scoped, public key-based credentials

This might also let us check off a few items in #358.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/pull/1031.html" title="Last updated on Aug 29, 2018, 12:29 AM GMT (48276cf)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/1031/d74f56b...48276cf.html" title="Last updated on Aug 29, 2018, 12:29 AM GMT (48276cf)">Diff</a>

PR #1048, opened 2018-08-27
PR 1031: Bound Credential <dfn>

emlun:
See https://github.com/w3c/webauthn/pull/1031#discussion_r212937971

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2184,7 +2184,7 @@ Note: The {{CollectedClientData}} may be extended in the future. Therefore it's 
                 </div>
 
             :   <dfn>id</dfn>
-            ::  This member MUST be present if {{TokenBinding/status}} is {{TokenBindingStatus/present}}, and MUST a [=base64url
+            ::  This member MUST be present if {{TokenBinding/status}} is {{TokenBindingStatus/present}}, and MUST be a [=base64url
                 encoding=] of the [=Token Binding ID=] that was used when communicating with the [=[RP]=].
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -485,8 +485,8 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
 : <dfn>Biometric Authenticator</dfn>
 :: Any [=authenticator=] that implements [=biometric recognition=].
 
-: Bound credential
-:: A [=public key credential source=] or [=public key credential=] is said to be <dfn>bound</dfn> to its [=managing
+: <dfn lt="bound">Bound credential</dfn>
+:: A [=public key credential source=] or [=public key credential=] is said to be [=bound=] to its [=managing
```
equalsJeffH: Yep, that's one way to do it -- another is: 
```
: <dfn>Bound credential</dfn>
:: A [=public key credential source=] or [=public key credential=] is said to be [=bound credential|bound=] to its ...
```
...so that if we end up using the term "[=bound credential=]" elsewhere, it just works

emlun: Hm, ok, I guess that seems fair. Let's go with that.

emlun:
(I accidentally pushed on the wrong branch, so GitHub didn't pick up commit d85bfbaa2906e13164892e94674c4f84456a3d1d which I added to this branch before merging into PR #1031)

PR #1050, opened 2018-08-29
Provide transport information during registration.

### LINKED ISSUE #851: Platform authenticators and key stores
Issue #420 (PR #708) details how a relying party who only wants platform authenticators can create authentication flows using the web authentication API:

* Relying party creates a credential with attachment=platform
* Relying party stores the credential ID (let's call it cred1) in HTML5 local storage
* Next time a user wants to authenticate, relying party passes [cred1] as allowCredentials in .get(), knowing that cred1 was created as a platform credential.

This method works fine most of the time. 

However, on some platforms, the key store holding the platform credential cred1 – or its decryption key, if it is a server-resident credential – may get wiped for a variety of reasons (e.g. last fingerprint removed, disabling / re-enabling screenlock, etc).

If we follow the WebAuthn specification as is, the following user flow will occur:

* Relying party creates a credential with attachment=platform
* Relying party stores the credential ID (cred1) in HTML5 local storage
* Day to day, relying party authenticates the user by passing [cred1] as allowCredentials in .get()
* One day, the user disables and re-enables the screenlock on their device. This deletes all the platform credentials in the device's key store, including cred1.
* Next time the relying party tries to authenticate the user, the platform will not know cred1 is a platform key, so the platform will display a UI asking the user to present a security key – which they never created.
  * User will try to cancel out of this security key UI
  * Relying party will get NotAllowedError which could mean anything from "timeout" to "user canceled authentication", so they can't tell that the user is in this state.
  * Thus, user will keep getting asked for a security key they never created every time they visit the relying party web site.

Is this a scenario that the working group cares about? If so, what are the recommendations to prevent the user experience described above?

### LINKED ISSUE #889: Return the employed AuthenticatorTransport when creating a new credential
@agl [originally wrote](https://github.com/w3c/webauthn/pull/882#issue-183773782):

FIDO U2F found it neccessary to wedge the authenticator transport in an
X.509 extension of the attestation certificate in order to communicate
this information to the RP.

In Webauthn, we currently note that it's possible that an RP might learn
this information from the attestation, but now have several kinds of
attestation and it seems dumb to define ways to wedge this information
in each.

Instead, have the client include the transport in the
AuthenticatorAttestationResponse. Also, define another transport type
for cases where a non-standard protocol is used to communicate with a
platform authenticator.

This issue pertains to PR #882 

assigning @agl to start because he originated PR #882

see also issues #851, #863, #867

agl:
This change adds a `getTransports` method to
`AuthenticatorAttestationResponse` that returns the
`AuthenticatorTransport` used to perform a registration, as well as
other transports that the user agent believes that the authenticator
supports.

Fixes #889
Fixes #851
See also #882

***
<a href="https://pr-preview.s3.amazonaws.com/agl/webauthn/pull/1050.html" title="Last updated on Sep 21, 2018, 9:20 PM GMT (52a5159)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/1050/035ea79...agl:52a5159.html" title="Last updated on Sep 21, 2018, 9:20 PM GMT (52a5159)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1194,12 +1194,14 @@ a numbered step. If outdented, it (today) is rendered as a bullet in the midst o
                         :   {{AuthenticatorAttestationResponse/attestationObject}}
                         ::  |attestationObject|
 
-                        :   {{AuthenticatorAttestationResponse/transports}}
-                        ::  A sequence of one or more {{AuthenticatorTransport}} values. The first value is set to the transport used to obtain this registration. Further values are appended to indicate other, distinct, transports that the |authenticator| is believed to support.
+                        :   {{AuthenticatorAttestationResponse/[[transports]]}}
+                        ::  A sequence of zero or more unique {{AuthenticatorTransport}} values, in lexicographical order, that the |authenticator| is believed to support.
 
-                            If a user agent does not wish to divulge this information it MAY choose not to include additional transports and MAY choose the first element to be a generic value rather than the actual transport used. But it MUST still include at least one value and it takes the risk that [=RP=] behaviour may be suboptimital.
+                            If a user agent does not wish to divulge this information it MAY substitute an arbitrary sequence designed to preserve privacy. This sequence MUST still be valid, i.e. lexicographically sorted and free of duplicates. For example, it may use the empty sequence. Either way, in this case the user agent takes the risk that [=[RP]=] behavior may be suboptimal.
 
-                            Note: How user agents discover other transports supported by a given [=authenticator=] is outside the scope of this specification, but may include information from an [=attestation certificate=] (for example [[FIDO-Transports-Ext]]), metadata communicated in an [=authenticator=] protocol such as CTAP2, or special-case knowledge about a [=platform authenticator=].
+                            If the user agent does not have any transport information, it SHOULD set this field to the empty sequence.
+
+                            Note: How user agents discover transports supported by a given [=authenticator=] is outside the scope of this specification, but may include information from an [=attestation certificate=] (for example [[FIDO-Transports-Ext]]), metadata communicated in an [=authenticator=] protocol such as CTAP2, or special-case knowledge about a [=platform authenticator=].
 
                     :   {{PublicKeyCredential/[[clientExtensionsResults]]}}
                     ::  A new {{ArrayBuffer}}, created using |global|'s [=%ArrayBuffer%=], containing the bytes of
@@ -1697,7 +1699,7 @@ during registration.
     ::  This operation returns the value of {{AuthenticatorAttestationResponse/[[transports]]}}.
 
     :   <dfn>\[[transports]]</dfn>
-    ::  This [=internal slot=] contains a sequence of one or more {{AuthenticatorTransport}} values. The first value is the transport used by the user agent to perform a {{PublicKeyCredential/[[Create]](origin, options, sameOriginWithAncestors)}} operation. Subsequent values are additional transports that the [=authenticator=] is believed to support. The sequence MUST NOT contain duplicate values.
+    ::  This [=internal slot=] contains a sequence of zero or more unique {{AuthenticatorTransport}} values in lexicographical order. These values are the transports that the [=authenticator=] is believed to support, or an empty sequence if the information is unavailable.
 </div>
 
 ### Web Authentication Assertion (interface <dfn interface>AuthenticatorAssertionResponse</dfn>) ### {#iface-authenticatorassertionresponse}
--- a/index.bs
+++ b/index.bs
@@ -1194,6 +1194,13 @@ a numbered step. If outdented, it (today) is rendered as a bullet in the midst o
                         :   {{AuthenticatorAttestationResponse/attestationObject}}
                         ::  |attestationObject|
 
+                        :   {{AuthenticatorAttestationResponse/transports}}
+                        ::  A sequence of one or more {{AuthenticatorTransport}} values. The first value is set to the transport used to obtain this registration. Further values are appended to indicate other, distinct, transports that the |authenticator| is believed to support.
+
+                            If a user agent does not wish to divulge this information it MAY choose not to include additional transports and MAY choose the first element to be a generic value rather than the actual transport used. But it MUST still include at least one value and it takes the risk that [=RP=] behaviour may be suboptimital.
+
+                            Note: How user agents discover other transports supported by a given [=authenticator=] is outside the scope of this specification, but may include information from an [=attestation certificate=] (for example [[FIDO-Transports-Ext]]), metadata communicated in an [=authenticator=] protocol such as CTAP2, or special-case knowledge about a [=platform authenticator=].
+
                     :   {{PublicKeyCredential/[[clientExtensionsResults]]}}
                     ::  A new {{ArrayBuffer}}, created using |global|'s [=%ArrayBuffer%=], containing the bytes of
                         <code>|credentialCreationData|.[=credentialCreationData/clientExtensionResults=]</code>.
@@ -1666,6 +1673,7 @@ during registration.
     [SecureContext, Exposed=Window]
     interface AuthenticatorAttestationResponse : AuthenticatorResponse {
         [SameObject] readonly attribute ArrayBuffer      attestationObject;
+        sequence<AuthenticatorTransport>                 getTransports();
     };
 </xmp>
 <div dfn-type="attribute" dfn-for="AuthenticatorAttestationResponse">
@@ -1684,6 +1692,12 @@ during registration.
         [=attestation statement=], as well as to decode and validate the [=authenticator data=] along with the
         [=JSON-serialized client data=]. For more details, see [[#sctn-attestation]], [[#generating-an-attestation-object]],
         and [Figure 5](#fig-attStructs).
+
+    :   {{AuthenticatorAttestationResponse/getTransports()}}
+    ::  This operation returns the value of {{AuthenticatorAttestationResponse/[[transports]]}}.
+
+    :   <dfn>\[[transports]]</dfn>
+    ::  This [=internal slot=] contains a sequence of one or more {{AuthenticatorTransport}} values. The first value is the transport used by the user agent to perform a {{PublicKeyCredential/[[Create]](origin, options, sameOriginWithAncestors)}} operation. Subsequent values are additional transports that the [=authenticator=] is believed to support. The sequence MUST NOT contain duplicate values.
 </div>
 
 ### Web Authentication Assertion (interface <dfn interface>AuthenticatorAssertionResponse</dfn>) ### {#iface-authenticatorassertionresponse}
@@ -5673,6 +5687,13 @@ for their contributions as our W3C Team Contacts.
     "title": "FIDO Privacy Principles",
     "href": "https://fidoalliance.org/wp-content/uploads/2014/12/FIDO_Alliance_Whitepaper_Privacy_Principles.pdf",
     "status": "FIDO Alliance Whitepaper"
+  },
+
+  "FIDO-Transports-Ext": {
+    "authors": ["FIDO Alliance"],
+    "title": "FIDO U2F Authenticator Transports Extension",
+    "href": "https://fidoalliance.org/specs/fido-u2f-v1.2-ps-20170411/fido-u2f-authenticator-transports-extension-v1.2-ps-20170411.html",
+    "status": "FIDO Alliance Proposed Standard"
   },
 
   "CDDL": {


agl:
This PR is not intended for PropRec, but rather for comments and for a future revision. It squashes some commits from #882 to make things clearer and easier to review. There are a few changes from that, older PR:

1. I've added some paragraphs about how a user agent might learn of additional transports and how Tor Browser might implement this, based on comments from JC on the previous PR.
1. Since we're returning a sequence of values, and `AuthenticatorAttestationResponse` is an interface, I've made `getTransports` a method that returns the value of an internal slot. This apes the whole discussion about whether we needed a method to return the extension results: since it's a sequence of transports I think we need a method here for the same reasons. Specifically, from [WebIDL](https://heycam.github.io/webidl/#idl-attributes):
> The type of the attribute, after resolving typedefs, must not be a nullable or non-nullable version of any of the following types: &hellip; a sequence type &hellip;

nadalin:
@akshayku please review and give OK as a general direction or post your issues why this si not the right direction

npdoty:
👍 that revealing this data only post authentication is a substantial mitigation. A couple questions:

1. How is the sequence of transports to be ordered (beyond the first transport, which is the one that was used for authenticating)? Non-functional ordering differences have been used in the past to fingerprint users without any functional benefit.

2. One suggested mitigation is to provide a "generic value", but it seems like `AuthenticatorTransport` is an enum without any generic values. Does the enum need an "unknown"/"generic" value added? Or does the IDL allow implementers to return a value not in the specified enumeration?

tomrittervg:
> after a final talk through with one of our Tor experts

Just to be clear, that was me, and I don't speak for Tor; I just gave an opinion. But the Tor folks have been emailed and may or may not weigh in themselves.

emlun:
>revealing this data only post authentication is a substantial mitigation.

To be precise, it doesn't look to me like that's what's being proposed. I'm guessing that what @jcjones meant by

>[...] this mechanism [...] is only available after the user explicitly intervenes w/ the authentication ceremony

is that it only happens after the user confirms consent to proceed with the _registration_ operation.

index.bs
```diff
@@ -1194,6 +1194,13 @@ a numbered step. If outdented, it (today) is rendered as a bullet in the midst o
                         :   {{AuthenticatorAttestationResponse/attestationObject}}
                         ::  |attestationObject|
 
+                        :   {{AuthenticatorAttestationResponse/transports}}
+                        ::  A sequence of one or more {{AuthenticatorTransport}} values. The first value is set to the transport used to obtain this registration. Further values are appended to indicate other, distinct, transports that the |authenticator| is believed to support.
+
+                            If a user agent does not wish to divulge this information it MAY choose not to include additional transports and MAY choose the first element to be a generic value rather than the actual transport used. But it MUST still include at least one value and it takes the risk that [=RP=] behaviour may be suboptimital.
```
akshayku: > generic value

Which Generic Value??

akshayku: > But it MUST still include at least one value

Why it is a MUST?

index.bs
```diff
@@ -1194,6 +1194,13 @@ a numbered step. If outdented, it (today) is rendered as a bullet in the midst o
                         :   {{AuthenticatorAttestationResponse/attestationObject}}
                         ::  |attestationObject|
 
+                        :   {{AuthenticatorAttestationResponse/transports}}
+                        ::  A sequence of one or more {{AuthenticatorTransport}} values. The first value is set to the transport used to obtain this registration. Further values are appended to indicate other, distinct, transports that the |authenticator| is believed to support.
```
akshayku: > The first value is set to the transport used to obtain this registration

This means that there is a value in providing what transport is used back to the RP. By that logic, does is makes sense that RP can choose which transport it wants to use in the first place as an input parameter for both MakeCredential and GetAssertion??

christiaanbrand: I don't think I care too much about the *specific* transport that was used during the operation, more about what's actually supported by the token, but if having both doesn't cause a privacy issue, I think it's good to have it.

>By that logic, does is makes sense that RP can choose which transport it wants to use in the first place as an input parameter for both MakeCredential and GetAssertion??
They can already choose during GetAssertion, but it's really more so the platform knows whether it can potentially satisfy the request (by doing an intersection between what it supports and what authenticators the user have registered).
During MakeCredential I'm not quite so sure it matters, apart from specifying the use case (which is why we have "attachment"). Is there a specific example which you think won't work using just "attachment" during create?

nadalin:
@akshayku Are you good now ?

akshayku:
Not yet. :)

index.bs
akshayku: > They can already choose during GetAssertion, but it's really more so the platform knows whether it can potentially satisfy the request (by doing an intersection between what it supports and what authenticators the user have registered).

No. You can't for username less flows as you don't know who the user is yet. That's why I always advocated for attachment in GetAssertion (which I will propose for next version as we ran out of time for this version). 

Adding a sequence of transport instead of attachment filter as a request option for both MakeCredential/GetAssertion makes it even more flexible. @christiaanbrand, @agl What do you think?

emlun: >The first value is set to the transport used to obtain this registration. Further values are appended to indicate other, distinct, transports [...]

I think we can leverage the types better to express this. I think that as a rule of thumb, particular items in a sequence should not be special because that speciality is not encoded in the type. I therefore suggest removing the significance of order in the list and instead adding a separate property that indicates the transport that was used this time. Something like:

```
{
  "transports": ["USB", "NFC"],
  "usedTransport": "USB"
}
```

This would also let us solve the issue of possible fingerprintability based on order, since we could then just specify that the list MUST be sorted lexicographically for example. The `usedTransport` member could also be optional instead of letting the client "choose the first element to be a generic value rather than the actual transport used".

Then again, the first element is special also in the `x5c` member of some attestation statement formats, so it's not a huge deal if we leave this as is.

emlun: In summary my position is that special cases inevitably cause bugs, so we should stay away from special cases whenever possible.

index.bs
emlun: s/[=RP=]/[=[RP]=]/

index.bs
```diff
@@ -1194,6 +1194,13 @@ a numbered step. If outdented, it (today) is rendered as a bullet in the midst o
                         :   {{AuthenticatorAttestationResponse/attestationObject}}
                         ::  |attestationObject|
 
+                        :   {{AuthenticatorAttestationResponse/transports}}
```
emlun: s/transports/[[transports]]/

agl:
(Replying to questions above. Will reply to inline questions separately.)

> How is the sequence of transports to be ordered (beyond the first transport, which is the one that was used for authenticating)? Non-functional ordering differences have been used in the past to fingerprint users without any functional benefit.

That's a good point. How about defining it to be lexicographical after the first element?

> One suggested mitigation is to provide a "generic value", but it seems like AuthenticatorTransport is an enum without any generic values. Does the enum need an "unknown"/"generic" value added? Or does the IDL allow implementers to return a value not in the specified enumeration?

At the highest level, there are two options for how a browser might withhold this information: one is to allow it a distinguished value (e.g. an empty list) which makes it clear that the value was withheld. The other is to have it provide a constant, plausible value which is unrelated to reality. I've chosen the latter here, but I really don't know what something like Tor Browser would prefer. I guess it's pretty obvious to sites that a client is Tor Browser so perhaps the former is preferred?

But, assuming the latter, the "generic value" is intended simply to communicate that the client may pick whatever they think provides the largest anonymity set. That may change over time so I didn't want to nail it down here. At the moment, I would suggest ["usb"] as probably the best option.

index.bs
agl: I disagree with this :) At the moment, there are two ways in which `transports` can be invalid: being empty or having duplicates. If the transport that was used is indicated by setting it as the value of a different member then there's a third possible problem: that the used transport isn't included in the list of transports.

Also, when this is exposed as API it has to be as a `getTransports` function because we can't have a sequence as a property of an interface. Since this isn't a feature of the baseline protocol, RP Javascript will have to test for `"getTransports" in response`, but now there'll be two properties to worry about. Or maybe return [usedTransport, transports] from `getTransports`? (But tuple-returns don't seem very common in Javascript.)

This is all pretty minor, but I do think that assigning semantics to the order of elements in a list is pretty standard and it's simpler to have a single list.

agl: > No. You can't for username less flows as you don't know who the user is yet. That's why I always advocated for attachment in GetAssertion.

I fear that I'm being slow and don't understand what you have in mind here. Would you mind expanding?

What I believe Christiaan has in mind here is the case where an RP requests an assertion and provides a credential that was created with a platform authenticator on another machine. If the RP can indicate that it's an &ldquo;internal&rdquo; transport then the browser can quickly establish that it doesn't have a matching private key and tell the user that. Otherwise it might prompt the user to insert their USB key or something, thus confusing them.

That obviously doesn't apply to usernameless logins however, so I imagine that you have some different scenario in mind that you want to address&mdash;I'm just not sure what it is.

index.bs
agl: See also the second part of my comment [here](https://github.com/w3c/webauthn/pull/1050#issuecomment-420714358), but basically I'm defining the transports list to be non-empty and telling browsers that they can fabricate plausible values if they want, but they can't return invalid ones.

agl: See also the second part of my comment [here](https://github.com/w3c/webauthn/pull/1050#issuecomment-420714358) but I'm deliberately not specifying the value. The idea is that browsers may fabricate a plausible list and the list with the largest anonymity set may change over time. Thus the best choice of fabrication may also change over time so I don't want to nail it down in the spec.

However, this is evidently a point where I've been unclear and I shall change the wording here to improve.

index.bs
```diff
@@ -1684,6 +1692,12 @@ during registration.
         [=attestation statement=], as well as to decode and validate the [=authenticator data=] along with the
         [=JSON-serialized client data=]. For more details, see [[#sctn-attestation]], [[#generating-an-attestation-object]],
         and [Figure 5](#fig-attStructs).
+
+    :   {{AuthenticatorAttestationResponse/getTransports()}}
+    ::  This operation returns the value of {{AuthenticatorAttestationResponse/[[transports]]}}.
+
+    :   <dfn>\[[transports]]</dfn>
+    ::  This [=internal slot=] contains a sequence of one or more {{AuthenticatorTransport}} values. The first value is the transport used by the user agent to perform a {{PublicKeyCredential/[[Create]](origin, options, sameOriginWithAncestors)}} operation. Subsequent values are additional transports that the [=authenticator=] is believed to support. The sequence MUST NOT contain duplicate values.
```
jcjones: Proposed per [https://github.com/w3c/webauthn/pull/1050/files#r216992616](emlun's comment): 
> Subsequent values are additional transports that the [=authenticator=] is believed to support, and MUST be sorted lexicographically.

akshayku:
@agl Let me try to explain the use cases

1. **With UserName Flow: User creates an platform credential on a mobile platform and want to use the same credential on other platforms.**
    - User tries to creates a platform/internal credential on one platform (Mobile). 
    - Platform creates the credential and tells RP all the transports supported by this credential (Lets say Internal, BLE, caBLE, NFC) in MakeCredential response. 
        - Whether we want to tell what transport was used is another question, but we are somewhat already telling this already in flags (2nd bit) for U2F authenticators.
    - User goes to another platform and types in his/her username
    - RP sees that the credentials created for the user can **_potentially_** be used over other transports in **_this platform_** and sends those credential list to the platform.
    - Platform can look for those credentials over other specified transports (NFC/BLE)

**_Spec changes required_**
- Supported transports needs to be passed back to the RP during MakeCredential

2. **With UserName Flow: User moving around between platforms and the RP only supports platform credentials**
    - User tries to create a platform/internal credential on a platform.
    - Platform creates the credential and tell the RP that the credential can only be used over internal transport as MakeCredential response.
    - User goes to another brand new machine and tries to login by entering the username.
    - RP sends all the known credentials for the user (which are all internal but platform may or may not have transports information as that is optional or not available in version 1 of the spec)
    - As Platform does not find any credential on this machine, with transports support, Platform can decide that this call in not going to succeed and tell the user about that and abort the operation.

**_Spec changes required_**
- Option 1: Supported transports needs to be passed back to the RP during MakeCredential 
OR
- Option 2: GetAssertion has a attachment property and RP sets that to "platform".

As we will have mixed authenticator credential registered between version 1 of the spec (which does not have MakeCredential transports response) and version 2 of the spec(which we are proposing), I would go for option 2 which works for all the cases.

3. **With UserName Flow: User moving around between platforms and the RP only supports cross-platform credentials**
    - User tries to create a cross-platform credential on a platform.
    - Platform creates the cross-platform credential and optionally tell the RP that the credential can only be used over internal transport as MakeCredential response.
    - User goes to another brand new machine and tries to login by entering the username.
    - RP sends all the known credentials for the user (which are all cross-platform but platform may or may not have transports information as that is optional or not available in version 1 of the spec)
    - If the transports information is available, platform only sees cross-platform credentials, Platform will not find any credential on this machine and can go directly to look for external authenticators. 

**_Spec changes required_**
- GetAssertion having an attachment property and RP sets that to "cross-platform". 
   - This gives flexibility to the platform to not even look at platform credentials and waste time/introduce unwarranted UIs. Then platform can directly go looking for external authenticators.

4. **Without UserName flow: User moving around between platforms. RP supports only platform credentials.** 
    - User goes to a machine and creates a platform credential.
    - User goes to another brand new machine and tries to login by clicking a link which kind of says "Login with platform or security keys". There is no username entry here. No cookies etc. 
    - As this is a brand new machine, there is no platform credential for this RP and platform does not know whether user wants to use platform credential or user has a roamable credential as this is username less flow. So platform has to go to look for external authenticators. 
    - Above step is very confusing for the user as he/she never had an external security keys. 
    - Adding a page which says no platform credentials are found and having a link which says go to external authenticators introduces a very bad friction for below use case. Having a link in itself where RP does not support about external devices is also not desired by the RPs as a feedback given to us.

**_Spec changes required_**
- GetAssertion has a attachment property and RP sets that to "platform".


5. **Without UserName flow: User moving around with external security keys. RP supports only cross-platform credentials.** 
    - User goes to a machine and creates a cross-platform credential.
    - User goes to another brand new machine and tries to login by clicking a link which kind of says "Login with platform or security keys". There is no username entry here. No cookies etc. 
    - As this is a brand new machine, there is no platform credential for this RP and platform does not know whether user wants to use platform credential or user has a roamable credential as this is username less flow. So platform has to go to look for external authenticators. 
    - Adding a page which says no platform credentials are found and having user clicking a link is a very bad user experience for a user/RP who only wants to deal with external security keys for any reason. For this reason, we are not introducing that page as it slows down the user login and that is not what user/RP wants and has given feedback to us.

**_Spec changes required_**
- GetAssertion has a attachment property and RP sets that to "cross-platform".


**_Summary_**
- Having a attachment property in GetAssertion is very important and absolutely required for various scenario especially with username less flows. It solves various ambiguous UI issues. We have already given flexibility of choosing platform vs cross-platform in MakeCredential 

- If we want to solve use case 1, then sending transports back to the RP is also required. For use case 2 of not showing external authenticators in **_identifiable platform only with username_** case, that can be solved better with attachment property in GetAssertion.

index.bs
akshayku: I prefer emlun approach. This is much cleaner. @agl, I still don't understand the "Generic Value" part of the PR. Can you explain what that value is as we cannot use "usb" as Generic Value.

emlun: @agl There would be equally many (2) invalid configurations, because the empty list would be valid. Even if this means there's two properties to worry about, I argue that it _reduces_ complexity because it _increases_ explicitness. And I'm not convinced that just because something is common practice that means it's good practice. :)

christiaanbrand:
Let me go back to the fundamental reason why we introduced an attachment property into the spec at the first place: the whole idea was that RPs can indicate which _use case_ they're trying to solve for. Either bootstrapping = x-plat, or reauthentication = platform.

During Get things are a little different: here we don't want to bother user with UI that has no relevance to them. This would include things like giving the user an option to insert an external authentication when an RP knows for a fact that they only ever created "platform" credentials (and never got a platform credential back that was annotated with a x-platform transport), or, giving the user the option to use an authenticator which is simply not available on this platform (for example, registering a BLE token on one machine, and then trying to use it on a machine without BLE. No point in asking the user to present that token).

I also don't really have sympathy for an argument which says: "but v1 of the spec doesn't include transports during MakeCredential, so we now have to do some weird, subpar workaround". We called this omission out in the working group as soon as we discovered it, and it was decided that it'll mess up the whole standardization process if we want to force it into v1. For that reason, we were okay deferring it to level 2 (or, really, version 1.0.1 - it really should be an extremely fast-follow), but we will pretend like it was just always there. If I knew that we were going to have to handle this in some backwards compatible fashion I would have insisted that we fix this before pushing the spec out.

Alright, now with all of the disclaimers out of the way, let's delve in to the 5 use cases.

Use case 1: Agreed
Use case 2: I prefer option 1. It gives us more flexibility regarding making decisions around what actual transports the system supports _at this point in time_. As I said earlier, there's no point in showing a user a dialog to "insert their external security key" if we know the user only has an external NFC token, and the platform we're on doesn't support that specific transport.
Use case 3: Having the transport information works fine for this use case too, right? You'll just not have "internal" as any of your transports in the credentialIDs.

Use case 4 and 5: These are basically the same, use case: Essentially, what you're saying is that, since transports information is a property of the credentialID, if there are no credentialIDs, we can't optimize the UX. I agree. I think the generic UI we had in mind actually worked pretty well for both these cases. Here was my thinking:
When doing an empty allowList, the browser will render an account chooser. If there are any credentials _locally on the system_ which can satisfy the request, they will be pre-populated in the list (I can see no reason why, if there are credentials, that an RP would willingly want to exclude them). If there are no credentials, the user will see an empty account selector. In both cases there will be a message below that says: "Please insert a security key with credentials for this RP" (or something to that effect) and the account chooser will then be populated as this happens. Granted, you might have had more discussions with RPs around the feasbility of this option, and if you absolutely believe that this won't work (I think we'll still try it), I can see a case for introducing the "attachment" property here, but it would **simply** be for the case where an empty allowList is specified. Of course, I'd still have to talk with the rest of the Google team to get their input here... especially since we'll probably handle both "attachment" cases exactly the same in order to _motivate_ RPs to not support only one of these use cases.

Makes sense?
P.S. I don't know where we've settled on the emlun and agl discussion. I don't really have any input there.

christiaanbrand:
So, folks. What's the verdict here? Are we good to merge?

equalsJeffH:
I would like to grok @akshayku's use cases and @christiaanbrand's reply in relation to the proposed text (which itsefl could use some modest polishing it seems on read-thru), hopefully can do this tonight or tomorrow morning.

nadalin:
@christiaanbrand This is a L2 PR so there will not me a merge until we finish L1

index.bs
agl: The empty list would not be valid based on the above. However, I think we can solve this by dropping the used transport completely. I had included it because it seemed that it might be useful, but we don't have a concrete use. Without it, the list is just a sorted list of transports. I think everyone will be happy with that and I'll upload a version with that in a sec.

I have made the empty list valid for a different reason however: I suspect that this information may not be valid in some cases. For example, platform APIs written against the first version of webauthn may not include this information and so the assumption that at least the transport used to register a credential is known, might not hold. Thus I've defined the empty list to mean handle that case.

index.bs

index.bs
agl: (This thread was continued at the top-level, below.)

index.bs

index.bs

index.bs
agl: The term "generic value" has been dropped.

agl:
@akshayku thank you for enumerating the cases that like.

Firstly, I think we're in agreement that passing the transports back to the RP during registration is necessary in order to address important use cases. However, I think your case 1 is too limited:

In your case 1, you consider an RP filtering credentials by transport in order to discover whether any could potentially be used by a user logging into a new platform. But the RP doesn't necessarily know whether the login is actually on a new platform, or whether it's the old platform with cookies cleared. Thus that decision is best deferred to the platform because only it can test whether a platform credential is locally available.

Next, the same pattern of the platform being able to act more intelligently when in possession of transport information about a credential repeats when considering not just platform _vs_ non-platform, but which transports are available. If the platform knows that a given credential is BLE-only, and it doesn't have a BLE adaptor, then there's no point bothering with it.

I believe the above argues for providing transport information for each credential when requesting an assertion. Thankfully, that's in the current spec ([here](https://w3c.github.io/webauthn/#dictdef-publickeycredentialdescriptor)). Additionally, since `internal` is now a valid transport type, this mechanism also subsumes the need to provide an attachment property when getting an assertion and so I believe that your suggested spec change there is already taken care of.

&nbsp;

More generally with your classification, you've split RPs into those that only accept platform credentials and cross-platform ones. But I don't see why any such RPs should exist. I think Christiaan is correct here to emphasise the two overarching uses: blessing new machines (i.e. bootstrapping) and reauthentication.

The latter is currently embodied by re-prompting for a password before doing significant account actions, even when logged in. The goal is to ensure that it's still the correct person behind the keyboard and this implies a user-verifying authenticator.

Blessing new machines implies using an external authenticator, which could imply a cross-platform authenticator. But that's fuzzy because platform-attached authenticators can be exposed by the platform over BLE etc. (See caBLE.) So I don't believe any RP would choose to accept only platform or cross-platform authenticators.

But, while I'm missing something there, I think the important thing is to focus on where I think we agree: we need transports to be provided during registration, and credentials should be able to indicate their attachment when getting an assertion&mdash;which is already handled in the spec by including transports in [PublicKeyCredentialDescriptor](https://w3c.github.io/webauthn/#dictdef-publickeycredentialdescriptor).

index.bs
rlin1: the requirement to sort ("and MUST be sorted lexicographically") doesn't seem to be reflected in the spec yet .

jcjones: > the requirement to sort ("and MUST be sorted lexicographically") doesn't seem to be reflected in the spec yet .

It doesn't have a `MUST`, but the list is defined that way each time. As that language is normative anyway, it'd be wrong to use ordering other than lexicographical. I think it's OK as-is, but feel free to object @rlin1, it'd be fine by me to use a `MUST` here, even if I don't think it's required.

jcjones:
(My approval is noted in https://github.com/w3c/webauthn/pull/1050#discussion_r227310751)

index.bs
```diff
@@ -1194,6 +1194,15 @@ a numbered step. If outdented, it (today) is rendered as a bullet in the midst o
                         :   {{AuthenticatorAttestationResponse/attestationObject}}
                         ::  |attestationObject|
 
+                        :   {{AuthenticatorAttestationResponse/[[transports]]}}
+                        ::  A sequence of zero or more unique {{AuthenticatorTransport}} values, in lexicographical order, that the |authenticator| is believed to support.
+
+                            If a user agent does not wish to divulge this information it MAY substitute an arbitrary sequence designed to preserve privacy. This sequence MUST still be valid, i.e. lexicographically sorted and free of duplicates. For example, it may use the empty sequence. Either way, in this case the user agent takes the risk that [=[RP]=] behavior may be suboptimal.
```
emlun: Is it still valuable to allow the client to return an arbitrary bogus sequence? Why not require the client to return either the true sequence, or an empty sequence if it wishes to keep the true sequence private?

emlun: So the edit I'm suggesting is to change this:

>If a user agent does not wish to divulge this information it MAY substitute an arbitrary sequence designed to preserve privacy. This sequence MUST still be valid, i.e. lexicographically sorted and free of duplicates. For example, it may use the empty sequence. Either way, in this case the user agent takes the risk that [=[RP]=] behavior may be suboptimal. 

to this:

>If a user agent does not wish to divulge this information it MAY set this to the empty sequence. In this case the user agent takes the risk that [=[RP]=] behavior may be suboptimal.

agl: If the client wishes to blend into the crowd, it would seem best to give it flexibility? Indeed, even if the spec says that the values must either be completely real or else empty, I expect that clients that wished to maintain privacy would probably do whatever they thought least conspicuous and most compatible irrespective of what the spec says.

I think @jcjones is the closest that we have to a rep for Tor Browser here, and I'll defer to him on this point if he has an opinion, but I'm not yet convinced about tightening this.

tomrittervg: Tor Browser historically chooses a value that will not cause breakage but allows them to look like the most popular <thing> in question. (E.g. choosing a version of the OS that is most popular.)  So I agree that if the spec requires either 'completely real or empty' that would probably be ignored.

emlun: Ok, that's enough for me. :)

equalsJeffH:
Shall we go ahead and merge this now?

PR #1052, opened 2018-09-04
fix #403: user handle - account relationship

### LINKED ISSUE #403: some RPs may wish to allow multiple registrations to same user account
we have this text in [4.3. User Account Information](https://w3c.github.io/webauthn/#iface-userinfo):
> The id member contains an identifier for the account, specified by the Relying Party. ... It is used by the Relying Party to control the number of credentials - an authenticator will never contain more than one credential for a given Relying Party under the same id.

This *could* be construed to mean that an RP may only map one credential to a given user account. Tho, if the RP is creative, and e.g. does not store their internal account identifier in the RelyingPartyUserInfo .id but rather stores some 2nd-level id there that they map to their account ID on their server-side, they can do so (yes?) -- seems to me we ought to document this.

equalsJeffH:
fixes #403

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -671,12 +671,13 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
     An [=authorization gesture=] is a [=ceremony=] component often employed to indicate [=user consent=].
 
 : <dfn>User Handle</dfn>
-:: The user handle is specified by a [=[RP]=] and is a unique identifier for a user account with that [=[RP]=]. A user handle is
-    an opaque [=byte sequence=] with a maximum size of 64 bytes.
+:: The user handle is specified by a [=[RP]=] and used to map a specific [=public key credential=]
+    to a specific user account with the [=[RP]=]. Authenticators in turn [=credentials map|map=] [=RP ID=]s and user handle pairs
+    to [=public key credential sources=].
 
-    The user handle is not meant to be displayed to the user, but is used by the [=[RP]=] to control the number of credentials - an
-    authenticator will never contain more than one credential for a given [=[RP]=] under the same user handle.
+    Note: [=[RPS]=] may map multiple [=public key credentials=] to a user account by mapping multiple user handles to the account.
 
+    A user handle is an opaque [=byte sequence=] with a maximum size of 64 bytes. User handles are not meant to be displayed to users.
     The user handle SHOULD NOT contain [PII] about the user, such as a username or e-mail address; see
     [[#sctn-user-handle-privacy]] for details.
 


index.bs
```diff
@@ -671,12 +671,13 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
...
 : <dfn>User Handle</dfn>
-:: The user handle is specified by a [=[RP]=] and is a unique identifier for a user account with that [=[RP]=]. A user handle is
-    an opaque [=byte sequence=] with a maximum size of 64 bytes.
+:: The user handle is specified by a [=[RP]=] and used to map a specific [=public key credential=]
+    to a specific user account with the [=[RP]=]. Authenticators in turn [=credentials map|map=] [=RP ID=]s and user handle pairs
+    to [=public key credential sources=].
 
-    The user handle is not meant to be displayed to the user, but is used by the [=[RP]=] to control the number of credentials - an
-    authenticator will never contain more than one credential for a given [=[RP]=] under the same user handle.
+    Note: [=[RPS]=] may map multiple [=public key credentials=] to a user account by mapping multiple user handles to the account.
```
emlun: Perhaps this could be better expressed as something like this?

>Note: [=[RPS]=] MAY map multiple [=public key credentials=] to one user account by supplying the same [=user handle=] as the \<code>{{PublicKeyCredentialCreationOptions/user}}.{{PublicKeyCredentialUserEntity/id}}\</code> argument in multiple calls to {{CredentialsContainer/get()}}.

emlun: I tried putting that in to see that it compiles, so I suppose I might as well push it as well: #1053

equalsJeffH: Hm, maybe I'm missing something, but given how the authnr's [credentials map](https://w3c.github.io/webauthn/#authenticator-credentials-map) is defined -- a (rpId, \[[userHandle](https://w3c.github.io/webauthn/#public-key-credential-source-userhandle)\]) pair yields a [credential source](https://w3c.github.io/webauthn/#public-key-credential-source) -- does that work?

emlun: Ah, good point. So that would be

>...MAY map multiple [=public key credential sources=] to one user account...

then?

emlun:
As we discussed on the 2018-09-05 WG call, it seems like we were discussing past each other a bit. The bit I find confusing about this PR is this (emphasis added):

>Note: [=[RPS]=] may [=credentials map|map=] multiple [=public key credentials=] to a user account **by mapping multiple user handles to the account**.

The points I was trying to make were:

- It's not necessary to have multiple user handles for the same account in order to have multiple credentials for that account
- Although one authenticator can indeed only have one credential for a given (rpId, userHandle) pair, it's possible to create another credential with the same (rpId, userHandle) pair on a different authenticator.
- Different user accounts are expected (by my estimation) to have different user handles.

The main counterpoint was:

- It's concievable that RPs may not want to use user handles, and opt to set the user handle to, for example, the empty byte array for all users. In this case it's not possible to have credentials for different accounts on the same authenticator, since they share the same user handle. The RP can work around this by mapping multiple user handles to the same account.

Would it be possible (and desirable) to get both of these subtleties across?

equalsJeffH:
thx @emlun .  Given our short timeframe for PropRec, perhaps we can just delete the Note from this PR,  not do PR #1053 (for now), and merge the remainder of the clarification of the user handle definition (as it presently is), and punt the subtleties to REC or Level 2 ?

in any case, WRT subtleties:
 I'm thinking that an _additional_ counterpoint is that an RP may wish to allow users to create multiple credentials on the same authnr (a platform authnr, say) that map to the same RP user account.  More generally, unless I'm misunderstanding things, it is an RP's decision whether to map user handle and RP user account as 1:1 or n:1. In the former case the RP could have the user handle equal RP user account identifier, in the latter case the RP would not.  Is this possible for RPs to do now if they wish?  Admittedly this is an edge case (?).

emlun:
Ok, I'm fine with punting my remaining comments to L2. I'll close #1053.

>RP may wish to allow users to create multiple credentials on the same authnr (a platform authnr, say) that map to the same RP user account

I really can't see how that would ever be useful - it would be like having two identical keys on the same keyring. If you're imagining, say, multiple users on the same laptop, then I'd say it's rather up to the OS to model separate OS accounts/fingerprints/whatever as different sub-authenticators (so that the "one key per (rpId, userHandle, authenticator)" rule still holds)... but I'll concede I might be arguing about irrelevant semantics here.

Anyway, please go ahead as you suggested. :)

equalsJeffH:
I'd proposed:

>Given our short timeframe for PropRec, perhaps we can just delete the Note from this PR, not do PR #1053 (for now), and merge the remainder of the clarification of the user handle definition (as it presently is),

@emlun replied:
> please go ahead as you suggested. :)

Ok, thx, will do.

> If you're imagining, say, multiple users on the same laptop, then I'd say it's rather up to the OS to model separate OS accounts/fingerprints/whatever as different sub-authenticators (so that the "one key per (rpId, userHandle, authenticator)" rule still holds)

yeah, that's another aspect of these subtleties it seems.

equalsJeffH:
merge this now?

nadalin:
@equalsJeffH Please merge

PR #1082, opened 2018-09-20
Clarify user identification in RP assertion verification operation

### LINKED ISSUE #1078: Clarify which user to authenticate if userHandle is not present
Maybe one should clarify Step 2 of [Verifying an authentication assetion](https://www.w3.org/TR/webauthn/#verifying-assertion) for the case, that userHandle is not present:

>2. If credential.response.userHandle is present, verify that the user identified by this value is the owner of the public key credential identified by credential.id.
**If credential.response.userHandle is not present, the owner of the public key credential identified by credential.id is the user to authenticate.**

Greetings, 
Miles

emlun:
This fixes #1078. Could arguably be technical and/or breaking, or arguably just editorial implementation consideration.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/pull/1082.html" title="Last updated on Dec 13, 2018, 12:18 PM GMT (6b3389c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/1082/daf9522...6b3389c.html" title="Last updated on Dec 13, 2018, 12:18 PM GMT (6b3389c)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -3501,9 +3501,19 @@ When verifying a given {{PublicKeyCredential}} structure (|credential|) and an {
     initiated, verify that <code>|credential|.{{Credential/id}}</code> identifies one of the [=public key credentials=] that were
     listed in {{PublicKeyCredentialRequestOptions/allowCredentials}}.
 
-1. If <code>|credential|.{{PublicKeyCredential/response}}.{{AuthenticatorAssertionResponse/userHandle}}</code> is present, verify
-    that the user identified by this value is the owner of the [=public key credential=] identified by
-    <code>|credential|.{{Credential/id}}</code>.
+1. Identify the user being authenticated and verify that this user is the owner of the [=public key credential source=]
+    |credentialSource| identified by <code>|credential|.{{Credential/id}}</code>:
+
+    <dl class="switch">
+        :  If the user was identified before the [=authentication ceremony=] was initiated,
+        :: verify that that user is the owner of |credentialSource|. If
+            <code>|credential|.{{PublicKeyCredential/response}}.{{AuthenticatorAssertionResponse/userHandle}}</code> is present,
+            verify that this value identifies the same user as was previously identified.
+
+        :  If the user was not identified before the [=authentication ceremony=] was initiated,
+        :: Verify that <code>|credential|.{{PublicKeyCredential/response}}.{{AuthenticatorAssertionResponse/userHandle}}</code> is
+            present, and that the user identified by this value is the owner of |credentialSource|.
+    </dl>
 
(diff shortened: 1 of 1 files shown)


emlun:
This can probably be punted to later if need be.

index.bs
```diff
@@ -3501,9 +3501,19 @@ When verifying a given {{PublicKeyCredential}} structure (|credential|) and an {
     initiated, verify that <code>|credential|.{{Credential/id}}</code> identifies one of the [=public key credentials=] that were
     listed in {{PublicKeyCredentialRequestOptions/allowCredentials}}.
 
-1. If <code>|credential|.{{PublicKeyCredential/response}}.{{AuthenticatorAssertionResponse/userHandle}}</code> is present, verify
-    that the user identified by this value is the owner of the [=public key credential=] identified by
-    <code>|credential|.{{Credential/id}}</code>.
+1. Identify the user being authenticated and verify that this user is the owner of the [=public key credential source=]
```
equalsJeffH: The RP _does not_ itself have access to the [=public key credential source=], yes?   Should that actually be [=public key credential=]?

equalsJeffH: wrt "identify the user":  I'm thinking we ought to clarify that we mean "identifying the user account in play", and that by this we mean utilizing whatever session maintenance facility the RP employs, eg, cookies, etc.

equalsJeffH:
I'm thinking we ought to just merge this and submit an issue to clean up the terminology in L2.  also ought to submit issue that this RP Ops section lacks description of Registration and Authn ceremonies initiations.

index.bs
emlun: Indees the RP does not, but I figured that `PublicKeyCredential.id` identifies a _credential source_ and not a _credential_. See top-level comment for more...

emlun:
Sorry for my late reply...

Continuing from https://github.com/w3c/webauthn/pull/1082#discussion_r221568308...

The definition of **Credential ID** reads (emphasis added):

>A probabilistically-unique byte sequence identifying a **public key credential source** and its authentication assertions.

So I think it's more appropriate to say "the public key credential source identified by |credential.id|" than "the public key credential identified by |credential.id|".

I agree it's probably a good idea to add "Initiating a(n) registration/authentication ceremony" subsections to the RP Operations section.

emlun:
However, the definition of `PublicKeyCredentialDescriptor.id` reads (emphasis added)

>id, of type BufferSource
>This member contains the credential ID of the **public key credential** the caller is referring to.

so maybe we should align that with the definition of Credential ID.

emlun:
@equalsJeffH Would you say we're ready to merge this? @nadalin @YubicoDemo Can we do that now or would we have to wait until after PR or until L2?

I would appreciate a review from @selfissued as well.

PR #1108, opened 2018-10-31
clarification of UP/UV flags in authenticator data structure

jericks-duo:
The User Presence (UP) and User Verification (UV) flags in the authenticator data structure (https://www.w3.org/TR/webauthn/#sec-authenticator-data) appear to have a similar purpose to the requireUserPresence and requireUserVerification input parameter booleans in the authenticatorMakeCredential operation. The requireUserPresence and requireUserVerification booleans are explicitly mutually exclusive -- if one is set the other must be unset. My understanding, after discussing the use case for the UP/UV flags, is that both MAY be set (i.e. not mutually exclusive). 

Example: The relying party may specify that user presence is required, but the authenticator may physically perform a user verification operation. In this case, the relying party may end up checking the UP flag and not the UV flag, so it seems like the authenticator should set both flags, not just the UV flag.

Just wanted to clarify this in the doc as there may be the potential for confusion during implementation. Or alternately, if there is a reason they should be mutually exclusive, the spec should probably specify that.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2576,7 +2576,7 @@ the requested [=public key credential|credential=] is [=scoped=] to exactly matc
 - Hash [=RP ID=] using SHA-256 to generate the [=rpIdHash=].
 
 - The `UP` [=flag=] SHALL be set if and only if the authenticator detected a user through an authenticator specific gesture. The
-    `RFU` bits SHALL be set to zero.
+    `RFU` bits SHALL be set to zero. The `UV` and `UP` [=flag=]s MAY both be set if the authenticator verified a user.
 
 - For [=attestation signatures=], the authenticator MUST set the AT [=flag=] and include the <code>[=attestedCredentialData=]</code>. 
     For [=assertion signature|authentication signatures=], the AT [=flag=] MUST NOT be set and the <code>[=attestedCredentialData=]</code> MUST NOT be included.


wseltzer:
@jericks-duo Are you affiliated with a W3C Member? If not, are you able to make a royalty-free patent licensing commitment on your contribution?

jericks-duo:
@wseltzer Hey Wendy. I am able to make such a commitment. I don't need or want any ownership. Just wanted to contribute a clarification.

Thanks,
- Jeremy

emlun:
Thanks for your proposal! I agree this is worthwhile clarifying, but I think it needs to be formulated differently to not open more opportunities for misunderstandings. With your proposal, the paragraph reads:

>The `UP` flag SHALL be set if and only if the authenticator detected a user through an authenticator specific gesture. The `RFU` bits SHALL be set to zero. The `UV` and `UP` flags MAY both be set if the authenticator verified a user.

This could be interpreted to mean that if the authenticator performed user verification, then it MAY set the `UV` flag. This should be that the `UV` flag MUST be set in that case, but that `UP` and `UV` MAY both be set at the same time. I'll draft an alternative proposal with reformulations to that effect.

emlun:
Alternate proposal: #1113

PR #1140, opened 2019-01-18
Let requireUserPresence always be true in authenticator operations

### LINKED ISSUE #1123: Clarify relationships between "uv"/"up" of the CTAP spec and "userVerification"/"userPresence"
In the CTAP spec (https://fidoalliance.org/specs/fido-v2.0-id-20180227/fido-client-to-authenticator-protocol-v2.0-id-20180227.html), we have the following tables:

WebAuthN authenticatorMakeCredentialoperation | CTAP authenticatorMakeCredential operation
--- | ---
`requireUserPresence` | Not present in the current version of CTAP. Authenticators are assumed to always check user presence.
`requireUserVerification` | `options.uv` or `pinAuth`/`pinProtocol`

WebAuthN authenticatorGetAssertion operation | CTAP authenticatorGetAssertion operation
--- | ---
`requireUserPresence` | `options.up`
`requireUserVerification` | `options.uv` or `pinAuth`/`pinProtocol`

That to me means `userVerification`/`userPresence` = `uv`/`up`.

Then we have this in the spec: https://www.w3.org/TR/webauthn/#getAssertion

>Let _userPresence_ be a Boolean value set to the inverse of _userVerification_.

I interpret this sentence as userVerification includes userPresence. When userVerification is required, userPresence is also required implicitly. Hence, the response from the authenticator will only have the following two valid combinations between `uv` and `up` ({uv, up}):

- {1, 1}
- {0, 1}

However, in the CTAP spec, it looks like requiring uv doesn't necessary mean requiring up.
In section 5.2, Step 7:

>Collect user consent if required. This step MUST happen before the following steps due to privacy reasons (i.e., authenticator cannot disclose existence of a credential until the user interacted with the device):
>- If the "uv" option was specified and set to true:
>   -  If device doesn’t support user-identifiable gestures, return the CTAP2_ERR_UNSUPPORTED_OPTION error.
>     - Collect a user-identifiable gesture. If gesture validation fails, return the CTAP2_ERR_OPERATION_DENIED error.
>- If the "up" option was specified and set to true, collect the user’s consent.
>   - If no consent is obtained and a timeout occurs, return the CTAP2_ERR_OPERATION_DENIED error.

I interpret the above as the authenticator will check `uv` and `up` independently according to the request's parameters.

If user agents send the request as suggested by both spec when userVerification is required from the RP, then an authenticator could send a response that includes the {uv, up} as {1, 0}.

So this really confuses me. I think we need some clarifications on either the WebAuthN spec or the CTAP spec. I will suggest we change the WebAuthN spec to:
"Let _userPresence_ be a Boolean value set to the inverse of _userVerification_." => "Let _userPresence_ be true."

emlun:
Fixes #1123.

An alternative solution to this would be to completely delete the `requireUserPresence` pseudo-parameters. I've left them in for now to make the mapping to CTAP slightly clearer.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/pull/1140.html" title="Last updated on Jan 18, 2019, 5:10 PM UTC (d9de125)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/1140/c610a95...d9de125.html" title="Last updated on Jan 18, 2019, 5:10 PM UTC (d9de125)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1153,8 +1153,6 @@ a numbered step. If outdented, it (today) is rendered as a bullet in the midst o
 
                     </dl>
 
-            1. Let |userPresence| be a Boolean value set to the inverse of |userVerification|.
-
             1. Let |excludeCredentialDescriptorList| be a new [=list=].
 
             1. [=list/For each=] credential descriptor |C| in <code>|options|.{{PublicKeyCredentialCreationOptions/excludeCredentials}}</code>:
@@ -1167,7 +1165,6 @@ a numbered step. If outdented, it (today) is rendered as a bullet in the midst o
                 |clientDataHash|,
                 <code>|options|.{{PublicKeyCredentialCreationOptions/rp}}</code>, <code>|options|.{{PublicKeyCredentialCreationOptions/user}}</code>,
                 <code>|options|.{{PublicKeyCredentialCreationOptions/authenticatorSelection}}.{{AuthenticatorSelectionCriteria/requireResidentKey}}</code>,
-                |userPresence|,
                 |userVerification|,
                 |credTypesAndPubKeyAlgs|,
                 |excludeCredentialDescriptorList|,
@@ -1502,8 +1499,6 @@ When this method is invoked, the user agent MUST execute the following algorithm
 
                     </dl>
 
-            1. Let |userPresence| be a Boolean value set to the inverse of |userVerification|.
-
             1. <span id="allowCredentialDescriptorListCreation"></span>
                 If <code>|options|.{{PublicKeyCredentialRequestOptions/allowCredentials}}</code>
                 <dl class="switch">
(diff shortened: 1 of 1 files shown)


emlun:
Would device authentication in IoT applications be running in a browser, though? We also have already foreclosed the possibility of silent authentication by always requiring at least one of UP and UV.

akshayku:
@selfissued This is fixing existing discrepency in the spec of what is available today. As and when, if we ever, allow silent authentication, we have to look at that in a broad sense and its use cases.

emlun:
Decided on 2019-02-13 WG call to go ahead with merging this if @equalsJeffH and @jcjones approve.

emlun:
As previously noted by @akshayku and @agl, this PR does not change our current stance on silent authentication - it only clarifies the present policy.

PR #1143, opened 2019-01-18
Determine appid extension output after authenticator returns

### LINKED ISSUE #1034: appid extension output is inaccurate in some corner cases
The client extension processing for the [`appid` extension][appid] reads (bold emphasis added):

>**When building _allowCredentialDescriptorList_**, if a U2F authenticator indicates that a credential is inapplicable (i.e. by returning `SW_WRONG_DATA`) then the client MUST **retry with the U2F application parameter set to the SHA-256 hash of _appId_. If this results in an applicable credential**, the client MUST include the credential in _allowCredentialDescriptorList_ and **set _output_ to `true`**. The value of appId then replaces the rpId parameter of authenticatorGetAssertion.

The value of _output_ is thus decided _before_ the client decides which assertion to return, if there are multiple authenticators present. Consider the following scenario:

1. The user has a U2F authenticator A plugged in, which has been registered via the U2F API (i.e., with AppID).
2. The user has a CTAP2 authenticator B plugged in, which has been registered via the WebAuthn API (i.e., with RP ID).
3. The user initiates an authentication ceremony and the RP sets the `appid` extension.
4. The client runs the above client processing and discovers that authenticator A does not contain a credential for the RP ID, and retries with the AppID. This succeeds, and the client sets the extension _output_ to `true`.
5. The client initiates authentication requests with both authenticator A and B, which both prompt the user for consent.
6. The user confirms user consent on authenticator B, which generates an assertion for the RP ID.
7. The client returns the assertion for the RP ID and the `appid` client extension output set to `true`.

So it's possible for the extension output to end up being `true` even though the RP should verify the assertion using the RP ID and not the AppID.

Practical impact
---
Likely low. It's a corner case with pretty specific requirements, and the [RP should arguably know from the credential ID][comment] which of the RP ID/AppID it needs to use.

[appid]: https://pr-preview.s3.amazonaws.com/agl/webauthn/pull/1033.html#sctn-appid-extension
[comment]: https://github.com/w3c/webauthn/issues/980#issuecomment-411884522

emlun:
Fixes #1034.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/pull/1143.html" title="Last updated on Jan 18, 2019, 7:01 PM UTC (776b7b1)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/1143/c610a95...776b7b1.html" title="Last updated on Jan 18, 2019, 7:01 PM UTC (776b7b1)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1579,7 +1579,8 @@ When this method is invoked, the user agent MUST execute the following algorithm
         :   If any |authenticator| indicates success,
         ::  1.  [=set/Remove=] |authenticator| from |issuedRequests|.
 
-            1.  Let <var ignore>assertionCreationData</var> be a [=struct=] whose [=items=] are:
+            1.  <span id="assertionCreationDataCreation"></span>
+                Let <var ignore>assertionCreationData</var> be a [=struct=] whose [=items=] are:
 
                 :   <code><dfn for="assertionCreationData">credentialIdResult</code>
                 ::  If <code>|savedCredentialIds|[|authenticator|]</code> exists, set the value of [=credentialIdResult=] to be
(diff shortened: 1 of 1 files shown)


nadalin:
@akshayku  please review

PR #1144, opened 2019-01-18
Make appid extension always return true

emlun:
This is an extension to PR #1143 and would merge into that.

I expect this might be a controversial proposal which might be closed with no action. Below is the motivation for it, which is also included as a commit message.

---

This greatly simplifies client implementation logic while leaving RP implementation arguably unaffected. The argument for the latter is as follows.

The [PropRec version of the spec][proprec] has some corner cases where the extension output could be `true` although the RP would in fact need to verify against the RP ID instead of the AppID (see issue #1034 and commit message 776b7b14d6e8f64b101db7e92318c877c588e861). In order to work around these corner cases, the RP has to always accept the RP ID as the `rpIdHash` even if the extension output alleges that the `rpIdHash` should be the hash of the AppID instead.

This means that for maximum compatibility with client implementation versions, the RP must keep this workaround behaviour even after the spec fix made in commit 776b7b14d6e8f64b101db7e92318c877c588e861. The precision of the appid extension output is therefore not very useful since it cannot be relied upon with all clients as long as at least one installation of a client with the old behaviour exists.

Therefore, this PR sacrifices the improved extension output accuracy for simplified client implementation logic.

[proprec]: https://www.w3.org/TR/2019/PR-webauthn-20190117/#sctn-appid-extension

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/pull/1144.html" title="Last updated on Jan 18, 2019, 7:38 PM UTC (20f0277)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/1144/776b7b1...20f0277.html" title="Last updated on Jan 18, 2019, 7:38 PM UTC (20f0277)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -4412,10 +4412,6 @@ JavaScript APIs.
         credential, the client MUST include the credential in
         <var ignore>allowCredentialDescriptorList</var>. The value of |appId| then replaces the `rpId`
         parameter of [=authenticatorGetAssertion=].
-    1. Let |output| be the Boolean value [FALSE].
-    1. When [creating assertionCreationData](#assertionCreationDataCreation),
-        if the [=assertion=] was created by a U2F authenticator with the U2F application parameter set to the SHA-256 hash of |appId|
-        instead of the SHA-256 hash of the [=RP ID=], set |output| to [TRUE].
 
 Note: In practice, several implementations do not implement steps four and onward of the
 algorithm for [=determining if a caller's FacetID is authorized for an AppID=].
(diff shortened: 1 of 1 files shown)


emlun:
RP implementation logic _before_ this change:

- **Extension output `appid: false`:**

  Ok, RP ID was definitely used, business as usual.

- **Extension output `appid: true`:**

  Ok, AppID is supposed have been used instead. But hmm, there's that [corner case](https://github.com/w3c/webauthn/issues/1034) in the L1 spec where the client might set `appid: true` even though the signature was computed over the RP ID instead of the AppID. Does this client implement the logic as specified in L1? **Better try the RP ID anyway if AppID fails.**

RP implementation logic _after_ this change:

- **Extension output `appid: false`:**

  Ok, RP ID was definitely used, business as usual.

- **Extension output `appid: true`:**

  Ok, spec says to **try both AppID and RP ID.**

I.e., RPs already kind-of have to do this. The motivation for the change is to simplify the client processing, and to also bring the RP processing specification in line with what RPs actually need to do to work around that corner case in L1 (which will in theory "always"(ish) be necessary since it's hard to know whether the client implements that logic as specified in L1, as specified in L2, or as was intended by L1).

In the end, though, the L1 corner case is a small one which in the worst case might hit the user with inexplicably failed authentications until they unplug all authenticators but one, so it's not a big problem, and it might be too much hassle to make this (on the surface) breaking change just to fix that.

selfissued:
I would suggest that we go with PR #1143 and fix the normative text - not changing the output.  This should be closed with no action.

emlun:
Agreed on 2019-02-06 WG call to close this.

PR #1182, opened 2019-03-12
Allow authenticators to do None instead of Self attestation

### LINKED ISSUE #978: MUST authenticators still perform self attestation?
[§6.3. Attestation][att] states (emphasis added):

>**Authenticators MUST also provide** some form of attestation. The basic requirement is that the authenticator can produce, for each credential public key, **an attestation statement verifiable by the Relying Party**. Typically, this attestation statement contains a signature by an attestation private key over the attested credential public key and a challenge, as well as a certificate or similar data providing provenance information for the attestation public key, enabling the Relying Party to make a trust decision. **However, if an attestation key pair is not available, then the authenticator MUST perform self attestation of the credential public key with the corresponding credential private key.** All this information is returned by authenticators any time a new public key credential is generated, in the overall form of an attestation object. The relationship of the attestation object with authenticator data (containing attested credential data) and the attestation statement is illustrated in figure 3, below.

Is this still accurate now that we have the [None][none] attestation type and -format? Should this not instead be something like the following (changes emphasized)?

>Authenticators **MAY** also provide some form of attestation. **If an authenticator does**, the basic requirement is that the authenticator can produce, for each credential public key, an attestation statement verifiable by the Relying Party. Typically, this attestation statement contains a signature by an attestation private key over the attested credential public key and a challenge, as well as a certificate or similar data providing provenance information for the attestation public key, enabling the Relying Party to make a trust decision. However, if an attestation key pair is not available, then the authenticator **MAY either** perform self attestation of the credential public key with the corresponding credential private key, **or otherwise perform no attestation**. All this information is returned by authenticators any time a new public key credential is generated, in the overall form of an attestation object. The relationship of the attestation object with authenticator data (containing attested credential data) and the attestation statement is illustrated in figure 3, below.
>
>**If an authenticator employs self attestation or no attestation, then no provenance information is provided for the Relying Party to base a trust decision on. In these cases, the authenticator provides no guarantees about its operation to the Relying Party.**

(The last paragraph could possibly be a Note:)

I'm not sure if this would be breaking or not. I'd say it's not important enough that it would be worth delaying the process for it, but it would be a shame if it is indeed inconsistent with the rest of the spec and we end up having to publish it.

[att]: https://w3c.github.io/webauthn/#sctn-attestation
[none]: https://w3c.github.io/webauthn/#none

emlun:
Fixes #978

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/pull/1182.html" title="Last updated on Mar 12, 2019, 12:37 PM UTC (88695f4)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/1182/3fc3b1e...88695f4.html" title="Last updated on Mar 12, 2019, 12:37 PM UTC (88695f4)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -3084,15 +3084,21 @@ or [=authenticatorGetAssertion=] operation currently in progress.
 
 ## Attestation ## {#sctn-attestation}
 
-[=Authenticators=] MUST also provide some form of [=attestation=]. The basic requirement is that the [=authenticator=] can
+[=Authenticators=] SHOULD also provide some form of [=attestation=], if possible.
+If an authenticator does, the basic requirement is that the [=authenticator=] can
 produce, for each [=credential public key=], an [=attestation statement=] verifiable by the [=[WRP]=]. Typically, this
 [=attestation statement=] contains a signature by an [=attestation private key=] over the attested [=credential public key=] and
 a challenge, as well as a certificate or similar data providing provenance information for the [=attestation public key=],
 enabling the [=[RP]=] to make a trust decision. However, if an [=attestation key pair=] is not available, then the authenticator
-MUST perform [=self attestation=] of the [=credential public key=] with the corresponding [=credential private key=]. All this
+MAY either perform [=self attestation=] of the [=credential public key=] with the corresponding [=credential private key=],
+or otherwise perform [=None|no attestation=]. All this
 information is returned by [=authenticators=] any time a new [=public key credential=] is generated, in the overall form of an
 <dfn>attestation object</dfn>. The relationship of the [=attestation object=] with [=authenticator data=] (containing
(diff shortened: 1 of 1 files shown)


emlun:
Ok to merge?

herrjemand:
Just pointing out: This still does not allow authenticators to return NONE as this needs to be synced with the common specs and FIDO2

emlun:
Yeah. This does apply to non-CTAP authenticators, but this was done mostly for internal consistency reasons since it seemed contradictory that "authenticators MUST also provide some form of attestation" while the spec also supports the None attestation format. It's probably not terribly useful to sync this into CTAP, since most of the point of hardware authenticators is that they're capable of protecting secrets such as an attestation key.

PR #1184, opened 2019-03-18
 Update isUVPAA() scenario to agree with spec 

### LINKED ISSUE #1178: Consistency between scenario "Registration specifically with UVPA" and former removal of confirmation prompt for isUVPAA()
_Warning: real newbie (and non native English writer) issue here, which I think is just a copy-edit issue. Please close this if necessary. I did not find any duplicate issue about that but correct me if needed._

While documenting the [`isUVPAA()`](https://w3c.github.io/webauthn/#dom-publickeycredential-isuserverifyingplatformauthenticatoravailable) method for [MDN](https://developer.mozilla.org/en-US/docs/Web/API/PublicKeyCredential/isUserVerifyingPlatformAuthenticatorAvailable), I stumbled across [the corresponding scenario](https://w3c.github.io/webauthn/#sample-registration-with-platform-authenticator).

In this [scenario](https://w3c.github.io/webauthn/#sample-registration-with-platform-authenticator):
* step 4 and 5 describe user involvment and consent
* in the example, the identifier `userIntent` is used
This led me to believe that there was indeed the need for user consent during the client-platform process. After reading through some articles reinforcing this idea, I found out https://github.com/w3c/webauthn/pull/904#pullrequestreview-119818144 and https://github.com/w3c/webauthn/commit/ad22fce9fbe6685490dd767bb52445e600c9af88 which removed such idea of any user consent. Therefore, the content of the scenario may not properly be consistent with the rest of the specification and especially with the description of `isUVPAA()`.

I think the corresponding fix could be:
* rewording steps 4 and 5 (maybe merge them into a single step)
* reword the `userIntent` variable in the code example
* reword the comment stating "If the user has affirmed willingness..." in the code example.

emlun:
Fixes #1178.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/pull/1184.html" title="Last updated on Mar 18, 2019, 1:56 PM UTC (bd67d13)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/1184/ce2b947...bd67d13.html" title="Last updated on Mar 18, 2019, 1:56 PM UTC (bd67d13)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -5114,35 +5114,40 @@ a [=user-verifying platform authenticator=].
 
 1. The [=[RP]=] script runs the code snippet below.
 
-1. The user agent asks the user whether they are willing to register with the [=[RP]=] using an available [=platform authenticator=].
+    1. The user agent checks if a [=user-verifying platform authenticator=] is available.
 
-1. If the user is not willing, terminate this flow.
+    1. The [=[RP]=] asks the user if they want to create a credential with it.
 
-1. The user is shown appropriate UI and guided in creating a credential using one of the available platform authenticators.
-    Upon successful credential creation, the [=[RP]=] script conveys the new credential to the server.
+    1. If no [=user-verifying platform authenticator=] is available,
+        or if the user does not want to create a credential, terminate this flow.
+
+    1. The user agent and/or operating system shows appropriate UI and guides the user in creating a credential
+        using one of the available platform authenticators.
+
+    1. Upon successful credential creation, the [=[RP]=] script conveys the new credential to the server.
 
 <pre class="example" highlight="js">
     if (!window.PublicKeyCredential) { /* Client not capable of the API. Handle error. */ }
 
     PublicKeyCredential.isUserVerifyingPlatformAuthenticatorAvailable()
-        .then(function (userIntent) {
-
-            // If the user has affirmed willingness to register with the 
-            // relying party using an available platform authenticator
-            if (userIntent) {
-                var publicKeyOptions = { /* Public key credential creation options. */};
-
-                // Create and register credentials.
-                return navigator.credentials.create({ "publicKey": publicKeyOptions });
-            } else {
-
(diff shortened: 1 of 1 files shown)


nadalin:
@equalsJeffH @jcjones  Please review

emlun:
Will merge this if at least one of @equalsJeffH or @jcjones approves.

index.bs
```diff
@@ -5114,35 +5114,40 @@ a [=user-verifying platform authenticator=].
 
 1. The [=[RP]=] script runs the code snippet below.
 
-1. The user agent asks the user whether they are willing to register with the [=[RP]=] using an available [=platform authenticator=].
+    1. The user agent checks if a [=user-verifying platform authenticator=] is available.
 
-1. If the user is not willing, terminate this flow.
+    1. The [=[RP]=] asks the user if they want to create a credential with it.
```
equalsJeffH: Hm, do we really need this check here -- the "appropriate UI ... guides user" is materialized down below in line 5124, yes?  Also, the thrust of issue #1178 is that we are not prompting the user, yes?

akshayku: Yup, we are not prompting the user.

emlun: Note that the subject here is the RP, not the client - this would happen before the RP makes the credentials.create() call. For example, by showing a nudge banner, interstitial page or similar.

akshayku: Ah.. ok.

equalsJeffH: Ok, understood, scratch this particular review.

equalsJeffH: The way this is written, does it suggest the RP shows UI asking the user about cred creation regardless of whether iUVPAA() returns true or false?  If so, do we want to do that?

emlun: Good catch, fixed in 07b58d51007d33dca4e620b2c3d1a2e6346d0775

nadalin:
@emlun Please merge

Issue #1149, opened 2019-01-29
Add notion of forbidding resident credential creation


equalsJeffH:
Update 16-May-2019: 
This issue was _improved_ by merging PR #1191, and _may_ be addressed (in practice) by the latter. Need to assess this. 

<hr>
**Original post 28-Jan-2019**:
Today, when making a credential, an RP can specify that they either require a [(client-side) resident credential](http://w3c.github.io/webauthn/#resident-credential), or that they can live with either that or the alternative server-side resident credential (think U2F). 

Thinking forward to how we’d like to migrate users with security keys to the “resident credential” world (and thus enabling password-less and typing-free flows), we propose that we add an additional value to the [`requireResidentKey`](http://w3c.github.io/webauthn/#dom-authenticatorselectioncriteria-requireresidentkey) parameter during Create. With this, we’ll end up with ResidentCredential being one of three values: `{Required, Preferred, Forbidden}`:

`Required` maps to the current setting of [`requireResidentKey = true`](http://w3c.github.io/webauthn/#dom-authenticatorselectioncriteria-requireresidentkey), explicitly requiring the authnr to create a [resident credential](http://w3c.github.io/webauthn/#resident-credential).

`Preferred`, indicates that the RP prefers to create a [resident credential](http://w3c.github.io/webauthn/#resident-credential), but would also be fine with a non-resident credential (aka "[server-side cred](http://w3c.github.io/webauthn/#server-side-credential-storage-modality)"), if that’s the only thing the authenticator supports. Presently, [`requireResidentKey = false`](http://w3c.github.io/webauthn/#dom-authenticatorselectioncriteria-requireresidentkey) largely maps to this.

`Forbidden` will map to a new, explicit sense of “false”&mdash;it’s better defined&mdash;it means that the authenticator must explicitly create a [server-side resident credential](http://w3c.github.io/webauthn/#server-side-credential-storage-modality). 

See also:  [Credential Storage Modality](http://w3c.github.io/webauthn/#sctn-credential-storage-modality) and [Authenticator Taxonomy](http://w3c.github.io/webauthn/#sctn-authenticator-taxonomy)

dwaite:
For clarification, why would a relying party require this Forbidden behavior?

emlun:
[...]

PR #1191, opened 2019-03-25
Initial proposal for modifications to resident key credential requirements

### LINKED ISSUE #991: Indicate resident key credential "preferred" during registration and find out what the authenticator offered
I'd like to revisit the scenario I was trying to achieve in the (now closed) issue #987 

Why can't the "requireResidentKey" authenticator selection criteria be a value like "preferred", "required" or "never" instead of a boolean true/false? In addition the registration response could indicate residentKey true/false as a flag similar to the userPresent and userVerified flags, or in an extension.

That way, without knowing ahead of time if an authenticator is capable of resident key (even a portable security key), the relying party could request it with fallback to derived credential (without a hard error), then notify the user of what actually happened and therefore what scenarios their key might be used for (replacement for regular username/password authentication, or only 2nd-factor scenarios)?

sbweeden:
Attempts to fix  #991  #1149 

Outstanding issues/questions:

In the user-agent processing algorithm described in section 5.1.3, please check if the modifications in step 20 are possible w.r.t. CTAP. Specifically is it possible for a user-agent (client) to determine in an authenticator is capable of resident (or non-resident) key credentials?

In the IANA considerations section (11.2) I have added the credProps extension. How do I really register this with WebAuthn-Registries?

***
<a href="https://pr-preview.s3.amazonaws.com/sbweeden/webauthn/pull/1191.html" title="Last updated on May 7, 2019, 12:57 AM UTC (466b563)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/1191/5d11e62...sbweeden:466b563.html" title="Last updated on May 7, 2019, 12:57 AM UTC (466b563)">Diff</a>


index.bs
```diff
@@ -2845,7 +2934,7 @@ It takes the following input parameters:
 : |userEntity|
 :: The user account's {{PublicKeyCredentialUserEntity}}, containing the [=user handle=] given by the [=[RP]=].
 : |requireResidentKey|
-:: The {{PublicKeyCredentialCreationOptions/authenticatorSelection}}.{{requireResidentKey}} value given by the [=[RP]=].
+:: The [=effective residenty key requirement for credential creation=], a Boolean value provided by the client.
```
emlun: s/residenty/resident/

index.bs
```diff
@@ -4902,6 +4991,55 @@ as candidates to be employed in a [=registration ceremony=].
 : Authenticator extension output
 :: None.
 
+
+## Credential Properties Extension (credProps) ## {#sctn-authenticator-credential-properties-extension}
+
+This extension allows [=[WRPS]=] to query additional properties of the authenticator's credential created during a
```
emlun: I would suggest: s/the authenticator's credential/the [=public key credential source=]/

index.bs
```diff
@@ -4902,6 +4991,55 @@ as candidates to be employed in a [=registration ceremony=].
...
+    };
+    </xmp>
+
+: Client extension processing
+:: None, except creating the authenticator extension input from the client extension input.
+
+: Client extension output
+:: Returns the dictionary of discovered credential properties as reported by either the authenticator itself or by the 
+    client if known, such as when a platform authenticator is used. This dictionary is initially defined with one credential
+    property rk to indicate if the authenticator has created or used a resident key credential. Other credential properties
```
emlun: `rk` -> `{{AuthenticationExtensionsClientInputs/rk}}`

index.bs
```diff
@@ -4902,6 +4991,55 @@ as candidates to be employed in a [=registration ceremony=].
...
+    partial dictionary AuthenticationExtensionsClientInputs {
+        boolean credProps;
+    };
+    </xmp>
+
+: Client extension processing
+:: None, except creating the authenticator extension input from the client extension input.
+
+: Client extension output
+:: Returns the dictionary of discovered credential properties as reported by either the authenticator itself or by the
```
emlun: I don't think it's necessary to have the client forward the authenticator extension outputs, since those are already returned in the authenticator data. Maybe something like this?

>Returns the credential properties as discovered by the client, if any, such as when a platform authenticator is used. The client MAY omit any properties at its discretion, and SHOULD omit any properties that would be duplicated in the authenticator extension output. This dictionary is initially defined [...]

index.bs
```diff
@@ -1129,13 +1130,56 @@ a numbered step. If outdented, it (today) is rendered as a bullet in the midst o
...
+                    [=iteration/continue=].
+                    ::  Note:  The value of <code>|options|.{{PublicKeyCredentialCreationOptions/authenticatorSelection}}.{{requireResidentKey}}</code> is to be ignored
+                        if a value for <code>|options|.{{PublicKeyCredentialCreationOptions/authenticatorSelection}}.{{residentKey}}</code> is supplied.
+                  1. If <code>|options|.{{PublicKeyCredentialCreationOptions/authenticatorSelection}}.{{residentKey}}</code> is set to
+                    {{ResidentKeyRequirement/forbidden}} and the |authenticator| is ONLY capable of storing a [=client-side-resident public key credential source=],
+                    [=iteration/continue=].
+                  1. If <code>|options|.{{PublicKeyCredentialCreationOptions/authenticatorSelection}}.{{residentKey}}</code> is set to
+                    {{ResidentKeyRequirement/preferred}} and the |authenticator| is not capable of storing a [=client-side-resident public key credential source=],
+                    and there are other known authenticators attached that are capable of storing a [=client-side-resident public key credential source=],
```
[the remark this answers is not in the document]
sbweeden: Anything for me to do here - or just leave it with you for now? If for me, I'll need a little more guidance.

index.bs
```diff
@@ -2094,6 +2146,43 @@ operation has no [=authenticator attachment modality=] selection option, so the
...
+        "preferred",
+        "required"
+    };
+</xmp>
+
+This enumeration's values describe the [=[RP]=]'s requirements for [=resident credentials=]:
+
+<div dfn-type="enum-value" dfn-for="ResidentKeyRequirement">
+    :   <dfn>forbidden</dfn>
+    ::  This value indicates the authenticator MUST only create a non-resident credential.
```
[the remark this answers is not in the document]
sbweeden: I'm not sure if you're asking me to take a stab at that or not? I'm happy to make an update to reference such a definition because I agree it's better than saying _MUST only create a non-resident credential_.

index.bs
```diff
@@ -2094,6 +2146,43 @@ operation has no [=authenticator attachment modality=] selection option, so the
...
+
+    :   <dfn>indifferent</dfn>
+    ::  This value indicates the client may use any authenticator. The authenticator MAY create a resident or non-resident credential at
```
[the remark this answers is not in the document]
sbweeden: fixed in subsequent commit

index.bs
```diff
@@ -1129,13 +1130,56 @@ a numbered step. If outdented, it (today) is rendered as a bullet in the midst o
...
+                    [=iteration/continue=].
+                  1. If <code>|options|.{{PublicKeyCredentialCreationOptions/authenticatorSelection}}.{{residentKey}}</code> is set to
+                    {{ResidentKeyRequirement/preferred}} and the |authenticator| is not capable of storing a [=client-side-resident public key credential source=],
+                    and there are other known authenticators attached that are capable of storing a [=client-side-resident public key credential source=],
                     [=iteration/continue=].
                   1. If <code>|options|.{{PublicKeyCredentialCreationOptions/authenticatorSelection}}.{{AuthenticatorSelectionCriteria/userVerification}}</code> is
                     set to {{UserVerificationRequirement/required}} and the |authenticator| is not capable of performing [=user
                     verification=], [=iteration/continue=].
 
+            1. Let |requireResidentKey| be the <dfn>effective residenty key requirement for credential creation</dfn>, a Boolean value,
```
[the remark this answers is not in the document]
sbweeden: fixed in subsequent commit

index.bs
```diff
@@ -2094,6 +2146,43 @@ operation has no [=authenticator attachment modality=] selection option, so the
...
+    ::  This value indicates the authenticator MUST only create a non-resident credential.
+
+    :   <dfn>indifferent</dfn>
+    ::  This value indicates the client may use any authenticator. The authenticator MAY create a resident or non-resident credential at 
+        its discretion.
+
+    :   <dfn>preferred</dfn>
+    ::  This value indicates the client should try to use an authenticator capable of [=resident credentials=] in preference to one that
+        is incapable of [=resident credentials=]. If no authenticator is capable of [=resident credentials=], the client may use any
+        authenticator capable of non-resident credentials.
```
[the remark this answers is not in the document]
sbweeden: Agreed - fixed in subsequent commit.

index.bs
sbweeden: Fixed in subsequent commit.

index.bs
sbweeden: Agreed - fixed in subsequent commit.

index.bs
sbweeden: Agreed - fixed in subsequent commit.

index.bs
sbweeden: Fixed in subsequent commit by it's actually: `rk -> {{CredentialPropertiesOutput/rk}}`

index.bs
```diff
@@ -2094,6 +2146,43 @@ operation has no [=authenticator attachment modality=] selection option, so the
...
+
+    :   <dfn>required</dfn>
+    ::  This value indicates the client MUST use an authenticator capable of [=resident credentials=]. The authenticator MUST create
+        [=resident credentials=] and MUST return an error if it cannot.
```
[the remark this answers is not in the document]
sbweeden: fixed in subsequent commit

index.bs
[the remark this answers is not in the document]
sbweeden: I've taken a stab at it. I've made the shorthand alias _Non-Resident Credential_ and understand this may be debatable, however after consideration there is a LOT of existing references to the term _resident credential_, and I believe that the _Non-Resident Credential_ shorthand is the most logically definition of one that is **not client-side resident**.

index.bs
```diff
@@ -2161,25 +2169,25 @@ This enumeration's values describe the [=[RP]=]'s requirements for [=resident cr
...
-    ::  This value indicates the client may use any authenticator. The authenticator MAY create a resident or non-resident credential at 
-        its discretion.
+    ::  This value indicates the client MAY use any authenticator. The authenticator MAY create a [=Resident Credential=] or 
+        [=Non-Resident Credential=] at its discretion.
 
     :   <dfn>preferred</dfn>
-    ::  This value indicates the client should try to use an authenticator capable of [=resident credentials=] in preference to one that
-        is incapable of [=resident credentials=]. If no authenticator is capable of [=resident credentials=], the client may use any
-        authenticator capable of non-resident credentials.
+    ::  This value indicates the client SHOULD try to use an authenticator capable of a [=resident credentials|Resident Credential=]
```
emlun: You don't need to specify `[=resident credentials|Resident Credential=]`. Bikeshed automatically equates simple -s plural forms and capitalizations, so this can just be `[=resident credential=]`.

index.bs
```diff
@@ -2161,25 +2169,25 @@ This enumeration's values describe the [=[RP]=]'s requirements for [=resident cr
 
 <div dfn-type="enum-value" dfn-for="ResidentKeyRequirement">
     :   <dfn>forbidden</dfn>
-    ::  This value indicates the authenticator MUST only create a non-resident credential.
+    ::  This value indicates the authenticator MUST create a [=Non-Resident Credential=].
```
emlun: We don't usually capitalize terms like this outside their definition, so I'd prefer `[=non-resident credential=]` and `[=resident credential=]` throughout.

index.bs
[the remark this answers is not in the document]
emlun: I'm working on it.

index.bs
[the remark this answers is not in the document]
emlun: Ok, that makes sense.

index.bs
```diff
@@ -601,6 +602,14 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
     By definition, the [=credential private key=] is always exclusively controlled by the [=authenticator=]. In the case of a
     [=resident credential=], the [=authenticator=] might offload storage of wrapped key material to the
     [=client device=], but the [=client device=] is not expected to offload the key storage to remote entities (e.g., [=[WRP]=] Server).
+    An authenticator is said to be using [=client-side credential storage modality=] when generating and storing a [=Resident Credential=].
+
+: <dfn>Non-Resident Credential</dfn>
+: <dfn>Server-side-resident Public Key Credential Source</dfn>
```
emlun: The terminology section is sorted, so I think we should move this to between "RP ID" and "Test of User Presence". It makes a lot of sense to have this directly following the Client-side-resident Public Key Credential Source definition, but I don't think it's worth breaking the ordering.

index.bs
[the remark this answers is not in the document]
emlun: Meta-PR open: https://github.com/sbweeden/webauthn/pull/2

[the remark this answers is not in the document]
sbweeden: Thanks. Accepted and incorporated.

index.bs
sbweeden: Thanks - done.

index.bs
sbweeden: Thanks - done.

index.bs
sbweeden: Ok - I did noticed the ordering was alphabetical, and had chosen to put them adjacent. No problem with moving it though. Done.

equalsJeffH:
thanks for submitting this @sbweeden!  I'm at IETF this week, wont be able to submit a review until next week...

index.bs
emlun: Oh yeah, good catch!

akshayku:
This requires modifications according to latest discussion in the issue. See https://github.com/w3c/webauthn/issues/991#issuecomment-477728368

sbweeden:
@akshayku, @emlun, @agl  I have changed the ResidentKeyRequirement enumeration back to my original proposal based on the comments from the 2019-04-03 WG call. Hoping this can be reviewed now.

emlun:
@akshayku talked earlier about having the client retry the authenticator invocation with different `requireResidentKey` arguments in the "preferred" and "discouraged" cases. Would it be useful to capture that retry logic here, or would that be too prescriptive?

index.bs
```diff
@@ -1129,13 +1138,68 @@ a numbered step. If outdented, it (today) is rendered as a bullet in the midst o
...
+
+                  1. If <code>|options|.{{PublicKeyCredentialCreationOptions/authenticatorSelection}}.{{residentKey}}</code>
+
+                      <dl class="switch">
+                          :   is set to {{ResidentKeyRequirement/required}}
+                          ::  If the |authenticator| is not capable of storing a [=client-side-resident public key credential
+                              source=], [=iteration/continue=].
+
+                          :   is set to {{ResidentKeyRequirement/preferred}}
+                          ::  If the |authenticator| is not capable of storing a [=client-side-resident public key credential
```
agl: I don't think this is the behaviour that we would actually implement. Rather, `preferred` would mean that a resident credential would be created if possible, but if the user selects an authenticator incapable of that, we're not going to tell them to use the other one but rather will create a non-resident credential.

akshayku:
@emlun That seems too prescriptive to me. Platforms/Browsers handling of these parameters to CTAP parameters is complex and I would not go over prescribing it. 

I would concentrate on defining the intent from the RP's perspective correctly and its use cases.

akshayku:
I have no objection to adding `indifferent`, if people see its value.

index.bs
```diff
@@ -1129,13 +1138,68 @@ a numbered step. If outdented, it (today) is rendered as a bullet in the midst o
...
+                              source=], and there are other known authenticators attached that are capable of storing a
+                              [=client-side-resident public key credential source=], [=iteration/continue=].
+
+                          :   is set to {{ResidentKeyRequirement/discouraged}}
+                          ::  If the |authenticator| is not capable of creating a [=server-side-resident public key credential
```
agl: This wording is based around how the private key is stored, rather than whether the credential may be returned with an empty allowList. As a platform, we don't actually know how the private key is stored. For cost reasons, current authenticators tend to wrap it in the credential ID, but they might not.

Rather, `discouraged` to me is an opinion about that friction involved in registration. I see this as the RP saying &ldquo;please don't bother the user with a PIN prompt or hassle them about a fingerprint being rejected&rdquo;.

index.bs
```diff
@@ -1129,13 +1138,68 @@ a numbered step. If outdented, it (today) is rendered as a bullet in the midst o
...
+                              source=], and there are other known authenticators attached that are capable of creating a
+                              [=server-side-resident public key credential source=], [=iteration/continue=].
+
+                          :   is [=not present=]
```
agl: Since `residentKey` is defined with a default value, I don't think &ldquo;not present&rdquo; is a state that it can have.

index.bs
```diff
@@ -1129,13 +1138,68 @@ a numbered step. If outdented, it (today) is rendered as a bullet in the midst o
...
+                                ::  Let |requireResidentKey| be [TRUE].
+
+                                :   is not capable of [=client-side-resident public key credential source=], or if the client cannot determine authenticator capability,
+                                ::  Let |requireResidentKey| be [FALSE].
+                            </dl>
+
+                        :   is set to {{ResidentKeyRequirement/required}}
+                        ::  Let |requireResidentKey| be [TRUE].
+
+                        :   is [=not present=]
```
agl: (Ditto about a default value excluding this case.)

index.bs
```diff
@@ -2043,9 +2108,16 @@ attributes.
         specified [[#attachment]].
 
     :   <dfn>requireResidentKey</dfn>
-    ::  This member describes the [=[RP]=]'s requirements regarding [=resident credentials=].
-        If the parameter is set to [TRUE], the authenticator MUST create a
-        [=client-side-resident public key credential source=] when creating a [=public key credential=].
+    ::  This member is retained for backwards compatibility but is deprecated in favour of providing *residentKey*.
+        It will be ignored if *residentKey* is supplied and understood by the user-agent. When provided, it describes the
```
agl: (I think this wording is also incompatible with setting a default value for `residentKey`.)

index.bs
```diff
@@ -1129,13 +1138,68 @@ a numbered step. If outdented, it (today) is rendered as a bullet in the midst o
...
                     verification=], [=iteration/continue=].
 
+            1. Let |requireResidentKey| be the <dfn>effective resident key requirement for credential creation</dfn>, a Boolean value,
+                as follows. 
+
+                If <code>|options|.{{PublicKeyCredentialCreationOptions/authenticatorSelection}}.{{residentKey}}</code>
+
+                    <dl class="switch">
+
+                        :   is set to {{ResidentKeyRequirement/discouraged}}
```
agl: (If you removed the default value for `residentKey` then I think these cases should be &ldquo;is present and set to &hellip;&rdquo;)

index.bs
```diff
@@ -2094,6 +2166,40 @@ operation has no [=authenticator attachment modality=] selection option, so the
...
+        "preferred",
+        "required"
+    };
+</xmp>
+
+This enumeration's values describe the [=[RP]=]'s requirements for [=resident credentials=]:
+
+<div dfn-type="enum-value" dfn-for="ResidentKeyRequirement">
+    :   <dfn>discouraged</dfn>
+    ::  This value indicates the client SHOULD try to use an authenticator capable of a [=non-resident credential=]
```
agl: (This wording is also based around storage of private keys, which the platform doesn't generally have insight into, vs whether a credential can be returned with an empty allowList. It's also based around authenticator selection but I don't think we would implement that way vs using this information to make a decision about the rk flag after the user has selected an authenticator.)

index.bs
```diff
@@ -4902,6 +5009,56 @@ as candidates to be employed in a [=registration ceremony=].
...
+
+: Client extension processing
+:: None, except creating the authenticator extension input from the client extension input.
+
+: Client extension output
+:: Returns the credential properties as discovered by the client, if any, such as when a platform authenticator is used. 
+    The client MAY omit any properties at its discretion, and SHOULD omit any properties that would be duplicated in the
+    authenticator extension output. This dictionary is initially defined with one credential property 
+    {{CredentialPropertiesOutput/rk}} to indicate if the authenticator has created or used a resident key 
+    credential. Other credential properties MAY be returned in this extension.
```
agl: &ldquo;Other credential properties MAY be returned in this extension&rdquo;, but they aren't defined. This appears to be giving implementations permission to invent new properties at will. I don't necessarily disagree with doing that but, if that's the intention, I believe it should be called out explicitly rather than hinted at.

index.bs
```diff
@@ -4902,6 +5009,56 @@ as candidates to be employed in a [=registration ceremony=].
...
+    partial dictionary AuthenticationExtensionsClientInputs {
+        boolean credProps;
+    };
+    </xmp>
+
+: Client extension processing
+:: None, except creating the authenticator extension input from the client extension input.
+
+: Client extension output
+:: Returns the credential properties as discovered by the client, if any, such as when a platform authenticator is used.
```
agl: Did you mean to include a boolean for whether a platform authenticator was used? (&ldquo;such as when a platform authenticator is used&rdquo;)

index.bs
akshayku: +1. That's what I inferred.

index.bs
akshayku: Whether one wants PIN or not, should be controlled by userVerification. Although, currently, that is not mapping to intended behavior as Christiaan wants. Meaning, even if RP said that it does not want UV (UVdiscouraged), we end up having UV because of how CTAP is designed at this moment. 

Combining this definition with whether PIN prompt will be shown or not is not correct. May be in future, we may allow creating resident keys also without PIN prompt as Christiaan wanted. 

This parameter is about whether to create a resident key or not and we should only talk about its behavior in terms or that. 

From an Authenticator's perspective, they are given a freedom right now to enhance protection if they want. Meaning, in CTAP spec, even if request is not asking for UV, authenticator may do UV and that security decision is left to it and will be reflected in the flags.

index.bs
akshayku: We have to think about the scenario where RPs have coded older RequiredResidentKeys behavior in their pages and when client's update, they should not break those RPs.

agl: Oh certainly, but just as a narrow technical matter, I think providing a default value means that it always has _a_ value w.r.t. Javascript.

sbweeden: No action taken - I believe @akshayku has correctly expressed why this may not have a value.

index.bs
sbweeden: I have changed the algorithm to say that if the value is preferred or discouraged, then "no effect" - i.e. don't ignore authenticators and continue in this case, but instead let the value drive how we instruct the authenticator to work in the next part of the algorithm.

index.bs
sbweeden: No action taken on this one based on comments from @akshayku

index.bs
sbweeden: Understood - removed default value for requireResidentKey

index.bs
sbweeden: Understood - removed default value.

index.bs
sbweeden: Agreed - changed as suggested

index.bs
sbweeden: I have changed the description for both discouraged and preferred to better convey the intent of client/authenticator behaviour. Will let the next round of review decide if that's ok.

index.bs
sbweeden: Modified wording to indicate that it's at the discretion of the client if they wish to add additional properties.

index.bs
sbweeden: No intention to include a boolean for whether a platform authenticator was used, although this is a perfectly reasonable suggestion and a possible additional property that could be included.  Let me know in next round of review (or via another issue) if you think that should be included as a credential property populated by the client.

The real reason I mentioned a platform authenticator is that generally speaking it's the client that implements or is strongly bound to the platform authenticator code. In that case the client (which is a superset of the authenticator) will have complete knowledge of what it's behaviour is and be the authoritative source for credential properties.

sbweeden:
Ready for re-review.

emlun:
Hm... looks like the pr-preview bot has failed to update the [Diff][old] link in the OP for quite a while - it's still pointing at commit 29a6701. Here's the most recent diff from commit e90506b: [Diff][new]

[old]: https://pr-preview.s3.amazonaws.com/w3c/webauthn/1191/8678b43...sbweeden:29a6701.html
[new]: https://pr-preview.s3.amazonaws.com/w3c/webauthn/1191/8678b43...sbweeden:e90506b.html

index.bs
```diff
@@ -2043,9 +2101,16 @@ attributes.
...
-    ::  This member describes the [=[RP]=]'s requirements regarding [=resident credentials=].
-        If the parameter is set to [TRUE], the authenticator MUST create a
-        [=client-side-resident public key credential source=] when creating a [=public key credential=].
+    ::  This member is retained for backwards compatibility but is deprecated in favour of providing *residentKey*.
+        It will be ignored if *residentKey* is supplied and understood by the user-agent. When provided, it describes the 
+        [=[RP]=]'s requirements regarding [=resident credentials=]. If the parameter is set to [TRUE], the authenticator MUST create 
+        a [=client-side-resident public key credential source=] when creating a [=public key credential=].
+
+    :   <dfn>residentKey</dfn>
+    ::  This member supercedes the use of *requireResidentKey* to describes the [=[RP]=]'s requirements regarding
```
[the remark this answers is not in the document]
emlun: I might be wrong here, but: s/supercedes/supersedes/ ?

index.bs
```diff
@@ -2043,9 +2101,16 @@ attributes.
         specified [[#attachment]].
 
     :   <dfn>requireResidentKey</dfn>
-    ::  This member describes the [=[RP]=]'s requirements regarding [=resident credentials=].
-        If the parameter is set to [TRUE], the authenticator MUST create a
-        [=client-side-resident public key credential source=] when creating a [=public key credential=].
+    ::  This member is retained for backwards compatibility but is deprecated in favour of providing *residentKey*.
+        It will be ignored if *residentKey* is supplied and understood by the user-agent. When provided, it describes the 
+        [=[RP]=]'s requirements regarding [=resident credentials=]. If the parameter is set to [TRUE], the authenticator MUST create 
+        a [=client-side-resident public key credential source=] when creating a [=public key credential=].
```
emlun: Throughout this paragraph and the next:
- `*requireResidentKey*` => `{{requireResidentKey}}`
- `*residentKey*` => `{{residentKey}}`

index.bs
```diff
@@ -2043,9 +2101,16 @@ attributes.
...
+
+    :   <dfn>residentKey</dfn>
+    ::  This member supercedes the use of *requireResidentKey* to describes the [=[RP]=]'s requirements regarding 
+        [=resident credentials=]. Only one of *residentKey* or *requireResidentKey* should be present in 
+        {{AuthenticatorSelectionCriteria}}, and use of *residentKey* is encouraged. If both are present 
+        and the user-agent supports it, *residentKey* takes precedence over *requireResidentKey*.
```
emlun: s/user-agent/[=client=]/

index.bs
```diff
@@ -2094,6 +2159,40 @@ operation has no [=authenticator attachment modality=] selection option, so the
...
+    ::  This value indicates the client, when communicating with an authenticator, SHOULD instruct the authenticator
+        to create a [=resident credential=]. If the authenticator is incapable of [=resident credentials=], 
+        the client MAY use the authenticator to create a [=non-resident credential=].
+
+    :   <dfn>required</dfn>
+    ::  This value indicates the client MUST use an authenticator capable of a [=resident credential=]. 
+        The authenticator MUST create a [=resident credential=] and MUST return an error if it cannot.
+</div>
+
+Note: It is intended for the client to use the [[#sctn-authenticator-credential-properties-extension]] in conjunction with the
```
emlun: s/intended for the client/intended for the [=[RP]=]/

index.bs
```diff
@@ -1129,13 +1138,61 @@ a numbered step. If outdented, it (today) is rendered as a bullet in the midst o
...
+                            <dl class="switch">
+                                :   is capable of [=client-side-resident public key credential source=]
+                                ::  Let |requireResidentKey| be [TRUE].
+
+                                :   is not capable of [=client-side-resident public key credential source=], or if the client cannot determine authenticator capability,
+                                ::  Let |requireResidentKey| be [FALSE].
+                            </dl>
+
+                        :   is present and set to {{ResidentKeyRequirement/required}}
+                        ::  Let |requireResidentKey| be [TRUE].
```
emlun: If you'll indulge me, would you arrange the cases in the same order as above (required, preferred, discouraged, not present)? :slightly_smiling_face:

emlun:
Should we also have a CDDL definition for the authenticator extension output? I'm afraid I'm not proficient with the syntax, so I can't propose one at this time.

index.bs
```diff
@@ -1129,13 +1138,61 @@ a numbered step. If outdented, it (today) is rendered as a bullet in the midst o
...
+
+                        :   is [=not present=]
+                        ::  If <code>|options|.{{PublicKeyCredentialCreationOptions/authenticatorSelection}}.{{requireResidentKey}}</code>
+                                is [=present=], let |requireResidentKey| be the value of
```
agl: nit: `requireResidentKey` has a default value and thus I believe is always considered to be present. So here you could just take the value of `requireResidentKey`.

index.bs
```diff
@@ -2043,9 +2101,16 @@ attributes.
...
-        If the parameter is set to [TRUE], the authenticator MUST create a
-        [=client-side-resident public key credential source=] when creating a [=public key credential=].
+    ::  This member is retained for backwards compatibility but is deprecated in favour of providing *residentKey*.
+        It will be ignored if *residentKey* is supplied and understood by the user-agent. When provided, it describes the 
+        [=[RP]=]'s requirements regarding [=resident credentials=]. If the parameter is set to [TRUE], the authenticator MUST create 
+        a [=client-side-resident public key credential source=] when creating a [=public key credential=].
+
+    :   <dfn>residentKey</dfn>
+    ::  This member supercedes the use of *requireResidentKey* to describes the [=[RP]=]'s requirements regarding 
+        [=resident credentials=]. Only one of *residentKey* or *requireResidentKey* should be present in
```
agl: Should we say that only one should be present? Providing both means being backwards compatible with level 1 of the spec so I might almost go the other way: &ldquo;if {{residentKey}} is set to `required` then {{requireResidentKey}} should be `true`, else it should be omitted&rdquo;. Don't feel strongly either way however.

index.bs
agl: I think &ldquo;such as&rdquo; may be imprecise here. What do you think about:

> Returns the credential properties as discovered by the client, if any. For example, a client is especially likely to have knowledge about properties of credentials stored in platform authenticators.

But your call, either way&mdash;I now understand how you intended it.

index.bs
```diff
@@ -4902,6 +5002,56 @@ as candidates to be employed in a [=registration ceremony=].
...
+## Credential Properties Extension (credProps) ## {#sctn-authenticator-credential-properties-extension}
+
+This extension allows [=[WRPS]=] to query additional properties of the [=public key credential source=] created during a 
+[=registration ceremony=] or used during an [=authentication ceremony=].
+
+: Extension identifier
+:: `credProps`
+
+: Operation applicability
+:: [=registration extension|Registration=] and [=authentication extension|Authentication=]
```
agl: I hadn't noticed that this was also defined at authentication time. What use did you have in mind for that?

equalsJeffH:
for some unknown-to-me reason, pr-preview's diff link (see the bottom of the original post for the "diff" link) seems to be stuck on a "latest commit" value from 26-Mar (29a6701) rather than being automagically updated with each commit. The latest diff link ought to be AFAICT:  e90506b.  note those commits are in Shane's fork repo: https://github.com/sbweeden/webauthn/commits/sbweeden-rk-991

equalsJeffH:
would be good to update shane's fork's master branch and merge-from-master into the sbweeden-rk-991 branch.

index.bs
sbweeden: residentKey no longer has a default value defined, so it is now possible that it is not present.

index.bs
sbweeden: Happy with this re-wording, applying.

index.bs
```diff
@@ -1146,15 +1146,8 @@ a numbered step. If outdented, it (today) is rendered as a bullet in the midst o
...
-                          :   is set to {{ResidentKeyRequirement/preferred}}
-                          ::  If the |authenticator| is not capable of storing a [=client-side-resident public key credential
-                              source=], and there are other known authenticators attached that are capable of storing a
-                              [=client-side-resident public key credential source=], [=iteration/continue=].
-
-                          :   is set to {{ResidentKeyRequirement/discouraged}}
-                          ::  If the |authenticator| is not capable of creating a [=server-side-resident public key credential
-                              source=], and there are other known authenticators attached that are capable of creating a
-                              [=server-side-resident public key credential source=], [=iteration/continue=].
+                          :   is set to {{ResidentKeyRequirement/preferred}} or {{ResidentKeyRequirement/discouraged}}
```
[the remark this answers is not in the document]
sbweeden: Agreed - updated.

index.bs
```diff
@@ -2180,14 +2173,14 @@ This enumeration's values describe the [=[RP]=]'s requirements for [=resident cr
 
 <div dfn-type="enum-value" dfn-for="ResidentKeyRequirement">
     :   <dfn>discouraged</dfn>
-    ::  This value indicates the client SHOULD try to use an authenticator capable of a [=non-resident credential=] 
-        in preference to one that only supports a [=resident credential=]. If no authenticator is capable of a 
-        [=non-resident credential=], the client MAY use any authenticator.
+    ::  This value indicates the client, when communicating with an authenticator, SHOULD NOT instruct the authenticator
+        to create a [=resident credential=]. The authenticator SHOULD create a [=non-resident credential=], however either
+        a [=resident credential=] or [=non-resident credential=] will be accepted by the client.
```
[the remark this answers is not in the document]
sbweeden: Agreed - fixed.

index.bs
```diff
@@ -2180,14 +2173,14 @@ This enumeration's values describe the [=[RP]=]'s requirements for [=resident cr
...
     :   <dfn>preferred</dfn>
-    ::  This value indicates the client SHOULD try to use an authenticator capable of a [=resident credential=] 
-        in preference to one that only supports a [=non-resident credential=]. If no authenticator is capable of a 
-        [=resident credential=], the client MAY use any authenticator.
+    ::  This value indicates the client, when communicating with an authenticator, SHOULD instruct the authenticator
+        to create a [=resident credential=]. If the authenticator is incapable of [=resident credentials=], 
+        the client MAY use the authenticator to create a [=non-resident credential=].
 
     :   <dfn>required</dfn>
     ::  This value indicates the client MUST use an authenticator capable of a [=resident credential=].
```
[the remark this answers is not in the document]
sbweeden: Agreed - fixed.

index.bs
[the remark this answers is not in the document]
sbweeden: No problem with changing that. Done.

index.bs
sbweeden: Agreed - done

index.bs
sbweeden: Changed both at this location, and in the preceding paragraph.

index.bs
sbweeden: Done.

index.bs
sbweeden: Done

index.bs
sbweeden: Agreed - simplified based on this assertion.

index.bs
sbweeden: I'm going to leave this for now. I feel it introduces ambiguity as to what `requireResidentKey` should be if supplied alongside `residentKey`=`preferred`. I feel it's better just to instruct the client to apply `residentKey` only (and ignore `requireResidentKey`) if it understand the L2 definition.

index.bs
sbweeden: The initial idea was that it could allow possible discovery of the type of credential for existing registrations created under L1 if the client is able to determine that. I guess this might be possible for platform authenticators, but probably not for portable authenticators. If other credential properties (i.e. new properties beyond `rk`) are supported by a client, then it might have other uses.

If you see no value in it, I've no issues with dropping the authentication usage.

sbweeden:
@emlun - I believe I have addressed all your review comments and requested changes.
@equalsJeffH - I have merged master into the sbweeden-rk-991 branch.
@agl - I've addressed your comments as well, let me know your thoughts on the one open discussion on the `credProps` extension in an authentication flow.

index.bs
agl: I would drop it for now because a) I don't know what I would do with it and b) it can always be added later if we add a property where it makes clear sense.

index.bs
agl: If, as an RP, I want to create a resident key, I could set `residentKey` to `required`, but that doesn't work for any browser only implementing Webauthn level one. So I could set `requireResidentKey` too to cover both, but this says that only one of them should be present, so I remove `residentKey`.

I think that's ok, but it does mean that we won't see the transition that we might otherwise.

index.bs
```diff
@@ -4900,6 +4995,57 @@ as candidates to be employed in a [=registration ceremony=].
...
+: Client extension processing
+:: None, except creating the authenticator extension input from the client extension input.
+
+: Client extension output
+:: Returns the credential properties as discovered by the client, if any. For example, a client is especially likely
+    to have knowledge about properties of credentials stored in [=platform authenticators=]. 
+    The client MAY omit any properties at its discretion, and SHOULD omit any properties that would be duplicated in the
+    authenticator extension output. This dictionary is initially defined with one credential property 
+    {{CredentialPropertiesOutput/rk}} to indicate if the authenticator has created or used a resident key 
+    credential. Other credential properties MAY be returned in this extension if client implementations wish to do so.
```
agl: > Other credential properties MAY be returned in this extension if client implementations wish to do so.

This seems to give clients carte blanche to add stuff to `CredentialPropertiesOutput`, which strikes me as odd for a W3C spec? Dunno.

index.bs
sbweeden: Ok - I will change the wording to be more accepting of both being present.

index.bs
sbweeden: Fair enough - removed.

index.bs
sbweeden: Me either - I'll remove it for now until there is a compelling reason to have more properties.

sbweeden:
@agl - Addressed those points. Let me know if you see anything else outstanding.

index.bs
sbweeden: Dropped use of this extension for assertion flow

emlun: For what it's worth, leaving it in would mean authenticators can start implementing it in authenticatorGetAssertion. It doesn't have to be left to the client to figure out what to do with it.

index.bs
```diff
@@ -4900,6 +4995,56 @@ as candidates to be employed in a [=registration ceremony=].
...
+
+: Client extension processing
+:: None, except creating the authenticator extension input from the client extension input.
+
+: Client extension output
+:: Returns the credential properties as discovered by the client, if any. For example, a client is especially likely
+    to have knowledge about properties of credentials stored in [=platform authenticators=]. 
+    The client MAY omit any properties at its discretion, and SHOULD omit any properties that would be duplicated in the
+    authenticator extension output. This dictionary is defined with one credential property 
+    {{CredentialPropertiesOutput/rk}} to indicate if the authenticator has created or used a resident key credential.
```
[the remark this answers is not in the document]
sbweeden: Accepted and delivered.

index.bs
```diff
@@ -4900,6 +4995,57 @@ as candidates to be employed in a [=registration ceremony=].
...
+
+    <xmp class="idl">
+    dictionary CredentialPropertiesOutput {
+        boolean rk;
+    };
+
+    partial dictionary AuthenticationExtensionsClientOutputs {
+        CredentialPropertiesOutput credProps;
+    };
+    </xmp>
```
equalsJeffH: I'm thinking we only really need to say this much about this extension given it's presently narrow purpose and single return value:

: Client extension output
:: This [=registration extension=] returns the newly-created [=credential=]'s properties to the calling [=[RP]=] [=web application=], if known by the [=client=]. For example, the [=client=] knows the parameters passed to the [=authenticatorMakeCredential=] operation and whether the operation was successful or not. The [=client=] can then use this knowledge to report to the caller.

At this time, one [=credential property=] is defined, {{CredentialPropertiesOutput/rk}}, which indicates whether or not the returned {{PublicKeyCredential}} is a [=resident credential=].

     <xmp class="idl">
    dictionary CredentialPropertiesOutput {
        boolean rk;
    };

        CredentialPropertiesOutput credProps;
    };
    </xmp>

index.bs
```diff
@@ -4983,6 +5129,12 @@ IANA "WebAuthn Extension Identifier" registry established by [[!WebAuthn-Registr
     The user verification method extension returns to the [=[WRP]=] which user verification methods (factors) were
     used for the WebAuthn operation.
 - Specification Document: Section [[#sctn-uvm-extension]] of this specification
+    <br/><br/>
+- WebAuthn Extension Identifier: credProps
+- Description: This [=registration extension=] and [=authentication extension=] enables reporting on credential properties.
+    The credential properties extension returns to the [=[WRP]=] information about the credential used for the WebAuthn operation.
```
equalsJeffH: this is only a registration extension, yes?

suggest:

- Description: This [=registration extension=] enables reporting of a newly-created [=credential=]'s properties to the calling [=[WRP]=] [=web application=].

index.bs
```diff
@@ -2043,9 +2101,16 @@ attributes.
...
     :   <dfn>requireResidentKey</dfn>
-    ::  This member describes the [=[RP]=]'s requirements regarding [=resident credentials=].
-        If the parameter is set to [TRUE], the authenticator MUST create a
-        [=client-side-resident public key credential source=] when creating a [=public key credential=].
+    ::  This member is retained for backwards compatibility but is deprecated in favour of providing *residentKey*.
+        It will be ignored if *residentKey* is supplied and understood by the user-agent. When provided, it describes the 
+        [=[RP]=]'s requirements regarding [=resident credentials=]. If the parameter is set to [TRUE], the authenticator MUST create 
+        a [=client-side-resident public key credential source=] when creating a [=public key credential=].
+
+    :   <dfn>residentKey</dfn>
```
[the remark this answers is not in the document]
emlun: I disagree; `residentKey: ResidentKeyRequirement` is at least more style-consistent with `userVerification: UserVerificationRequirement` and `attestation: AttestationConveyancePreference`.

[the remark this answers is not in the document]
sbweeden: I'm with @emlun on this one. I've tried to remain consistent with the established convention for other parameters.

index.bs
```diff
@@ -2033,6 +2090,7 @@ attributes.
     dictionary AuthenticatorSelectionCriteria {
         AuthenticatorAttachment      authenticatorAttachment;
         boolean                      requireResidentKey = false;
+        ResidentKeyRequirement       residentKey;
```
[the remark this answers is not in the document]
sbweeden: See previous explanation

index.bs
```diff
@@ -601,6 +602,7 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
     By definition, the [=credential private key=] is always exclusively controlled by the [=authenticator=]. In the case of a
     [=resident credential=], the [=authenticator=] might offload storage of wrapped key material to the
     [=client device=], but the [=client device=] is not expected to offload the key storage to remote entities (e.g., [=[WRP]=] Server).
+    An authenticator is said to be using [=client-side credential storage modality=] when generating and storing a [=Resident Credential=].
```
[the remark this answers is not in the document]
sbweeden: Agreed - changed to the _See also_ suggestion.

index.bs
```diff
@@ -734,6 +736,13 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
         [=document.domain=]'s setter provides.
     </div>
 
+: <dfn>Non-Resident Credential</dfn>
+: <dfn>Server-side-resident Public Key Credential Source</dfn>
+:: A [=Server-side-resident Public Key Credential Source=], or [=Non-Resident Credential=] for short, is a [=public key credential
+    source=] whose [=credential private key=] is NOT stored in the [=authenticator=], [=client=] or [=client device=], but is instead
```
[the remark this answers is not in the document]
sbweeden: Agreed - fixed.

index.bs
```diff
@@ -4900,6 +4995,57 @@ as candidates to be employed in a [=registration ceremony=].
 : Authenticator extension output
 :: None.
 
+
+## Credential Properties Extension (credProps) ## {#sctn-authenticator-credential-properties-extension}
```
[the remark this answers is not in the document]
sbweeden: The approach I took is to be consistent with that of all the other extensions. If we want to establish a new reference pattern, can I suggest it be done in a new defect and PR?

index.bs
```diff
@@ -2084,6 +2144,40 @@ operation has no [=authenticator attachment modality=] selection option, so the
...
+    :   <dfn>required</dfn>
+    ::  This value indicates the client MUST use an authenticator capable of [=resident credentials=]. 
+        The authenticator MUST create a [=resident credential=] and MUST return an error if it cannot.
+</div>
+
+Note: It is intended for the [=[RP]=] to use the [[#sctn-authenticator-credential-properties-extension]] in conjunction with the 
+<code>|options|.{{PublicKeyCredentialCreationOptions/authenticatorSelection}}.{{residentKey}}</code> value to determine whether or not 
+the authenticator provisioned a [=resident credential=]. This is particularly useful when values of 
+{{ResidentKeyRequirement/discouraged}} or {{ResidentKeyRequirement/preferred}} are supplied.
+
```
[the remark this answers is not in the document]
sbweeden: I have done some re-wording to align with your suggestion.

index.bs
sbweeden: I have made a few small updates, and would like you to take another look. Keep in mind that I don't want to preclude the idea that authenticators themselves (in future) could actually return this extension data - not just the client. 

There is also not a current definition for `[=credential property=]`.

index.bs
sbweeden: Good pickup. This was original defined as both registration and authentication extension. Now fixed.

sbweeden:
@equalsJeffH - can you please take another look. There are a couple of open discussions but nothing major.

index.bs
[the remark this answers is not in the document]
equalsJeffH: ok, I suppose.  In hindsight, I'd rename {residentKey, userVerification, attestation} to indicate that they are the actual expressions of requirements/preferences, but I spose we're stuck.

index.bs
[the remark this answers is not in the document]
equalsJeffH: yep, fuggedaboutit.

index.bs
```diff
@@ -734,6 +736,13 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
         [=document.domain=]'s setter provides.
     </div>
 
+: <dfn>Non-Resident Credential</dfn>
+: <dfn>Server-side-resident Public Key Credential Source</dfn>
+:: A [=Server-side-resident Public Key Credential Source=], or [=Non-Resident Credential=] for short, is a [=public key credential
+    source=] whose [=credential private key=] is NOT stored in the [=authenticator=], [=client=] or [=client device=], but is instead 
+    encrypted (i.e. wrapped) and returned to the server as the [=credential ID=]. An authenticator is said to be using 
+    [=server-side credential storage modality=] when generating a [=Non-Resident Credential=].
```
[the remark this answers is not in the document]
equalsJeffH: what I mean here is I'd change:
> An authenticator is said to be using [=server-side credential storage modality=] when generating a [=Non-Resident Credential=].

..to be..
> See also: [=server-side credential storage modality=]

..because that sentence reads like a definition, and it's already defined elsewhere.  this is a nit I suppose. one could also move the definition to be that sentence and change the current dfn sentence to reference the sentence here.

index.bs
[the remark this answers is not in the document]
equalsJeffH: Actually, we have the \<dfn> in the title for the AppId extension, so there is precedent for doing this.   The AppId extension is referred to in this fashion: [=AppID|AppID extension=].  we haven't so modified the other extension titles (by adding \<dfn>s) becasesAFAICT we do not refer to any of the other extensions from elsewhere in the spec text.  So, anyway, I'd do this.

index.bs
```diff
@@ -4900,6 +4995,56 @@ as candidates to be employed in a [=registration ceremony=].
 : Authenticator extension output
 :: None.
 
+
+## Credential Properties Extension (credProps) ## {#sctn-authenticator-credential-properties-extension}
+
+This extension allows [=[WRPS]=] to query additional properties of the [=public key credential source=] created during a 
+[=registration ceremony=].
```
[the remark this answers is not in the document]
equalsJeffH: Here's a suggested definition for "credential properties":
```
: <dfn>Credential  Properties</dfn>
:: A [=credential property=] is some characteristic property of a credential, such as whether it is a [=resident credential=] or a [=non-resident=] credential.

File: index.bs (hunk 36 above)
Comment (equalsJeffH) [reply index.bs]: OK, well, I'd also add the "At this time, ..." sentence to define what "rk" means, otherwise I fear it won't be clear to the reader.  

>   I don't want to preclude the idea that authenticators themselves (in future) could actually return this extension data - not just the client.

Hm.. I do not understand...  "Authenticator extension output" below says that the authnr returns the credProps, yes?

> There is also not a current definition for [=credential property=].

correct, we should create one.  I've suggested one up above.

File: index.bs (hunk 44 above)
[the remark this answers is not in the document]
Comment (sbweeden) [reply index.bs]: Accepted and updated.

File: index.bs [hunk 46]
```diff
@@ -1129,13 +1138,56 @@ a numbered step. If outdented, it (today) is rendered as a bullet in the midst o
...
+                              is set to [TRUE] and the |authenticator| is not capable of storing a [=client-side-resident public
+                              key credential source=], [=iteration/continue=].
+                      </dl>
+
                   1. If <code>|options|.{{PublicKeyCredentialCreationOptions/authenticatorSelection}}.{{AuthenticatorSelectionCriteria/userVerification}}</code> is
                     set to {{UserVerificationRequirement/required}} and the |authenticator| is not capable of performing [=user
                     verification=], [=iteration/continue=].
 
+            1. Let |requireResidentKey| be the <dfn>effective resident key requirement for credential creation</dfn>, a Boolean value,
+                as follows.
```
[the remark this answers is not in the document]
sbweeden: Accepted and updated

index.bs
```diff
@@ -1129,13 +1138,56 @@ a numbered step. If outdented, it (today) is rendered as a bullet in the midst o
...
+                    <dl class="switch">
+
+                        :   is present and set to {{ResidentKeyRequirement/required}}
+                        ::  Let |requireResidentKey| be [TRUE].
+
+                        :   is present and set to {{ResidentKeyRequirement/preferred}}
+                        ::  If the |authenticator|
+
+                            <dl class="switch">
+                                :   is capable of [=client-side-resident public key credential source=]
```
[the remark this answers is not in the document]
sbweeden: Accepted and updated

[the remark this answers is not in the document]
sbweeden: Accepted and updated

index.bs
```diff
@@ -1129,13 +1138,56 @@ a numbered step. If outdented, it (today) is rendered as a bullet in the midst o
...
+                                ::  Let |requireResidentKey| be [TRUE].
+
+                                :   is not capable of [=client-side-resident public key credential source=], or if the client cannot determine authenticator capability,
```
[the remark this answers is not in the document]
sbweeden: Accepted and updated

index.bs
```diff
@@ -2043,9 +2096,16 @@ attributes.
         specified [[#attachment]].
 
     :   <dfn>requireResidentKey</dfn>
-    ::  This member describes the [=[RP]=]'s requirements regarding [=resident credentials=].
-        If the parameter is set to [TRUE], the authenticator MUST create a
-        [=client-side-resident public key credential source=] when creating a [=public key credential=].
+    ::  This member is retained for backwards compatibility but is deprecated in favour of providing {{residentKey}}.
```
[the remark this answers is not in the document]
sbweeden: Accepted and updated.

index.bs
```diff
@@ -2043,9 +2096,16 @@ attributes.
...
+        It will be ignored if {{residentKey}} is supplied and understood by the [=client=]. When provided, it describes the
```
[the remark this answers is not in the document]
sbweeden: Accepted and updated.

index.bs
```diff
@@ -2084,6 +2144,40 @@ operation has no [=authenticator attachment modality=] selection option, so the
...
+    };
+</xmp>
+
+This enumeration's values describe the [=[RP]=]'s requirements for [=resident credentials=]:
+
+<div dfn-type="enum-value" dfn-for="ResidentKeyRequirement">
+    :   <dfn>discouraged</dfn>
+    ::  This value indicates the client, when communicating with an authenticator, SHOULD NOT instruct the authenticator
+        to create a [=resident credential=]. The authenticator SHOULD create a [=non-resident credential=], however either
+        a [=resident credential=] or [=non-resident credential=] will be accepted by the [=[RP]=].
```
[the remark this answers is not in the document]
sbweeden: Accepted - taken a shot.

index.bs
```diff
@@ -2084,6 +2144,40 @@ operation has no [=authenticator attachment modality=] selection option, so the
...
+
+    :   <dfn>preferred</dfn>
+    ::  This value indicates the client, when communicating with an authenticator, SHOULD instruct the authenticator
+        to create a [=resident credential=]. If the authenticator is incapable of [=resident credentials=], 
+        the client MAY use the authenticator to create a [=non-resident credential=].
```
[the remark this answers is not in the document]
sbweeden: Accepted and updated.

index.bs
```diff
@@ -2084,6 +2144,40 @@ operation has no [=authenticator attachment modality=] selection option, so the
...
+
+    :   <dfn>required</dfn>
+    ::  This value indicates the client MUST use an authenticator capable of [=resident credentials=]. 
+        The authenticator MUST create a [=resident credential=] and MUST return an error if it cannot.
```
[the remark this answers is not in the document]
sbweeden: Accepted and updated with s/the the/the/

index.bs
```diff
@@ -2084,6 +2144,42 @@ operation has no [=authenticator attachment modality=] selection option, so the
...
+</div>
+
+Note: The [=[RP]=] can use a combination of the value provided for |options|.{{PublicKeyCredentialCreationOptions/authenticatorSelection}}.{{residentKey}} 
+and the [[#sctn-authenticator-credential-properties-extension]] [=registration extension=]'s return value to determine
```
equalsJeffH: strongly suggest s/ [[#sctn-authenticator-credential-properties-extension]] / [=credProps=]  /   otherwise it'll read quite oddly with a section number and full title in there....

index.bs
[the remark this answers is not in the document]
equalsJeffH: thx, tho pls see above comment.

index.bs
```diff
@@ -2835,7 +2930,7 @@ It takes the following input parameters:
 : |userEntity|
 :: The user account's {{PublicKeyCredentialUserEntity}}, containing the [=user handle=] given by the [=[RP]=].
 : |requireResidentKey|
-:: The {{PublicKeyCredentialCreationOptions/authenticatorSelection}}.{{requireResidentKey}} value given by the [=[RP]=].
+:: The [=effective resident key requirement for credential creation=], a Boolean value provided by the client.
```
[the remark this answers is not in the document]
sbweeden: I have changed the provided->determined wording, both for this definition and for effective user verification requirement.

There is already an anchoring hyperlink in the text which links back to step 20.2 by use of [=effective resident key requirement for credential creation=]. This follows the same pattern established for effective user verification requirement a couple of lines further down. I am not [yet] seeing extra value in the additional anchoring pattern you have suggested. Is there really value in that, or is it just another way?

index.bs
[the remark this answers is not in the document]
sbweeden: Accepted - done. Also updated where it was referenced from to be:

the [=credProps|Credential Properties Extension=]'s return value

index.bs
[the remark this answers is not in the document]
sbweeden: Accepted and updated with:

s/[=non-resident=]/[=non-resident credential=]/

index.bs
sbweeden: Ok - I obviously missed a bunch of the earlier comments. I've made the changes you've suggested, but left in this sentence:

```
The client MAY omit any properties at its discretion, and SHOULD omit any 
    properties that would be duplicated in the authenticator extension output.
```

The reason for that sentence is that if an authenticator itself reported on the rk property, the client would not need to and should not duplicate it. That's what I meant when I said that authenticators themselves could return the extension data. So might the client ... so the idea was to instruct the client not to duplicate extension parameters that an authenticator supplies.

index.bs
```diff
@@ -2043,9 +2096,16 @@ attributes.
...
+    ::  This member is retained for backwards compatibility but is deprecated in favour of providing {{residentKey}}.
+        It will be ignored if {{residentKey}} is supplied and understood by the [=client=]. When provided, it describes the 
+        [=[RP]=]'s requirements regarding [=resident credentials=]. If the parameter is set to [TRUE], the authenticator MUST create 
+        a [=client-side-resident public key credential source=] when creating a [=public key credential=].
+
+    :   <dfn>residentKey</dfn>
+    ::  This member supersedes the use of {{requireResidentKey}} to describe the [=[RP]=]'s requirements regarding 
+        [=resident credentials=]. Only one of {{residentKey}} or {{requireResidentKey}} should be present in 
+        {{AuthenticatorSelectionCriteria}}, and use of {{residentKey}} is encouraged. If both are present 
+        and the [=client=] supports it, {{residentKey}} takes precedence over {{requireResidentKey}}.
```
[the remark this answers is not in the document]
sbweeden: I have updated this - I put the very last sentence outside the `Note:`

index.bs
sbweeden: This has been fixed.

sbweeden:
@equalsJeffH  - ready for another round. Sorry I missed all those earlier review comments. Couple of questions I've left open for discussion.

index.bs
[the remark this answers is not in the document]
equalsJeffH: LGTM, thx.

index.bs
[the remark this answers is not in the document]
equalsJeffH: LGTM, thx.

index.bs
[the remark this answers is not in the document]
equalsJeffH: LGTM, thx  :)

index.bs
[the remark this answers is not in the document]
equalsJeffH: > There is already an anchoring hyperlink in the text...

ah, I'd overlooked that linkage, this LGTM, thx!

index.bs
equalsJeffH: Ok, thanks, silly me, I think I understand now.   

Given that authnr extn outputs are distinct from client extn outputs (tho one sure needs to read this spec carefully to figure that out),  there's a few things we need to fix and further clarify in specifying this extension.

1.  I'm assuming your intent is for the client to report on whether the authn created a resident key in the event that the authnr does not support the extension and does not itself return an output value.  

2. The authnr extn output is a CBOR-encoded value that we need to express in prose and/or CDDL, rather than webIDL.  We do not have any defined means for translating from WebIDL-expressed data structures to CBOR, so we need to express the authnr extn output here in CBOR/CDDL.  E.g., see the authnr extn output for the "exts" and "txAuthSimple" extensions. 

Thoughts?

I'm working on a rewrite of this extension section which addresses the above. I'll submit it as a pull request to your sbweeden/webauthn repo later today (I hope).

sbweeden: Yes, that's the idea. Thanks for working on the PR - I'll check it out when available.

index.bs
```diff
@@ -619,6 +621,10 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
 
     [=[RPS]=] do not need to distinguish these two [=Credential ID=] forms.
 
+: <dfn>Credential Properties</dfn>
+:: A [=credential property=] is some characteristic property of a credential, such as whether it is a 
+    [=resident credential=] or a [=non-resident credential=] credential.
```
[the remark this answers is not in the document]
sbweeden: Good pickup - fixed.

equalsJeffH:
Hi @sbweeden -- please see: https://github.com/sbweeden/webauthn/pull/3

sbweeden:
I merged https://github.com/sbweeden/webauthn/pull/3 however made two small subsequent changes. The first was to fix a compile error related to indenting, and the other a small grammatical error of "a" vs "an" before a word starting with a vowel.

equalsJeffH:
@sbweeden -- great, thx, this is now ready for some re-review from others: @emlun @agl @akshayku others?

index.bs
```diff
@@ -4900,6 +5008,83 @@ as candidates to be employed in a [=registration ceremony=].
...
+
+: Authenticator extension processing
+:: None, other than to report on credential properties in the output.
+
+: Authenticator extension output
+:: If the [=authenticator=] supports this extension it returns a CBOR map with one entry:
+	<pre>
+	CDDL:
+	{
+	  rk: bool
```
akshayku: "rk" key is a string right?

index.bs
```diff
@@ -4900,6 +5008,83 @@ as candidates to be employed in a [=registration ceremony=].
...
+
+Generally, this extension allows for both, or either, the [=authenticator=] or the [=client=] to report on [=credential properties=]. However, certain [=credential properties=], known as <dfn>source-specific credential properties</dfn>, are defined to be reported on by only the [=authenticator=] or by the [=client=].
+
+In the opposite case, known as <dfn>all-sources credential properties</dfn>, it is possible for either the [=authenticator=] or the [=client=] to report on a certain [=credential property=]. Additionally, the [=client=] MAY so report if the [=client=] determines the [=authenticator=] failed to report, and the [=client=] possesses the appropriate data to report.
+
+At this time, one [=credential property=], which is an [=all-sources credential property=], is defined. It is known abstractly as the <dfn>resident key credential property</dfn>. It is a boolean value indicating whether or not the {{PublicKeyCredential}} returned as a result of a [=registration ceremony=] is a [=resident credential=] or a [=non-resident credential=].  
+
+: Extension identifier
+:: `credProps`
```
akshayku: This also needs to be indicated in authenticatorGetInfo. May be open a seperate PR in FIDO2 specs or mention here that this identifier will be present in authenticatorGetInfo.

index.bs
```diff
@@ -4900,6 +5008,83 @@ as candidates to be employed in a [=registration ceremony=].
...
+    dictionary CredentialPropertiesOutput {
+        boolean rk;
+    };
+
+    partial dictionary AuthenticationExtensionsClientOutputs {
+        CredentialPropertiesOutput credProps;
+    };
+    </xmp>
+
+: Authenticator extension input
```
akshayku: Actually, why do we need authenticator extension at all.. I thought platform is going to figure out which one to create by inspecting authenticators and authenticators don't have to change.??

akshayku: I would remove Authenticator extension input/output sections for now. You already has "Client Extension Output" in your design. There is no need to add "Authenticator extension input/output here". Platform can figure out via other means. CTAP spec is the right place to figure out how best this information to be communicated back to the platform.

index.bs
equalsJeffH: good catch, thx. see https://github.com/fido-alliance/fido-2-specs/issues/660

index.bs
sbweeden: The format seems consistent with other CDDL definitions in the spec.

index.bs
emlun: All extensions are optional, so authenticators do not have to change. But the authenticator has the best knowledge of the credential it created, so why not allow the authenticator report that back if it can?

There's also the difference that authenticator extension outputs are signed, while client extension outputs are not.

index.bs
sbweeden: Removing authenticator extension component of the credProps extension per WG call on May 1, 2019 so this is now moot.

index.bs
sbweeden: Removing authenticator extension component of the credProps extension per WG call on May 1, 2019 so this is now moot.

index.bs
sbweeden: Removing authenticator extension component of the credProps extension per WG call on May 1.

sbweeden:
Ready for re-review following removal of authenticator extension component of the credProps extension per WG call on May 1, 2019.

index.bs
```diff
@@ -621,6 +622,10 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
 
     [=[RPS]=] do not need to distinguish these two [=Credential ID=] forms.
 
+: <dfn>Credential Properties</dfn>
+:: A [=credential property=] is some characteristic property of a credential, such as whether it is a
```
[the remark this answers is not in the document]
equalsJeffH: Yes, good catch, thx :)

index.bs
```diff
@@ -4983,6 +5138,12 @@ IANA "WebAuthn Extension Identifier" registry established by [[!WebAuthn-Registr
     The user verification method extension returns to the [=[WRP]=] which user verification methods (factors) were
     used for the WebAuthn operation.
 - Specification Document: Section [[#sctn-uvm-extension]] of this specification
+    <br/><br/>
+- WebAuthn Extension Identifier: credProps
+- Description: This [=client extension|client=] and [=authenticator extension|authenticator=] [=registration extension=] enables reporting of a newly-created [=credential=]'s properties to the
```
equalsJeffH: s/ and [=authenticator extension|authenticator=] //    -- ie delete it  :)

additionally, suggest:
s/ [=credential=]'s properties / [=credential=]'s properties, as determined by the [=client=] /

index.bs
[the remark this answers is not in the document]
sbweeden: Accepted and updated.

index.bs
sbweeden: Accepted and updated.

sbweeden:
Thanks for latest review @emlun and @equalsJeffH. I have addressed both the pick-ups, and found one additional dfn reference error which I fixed as well.

emlun:
With the changes @agl proposes in fido-alliance/fido-2-specs#661 it looks to me like we actually could support `forbidden` here. Should we? (We could do that in a separate PR if we don't want to hold this one up for it)

index.bs
```diff
@@ -5140,8 +5140,8 @@ IANA "WebAuthn Extension Identifier" registry established by [[!WebAuthn-Registr
 - Specification Document: Section [[#sctn-uvm-extension]] of this specification
     <br/><br/>
 - WebAuthn Extension Identifier: credProps
-- Description: This [=client extension|client=] and [=authenticator extension|authenticator=] [=registration extension=] enables reporting of a newly-created [=credential=]'s properties to the 
-    calling [=[WRP]=]'s [=web application=].
+- Description: This [=client extension|client=] [=registration extension=] enables reporting of a newly-created [=credential=]'s properties,
+    as determine by the [=client=] to the calling [=[WRP]=]'s [=web application=].
```
emlun: s/determine/determined/

emlun: Also, I think we need a comma after `[=client=]`?

equalsJeffH: yep, thanks :)

equalsJeffH:
@sbweeden 
> I have addressed both the pick-ups, and found one additional dfn reference error which I fixed as well.

Yep, looks good.  changing 'credential' to 'pk cred source' is good catch, thx :)
Would be good to add the comma and fix grammar error that @emlun caught.

@emlun 
> With the changes @agl proposes in fido-alliance/fido-2-specs#661 it looks to me like we actually could support forbidden here. Should we? (We could do that in a separate PR if we don't want to hold this one up for it)

you mean add "forbidden" as a fourth value of [ResidentKeyRequirement](file:///usr/local/google/home/jdhodges/_stds/w3c/webauthn/equalsjeffh/fork-sbweeden-webauthn/index.html#enumdef-residentkeyrequirement) ?

emlun:
>@equalsJeffH: you mean add "forbidden" as a fourth value of ResidentKeyRequirement ?

Yes.

equalsJeffH:
I think we should merge this and separately consider possibly adding "forbidden" as a fourth value of ResidentKeyRequirement.

index.bs
sbweeden: Agreed and updated.

PR #1095, opened 2018-10-09
Rewrite security consideration on attestation and MitM attacks

emlun:
This fixes #1088.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/pull/1095.html" title="Last updated on May 2, 2019, 11:36 AM UTC (a82ea4c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/1095/5d11e62...a82ea4c.html" title="Last updated on May 2, 2019, 11:36 AM UTC (a82ea4c)">Diff</a>

PR #1203, opened 2019-04-23
Add some embellishments to @equalsJeffH's suggestions

emlun:
@equalsJeffH How about this?

***
<a href="https://pr-preview.s3.amazonaws.com/emlun/webauthn/pull/1203.html" title="Last updated on May 2, 2019, 11:34 AM UTC (a82ea4c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/1203/12cd06a...emlun:a82ea4c.html" title="Last updated on May 2, 2019, 11:34 AM UTC (a82ea4c)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -5402,7 +5402,7 @@ For example, the [=authenticator=] model, or how it stores and protects [=creden
 However, it is important to note that an [=attestation statement=], on its own,
 provides no means for a [=[RP]=] to verify that an [=attestation object=] was generated
 by the [=authenticator=] the user intended, and not by a [=man-in-the-middle attack|man-in-the-middle attacker=].
-For example, malicious code injected into [=[RP]=] script.
+For example, such an attacker could use malicious code injected into [=[RP]=] script.
 The [=[RP]=] must therefore rely on other means, e.g., TLS and related technologies,
 to protect the [=attestation object=] from [=man-in-the-middle attacks=].
 
@@ -5417,7 +5417,7 @@ same [=[RP]=] and passing through the same attacker.
 
 Such an attack would potentially be detectable; since the [=[RP]=] has registered the attacker's [=credential public key=] rather
 than the user's, the attacker must tamper with all subsequent [=authentication ceremonies=] with that [=[RP]=]: unscathed
-ceremonies will fail.
+ceremonies will fail, risking to reveal the attack.
 
 [=Attestation types=] other than [=Self Attestation=] and [=None=] can increase the difficulty of such attacks, since [=[RPS]=]
 can possibly display [=authenticator=] information, e.g., model designation, to the user. An attacker might therefore need to use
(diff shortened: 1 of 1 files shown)


emlun:
Forgot to mention: this would merge into #1095.

index.bs
```diff
@@ -5417,15 +5417,15 @@ same [=[RP]=] and passing through the same attacker.
 
 Such an attack would potentially be detectable; since the [=[RP]=] has registered the attacker's [=credential public key=] rather
 than the user's, the attacker must tamper with all subsequent [=authentication ceremonies=] with that [=[RP]=]: unscathed
-ceremonies will fail.
+ceremonies will fail, risking to reveal the attack.
```
equalsJeffH: nice, tho:
s/ , risking to reveal / , potentially revealing the attack. /

index.bs
```diff
@@ -5417,15 +5417,15 @@ same [=[RP]=] and passing through the same attacker.
...
 
 [=Attestation types=] other than [=Self Attestation=] and [=None=] can increase the difficulty of such attacks, since [=[RPS]=]
 can possibly display [=authenticator=] information, e.g., model designation, to the user. An attacker might therefore need to use
 a genuine [=authenticator=] of the same model as the user's [=authenticator=], or the user might notice that the [=[RP]=] reports
 a different [=authenticator=] model than the user expects.
 
 Note: All variants of [=man-in-the-middle attacks=] described above are more difficult for an attacker to mount
-than a [=man-in-the-middle attack=] against a [=[RP]=] employing conventional password authentication.
+than a [=man-in-the-middle attack=] against a [=[RP]=] employing conventional password authentication only.
```
equalsJeffH: I dunno if I'd say "only" because RPs that adopt webauthn/fido are, it seems, likely to support legacy uname/password along with webauthn/fido for at least a while....

index.bs
```diff
@@ -5402,7 +5402,7 @@ For example, the [=authenticator=] model, or how it stores and protects [=creden
 However, it is important to note that an [=attestation statement=], on its own,
 provides no means for a [=[RP]=] to verify that an [=attestation object=] was generated
 by the [=authenticator=] the user intended, and not by a [=man-in-the-middle attack|man-in-the-middle attacker=].
-For example, malicious code injected into [=[RP]=] script.
+For example, such an attacker could use malicious code injected into [=[RP]=] script.
```
equalsJeffH: nice, thx

index.bs
emlun: Hm, yes, that's what I was thinking too... Do you read it as "more difficult only if the RP uses password auth"?

Maybe the "only" should be between "employing" and "conventional" instead:

>more difficult for an attacker to mount than a [=man-in-the-middle attack=] against a [=[RP]=] employing only conventional password authentication.

emlun: I think we've identified the same ambiguity but are approaching it from opposite directions. :) So maybe a better solution is something like this:

>Note: All variants of [=man-in-the-middle attacks=] described above are more difficult for an attacker to mount than a [=man-in-the-middle attack=] against conventional password authentication.

equalsJeffH: LGTM, thx!

index.bs
emlun: Done, thanks!

Issue #911, opened 2018-05-15
Integrate with Feature Policy and possibly Permissions and define appropriate identifier value for WebAuthn


equalsJeffH:
[update 3-Apr-2019](https://github.com/w3c/webauthn/issues/911#issuecomment-479596069): wrt @agl's [comment](https://github.com/w3c/webauthn/issues/911#issuecomment-470718913):
Yes wrt [Permissions](https://w3c.github.io/permissions/), tho this general "how to safely orchestrate cross-origin + iframe webauthn usage" topic, along with the potentially enabling technologies ([feature policy](https://w3c.github.io/webappsec-feature-policy/), [permissions](https://w3c.github.io/permissions/), [digital asset links](https://developers.google.com/digital-asset-links/), et al), is complex (cf. the [original post](https://github.com/w3c/webauthn/issues/911#issue-323420798)). I expect that we have more investigation to do here wrt feature policy vs permissions and such before deciding on a particular solution approach.

[update 24-Jul-2018](https://github.com/w3c/webauthn/issues/911#issuecomment-407582250): Recent further investigation reveals that this is actually a technical issue if one wishes to actually integrate with [Feature Policy](https://wicg.github.io/feature-policy/), i.e., defining a [feature-identifier value](https://github.com/WICG/feature-policy/blob/master/features.md) is only one aspect. One also posibly needs to update their algorithms to check feature policy and act accordingly. See the [payment-handler integration with feature-policy](https://github.com/w3c/payment-handler/pull/282/files), see also how [picture-in-picture integrates feature policy evaluation in its algorithm](https://wicg.github.io/picture-in-picture/#request-pip), and [defines its feature name](https://wicg.github.io/picture-in-picture/#feature-policy).

----
**original post (OP):**

(framed?) webapps may request permission to use the PaymentRequest interface via the [feature-policy](https://wicg.github.io/feature-policy/) [feature-identifier value](https://github.com/WICG/feature-policy/blob/master/features.md) `payment`.  

Given the desire for webauthn to be usable in a webpayment context and its being a "powerful feature", ought we mint a `webauthn` [feature-identifier value](https://github.com/WICG/feature-policy/blob/master/features.md) ?

this may depend to some degree upon resolution of w3c/permissions#177

see also wicg/feature-policy#168, wicg/feature-policy#166,  wicg/feature-policy#75

see also discussion here:  [Relationship of Permissions, Feature Policy, Origin Policy?
[...]

PR #1214, opened 2019-05-09
feature policy integration

equalsJeffH:
An initial pass at feature policy integration. Possibly is work-in-progress to a modest extent given how associated mods to credman (see w3c/webappsec-credential-management#138) shakes out.  I.e., we may (or may not) need to add a new parameter to the `[[Create]]()` and `[[DiscoverFrom...]]()` internal methods in addition to `SameOriginWithAncestors`, we'll see...

in its present state on 15-May-2019, this PR improves #911

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/pull/1214.html" title="Last updated on May 15, 2019, 6:44 PM UTC (7f8e256)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/1214/1b13d2e...7f8e256.html" title="Last updated on May 15, 2019, 6:44 PM UTC (7f8e256)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2541,7 +2541,7 @@ to successfully invoke the [=Web Authentication API=], i.e., via
 <code><a idl for="CredentialsContainer" lt="create()">navigator.credentials.create({publicKey:..., ...})</a></code> and
 <code><a idl for="CredentialsContainer" lt="get()">navigator.credentials.get({publicKey:..., ...})</a></code>.
 If disabled in any document, no content in the document will be [=allowed to use=]
-the foregoing methods: attempting to do so will throw an error.
+the foregoing methods: attempting to do so will [return an error](https://www.w3.org/2001/tag/doc/promises-guide#errors).
 
 Note: Algorithms specified in [[!CREDENTIAL-MANAGEMENT-1]] perform the actual feature policy evaluation. This is because such policy evaluation needs to occur when there is access to the [=current settings object=]. The {{PublicKeyCredential/[[Create]](origin, options, sameOriginWithAncestors)}} and {{PublicKeyCredential/[[DiscoverFromExternalSource]](origin, options, sameOriginWithAncestors)}} [=internal methods=] do not have such access since they are invoked [=in parallel=] (by algorithms specified in [[!CREDENTIAL-MANAGEMENT-1]]).
 
--- a/index.bs
+++ b/index.bs
@@ -2428,10 +2428,10 @@ needs.
 ## Feature Policy integration ## {#sctn-feature-policy}
 
 This specification defines a [=policy-controlled feature=] identified by
-the string "<code><dfn data-lt="payment-feature">webauthn</dfn></code>".
+the string "<code><dfn data-lt="webauthn-feature">webauthn</dfn></code>".
 Its [=default allowlist=] is '<code>self</code>'. [[!Feature-Policy]]
 
-A {{Document}}'s [=Document/feature policy=] determines whether any content in that document is allowed
+A {{Document}}'s [=Document/feature policy=] determines whether any content in that <a href="https://html.spec.whatwg.org/multipage/dom.html#documents">document</a> is allowed
 to successfully invoke the [=Web Authentication API=]&mdash;i.e.,
 {{CredentialsContainer/create()|navigator.credentials.create({"publicKey"})}} and
 {{CredentialsContainer/get()|navigator.credentials.get({"publicKey"})}}.
--- a/index.bs
+++ b/index.bs
@@ -92,6 +92,15 @@ figure.table .overlarge {
 </style>
 
 
+<pre class="link-defaults">
+
+spec: dom; type: interface; for:/; text: Document
+
+spec: html; urlPrefix: https://html.spec.whatwg.org/multipage/dom.html; type: dfn; url: concept-document-feature-policy; text: feature policy
+
+</pre>
+
+
 <!-- TODO: Clean out these anchor lists once they appear in Shepherd -->
 <pre class="anchors">
 
@@ -154,7 +163,7 @@ spec: page-visibility; urlPrefix: https://www.w3.org/TR/page-visibility/
         text: visibility states
 
 spec: WHATWG HTML; urlPrefix: https://html.spec.whatwg.org/
-    type: dfn 
+    type: dfn
         text: focus
         text: username; url: attr-fe-autocomplete-username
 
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -2527,6 +2530,31 @@ needs.
...
 
 
+## Feature Policy integration ## {#sctn-feature-policy}
+
+This specification defines a [=policy-controlled feature=] identified by
+the feature policy token "<code><dfn data-lt="webauthn-feature">webauthn</dfn></code>".
+Its [=default allowlist=] is '<code>self</code>'. [[!Feature-Policy]]
+
+A {{Document}}'s [=Document/feature policy=] determines whether any content in that <a href="https://html.spec.whatwg.org/multipage/dom.html#documents">document</a> is allowed
+to successfully invoke the [=Web Authentication API=], i.e., via
```
agl: &ldquo;i.e.&rdquo; to &ldquo;e.g.&rdquo;, since there are other Webauthn calls (IsUVPAA) and there may be others in the future?

index.bs
```diff
@@ -2527,6 +2530,31 @@ needs.
...
+<code><a idl for="CredentialsContainer" lt="create()">navigator.credentials.create({publicKey:..., ...})</a></code> and
+<code><a idl for="CredentialsContainer" lt="get()">navigator.credentials.get({publicKey:..., ...})</a></code>.
+If disabled in any document, no content in the document will be [=allowed to use=]
+the foregoing methods: attempting to do so will throw an error.
```
agl: Not just the foregoing methods, but any Webauthn method, no?

equalsJeffH: Ah, good question, thx. Presently, [FP](https://www.w3.org/TR/feature-policy/) is integrated into only the `[[Create]]()` and `[[DiscoverFrom...]]()` internal methods.  I neglected `IsUVPAA()`, will look into it.   

I am guessing we do not need to address `PublicKeyCredential.getClientExtensionResults()` because if it was not minted by  the `[[Create]]()` and `[[DiscoverFrom...]]()` internal methods (ie some random JS code does a `new PublicKeyCredential`, there's nothing to return from `getClientExtensionResults()`.  And such random code would be able to mint a `PublicKeyCredential` regardless of feature-policy, yes?  

Similar argument for `AuthenticatorAttestationResponse.getTransports()`.

I do not see any other WebAuthn API-specific methods... ?

equalsJeffH:
WRT integrating [Feature Policy (FP)](https://www.w3.org/TR/feature-policy/) with  [`isUserVerifyingPlatformAuthenticatorAvailable()`](http://w3c.github.io/webauthn/#isUserVerifyingPlatformAuthenticatorAvailable): the latter's impl is [client-platform](http://w3c.github.io/webauthn/#client-platform)-specific.

I _suppose_ we would want it to [return an error](https://www.w3.org/2001/tag/doc/promises-guide#errors) if webauthn is "Disabled" in the calling context?

If so, would this build on the present text in the [iUVPAA() section](http://w3c.github.io/webauthn/#isUserVerifyingPlatformAuthenticatorAvailable) work?  One issue is that [iUVPAA()](http://w3c.github.io/webauthn/#isUserVerifyingPlatformAuthenticatorAvailable) returns a boolean -- can that also be morphed into an error code?  [Promises are only sposed to _return_ errors, not throw them](https://www.w3.org/2001/tag/doc/promises-guide#always-return-promises). First parag is unchanged, only added the Note:

> WebAuthn Relying Parties use this method to determine whether they can create a new credential using a user-verifying platform authenticator. Upon invocation, the client employs a client platform-specific procedure to discover available user-verifying platform authenticators. If any are discovered, the promise is resolved with the value of true. Otherwise, the promise is resolved with the value of false. Based on the result, the Relying Party can take further actions to guide the user to create a credential.
>
> Note: Invoking this method from a [=browsing context=] where the [=Web Authentication API=] is "disabled" according to the [=allowed to use=] algorithm&mdash;i.e., by a [=feature policy=]&mdash;will result in the return of an error code equivalent to "{{NotAllowedError}}".
>
> This method has no arguments and returns a Boolean value.

index.bs
equalsJeffH: @agl see below https://github.com/w3c/webauthn/pull/1214#issuecomment-491014211 wrt IsUVPAA().

emlun:
>One issue is that [iUVPAA()](http://w3c.github.io/webauthn/#isUserVerifyingPlatformAuthenticatorAvailable) returns a boolean -- can that also be morphed into an error code? [Promises are only sposed to _return_ errors, not throw them](https://www.w3.org/2001/tag/doc/promises-guide#always-return-promises). First parag is unchanged, only added the Note:

I don't think that is a problem. The guideline you quote says that any _function_ that returns a promise should always return a promise and never throw an exception, but a promise can be either _resolved_ with a success value or _rejected_ with a failure value (usually an exception). So it should be fine to have `isUVPAA()` return a rejected promise containing an exception. One could argue that this would be a breaking API change since the current description of the function reads:

>Upon invocation, the client employs a client platform-specific procedure to discover available user-verifying platform authenticators. If any are discovered, the promise is resolved with the value of `true`. Otherwise, the promise is resolved with the value of `false`.

...so as written it seems to indicate it will never fail. But one could also argue that this is not a breaking change since there always was the theoretical possibility that it would return a rejected promise.

emlun:
De we also need to say something about what this means for what RP IDs are permitted in various contexts?

equalsJeffH:
@emlun 
> Do we also need to say something about what this means for what RP IDs are permitted in various contexts?

short answer: I don't think so.  

long answer: The [RP ID](https://w3c.github.io/webauthn/#rp-id) text itself takes into account only the caller's effective domain et al. It does not discuss cross-originess.

longer answer: we need to think thru the ramifications of feature-policy's cross-origin enablement-ness.  e.g., if  credman+webauthn is invoked from a context that is _not_ `SameOriginWithAncestors`, though, _is_ "allowed to run", _should_ it run?  Various folks are asking for the answer to be "yes, it should run".  This would impact the language wrt `SameOriginWithAncestors` within the #createCred and #getAssertion algs.

Also, some guidance we've heard wrt enabling webauthn in cross-origin nested browsing contexts is that a key aspect to it will be materializing some guidance to the user, like having the UA and/or authenticator materialize a UX that clearly designates the context of such cross-origin interactions to users, I.e., we need to explain to users that their login to example.com is being handled by foobar.baz and it's legit, and that it's under UA control and not a phish.

index.bs
equalsJeffH: in any case, done.

equalsJeffH:
attempt to address @agl's https://github.com/w3c/webauthn/pull/1214#pullrequestreview-235709437 including @emlun's guidance https://github.com/w3c/webauthn/pull/1214#issuecomment-491299616 is in 7f8e256.

index.bs
agl: > I am guessing we do not need to address `PublicKeyCredential.getClientExtensionResults()` [&hellip;] similar argument for `AuthenticatorAttestationResponse.getTransports()`

Yep, I think that's correct.

emlun:
> I don't think this PR need include it, but I did wonder whether the [collectedClientData](https://w3c.github.io/webauthn/#dictdef-collectedclientdata) should include an optional `topLevelOrigin` so that it's clear when Webauthn is operating inside an iframe.

This sounds like a great idea to me.

nadalin:
@akshayku Please review

nadalin:
@emlun Please merge

PR #1219, opened 2019-05-15
Browser capability detection.

### LINKED ISSUE #1199: How to get supported client extension
Currently, the WebAuthn does not define the way for RP to get supported extension that the client supports.
For the authenticator extension support, there is a _exts_ extension so that RP may obtain such information during the create() call.
Similarly, the RP would like to know which extensions are supported by user's client.

### LINKED ISSUE #1202: Supported features detection from the browser
This issue is related to the #1199 and this is more broader cases.
We, LINE, are trying to leverage WebAuthn for our login flow and suffering some issues handling different browsers/platforms.
RPs supporting WebAuthn already may have some logics to enable or disable supporting WebAuthn by looking at user agent information.
For real word implementation, it is very hard for us to detect such supported or not-supported features from UA.
It is better to have some APIs to provide supported features for RP to easily leverage the features.

### LINKED ISSUE #1218: Provide API for indicating whether the platform has attached(bound) authenticator or not
WebAuthn API has [isUVPAA()](https://www.w3.org/TR/webauthn/#isUserVerifyingPlatformAuthenticatorAvailable) for RP to check whether the current user device has available platform authenticator.
This API may return true if the following conditions are met:

1. The platform provides or has attached (bound) WebAuthn authenticators
2. The authenticator is enabled (feature turned on) and user verification method (like PIN or biometrics) is enrolled on the authenticator if it supports uv.

Sometime RP wants to promote WebAuthn even in case where the user has platform authenticator but it is not enabled.
If there is such API, RPs are willing to check and show some guides for users to enable the authenticator depending on the policy.

I'm not sure how much of the users' devices do have the platform authenticator but not enabled for Android/Win10.

jafisher-microsoft:
Fixes #1218. Fixes #1202. Fixes #1199.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/pull/1219.html" title="Last updated on Jun 18, 2019, 7:41 PM UTC (1d414c6)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/1219/db94f4a...1d414c6.html" title="Last updated on Jun 18, 2019, 7:41 PM UTC (1d414c6)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -46,6 +46,7 @@ Former Editor: Angelo Liao, w3cid 94342, Microsoft, huliao@microsoft.com
 !Contributors: <a href="mailto:mkwst@google.com">Mike West</a> (Google)
 !Contributors: <a href="mailto:jyasskin@google.com">Jeffrey Yasskin</a> (Google)
 !Contributors: <a href="mailto:sweeden@au1.ibm.com">Shane Weeden</a> (IBM)
+!Contributors: <a href="mailto:jafisher@microsoft.com">Jason Fisher</a> (Microsoft)
 group: webauthn
 Issue Tracking: GitHub https://github.com/w3c/webauthn/issues
 !Tests: <a href=https://github.com/web-platform-tests/wpt/tree/master/webauthn>web-platform-tests webauthn/</a> (<a href=https://github.com/web-platform-tests/wpt/labels/webauthn>ongoing work</a>)
@@ -1776,6 +1777,29 @@ This [=internal method=] accepts no arguments.
 
 </div>
 
+### Availability of [=User-Verifying Platform Authenticator=] (enum <dfn enum>UserVerifyingPlatformAuthenticatorAvailability</dfn>) ### {#enum-userVerifyingPlatformAuthenticatorAvailability}
+
+<xmp class="idl">
+    enum UserVerifyingPlatformAuthenticatorAvailability {
+        "enabled",
+        "disabled",
+        "unavailable"
+    };
+</xmp>
+
+This enumeration's values describe the availability of a [=User-Verifying Platform Authenticator=] on the [=client platform=].
+
+<div dfn-type="enum-value" dfn-for="UserVerifyingPlatformAuthenticatorAvailability">
+    :   <dfn>enabled</dfn>
+    ::  This value indicates that a [=User-Verifying Platform Authenticator=] is available and enabled on the [=client platform=].
+
+    :   <dfn>disabled</dfn>
+    ::  This value indicates that a [=User-Verifying Platform Authenticator=] is available, but not enabled on the [=client platform=].
+
+    :   <dfn>unavailable</dfn>
+    ::  This value indicates that a [=User-Verifying Platform Authenticator=] is not available on the [=client platform=].
+</div>
+
 ### Availability of [=User-Verifying Platform Authenticator=] - PublicKeyCredential's `isUserVerifyingPlatformAuthenticatorAvailable()` Method ### {#sctn-isUserVerifyingPlatformAuthenticatorAvailable}
 
 <div link-for-hint="WebAuthentication/isUserVerifyingPlatformAuthenticatorAvailable">
@@ -1797,6 +1821,66 @@ This method has no arguments and returns a Boolean value.
 Note: Invoking this method from a [=browsing context=] where the [=Web Authentication API=] is "disabled" according to the [=allowed to use=] algorithm&mdash;i.e., by a [=feature policy=]&mdash;will result in the promise being rejected with a [=DOMException=] whose name is "{{NotAllowedError}}".
 
 </div>
+
+### Client Supported Capabilities (dictionary <dfn dictionary>SupportedCapabilities</dfn>) ### {#dictionary-supportedcapabilities}
+
+<xmp class="idl">
+    dictionary SupportedCapabilities {
+        boolean areU2FDevicesSupported;
+        boolean areCTAPDevicesSupported;
+        boolean areResidentKeysSupported;
+        boolean isUserVerificationSupported;
+        boolean isTokenBindingSupported;
+        UserVerifyingPlatformAuthenticatorAvailability userVerifyingPlatformAuthenticatorAvailability;
+        sequence<AuthenticatorTransport> transportsSupported;
+        sequence<DOMString> extensionsSupported;
+    };
+</xmp>
+
+This dictionary is used by [=[WRPS]=] to determine the supported capabilities of the [=client=].
+
+<div dfn-type="dict-member" dfn-for="SupportedCapabilities">
+    :   <dfn>areU2FDevicesSupported</dfn>
+    ::  This attribute indicates whether the [=client=] supports communicating with [[FIDO-U2F-Message-Formats]] devices.
+
+    :   <dfn>areCTAPDevicesSupported</dfn>
+    ::  This attribute indicates whether the [=client=] supports communicating with [[FIDO-CTAP]] devices.
+
+    :   <dfn>areResidentKeysSupported</dfn>
+    ::  This attribute indicates whether the [=client=] supports creating [=resident credentials=].
+
+    :   <dfn>isUserVerificationSupported</dfn>
+    ::  This attribute indicates whether the [=client=] supports [=user verification=].
+
+    :   <dfn>isTokenBindingSupported</dfn>
+    ::  This attribute indicates whether the [=client=] supports the [=Token Binding=] protocol.
+
+    :   <dfn>userVerifyingPlatformAuthenticatorAvailability</dfn>
+    ::  This attribute indicates the availability of a [=user-verifying platform authenticator=] on the [=client=].
+
+    :   <dfn>transportsSupported</dfn>
+    ::  This attribute indicates the {{AuthenticatorTransport}} values supported by the [=client=].
+
+    :   <dfn>extensionsSupported</dfn>
+    ::  This attribute indicates the [=extension identifiers=] supported by the [=client=].
+</div>
+
+### Capability Support - PublicKeyCredential's `getSupportedCapabilities()` Method ### {#sctn-getSupportedCapabilities}
+
+<div link-for-hint="WebAuthentication/getSupportedCapabilities">
+
+[=[WRPS]=] use this method to determine what capabilities are supported by the [=client=].
+Upon invocation, the [=client=] employs a [=client platform=]-specific procedure to discover supported capabilities.
+The promise is resolved with the capabilities that are supported.
+Based on the result, the [=[RP]=] can take further actions to guide the user to create a credential.
+
+This method has no arguments and returns a {{SupportedCapabilities}} value.
+
+<xmp class="idl">
+    partial interface PublicKeyCredential {
+        static Promise<SupportedCapabilities> getSupportedCapabilities();
+    };
+</xmp>
 
 ## Authenticator Responses (interface <dfn interface>AuthenticatorResponse</dfn>) ## {#iface-authenticatorresponse}
 
--- a/index.bs
+++ b/index.bs
@@ -1777,29 +1777,6 @@ This [=internal method=] accepts no arguments.
 
 </div>
 
-### Availability of [=User-Verifying Platform Authenticator=] (enum <dfn enum>UserVerifyingPlatformAuthenticatorAvailability</dfn>) ### {#enum-userVerifyingPlatformAuthenticatorAvailability}
-
-<xmp class="idl">
-    enum UserVerifyingPlatformAuthenticatorAvailability {
-        "enabled",
-        "disabled",
-        "unavailable"
-    };
-</xmp>
-
-This enumeration's values describe the availability of a [=User-Verifying Platform Authenticator=] on the [=client platform=].
-
-<div dfn-type="enum-value" dfn-for="UserVerifyingPlatformAuthenticatorAvailability">
-    :   <dfn>enabled</dfn>
-    ::  This value indicates that a [=User-Verifying Platform Authenticator=] is available and enabled on the [=client platform=].
-
-    :   <dfn>disabled</dfn>
-    ::  This value indicates that a [=User-Verifying Platform Authenticator=] is available, but not enabled on the [=client platform=].
-
-    :   <dfn>unavailable</dfn>
-    ::  This value indicates that a [=User-Verifying Platform Authenticator=] is not available on the [=client platform=].
-</div>
-
 ### Availability of [=User-Verifying Platform Authenticator=] - PublicKeyCredential's `isUserVerifyingPlatformAuthenticatorAvailable()` Method ### {#sctn-isUserVerifyingPlatformAuthenticatorAvailable}
 
 <div link-for-hint="WebAuthentication/isUserVerifyingPlatformAuthenticatorAvailable">
@@ -1822,49 +1799,6 @@ Note: Invoking this method from a [=browsing context=] where the [=Web Authentic
 
 </div>
 
-### Client Supported Capabilities (dictionary <dfn dictionary>SupportedCapabilities</dfn>) ### {#dictionary-supportedcapabilities}
-
-<xmp class="idl">
-    dictionary SupportedCapabilities {
-        boolean areU2FDevicesSupported;
-        boolean areCTAPDevicesSupported;
-        boolean areResidentKeysSupported;
-        boolean isUserVerificationSupported;
-        boolean isTokenBindingSupported;
-        UserVerifyingPlatformAuthenticatorAvailability userVerifyingPlatformAuthenticatorAvailability;
-        sequence<AuthenticatorTransport> transportsSupported;
-        sequence<DOMString> extensionsSupported;
-    };
-</xmp>
-
-This dictionary is used by [=[WRPS]=] to determine the supported capabilities of the [=client=].
-
-<div dfn-type="dict-member" dfn-for="SupportedCapabilities">
-    :   <dfn>areU2FDevicesSupported</dfn>
-    ::  This attribute indicates whether the [=client=] supports communicating with [[FIDO-U2F-Message-Formats]] devices.
-
-    :   <dfn>areCTAPDevicesSupported</dfn>
-    ::  This attribute indicates whether the [=client=] supports communicating with [[FIDO-CTAP]] devices.
-
-    :   <dfn>areResidentKeysSupported</dfn>
-    ::  This attribute indicates whether the [=client=] supports creating [=resident credentials=].
-
-    :   <dfn>isUserVerificationSupported</dfn>
-    ::  This attribute indicates whether the [=client=] supports [=user verification=].
-
-    :   <dfn>isTokenBindingSupported</dfn>
-    ::  This attribute indicates whether the [=client=] supports the [=Token Binding=] protocol.
-
-    :   <dfn>userVerifyingPlatformAuthenticatorAvailability</dfn>
-    ::  This attribute indicates the availability of a [=user-verifying platform authenticator=] on the [=client=].
-
-    :   <dfn>transportsSupported</dfn>
-    ::  This attribute indicates the {{AuthenticatorTransport}} values supported by the [=client=].
-
-    :   <dfn>extensionsSupported</dfn>
-    ::  This attribute indicates the [=extension identifiers=] supported by the [=client=].
-</div>
-
 ### Capability Support - PublicKeyCredential's `getSupportedCapabilities()` Method ### {#sctn-getSupportedCapabilities}
 
 <div link-for-hint="WebAuthentication/getSupportedCapabilities">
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -1798,6 +1822,66 @@ Note: Invoking this method from a [=browsing context=] where the [=Web Authentic
...
+</div>
+
+### Capability Support - PublicKeyCredential's `getSupportedCapabilities()` Method ### {#sctn-getSupportedCapabilities}
+
+<div link-for-hint="WebAuthentication/getSupportedCapabilities">
+
+[=[WRPS]=] use this method to determine what capabilities are supported by the [=client=].
+Upon invocation, the [=client=] employs a [=client platform=]-specific procedure to discover supported capabilities.
+The promise is resolved with the capabilities that are supported.
+Based on the result, the [=[RP]=] can take further actions to guide the user to create a credential.
```
Kieun: RP may also invoke this method for getting a credential.

index.bs
```diff
@@ -1798,6 +1822,66 @@ Note: Invoking this method from a [=browsing context=] where the [=Web Authentic
...
+    :   <dfn>areU2FDevicesSupported</dfn>
+    ::  This attribute indicates whether the [=client=] supports communicating with [[FIDO-U2F-Message-Formats]] devices.
+
+    :   <dfn>areCTAPDevicesSupported</dfn>
+    ::  This attribute indicates whether the [=client=] supports communicating with [[FIDO-CTAP]] devices.
+
+    :   <dfn>areResidentKeysSupported</dfn>
+    ::  This attribute indicates whether the [=client=] supports creating [=resident credentials=].
+
+    :   <dfn>isUserVerificationSupported</dfn>
```
Kieun: Is this different from isUserVerifyingPlatformAuthenticatorAvailable()?

jafisher-microsoft: isUserVerifyingPlatformAuthenticatorAvailable currently indicates the enabled state of the UVPA, but the RP cannot distinguish between disabled/unavailable.  I left isUserVerifyingPlatformAuthenticatorAvailable alone as RPs are currently using this API, but reflected the tri-state in this new API.

index.bs
jafisher-microsoft: Agreed, this is also useful for login scenarios.  I can update the text to reflect that.

index.bs
Kieun: I was asking about the purpose of isUserVerificationSupported() function. We have userVerifyingPlatformAuthenticatorAvailability() and isUVPAA(). What is the difference?

jasonf1990: Thanks for clarifying.  The intention of isUserVerificationSupported is so that the client can covey whether it supports the UserVerificationRequirement enum.  On the other hand, userVerifyingPlatformAuthenticatorAvailability is intended to indicate the specific state of the UVPA on the client platform.  Does that make sense?

Kieun: Make sense.
So, this method is for indicating the client supports userVerification option. IMO, it is better to refer [userVerification](https://w3c.github.io/webauthn/#dom-authenticatorselectioncriteria-userverification) member instead.

index.bs
```diff
@@ -1798,6 +1822,66 @@ Note: Invoking this method from a [=browsing context=] where the [=Web Authentic
 
 </div>
 
+### Client Supported Capabilities (dictionary <dfn dictionary>SupportedCapabilities</dfn>) ### {#dictionary-supportedcapabilities}
+
+<xmp class="idl">
+    dictionary SupportedCapabilities {
+        boolean areU2FDevicesSupported;
+        boolean areCTAPDevicesSupported;
```
leshi: Can you explain the user flow for how you plan to use `areU2FDevicesSupported` and `areCTAPDevicesSupported`?

index.bs
leshi: I suggest that the for this (and all other similar functionality detection), we remove these methods and say that if a particular capability is not part of  the response from `getSupportedCapabilities()`, then it's not supported.

index.bs
jafisher-microsoft: CTAP devices are not supported on every browser.  RP will look into this field to decide what they want to support and what they want to show on their page.  User Agent detection is too fragile to make these decisions and it requires constant monitoring of what browser/platform combination works.

index.bs
jafisher-microsoft: If I understand the comment of saying to remove isUserVerifyingPlatformAuthenticatorAvailable(), browsers already implement this function and RPs already depend on it.  I don't think we should remove it.

equalsJeffH:
Here's an overall comment on "browser capability detection': 

I asked [Tab Atkins](https://github.com/tabatkins) (CSS spec editor) about this (back in April), and he said that the
```
if (foo.bar()) /* do something */ else /* not there, go figure */
```
approach, colloquially known as "feature detection", often is the best one can do from JS within a user agent. 

There are feature detection JS libraries &mdash; e.g., modernizr, yepnope &mdash; but results with them can be mixed. Tab says that SVG tried specifying feature label strings (i suppose one asked the svg impl 'what is supported?' and got a list of strings in return), and that sorta worked for a very short time and then implementations would lie about what they supported, then the number of features outgrew the set of strings, etc.  

Tab asserts similar efforts occurred in various portions of the web platform with lousy results overall.  In CSS, they have a `@support <css stuff to test here>` rule, which nominally works because CSS is a language, they have a parser, and the assumption is: if the @support rule can successfully parse the `<css stuff to test here>`, then it's supported.

WebAuthn/FIDO2 is an especially tough problem because we have a browser->protocols->authnr stack and the answer to "is foo feature supported" question may involve component answers from several layers in the stack.

dwaite:
The transports result might need to be defined as what the client + operating system can support, and not what currently is supported by available hardware and settings. Examples what this may be appropriate:

- CoreBluetooth on the Mac is available, but the user may have the radio turned off
- NFC support is not built into the mac, but could hypothetically be provided by an external reader. Like above, this could mean NFC support comes and goes during authentication

I suppose the user might toggle the UVPA as well.

index.bs
Kieun: I would prefer the way suggested by @leshi . WebAuthn spec will keep introducing new features. Returning the supported features as a list of string is better and easy to extend.

equalsJeffH: yes, IIUC, whatever we do, we need to leave `isUserVerifyingPlatformAuthenticatorAvailable()` nominally intact, and if we do decide to deprecate it in light of other new functionality we define, we ought to mark it (in the spec) as deprecated and point to the new functionality.

equalsJeffH:
just fyi; @agl opened: w3ctag/design-reviews#383 regarding this here PR.

index.bs
```diff
@@ -1798,6 +1822,66 @@ Note: Invoking this method from a [=browsing context=] where the [=Web Authentic
 
 </div>
 
+### Client Supported Capabilities (dictionary <dfn dictionary>SupportedCapabilities</dfn>) ### {#dictionary-supportedcapabilities}
+
+<xmp class="idl">
+    dictionary SupportedCapabilities {
+        boolean areU2FDevicesSupported;
```
jcjones: Will this one ever be used? Can we imagine this being set to `false` and for RPs to code themselves to handle it?

index.bs
```diff
@@ -1798,6 +1822,66 @@ Note: Invoking this method from a [=browsing context=] where the [=Web Authentic
...
+
+<div dfn-type="dict-member" dfn-for="SupportedCapabilities">
+    :   <dfn>areU2FDevicesSupported</dfn>
+    ::  This attribute indicates whether the [=client=] supports communicating with [[FIDO-U2F-Message-Formats]] devices.
+
+    :   <dfn>areCTAPDevicesSupported</dfn>
+    ::  This attribute indicates whether the [=client=] supports communicating with [[FIDO-CTAP]] devices.
+
+    :   <dfn>areResidentKeysSupported</dfn>
+    ::  This attribute indicates whether the [=client=] supports creating [=resident credentials=].
```
jcjones: If you support CTAP devices, isn't this always going to be `true`? Or are clients supposed to guess what users' devices in their pocket can do, and fuzzily adjust depending on what's been paired/plugged in of late?

index.bs
jcjones: Point of order: U2F == CTAP1, and CTAP2 is the new thing, right?

Seems like this should be `areCTAP2DevicesSupported`? Or maybe it should just be `minimumCTAPVersionSupported` and make it a float?

index.bs
```diff
@@ -1798,6 +1822,66 @@ Note: Invoking this method from a [=browsing context=] where the [=Web Authentic
...
+    :   <dfn>isUserVerificationSupported</dfn>
+    ::  This attribute indicates whether the [=client=] supports [=user verification=].
+
+    :   <dfn>isTokenBindingSupported</dfn>
+    ::  This attribute indicates whether the [=client=] supports the [=Token Binding=] protocol.
+
+    :   <dfn>userVerifyingPlatformAuthenticatorAvailability</dfn>
+    ::  This attribute indicates the availability of a [=user-verifying platform authenticator=] on the [=client=].
+
+    :   <dfn>transportsSupported</dfn>
```
jcjones: We have to define the strict ordering of these two sequences to reduce fingerprinting opportunity.

index.bs
```diff
@@ -1776,6 +1777,29 @@ This [=internal method=] accepts no arguments.
...
 
+### Availability of [=User-Verifying Platform Authenticator=] (enum <dfn enum>UserVerifyingPlatformAuthenticatorAvailability</dfn>) ### {#enum-userVerifyingPlatformAuthenticatorAvailability}
+
+<xmp class="idl">
+    enum UserVerifyingPlatformAuthenticatorAvailability {
+        "enabled",
+        "disabled",
+        "unavailable"
+    };
+</xmp>
```
emlun: This differentiation of whether the platform authenticator is configured is one bit of information that's not determined by the User-Agent header. On the other hand, I guess the same conclusions can probably be drawn already from a combination of User-Agent and `isUserVerifyingPlatformAuthenticatorAvailable()` result? (i.e., this leaks fingerprinting information, but no more than we already do?)

index.bs
```diff
@@ -1798,6 +1822,66 @@ Note: Invoking this method from a [=browsing context=] where the [=Web Authentic
...
+
+    :   <dfn>extensionsSupported</dfn>
+    ::  This attribute indicates the [=extension identifiers=] supported by the [=client=].
+</div>
+
+### Capability Support - PublicKeyCredential's `getSupportedCapabilities()` Method ### {#sctn-getSupportedCapabilities}
+
+<div link-for-hint="WebAuthentication/getSupportedCapabilities">
+
+[=[WRPS]=] use this method to determine what capabilities are supported by the [=client=].
```
emlun: s/client/client platform/

index.bs
emlun: Would it make sense to define them as [ordered sets](https://infra.spec.whatwg.org/#sets)? I agree we should probably also explicitly define the order like we have for [`[[transports]]`](https://www.w3.org/TR/webauthn-2/#dom-authenticatorattestationresponse-transports-slot).

index.bs
```diff
@@ -1798,6 +1822,66 @@ Note: Invoking this method from a [=browsing context=] where the [=Web Authentic
...
+        boolean areResidentKeysSupported;
+        boolean isUserVerificationSupported;
+        boolean isTokenBindingSupported;
+        UserVerifyingPlatformAuthenticatorAvailability userVerifyingPlatformAuthenticatorAvailability;
+        sequence<AuthenticatorTransport> transportsSupported;
+        sequence<DOMString> extensionsSupported;
+    };
+</xmp>
+
+This dictionary is used by [=[WRPS]=] to determine the supported capabilities of the [=client=].
```
emlun: Throughout this section: s/[=client=]/[=client platform=]/

emlun:
As was discussed on recent WG calls, I think that `isUserVerificationSupported` is really getting at whether the client supports PIN entry - because on-authenticator UV such as a fingerprint scanner requires no special support from the client, and UV on a platform authenticator is supported if and only if `isUVPAA()` returns true.

However: suppose we rename `isUserVerificationSupported` to `isPinEntrySupported`. Is that actually useful to the RP? What if the user wants to use a biometric authenticator? Since there's no way for the RP to control whether to use PIN or other UV, won't the RP have to resort to `userVerification: "preferred"/"required"` anyway?

index.bs
agl: Agreed about saying &ldquo;CTAP2&rdquo;.

Minimum version might be a problem for Safari: don't they _only_ support CTAP2 devices at the moment?

index.bs
```diff
@@ -1798,6 +1822,66 @@ Note: Invoking this method from a [=browsing context=] where the [=Web Authentic
...
+    :   <dfn>transportsSupported</dfn>
+    ::  This attribute indicates the {{AuthenticatorTransport}} values supported by the [=client=].
+
+    :   <dfn>extensionsSupported</dfn>
+    ::  This attribute indicates the [=extension identifiers=] supported by the [=client=].
+</div>
+
+### Capability Support - PublicKeyCredential's `getSupportedCapabilities()` Method ### {#sctn-getSupportedCapabilities}
+
+<div link-for-hint="WebAuthentication/getSupportedCapabilities">
```
agl: Will Microsoft's webauthn.dll export a reflection of this function for browsers running on >= 19H1? Some of the fields are browser specific, but some depend on the platform.

index.bs
```diff
@@ -1798,6 +1822,66 @@ Note: Invoking this method from a [=browsing context=] where the [=Web Authentic
...
 </div>
 
+### Client Supported Capabilities (dictionary <dfn dictionary>SupportedCapabilities</dfn>) ### {#dictionary-supportedcapabilities}
+
+<xmp class="idl">
+    dictionary SupportedCapabilities {
+        boolean areU2FDevicesSupported;
+        boolean areCTAPDevicesSupported;
+        boolean areResidentKeysSupported;
+        boolean isUserVerificationSupported;
```
agl: As I think was mentioned in the call, `isUserVerificationSupported` seems odd and I think you might want `isClientPINSupported`.

index.bs
agl: Chrome at the moment (version 75) supports CTAP2 devices, but doesn't have support for account selection etc and so would return `false` for this because resident keys aren't actually usable. I.e. I think these fields are reporting on browser capabilities so it's &ldquo;assuming that the user has an authenticator that does resident keys, will it actually work with this browser?&rdquo;.

index.bs
agl: For a list, I always default to saying that it should be ordered lexicographically and not contain duplicates. Looks like [ordered sets](https://infra.spec.whatwg.org/#sets) imply the uniqueness condition, so that would work.

index.bs
Kieun: I think **isUserVerificationSupported** is more about whether the client (browser) supports **userVerification** option for create() and get() call. Plus, it's better to have isClientPINSupported as well.

index.bs
leshi: But why do RPs care about CTAP2 vs CTAP1 (U2F)?  They care about *capabilities* and *features* -- not transports.  Both of which are covered below.

index.bs
leshi: Do we have other forms of user verification that hte client can do besides clientPin?  If not, I think `isClientPINSupported` is the correct name here.

emlun:
Edited OP to add close macros for all related issues.

index.bs
akshayku: This is for the scenario where RP can actually promote Platform authenticator if not configured. It is the same info from  user-agent and `isUserVerifyingPlatformAuthenticatorAvailable`. We just want to get away from user-agent handling all together. That is too fragile.

index.bs
akshayku: Yes

index.bs
emlun: Sorry, I didn't mean to say User-Agent matching is preferable, just pondering whether this would leak additional fingerprinting information. I think it won't since we already have `isUVPAA()`, but I figured I'd think aloud a bit in case someone else would notice something I didn't.

index.bs
jafisher-microsoft: I have removed references to an updated UVPAA check, as we currently don't have scenarios to leverage this additional granularity.  I have also adopted the style used by Transports to return the list of supported capabilities.

index.bs
jafisher-microsoft: Agreed that we do not currently have a use case for such granularity.  I have removed these particular capabilities.

index.bs
jafisher-microsoft: Agreed, this has been removed.

index.bs
jafisher-microsoft: Resident key support is still being returned as we have a use case for this, and browsers do react differently today.

index.bs
jafisher-microsoft: I have applied the same logic as was previously done for transports.

index.bs
jafisher-microsoft: UVPAA modifications have been removed as we don't have a concrete use case for this today.

index.bs
jafisher-microsoft: Updated for clientPin as opposed to UV.

jafisher-microsoft:
@emlun @leshi @Kieun @jcjones @agl , I have made changes based on feedback.  In particular, the capabilities have been scoped down significantly to concrete use cases that our Microsoft RP (and other RPs) have requested to take advantage of.  Namely whether the client platform will support Client PIN and whether or not it supports Resident Keys.  Both of these currently have states where some platform + browser + version combinations will say true vs. false.  Look forward to more discussions on this.

index.bs
Kieun: We have concrete use cases for this. RP tending to leverage the platform authenticator and promote it to users has no means to promote the user if they do not turn on the feature even the platform has the function.
For example, we would like to show following information.
1. If the user has platform authenticator and is available: "Try now"
2. If the user has platform authenticator and is disabled: "Your device has the best option for protecting your account. For trying this, please go to setup and turn on the feature. Then, try again"
3. If the user has no platform authenticator, never show any authentication promotion

We have seen many of people haven't setup Windows Hello feature on their device.

Kieun:
@jafisher-microsoft  Is `extensionsSupported` removed intentionally since there is no concrete usecase for this?

index.bs
```diff
@@ -2626,6 +2560,24 @@ needs.
...
+    };
+</xmp>
+
+<div dfn-type="enum-value" dfn-for="SupportedCapability">
+    Each enum value indicates a supported capability of the [=client platform=] for use by [=[WRPS]=].
+    :   <dfn>clientPin</dfn>
+    ::  This attribute indicates whether the [=client platform=] supports [=user verification=] via client PIN.
+
+    :   <dfn>residentKeys</dfn>
+    ::  This attribute indicates whether the [=client platform=] supports creating [=resident credentials=].
```
emlun: This is also about whether the client supports resident credential selection (for external authenticators without their own display), right? Or should that be a separate capability?

index.bs
```diff
@@ -2542,6 +2560,24 @@ needs.
...
+    enum SupportedCapability {
+        "clientPin",
+        "residentKeys"
+    };
+</xmp>
+
+<div dfn-type="enum-value" dfn-for="SupportedCapability">
+    Each enum value indicates a supported capability of the [=client platform=] for use by [=[WRPS]=].
+    :   <dfn>clientPin</dfn>
+    ::  This attribute indicates whether the [=client platform=] supports [=user verification=] via client PIN.
```
```suggestion
    ::  This value indicates that the [=client platform=] supports [=user verification=] via client PIN.
```

index.bs
```diff
@@ -2542,6 +2560,24 @@ needs.
...
+
+    :   <dfn>residentKeys</dfn>
+    ::  This attribute indicates whether the [=client platform=] supports creating [=resident credentials=].
```
```suggestion
    ::  This value indicates that the [=client platform=] supports creating [=resident credentials=].
```

rlin1:
What happens if 
(1) the platform doesn't support clientPIN, but
(2) a connected roaming authenticator support user verification through a fingerprint sensor?

equalsJeffH:
on webauthn call today:
@akshayku's described his use case (a primary one?): user logs in, is shown promo "want to add SK?" but wants to know what the browser supports in order to tailor the promo and not have the user go thru a bunch of the ceremony to then have the ceremony fail.

@leshi relates that in the google security key experience, they learned that users were really confused if they were offered to register a security key in one browser, but not even offered in another browser. the lesson was that it was more clear to users if they go thru part of the ceremony and if the browser does not support it then fail. 

overall situation seems to be:  
@akshayku is trying to work around the situation that different versions of different browsers support none, some, (nearly) all of the webauthn spec.  and then deployers are trying to workaround these browser differences.  observed that even if this PR is landed, it would be hopefully easier to impl than the features themselves, but there of course is the possibility that there'd be browser impl skew for even this feature-discovery facility.

sbweeden:
In general I think an RP has to deal with the case that the browser doesn't even support *this* proposed API anyway, and potential variances in behaviour of this API, so I think it's better that the RP always deals with the error case rather than trying to pre-emptively render different UX based on what this API is proposing.

index.bs
equalsJeffH: @leshi, what do you mean by "account selection" ?

emlun: (5), key management, isn't standardised in WebAuthn (yet?), so I'd say we can't make that requirement at this time.

akshayku:
@sbweeden has a very good point due to how the platforms implement the spec (only released published final version vs draft versions) and their support on various versions in different versions of the platform. And this API may not be present. Given this, I am not sure how much we are moving the needle and RPs has to manually be testing browser/platform combination for a while and this PR probably won't help that much. 

@jafisher-microsoft, What do you think?

akshayku:
Closing this down as right now things are not clear at this point in time..

PR #1232, opened 2019-06-05
Pass through other assertion formats

### LINKED ISSUE #465: ensure non-native Signature Assertions are passed-through
add support for UAF signature and assertion formats -- this will facilitate the **millions of already-deployed UAF-capable smartphones** to be usable with a WebAuthn-capable browser via CTAP (yes, a separate step is enabling the already-deployed UAF-capable smartphones to speak CTAP, which is possible)

this issue is addressed by these open PRs: #407 #408

rlin1:
Close #465

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/pull/1232.html" title="Last updated on Jun 5, 2019, 11:32 AM UTC (3981b67)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/1232/db94f4a...3981b67.html" title="Last updated on Jun 5, 2019, 11:32 AM UTC (3981b67)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1673,7 +1673,7 @@ When this method is invoked, the user agent MUST execute the following algorithm
                 ::  whose value is the bytes of the [=authenticator data=] returned by the [=authenticator=].
 
                       Note: Web Browsers shall tolerate assertion formats that differ from the one described 
-                      in [[#sec-authenticator-data]].
+                      in [[#sctn-authenticator-data]], e.g. FIDO UAF assertion [[FIDO-UAF-AUTHNR-CMDS]].
 
                 :   <code><dfn for="assertionCreationData">signatureResult</dfn></code>
                 ::  whose value is the bytes of the signature value returned by the [=authenticator=].
--- a/index.bs
+++ b/index.bs
@@ -1591,6 +1591,9 @@ When this method is invoked, the user agent MUST execute the following algorithm
                 :   <code><dfn for="assertionCreationData">authenticatorDataResult</dfn></code>
                 ::  whose value is the bytes of the [=authenticator data=] returned by the [=authenticator=].
 
+                      Note: Web Browsers shall tolerate assertion formats that differ from the one described 
+                      in [[#sec-authenticator-data]].
+
                 :   <code><dfn for="assertionCreationData">signatureResult</dfn></code>
                 ::  whose value is the bytes of the signature value returned by the [=authenticator=].
 


index.bs
```diff
@@ -1672,6 +1672,9 @@ When this method is invoked, the user agent MUST execute the following algorithm
                 :   <code><dfn for="assertionCreationData">authenticatorDataResult</dfn></code>
                 ::  whose value is the bytes of the [=authenticator data=] returned by the [=authenticator=].
 
+                      Note: Web Browsers shall tolerate assertion formats that differ from the one described
```
agl: If &ldquo;shall&rdquo; is intended to be binding, should it be &ldquo;SHALL&rdquo;?

rlin1: I am happy to make that change (plus remove the "note")

rlin1:
Could we return the UAF assertion in an extension that is included in a well-formed FIDO2 signature assertion?

agl:
From call of 2019-06-05:

Rolf clarified that registration would be handled out-of-band.

I, for Chrome, said that arbitrary Authenticator Data would be very problematic. Akshay, I think, said something similar. I suggested putting the necessary information in an extension and having the RP reconstruct the signed message from that.

There was worry about this being a channel for arbitrary data though the browser. Several such channels already exist, but people are thrilled about adding more.

J. C. queries how a reader of the spec would be able to figure out the necessary supporting documents to be able to use this. (I.e. the UAF specs.)

John asks about UAF extensions themselves, which currently aren't defined for WebAuthn and so couldn't be passed in.

Mike notes that no UAF authenticators work with WebAuthn today, so new code will be needed. But if a change is needed why not make these authenticators real CTAP2 devices? Rolf asserts that parts of this ecosystem are pretty hard to change, but adding an interop layer would be much easier.

John, J. C., and Mike ask about the number of devices that would partake in order to judge the value.

ve7jtb:
My more basic question is who really needs this? 
The use case seems to be for phones that have UAF authenticators in ROE that someone wants to wrap a BLE transport around so the phone can be a CTAP2 authenticator for a desktop.

As I understand it all Android phones 7+ have Webauthn platform authenticators now and Google is in the process of adding CTAP over BLE and caBLE so that they can be remote authenticators.   So current Android phones don't really seem to fit this description. 

On iOS apple may take a bit longer to roll out their platform authenticator, so this might be a target.  However, all Fido authenticators on iOS are in software not in a ROE other than using the keychain.  Those could just as easily be updated to Fido2 native authenticators rather than doing this wraping.

I am left to conclude that we are talking about a small number of phones that may have UAF built in perhaps on Samsung?  But those would have to be pre android 7.  Anything later and the android platform is going to be doing Fido over BLE.  I don't even know if a non core OS app could also do Fido over BLE at the same time without breaking things.  

This seems a lot of work for potentially a small to non-existent number of devices.

Can someone point to a real example of a phone that would do this?
If not it seems a lot of work for a hypothetical problem.

John B.

akshayku:
Thinking more about this, may be upgrading such implementations to do CTAP2 is more achievable. Changing webauthn and faking the authdata, passing info via an extension, does not seems right.

rlin1:
No update today, so I untied it from milestone L2-WD-02

nicksteele:
As per 3/3/21 Meeting: This pull request would need to be heavily updated to include further explanation of what's being proposed, justification of its purpose , and amendments to sections of the spec (such as those regarding clients and authentication servers) that would be affected by this change.

PR #1238, opened 2019-06-19
Clarify that a single user might be several persons

### LINKED ISSUE #1162: Refine User Verification description
The description of [User Verification ](https://w3c.github.io/webauthn/#user-verification)includes following sentence.

> The intent is to be able to distinguish individual users.

In fact, If we enroll multiple fingerprints of family members to the authenticator (device), the authenticator never identify the each users. But, this sentence may mislead to readers that the authenticator can distinguish or identify the users.

It is more like the process that checks the user's authority to register the credential or authenticate with the credential.

emlun:
Fixes #1162

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/pull/1238.html" title="Last updated on Jun 26, 2019, 4:08 PM UTC (a30bdf7)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/1238/00127b7...a30bdf7.html" title="Last updated on Jun 26, 2019, 4:08 PM UTC (a30bdf7)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -821,7 +821,7 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
         implies use of key material managed by the authenticator.
 
         Also, for security, [=user verification=] and use of [=credential private keys=]
-        must occur within a single logical security boundary within the [=authenticator=].
+        must occur within a single logical security boundary defining the [=authenticator=].
     </div>
 
     [=User verification=] procedures MAY implement [=rate limiting=] as a protection against brute force attacks.
--- a/index.bs
+++ b/index.bs
@@ -811,7 +811,7 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
 
     Note: Distinguishing natural persons depends in significant part upon the [=client platform=]'s
     and [=authenticator=]'s capabilities.
-    For example, some devices are intended to be used by one distinct individual,
+    For example, some devices are intended to be used by a single individual,
     yet they may allow multiple natural persons to enroll fingerprints
     and thus access the same [=[RP]=] account(s) using that device.
     See also [[#sctn-uvi-extension]].
--- a/index.bs
+++ b/index.bs
@@ -821,7 +821,7 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
         implies use of key material managed by the authenticator.
 
         Also, for security, [=user verification=] and use of [=credential private keys=]
-        must occur within a single logical security boundary defining the [=authenticator=].
+        must all occur within the logical security boundary defining the [=authenticator=].
     </div>
 
     [=User verification=] procedures MAY implement [=rate limiting=] as a protection against brute force attacks.
--- a/index.bs
+++ b/index.bs
@@ -821,7 +821,7 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
         implies use of key material managed by the authenticator.
 
         Also, for security, [=user verification=] and use of [=credential private keys=]
-        must occur within a single logical security boundary defining the [=authenticator=].
+        must occur within a single logical security boundary within the [=authenticator=].
     </div>
 
     [=User verification=] procedures MAY implement [=rate limiting=] as a protection against brute force attacks.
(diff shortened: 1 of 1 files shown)


emlun:
I suggested something like this in https://github.com/w3c/webauthn/issues/1162#issuecomment-467981906, but on further consideration I don't think this is the right way to go. This makes it seem like we're supporting account sharing as a good idea, but in many cases it would be better to have separate accounts for each person but have them set up to access the same things. People can still choose to share accounts anyway, of course, but I'm not sure we need to call that out in the spec as a supported use case.

I think it might be best to close this and #1162 with no further action, but I'd like to hear what others think.

index.bs
```diff
@@ -807,7 +807,11 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
     [=authenticatorMakeCredential=] and [=authenticatorGetAssertion=] operations. [=User verification=] MAY be instigated
     through various [=authorization gesture=] modalities; for example, through a touch plus pin code, password entry, or
     [=biometric recognition=] (e.g., presenting a fingerprint) [[ISOBiometricVocabulary]]. The intent is to be able to
-    distinguish individual users. Note that invocation of the [=authenticatorMakeCredential=] and [=authenticatorGetAssertion=]
+    distinguish individual users. This typically means distinguishing individual persons,
+    but in some cases a single user might be several persons,
```
equalsJeffH: s/ might be several persons /  might actually be several different natural persons /   ?

emlun:
@equalsJeffH Do you [think we should do this](https://github.com/w3c/webauthn/pull/1238#issuecomment-503676500)?

equalsJeffH:
> @equalsJeffH Do you [think we should do this](https://github.com/w3c/webauthn/pull/1238#issuecomment-503676500)?

Well, I'm thinking @Kieun has a good point that the "The intent is to be able to distinguish individual users" statement (within the [user verification definition](https://w3c.github.io/webauthn/#user-verification)) encompasses several subtle-but-important nuances. 

How about this:
```
: <dfn>User Verification</dfn>
:: The technical process by which an [=authenticator=] <em>locally authorizes</em> the invocation of the
    [=authenticatorMakeCredential=] and [=authenticatorGetAssertion=] operations. [=User verification=] MAY be instigated
    through various [=authorization gesture=] modalities; for example, through a touch plus pin code, password entry, or
    [=biometric recognition=] (e.g., presenting a fingerprint) [[ISOBiometricVocabulary]]. The intent is to
    distinguish individual users.

    <div class="note">
        Note: Distinguishing natural persons depends in significant part upon the [=client platform=]'s
        and [=authenticator=]'s capabilities. For example, some devices are intended to be used
        by one distinct individual, yet they may allow multiple natural persons to enroll fingerprints and thus
        access the same [=[RP]=] account(s) using that device. See also [[#sctn-uvi-extension]].
    </div>

    <div class="note">
        Note: Invocation of the [=authenticatorMakeCredential=] and [=authenticatorGetAssertion=]
        operations implies use of key material managed by the authenticator.

        Also, for security, [=user verification=] and use
        of [=credential private keys=] must occur within a single logical security boundary defining the [=authenticator=].
    </div>

    [=User verification=] procedures MAY implement [=rate limiting=] as a protection against brute force attacks.
```
...?

emlun:
@equalsJeffH Thanks, your proposal looks much better! :slightly_smiling_face:

Also while we're making changes here anyway, I took the liberty of changing the last note to "a single logical security boundary **within** the authenticator", because I can't make sense of what a "security boundary **defining** the authenticator" is supposed to mean.

equalsJeffH:
> what a "security boundary defining the authenticator" is supposed to mean

Oh, that's an attempt to encompass both notions of...
* a _physically monolithic authnr_, e.g., roaming usb/ble/nfc security keys, phone-as-a-security-key, etc.
* a _logical authnr_, e.g., stitched together platform biometric sensor(s), display, etc, via software, perhaps running in a TEE, or perhaps just a userland software authnr.

emlun:
Yes, ok, what I was trying to say is that even though I understand what it means, the formulation "a security boundary defining the authenticator" sounds unnecessarily confusing since we've already defined what the authenticator is. Another way I think it could be a bit better is by just changing it to "the security boundary defining the authenticator", so it refers to something existing ("the") instead of suddenly introducing something new ("a").

equalsJeffH:
go for it :)

emlun:
@equalsJeffH Does that look okay?

PR #1241, opened 2019-06-21
bikeshed cleanup: refs

### LINKED ISSUE #872: cleanup: mark & structure biblio references consistently
Presently, we've (myself included) become somewhat sloppy with our biblio references and auto-link entries.  There's at least two broad categories:

The first, for example, is that we are not consistent in marking citations as **(non-)normative**, e.g.:
```
[...]
        <td>
            The ...
            as defined in Section 7 of [[RFC8152]] ...
[...]
    1. Convert the COSE_KEY ... (see Section 7 of [[!RFC8152]]) to CTAP1/U2F user public key ...
[...]
```
...note the inconsistency in using the normative exclamation mark citation signal **`!`** in the above cites of RFC8152.  There are similar inconsistencies with cites of other specs.

It appears that if a referenced spec is marked normative (at least once?) then it is placed in the "Normative" biblio section, thus cleaning up is not high priority, but probably something we ought to do.

The second cleanup aspect of this is we are creating cross-spec links that are syntactically  inconsistent. E.g.:
```
<pre class="anchors">
[...]
spec: TokenBinding; urlPrefix: https://tools.ietf.org/html/draft-ietf-tokbind-protocol#
    type: dfn
        text: Token Binding
        text: Token Binding ID; url: section-3.2                   [A]
[...]
spec: FIDO-CTAP; urlPrefix: https://fidoalliance.org/specs/fido-v2.0-ps-20170927/fido-client-to-authenticator-protocol-v2.0-ps-20170927.html
    type: dfn
        text: CTAP2 canonical CBOR encoding form; 
           url: ctap2-canonical-cbor-encoding-form                [B]
[...]
</pre> <!-- class=anchors -->

[...]
[[RFC8152]] [Section 7](https://tools.ietf.org/html/rfc8152#section-7)      [C]
...
(see [[RFC8152]]
[Section 13.1](https://tools.ietf.org/html/rfc8152#section-13.1))          [D]
[...]
```
[IIUC](https://en.wiktionary.org/wiki/IIUC),  we ought to be able to craft link anchors in the same fashion as [A] and [B], for the cases of [C] and [D] such that we are using a consistent cross-spec linking approach.

equalsJeffH:
fixes #872 

note:  let's merge PR #1240 first, then merge-from-master into this branch which will clean up all the whitespace removal and make review easier.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/pull/1241.html" title="Last updated on Jun 26, 2019, 11:25 PM UTC (2fa1bee)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/1241/78ed130...2fa1bee.html" title="Last updated on Jun 26, 2019, 11:25 PM UTC (2fa1bee)">Diff</a>


index.bs
```diff
@@ -213,10 +220,15 @@ spec: SP800-800-63r3; urlPrefix: https://pages.nist.gov/800-63-3/sp800-63-3.html
...
         text: something you are; url: af
+
 spec: webidl; urlPrefix: https://heycam.github.io/webidl
     type: dfn;
         text: get a copy of the bytes held by the buffer source; url: dfn-get-buffer-source-reference
 
+spec: UTR29
+    urlPrefix: https://unicode.org/reports/tr29/
+        type: dfn; for:/; url: Grapheme_Cluster_Boundaries; text: grapheme cluster
+
```
emlun: Why is this biblio separate from the one at the end of the file?

equalsJeffH: because the UTR29 at end of file is a biblio entry, while the above is a cross-spec link specifically for "grapheme cluster"

emlun: Whoops, I put this on the wrong diff hunk...

index.bs
```diff
@@ -242,6 +254,11 @@ spec:webidl; type:interface; text:Promise
         "publisher": "IANA"
     },
 
+    "UTR29": {
+        "href": "http://www.unicode.org/reports/tr29/",
+        "title": "UNICODE Text Segmentation",
+        "publisher": "UNICODE Consortium"
+    },
```
emlun: Here https://github.com/w3c/webauthn/pull/1241/files#r297719677

Same thing?

emlun: I mean the whole `<pre class=biblio>`, not just this entry.

equalsJeffH: I'm sorry, I don't understand...?

emlun: :)

There are two of these blocks:

- https://github.com/w3c/webauthn/blob/master/index.bs#L237-L253
- https://github.com/w3c/webauthn/blob/master/index.bs#L6027-L6230

Why is that?

equalsJeffH: doh!  just an oversight me thinks -- I think it was me that created the 2nd `<pre class=biblio>` block. thx for pointing it out explicitly.  o_0

I have fixed it in subsequent commits...

equalsJeffH:
ok, can we merge this now?

PR #1239, opened 2019-06-19
Correctly refer to the FIDO U2F JavaScript API and informatively refe…

futureimperfect:
…rence it. Fixes #1233.

***
<a href="https://pr-preview.s3.amazonaws.com/futureimperfect/webauthn/pull/1239.html" title="Last updated on Jun 19, 2019, 9:10 PM UTC (6437b1f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/1239/00127b7...futureimperfect:6437b1f.html" title="Last updated on Jun 19, 2019, 9:10 PM UTC (6437b1f)">Diff</a>

PR #1244, opened 2019-06-24
Add “appidExclude” extension.

### LINKED ISSUE #1235: Can't exclude U2F credentals
There's no `appid` extension when making a credential. Therefore the excludeList, if any, is only excluding WebAuthn credentials. So sites doing the transition cannot exclude authenticators by U2F credential ID.

We could fix this in WebAuthn if we wished. From the call of 2019-06-12, it's not yet clear how people feel about it.

agl:
This change adds a registration-only “appidExclude” extension that allows platforms to make excludeCredentials effective when transitioning from U2F.

It's a separate extension from “appid” because that extension was documented in level one as being authentication-only and some implementations thus reject it for registration. Also, having an “appid” extension that's effective during registration invites people to believe that it does the “obvious” thing and allows the creation of U2F credentials, which isn't true.

Fixes #1235.

***
<a href="https://pr-preview.s3.amazonaws.com/agl/webauthn/pull/1244.html" title="Last updated on Jun 27, 2019, 10:21 PM UTC (916aad6)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/1244/78ed130...agl:916aad6.html" title="Last updated on Jun 27, 2019, 10:21 PM UTC (916aad6)">Diff</a>


index.bs
```diff
@@ -4675,6 +4675,45 @@ Instead, in step three, the comparison on the host is relaxed to accept hosts on
...
+: Client extension processing
+::  1. Let |facetId| be the result of passing the caller's [=origin=] to the FIDO algorithm for [=determining the FacetID of a calling application=].
+    1. Let |appId| be the extension input.
+    1. Pass |facetId| and |appId| to the FIDO algorithm for [=determining if a caller's FacetID is authorized for an AppID=]. If that algorithm rejects |appId| then return a "{{SecurityError}}" {{DOMException}}.
+    1. For each authenticator that supports the U2F protocol, if |excludeCredentialDescriptorList| is not empty then, before invoking [=authenticatorMakeCredential=], attempt to obtain a U2F assertion (without user-presence), for each element of |excludeCredentialDescriptorList|, using the SHA-256 hash of |appId| as the application parameter. If unsuccessful, continue with invoking [=authenticatorMakeCredential=]. Otherwise, cease normal processing of this authenticator and indicate in a platform-specific manner that the authenticator is inapplicable. For example, this could be in the form of UI, or could involve requesting user-consent from the authenticator and, upon receipt, treating it as if the authenticator had returned {{InvalidStateError}}.
```
equalsJeffH: I had to dig around to figure out how this works. IIUC, it seems that steps 1 - 3 (above) probably ought to occur around step 19 of [\[\[create\]\]()](https://w3c.github.io/webauthn/#sctn-createCredential), and step 4 (above) ought to occur either after or during Step 20.5 of [\[\[create\]\]()](https://w3c.github.io/webauthn/#sctn-createCredential).  Perhaps we oughta add links into those places in \[\[create\]\]()](https://w3c.github.io/webauthn/#sctn-createCredential) from the above. I have some alg-smithing suggestions for step 4 yet to write up. I can help with crafting this PR if you wish.

index.bs
```diff
@@ -4675,6 +4675,45 @@ Instead, in step three, the comparison on the host is relaxed to accept hosts on
 : Authenticator extension output
 :: None.
 
+## FIDO AppID Exclusion Extension (appidExclude) ## {#sctn-appid-exclude-extension}
+
+This registration extension allows [=[WRPS]=] to exclude authenticators which contain specified credentials that were created with FIDO APIs.
```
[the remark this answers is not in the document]

index.bs
agl: Indeed, there were a lot of commas in there. I've restructured this a bit, see if you like it better. You also have direct push access to this repo. (I.e. `git remote add agl git@github.com:agl/webauthn.git` then `git push agl HEAD:appidexclude` to push the current branch to this PR.)

equalsJeffH:
agreed that we should merge PR #1239 first

equalsJeffH:
PR #1239 has been merged.  Plus, I did a merge-from-master since master changed today.

index.bs
```diff
@@ -4990,6 +4990,85 @@ Instead, in step three, the comparison on the host is relaxed to accept hosts on
 : Authenticator extension output
 :: None.
 
+## FIDO AppID Exclusion Extension (appidExclude) ## {#sctn-appid-exclude-extension}
+
+This registration extension allows [=[WRPS]=] to exclude authenticators which contain specified credentials that were created with the legacy FIDO U2F JavaScript API [[FIDOU2FJavaScriptAPI]].
+
+During a transition from legacy FIDO JavaScript APIs, a [=[RP]=] may have a population of users with legacy credentials already registered. The [appid](#sctn-appid-extension) extension allows the sign-in flow to be transitioned smoothly but, when transitioning the registration flow, the [excludeCredentials](#dom-publickeycredentialcreationoptions-excludecredentials) field will not be effective in excluding authenticators with legacy credentials because its contents are taken to be WebAuthn credentials. This extension directs [=client platforms=] to consider the contents of [excludeCredentials](#dom-publickeycredentialcreationoptions-excludecredentials) as both WebAuthn and legacy FIDO credentials.
```
```suggestion
During a transition from the FIDO U2F JavaScript API, a [=[RP]=] may have a population of users with legacy credentials already registered. The [appid](#sctn-appid-extension) extension allows the sign-in flow to be transitioned smoothly but, when transitioning the registration flow, the [excludeCredentials](#dom-publickeycredentialcreationoptions-excludecredentials) field will not be effective in excluding authenticators with legacy credentials because its contents are taken to be WebAuthn credentials. This extension directs [=client platforms=] to consider the contents of [excludeCredentials](#dom-publickeycredentialcreationoptions-excludecredentials) as both WebAuthn and legacy FIDO credentials.
```

index.bs
```diff
@@ -4990,6 +4990,85 @@ Instead, in step three, the comparison on the host is relaxed to accept hosts on
...
+      USVString appidExclude;
+    };
+    </xmp>
+
+: Client extension processing
+:: When [[#sctn-createCredential|creating a new credential]]:
+        1. Just after [establishing the RP ID](#CreateCred-DetermineRpId) perform these steps:
+            1. Let |facetId| be the result of passing the caller's [=origin=] to the FIDO algorithm
+                for [=determining the FacetID of a calling application=].
+            1. Let |appId| be the value of the extension input {{appidExclude}}
```
```suggestion
            1. Let |appId| be the value of the extension input {{appidExclude}}.
```

index.bs
```diff
@@ -4990,6 +4990,85 @@ Instead, in step three, the comparison on the host is relaxed to accept hosts on
...
+                        ::  32 random bytes
+
+                        :   <var ignore>application parameter</var>
+                        ::  SHA-256 hash of |appId|
+
+                        :   <var ignore>key handle length</var>
+                        ::  Length of <code>|C|.{{PublicKeyCredentialDescriptor/id}}</code> (in bytes)
+
+                        :   <var ignore>key handle</var>
+                        ::  <code>|C|.{{PublicKeyCredentialDescriptor/id}}</code>'s value, i.e., the [=credential id=].
```
```suggestion
                        ::  The value of <code>|C|.{{PublicKeyCredentialDescriptor/id}}</code>, i.e., the [=credential id=].
```

index.bs
```diff
@@ -4990,6 +4990,85 @@ Instead, in step three, the comparison on the host is relaxed to accept hosts on
...
+                        ::  `0x07` ("check-only")
+
+                        :   <var ignore>challenge parameter</var>
+                        ::  32 random bytes
+
+                        :   <var ignore>application parameter</var>
+                        ::  SHA-256 hash of |appId|
+
+                        :   <var ignore>key handle length</var>
+                        ::  Length of <code>|C|.{{PublicKeyCredentialDescriptor/id}}</code> (in bytes)
```
```suggestion
                        ::  The length of <code>|C|.{{PublicKeyCredentialDescriptor/id}}</code> (in bytes)
```

emlun:
Perhaps we should define a proper term for "legacy credential", and use that in both this and the appid extension? I'm thinking something like:

- **AppID Credential:** A credential scoped to a FIDO AppID, unlike an _RP ID credential_ which is scoped to an RP ID. AppID credentials can be created via the FIDO U2F JavaScript API and cannot be created via the WebAuthn API, but the `appid` extension enables using them to generate WebAuthn assertion signatures.
- **RP ID Credential:** A credential scoped to an RP ID. This is the default kind of credential in WebAuthn, and is the only kind of credential that can be created via the WebAuthn API. See also _AppID credential_.

equalsJeffH:
@emlun suggested "AppID Credential" and "RP ID Credential" terms....

hm... well, I spose I'm thinking that we already have a formal "public key credential" name as well as various informal monikers in use: "webauthn credential", "fido2 credential", "U2F credential"...

Since "legacy" is only used, in terms of U2F credentias and JS API, in the two extension definitions (AppId, appidExclude), I'm thinking things are defined clearly enough at this point and we ought not grant legacy U2F creds any further formal status in the context of the webauthn spec.

emlun:
@equalsJeffH Alright, sounds good.

jcjones:
As an extension, this is going to be best-effort.  With no output, it's not going to be clear to the RP whether it was honored or not -- is that acceptable to RPs?

index.bs
```diff
@@ -4990,6 +4990,85 @@ Instead, in step three, the comparison on the host is relaxed to accept hosts on
...
+                    could involve requesting [=user consent=] from |authenticator| and, upon receipt, treating
+                    it as if the authenticator had returned {{InvalidStateError}}. Requesting [=user consent=]
+                    can be accomplished by sending another `U2F_AUTHENTICATE` message to |authenticator| as
+                    above except for setting |control byte| to `0x03` ("enforce-user-presence-and-sign"),
+                    and ignoring the response.
+
+            1. Continue with normal processing.
+
+: Client extension output
+:: None.
```
emlun: @jcjones [Raises a good point](https://github.com/w3c/webauthn/pull/1244#issuecomment-508227696) - this should probably be a constant `true` output, like the `appid` extension had previously (and like `authnSel` and `biometricPerfBounds` have currently).

agl:
> As an extension, this is going to be best-effort. With no output, it's not going to be clear to the RP whether it was honored or not -- is that acceptable to RPs?

I think the output of `appid` was a mistake and so I've got this extension without an output so far. @emlun suggests a constant `true` output, which isn't too expensive, but I still think it's superfluous.

As an RP transitioning to WebAuthn, if they don't care about duplicate registrations on a single authenticator then they're already happy, so let's assume that they do care. They'll include the current credential IDs in the exclude list and set this extension. Then, if there's an output from this extension, they get to find out, after the registration, whether this extension was processed or not. But what are they going to do with that information? Reject the registration because it might be duplicative and tell users to update their browser? That doesn't seem like something we want to work to support. Worse yet they might try some warning: &ldquo;due to browser limitations we can't ensure that you didn't just register the same token again because it might duplicate some registration that you made before we switched to WebAuthn, so hopefully you didn't do that, right?&rdquo;.

So I can only see bad things happening with this information, so why waste time plumbing it in?

Even if we do believe that either of the above are sensible, surely the RP would want to know about support _before_ the registration operation? That would suggest putting it in #1219 .

nadalin:
@selfissued Please review

agl:
(J.C. suggests adding a string for this in the capability detection PR.)

index.bs
emlun: @agl Makes a good [counter-argument](https://github.com/w3c/webauthn/pull/1244#issuecomment-509316913) to this.

PR #1250, opened 2019-07-01
Add privacy considerations about credential IDs

### LINKED ISSUE #1246: Privacy risk from revealing allowed credentials
A relying party wanting to use WebAuthn Authenticators (including those with non-resident keys) as a 1st factor needs to prompt for a username (eg email address) then reveal that user's credential ids by listing them in the allowedCredentials argument to the navigator.credentials.get() call. Hence, a user's credential ids are revealed to anyone who enters their username.

This is not a security problem (revealing a credential id does not allow anyone to authenticate with that credential). But it might be a slight privacy issue: it reveals how many authenticators a user has registered with this RP; and the length of each credential id gives a hint about its type. While the credential ids for the same user at another RP will have different values, it is likely that there will be the same number and they will have the same length.

[§13.10 Username Enumeration](https://w3c.github.io/webauthn/#sctn-username-enumeration) already suggests an RP return a fake credential id when it has no account for a given username (see [issue 1014](https://github.com/w3c/webauthn/issues/1014)). But RPs may not do that as many don't bother preventing username enumeration.

An RP could include fake credential ids alongside the real credential ids in the allowedCredentials argument (eg so everyone appears to have, say, 10 authenticators). That should be able to address most of the privacy risk (if the fakes are carefully chosen).

Is this worth mentioning in the Privacy Considerations section?
Could extra fake credentials harm the WebAuthn experience (either affecting the user interface, slowing performance, or causing errors)?

### LINKED ISSUE #1311: Privacy leak via credential IDs also applies to resident credentials
> This privacy consideration applies to Relying Parties supporting single-factor authentication with non-resident credentials.

This should apply not only to non-resident credentials but also resident credentials if credentialIds are sent. Sending credentialIds in an allowCredentials is a legitimate flow for authentication. Thus, I suggest to change "with non-resident credentials" to something like "when sending an allowCredentials with credentialIds."

emlun:
Fixes #1246. Fixes #1311.

Things to consider:

- Does this belong in the spec or in separate explainer documentation?
- Is this unnecessarily verbose?
- This is all conjecture as I'm not aware of any quantification of the severity of this kind of information leak, nor of how effective the suggested mitigation would be. Should we refrain from putting that in?

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/pull/1250.html" title="Last updated on Oct 8, 2019, 10:53 AM UTC (e7e11c4)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/1250/4b36388...e7e11c4.html" title="Last updated on Oct 8, 2019, 10:53 AM UTC (e7e11c4)">Diff</a>


index.bs
```diff
@@ -6003,6 +6007,36 @@ leakage due to such an attack:
...
+if the user can initiate an [=authentication ceremony=] by only providing a username.
+[=Credential IDs=] are designed to not be correlatable between [=[RPS]=],
+but the length of a [=credential ID=] might be a hint as to what type of [=authenticator=] created it.
+It is likely that a user will use the same username and set of [=authenticators=] for several [=[RPS]=],
+so the number of [=credential IDs=] in {{PublicKeyCredentialRequestOptions/allowCredentials}} and their lengths
+might serve as a global correlation handle to de-anonymize the user.
+
+The simplest ways to prevent this information leak
+is to not support [=single-factor=] authentication with [=non-resident credentials=],
+for example by:
```
manger: This suggestion -- while simple to describe -- harms interop across the ecosystem. Some authenticators don't work at some sites. That is a poor trade-off for many.

index.bs
```diff
@@ -6003,6 +6007,36 @@ leakage due to such an attack:
...
+
+This privacy consideration applies to [=[RPS]=] supporting [=single-factor=] authentication with [=non-resident credentials=].
+
+In this case the {{PublicKeyCredentialRequestOptions/allowCredentials}} argument risks leaking [PII],
+if the user can initiate an [=authentication ceremony=] by only providing a username.
+[=Credential IDs=] are designed to not be correlatable between [=[RPS]=],
+but the length of a [=credential ID=] might be a hint as to what type of [=authenticator=] created it.
+It is likely that a user will use the same username and set of [=authenticators=] for several [=[RPS]=],
+so the number of [=credential IDs=] in {{PublicKeyCredentialRequestOptions/allowCredentials}} and their lengths
+might serve as a global correlation handle to de-anonymize the user.
```
manger: Perhaps we should recommend a 1 or 2 lengths for credential ids. That would reduce the slight leak.

index.bs
emlun: This is not about blocking some authenticators, but recommending against some use cases. Non-resident authenticators are perfectly fine to use for second-factor authentication.

index.bs
emlun: Even if we were to recommend that, I don't think it would have much effect in practice. There are already lots of authenticators in the field that cannot change, so we're still going to have to deal with different credential ID lengths for the foreseeable future.

index.bs
```diff
@@ -5997,6 +6001,41 @@ leakage due to such an attack:
...
+
+The simplest ways to prevent this information leak
+is to not support [=single-factor=] authentication with [=non-resident credentials=],
```
equalsJeffH: I'm thinking that @manger [has a point](https://github.com/w3c/webauthn/pull/1250/files#r299430803) in the sense that we ought to distinguish between "sign-in"&mdash;when there is no "ambient" notion of who the user is (eg: cookies&mdash;and "re-authentication", when there _is_ an "ambient" notion of who the user is.

It is the "sign-in" case where a malicious client might be able to contact the RP server and request sign-in and receive `credentialIds` in the `allowList`, yes?  In contrast, in the "re-authn" case, a malicious client will have to have captured whatever "ambient creds" (typically cookies) that the RP employs and replay them, yes?

If the above is nominally correct, then we ought to distinguish them.   This in general harkens back to the sorts of subtleties  issue #334 is arguing we ought to clarify in the spec.

index.bs
manger: We should not recommend against this uses case: accepting authenticators with non-resident credentials as 1st factors.
Using a phone or laptop with a builtin platform authenticator, plus having a Security Key for bootstrap & backup, is a great combination that we should encourage.

emlun: Well, you can't have it both ways. Either you're leaking information via `allowCredentials`, or you don't support exposing `allowCredentials` without some form of preceding authentication step (password or cookie, for example).

index.bs
```diff
@@ -5997,6 +6001,41 @@ leakage due to such an attack:
     - When verifying an {{AuthenticatorAssertionResponse}} response from the [=authenticator=], make it indistinguishable whether
           verification failed because the signature is invalid or because no such user or credential is registered.
 
+    - Perform a different authentication step, such as username and password authentication,
```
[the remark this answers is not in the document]
emlun: Yes, that's a good suggestion, thanks!

index.bs
emlun: I added mentions of session cookies as a first authentication step, what do you think?

emlun:
@manger Thanks for the suggestion, I'll see if I can incorporate some of that.

emlun:
@agl Thanks, I added a mention of this.

@manger Going through your proposal again it seems like the main points are already addressed by the text as currently written, so I ended up not making any changes to it.

With that, unless anyone wants to re-review the addition of [@agl's comment](https://github.com/w3c/webauthn/pull/1250#pullrequestreview-276192920), I think we're good to merge as soon as @equalsJeffH finishes his review.

index.bs
manger: "single-factor" is wrong. The info leak also occurs for a Security Key that does multi-factor auth with a non-resident key.

emlun: Hm, you're right. Good catch, thanks!

emlun: Addressed in bba65aa

maxhata:
> In this case the {{PublicKeyCredentialRequestOptions/allowCredentials}} argument risks leaking [PII],
> if the user can initiate an [=authentication ceremony=] by only providing a username.

"by only providing a username" may sound like it excludes the case where 
a username is derived from "ambient credentials" such as cookies.
To eliminate this concern, how about removing "providing" or something else?

emlun:
> "by only providing a username" may sound like it excludes the case where
> a username is derived from "ambient credentials" such as cookies.

It does exclude that case - unless the "ambient credential" is just an unauthenticated username with no session key, which wouldn't really make much sense.

maxhata:
There are many other ways to use "ambien credentials". For example, you could store a hash of credentialIDs instead of username. All these cases should be part of this privacy concern since no matter what you do, you end up sending credentialIDs for a get().

emlun:
@maxhata I don't think I quite understand what you mean, but I changed the quoted sentence to

>In this case the `allowCredentials` argument risks leaking personally identifying information, since it exposes the user’s credential IDs to an unauthenticated caller.

This is both more general by not emphasising a particular interaction flow, and more precise in what the problem is, and I think it should also address your concern, right?

maxhata:
This is how it started the discussion.

index.bs
equalsJeffH: LGTM, thx :)

emlun:
Hm, looks I forgot that I branched PR #1270 off from this, and since that is now merged the diff here looks a bit strange. Sorry about the mess...

equalsJeffH:
are we waiting on merging this for @manger to approve?

emlun:
There's also a review requested from @akshayku.

akshayku:
I need more time to see this. I started looking what we have specified regarding Credential ID privacy and there are more points to review related to username enumeration. 

I will finish this by TPAC.

index.bs
```diff
@@ -6152,37 +6152,40 @@ leakage due to such an attack:
     - When verifying an {{AuthenticatorAssertionResponse}} response from the [=authenticator=], make it indistinguishable whether
           verification failed because the signature is invalid or because no such user or credential is registered.
 
-    - Perform a different authentication step, such as username and password authentication,
-        before initiating the WebAuthn [=authentication ceremony=].
-        This moves the username enumeration problem from the WebAuthn [=authentication ceremony=]
+    - Perform a multi-step [=authentication ceremony=], e.g., beginning with supplying username and password or a session cookie,
```
akshayku: Why? Actually, why not we combine this whole argument to, During authentication, whether or not account exists or not, RP should try to get the assertion from the signature.

akshayku: Also, I am not sure about username recommendation we give in this section which says for one time OTP and other things.. Probably a good idea to review username enumeration issue in another PR??

index.bs
```diff
@@ -6152,37 +6152,40 @@ leakage due to such an attack:
...
 It is likely that a user will use the same username and set of [=authenticators=] for several [=[RPS]=],
 so the number of [=credential IDs=] in {{PublicKeyCredentialRequestOptions/allowCredentials}} and their lengths
 might serve as a global correlation handle to de-anonymize the user.
+Knowing a user's [=credential IDs=] also makes it possible to confirm guesses about the user's identity
+given only momentary physical access to one of the user's [=authenticators=].
 
-The simplest ways to prevent this information leak
-is to not support [=single-factor=] authentication with [=non-resident credentials=],
-for example by:
+The simplest way to prevent this information leak
```
akshayku: No.. With non-resident credentials also you should be able to do single-factor/2FA authentication.. 

How about we collapse this to, for authentication ceremory, add random credentials with random length to the list??

akshayku: Actually, we should talk in terms of second factor vs first factor.. and not in terms of whether credential is resident or not..

emlun: >No.. With non-resident credentials also you should be able to do single-factor/2FA authentication..
>How about we collapse this to, for authentication ceremory, add random credentials with random length to the list??

The last paragraph of the section is about this, is it enough?

>Actually, we should talk in terms of second factor vs first factor.. and not in terms of whether credential is resident or not..

I don't think so - these are privacy considerations about exposing an `allowCredentials` list to an unauthenticated user, so it's precisely about non-resident keys whether they are MFA (UV) credentials or not.

index.bs
emlun: >During authentication, whether or not account exists or not, RP should try to get the assertion from the signature.

I don't understand what you mean, could you elaborate on this?

>Also, I am not sure about username recommendation we give in this section which says for one time OTP and other things.. Probably a good idea to review username enumeration issue in another PR??

Yeah, I'm also a bit ambivalent on it. You're welcome to open a new issue with any comments you have.

index.bs
maxhata: > Actually, we should talk in terms of second factor vs first factor.. and not in terms of whether credential is resident or not..

I agree with this idea.

The issue is about whether or not you need to send credentialIDs with allowCredentials or not. 
By sending credentialIDs, (thereby exposing credentialIDs), regardless of resident credentials or non-resident credentials, we can have the "with usename" flows that have a FIDO credential and typically User Verification-- 2 factors.

emlun: Ok, on second thought you are both right that this would also apply to resident credentials used in second-factor mode. I'll see if I can work that in...

emlun: I made an attempt at this in e7e11c4, please have another look and let me know what you think!

index.bs
akshayku: > I don't understand what you mean, could you elaborate on this?

Actually, what I meant was already present in first option in authentication ceremonies.

akshayku: > Yeah, I'm also a bit ambivalent on it. You're welcome to open a new issue with any comments you have.

I will look and open up clarification on the second point later. Thanks.

PR #1256, opened 2019-07-10
Add a WebDriver Virtual Authenticator extension

### LINKED ISSUE #1236: Create a WebDriver Virtual Authenticator extension to allow automated testing
Currently there is no way for web developers to write automated tests that exercise the WebAuthn API. This problem is echoed on the [WebAuthn WPT test suite](https://github.com/web-platform-tests/wpt/tree/master/webauthn) - many of the tests assume mock authenticators that are not set up by any platform and so [fail outright on all of them](https://wpt.fyi/results/webauthn?label=master&product=chrome%5Bexperimental%5D&product=edge&product=firefox%5Bexperimental%5D&product=safari%5Bexperimental%5D&aligned&q=webauthn).

I propose creating a [WebDriver extension](https://w3c.github.io/webdriver/#dfn-extension-command) defining an API that lets clients create and manipulate virtual authenticators that act transparently as any other authenticator on calls to the WebAuthn API. This would allow web developers to write tests and it would also allow rewriting the WPTs to use this API through [testdriver.js](https://web-platform-tests.org/writing-tests/testdriver-tutorial.html).

The API at minimum should have commands to create and remove virtual authenticators, and to inject, inspect and remove credentials associated to them.

This would be similar to how the [Generic Sensor API is automated](https://www.w3.org/TR/generic-sensor/#automation).

nsatragno:
This change introduces a [WebDriver Extension](https://w3c.github.io/webdriver/#dfn-extension-command) that lets clients create and manipulate virtual authenticators. This API can be used by web developers to write tests against the WebAuthn API and it's also possible to leverage it on [WPTs](https://github.com/web-platform-tests/wpt/tree/master/webauthn) through [testdriver.js](https://web-platform-tests.org/writing-tests/testdriver-tutorial.html). Being able to instantiate virtual authenticators on WPTs would be a major step towards enabling testing a larger surface area of the API across multiple browsers.

Fixes #1236

***
<a href="https://pr-preview.s3.amazonaws.com/nsatragno/webauthn/pull/1256.html" title="Last updated on Jul 30, 2019, 6:30 PM UTC (9be5c17)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/1256/126223e...nsatragno:9be5c17.html" title="Last updated on Jul 30, 2019, 6:30 PM UTC (9be5c17)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -5499,19 +5499,19 @@ At this time, one [=credential property=] is defined: the [=resident key credent
 
 # Automation # {#sctn-automation}
 
-For the purposes of user-agent automation and application testing, this document defines a number of <a>extension
-commands</a> for the [[WebDriver]] specification. 
+For the purposes of user-agent automation and application testing, this document defines a number of [=extension
+commands=] for the [[WebDriver]] specification. 
 
 ## <dfn>Virtual Authenticators</dfn> ## {#sctn-automation-virtual-authenticators}
 
-These <a>extension commands</a> create and interact with <a>Virtual Authenticators</a>: software implementations of the
+These [=extension commands=] create and interact with [=Virtual Authenticators=]: software implementations of the
 <a>Authenticator Model</a>. <a>Virtual Authenticators</a> are stored in a <dfn>Virtual Authenticator Database</dfn> and
 have the following properties:
 
 : |authenticatorId|
-:: An arbitrary non-null string that uniquely identifies the <a>Virtual Authenticator</a>.
+:: An arbitrary non-null string that uniquely identifies the [=Virtual Authenticator=].
 : |protocol|
-:: The protocol the <a>Virtual Authenticator</a> uses to encode its messages. Either `ctap2` for the [[FIDO-CTAP]] format
+:: The protocol the [=Virtual Authenticator=] uses to encode its messages. Either `ctap2` for the [[FIDO-CTAP]] format
     or `u2f` for the [[FIDO-U2F-Message-Formats]].
 : |transport|
 :: The {{AuthenticatorTransport}} simulated. If the |transport| is set to {{AuthenticatorTransport/internal}}, the
@@ -5521,14 +5521,14 @@ have the following properties:
 : |hasUserVerification|
 :: If set to `true`, the authenticator will support [=user verification=].
 : |isUserVerified|
-:: Determines the result of <a>User Verification</a> performed on the <a>Virtual Authenticator</a>. If set to `true`,
-    <a>User Verification</a> will always succeed. If set to `false`, it will fail. This property has no effect if
+:: Determines the result of [=User Verification=] performed on the [=Virtual Authenticator=]. If set to `true`,
+    [=User Verification=] will always succeed. If set to `false`, it will fail. This property has no effect if
     |hasUserVerification| is set to `false`.
 
 ## <dfn>Add Virtual Authenticator</dfn> ## {#sctn-automation-add-virtual-authenticator}
 
-The <a>Add Virtual Authenticator</a> <a>extension command</a> creates a software <a>Virtual Authenticator</a>. The
-<a>extension command</a> is defined as follows:
+The [=Add Virtual Authenticator=] [=extension command=] creates a software [=Virtual Authenticator=]. The
+[=extension command=] is defined as follows:
 
     <xmp class="idl">
     enum AuthenticatorProtocol {
@@ -5562,26 +5562,26 @@ The <a>Add Virtual Authenticator</a> <a>extension command</a> creates a software
     </table>
 </figure>
 
-The <a>remote end steps</a> are:
+The [=remote end steps=] are:
 
- 1. If |parameters| is not a JSON <a>Object</a>, return a <a>WebDriver error</a> with <a>WebDriver error code</a>
-     <a>invalid argument</a>.
- 1. Let |protocol| be the result of <a>trying</a> to get the |parameter|'s {{AddVirtualAuthenticatorParameters/protocol}}
+ 1. If |parameters| is not a JSON [=Object=], return a [=WebDriver error=] with [=WebDriver error code=]
+     [=invalid argument=].
+ 1. Let |protocol| be the result of [=trying=] to get the |parameter|'s {{AddVirtualAuthenticatorParameters/protocol}}
     property.
- 1. If |protocol| is null, return a <a>WebDriver error</a> with <a>WebDriver error code</a> <a>invalid argument</a>.
- 1. Let |transport| be the result of <a>trying</a> to get the |parameter|'s {{AddVirtualAuthenticatorParameters/transport}}
+ 1. If |protocol| is null, return a [=WebDriver error=] with [=WebDriver error code=] [=invalid argument=].
+ 1. Let |transport| be the result of [=trying=] to get the |parameter|'s {{AddVirtualAuthenticatorParameters/transport}}
     property.
- 1. If |transport| is null, return a <a>WebDriver error</a> with <a>WebDriver error code</a> <a>invalid argument</a>.
+ 1. If |transport| is null, return a [=WebDriver error=] with [=WebDriver error code=] [=invalid argument=].
  1. Generate an arbitrary and unique string |authenticatorId|.
- 1. Create a new <a>Virtual Authenticator</a> with the fields |authenticatorId|, |protocol|, |transport|,
+ 1. Create a new [=Virtual Authenticator=] with the fields |authenticatorId|, |protocol|, |transport|,
     |hasResidentKey|, |hasUserVerification|, and |isUserVerified|.
- 1. Store the authenticator in the <a>Virtual Authenticator Database</a>.
- 1. Return <a>success</a> with |authenticatorId|.
+ 1. Store the authenticator in the [=Virtual Authenticator Database=].
+ 1. Return [=success=] with |authenticatorId|.
 
 ## <dfn>Remove Virtual Authenticator</dfn> ## {#sctn-automation-remove-virtual-authenticator}
 
-The <a>Remove Virtual Authenticator</a> <a>extension command</a> removes a previously created <a>Virtual Authenticator</a>.
-The <a>extension command</a> is defined as follows:
+The [=Remove Virtual Authenticator=] [=extension command=] removes a previously created [=Virtual Authenticator=].
+The [=extension command=] is defined as follows:
 
 <figure id="table-removeVirtualAuthenticator" class="table">
     <table class="data">
@@ -5600,17 +5600,17 @@ The <a>extension command</a> is defined as follows:
     </table>
 </figure>
 
-The <a>remote end steps</a> are:
+The [=remote end steps=] are:
 
- 1. If |authenticatorId| does not match any <a>Virtual Authenticator</a> stored in the <a>Virtual Authenticator
-     Database</a>, return a <a>WebDriver error</a> with <a>WebDriver error code</a> <a>invalid argument</a>.
- 1. Remove the <a>Virtual Authenticator</a> identified by |authenticatorId| from the <a>Virtual Authenticator Database</a>
- 1. Return <a>success</a>.
+ 1. If |authenticatorId| does not match any [=Virtual Authenticator=] stored in the [=Virtual Authenticator
+     Database=], return a [=WebDriver error=] with [=WebDriver error code=] [=invalid argument=].
+ 1. Remove the [=Virtual Authenticator=] identified by |authenticatorId| from the [=Virtual Authenticator Database=]
+ 1. Return [=success=].
 
 ## <dfn>Add Credential</dfn> ## {#sctn-automation-add-credential}
 
-The <a>Add Credential</a> <a>extension command</a> injects a <a>Credential Key Pair</a> and associated <a>Authenticator Data</a>
-into an existing <a>Virtual Authenticator</a>. The <a>extension command</a> is defined as follows:
+The [=Add Credential=] [=extension command=] injects a [=Credential Key Pair=] and associated [=Authenticator Data=]
+into an existing [=Virtual Authenticator=]. The [=extension command=] is defined as follows:
 
     <xmp class="idl">
     dictionary CredentialParams {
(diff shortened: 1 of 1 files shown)


jcjones:
I'm going to need to get more expert eyes than my own on this.

My initial question is: Do we need to define what these operations _do_? For example, if one calls `Add Virtual Authenticator`, I presume the user agent would have an authenticator now that could respond to a makeCredential. But will it? Does it need to be selective in some way about its virtual attachment?

What happens if I do `Add Virtual Authenticator` twice? 

I'm guessing that someone would implement this as a soft-HID device, maybe, that listened for these commands? Or does the user agent need to do that?

nsatragno:
> I presume the user agent would have an authenticator now that could respond to a makeCredential. But will it? Does it need to be selective in some way about its virtual attachment?

It's possible to control if the authenticator responds to a makeCredential request with `isUserPresent` - the authenticator "simulates" a test of user presence and that flag indicates whether it passes (say, pretending a user tapped a security key) or fails.

There is another layer to selectivity which is WebDriver constrains the actions to the current [session](https://w3c.github.io/webdriver/#sessions), a particular instantiation of a user agent.

> What happens if I do Add Virtual Authenticator twice?

Presumably the same thing that would happen if you had two real authenticators plugged in your machine. If both match the `authenticatorSelectionCriteria` and both have the `isUserPresent` bit set to `true`, it would probably be acceptable to have a race between them. We could also specify more strict behaviour (e.g. the last authenticator added resolves the request). I avoided doing this because in general I don't think it's a useful case to test.

> I'm guessing that someone would implement this as a soft-HID device, maybe, that listened for these commands? Or does the user agent need to do that?

WebDriver extensions are implemented by user agents, so the user agent has to provide this virtual authenticator. This is similar to how the [Generic Sensor API](https://www.w3.org/TR/generic-sensor/#automation) provides automation capabilities for testing.

index.bs
```diff
@@ -5483,6 +5500,377 @@ At this time, one [=credential property=] is defined: the [=resident key credent
...
+: |transport|
+:: The {{AuthenticatorTransport}} simulated. If the |transport| is set to {{AuthenticatorTransport/internal}}, the
+    authenticator simulates [=platform attachment=]. Otherwise, it simulates [=cross-platform attachment=].
+: |hasResidentKey|
+:: If set to [TRUE] the authenticator will support [=resident credentials=].
+: |hasUserVerification|
+:: If set to [TRUE], the authenticator supports [=user verification=].
+: |isUserPresent|
+:: Determines the result of a [=test of user presence=] performed on the [=Virtual Authenticator=]. If set to [TRUE],
+    a [=test of user presence=] will always succeed. If set to [FALSE], it will fail.
```
emlun: I assume this also defines the outcome of any other kind of request for user consent? Should that then be stated explicitly?

index.bs
```diff
@@ -5483,6 +5500,377 @@ At this time, one [=credential property=] is defined: the [=resident key credent
...
+     [=invalid argument=].
+     Note: |parameters| is a [=Credential Parameters=] object.
+ 1. Let |credentialId| be the result of decoding [=Base64url Encoding=] on the |parameters|' |credentialId| property.
+ 1. If |credentialId| is failure, return a [=WebDriver error=] with [=WebDriver error code=] [=invalid argument=].
+ 1. Let |rpId| be the the |parameters|' |rpId| property.
+ 1. If |rpId| is not a valid [=RP ID=], return a [=WebDriver error=] with [=WebDriver error code=] [=invalid argument=].
+ 1. Let |privateKey| be the result of decoding [=Base64url Encoding=] on the |parameters|' |privateKey| property.
+ 1. If |privateKey| is failure, return a [=WebDriver error=] with [=WebDriver error code=] [=invalid argument=].
+ 1. If |privateKey| is not a validly-encoded asymmetric key package containing a single private key per [[RFC5958]],
+     return a [=WebDriver error=] with [=WebDriver error code=] [=invalid argument=].
```
emlun: Since we're using COSE key encoding in the main API, thoughts on using that here as well?

index.bs
```diff
@@ -5483,6 +5500,377 @@ At this time, one [=credential property=] is defined: the [=resident key credent
...
+    : [=public key credential source/type=]
+    :: {{public-key}}
+    : [=public key credential source/id=]
+    :: |credentialId|
+    : [=public key credential source/privateKey=]
+    :: |privateKey|
+    : [=public key credential source/rpId=]
+    :: |rpId|
+    : [=public key credential source/userHandle=]
+    :: `null`
```
emlun: Why is the userHandle always null? Won't that make it harder to test allowCredentials-less authentication scenarios?

index.bs
```diff
@@ -5483,6 +5500,377 @@ At this time, one [=credential property=] is defined: the [=resident key credent
...
+ 1. If |authenticatorId| does not match any [=Virtual Authenticator=] stored in the [=Virtual Authenticator
+     Database=], return a [=WebDriver error=] with [=WebDriver error code=] [=invalid argument=].
+ 1. Let |credentialsArray| be an empty array.
+ 1. For each [=Public Key Credential Source=] |credential|, managed by the authenticator identified by |authenticatorId|,
+    construct a corresponding [=Credential Parameters=] [=Object=] and add it to |credentialsArray|.
+ 1. Return [=success=] with data containing |credentialsArray|.
+
+## <dfn>Remove Credential</dfn> ## {#sctn-automation-remove-credential}
+
+The [=Remove Credential=] WebDriver [=extension command=] removes a [=Public Key Credential Sources=] stored on a
```
```suggestion
The [=Remove Credential=] WebDriver [=extension command=] removes a [=Public Key Credential Source=] stored on a
```

index.bs
```diff
@@ -5483,6 +5500,377 @@ At this time, one [=credential property=] is defined: the [=resident key credent
...
+                <td>POST</td>
+                <td>`/session/{session id}/webauthn/authenticator/{authenticatorId}/uv`</td>
+            </tr>
+        </tbody>
+    </table>
+</figure>
+
+The [=remote end steps=] are:
+
+ 1. If |parameters| is not a JSON [=Object=], return a [=WebDriver error=] with [=WebDriver error code=]
```
emlun: It's obvious what this means, but _parameters_ isn't explicitly defined here, unlike in the other operations.

index.bs
```diff
@@ -5483,6 +5500,377 @@ At this time, one [=credential property=] is defined: the [=resident key credent
...
+
+ 1. If |authenticatorId| does not match any [=Virtual Authenticator=] stored in the [=Virtual Authenticator
+     Database=], return a [=WebDriver error=] with [=WebDriver error code=] [=invalid argument=].
+ 1. Remove all [=Public Key Credential Sources=] managed by the [=Virtual Authenticator=] identified by |authenticatorId|.
+ 1. Return [=success=].
+
+## <dfn>Set User Verified</dfn> ## {#sctn-automation-set-user-verified}
+
+The [=Set User Verified=] [=extension command=] sets the |isUserVerified| property on the [=Virtual Authenticator=]. It
+is defined as follows:
```
emlun: Would it make sense to also have a similar command for `isUserPresent`? Maybe combined into a single command like `/session/{session id}/webauthn/authenticator/{authenticatorId}/state` with both `isUserPresent` and `isUserVerified` parameters?

index.bs
```diff
@@ -5483,6 +5500,377 @@ At this time, one [=credential property=] is defined: the [=resident key credent
...
+                <td>POST</td>
+                <td>`/session/{session id}/webauthn/authenticator/{authenticatorId}/credential`</td>
+            </tr>
+        </tbody>
+    </table>
+</figure>
+
+The <dfn>Credential Parameters</dfn> is a JSON [=Object=] passed to the [=remote end steps=] as |parameters|. It contains the following |key| and |value| pairs:
+
+<figure id="table-credentialParameters" class="table">
```
emlun: Since this table is referenced in more than one operation, I think it would be useful to move/copy some of the content descriptions/requirements (base64, key encoding etc) from the below algorithm into a third column in the table, no?

index.bs
equalsJeffH: Well, hm, a subtle p[oint there -- it seems from the "Flags" table a the very end of section [5.2. authenticatorGetAssertion (0x02)](https://drafts.fidoalliance.org/fido-2/stable-links-to-latest/fido-client-to-authenticator-protocol.html#authenticatorGetAssertion) that it is possible for `uv` to be returned true and `up` false, and the definition of [User Consent](https://www.w3.org/TR/webauthn-2/#user-consent) punts to  [authorization gesture](https://www.w3.org/TR/webauthn-2/#authorization-gesture) which say it could be one or the other.

index.bs
nsatragno: I specifically left `isUserPresent` / `isUserConsenting` as a flag set when the authenticator is created that cannot be changed. Otherwise, a client of the API might expect something like this to work:

```javascript
let authenticatorId = AddVirtualAuthenticator({isUserConsenting: false})
navigator.credentials.create(...).then(resolvePromise);
SetIsUserConsenting({authenticatorId, isConsenting: true});
// promise resolves.
```

Supporting a case like this for CTAP2 software authenticators would add complexity for little value for clients of the API. For a client, an authenticator present that does not collect a gesture indicating user consent or the lack of plugged in authenticators should have the same results.

index.bs
agl: In order to make this as usable as possible, I think a more common private key format would be preferable. Web Crypto, for example, [exports](https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/exportKey) in PKCS#8 and, whilst PKCS#8 is hardly a good format, nor is COSE, and PKCS#8 is the devil that everybody already supports.

nsatragno: Agreed, using something that web developers can get straight out of web crypto is more usable and will make writing web platform tests smoother. I also constrained further the key type to simplify the implementation.

jcjones:
Note: I've asked @AutomatedTester to do a review pass here from the Mozilla side.

index.bs
```diff
@@ -5483,6 +5500,415 @@ At this time, one [=credential property=] is defined: the [=resident key credent
...
+                <td>DELETE</td>
+                <td>`/session/{session id}/webauthn/authenticator/{authenticatorId}/credentials`</td>
+            </tr>
+        </tbody>
+    </table>
+</figure>
+
+The [=remote end steps=] are:
+
+ 1. If |authenticatorId| does not match any [=Virtual Authenticator=] stored in the [=Virtual Authenticator
```
AutomatedTester: Same question as above?

index.bs
```diff
@@ -5483,6 +5500,415 @@ At this time, one [=credential property=] is defined: the [=resident key credent
...
+                <td>DELETE</td>
+                <td>`/session/{session id}/webauthn/authenticator/{authenticatorId}/credentials/{credentialId}`</td>
+            </tr>
+        </tbody>
+    </table>
+</figure>
+
+The [=remote end steps=] are:
+
+ 1. If |authenticatorId| does not match any [=Virtual Authenticator=] stored in the [=Virtual Authenticator
```
AutomatedTester: Do we really need to error here when there are not items that match. Can't we just try do the right thing and if not just go meh?

index.bs
```diff
@@ -5483,6 +5500,415 @@ At this time, one [=credential property=] is defined: the [=resident key credent
...
+    : [=public key credential source/rpId=]
+    :: |rpId|
+    : [=public key credential source/userHandle=]
+    :: |userHandle|
+ 1. Associate a [=signature counter=] |counter| to the |credential| with a starting value equal to the |parameters|'
+     |signCount| or `0` if |signCount| is `null`.
+ 1. Store the |credential| and |counter| in the database of the |authenticator|.
+ 1. Return [=success=].
+
+## <dfn>Get Credentials</dfn> ## {#sctn-automation-get-credentials}
```
AutomatedTester: Is there a use case for this? The webdriver philosophy is to only add a `getter` type command if there is a sufficient need and that tests aren't required to be aware of this info in the first place. This feels like you might be setting it once per test and might never call this API but happy to be corrected

index.bs
```diff
@@ -5483,6 +5500,415 @@ At this time, one [=credential property=] is defined: the [=resident key credent
...
+The [=remote end steps=] are:
+
+ 1. If |authenticatorId| does not match any [=Virtual Authenticator=] stored in the [=Virtual Authenticator
+     Database=], return a [=WebDriver error=] with [=WebDriver error code=] [=invalid argument=].
+ 1. Remove the [=Virtual Authenticator=] identified by |authenticatorId| from the [=Virtual Authenticator Database=]
+ 1. Return [=success=].
+
+## <dfn>Add Credential</dfn> ## {#sctn-automation-add-credential}
+
+The [=Add Credential=] WebDriver [=extension command=] injects a [=Public Key Credential Source=] into an existing
```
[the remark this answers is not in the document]
nsatragno: The general implementation is based on the permissions API, [generic sensor API](https://www.w3.org/TR/generic-sensor/#automation) and [reporting API](https://w3c.github.io/reporting/#automation). The only major stylistic difference is we don't use IDL here to describe the parameters.

index.bs
nsatragno: What do you mean exactly? If the `authenticatorId` doesn't match, just don't do anything and return `success`? If that's the case, that would potentially be a source of confusion for developers that's easily avoided by returning an error... the UA has to check for the existance of the authenticator anyway. Maybe I'm misunderstanding you?

index.bs
nsatragno: A getter for the credentials gives a client of the API more granularity. For example, a relying party might want to run a test where they create a credential via `navigator.credentials.create` and then produce an assertion with `navigator.credentials.get`. Being able to inspect the credential stored in the virtual authenticator by `navigator.credentials.create` lets them verify the creation ran as expected, so if the assertion fails the problem is with the `get` call.

A big motivation for this API is to allow RPs to test their webauthn client code, as well as allow end-to-end tests. We should probably avoid being too restrictive on its design.

index.bs
emlun: Hm, good point. Maybe leaving it explicitly coupled to just `up` is a better way to go after all?

index.bs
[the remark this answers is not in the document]
AutomatedTester: 👍🏽

index.bs
AutomatedTester: Since we are deleting it, if it is not there does it really matter is what I mean. If you feel it does matter then leave it in.

index.bs
AutomatedTester: Since we are deleting it, if it is not there does it really matter is what I mean. If you feel it does matter then leave it in.

index.bs
AutomatedTester: ok.

jcjones:
Just a question @nsatragno: 

In the algorithm for adding a Virtual Authenticator, we emit a WebDriver error when the parameter doesn't have a valid value. If, for example, a browser initially only supported `protocol` being `ctap1/u2f`, would emitting a WebDriver error for all other `protocol`s (e.g. `ctap2`) permit the relevant tests to continue on, or is such an error fatal?

jcjones:
> Just a question @nsatragno:
> 
> In the algorithm for adding a Virtual Authenticator, we emit a WebDriver error when the parameter doesn't have a valid value. If, for example, a browser initially only supported `protocol` being `ctap1/u2f`, would emitting a WebDriver error for all other `protocol`s (e.g. `ctap2`) permit the relevant tests to continue on, or is such an error fatal?

To clarify, I think the WebDriver implementation for WebAuthn could be used then for FIDO U2F's JS API, which would let us write Web Platform Tests for U2F. Which I want to get, if at all possible.

Other than this question, I am ready to grant this review from the Mozilla side.

index.bs
nsatragno: I think it does matter. The most likely scenario for trying to delete something that is not there is a developer oversight. Catching that early means less headaches. And if say a library wanted that functionality, they could catch the error and pretend nothing happened.

jcjones:
> What do you think?

I think I hit 'approve' :)

index.bs
equalsJeffH: This and the above conversation can be set as resolved now?

equalsJeffH:
Hi @nsatragno -- IIUC, your answer to @jcjones in https://github.com/w3c/webauthn/pull/1256#pullrequestreview-284336780 implies that returning a webdriver error does not necessarily halt a testing run, correct?

Also, IIUC, you have not as yet added the step proposed in https://github.com/w3c/webauthn/pull/1256#pullrequestreview-284336780 to the AddVirtualAuthenticator  algo. Is @jcjones' appoval contingent on that being added?

HTH, equalsJeffH

jcjones:
The clarification doesn't sound normative to me, but I'd be happy to wait until it is available before merging rather than worrying about it more in the future.

jabazorius:
help

akshayku:
Lets discuss it next week in TPAC where more Edge folks will be present and can review it more thoroughly.

PR #1259, opened 2019-07-13
make UV requirement default 'discouraged'

### LINKED ISSUE #1253: The user-verification default of “preferred” is catching most sites out.
With the launch of PIN support in Chrome 75, we received a bug report that Chrome was now asking for a PIN when logging into Google. It turns out that our server-side team had missed that the default value for userVerification was “preferred” and weren't setting a value. The user in question was unaware that their authenticator had a PIN set, but it did and thus we were asking for it.

Dropbox and Twitter also have no value set for userVerification and that appears to be the same error. (Indeed, it's unclear to me what site would want the behaviour of “preferred”.)

Given that this is catching everyone out, and that setting the default to “discouraged” is backwards compatible, perhaps we should do that.

equalsJeffH:
fixes #1253 

see also: https://chromium.googlesource.com/chromium/src/+/master/content/browser/webauth/uv_preferred.md

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/pull/1259.html" title="Last updated on Jul 13, 2019, 12:51 AM UTC (5658427)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/1259/d2a4543...5658427.html" title="Last updated on Jul 13, 2019, 12:51 AM UTC (5658427)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2434,7 +2434,7 @@ attributes.
         AuthenticatorAttachment      authenticatorAttachment;
         boolean                      requireResidentKey = false;
         ResidentKeyRequirement       residentKey;
-        UserVerificationRequirement  userVerification = "preferred";
+        UserVerificationRequirement  userVerification = "discouraged";
     };
 </xmp>
 


akshayku:
This is a breaking change which we don't like. I would go for more clearer explanation in the spec as well as in the browser (as chrome is proposing). Saying that people don't read spec and lets change the spec due to this, is not the way to approach this.

PR #1264, opened 2019-07-25
Add lightning transport

### LINKED ISSUE #1261: Add two new transport types to section 5.10.4
In order for RP to know if a credential can be used on a iOS device, we need to add new transport enumerators to Section 5.10.4.

We need one for the physical lightning connector, and one for the HID transport encapsulation.

Currently, we have usb, ble, nfc, and internal.

Strings to add 
lightning
tunnel           (it could be more specific but I don't want to step on any trademarks for the External Accessory Framework.)

Are we missing an entry for caBLE?

This should be relatively uncontroversial, however agreeing on the sting value could take us months:)

ve7jtb:
Add lightning transport

Fixes #1261

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2824,6 +2824,7 @@ parameter to the {{CredentialsContainer/create()}} or {{CredentialsContainer/get
         "nfc",
         "ble",
         "internal"
+        "lightning"
     };
 </xmp>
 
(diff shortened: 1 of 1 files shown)


jcjones:
Per https://heycam.github.io/webidl/#idl-enums , Example 33, providing "lightning" to a browser that doesn't have that part of the enum will throw a TypeError.

Perhaps we want to keep the enum as a definition, but change the structures that specify it to instead ask for a String-ish type that can be carefully cast to an enum, rather than letting the WebIDL parser throw.

PR #1272, opened 2019-08-09
Add note about decoding U2F keyhandles.

agl:
U2F key handles need to be base64url decoded before being used as
ArrayBuffers in WebAuthn calls. This isn't completely obvious so this
change adds a clarification to the appid and appidExclude extensions.

***
<a href="https://pr-preview.s3.amazonaws.com/agl/webauthn/pull/1272.html" title="Last updated on Aug 20, 2019, 10:13 PM UTC (cf05efe)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/1272/4561e48...agl:cf05efe.html" title="Last updated on Aug 20, 2019, 10:13 PM UTC (cf05efe)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -4930,7 +4930,7 @@ in order to allow users to [=authentication|authenticate=] using their registere
      of the {{CredentialsContainer/get()}} method:
 
      - Set the {{PublicKeyCredentialDescriptor/type}} members to {{PublicKeyCredentialType/public-key}}.
-     - Set the {{PublicKeyCredentialDescriptor/id}} members to the respective U2F key handles of the desired credentials.
+     - Set the {{PublicKeyCredentialDescriptor/id}} members to the respective U2F key handles of the desired credentials. Note that U2F key handles are commonly base64url encoded but must be decoded to their binary form when used in {{PublicKeyCredentialDescriptor/id}}.
 
      {{PublicKeyCredentialRequestOptions/allowCredentials}} MAY contain a mixture
      of both WebAuthn [=credential IDs=] and U2F key handles;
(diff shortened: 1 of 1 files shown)


baloo:
thank you!

index.bs
```diff
@@ -4930,7 +4930,7 @@ in order to allow users to [=authentication|authenticate=] using their registere
      of the {{CredentialsContainer/get()}} method:
 
      - Set the {{PublicKeyCredentialDescriptor/type}} members to {{PublicKeyCredentialType/public-key}}.
-     - Set the {{PublicKeyCredentialDescriptor/id}} members to the respective U2F key handles of the desired credentials.
+     - Set the {{PublicKeyCredentialDescriptor/id}} members to the respective U2F key handles of the desired credentials. Note that U2F key handles are commonly base64url encoded but must be decoded to their binary form when used in {{PublicKeyCredentialDescriptor/id}}.
```
baloo:
```suggestion
     - Set the {{PublicKeyCredentialDescriptor/id}} members to the respective U2F key handles of the desired credentials. Note that U2F key handles are commonly websafe-base64 encoded but must be decoded to their binary form when used in {{PublicKeyCredentialDescriptor/id}}.
```

index.bs
```diff
@@ -5011,7 +5011,8 @@ Instead, in step three, the comparison on the host is relaxed to accept hosts on
...
+During a transition from the FIDO U2F JavaScript API, a [=[RP]=] may have a population of users with legacy credentials already registered. The [appid](#sctn-appid-extension) extension allows the sign-in flow to be transitioned smoothly but, when transitioning the registration flow, the [excludeCredentials](#dom-publickeycredentialcreationoptions-excludecredentials) field will not be effective in excluding authenticators with legacy credentials because its contents are taken to be WebAuthn credentials. This extension directs [=client platforms=] to consider the contents of [excludeCredentials](#dom-publickeycredentialcreationoptions-excludecredentials) as both WebAuthn and legacy FIDO credentials. Note that U2F key handles are commonly base64url encoded but must be decoded to their binary form when used in [excludeCredentials](#dom-publickeycredentialcreationoptions-excludecredentials).
```
baloo:
```suggestion
During a transition from the FIDO U2F JavaScript API, a [=[RP]=] may have a population of users with legacy credentials already registered. The [appid](#sctn-appid-extension) extension allows the sign-in flow to be transitioned smoothly but, when transitioning the registration flow, the [excludeCredentials](#dom-publickeycredentialcreationoptions-excludecredentials) field will not be effective in excluding authenticators with legacy credentials because its contents are taken to be WebAuthn credentials. This extension directs [=client platforms=] to consider the contents of [excludeCredentials](#dom-publickeycredentialcreationoptions-excludecredentials) as both WebAuthn and legacy FIDO credentials. Note that U2F key handles are commonly websafe-base64 encoded but must be decoded to their binary form when used in [excludeCredentials](#dom-publickeycredentialcreationoptions-excludecredentials).
```

index.bs
s/ base64url encoded / [=base64url encoding|base64url encoded=] /

note: using just [=base64url encoded=] may do the right thing, would have to try it and see.

index.bs
s/ base64url encoded / [=base64url encoding|base64url encoded=] /

index.bs
peteroupc: > 
> 
> suggest:
> s/ base64url encoded / [=base64url encoding|base64url encoded=] /
> 
> note: using just [=base64url encoded=] may do the right thing, would have to try it and see.

No it wouldn't. "Base64url Encoding" is a specifically-defined term in WebAuthn, and base64url encoding uses padding at the end unless stated otherwise by the specification that uses it (see RFC 4648 section 3.2), for example, in WebAuthn's "Base64url Encoding" definition.

agl: Have used `[=base64url encoding=]`. It links [here](https://w3c.github.io/webauthn/#base64url-encoding) in the WebAuthn spec, which defines a base64 encoding with padding removed, which is correct for U2F.

index.bs

PR #1276, opened 2019-08-14
add indication of cross-origin operation in `collectedClientData`

### LINKED ISSUE #1271: add indication of embedded and cross-origin operation in `collectedClientData`
On the [2019-08-07 webauthn call](https://html.spec.whatwg.org/#child-browsing-context), we discussed whether we ought to capture an indication of whether the browsing context invoking the webauthn API is a [child browsing context](https://html.spec.whatwg.org/#child-browsing-context), i.e., whether it is framed and thus is within a [browsing context container](https://html.spec.whatwg.org/#browsing-context-container) (see [minutes](https://www.w3.org/2019/08/07-webauthn-minutes.html), search for "iframes" down near the bottom). If it is a [child browsing context](https://html.spec.whatwg.org/#child-browsing-context), then whether it is [same-origin with its ancestors](https://w3c.github.io/webappsec-credential-management/#same-origin-with-its-ancestors) is also of interest. 

The latter is already detected within the [credential management framework machinery](https://w3c.github.io/webappsec-credential-management/) and passed into the WebAuthn API's (internal-to-Credential-Management) methods, so at a minimum addressing this issue involves determining whether the current browsing context is framed or not and recording that in the [`collectedClientData`](https://www.w3.org/TR/webauthn-2/#dictdef-collectedclientdata).

equalsJeffH:
fixes #1271 #911

This webauthn PR is associated with w3c/webappsec-credential-management#138. It eliminates the `sameOriginWithAncestors` check from both `[[Create]]()` and `[[DiscoverFrom...]]()` and instead adds the inverse of it's value to `collectedClientData` in the form of the `crossOrigin` boolean.

Cross-origin usage is now gated upon feature policy, per w3c/webappsec-credential-management#138.  Credman+WebAuthn default behavior remains the same as before:  allowed in same-origin contexts, disallowed in cross-origin contexts. Cross-origin usage can now be attained by the RP webdev/author by explicitly setting the `publickey-credentials` feature policy.

see also: https://github.com/w3c/webappsec-credential-management/pull/138#issuecomment-547247234

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/pull/1276.html" title="Last updated on Oct 30, 2019, 6:44 PM UTC (de32639)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/1276/5dbea68...de32639.html" title="Last updated on Oct 30, 2019, 6:44 PM UTC (de32639)">Diff</a>


emlun:
Blocked awaiting work in CredMan

index.bs
```diff
@@ -1324,9 +1324,13 @@ This [=internal method=] accepts three arguments:
...
 
-    :   <dfn>sameOriginWithAncestors</dfn>
-    ::  This argument is a Boolean value which is [TRUE] if and only if the caller's [=environment settings object=] is
-        [=same-origin with its ancestors=].
+    :   <dfn>crossOrigin</dfn>
+    ::  This argument is a Boolean value which is [TRUE] if this [=internal method=] is being
+        invoked from a cross-origin [=browsing context=].
+
+        Note: Invocation of this [=internal method=] indicates that it was allowed by
+        [=feature policy=], which is evauated at the [[!CREDENTIAL-MANAGEMENT-1]] level.
```
```suggestion
        [=feature policy=], which is evaluated at the [[!CREDENTIAL-MANAGEMENT-1]] level.
```

index.bs
```diff
@@ -1341,18 +1345,6 @@ When this method is invoked, the user agent MUST execute the following algorithm
 
 1. Assert: <code>|options|.{{CredentialCreationOptions/publicKey}}</code> is [=present=].
 
-1. If <var ignore>sameOriginWithAncestors</var> is [FALSE], return a "{{NotAllowedError}}" {{DOMException}}.
```
emlun: The equivalent of this check will be moved to CredMan instead, right?

agl:
> This is a breaking change, right?

No? It allows operations that were previously forbidden, I don't think it stops anything that was previously allowed.

ve7jtb:
It shouldn't break anything.  The one exception to that might be some servers that won't understand the new elements in client data.  But only servers explicitly allowing this should ever see that.  
This should not change anything for the normal case.

emlun:
Yeah, ok, it won't break the WebAuthn API, but I was thinking it's a breaking change to the internal browser APIs. But I guess that is already the case for L2 anyway.

equalsJeffH:
I /think/ this is largely ready to re-review in conjunction with https://github.com/w3c/webappsec-credential-management/pull/138 (I'm pretty sure the latter is ready), but it's late and I'll do my fine-grained double-checking tomorrow. Note: I've updated the original post and this PR's title to reflect the current state of things.

index.bs
equalsJeffH: yes. See https://github.com/w3c/webappsec-credential-management/pull/138

equalsJeffH:
this is **ready to re-review** in conjunction with https://github.com/w3c/webappsec-credential-management/pull/138.

Note: I've updated the original post and this PR's title to reflect the current state of things.

nadalin:
@akshayku Please review so we can close this out or move it to next WD

emlun:
Could someone help refresh my memory on this? At some point we discussed making `origin` a list of the origins in the hierarchy instead of just indicating cross-origin with a Boolean flag. What was the rationale for giving the RP only the innermost-embedded origin, and not the whole hierarchy?

Issue #1292, opened 2019-09-04
Supply an “intention" member in PublicKeyCredentialCreationOptions dictionary


alanwaketan:
a) Proposed changes to the spec:

```
dictionary PublicKeyCredentialCreationOptions {
    …
    CredentialCreationIntention intention;
};
```

**intention, of type CredentialCreationIntention**
&nbsp;&nbsp;&nbsp;&nbsp;This OPTIONAL member contains the purpose of the relying party to create the public key credential. This allows the user agent to better describe the intention in its own UI.

```
enum CredentialCreationIntention  {
    “sole-first-factor”,
    “alternative-first-factor”,
    “second-factor"
};
```

A public key credential created by this API can be used for various different purposes. This enumeration defines intentions as to how the about-to-create credential will be used.

**_sole-first-factor_**
&nbsp;&nbsp;&nbsp;&nbsp;Indicates the intention is to use the public key credential as the only factor for authentication. If a password is used previously, this indicates the about-to-create credential will replace that password.
**_alternative-first-factor_**
&nbsp;&nbsp;&nbsp;&nbsp;Indicates the intention is to use the public key credential as an alternative method for authentication. If a password is used previously, this indicates that either the password or the about-to-create credential could be used for any given authentication.
**_second-factor_**
&nbsp;&nbsp;&nbsp;&nbsp;Indicates the intention is to use the public key credential as a second factor together with other factors for authentication. If a password is used previously, this indicates the about-to-create credential will be used together with that password.

b) Reasons:

The “intension" member is to describe the purpose of the relying party to create the public key credential, for example, a password alternative, a password replacement or a second factor. Such string is useful for user agents that provide a dedicated UI for instructing users to use WebAuthn, so that user agents can provide more context to their users of what this operation is intended to do instead of providing a generic string. For example, instead of “Use your security key with example.com”, user agents now could suggest “Do you want to _create a second factor_ using your security key with example.com?”. From the user perspective, users can now reason what the user agent asks for instead of guessing. Even though relying parties could have already explained the intention to users before calling WebAuthn API, it is still very useful to channel this intention all the way to the screen where users are actually taking the final action.

lgarron:
While there are lots of knobs in the spec (user verification, platform auth, resident key), this kind of intention is definitely a useful abstraction from a website implementation — a lot of our work at GitHub involved trying to map the knobs onto recommended values for this "intention" abstraction.

[...]

PR #1300, opened 2019-09-12
WIP: More explicitly document use cases.

### LINKED ISSUE #1389: More explicitly document use cases
We need to provide more specific guidance for RPs about how to configure the several options for `navigator.credentials.{create,get}` and, secondly, to provide a reserve mapping of options back to intents.

See PR #1300

agl:
The aims of this PR (which is currently incomplete) are, firstly, to provide more specific guidance for RPs about how to configure the several options in WebAuthns and, secondly, to provide a reserve mapping of options back to intents.

On the first motive: the feedback that we're getting from several RPs is that the WebAuthn spec is dauntingly complex. They go into it with a certain user experience in mind but the translation of that intent to concrete options is extremely challenging. While third-party documentation can also help address this, the spec should stand alone.

On the second motive: #1292 contains a request from a browser to add an extra option to translate the other options into a higher-level motive. The hope is that, in defining a mapping fro high-level intent to specific options, we also implicitly define the reverse mapping and so browsers can also extract the intent from the options.

It is unlikely that the reverse mapping will be total, rather some combinations of options will not map to any higher-level intent. Considering that set is hoped to be useful.

This PR is not yet complete and is being submitted in its current form ahead of TPAC to help inform the discussion around #1292.

Fixes #1389

***
<a href="https://pr-preview.s3.amazonaws.com/agl/webauthn/pull/1300.html" title="Last updated on Mar 18, 2020, 6:26 PM UTC (c2404d3)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/1300/32c216c...agl:c2404d3.html" title="Last updated on Mar 18, 2020, 6:26 PM UTC (c2404d3)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -338,42 +338,61 @@ refer to [[WebAuthnAPIGuide]] for an overall tutorial.
 
 ## Use Cases ## {#sctn-use-cases}
 
-The below use case scenarios illustrate use of two very different types of [=authenticators=], as well as outline further
-scenarios. Additional scenarios, including sample code, are given later in [[#sctn-sample-scenarios]].
+Several user experiences can be crafted using this specification. This section enumerates the most common and specifies the values for the numerous options in {{PublicKeyCredentialCreationOptions}} and {{PublicKeyCredentialRequestOptions}} that will invoke them. By implementing one of these patterns, a Relying Party should find that users suffer less confusion as the structure of the interaction should be more familiar to them.
 
-### Registration ### {#sctn-usecase-registration}
+Issue: credProtect is not currently discussed.
 
-- On a phone:
-    * User navigates to example.com in a browser and signs in to an existing account using whatever method they have been using
-        (possibly a legacy method such as a password), or creates a new account.
-    * The phone prompts, "Do you want to register this device with example.com?"
-    * User agrees.
-    * The phone prompts the user for a previously configured [=authorization gesture=] (PIN, biometric, etc.); the user
-        provides this.
-    * Website shows message, "Registration complete."
+Issue: “tap-and-go” uses are not currently discussed.
+
+Issue: “passwordless” (where a username is entered and the RP requests an assertion with UV for a list of registered credentials) is not discussed. This term is sometimes confused with “typing-free”.
+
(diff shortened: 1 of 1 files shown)


balfanz:
Thanks for putting this together. It seems to me, though, that there are even more variations of parameters to consider, which map to more and/or other use cases. For example:

- What kind of attestation is requested (none vs indirect vs direct) may affect whether the RP can "inventory" their authenticators (i.e., imagine an RP that wants to ensure users use authenticators from a defined inventory).

- Whether or not the RP uses the hmacCreateSecret extension affects whether the credential/authenticator can be used to unlock devices.

- Whether or not the RP uses the credProtect extension, and which of its possible six values it selects, affects whether the authenticator can be used in certain tap-to-sign-in scenarios.

These are just off the top of my head - there might be more parameters or other extensions that also induce separate use cases that RPs may want to consider.

index.bs
```diff
@@ -338,42 +338,61 @@ refer to [[WebAuthnAPIGuide]] for an overall tutorial.
 
 ## Use Cases ## {#sctn-use-cases}
 
-The below use case scenarios illustrate use of two very different types of [=authenticators=], as well as outline further
-scenarios. Additional scenarios, including sample code, are given later in [[#sctn-sample-scenarios]].
+Several user experiences can be crafted using this specification. This section enumerates the most common and specifies the values for the numerous options in {{PublicKeyCredentialCreationOptions}} and {{PublicKeyCredentialRequestOptions}} that will invoke them. By implementing one of these patterns, a Relying Party should find that users suffer less confusion as the structure of the interaction should be more familiar to them.
```
equalsJeffH: rather than jumping directly into API options settings and named flows with little real-world context, I am thinking that we might largely retain the present [Use Cases section](https://w3c.github.io/webauthn/#sctn-use-cases), though appropriately edit it (move the passwordless/Bluetooth example to end, add a distinct 2nd factor scenario(?)), and incorporate links to further sections that describe the options settings yielding said use case scenarios. WDYT?

PR #1307, opened 2019-09-24
Add note about use of floating point in extensions passed through that clients do not recognize

### LINKED ISSUE #1273: Certain issues in client extension pass-through specification
There are two issues with Section 9, which discusses how clients pass through unrecognized extensions.

1.  In the algorithm for converting client extension inputs to CBOR, section 9 currently says: "When the JavaScript value _is a non-integer number_, it is converted to a 64-bit CBOR floating point number."  But until recently, JavaScript has only one number type, namely a 64-bit floating-point number of the class `Number`, and no integer types such as 32- or 64-bit integers.  And currently, of the defined WebAuthn extensions, only biometricPerfBounds includes number types, namely `float`, in client extension inputs. This suggests to me that—

    - section 9 can simply say "When the JavaScript value is a `float` [or `float` or `double`]..." rather than "When the JavaScript value is a non-integer number", and
    - WebAuthn should discourage new extension definitions that use any number type other than float (or `double`) in their client extension inputs (due to the [lack of a threshold](https://github.com/w3c/webauthn/issues/1044) above which integers are converted to floating-point numbers rather than CBOR integers).

2.  When a client passes through an unrecognized extension, the special cases for ArrayBuffers, non-integer numbers, and CBOR byte strings are probably intended to apply when those values appear as keys and values of arrays and maps, rather than just when they stand alone (one possible interpretation due to how this section incorporates RFC7049 by reference).  This can be addressed, for example, by adding text as follows: "However, if the JavaScript value contains ArrayBuffers and/or non-integer numbers [contains ArrayBuffers, `float`s, and/or `double`s], they are converted as described above rather than as given in that section."

selfissued:
Fixes #1273

***
<a href="https://pr-preview.s3.amazonaws.com/selfissued/webauthn/pull/1307.html" title="Last updated on Oct 9, 2019, 7:41 PM UTC (d1dc052)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/1307/2b435a7...selfissued:d1dc052.html" title="Last updated on Oct 9, 2019, 7:41 PM UTC (d1dc052)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -4827,7 +4827,7 @@ operation. Likewise, clients can choose to produce a [=client extension output=]
 understand by encoding the [=authenticator extension output=] value into JSON, provided that the CBOR output uses only types
 present in JSON.
 
-When clients choose to pass through extensions they do not recognize,
+When [=clients=] choose to pass through extensions they do not recognize,
 the JavaScript values in the [=client extension inputs=]
 are converted to [=CBOR=] values in the [=authenticator extension inputs=].
 When the JavaScript value is an [=%ArrayBuffer%=], it is converted to a [=CBOR=] byte array.
@@ -4835,6 +4835,12 @@ When the JavaScript value is a non-integer number, it is converted to a 64-bit C
 Otherwise, when the JavaScript type corresponds to a JSON type, the conversion is done
 using the rules defined in Section 4.2 of [[!RFC7049]] (Converting from JSON to CBOR),
 but operating on inputs of JavaScript type values rather than inputs of JSON type values.
+Note that the numeric conversion rules have the consequence that
+when a client passes through an extension it does not recognize,
+if the extension uses floating point values,
+[=authenticators=] need to be prepared to receive those values as [=CBOR=] integers
+when the floating point values used happen to be integers,
+should the [=authenticator=] want the extension to always work without actual [=client=] support for it.
 Once these conversions are done,
 canonicalization of the resulting [=CBOR=] MUST be performed using the [=CTAP2 canonical CBOR encoding form=].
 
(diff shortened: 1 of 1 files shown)


emlun:
@peteroupc I don't think that should be a problem, at least for WebAuthn ceremonies. The RP has no direct control of the exact bytes being sent to the authenticator anyway (because the client can add random extra stuff to the CollectedClientData, for example), so you can't rely on the same input producing the exact same outputs even with deterministic signatures.

peteroupc:
If that's the case, then there should be an additional note that different clients might choose different integer thresholds for converting floating-point values to integers, and that this is not a problem because canonicalized CTAP2 messages generated by the protocol are not meant to be compared bit-for-bit with each other (either directly or via hash codes).  This is unlike other protocols like DKIM and XML Canonicalization, which do use canonicalization for bit-for-bit data comparison.

index.bs
```diff
@@ -4827,14 +4827,20 @@ operation. Likewise, clients can choose to produce a [=client extension output=]
...
 When the JavaScript value is a non-integer number, it is converted to a 64-bit CBOR floating point number.
 Otherwise, when the JavaScript type corresponds to a JSON type, the conversion is done
 using the rules defined in Section 4.2 of [[!RFC7049]] (Converting from JSON to CBOR),
 but operating on inputs of JavaScript type values rather than inputs of JSON type values.
+Note that the numeric conversion rules have the consequence that
+when a client passes through an extension it does not recognize,
+if the extension uses floating point values,
+[=authenticators=] need to be prepared to receive those values as [=CBOR=] integers
+when the floating point values used happen to be integers,
+should the [=authenticator=] want the extension to always work without actual [=client=] support for it.
```
equalsJeffH: This is all one long sentence -- suggested rewrite (plus I'd move this Note to under line 4834 where numeric conversion is discussed):

Note: Due to Javascript's numeric conversion rules, any floating point values may be converted to integers. Thus, [=authenticators=] need to be prepared to receive those values as either [=CBOR=] floating-point or as [=CBOR=] integers. If this is not feasible, the extension may be designed with the [=client=] performing any necessary value type conversions.

peteroupc:
For your information: See cbor-wg/CBORbis#117, which modifies the CBOR revision to include defaults for encoding JSON numbers as CBOR.

Issue #1306, opened 2019-09-21
Conflict between WebAuthn and CTAP2 handling of userVerification mapping


FabianHenneke:
As explained in [this comment by @akshayku](https://github.com/w3c/webauthn/issues/1123#issuecomment-448198487) and [this comment by @agl](https://bugs.chromium.org/p/chromium/issues/detail?id=1005744#c2), the current (non-public) version of the CTAP2 spec appears to demand that all `authenticatorMakeCredential` requests are made with `userVerification` set to `true` if the authenticator has UV configured, which is in conflict [5.1.3, Step 20](https://www.w3.org/TR/webauthn-2/#sctn-createCredential) of the WebAuthn spec.

To me this seems to be due to different points of view on user verification underlying the CTAP2 and WebAuthn spec:
In CTAP2, the scope of user verification, for example via client PIN, is the entire authenticator: In order to ensure that the same user is generating and using credentials, any sensitive operation (i.e. any call to `authenticatorMakeCredential` and any user-verifying call to `authenticatorGetAssertion`) requires the authenticator to be unlocked, e.g., by entering the PIN.
In WebAuthn, the scope of user verification is not discussed in detail, but could potentially be per credential. For example, a software authenticator backed by the Android Keystore could maintain both user-verifying and non-user-verifying credentials simultaneously, and both types of credentials could be added independently. In particular, an authenticator of this type could both be advertising that it is capable of user verification and handle `authenticatorMakeCredential` requests with `userVerification` set to `false`, which matches the behavior mandated by the current version of the spec.

The mismatch between the two approaches leads to rather unsatisfying workarounds, such as Chrome falling back to U2F to bypass PIN prompts for `authenticatorMakeCredential` requests which use none of the CTAP2-only features. 

A possible, partial solution to this issue that would require no major changes to any spec would be to add a WebAuthn extension (or, even better, an `authenticatorGetInfo` response map entry) that allows an authenticator to switch between the two modes described above. This could be called `userVerificationScope`, taking the value `authenticator` on low-power authenticators such as hardware security keys and `credential` on authenticators that are e.g. backed by the Android Keystore, macOS' TouchID or a generic TPM. The value `authenticator`, which would also be the default, would keep the current CTAP2 behavior, whereas the value `credential` would make the client follow the current WebAuthn spec.

I would be grateful for any feedback on this proposal and the general situation around this part of the two specifications.

[...]

PR #1313, opened 2019-10-07
Clarify assumptions about single-user authenticator and relation to UV

### LINKED ISSUE #1305: Clarify relation between requireUserVerification values for MakeCredential and GetAssertion
In [6.3.3 The authenticatorGetAssertion Operation, Step 7](https://www.w3.org/TR/webauthn-2/#sctn-op-get-assertion), it is specified that if user verification is performed, it has to be ensured that the user verified is the same as the user verified during the corresponding call to the `authenticatorMakeCredential` operation. As a consequence of this requirement, user verification had to be performed during the corresponding call to `authenticatorMakeCredential`, so that this part of the spec can only apply to the situation where `requireUserVerification` had been set to `true` for both calls.

Now there are three other cases for the combination of the values of `requireUserVerification` in the `authenticatorMakeCredential` and `authenticatorGetAssertion` calls, two of which I believe may not be handled in sufficient detail by the spec:

1. `requireUserVerification` set to `false` in both calls: No user verification required at any step, behaves analogously to plain U2F.
2. `requireUserVerification` set to `true` in `authenticatorMakeCredential` and to `false` in `authenticatorGetAssertion`: In this case, I think that it is not sufficiently clear whether the authenticator is expected to a) allow the use of the credential without performing user verification and note this by setting the UV flag to 0 or b) decline to use any credential created with user verification, potentially erroring out if there are no credentials that can be used without user verification.
3. `requireUserVerification` set to `false` in `authenticatorMakeCredential` and to `true` in `authenticatorGetAssertion`: Because of the requirement cited above, it is unclear how to treat this situation since it cannot be ensured that the current user is the one who created the credential, but at the same time user verification must be performed. Does this mean that an error should be returned and if so, which one?

emlun:
Fixes #1305.

This is what I gathered of the opinions expressed on the 2019-10-02 WG call. Please point out anything I misunderstood or that needs further clarification.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/pull/1313.html" title="Last updated on Oct 14, 2019, 10:52 AM UTC (101146b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/1313/976391d...101146b.html" title="Last updated on Oct 14, 2019, 10:52 AM UTC (101146b)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -883,6 +883,13 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
     A [=[WAA]=] could be a [=roaming authenticator=], a dedicated hardware subsystem integrated into the [=client device=],
     or a software component of the [=client=] or [=client device=].
 
+    In general, an [=authenticator=] is assumed to have only one user.
+    If multiple natural persons share access to an [=authenticator=],
+    they are considered to represent the same user in the context of that [=authenticator=].
+    If an [=authenticator=] implementation supports multiple users in separated compartments,
+    then each compartment is considered a separate [=authenticator=]
+    with a single user with no access to other users' [=credentials=].
+
 : <dfn>Authorization Gesture</dfn>
 :: An [=authorization gesture=] is a physical interaction performed by a user with an authenticator as part of a [=ceremony=],
     such as [=registration=] or [=authentication=]. By making such an [=authorization gesture=], a user [=user consent|provides
@@ -1142,10 +1149,15 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
     [=biometric recognition=] (e.g., presenting a fingerprint) [[ISOBiometricVocabulary]]. The intent is to
     distinguish individual users.
 
+    Note that [=user verification=] does not give the [=[RP]=] a concrete identification of the user,
+    only an assertion that multiple instances of [=user verification=] for the same [=credential=] are performed by the same user.
+    The same user might not always be the same natural person, however,
+    if multiple natural persons share access to the same [=authenticator=].
+
     Note: Distinguishing natural persons depends in significant part upon the [=client platform=]'s
     and [=authenticator=]'s capabilities.
     For example, some devices are intended to be used by a single individual,
-    yet they may allow multiple natural persons to enroll fingerprints
+    yet they may allow multiple natural persons to enroll fingerprints or know the same PIN
     and thus access the same [=[RP]=] account(s) using that device.
     See also [[#sctn-uvi-extension]].
 
--- a/index.bs
+++ b/index.bs
@@ -1150,7 +1150,8 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
     distinguish individual users.
 
     Note that [=user verification=] does not give the [=[RP]=] a concrete identification of the user,
-    only an assertion that multiple instances of [=user verification=] for the same [=credential=] are performed by the same user.
+    but when 2 or more ceremonies with [=user verification=] have been done with that [=credential=]
+    it expresses that it was the same user that performed all of them.
     The same user might not always be the same natural person, however,
     if multiple natural persons share access to the same [=authenticator=].
 
--- a/index.bs
+++ b/index.bs
@@ -3567,10 +3567,6 @@ When this method is invoked, the [=authenticator=] MUST perform the following pr
 
     If |requireUserVerification| is [TRUE], the [=authorization gesture=] MUST include [=user verification=].
 
-        Note: For the overall WebAuthn security properties to hold, the user verified here must be the same user that was
-        verified when this [=authenticator=] created this [=public key credential=] in
-        <a href='#op-makecred-step-user-consent'>Step 6</a> of [[#sctn-op-make-cred]].
-
     If |requireUserPresence| is [TRUE], the [=authorization gesture=] MUST include a
         [=test of user presence=].
 


index.bs
```diff
@@ -1142,10 +1149,15 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
     [=biometric recognition=] (e.g., presenting a fingerprint) [[ISOBiometricVocabulary]]. The intent is to
     distinguish individual users.
 
+    Note that [=user verification=] does not give the [=[RP]=] a concrete identification of the user,
+    only an assertion that multiple instances of [=user verification=] for the same [=credential=] are performed by the same user.
```
akshayku: Multiple?

FabianHenneke: How about "all instances of (successful?) [=user verification=] have been performed by the same user"?

> but when you have done 2 or more UV ceremonies with that credential it expresses that it was the same user that performed all of them.

Just to clarify a subtle point: Would this guarantee about the user also be expected to apply in the case where the credential was created without a UV ceremony, but is then used multiple times to generate an assertion with UV requested?

emlun: >but when you have done 2 or more UV ceremonies with that credential it expresses that it was the same user that performed all of them.

I like this idea, thanks!

>Would this guarantee about the user also be expected to apply in the case where the credential was created without a UV ceremony, but is then used multiple times to generate an assertion with UV requested?

Yes, that is my intent.

FabianHenneke: > Yes, that is my intent.

I asked for clarification on this point because I believe that it could unnecessarily limit future authenticator designs. For example, on a high-powered authenticator (like a phone with a TEE or SE), UV would probably be tracked by binding an individual credential created with UV requested to the current user at the time or even their concrete biometric information. Introducing a meaningful user binding after creation time is often not possible or at least not well supported in hardware by existing key store implementations (like the Android Keystore or iOS Keychain). If UV were not a flag, but instead a cryptographic fingerprint like the one provided by the "uvi" extension, this would be less of an issue.

It is also not clear to me what kind of application could make use of the guarantee that the same user requested multiple assertions for a credential that may have been created by someone else. Is this something you could see a potential use case for? If not, it may be worthwhile to make the guarantee here slightly more restrictive in order to not complicate authenticators in an unnecessary way.

emlun: > It is also not clear to me what kind of application could make use of the guarantee that the same user requested multiple assertions for a credential that may have been created by someone else.

Indeed, that's not likely to be an explicit use case. I wrote this mostly to reflect how most roaming authenticators work: you can create credentials without setting up UV (often by PIN) on the authenticator, but once you have UV set up you can't disable it without destroying all your credential keys.

So it's not so much a use case as an emergent consequence of how implementations interact. It's also not unthinkable that an RP might start out with `userVerification: "discouraged"`, but later decide to move more towards passwordless MFA and change to `userVerification: "preferred"`. In that case they probably have a bunch of users whose credentials were created without UV. It doesn't really matter if UV wasn't set up when the credential was created - the RP can still use UV in place of a password for future ceremonies, as long as the _first_ UV ceremony is also authenticated by the password (or some other second factor).

> Introducing a meaningful user binding after creation time is often not possible or at least not well supported in hardware by existing key store implementations (like the Android Keystore or iOS Keychain).

Ok, but then that sounds to me like they wouldn't be able to support UV for credentials created without UV, correct? I.e., they'd be skipped in [step 18 of `get()`](https://www.w3.org/TR/2019/WD-webauthn-2-20190604/#ref-for-dom-publickeycredentialrequestoptions-userverification):

>If `options.userVerification` is set to `required` and the authenticator is not capable of performing user verification, continue.

FabianHenneke: > So it's not so much a use case as an emergent consequence of how implementations interact. It's also not unthinkable that an RP might start out with `userVerification: "discouraged"`, but later decide to move more towards passwordless MFA and change to `userVerification: "preferred"`. In that case they probably have a bunch of users whose credentials were created without UV. It doesn't really matter if UV wasn't set up when the credential was created - the RP can still use UV in place of a password for future ceremonies, as long as the _first_ UV ceremony is also authenticated by the password (or some other second factor).

This is a situation I hadn't thought about, thanks for the example. I agree that this is a situation in which allowing to ascertain UV at a later point in time is advantageous.

> Ok, but then that sounds to me like they wouldn't be able to support UV for credentials created without UV, correct? I.e., they'd be skipped in step 18 of get():

So far I have taken this step to not depend on the registered credentials, but only care about whether the authenticator is configured to support UV (e.g. has a fingerprint registered) at the moment. I may have misunderstood this part of the spec though, please let me know if it's supposed to say something else.

My expectation would rather have been that the authenticator would simply not return the UV flag for credentials registered without UV, regardless of whether UV was required or not. Would this be a good way for authenticators to behave if they cannot track users of non-UV keys, but still support UV in general?

emlun: > So far I have taken this step to not depend on the registered credentials, but only care about whether the authenticator is configured to support UV (e.g. has a fingerprint registered) at the moment. I may have misunderstood this part of the spec though, please let me know if it's supposed to say something else.

I don't quite understand what you mean here. WebAuthn ceremonies are all about credentials, you can't do anything without one.

> My expectation would rather have been that the authenticator would simply not set the UV flag in GetAssertion responses for credentials created without UV, even if UV is required for the corresponding request. Would this be a good way for authenticators to behave if they cannot track users of non-UV keys, but still support UV in general?

Yes, that's exactly right. The UV flag in the authenticator response indicates whether UV was used in this particular ceremony. If UV=1, then the authenticator must of course have UV configured, but UV=0 does not imply that the authenticator is not configured or not capable of UV.

FabianHenneke: > Yes, that's exactly right. The UV flag in the authenticator response indicates whether UV was used in this particular ceremony. If UV=1, then the authenticator must of course have UV configured, but UV=0 does not imply that the authenticator is not configured or not capable of UV.

Thanks for the confirmation, this clears up the potential issues I saw with the user-tracking guarantee explained in this PR as authenticators have a way to gracefully decline tracking a user if the credential was created without UV. Additionally, being allowed to return UV=0 even if the request has set UV to required helps with #1306 to some extent.

> I don't quite understand what you mean here. WebAuthn ceremonies are all about credentials, you can't do anything without one.

I apologize for not being clear enough in my reply. I just wanted to remark that the part of the spec you referenced is part of the steps to take "If an authenticator becomes available on this client device,", and hence the information retrieved from the authenticator in this step would probably only depend on the general state and configuration of the authenticator rather than the properties of a particular credential. But this doesn't really matter anymore thanks to your clarification above.

emlun: > Additionally, being allowed to return UV=0 even if the request has set UV to required helps with #1306 to some extent.

Actually, this isn't allowed - if the request sets `userVerification: "required"` then the authenticator must perform UV or fail. It is fine to return UV=0 for `userVerification: "preferred"`, though.

FabianHenneke: > Actually, this isn't allowed - if the request sets `userVerification: "required"` then the authenticator must perform UV or fail. It is fine to return UV=0 for `userVerification: "preferred"`, though.

Since the mapping of `userVerification: preferred` to either `true` or `false` will loose this information before it reaches the authenticator, this may unfortunately not help much. Thank you very much for all the clarifications though!

It seems to me that a "true" solution to this and other issues (#1306 + the somewhat hacky U2F fallback in Chrome, and also the client not being able to judge whether UV is "low-effort" on the authenticator) would be to pass the full trivalent `userVerification` set by the RP through to the authenticator via an extension. I would gladly work on this if you should find it a reasonable proposal, but I can understand if this is not a priority because current authenticator designs would not be able to make use of it.

emlun: Ok. Let's close this discussion, then. You're welcome to open a new issue for further clarifications/improvements.

index.bs
```diff
@@ -883,6 +883,13 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
     A [=[WAA]=] could be a [=roaming authenticator=], a dedicated hardware subsystem integrated into the [=client device=],
     or a software component of the [=client=] or [=client device=].
 
+    In general, an [=authenticator=] is assumed to have only one user.
+    If multiple natural persons share access to an [=authenticator=],
+    they are considered to represent the same user in the context of that [=authenticator=].
+    If an [=authenticator=] implementation supports multiple users in separated compartments,
+    then each compartment is considered a separate [=authenticator=]
+    with a single user with no access to other users' [=credentials=].
```
equalsJeffH: perhaps "single" should be "given" ?

index.bs
```diff
@@ -1142,10 +1149,16 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
     [=biometric recognition=] (e.g., presenting a fingerprint) [[ISOBiometricVocabulary]]. The intent is to
     distinguish individual users.
 
+    Note that [=user verification=] does not give the [=[RP]=] a concrete identification of the user,
+    but when 2 or more ceremonies with [=user verification=] have been done with that [=credential=]
+    it expresses that it was the same user that performed all of them.
+    The same user might not always be the same natural person, however,
+    if multiple natural persons share access to the same [=authenticator=].
```
equalsJeffH: perhaps add an example? 

"For example, by sharing the authentictor PIN."

emlun: There's a Note: directly following with some examples. Is that enough? :slightly_smiling_face:

index.bs
emlun: I think "single" better emphasises that there's only one user per semantic authenticator.

index.bs
equalsJeffH: doh!  :-P

index.bs
```diff
@@ -1142,10 +1149,16 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
...
+
     Note: Distinguishing natural persons depends in significant part upon the [=client platform=]'s
     and [=authenticator=]'s capabilities.
     For example, some devices are intended to be used by a single individual,
-    yet they may allow multiple natural persons to enroll fingerprints
+    yet they may allow multiple natural persons to enroll fingerprints or know the same PIN
```
equalsJeffH: good addiion, this.  :)

emlun: Thanks! :)

PR #1316, opened 2019-10-15
Truncate strings for authenticators where needed.

### LINKED ISSUE #1296: Consider whether browsers should truncate `name` and `displayName`.
We say of [PublicKeyCredentialEntity](https://www.w3.org/TR/webauthn/#dictionary-pkcredentialentity)'s `name` and [PublicKeyCredentialUserEntity](https://www.w3.org/TR/webauthn/#sctn-user-credential-params)'s `displayName`, that:

> Authenticators MAY truncate [&hellip;] to a length equal to or greater than 64 bytes.

**Some authenticators do, indeed, truncate these strings at 64 bytes**. This may cut a UTF-8 sequence and thus produce invalid strings in responses. Because of this, WebAuthn now has [a section](https://w3c.github.io/webauthn/#sctn-strings) on handling these truncated strings.

However, some authenticators will also _**reject any operations where the strings are longer than 64 bytes**_. Whatever the fidelity of that behaviour to the spec, it's now reality. _Therefore we should consider whether it's the browser's job to do the truncation in order to ensure that requests work_. If we do not, then RPs will probably do it themselves because things will fail otherwise and once RPs do it, we'll not be able to add support for future authenticators which have more storage.

(Since there exist truncating authenticators out there, [the rules](https://w3c.github.io/webauthn/#sctn-strings) around string handling still apply because those authenticators aren't going away either.)

(I don't know, off-hand, what the behaviour for `icon` is in practice.)

(This relates to fido-alliance/fido-2-specs#673.)

agl:
There exist a significant number of authenticators that do not conform
to the current WebAuthn requirements in that they fail requests with
name/displayName strings longer than 64 bytes, rather than truncating
them.

This change adds a new requirement on user-agents that they maintain the
authenticator model for RPs by doing the truncation on their behalf in
this case. The alternative is that each RP will hit this edge-case and
do the truncation itself, thus the ecosystem will never be able to
support longer strings.

Since user-agents may now be doing truncation, this change also permits
truncation at the level of grapheme clusters (since user-agents
presumably have Unicode tables available).

Fixes #1296.

***
<a href="https://pr-preview.s3.amazonaws.com/agl/webauthn/pull/1316.html" title="Last updated on Oct 17, 2019, 8:57 PM UTC (e75a6f0)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/1316/36fe4ed...agl:e75a6f0.html" title="Last updated on Oct 17, 2019, 8:57 PM UTC (e75a6f0)">Diff</a>


jcjones:
Note: SVG is viewable here https://github.com/w3c/webauthn/blob/c7406dde0eacd1b8b6d2698a89dd324377f52409/images/string-truncation.svg

index.bs
```diff
@@ -3651,11 +3649,18 @@ or [=authenticatorGetAssertion=] operation currently in progress.
...
+If string value truncation is the chosen accommodation then the amount truncated MUST be, at most, the minimum amount neccessary while respecting UTF-8 sequence and [=grapheme cluster=] boundaries. For example, in the following example the string is 65 bytes long. If truncating to 64 bytes then the final 0x88 byte must be dropped purely because of space reasons. Since that leaves a partial UTF-8 sequence an authenticator MAY drop the remainder of that sequence. Since that leaves a partial [=grapheme cluster=] an authenticator MAY drop the remainder of that cluster. That defines that maximum truncation permitted and authenticators MUST NOT truncate further.
 
-Additionally, a [=grapheme cluster=] may be truncated [[UTR29]], <i>altering</i> the resulting character in the mind of the user.
+<figure id="fig-stringTruncation">
```
equalsJeffH: Adding this figure here causes it to be "figure 5" and thus causes renumbering of the former figure 5 to be figure 6.  We need to update all four of the occurrances:

`s/ [Figure 5](#fig-attStructs) / [Figure 6](#fig-attStructs) /`

index.bs
```diff
@@ -3651,11 +3649,18 @@ or [=authenticatorGetAssertion=] operation currently in progress.
...
+    <img src="images/string-truncation.svg"/>
+    <figcaption>A UTF-8 encoded string showing the positions where different truncation algorithms would cut it.</figcaption>
```
equalsJeffH: s/ A UTF-8 / The end of a UTF-8 /   ...?

s/ showing the positions where different truncation algorithms would cut it / showing the positions of different trucation boundaries /    ....?

index.bs
```diff
@@ -3651,11 +3649,18 @@ or [=authenticatorGetAssertion=] operation currently in progress.
...
+</figure>
+
+[=Conforming User Agents=] are responsible for ensuring that the authenticator behaviour observed by [=[RPS]=] conforms to this specification with respect to string handling. For example, if an authenticator is known to behave incorrectly when asked to store large strings, the user-agent should perform the truncation for it in order to maintain the model from the point of view of the [=[RP]=]. User-agents that do this SHOULD truncate at [=grapheme clusters=].
```
equalsJeffH: we do not hyphenate "user-agent" in the rest of the spec  :)

index.bs
```diff
@@ -3651,11 +3649,18 @@ or [=authenticatorGetAssertion=] operation currently in progress.
...
+
+Authenticators that truncate on UTF-8 sequences alone may cause a [=grapheme cluster=] to be truncated [[UTR29]], <i>altering</i> the resulting character in the mind of the user.
```
equalsJeffH: Perhaps this admonition should be generally applicable to whichever actor is performing truncation?  Perhaps this accomplishes that:

s/ Authenticators that truncate on UTF-8 sequences alone / Truncation on UTF-8 sequence boundaries alone /

...?

index.bs
equalsJeffH: s/ to store large strings / to store strings larger that its maximum string length capability /    ...?

index.bs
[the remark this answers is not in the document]
emlun: "The resulting character in the mind of the user" sounds a bit weird to me. How about something like this?

```suggestion
Authenticators that truncate on UTF-8 sequences alone may cause a [=grapheme cluster=] to be truncated, but still valid [[UTR29]]. This could make the grapheme cluster render as a different valid glyph instead of removing the glyph entirely.
```

index.bs
```diff
@@ -3651,11 +3649,18 @@ or [=authenticatorGetAssertion=] operation currently in progress.
...
+If string value truncation is the chosen accommodation then the amount truncated MUST be, at most, the minimum amount neccessary while respecting UTF-8 sequence and [=grapheme cluster=] boundaries. For example, in the following example the string is 65 bytes long. If truncating to 64 bytes then the final 0x88 byte must be dropped purely because of space reasons. Since that leaves a partial UTF-8 sequence an authenticator MAY drop the remainder of that sequence. Since that leaves a partial [=grapheme cluster=] an authenticator MAY drop the remainder of that cluster. That defines that maximum truncation permitted and authenticators MUST NOT truncate further.
```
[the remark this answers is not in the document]
agl: Done. (Tweaked first sentence a bit. Let me know if it doesn't make sense.)

index.bs
agl: Ugh. Having to maintain these things manually is unfortunate. Done.

index.bs

index.bs
agl: Fixed.

index.bs

index.bs
agl: (Didn't do this one. Extra words didn't seem to carry their weight.)

index.bs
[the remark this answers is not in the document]

index.bs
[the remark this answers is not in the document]

index.bs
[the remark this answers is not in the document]

agl:
Built a way to eliminate some of the manual figure references on top of this change in #1323.

agl:
For ease of reviewing, here is the updated image from this change

image

index.bs

equalsJeffH:
LGTM @agl -- the figure is a great addition

equalsJeffH:
shouldn't this be merged by now?

PR #1317, opened 2019-10-16
Timeout considerations

### LINKED ISSUE #1286: Low timeout bounds for inline bio enrollment of FIDO2 keys
We are looking into inline bio enrollment of FIDO2 keys during webauthn calls. Current time bound in webauthn is 15-120 seconds, which when we decided was a guess for a single touch or PIN input. Browsers have been hooked up to cancel the transaction when timeout happens. 

Many RPs don't configure the timeout, resulting in 15 seconds. Even 120 seconds is not enough given our user studies where user has to figure out which authenticator he/she wants to use, plug that in, setup PIN if not present, setup fingerprint which requires multiple samples. Last step also depends on how many samples authenticator wants. 

Want to take suggestions on what the typical timeout should be to support this case. I am thinking 180-300 seconds. 

Thoughts?

akshayku:
This fixes #1286

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/pull/1317.html" title="Last updated on Oct 18, 2019, 3:56 AM UTC (ea3419c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/1317/36fe4ed...ea3419c.html" title="Last updated on Oct 18, 2019, 3:56 AM UTC (ea3419c)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1388,8 +1388,17 @@ When this method is invoked, the user agent MUST execute the following algorithm
     |lifetimeTimer| to this adjusted value. If the {{PublicKeyCredentialCreationOptions/timeout}} member of |options| is [=present|not
     present=], then set |lifetimeTimer| to a [=client=]-specific default.
 
-        Note: A suggested reasonable range for the {{PublicKeyCredentialCreationOptions/timeout}} member of |options| is
-        15 seconds to 120 seconds.
+      * Recommended ranges for the {{PublicKeyCredentialCreationOptions/timeout}} member of |options| are as follows.
+          If <code>|options|.{{PublicKeyCredentialCreationOptions/authenticatorSelection}}.{{AuthenticatorSelectionCriteria/userVerification}}</code>
+            <dl class="switch">
+                :   is set to {{UserVerificationRequirement/discouraged}}
+                ::  Recommended range: 30 seconds to 180 seconds.
+                ::  Default recommended value for user agents if {{PublicKeyCredentialCreationOptions/timeout}} is not specified: 120 seconds.
+
+                :   is set to {{UserVerificationRequirement/required}} or {{UserVerificationRequirement/preferred}}
+                ::  Recommended range: 30 seconds to 600 seconds.
+                ::  Default recommended value for user agents if {{PublicKeyCredentialCreationOptions/timeout}} is not specified: 300 seconds.
+            </dl>
 
         Note: The user agent should take cognitive guidelines into considerations regarding timeout for users with special needs.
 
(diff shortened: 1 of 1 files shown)


agl:
From the call of 2019-10-16: J. C. notes that that this could use a web-platform test. (@nsatragno.)

index.bs
```diff
@@ -1388,8 +1388,17 @@ When this method is invoked, the user agent MUST execute the following algorithm
     |lifetimeTimer| to this adjusted value. If the {{PublicKeyCredentialCreationOptions/timeout}} member of |options| is [=present|not
     present=], then set |lifetimeTimer| to a [=client=]-specific default.
 
-        Note: A suggested reasonable range for the {{PublicKeyCredentialCreationOptions/timeout}} member of |options| is
-        15 seconds to 120 seconds.
+      * Recommended ranges for the {{PublicKeyCredentialCreationOptions/timeout}} member of |options| are as follows.
```
agl: I don't think this needs a lone bullet point? Just a paragraph would seem ok.

(Here and below.)

agl:
LGTM. Pending chromium change to reflect this in https://chromium-review.googlesource.com/c/chromium/src/+/1865755

index.bs
equalsJeffH: agree w/@agl

equalsJeffH:
one thought I just had when looking at examples 1, 3, & 4 is that `|options|` timeout member's value's units is milliseconds, and the guidance we are giving in the new spec text herein is in seconds.  I wonder whether we ought to note that within these new recommendations somehow.

index.bs
emlun: Maybe?

```suggestion
      * Recommended ranges and defaults for the {{PublicKeyCredentialCreationOptions/timeout}} member of |options| are as follows.
```

index.bs
```diff
@@ -1388,8 +1388,17 @@ When this method is invoked, the user agent MUST execute the following algorithm
...
+          If <code>|options|.{{PublicKeyCredentialCreationOptions/authenticatorSelection}}.{{AuthenticatorSelectionCriteria/userVerification}}</code>
+            <dl class="switch">
+                :   is set to {{UserVerificationRequirement/discouraged}}
+                ::  Recommended range: 30 seconds to 180 seconds.
+                ::  Default recommended value for user agents if {{PublicKeyCredentialCreationOptions/timeout}} is not specified: 120 seconds.
```
emlun: I think this could be reduced to just:
```suggestion
                ::  Recommended default value: 120 seconds.
```

index.bs
```diff
@@ -1388,8 +1388,17 @@ When this method is invoked, the user agent MUST execute the following algorithm
...
+
+                :   is set to {{UserVerificationRequirement/required}} or {{UserVerificationRequirement/preferred}}
+                ::  Recommended range: 30 seconds to 600 seconds.
+                ::  Default recommended value for user agents if {{PublicKeyCredentialCreationOptions/timeout}} is not specified: 300 seconds.
```
emlun: I think this could be reduced to just:
```suggestion
                ::  Recommended default value: 300 seconds.
```

index.bs
```diff
@@ -1814,8 +1823,17 @@ When this method is invoked, the user agent MUST execute the following algorithm
     Set a timer |lifetimeTimer| to this adjusted value. If the {{PublicKeyCredentialRequestOptions/timeout}} member of
     |options| is [=present|not present=], then set |lifetimeTimer| to a [=client=]-specific default.
 
-        Note: A suggested reasonable range for the {{PublicKeyCredentialCreationOptions/timeout}} member of |options| is
-        15 seconds to 120 seconds.
+      * Recommended ranges for the {{PublicKeyCredentialRequestOptions/timeout}} member of |options| are as follows.
```
emlun: Maybe?

```suggestion
      * Recommended ranges and defaults for the {{PublicKeyCredentialRequestOptions/timeout}} member of |options| are as follows.
```

index.bs
```diff
@@ -1814,8 +1823,17 @@ When this method is invoked, the user agent MUST execute the following algorithm
...
+          If <code>|options|.{{PublicKeyCredentialRequestOptions/userVerification}}</code>
+            <dl class="switch">
+                :   is set to {{UserVerificationRequirement/discouraged}}
+                ::  Recommended range: 30 seconds to 180 seconds.
+                ::  Default recommended value for user agents if {{PublicKeyCredentialRequestOptions/timeout}} is not specified: 120 seconds.
```
emlun: I think this could be reduced to just:
```suggestion
                ::  Recommended default value: 120 seconds.
```

index.bs
```diff
@@ -1814,8 +1823,17 @@ When this method is invoked, the user agent MUST execute the following algorithm
...
+
+                :   is set to {{UserVerificationRequirement/required}} or {{UserVerificationRequirement/preferred}}
+                ::  Recommended range: 30 seconds to 600 seconds.
+                ::  Default recommended value for user agents if {{PublicKeyCredentialRequestOptions/timeout}} is not specified: 300 seconds.
```
emlun: I think this could be reduced to just:

```suggestion
                ::  Recommended default value: 300 seconds.
```

index.bs
akshayku: Changed

akshayku:
Thanks @equalsJeffH and @emlun . Incorporated your comments and also fixing examples.

PR #1319, opened 2019-10-17
Update timeout examples to better agree with guidelines

### LINKED ISSUE #1318: update timeout values in examples (due to new timeout value recommendations) ?
examples 1, 3, 4 all give a timeout value of 1 minute (60000 ms) -- shall we update the examples to vary the timeout value per `options.authenticatorSelection.userVerification`'s setting per PR #1317 ?

emlun:
This would merge into PR #1317. Fixes #1318.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/pull/1319.html" title="Last updated on Oct 23, 2019, 8:03 PM UTC (235385c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/1319/fc62216...235385c.html" title="Last updated on Oct 23, 2019, 8:03 PM UTC (235385c)">Diff</a>


index.bs
```diff
@@ -523,7 +523,7 @@ The sample code for generating and registering a new key follows:
         }
       ],
 
-      timeout: 60000,  // 1 minute
+      timeout: 3600000,  // 6 minutes
```
akshayku: 300000, // 5 minutes

Had recommended 5 minutes as default in PR #1317

emlun: I know, but should we use the defaults as the example argument? It seems a bit strange to me since a better example of that would be to just leave the parameter out. I agree 5 minutes might be better for consistency, but I'm not sure which way to go on this...

akshayku: Actually you had 60 minutes in your example. :)
Either we leave the timeout or we add user verifcation required in above example. Both are fine with me.

emlun: > Actually you had 60 minutes in your example. :)

Oh, oops! Thanks! :smile:

Adding UV seems like a good idea.

PR #1322, opened 2019-10-17
revise draft-hodges-webauthn-registries per AD review and initiate work on rev -03

equalsJeffH:
This set of commits represent updates to draft-hodges-webauthn-registries per AD Ben Kaduk's review. the new target published revision number is -03 (the presently-published rev is -02), so this set of commits is referred to as -03a.  Target publication date for -03 is by 4-Nov-2019 (which is the Internet-Draft cut-off for IETF-106)


equalsJeffH:
I pushed another commit that fixes-up line-ending char, indentation, some section anchors.  I do not recommend trying to  review this, especially in github's diff. we ought to just merge this branch at this time, the draft-hodges-webauthn-registries{.xml,.txt,.html} files are all WIP (on rev -03) in any case. 

For review  by folks other than the editors of these files, they ought to see https://tools.ietf.org/html/draft-hodges-webauthn-registries which will be updated to -03 in due course (it is presently -02).

PR #1323, opened 2019-10-17
Create a way to reference the following figure

agl:
The fact that we have to keep the figure numbers in sync manually annoyed me. Bikeshed isn't helping out here so I don't see a way to avoid this for long-range links, but this change eliminates the need to manually maintain some of the absolute numbers by providing a way to insert the number of the following figure when the reference and figure are close by.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/pull/1323.html" title="Last updated on Oct 29, 2019, 11:00 PM UTC (0f6d96a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/1323/428bf82...0f6d96a.html" title="Last updated on Oct 29, 2019, 11:00 PM UTC (0f6d96a)">Diff</a>

PR #1324, opened 2019-10-18
Use CSS numbering for table references too

emlun:
This would merge into PR #1323.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/pull/1324.html" title="Last updated on Oct 23, 2019, 7:36 PM UTC (847ae92)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/1324/3f1f35a...847ae92.html" title="Last updated on Oct 23, 2019, 7:36 PM UTC (847ae92)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -86,7 +86,7 @@ figure.table figcaption {
     counter-increment: none;
 }
 figure.table figcaption:not(.no-marker)::before {
-    content: counter(table);
+    content: "Table " counter(table) " ";
 }
 figure.table .overlarge {
     max-width: 50em;
--- a/index.bs
+++ b/index.bs
@@ -101,6 +101,20 @@ figure.table .overlarge {
     counter-increment: figure -1;
     content: ""
 }
+
+.table-ref-previous::before {
+    content: "Table " counter(table)
+}
+
+.table-ref-following::before {
+    counter-increment: table;
+    content: "Table " counter(table)
+}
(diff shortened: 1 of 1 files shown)


emlun:
Yeah, good points, I noticed that too. I'll roll that part back.

index.bs
```diff
@@ -86,7 +86,7 @@ figure.table figcaption {
     counter-increment: none;
 }
 figure.table figcaption:not(.no-marker)::before {
-    content: "Table " counter(table) " ";
+    content: counter(table);
```
jcjones: Why strip the `Table ` portion? It looks a little odd to have these naked table numbers.
image

agl: (I think the "Table" might need to be put back in manually otherwise you can't search for it and nor can one choose the capitalisation at will.)

emlun: D'oh. Good catch, thanks!

emlun: Yes, searchability was a main motivation for this, but I accidentally removed the prefix from the `<figcaption>` rule as well. Probably in part because there's no corresponding rule for figures (`<figure>`s without `.table`), I guess `<figcaption>`s have a default "Figure N" prefix built in. So both before and after fixing this, the prefix in the `<figcaption>` isn't searchable, but the references are.

Issue #1297, opened 2019-09-10
Consider rewording restrictions around origin


sbweeden:
At present section https://w3c.github.io/webauthn/#relying-party-identifier states a restriction in the green Note box about origin scheme must be https. 

Whilst this is true when the RP is only used by browsers using WebAuthn, the platform authenticator of Android when used in a native app doesn't do this - it uses a different URI for origin and a hosted assetlinks.json file to bind this to the RPID.

Should the WebAuthn spec be inclusive of this type of use case, or at least mention that RPs MAY accept other origins for non-browser clients provided there is a trust mechanism in place to ensure the client validates that the origin is permitted by the RPs domain?

agl:
WebAuthn is a W3C spec and thus it should probably only deal with the web. You might be interested in https://github.com/fido-alliance/fido-2-specs/issues/682 however.

sbweeden:
I noticed in that issue:

_Credentials for these contexts will want to be separated such that they can never be used unexpectedly by the web._

In the Android case it is explicitly a goal that the credentials can be shared between the browser and native mobile contexts. That’s in part why I think it does relate to WebAuthn.

herrjemand:
@agl I am a little bit lost. I thought that WebAuthn is the document that defines server processing. And now we gonna have FIDO2 specs contains "exception" to the WebAuthn specs?

sbweeden:
@herrjemand there are other gaps in that space. For example FIDO2 conformance requires a server to be able to support use cases for UP=false whereas WebAuthn explicitly requires UP=true always. How would you propose to resolve that irregularity?

herrjemand:
@sbweeden That comes from authenticators supporting silent authentication, and we test it because this feature is explicitly in development.

sbweeden:
It is not mentioned in any normative server specification.

equalsJeffH:
As discussed in the webauthn f2f 20-Sep-2019 -- we probably ought to put a Note: in the RP assertion validation rules that points the RP dev's to whichever platform rules for assertion validation is appropriate, eg, Android, Windows, etc...

nadalin:
@sbweeden Please create PR per @equalsJeffH  note

sbweeden:
> As discussed in the webauthn f2f 20-Sep-2019 -- we probably ought to put a Note: in the RP assertion validation rules that points the RP dev's to whichever platform rules for assertion validation is appropriate, eg, Android, Windows, etc...

[...]

PR #1332, opened 2019-10-21
Update note in RP definition to indicate that non-WebAuthn FIDO clien…

sbweeden:
…ts MAY use origin values that are different than those specified for WebAuthn.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/pull/1332.html" title="Last updated on Oct 30, 2019, 1:02 AM UTC (a14e11d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/1332/a1d4e06...a14e11d.html" title="Last updated on Oct 30, 2019, 1:02 AM UTC (a14e11d)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -43,9 +43,9 @@ Former Editor: Angelo Liao, w3cid 94342, Microsoft, huliao@microsoft.com
 !Contributors: <a href="mailto:cbrand@google.com">Christiaan Brand</a> (Google)
 !Contributors: <a href="mailto:agl@google.com">Adam Langley</a> (Google)
 !Contributors: <a href="mailto:mandyam@qti.qualcomm.com">Giridhar Mandyam</a> (Qualcomm)
+!Contributors: <a href="mailto:sweeden@au1.ibm.com">Shane Weeden</a> (IBM)
 !Contributors: <a href="mailto:mkwst@google.com">Mike West</a> (Google)
 !Contributors: <a href="mailto:jyasskin@google.com">Jeffrey Yasskin</a> (Google)
-!Contributors: <a href="mailto:sweeden@au1.ibm.com">Shane Weeden</a> (IBM)
 group: webauthn
 Issue Tracking: GitHub https://github.com/w3c/webauthn/issues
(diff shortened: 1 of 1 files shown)


sbweeden:
Addresses issue #1297

index.bs
```diff
@@ -1082,6 +1082,8 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
         This is done in order to match the behavior of pervasively deployed ambient credentials (e.g., cookies, [[RFC6265]]).
         Please note that this is a greater relaxation of "same-origin" restrictions than what
         [=document.domain=]'s setter provides.
+
+        These restrictions on origin values apply to [=WebAuthn Clients=]. Other types of FIDO clients may use different origin URI formats, and have different rules for binding the origin URI to a [=Relying Party Identifier=], however the definition of these is outside the scope of this specification.
```
emlun: >Other types of FIDO clients

I don't think we need to call out FIDO specifically here - there's (in theory) nothing FIDO-specific about non-WebAuthn integrations.

>may use different origin URI formats, and have different rules for binding the origin URI to a [=Relying Party Identifier=]

It doesn't necessarily need to be URIs, does it? I'm not actually familiar with how mobile apps are identified in app stores, for example, but I wouldn't assume they always have to be URIs.

With both of those things in mind, how about something like this?

```suggestion
        These restrictions on origin values apply to [=WebAuthn Clients=]. If other specifications mimic the [=Web Authentication API=] to make WebAuthn [=credentials=] usable on other platforms, they might have different rules for binding a caller to a [=Relying Party Identifier=]; they are outside the scope of this specification.
```

agl: >  These restrictions on origin values

(nit: RP IDs aren't origins. Also, RP IDs as defined here aren't URIs, I think&mdash;although URI is pretty general so not totally sure about that.)

jcjones:
> I can go along with this pithy re-statement in terms of what we place in the WebAuthn spec proper (good job boiling it down @sbweeden ;-)
> 
> Though, a more full-featured treatment of this topic oughta be published somewhere... MDN?

MDN makes sense to me.

PR #1333, opened 2019-10-22
Add to sec cons a brief discussion of the sec properties accrued by authnr & client platform proximity

### LINKED ISSUE #1257: Add to sec cons a brief discussion of the sec properties accrued by authnr & client platform proximity
Add to sec cons a brief discussion of the sec properties accrued by authnr & client platform proximity

emlun:
How about something like this?

Fixes #1257.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/pull/1333.html" title="Last updated on Jan 23, 2020, 11:34 AM UTC (91777f7)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/1333/88468ca...91777f7.html" title="Last updated on Jan 23, 2020, 11:34 AM UTC (91777f7)">Diff</a>


index.bs
```diff
@@ -6286,6 +6290,58 @@ the wrong [=credential ID=], or if an attacker intercepts and manipulates the [=
...
+it might be possible for a malicious actor to trigger (3) the [=authenticator=] prompt (2),
+receive the [=authentication assertion=] and use it to authenticate to the [=[RP]=] (4).
+
+In order to enforce the [=scope=] restrictions in this example,
+the [=client=] would need to authenticate itself to the authenticator service (1)
+before the authenticator service allows access to the user's [=authenticator=] (2).
+This would ensure that the communication with the [=authenticator=]
+is mediated by the user's [=client=] so that the [=client=] can enforce the [=credential=] [=scope=] restrictions.
+
+Implementers of similar protocols SHOULD ensure that the [=client=] can still enforce [=credential=] [=scope=] restrictions
```
equalsJeffH: s/Implementors/Designers/ ..?

index.bs
```diff
@@ -6286,6 +6290,58 @@ the wrong [=credential ID=], or if an attacker intercepts and manipulates the [=
...
+despite not communicating with the [=authenticator=] directly.
```
equalsJeffH: s/ directly. / directly, or they should establish the means to create subsequent secure channels between the client and authnr via some initial channel that is not vulnerable to MITM attacks. E.g., exchanging pre-shared keys between the client and authnr over a direct physical USB connection or similar. /

emlun: Nice, thanks! I rearranged the sentence a structure a bit to make it less ambiguous what left-hand-side the "via" refers to.

emlun:
Rebased onto master.

nadalin:
@akshayku Please review so we can merge this or move it to next WD

equalsJeffH:
AFAIK @agl doesn't have time in the next 30 min to review this, and it was my thought to have him do so, thus I've removed him from reviewer list, and suggest we land this -- can always revise if necessary in the run-up to WD-03, CR, etc.

nadalin:
@agl @leshi @akshayku  Please review

agl:
I think something like this is valuable but I'm not sure that the current working is quite capturing what I see as the essence. (Or, perhaps nobody else views the essence in the same way, in which case LGTM.)

As currently written, this text emphasises that client and RP enforcement of the RP ID is critical to security. Absolutely agree with this part. Then, in my mind the critical point is that the authenticator (which is the trusted device here, assuming that the phisher controls their own client machine) gets assurances by transmitting over a medium that has limited range. (Direct USB connections have the most limited range, but BLE is still local.)

This ensures that an attacker must have a subverted a device physically close to the authenticator, which is a much higher bar than if the authenticator is willing to communicate across the internet.

nickmooney:
I think @agl's concerns are partially addressed by the section that discusses what tradeoffs may be made if the client and the authenticator are not being discussed directly. I still see compelling use cases for authentications where a "proximal" transport isn't possible (and I would argue that proximity is a bandaid, not a guarantee).

It would be reasonable to add some discussion more directly of what the benefits of proximity are;  ultimately I think it's a choice that is best left up to implementors equipped with an understanding of the security tradeoffs involved rather than being mandated in any way.

emlun:
Thanks @agl, that's a very good point and I agree. I've rewritten the section (now much shorter!) with equal focus on both the physical proximity aspect and the direct communication aspect.

index.bs
```diff
@@ -5851,6 +5855,36 @@ the wrong [=credential ID=], or if an attacker intercepts and manipulates the [=
 (a.k.a., [=assertion=]), and thus the interaction would end in an error.
 
 
+## Physical Proximity between Client and Authenticator ## {#sctn-client-authenticator-proximity}
+
+In the WebAuthn [=authenticator model=], it is generally assumed that [=roaming authenticators=]
+are physically close to and communicate directly with the [=client=].
```
agl:
```suggestion
are physically close to, and communicate directly with, the [=client=].
```

equalsJeffH:
on 2020-02-19 call: akshay reports msft looking into using remote SKs to remote desktop and in the desktop do things via webauthn.  AGL reports that this use case is supported in chrome remote desktop. threat model remains that an attacker must compromise something nearby the user.  Ought to add something to client data when operating in such a scenario. Ie. this PRs language does not preclude remote desktop use cases.

PR #1335, opened 2019-10-24
name the protocol

equalsJeffH:
we note in the webauthn spec that the spec defines a protocol as well as an API, but do not name the  protocol. How about the "WebAuthn/FIDO2 protocol" ?

[this name is used in text I'm proposing in my review of PR #1332]

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/pull/1335.html" title="Last updated on Oct 24, 2019, 11:31 PM UTC (a3991a4)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/1335/5dbea68...a3991a4.html" title="Last updated on Oct 24, 2019, 11:31 PM UTC (a3991a4)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -322,7 +322,8 @@ Beyond that, the intended audiences for this document are the following main gro
 
 <div class="note">
     Note: Along with the [[#sctn-api|Web Authentication API]] itself, this specification defines a
-    request-response <em>cryptographic protocol</em>
+    request-response <em>cryptographic protocol</em>&mdash;the
+    <dfn export>WebAuthn/FIDO2 protocol</dfn>&mdash;
     between a [=[WRP]=] server and an [=authenticator=], where the [=[RP]=]'s request consists of a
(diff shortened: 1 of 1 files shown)


emlun:
Looks fair. This term would be the only occurrence of "FIDO2" throughout the whole document, though - maybe we should drop that? But on the other hand, it's arguably more correct that FIDO2 is the more generic protocol of which WebAuthn is one (partial) implementation...

equalsJeffH:
> it's arguably more correct that FIDO2 is the more generic protocol of which WebAuthn is one (partial) implementation

Yes, the WebAuthn work is a facet of the overall [FIDO2 project](https://fidoalliance.org/fido2/).

Why don't we land this and see if anyone throws a rod over it?

jcjones:
I'd rather it not be "FIDO2".

PR #1342, opened 2019-11-17
'loc' extension: cite permissions and geolocation specs

equalsJeffH:
The IETF Sec Area Director (AD) sponsoring [`draft-hodges-webauthn-registries`](https://tools.ietf.org/html/draft-hodges-webauthn-registries) recommended that we bolster the sec & priv considerations for the "loc" (location) extension in the webauthn spec, hence this PR.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/pull/1342.html" title="Last updated on Nov 27, 2019, 7:51 PM UTC (8aaa6ca)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/1342/8927216...8aaa6ca.html" title="Last updated on Nov 27, 2019, 7:51 PM UTC (8aaa6ca)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1516,7 +1516,7 @@ a numbered step. If outdented, it (today) is rendered as a bullet in the midst o
     : {{CollectedClientData/origin}}
     :: The [=ascii serialization of an origin|serialization of=] |callerOrigin|.
     : {{CollectedClientData/crossOrigin}}
-    :: The inverse of the value of the 
+    :: The inverse of the value of the
         {{PublicKeyCredential/[[Create]](origin, options, sameOriginWithAncestors)/sameOriginWithAncestors}}
         argument passed to this [=internal method=].
     : {{CollectedClientData/tokenBinding}}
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -5543,7 +5543,14 @@ This extension enables use of a user verification index.
 
 ## Location Extension (loc) ## {#sctn-location-extension}
 
-This extension provides the [=authenticator=]'s current location to the WebAuthn [=[WRP]=].
+This extension provides the [=authenticator=]'s current geographic location to the [=[WRP]=].
+
+Note: In almost all cases, this information also discloses the location of the user of the device,
+    thereby potentially compromising the user's privacy. A [=conforming user agent=] must provide a mechanism that protects the user's privacy and this mechanism
```
```suggestion
    thereby potentially compromising the user's privacy. A [=conforming user agent=] MUST provide a mechanism that protects the user's privacy and this mechanism
```
...?

index.bs
```diff
@@ -5543,7 +5543,14 @@ This extension enables use of a user verification index.
...
+    MUST ensure that no location information is made available through this API without the user's
+    express permission [[Permissions]]. Further, the security and privacy considerations
```
emlun: The location is encoded into the signed authenticator extension output, so it is not possible to mask it out of the response without breaking the signature. The client would have to strip the `loc` extension input from the request to prevent the authenticator to add the output in the first place; maybe we should call that out explicitly.

index.bs
equalsJeffH: done, thx

index.bs
equalsJeffH: hm, yep, thx, I took a stab at it in 8aaa6ca

PR #1354, opened 2019-12-11
Remove the requirement accept and store a 64-byte minimum length for a name member’s value

### LINKED ISSUE #1352: Remove requirement for Authenticators to store PublicKeyCredentialRpEntity name
Browsers don't display this value if it is returned, they are displaying RPID.  
The requirement for the authenticator to always accept and store a minimum 64byte string is not useful for authenticators with no internal display.

Some Authenticators with internal displays do use this value for display.  
The proposal is to still have it required for the RP to send but optional for the Authenticator to store. 

The test is in Section 5.4.1 last paragraph.

ve7jtb:
Remove the requirement accept and store a 64-byte minimum length for a name member’s value
Fixes #1352

***
<a href="https://pr-preview.s3.amazonaws.com/ve7jtb/webauthn/pull/1354.html" title="Last updated on Feb 19, 2020, 9:44 PM UTC (6955b9c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/1354/90aaad9...ve7jtb:6955b9c.html" title="Last updated on Feb 19, 2020, 9:44 PM UTC (6955b9c)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2448,7 +2448,7 @@ associated with or [=scoped=] to, respectively.
 
         When [=clients=], [=client platforms=], or [=authenticators=] display a {{PublicKeyCredentialEntity/name}}'s value, they should always use UI elements to provide a clear boundary around the displayed value, and not allow overflow into other elements [[css-overflow-3]].
 
-        Authenticators MAY truncate a {{PublicKeyCredentialEntity/name}} member's value so that it fits within 64 bytes. See [[#sctn-strings]] about truncation and other considerations.
+        Authenticators MAY truncate a {{PublicKeyCredentialEntity/name}} member's value so that it fits within 64 bytes, if the authenticator stores the value. See [[#sctn-strings]] about truncation and other considerations.
 </div>
 
 
--- a/index.bs
+++ b/index.bs
@@ -2448,8 +2448,7 @@ associated with or [=scoped=] to, respectively.
 
         When [=clients=], [=client platforms=], or [=authenticators=] display a {{PublicKeyCredentialEntity/name}}'s value, they should always use UI elements to provide a clear boundary around the displayed value, and not allow overflow into other elements [[css-overflow-3]].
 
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -2448,8 +2448,7 @@ associated with or [=scoped=] to, respectively.
 
         When [=clients=], [=client platforms=], or [=authenticators=] display a {{PublicKeyCredentialEntity/name}}'s value, they should always use UI elements to provide a clear boundary around the displayed value, and not allow overflow into other elements [[css-overflow-3]].
 
-        [=Authenticators=] MUST accept and store a 64-byte minimum length for a {{PublicKeyCredentialEntity/name}} member's
-        value. Authenticators MAY truncate a {{PublicKeyCredentialEntity/name}} member's value so that it fits within 64 bytes. See [[#sctn-strings]] about truncation and other considerations.
+        Authenticators MAY truncate a {{PublicKeyCredentialEntity/name}} member's value so that it fits within 64 bytes. See [[#sctn-strings]] about truncation and other considerations.
```
equalsJeffH: perhaps we ought to be explicit that authnrs MAY choose not to store the `{{PublicKeyCredentialEntity/name}} member's value` ?  I can go either way.

ve7jtb: There are other parameters that we are not forced to store and they are not explicitly called out. 
This was sort of an exception saying we had to store up to 64 bytes.   I don't think it needs more but can add something if people want.

equalsJeffH: adding "if the authenticator stores the value" seems reasonable, but as you note, not required given precedent.

agl: Since previous versions of the spec required this to be stored, I think it's sensible to emphasise that we've changed our minds in the wording.

equalsJeffH: ok, so is @ve7jtb going to add "if the authenticator stores the value" and then we can merge this?

ve7jtb: done

PR #1366, opened 2020-01-22
Add “enterprise” attestation type.

### LINKED ISSUE #1147: add notion of "enterprise" attestation
We are proposing adding another [`PublicKeyCredentialCreationOptions`](https://w3c.github.io/webauthn/#dictdef-publickeycredentialcreationoptions) parameter (for [`[[Create]]()`](https://w3c.github.io/webauthn/#createCredential)) such that an RP can indicate it wishes the authenticator to provide "enterprise attestation" which allows an authenticator to send back special attestation data (might even be individual attestation) .  This request will only be honored by the [client platform](https://w3c.github.io/webauthn/#client-platform) if there exists some enterprise policy stating: "yes, it's okay for this RPID".

See also:  fido-alliance/fido-2-specs#596

agl:
In controlled deployments, organisations may wish to tie specific
registrations back to individual authenticators. Obviously this has
privacy concerns and needs to be gated on local configuration, or
special configuration on the authenticator. However, as cloud services
are increasingly used, RP IDs are no longer neatly divided into
enterprise and consumer contexts, and the RP might _not_ wish to receive
the enterprise attestation when used in a consumer context.

This change adds a new level of attestation, “enterprise”, which allows
RPs to indicate when they would like to, possibly, receive an
attestation that may include uniquely identifying information. This
leaves “direct” with its current, less privacy-impacting meaning.

Fixes #1147
Improves  #1376

***
<a href="https://pr-preview.s3.amazonaws.com/agl/webauthn/pull/1366.html" title="Last updated on Apr 27, 2020, 8:47 PM UTC (a3c9262)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/1366/6349d24...agl:a3c9262.html" title="Last updated on Apr 27, 2020, 8:47 PM UTC (a3c9262)">Diff</a>


index.bs
```diff
@@ -1747,6 +1747,9 @@ a numbered step. If outdented, it (today) is rendered either as a bullet in the
 
                         :   "direct"
```
equalsJeffH: s/ "direct" / "direct" or "enterprise" /

index.bs
```diff
@@ -1747,6 +1747,9 @@ a numbered step. If outdented, it (today) is rendered either as a bullet in the
...
                         ::  Convey the [=authenticator=]'s [=AAGUID=] and [=attestation statement=], unaltered, to the [=[RP]=].
+
+                        :   "enterprise"
+                        ::  If user agent or authenticator configuration permits it for the requested [=RP ID=], signal to the authenticator that enterprise attestation is permitted and convey the resulting [=AAGUID=] and [=attestation statement=], unaltered, to the [=[RP]=]. If not permitted, act the same as if the value were "direct".
```
equalsJeffH: hm, the imperative "signal to the authenticator" belongs down in the explanation for the `AttestationConveyancePreference` enum because where it is now up in the `[[create]]()` alg, the authnr has _already been invoked_ and we're just  working with the data it has returned, we can't signal it there, it's already been done.

I suggest here we just say the "direct" and "enterprise" do that same thing (see above).

```suggestion
```

index.bs
```diff
@@ -2663,6 +2667,8 @@ during credential generation.
     ::  This value indicates that the [=[RP]=] wants to receive the [=attestation statement=] as generated by the
         [=authenticator=].
 
+    :   <dfn>enterprise</dfn>
+    ::  This value indicates that the [=[RP]=] wants to receive an [=attestation statement=] that may include uniquely identifying information. This is intended for controlled deployments within an enterprise where the organization wishes to tie registrations to specific authenticators. User agents MUST NOT provide such an attestation unless user agent or authenticator configuration permits it for the requested [=RP ID=]. If not permitted, the user agent SHOULD treat this value the same as `direct`.
```
equalsJeffH:
```suggestion
    ::  This value indicates that the [=[RP]=] wants to receive an [=attestation statement=] that may include uniquely identifying information. This is intended for controlled deployments within an enterprise where the organization wishes to tie registrations to specific authenticators. User agents MUST NOT provide such an attestation unless user agent or authenticator configuration permits it for the requested [=RP ID=]. If not permitted, the user agent SHOULD treat this value the same as `direct`. 
    
    If permitted, the user agent SHOULD signal to the authenticator (at [=#CreateCred-InvokeAuthnrMakeCredauthenticator|invocation time=]) that enterprise attestation is requested, and convey the resulting [=AAGUID=] and [=attestation statement=], unaltered, to the [=[RP]=]. 
```

emlun: >If not permitted, the user agent SHOULD treat this value the same as `direct`.

Hm, could this "SHOULD" (as opposed to MUST) open opportunities for leaking the user's identity? If I put some known credential IDs in `excludeCredentials`, set `attestation: "enterprise"`, and the client refuses but does not treat it like `direct`...

If the client immediately returns failure, the RP knows that enterprise attestation was refused. If the client proceeds to the authenticator's test of user presence, the RP as usual does not learn if the authenticator does or does not have one of the `excludeCredentials`.

So probably not an issue?

emlun: Sure this shouldn't be a new parameter of `authenticatorMakeCredential`?

index.bs

index.bs
agl: Good point, thanks. I've (maybe) fixed this change to signal at the correct time.

index.bs
agl: Maybe, but I don't think so? For example, what would a call with enterprise + none-attestation mean? I.e. I don't think the cross-product is meaning in several cases thus I'm casting this as an alternative.

agl: Have changed to a MUST because the SHOULD introduces unneeded ambiguity. I'm not actually sure whether fall back to &ldquo;direct&rdquo; or immediate rejection is actually the right behaviour. However, if we pick the former then we should firmly pick it.

agl:
> Overall LGTM, thanks. One necessary fixup since we're technically not origin-based, but RP ID-based. Also, in a [PR on `github.com/alg/webauthn:ep`](https://github.com/agl/webauthn/pull/4), I fixed-up a link error and clarified how `enterpriseAttestationPossible` is factored into the `authenticatorMakeCredential` operation. HTH

That PR was pulling in some other changes so I've just rebased it into this change. I believe everything should be included here now.

equalsJeffH:
on 2020-03-18 call:  @jcjones please re-review.  also @akshayku, please.  this should be ready to land AFAWK.

equalsJeffH:
on 2020-04-01 call: reviewers please review :)

index.bs
```diff
@@ -2663,6 +2680,10 @@ during credential generation.
     ::  This value indicates that the [=[RP]=] wants to receive the [=attestation statement=] as generated by the
         [=authenticator=].
 
+    :   <dfn>enterprise</dfn>
+    ::  This value indicates that the [=[RP]=] wants to receive an [=attestation statement=] that may include uniquely identifying information. This is intended for controlled deployments within an enterprise where the organization wishes to tie registrations to specific authenticators. User agents MUST NOT provide such an attestation unless user agent or authenticator configuration permits it for the requested [=RP ID=]. If not permitted, the user agent MUST treat this value the same as `direct`.
+
```
akshayku, line 2685: Does It happen before or after the request is sent to the authenticator? We will plan to ask for permission before. 

If it happens after, because platform does not know whether authenticator will return enterprise attestation or not, doesn't treating it as `direct` in case user does not permit it disallows clients/platforms to change attestation to `none` or something anyway?

Do we really need a MUST here?

agl:
(From call: I need to update this.)

index.bs
equalsJeffH: on 2020-04-15 call:  (see minutes)  at minimum, we're going to delete the "If not permitted, the user agent MUST treat this value the same as `direct`." sentence.

agl: Have deleted that sentence.

equalsJeffH:
@akshayku & @jcjones -- please review!

index.bs
```diff
@@ -2663,6 +2680,10 @@ during credential generation.
     ::  This value indicates that the [=[RP]=] wants to receive the [=attestation statement=] as generated by the
         [=authenticator=].
 
+    :   <dfn>enterprise</dfn>
+    ::  This value indicates that the [=[RP]=] wants to receive an [=attestation statement=] that may include uniquely identifying information. This is intended for controlled deployments within an enterprise where the organization wishes to tie registrations to specific authenticators. User agents MUST NOT provide such an attestation unless user agent or authenticator configuration permits it for the requested [=RP ID=].
```
jcjones, line 2684:
```suggestion
    ::  This value indicates that the [=[RP]=] wants to receive an [=attestation statement=] that may include uniquely identifying information. This is intended for controlled deployments within an enterprise where the organization wishes to tie registrations to specific authenticators. User agents MUST NOT provide such an attestation unless the user agent or authenticator configuration permits it for the requested [=RP ID=].
```

Usually we say `the user agent`, fwiw.

equalsJeffH:
All comments are resolved, are we ok to merge this now?

PR #1369, opened 2020-01-30
Add clearer recommendation on what to do with transport hints

### LINKED ISSUE #1368: Clarity on transport hints
John says that the wording about transport hints in the spec isn't very accurate now. We should tell RPs to store the hints with the credential, return them during getAssertion, and otherwise never look at or touch them.

emlun:
Fixes #1368.

With this we have

- A normative RECOMMENDation to store the transport hints in the registration operation
- A normative SHOULD to put them in as arguments to `get()`
- A mention of `getTransports()` in the [`PublicKeyCredentialDescriptor.transports`](https://w3c.github.io/webauthn/#dom-publickeycredentialdescriptor-transports) description
- A mention of `getTransports()` in the [`AuthenticatorTransport`](https://w3c.github.io/webauthn/#enumdef-authenticatortransport) description

Should we add hooks anywhere else too?

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/pull/1369.html" title="Last updated on Feb 27, 2020, 12:33 PM UTC (eb02738)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/1369/66ad76b...eb02738.html" title="Last updated on Feb 27, 2020, 12:33 PM UTC (eb02738)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2904,7 +2904,13 @@ parameter to the {{CredentialsContainer/create()}} or {{CredentialsContainer/get
     :   <dfn>transports</dfn>
     ::  This OPTIONAL member contains a hint as to how the [=client=] might communicate with the [=managing authenticator=] of the
         [=public key credential=] the caller is referring to. The values SHOULD be members of {{AuthenticatorTransport}} but [=client platforms=] MUST ignore unknown values.
+
         The {{AuthenticatorAttestationResponse/getTransports()}} operation can provide suitable values for this member.
+        When [[#sctn-registering-a-new-credential|registering a new credential]],
+        the [=[RP]=] SHOULD store the value returned from {{AuthenticatorAttestationResponse/getTransports()}}.
+        When creating a {{PublicKeyCredentialDescriptor}} for that credential,
+        the [=[RP]=] SHOULD retrieve that stored value
+        and set it as the {{PublicKeyCredentialDescriptor/transports}} member.
 </div>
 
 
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -2904,7 +2904,13 @@ parameter to the {{CredentialsContainer/create()}} or {{CredentialsContainer/get
     :   <dfn>transports</dfn>
     ::  This OPTIONAL member contains a hint as to how the [=client=] might communicate with the [=managing authenticator=] of the
         [=public key credential=] the caller is referring to. The values SHOULD be members of {{AuthenticatorTransport}} but [=client platforms=] MUST ignore unknown values.
+
         The {{AuthenticatorAttestationResponse/getTransports()}} operation can provide suitable values for this member.
+        When [[#sctn-registering-a-new-credential|registering a new credential]],
+        the [=[RP]=] SHOULD store the value returned from {{AuthenticatorAttestationResponse/getTransports()}}.
+        When creating a {{PublicKeyCredentialDescriptor}} for that credential,
+        the [=[RP]=] SHOULD retrieve that stored value
+        and set it as the {{PublicKeyCredentialDescriptor/transports}} member.
```
equalsJeffH: s/ and set it as the / and set it as the value of the /  ..?

arshadnoor:
You might want to consider adding a note that not all RPs will have to deal with this - only those who choose to create their own FIDO2 Servers. Third-party commercial and open-source FIDO2 Servers are more than likely capable of handling the transport hints, thus allowing RPs to focus on their business application rather than protocol semantics.

emlun:
Thanks, but I don't think such a distinction is necessary to have in the spec. The party making the JavaScript calls is the one that will need to call `getTransports()` and know what to do with the result, even if what they need to do is offload it to a separate third-party server.

PR #1375, opened 2020-02-24
Add more requirements for ClientDataJSON serialisation.

agl:
ClientDataJSON is currently defined to be the JSON encoding of the
CollectedClientData. This implies that validators require a full JSON
parsing library to check needed entries in the ClientDataJSON such as
the challenge, type, and origin.

This is a problematic dependency in some cases. This change seeks to
address that by being stricter about the encoding, while still
generating JSON. Thus existing validators do not need to change but
those Relying Parties' webapps that are willing to require the use recent 
WebAuthn-implementing browsers can avoid the full generality of JSON.

***
<a href="https://pr-preview.s3.amazonaws.com/agl/webauthn/pull/1375.html" title="Last updated on Apr 29, 2020, 6:40 PM UTC (d4300ee)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/1375/dbff4e5...agl:d4300ee.html" title="Last updated on Apr 29, 2020, 6:40 PM UTC (d4300ee)">Diff</a>


index.bs
```diff
@@ -2857,14 +2857,48 @@ Note: The {{CollectedClientData}} may be extended in the future. Therefore it's
...
+    1. Append &lsquo;`false`&rsquo;.
+1. Otherwise:
+    1. Append &lsquo;`true`&rsquo;.
+1. Create a temporary copy of the {{CollectedClientData}} and remove the fields `type`, `challenge`, `origin`, and `crossOrigin` (if present). Invoke [=serialize JSON to bytes=] on it to produce a byte string `remainder`.
+1. If `remainder` is equal to 0x7b7d (i.e. "{}") then:
+    1. Append &lsquo;`}`&rsquo;.
+1. Otherwise:
+    1. Append &lsquo;`,`&rsquo;.
+    1. Remove the leading byte from `remainder`.
+    1. Append `remainder` to the partial result.
```
emlun: The use of single quotes for string literals is inconsistent with other parts of the spec (search for "the string" in the doc). For example:

>***type*, of type DOMString**
>
>This member contains the string "webauthn.create" when creating new credentials, and "webauthn.get" when getting an assertion from an existing credential. [...]

Admittedly there are already some preexisting inconsistencies we should sort out independently of this, but double-quotes is most common right now.

I realize that double quotes would be cumbersome here since those also appear in the values. Maybe we can make a special case for this section, but then we might need to define the syntax.

index.bs
```diff
@@ -2857,14 +2857,48 @@ Note: The {{CollectedClientData}} may be extended in the future. Therefore it's
...
+1. Append [=CCDToString=]({{CollectedClientData/challenge}}).
+1. Append &lsquo;`,"origin":`&rsquo;.
+1. Append [=CCDToString=]({{CollectedClientData/origin}}).
+1. Append &lsquo;`,"crossOrigin":`&rsquo;.
+1. If {{CollectedClientData/crossOrigin}} is not included, or is `false`:
+    1. Append &lsquo;`false`&rsquo;.
+1. Otherwise:
+    1. Append &lsquo;`true`&rsquo;.
+1. Create a temporary copy of the {{CollectedClientData}} and remove the fields `type`, `challenge`, `origin`, and `crossOrigin` (if present). Invoke [=serialize JSON to bytes=] on it to produce a byte string `remainder`.
+1. If `remainder` is equal to 0x7b7d (i.e. "{}") then:
```
```suggestion
1. If |remainder| is equal to 0x7b7d (i.e., "{}") then:
```

index.bs
```diff
@@ -2857,14 +2857,48 @@ Note: The {{CollectedClientData}} may be extended in the future. Therefore it's
...
+1. Append &lsquo;`,"challenge":`&rsquo;.
+1. Append [=CCDToString=]({{CollectedClientData/challenge}}).
+1. Append &lsquo;`,"origin":`&rsquo;.
+1. Append [=CCDToString=]({{CollectedClientData/origin}}).
+1. Append &lsquo;`,"crossOrigin":`&rsquo;.
+1. If {{CollectedClientData/crossOrigin}} is not included, or is `false`:
+    1. Append &lsquo;`false`&rsquo;.
+1. Otherwise:
+    1. Append &lsquo;`true`&rsquo;.
+1. Create a temporary copy of the {{CollectedClientData}} and remove the fields `type`, `challenge`, `origin`, and `crossOrigin` (if present). Invoke [=serialize JSON to bytes=] on it to produce a byte string `remainder`.
```
```suggestion
1. Create a temporary copy of the {{CollectedClientData}} and remove the fields {{CollectedClientData/type}}, {{CollectedClientData/challenge}}, {{CollectedClientData/origin}}, and {{CollectedClientData/crossOrigin}} (if present). Invoke [=serialize JSON to bytes=] on it to produce a byte string |remainder|.
```

index.bs
```diff
@@ -2857,14 +2857,48 @@ Note: The {{CollectedClientData}} may be extended in the future. Therefore it's
...
+1. If `remainder` is equal to 0x7b7d (i.e. "{}") then:
+    1. Append &lsquo;`}`&rsquo;.
+1. Otherwise:
+    1. Append &lsquo;`,`&rsquo;.
+    1. Remove the leading byte from `remainder`.
```
```suggestion
    1. Remove the leading byte from |remainder|.
```

index.bs
```suggestion
    1. Append |remainder| to the partial result.
```

index.bs
```diff
@@ -2857,14 +2857,48 @@ Note: The {{CollectedClientData}} may be extended in the future. Therefore it's
...
+    1. Append `remainder` to the partial result.
+
+The function <dfn>CCDToString</dfn> is used in the above algorithm and is defined as:
```
equalsJeffH: this function seems to have an implied accumulator variable that perhaps oughta be explicit.

index.bs
```diff
@@ -2857,14 +2857,48 @@ Note: The {{CollectedClientData}} may be extended in the future. Therefore it's
...
+    :: This is the result of performing the [JSON-compatible serialization algorithm](#clientdatajson-serialization) on the {{CollectedClientData}} dictionary.
 
     : <dfn dfn>Hash of the serialized client data</dfn>
-    :: This is the hash (computed using SHA-256) of the [=JSON-serialized client data=], as constructed by the client.
+    :: This is the hash (computed using SHA-256) of the [=JSON-compatible serialization of client data=], as constructed by the client.
 </div>
 
+#### Serialization #### {#clientdatajson-serialization}
+
+The serialization of the {{CollectedClientData}} is a subset of the algorithm for [=serialize JSON to bytes|JSON-serializing to bytes=]. I.e. it produces a valid JSON encoding of the {{CollectedClientData}} but also provides additional structure that may be exploited by verifiers to avoid integrating a full JSON parser.
```
equalsJeffH: what explicitly is the "additional structure" and how might it help a verifier?

I'd s/ JSON encoding / JSON serialization /

index.bs
```diff
@@ -2857,14 +2857,48 @@ Note: The {{CollectedClientData}} may be extended in the future. Therefore it's
...
+    1. Append &lsquo;`,`&rsquo;.
+    1. Remove the leading byte from `remainder`.
+    1. Append `remainder` to the partial result.
+
+The function <dfn>CCDToString</dfn> is used in the above algorithm and is defined as:
+
+1. Append &lsquo;`"`&rsquo;.
+1. Invoke [ToString](https://tc39.es/ecma262/#sec-tostring) on the given object to convert to a string.
+1. For each code point in the resulting string:
+    1. If the code point is in the set {0x20, 0x21, 0x23&ndash;0x5b, 0x5d&ndash;0x10ffff} then append the UTF-8 encoding of that code point.
```
equalsJeffH: since these are  code points, ought they be shown in code point syntax, eg: U+0020, U+0021, U+0023&ndash;U+005B, U+005Dndash;U+10ffff ?

index.bs
emlun: Explicitly defined order of keys, I think. Allows for fewer loops and conditional branches.

index.bs
emlun: Yes, that's what I mean.

equalsJeffH:
in 2020-02-26 meeting, feedback is that this needs further Note:/comments targeted at implementers that:
* full JSON parser is preferred impl
* add a validators' algorithm for those who wish to take advantage of the special sauce this limited & deterministic encoding provides
    * this alg needs to be crafted such that futue extensions to clientDataJSON  client platforms may employ will not break existing 
* add further explanation of the "hooks" this provides, eg fixed ordering of the fixed four key-value pairs, etc.
* explicitly state that these four fields are fixed forever, we can never change them in future spec versions
* explain the consequences of any future fifth (or more) key-value pairs added to the spec in terms of new browsers and RPs --- (consideration: overall in webauthn we're commiting to not make bkwards incompatible changes)
* perhaps more to add here....

equalsJeffH:
On 2020-03-11 call @agl reports that he has tugged the OpenSSH folks' sleeves and they are thinking about this..... more news later....

index.bs
agl: Have switched to writing out the hex strings explicitly, with illustrative strings following to aid understanding.

index.bs
agl: (Mooted: changed how this is detected because I believe a valid JSON encoder could add whitespace.)

index.bs

index.bs

index.bs
```diff
@@ -2251,7 +2251,7 @@ Note: Invoking this method from a [=browsing context=] where the [=Web Authentic
 </xmp>
 <div dfn-type="attribute" dfn-for="AuthenticatorResponse">
     :   <dfn>clientDataJSON</dfn>
-    ::  This attribute contains a [=JSON-serialized client data|JSON serialization=] of the [=client data=] passed to the
+    ::  This attribute contains a [JSON-compatible serialization](#clientdatajson-serialization) of the [=client data=] passed to the
```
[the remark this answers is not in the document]

index.bs
```diff
@@ -2857,14 +2857,48 @@ Note: The {{CollectedClientData}} may be extended in the future. Therefore it's
...
     : <dfn dfn>Hash of the serialized client data</dfn>
-    :: This is the hash (computed using SHA-256) of the [=JSON-serialized client data=], as constructed by the client.
+    :: This is the hash (computed using SHA-256) of the [=JSON-compatible serialization of client data=], as constructed by the client.
 </div>
 
+#### Serialization #### {#clientdatajson-serialization}
+
+The serialization of the {{CollectedClientData}} is a subset of the algorithm for [=serialize JSON to bytes|JSON-serializing to bytes=]. I.e. it produces a valid JSON encoding of the {{CollectedClientData}} but also provides additional structure that may be exploited by verifiers to avoid integrating a full JSON parser.
+
+The serialization algorithm works by appending successive byte strings to an, initially empty, partial result until the complete result is obtained. If a step in the following algorithm suggests appending a string, then the byte string to append is formed by [UTF-8 encoding](https://encoding.spec.whatwg.org/#utf-8-encode) that string.
```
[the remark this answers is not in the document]
agl: Have made accumulator values explicit.

index.bs
agl: Have enumerated the requirements for the limited verification algorithm.

index.bs
agl: Yep, done. Thanks.

index.bs
agl: Have made accumulator values explicit.

index.bs

equalsJeffH:
on 2020-03-18 call, @agl reports SSH people are chewing on it, still on hold.

equalsJeffH:
on 2020-04-08 call: @agl reports that ssh folk think this is ok.

agl:
This is now good to review; we believe that it all works in practice.

index.bs
```diff
@@ -2857,14 +2857,105 @@ Note: The {{CollectedClientData}} may be extended in the future. Therefore it's
...
+        : is 0x2c
+        :: The verification is successful.
+
+        : otherwise
+        :: The verification has failed.
+    </dl>
+
+#### Future development #### {#clientdatajson-development}
+
+In order to remain compatible with the [limited verification algorithm](#clientdatajson-verification), future versions of this specification may not remove any of the fields {{CollectedClientData/type}}, {{CollectedClientData/challenge}}, {{CollectedClientData/origin}}, or {{CollectedClientData/crossOrigin}} from {{CollectedClientData}}. They may also not change the [serialization algorithm](#clientdatajson-verification) to change the order in which those fields are serialized.
```
equalsJeffH, line 2956:
```suggestion
In order to remain compatible with the [limited verification algorithm](#clientdatajson-verification), future versions of this specification must not remove any of the fields {{CollectedClientData/type}}, {{CollectedClientData/challenge}}, {{CollectedClientData/origin}}, or {{CollectedClientData/crossOrigin}} from {{CollectedClientData}}. They also must not change the [serialization algorithm](#clientdatajson-verification) to change the order in which those fields are serialized.
```

index.bs
```diff
@@ -2857,14 +2857,105 @@ Note: The {{CollectedClientData}} may be extended in the future. Therefore it's
...
+        : is U+005C
+        :: Append 0x5c5c (<kbd>\\</kbd>) to |encoded|.
+
+        : otherwise
+        :: Append 0x5c75 (`\u`) to |encoded|, followed by four, lower-case hex digits that, when interpreted as a base-16 number, represent that code point.
+    </dl>
+1. Append 0x22 (`"`) to |encoded|.
+1. The result of this function is the value of |encoded|.
+
+#### Verification #### {#clientdatajson-verification}
```
equalsJeffH, line 2917:
```suggestion
#### Limited Verification Algorithm #### {#clientdatajson-verification}
```

quartzjer:
As an outsider to this process (though one with some protocol design experience), this entire proposal/workaround seems deeply dangerous at worst, and an ugly wart at best.

There should be _no_ justification for not requiring a standard JSON serializer and parser, that is just asking for unknown and unforeseen trouble in a spec.  JSON is used in all manners of security sensitive and even constrained protocols.

Asking implementors to do *manual* serialization of part of a JSON document is almost guaranteed to result in one or more vulnerabilities.  I’m disappointed that anyone from SSH is even asking for or expecting this kind of compromise just for their benefit.  There are many formal and audited codebases that include JSON parsers, and even some parsers that will tokenize side-effect free without any memory allocations/operations.

It took a lot for me to decide to say something here after this was pointed out to me since I’ve not participated in any of these processes and don’t have a full background, so I’ll understand if I’m off track.  What I’ve read here though was shocking enough that I felt compelled to speak just in case I’m not the only observer who has these concerns and hope that the whole spec isn’t cobbled by a single particularly stubborn use-case.

dwaite:
Do we also need to make comments around verification of extensions and attestations, given that an implementation without JSON is somewhat likely to also not have an implementation of CBOR?

agl:
> Do we also need to make comments around verification of extensions and attestations, given that an implementation without JSON is somewhat likely to also not have an implementation of CBOR?

The differences stem from different sensitivities of different bits of code. Registering a new authenticator (during which attestation processing may be done) is usually less sensitive than pre-auth verification of a signature. Extension processing could be a problem, although the CBOR used is already significantly subsetted by CTAP2 and currently no authentication extensions are commonly used.

> There should be no justification for not requiring a standard JSON serializer and parser.

I roughly agree that JSON is not an unreasonably complex format, even in sensitive locations. I wrote a, I think, demonstrably memory-safe JSON parser as an example of how it could be done. However, it's not my decision and I must admit that, compared to SSH's length-prefixed bytestrings, JSON _is_ a lot more heavyweight and would significantly increase the amount of code in the highly-sensitive preauth flow.

(For those looking for a safe JSON parser in C, I suggest looking at the [Wuffs](https://github.com/google/wuffs) framework, on which the Chromium's future JSON parser may well be based.)

> Asking implementors to do manual serialization of part of a JSON document is almost guaranteed to result in one or more vulnerabilities.

I disagree here. While things could get hairy in general, the input strings here are either static values from the browser, base64url-encoded values, or punycoded origin strings. Of these, only the last is not trivially safe for plain template insertion. This is not a general data structure that's being encoded, it has a few, known keys and values. Lastly, there are few browser implementations and they have to do much more complex and subtle things to ensure safety in WebAuthn than this.

It's worth emphasing that no site needs to worry about this, and are recommended not to worry about this. No site need touch anything other than a standard JSON parser.

equalsJeffH:
Good catch @jcjones -- added in d4300ee -- thx!

PR #1387, opened 2020-03-11
Default to ES256 and RS256 if pubKeyCredParams is empty

### LINKED ISSUE #1383: `credentials.create` should default to ES256 if `options.pubKeyCredParams` is empty
On `[[Create]]` method, step 9:

> if credTypesAndPubKeyAlgs is empty and options.pubKeyCredParams is not empty, return a DOMException whose name is "NotSupportedError", and terminate this algorithm.

"NotSupportedError" should be returned if `credTypesAndPubKeyAlgs` is empty regardless of whether `options.pubKeyCredParams` is empty or not. This is because an empty list will fail when sent to authenticators anyway (step 2 of the authenticatorMakeCredential operation), so there is no point in sending the empty list.

At the moment chrome is immediately throwing `NotSupportedError` (which doesn't match the spec, it should wait for the user to tap an authenticator and return `NotAllowedError` instead) and firefox is defaulting to `{ type: "public-key", alg: -7 }`

nsatragno:
Default to algorithm -7 ("ES256") when `options.pubKeyCredParams` is
empty.

Fixes #1383.

***
<a href="https://pr-preview.s3.amazonaws.com/nsatragno/webauthn/pull/1387.html" title="Last updated on Mar 25, 2020, 7:46 PM UTC (2feeac6)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/1387/a0249af...nsatragno:2feeac6.html" title="Last updated on Mar 25, 2020, 7:46 PM UTC (2feeac6)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1490,18 +1490,21 @@ a numbered step. If outdented, it (today) is rendered as a bullet in the midst o
 1. Let |credTypesAndPubKeyAlgs| be a new [=list=] whose [=list/items=] are pairs of {{PublicKeyCredentialType}} and
     a {{COSEAlgorithmIdentifier}}.
 
-1. If <code>|options|.{{PublicKeyCredentialCreationOptions/pubKeyCredParams}}</code> [=list/is empty=]:
+1. If <code>|options|.{{PublicKeyCredentialCreationOptions/pubKeyCredParams}}</code>'s [=list/size=]
+    <dl class="switch">
+        :   is zero
+        ::  [=list/Append=] the pair of {{PublicKeyCredentialType}} with value {{public-key}} and a
+            {{COSEAlgorithmIdentifier}} with value <code>-7</code> ("ES256") to |credTypesAndPubKeyAlgs|.
 
-    1. [=list/Append=] the pair of {{PublicKeyCredentialType}} with value {{public-key}} and a
-        {{COSEAlgorithmIdentifier}} with value <code>-7</code> ("ES256") to |credTypesAndPubKeyAlgs|.
+        :   is non-zero
+        ::  [=list/for each=] |current| of <code>|options|.{{PublicKeyCredentialCreationOptions/pubKeyCredParams}}</code>:
 
-1. Otherwise, [=list/for each=] |current| of <code>|options|.{{PublicKeyCredentialCreationOptions/pubKeyCredParams}}</code>:
-
-    1. If <code>|current|.{{PublicKeyCredentialParameters/type}}</code> does not contain a {{PublicKeyCredentialType}} supported
-        by this implementation, then [=continue=].
-    1. Let |alg| be <code>|current|.{{PublicKeyCredentialParameters/alg}}</code>.
-    1. [=list/Append=] the pair of <code>|current|.{{PublicKeyCredentialParameters/type}}</code> and |alg| to
-        |credTypesAndPubKeyAlgs|.
+            1. If <code>|current|.{{PublicKeyCredentialParameters/type}}</code> does not contain a {{PublicKeyCredentialType}} supported
+                by this implementation, then [=continue=].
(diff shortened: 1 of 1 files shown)


nsatragno:
@equalsJeffH, please take a look. Thanks!

index.bs
```diff
@@ -1490,16 +1490,21 @@ a numbered step. If outdented, it (today) is rendered as a bullet in the midst o
...
     1. If <code>|current|.{{PublicKeyCredentialParameters/type}}</code> does not contain a {{PublicKeyCredentialType}} supported
         by this implementation, then [=continue=].
     1. Let |alg| be <code>|current|.{{PublicKeyCredentialParameters/alg}}</code>.
     1. [=list/Append=] the pair of <code>|current|.{{PublicKeyCredentialParameters/type}}</code> and |alg| to
         |credTypesAndPubKeyAlgs|.
 
-1. If |credTypesAndPubKeyAlgs| [=list/is empty=] and <code>|options|.{{PublicKeyCredentialCreationOptions/pubKeyCredParams}}</code>
-    [=list/is not empty=], return a {{DOMException}} whose name is "{{NotSupportedError}}", and terminate this algorithm.
+1. If |credTypesAndPubKeyAlgs| [=list/is empty=], return a {{DOMException}} whose name is
+    "{{NotSupportedError}}", and terminate this algorithm.
```
equalsJeffH: Is this step even needed any longer?  ISTM that `credTypesAndPubKeyAlgs` will now never be empty ...?

nsatragno: It's necessary. It can still be empty if `pubKeyCredParams` is not empty and none of its algorithms are supported.

nsatragno: What I'm thinking now is that perhaps it's better to move it to the "is non-zero" branch of the switch.

equalsJeffH: > It's necessary. It can still be empty if `pubKeyCredParams` is not empty and none of its algorithms are supported.

Right. doh.

equalsJeffH: fyi/fwiw, I did not obviously see support for `[=list/is not empty=]` in infra.spec.whatwg.org (but should have looked at index.bs more closely because that construct is in there (doh)), hence suggesting leveraging the list's `size` for the switch.  Either works so this is fine but would not object to using `is (not) empty` instead.  no big deal either way.

nsatragno:
@akshayku updated the PR to include a list of algorithms client platforms can pick (ECC and RSA), please take another look.

index.bs
```diff
@@ -1490,16 +1490,29 @@ a numbered step. If outdented, it (today) is rendered as a bullet in the midst o
...
+
+            1. If <code>|current|.{{PublicKeyCredentialParameters/type}}</code> does not contain a {{PublicKeyCredentialType}} supported
+                by this implementation, then [=continue=].
+            1. Let |alg| be <code>|current|.{{PublicKeyCredentialParameters/alg}}</code>.
+            1. [=list/Append=] the pair of <code>|current|.{{PublicKeyCredentialParameters/type}}</code> and |alg| to
+                |credTypesAndPubKeyAlgs|.
+
+            If |credTypesAndPubKeyAlgs| [=list/is empty=], return a {{DOMException}} whose name is
+               "{{NotSupportedError}}", and terminate this algorithm.
+    </dl>
```
akshayku: Not asking that client chooses one of the algorithm at its descretion. Would like the list be of two elements. 

We just need to add one more step in existing flow. `If <code>|options|.{{PublicKeyCredentialCreationOptions/pubKeyCredParams}}</code> is empty, append <code>-7</code> ("ES256") AND <code>-257</code> ("RS256") to the list.`

nsatragno: Oh I get you now. Updated the PR to append both `-7` and `-257` to the list.

Issue #1376, opened 2020-02-26
Inputs to WebAuthn methods should not include strict Enum types


jcjones:
We need to audit all enums passed into WebAuthn to ensure that WebAuthn can be extended without causing WebIDL failures for lagging implementations.

What we did for Transport was to define that the input values be DOMString and coerced when possible into the acceptable enum values. At the 26-02-2020 face-to-face we agreed that in general, enums input into WebAuthn, if unknown, be treated as if the value was not set (e.g., the default value).

This affects #1366.

equalsJeffH:
on 2020-03-11 call:  we just noted the verbal agreement we had that @agl to address the [`AttestationConveyancePreference`](https://www.w3.org/TR/webauthn-2/#enum-attestation-convey) enum as a part of completing PR #1366, and then @jcjones was intending to address the remaining enums in the spec.

equalsJeffH:
Note: PR #1366 now addresses changing `PublicKeyCredentialCreationOptions.attestation` to a `DOMString`.

PR #1392, opened 2020-03-24
Substitute enum types in dictionaries with DOMStrings

jcjones:
This fixes issue #1376, auditing all enums passed into WebAuthn to ensure that WebAuthn can be extended without causing WebIDL failures for lagging implementations.

I believe I caught them all. I prefixed all the targets with `>>` in the list below:

```
dictionary PublicKeyCredentialCreationOptions {
    required PublicKeyCredentialRpEntity         rp;
    required PublicKeyCredentialUserEntity       user;

    required BufferSource                             challenge;
    required sequence<PublicKeyCredentialParameters>  pubKeyCredParams;

    unsigned long                                timeout;
    sequence<PublicKeyCredentialDescriptor>      excludeCredentials = [];
    AuthenticatorSelectionCriteria               authenticatorSelection;
>>    AttestationConveyancePreference              attestation = "none";
    AuthenticationExtensionsClientInputs         extensions;
};


dictionary PublicKeyCredentialRequestOptions {
    required BufferSource                challenge;
>>    UserVerificationRequirement          userVerification = "preferred";
    AuthenticationExtensionsClientInputs extensions;
};

>>    AuthenticatorAttachment      authenticatorAttachment;
>>    UserVerificationRequirement  userVerification = "preferred";
};

dictionary PublicKeyCredentialDescriptor {
>>    required PublicKeyCredentialType      type;
    required BufferSource                 id;
    sequence<DOMString>                   transports;
};

dictionary TokenBinding {
>>    required TokenBindingStatus status;
    DOMString id;
};

dictionary PublicKeyCredentialParameters {
>>    required PublicKeyCredentialType      type;
    required COSEAlgorithmIdentifier      alg;
};
```

***
<a href="https://pr-preview.s3.amazonaws.com/jcjones/webauthn/pull/1392.html" title="Last updated on May 15, 2020, 6:04 PM UTC (42dbe07)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/1392/d530669...jcjones:42dbe07.html" title="Last updated on May 15, 2020, 6:04 PM UTC (42dbe07)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2689,7 +2689,7 @@ an assertion. Its {{PublicKeyCredentialRequestOptions/challenge}} member MUST be
         unsigned long                        timeout;
         USVString                            rpId;
         sequence<PublicKeyCredentialDescriptor> allowCredentials = [];
-        UserVerificationRequirement          userVerification = "preferred";
+        DOMString                            userVerification = "preferred";
         AuthenticationExtensionsClientInputs extensions;
     };
 </xmp>
@@ -2715,8 +2715,7 @@ an assertion. Its {{PublicKeyCredentialRequestOptions/challenge}} member MUST be
 
     :   <dfn>userVerification</dfn>
     ::  This OPTIONAL member describes the [=[RP]=]'s requirements regarding [=user verification=] for the
-        {{CredentialsContainer/get()}} operation. Eligible authenticators are filtered to only those capable of satisfying this
-        requirement.
+        {{CredentialsContainer/get()}} operation. The value SHOULD be a member of {{UserVerificationRequirement}} but [=client platforms=] MUST ignore unknown values, treating an unknown value as if the member was unset. Eligible authenticators are filtered to only those capable of satisfying this requirement.
 
     :   <dfn>extensions</dfn>
     ::  This OPTIONAL member contains additional parameters requesting additional processing by the client and authenticator.
--- a/index.bs
+++ b/index.bs
@@ -2371,7 +2371,7 @@ optionally evidence of [=user consent=] to a specific transaction.
         unsigned long                                timeout;
         sequence<PublicKeyCredentialDescriptor>      excludeCredentials = [];
         AuthenticatorSelectionCriteria               authenticatorSelection;
-        AttestationConveyancePreference              attestation = "none";
+        DOMString                                    attestation = "none";
         AuthenticationExtensionsClientInputs         extensions;
     };
 </xmp>
@@ -2418,7 +2418,7 @@ optionally evidence of [=user consent=] to a specific transaction.
 
     :   <dfn>attestation</dfn>
     ::  This member is intended for use by [=[RPS]=] that wish to express their preference for [=attestation conveyance=].
-        The default is {{AttestationConveyancePreference/none}}.
+        The value SHOULD be a member of {{AttestationConveyancePreference}} but [=client platforms=] MUST ignore unknown values, treating an unknown value as if the member was unset. The default is {{AttestationConveyancePreference/none}}.
 
     :   <dfn>extensions</dfn>
     ::  This member contains additional parameters requesting additional processing by the client and authenticator. For
@@ -2915,7 +2915,7 @@ parameter to the {{CredentialsContainer/create()}} or {{CredentialsContainer/get
 
     :   <dfn>transports</dfn>
     ::  This OPTIONAL member contains a hint as to how the [=client=] might communicate with the [=managing authenticator=] of the
-        [=public key credential=] the caller is referring to. The values SHOULD be members of {{AuthenticatorTransport}} but [=client platforms=] MUST ignore unknown values.
+        [=public key credential=] the caller is referring to. The values SHOULD be members of {{AuthenticatorTransport}} but [=client platforms=] MUST ignore unknown values, treating an unknown value as if the member was unset.
 
         The {{AuthenticatorAttestationResponse/getTransports()}} operation can provide suitable values for this member.
         When [[#sctn-registering-a-new-credential|registering a new credential]],
--- a/index.bs
+++ b/index.bs
@@ -2348,7 +2348,7 @@ optionally evidence of [=user consent=] to a specific transaction.
     This dictionary is used to supply additional parameters when creating a new credential.
 
     :   <dfn>type</dfn>
-    ::  This member specifies the type of credential to be created. The value SHOULD be a member of {{PublicKeyCredentialType}} but [=client platforms=] MUST ignore unknown values, treating an unknown value as if the member was unset.
+    ::  This member specifies the type of credential to be created. The value SHOULD be a member of {{PublicKeyCredentialType}} but [=client platforms=] MUST ignore unknown values, ignoring any {{PublicKeyCredentialParameters}} with an unknown {{PublicKeyCredentialParameters/type}}.
 
     :   <dfn>alg</dfn>
     ::  This member specifies the cryptographic signature algorithm with which the newly generated credential will be used, and
@@ -2907,14 +2907,14 @@ parameter to the {{CredentialsContainer/create()}} or {{CredentialsContainer/get
 
 <div dfn-type="dict-member" dfn-for="PublicKeyCredentialDescriptor">
     :   <dfn>type</dfn>
-    ::  This member contains the type of the [=public key credential=] the caller is referring to. The value SHOULD be a member of {{PublicKeyCredentialType}} but [=client platforms=] MUST ignore unknown values, treating an unknown value as if the member was unset.
+    ::  This member contains the type of the [=public key credential=] the caller is referring to. The value SHOULD be a member of {{PublicKeyCredentialType}} but [=client platforms=] MUST ignore any {{PublicKeyCredentialDescriptor}} with an unknown {{PublicKeyCredentialDescriptor/type}}.
 
     :   <dfn>id</dfn>
     ::  This member contains the [=credential ID=] of the [=public key credential=] the caller is referring to.
 
     :   <dfn>transports</dfn>
     ::  This OPTIONAL member contains a hint as to how the [=client=] might communicate with the [=managing authenticator=] of the
-        [=public key credential=] the caller is referring to. The values SHOULD be members of {{AuthenticatorTransport}} but [=client platforms=] MUST ignore unknown values, treating an unknown value as if the member was unset.
+        [=public key credential=] the caller is referring to. The values SHOULD be members of {{AuthenticatorTransport}} but [=client platforms=] MUST ignore unknown values.
 
         The {{AuthenticatorAttestationResponse/getTransports()}} operation can provide suitable values for this member.
         When [[#sctn-registering-a-new-credential|registering a new credential]],
--- a/index.bs
+++ b/index.bs
@@ -799,6 +799,16 @@ would be obtained by the specification's algorithms.
 A conforming User Agent MUST also be a conforming implementation of the IDL fragments of this specification, as described in the
 “Web IDL” specification. [[!WebIDL]]
 
+### Enumerations as DOMString types ### {#sct-domstring-backwards-compatibility}
+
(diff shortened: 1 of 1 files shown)


jcjones:
This will need to rebase on https://github.com/w3c/webauthn/pull/1393.

index.bs
```diff
@@ -2899,14 +2898,14 @@ parameter to the {{CredentialsContainer/create()}} or {{CredentialsContainer/get
 
 <div dfn-type="dict-member" dfn-for="PublicKeyCredentialDescriptor">
     :   <dfn>type</dfn>
-    ::  This member contains the type of the [=public key credential=] the caller is referring to.
+    ::  This member contains the type of the [=public key credential=] the caller is referring to. The value SHOULD be a member of {{PublicKeyCredentialType}} but [=client platforms=] MUST ignore unknown values, treating an unknown value as if the member was unset.
```
agl: Since this is a required member, treating it &ldquo;as if the member was unset&rdquo; would mean a parse error, which is what we're trying to avoid. How about &ldquo;ignoring any [[PublicKeyCredentialDescriptor]] with an unknown type&rdquo;.

index.bs
```diff
@@ -2899,14 +2898,14 @@ parameter to the {{CredentialsContainer/create()}} or {{CredentialsContainer/get
...
 
     :   <dfn>id</dfn>
     ::  This member contains the [=credential ID=] of the [=public key credential=] the caller is referring to.
 
     :   <dfn>transports</dfn>
     ::  This OPTIONAL member contains a hint as to how the [=client=] might communicate with the [=managing authenticator=] of the
-        [=public key credential=] the caller is referring to. The values SHOULD be members of {{AuthenticatorTransport}} but [=client platforms=] MUST ignore unknown values.
+        [=public key credential=] the caller is referring to. The values SHOULD be members of {{AuthenticatorTransport}} but [=client platforms=] MUST ignore unknown values, treating an unknown value as if the member was unset.
```
agl: I think just leaving it at &ldquo;ignore unknown values&rdquo; is fine? Since it's a list, we just want to filter unknown values rather than discard the whole list.

index.bs
```diff
@@ -2837,7 +2836,7 @@ Note: The {{CollectedClientData}} may be extended in the future. Therefore it's
 
         <div dfn-type="dict-member" dfn-for="TokenBinding">
             :   <dfn>status</dfn>
-            ::  This member is one of the following:
+            ::  This member SHOULD be a member of {{TokenBindingStatus}} but [=client platforms=] MUST ignore unknown values, treating an unknown value as if the {{CollectedClientData/tokenBinding}} member was absent. When known, this member is one of the following:
```
agl: This is a required member so I think the client would need to treat unknown values as if the whole `TokenBinding` structure was absent.

index.bs
```diff
@@ -2339,7 +2339,7 @@ optionally evidence of [=user consent=] to a specific transaction.
     This dictionary is used to supply additional parameters when creating a new credential.
 
     :   <dfn>type</dfn>
-    ::  This member specifies the type of credential to be created.
+    ::  This member specifies the type of credential to be created. The value SHOULD be a member of {{PublicKeyCredentialType}} but [=client platforms=] MUST ignore unknown values, treating an unknown value as if the member was unset.
```
agl: Since this is a required member, treating it “as if the member was unset” would mean a parse error, which is what we're trying to avoid. How about “ignoring any [[PublicKeyCredentialParameters]] with an unknown type”.

index.bs
jcjones: Hmm, I thought it was a bit ambiguous as it was, but I'll revert it. Thanks!

index.bs
jcjones: That's what this says, is it not?

index.bs
jcjones: Agreed, thanks.

index.bs
agl: Yes, sorry.

index.bs
```diff
@@ -1783,7 +1783,7 @@ a numbered step. If outdented, it (today) is rendered either as a bullet in the
                         ::  |attestationObject|
 
                         :   {{AuthenticatorAttestationResponse/[[transports]]}}
-                        ::  A sequence of zero or more unique {{DOMString}}s, in lexicographical order, that the |authenticator| is believed to support. The values SHOULD be members of {{AuthenticatorTransport}}.
+                        ::  A sequence of zero or more unique {{DOMString}}s, in lexicographical order, that the |authenticator| is believed to support. The values SHOULD be members of {{AuthenticatorTransport}}, but [=client platforms=] MUST ignore unknown values.
```
equalsJeffH, line 1786: down below where `AuthenticatorAttestationResponse` interface is devined, the language for `[[transports]]` says: 
> The values SHOULD be members of AuthenticatorTransport but Relying Parties MUST **accept** unknown values.

Should we `s/ accept / ignore /` in the latter (it is presently untouched by this PR) or instead alter the above change to be:
> , but [=client platforms=] MUST **accept** unknown values.

..?

I.e., ought the language be the same in both places?

jcjones: It should indeed be the same in both, and I don't find myself having a definite preference, though I think `ignore unknown values` might be clearer. @agl?

equalsJeffH: thx, `ignore unknown values` is fine by me, and we should make the language the same in both places.

jcjones: Fixed in https://github.com/w3c/webauthn/pull/1392/commits/8b0dbbc2cf7a7b6054dfadf49af920977d3bd4f8

index.bs
```diff
@@ -2544,17 +2544,17 @@ attributes.
 
 <xmp class="idl">
     dictionary AuthenticatorSelectionCriteria {
-        AuthenticatorAttachment      authenticatorAttachment;
+        DOMString                    authenticatorAttachment;
         boolean                      requireResidentKey = false;
         ResidentKeyRequirement       residentKey;
```
emlun, line 2549: `residentKey` also needs updating, doesn't it?

index.bs
```diff
@@ -2339,7 +2339,7 @@ optionally evidence of [=user consent=] to a specific transaction.
 
 <xmp class="idl">
     dictionary PublicKeyCredentialParameters {
-        required PublicKeyCredentialType      type;
+        required DOMString                    type;
```
emlun, line 2342: The corresponding one in `credTypesAndPubKeyAlgs` in [`authenticatorMakeCredential`](https://pr-preview.s3.amazonaws.com/jcjones/webauthn/pull/1392.html#sctn-op-make-cred) can remain a strict enum, right?

jcjones: I think that's fine. By the time we're running the Authenticator-level algorithm, the User Agent has already ignored unknown values.

index.bs
jcjones: Yes, it does! Good catch, thank you!

equalsJeffH:
where you are using the term "unset", do you mean by that "does not [exist](https://infra.spec.whatwg.org/#map-exists)" (note link to infra spec)?

index.bs
emlun: Yep, sounds good!

jcjones:
> where you are using the term "unset", do you mean by that "does not [exist](https://infra.spec.whatwg.org/#map-exists)" (note link to infra spec)?

Yes, I'll update all that wordage - thank you for the suggestion!

index.bs
```diff
@@ -2418,7 +2418,7 @@ optionally evidence of [=user consent=] to a specific transaction.
 
     :   <dfn>attestation</dfn>
     ::  This member is intended for use by [=[RPS]=] that wish to express their preference for [=attestation conveyance=].
-        The default is {{AttestationConveyancePreference/none}}.
+        The value SHOULD be a member of {{AttestationConveyancePreference}} but [=client platforms=] MUST ignore unknown values, treating an unknown value as if the member was unset. The default is {{AttestationConveyancePreference/none}}.
```
jcjones: @equalsJeffH: The `default is...` text is unique to this entry, and in the actual generated doc quite duplicative since the default is shown. I was going to remove it from this sentence while I was here. Thoughts?

index.bs
```diff
@@ -1819,7 +1819,7 @@ a numbered step. If outdented, it (today) is rendered either as a bullet in the
                         ::  |attestationObject|
 
                         :   {{AuthenticatorAttestationResponse/[[transports]]}}
-                        ::  A sequence of zero or more unique {{DOMString}}s, in lexicographical order, that the |authenticator| is believed to support. The values SHOULD be members of {{AuthenticatorTransport}}.
+                        ::  A sequence of zero or more unique {{DOMString}}s, in lexicographical order, that the |authenticator| is believed to support. The values SHOULD be members of {{AuthenticatorTransport}}, but [=client platforms=] MUST ignore unknown values.
```
emlun, line 1822: Isn't this value set by the client? Then it seems a bit odd that the client would be the one to ignore unknown values, but I'm not sure what else would make sense to say here.

index.bs
```diff
@@ -2335,7 +2335,7 @@ during registration.
     ::  This operation returns the value of {{AuthenticatorAttestationResponse/[[transports]]}}.
 
     :   <dfn>\[[transports]]</dfn>
-    ::  This [=internal slot=] contains a sequence of zero or more unique {{DOMString}}s in lexicographical order. These values are the transports that the [=authenticator=] is believed to support, or an empty sequence if the information is unavailable. The values SHOULD be members of {{AuthenticatorTransport}} but [=[RPS]=] MUST accept unknown values.
+    ::  This [=internal slot=] contains a sequence of zero or more unique {{DOMString}}s in lexicographical order. These values are the transports that the [=authenticator=] is believed to support, or an empty sequence if the information is unavailable. The values SHOULD be members of {{AuthenticatorTransport}} but [=[RPS]=] MUST ignore unknown values.
```
emlun, line 2338: Is this right? Shouldn't the RP just pass all values through (known or unknown) and let the client ignore the ones it doesn't need?

index.bs
jcjones: It's definitely awkward, but an authenticator could betray the spec and return an unknown value, which the system needs to handle.

index.bs
jcjones: But what does the RP do with an unknown value?

The authenticator says it supports USB and ZWave. Does the RP need to permit ZWave, or just ignore it?

jcjones:
@equalsJeffH : Please take a look at https://github.com/w3c/webauthn/pull/1392/commits/42dbe07fa4e9661f46ff61ea5b9f40dcddf4ecbf , as the change is perhaps larger than we had initially discussed but I like the way this ties together better.

@emlun: Please see responses above. I'm OK with changing things, but maybe you can propose the changes you'd like to see?

Thanks and have a good weekend!

index.bs
emlun: Yeah, that makes sense under the assumption that the client asks the authenticator to generate this value, but it's not really evident here that that is the expected case... Anyway, this is probably fine, just something to consider.

index.bs
emlun: Since this value isn't signed, the RP should never make permission decisions based on it, right  (such decisions should be based on data derived from the attestation cert)? So I don't know what else the RP would do with this value other than echo it back to the client.

index.bs
```diff
@@ -2335,7 +2345,7 @@ during registration.
...
```
emlun, line 2348: I'm thinking something like this:
```suggestion
    ::  This [=internal slot=] contains a sequence of zero or more unique {{DOMString}}s in lexicographical order. These values are the transports that the [=authenticator=] is believed to support, or an empty sequence if the information is unavailable. The values SHOULD be members of {{AuthenticatorTransport}} but [=[RPS]=] SHOULD accept unknown values
        and return them in {{PublicKeyCredentialRequestOptions/allowCredentials}} in {{CredentialsContainer/get()}} operations.
```

equalsJeffH: we are going to land this without this suggestion, and submit an issue for this suggested change.

PR #1393, opened 2020-03-24
Use Python 3.7 for Travis-CI

jcjones:
image

PR #1394, opened 2020-03-24
Prohibit Create Credential from cross-origin iframes

jcjones:
This reverts part of PR #1276, again prohibiting the use of the Create method when `sameOriginWithAncestors` is `false`. The `Note` is simplified, since the integration between Credential Management and Feature Policy is now complete.

***
<a href="https://pr-preview.s3.amazonaws.com/jcjones/webauthn/pull/1394.html" title="Last updated on Apr 8, 2020, 7:02 PM UTC (1c16985)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/1394/a636817...jcjones:1c16985.html" title="Last updated on Apr 8, 2020, 7:02 PM UTC (1c16985)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -3009,9 +3009,9 @@ needs.
 
 ## Feature Policy integration ## {#sctn-feature-policy}
 
-This specification defines a [=policy-controlled feature=] identified by
-the feature-identifier token "<code><dfn data-lt="publickey-credentials-feature" export>publickey-credentials</dfn></code>".
-Its [=default allowlist=] is '<code>self</code>'. [[!Feature-Policy]]
+This specification defines two [=policy-controlled features=] identified by
+the feature-identifier tokens "<code><dfn data-lt="publickey-credentials-get-feature" export>publickey-credentials-get</dfn></code>" and "<code><dfn data-lt="publickey-credentials-create-feature" export>publickey-credentials-create</dfn></code>".
+Their [=default allowlists=] are '<code>self</code>'. [[!Feature-Policy]]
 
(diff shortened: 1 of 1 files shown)


jcjones:
This will also need to rebase on #1393.

jcjones:
> We also need to split the feature-policy, right?

Do we?

kenrb:
Or rename the feature policy, at least? I thought the idea was that if we decide later to allow MakeCredential from cross-origin iframes, we would need a new feature policy, so as not to change the behavior of an existing one for sites that might be using it. So the feature policy being added for this should correspond specifically to GetAssertion, if I understand correctly.

ve7jtb:
We agreed at the F2F to split the policy so we could add make credential later if we decide to without breaking the current get behavior.

agl:
Discussed on the call: we're going to rename with, for example, -create and -get.

jcjones:
Thanks for the reviews; see what you think of this revision.

equalsJeffH:
on 2020-04-01 call: @equalsJeffH & @akshayku to review

equalsJeffH:
on 2020-04-08 call: @equalsJeffH to re-review and merge if good

equalsJeffH:
feature policy name change PR submitted: https://github.com/w3c/webappsec-feature-policy/pull/370 "Update publickey-credentials to be "get-only" policy"

equalsJeffH:
wrt https://github.com/w3c/webauthn/pull/1394#issuecomment-611635942:

https://github.com/w3c/webappsec-feature-policy/pull/370 -- landed; 
 see also: https://github.com/w3c/webappsec-feature-policy/blob/master/features.md

PR #1395, opened 2020-03-25
Add getPublicKey method.

### LINKED ISSUE #1363: Provide the public key in `AuthenticatorAttestationResponse`
The public key for a freshly created credential is provided inside of the [attestation object](https://w3c.github.io/webauthn/#dom-authenticatorattestationresponse-attestationobject). However, that is a somewhat complex format that involves decoding CBOR in order to read the public key. If a site doesn't care about attestation (as many won't) we might usefully be able to have browsers provide fields of this structure more directly.

Assumption: absent attestation, web site implementations wouldn't need CBOR if we did this. This appears to be true at first glance since the [authenticator data](https://w3c.github.io/webauthn/#authenticator-data) is a fixed-offset binary format (not including extensions).

A reason not to do this would be that it encourages sites to depend on these additional fields, which will only be available in newer browsers. Thus people with older browsers might not be able to use WebAuthn, even though they could if sites put in more work. However, this argument applies to any such ergonomic improvement to the API and so, if we buy it, we're forced to conclude that they're mostly a bad idea as a class.

agl:
This change adds a getPublicKey method to the
AuthenticatorAttestationResponse to save some users from having to parse
out and handle COSE keys.

(See linked issue for background.)

Fixes #1363

***
<a href="https://pr-preview.s3.amazonaws.com/agl/webauthn/pull/1395.html" title="Last updated on May 14, 2020, 10:51 PM UTC (0e3c67b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/1395/ae29ff0...agl:0e3c67b.html" title="Last updated on May 14, 2020, 10:51 PM UTC (0e3c67b)">Diff</a>


agl:
For discussion: if we want to avoid uses having to grub around in the authenticator data, should we also provide methods to get the signature count and UP/UV flags?

agl:
From the call: yes, we should add more accessors for other stuff that people are likely to need.

equalsJeffH:
on 2020-03-25 call:  noted that RPs that come to rely upon these accessors will need to do feature detection appropriately in case they end up running in older browsers -- call this out in spec in a Note at least.  also, add accessors for select other portions of the attestationObject -- tho how many/which ones?

equalsJeffH:
on 2020-04-01 call: @agl to update, then reviewers to review :)

agl:
* Added note about feature detection.
   * Added `getAuthenticatorData` to extract that from the attestation object.
   * Added note about why accessors for the flags and signature counter would not be helpful.

index.bs
```diff
@@ -2289,10 +2297,33 @@ during registration.
...
 </div>
 
+#### Easily accessing credential data #### {#sctn-public-key-easy}
+
+Every user of the [[#sctn-createCredential|create method]] will need to parse and store the returned public key in order to verify future assertions. However, the public key is in [[!RFC8152]] (COSE) format, inside the [=credentialPublicKey=] member of the [=attestedCredentialData=], inside the [=authenticator data=], inside the [=attestation object=] from {{AuthenticatorAttestationResponse/attestationObject}}. Users that wish to use [=attestation=] MUST do the work of parsing that out because that copy of the public-key is the one that the [=authenticator=] has [signed](#signing-procedure). However, many valid uses of WebAuthn do not need [=attestation=]. For those uses, user agents can do the work of parsing, expose the [=authenticator data=] directly, and translate the public key into a more convenient format.
```
equalsJeffH, line 2312: We use `[=base64url encoding=]` throughout the spec. Ought we also here in this section?  If not we perhaps ought to note that and explain why.

Editorial polishing suggestion to more align with rest of spec:
```suggestion
Every user of the {{PublicKeyCredential/[[Create]](origin, options, sameOriginWithAncestors)}} method will need to parse and store the returned [=credential public key=] in order to verify future [=authentication assertions=]. However, the [=credential public key=] is in [[!RFC8152]] (COSE) format, inside the [=credentialPublicKey=] member of the [=attestedCredentialData=], inside the [=authenticator data=], inside the [=attestation object=] conveyed by {{AuthenticatorAttestationResponse}}.{{AuthenticatorAttestationResponse/attestationObject}}. [=[RPS]=] wishing to use [=attestation=] are obliged to do the work of parsing the {{AuthenticatorAttestationResponse/attestationObject}} and obtaining the [=credential public key=]  because that public key copy is the one the [=authenticator=] [signed](#signing-procedure). However, many valid WebAuthn use cases do not require [=attestation=]. For those uses, user agents can do the work of parsing, expose the [=authenticator data=] directly, and translate the [=credential public key=] into a more convenient format.
```

index.bs
```diff
@@ -2289,10 +2297,33 @@ during registration.
...
+
+The {{AuthenticatorAttestationResponse/getPublicKey()}} operation thus returns the credential public key as a [=SubjectPublicKeyInfo=]. After base64 decoding, this format can, for example, be passed to Java's `java.security.spec.X509EncodedKeySpec`, .NET's `System.Security.Cryptography.ECDsa.ImportSubjectPublicKeyInfo`, or Go's `crypto/x509.ParsePKIXPublicKey`.
```
equalsJeffH, line 2314:
```suggestion
The {{AuthenticatorAttestationResponse/getPublicKey()}} operation thus returns the [=credential public key=] as a [=SubjectPublicKeyInfo=]. After base64 decoding, this format can, for example, be passed to Java's `java.security.spec.X509EncodedKeySpec`, .NET's `System.Security.Cryptography.ECDsa.ImportSubjectPublicKeyInfo`, or Go's `crypto/x509.ParsePKIXPublicKey`.
```

index.bs
```diff
@@ -2289,10 +2297,33 @@ during registration.
...
+
+Use of this function does impose some limitations: by using {{PublicKeyCredentialCreationOptions/pubKeyCredParams}}, a [=[RP]=] can negotiate with the [=authenticator=] to use public-key algorithms that the user agent may not understand. However, if it does so, the user agent will not be able translate the resulting public key into [=SubjectPublicKeyInfo=] format and the return value of {{AuthenticatorAttestationResponse/getPublicKey()}} will be an empty string.
```
equalsJeffH, line 2316:
```suggestion
Use of {{AuthenticatorAttestationResponse/getPublicKey()}} does impose some limitations: by using {{PublicKeyCredentialCreationOptions/pubKeyCredParams}}, a [=[RP]=] can negotiate with the [=authenticator=] to use public key algorithms that the user agent may not understand. However, if it does so, the user agent will not be able translate the resulting [=credential public key=] into [=SubjectPublicKeyInfo=] format and the return value of {{AuthenticatorAttestationResponse/getPublicKey()}} will be an empty string.
```

index.bs
```diff
@@ -2289,10 +2297,33 @@ during registration.
...
+
+User agents MUST be able to provide a non-empty value for {{AuthenticatorAttestationResponse/getPublicKey()}} when the [=credential public key=] has a {{COSEAlgorithmIdentifier}} value of:
```
equalsJeffH, line 2318:
```suggestion
User agents MUST be able to return a non-empty value for {{AuthenticatorAttestationResponse/getPublicKey()}} when the [=credential public key=] has a {{COSEAlgorithmIdentifier}} value of:
```

index.bs
```diff
@@ -2289,10 +2297,33 @@ during registration.
...
+   * -7 (ECDSA with SHA-256), where [=kty=] is 2 (uncompressed points) and [=crv=] is 1 (P-256).
```
equalsJeffH, line 2319: just a headsup that we note -7 as "ES256" elsewhere in the spec. That's the name for it given in https://tools.ietf.org/html/rfc8152#section-8.1

index.bs
```diff
@@ -2289,10 +2297,33 @@ during registration.
...
+   * -257 (RSA PKCS#1 with SHA-256).
```
equalsJeffH, line 2320: similarly: "RS256"  (https://www.iana.org/assignments/cose/cose.xhtml#algorithms)

index.bs
```diff
@@ -2289,10 +2297,33 @@ during registration.
...
+
+Note: {{AuthenticatorAttestationResponse/getPublicKey()}} and {{AuthenticatorAttestationResponse/getAuthenticatorData()}} were only added in level two of this spec. [=[RPS]=] SHOULD use feature detection being using the function by testing the value of `'getPublicKey' in AuthenticatorAttestationResponse.prototype`. [=[RPS]=] that require this function to exist may not interoperate with older user-agents.
```
equalsJeffH, line 2325: Did you intend:
```suggestion
Note: {{AuthenticatorAttestationResponse/getPublicKey()}} and {{AuthenticatorAttestationResponse/getAuthenticatorData()}} were only added in level two of this spec. [=[RPS]=] SHOULD use feature detection before using these functions by testing the value of `'getPublicKey' in AuthenticatorAttestationResponse.prototype`. [=[RPS]=] that require this function to exist may not interoperate with older user-agents.
```
...?

equalsJeffH:
on 2020-04-08 call: please review!

agl:
(Just adding a note here that, while people should review this, I think I should sketch an implementation before we land this in the spec to ensure that there aren't any unexpected gotchas.)

index.bs
emlun: > We use `[=base64url encoding=]` throughout the spec. Ought we also here in this section?

I'm inclined to agree, it seems good to stick with base64url as much as possible.

index.bs
emlun: There's also a missing "to" in "be able translate":

```suggestion
Use of {{AuthenticatorAttestationResponse/getPublicKey()}} does impose some limitations: by using {{PublicKeyCredentialCreationOptions/pubKeyCredParams}}, a [=[RP]=] can negotiate with the [=authenticator=] to use public key algorithms that the user agent may not understand. However, if it does so, the user agent will not be able to translate the resulting [=credential public key=] into [=SubjectPublicKeyInfo=] format and the return value of {{AuthenticatorAttestationResponse/getPublicKey()}} will be an empty string.
```

index.bs
```diff
@@ -2289,10 +2297,33 @@ during registration.
...
+
+User agents MUST be able to provide a non-empty value for {{AuthenticatorAttestationResponse/getPublicKey()}} when the [=credential public key=] has a {{COSEAlgorithmIdentifier}} value of:
+   * -7 (ECDSA with SHA-256), where [=kty=] is 2 (uncompressed points) and [=crv=] is 1 (P-256).
+   * -257 (RSA PKCS#1 with SHA-256).
+   * -8 (EdDSA), where [=crv=] is 6 (Ed25519).
+
+To remove the need to parse CBOR completely in many cases, {{AuthenticatorAttestationResponse/getAuthenticatorData()}} returns the [=authenticator data=] from {{AuthenticatorAttestationResponse/attestationObject}}. The [=authenticator data=] contains other fields that are encoded in a binary format. However, helper functions are not provided to access them because [=[RPS]=] already need to extract those fields when [getting an assertion](#sctn-getAssertion). In contrast to [credential creation](#sctn-createCredential), where signature verification is [optional](#enumdef-attestationconveyancepreference), [=[RPS]=] should always be verifying signatures from an assertion and thus must extract fields from the signed [=authenticator data=]. The same functions used there will also serve during credential creation.
```
emlun, line 2323: Maybe?
(Change "parse CBOR completely" to "parse CBOR at all")

```suggestion
To remove the need to parse CBOR at all in many cases, {{AuthenticatorAttestationResponse/getAuthenticatorData()}} returns the [=authenticator data=] from {{AuthenticatorAttestationResponse/attestationObject}}. The [=authenticator data=] contains other fields that are encoded in a binary format. However, helper functions are not provided to access them because [=[RPS]=] already need to extract those fields when [getting an assertion](#sctn-getAssertion). In contrast to [credential creation](#sctn-createCredential), where signature verification is [optional](#enumdef-attestationconveyancepreference), [=[RPS]=] should always be verifying signatures from an assertion and thus must extract fields from the signed [=authenticator data=]. The same functions used there will also serve during credential creation.
```

index.bs
```diff
@@ -2289,10 +2297,33 @@ during registration.
     :   {{AuthenticatorAttestationResponse/getTransports()}}
     ::  This operation returns the value of {{AuthenticatorAttestationResponse/[[transports]]}}.
 
+    :   {{AuthenticatorAttestationResponse/getAuthenticatorData()}}
+    ::  This operation returns the [=authenticator data=] contained within {{AuthenticatorAttestationResponse/attestationObject}}. See [[#sctn-public-key-easy]].
+
+    :   {{AuthenticatorAttestationResponse/getPublicKey()}}
+    ::  This operation returns a string that contains the base64-encoded, DER [=SubjectPublicKeyInfo=] of the new credential, or an empty string if this is not available. See [[#sctn-public-key-easy]].
```
emlun, line 2304: Are we sure empty string is the best way to signal "not supported"? What about `null`, `undefined` or `false`? Changing the type is a bit ugly, but I think it's usually better than special sentinel values of the same type.

How about `getAuthenticatorData()` and `getPublicKey()` both return `ArrayBuffer` or both return base64(url?) `DOMString`? We use `ArrayBuffer` for all other byte arrays, right?

index.bs
emlun: Actually, on second thought, maybe this should be raw `ArrayBuffer`? See: https://github.com/w3c/webauthn/pull/1395#discussion_r409055892

lgarron: (I'm keeping an eye on this for https://github.com/github/webauthn-json)

While I'm not a fan of all the `ArrayBuffer`s, everything binary that is passed/returned in the browser API is an `ArrayBuffer` and I think by far the least confusing thing would be to stay consistent. It might also save code for application who want to process on the frontend but don't otherwise need a `base64url` decoder.

index.bs
```diff
@@ -2267,6 +2273,8 @@ during registration.
     interface AuthenticatorAttestationResponse : AuthenticatorResponse {
         [SameObject] readonly attribute ArrayBuffer      attestationObject;
         sequence<DOMString>                              getTransports();
+        ArrayBuffer                                      getAuthenticatorData();
+        DOMString                                        getPublicKey();
```
jcjones, line 2277: I'm in favor, with the commentors below, that this be a nullable ArrayBuffer with the DER-encoded bytes of the SPKI.

```suggestion
        ArrayBuffer?                                        getPublicKey();
```

index.bs
jcjones: I vote for ArrayBuffer, and make it nullable rather than checking len>0.

index.bs
jcjones, line 2314: If we use ArrayBuffer, then: 
```suggestion
The {{AuthenticatorAttestationResponse/getPublicKey()}} operation thus returns the credential public key as a [=SubjectPublicKeyInfo=]. This [=ArrayBuffer=] can, for example, be passed to Java's `java.security.spec.X509EncodedKeySpec`, .NET's `System.Security.Cryptography.ECDsa.ImportSubjectPublicKeyInfo`, or Go's `crypto/x509.ParsePKIXPublicKey`.
```

index.bs
agl: Will make it nullable.

agl:
Latest pushes switch to ArrayBuffer for the public key, apply all suggestions, and add `getPublicKeyAlgorithm` to convey the COSE algorithm that's in a COSE Key, but not a SubjectPublicKeyInfo.

equalsJeffH:
on 2020-04-29 call, @agl still working on this and an implementation, so remains a moving target...

agl:
I've fleshed out an implementation of this to the point where I believe it'll work and thus think this PR is in good shape.

index.bs
```diff
@@ -2289,10 +2298,38 @@ during registration.
...
+
+The {{AuthenticatorAttestationResponse/getPublicKey()}} operation thus returns the [=credential public key=] as a [=SubjectPublicKeyInfo=]. This [=ArrayBuffer=] can, for example, be passed to Java's `java.security.spec.X509EncodedKeySpec`, .NET's `System.Security.Cryptography.ECDsa.ImportSubjectPublicKeyInfo`, or Go's `crypto/x509.ParsePKIXPublicKey`.
+
+Use of {{AuthenticatorAttestationResponse/getPublicKey()}} does impose some limitations: by using {{PublicKeyCredentialCreationOptions/pubKeyCredParams}}, a [=[RP]=] can negotiate with the [=authenticator=] to use public key algorithms that the user agent may not understand. However, if it does so, the user agent will not be able to translate the resulting [=credential public key=] into [=SubjectPublicKeyInfo=] format and the return value of {{AuthenticatorAttestationResponse/getPublicKey()}} will be null.
```
emlun, line 2320: Maybe?

```suggestion
Use of {{AuthenticatorAttestationResponse/getPublicKey()}} does impose some limitations: by using {{PublicKeyCredentialCreationOptions/pubKeyCredParams}}, a [=[RP]=] can negotiate with the [=authenticator=] to use public key algorithms that the user agent may not understand. However, if the [=[RP]=] does so, the user agent will not be able to translate the resulting [=credential public key=] into [=SubjectPublicKeyInfo=] format and the return value of {{AuthenticatorAttestationResponse/getPublicKey()}} will be null.
```

jcjones:
Per the 13 May call, merging!

Issue #98, opened 2016-05-13
Add opaque data extension


gmandyam:
An RP may send opaque data to an authenticator via an extension that requires no client processing.  This should be a pre-registered extension type and would be passed directly to the authenticator from the client.

equalsJeffH:
Does this last paragraph of the current section 5 {#extension-request-parameters} address this use case?

```
For extensions that specify additional authenticator processing only, it is desirable that the platform need not know the extension. To support this, platforms SHOULD pass the client argument of unknown extension as the authenticator argument unchanged, under the same extension identifier. The authenticator argument should be the CBOR encoding of the client argument, as specified in Section 4.2 of [RFC7049]. Clients SHOULD silently drop unknown extensions whose client argument cannot be encoded as a CBOR structure.
```

equalsJeffH:
shall we close this issue?

gmandyam:
Closing issue.  Note that as long as we do not put normative requirements on how UVI is generated by an authenticator, it is for all intents and purposes an opaque data extension.

PR #1402, opened 2020-04-02
Add `largeBlob` extension.

agl:
This allows “RPs” who, in this case will likely be browser extensions or
PWAs in offline mode, to store small amounts of arbitrary data with a
credential.

***
<a href="https://pr-preview.s3.amazonaws.com/agl/webauthn/pull/1402.html" title="Last updated on Jun 5, 2020, 11:03 PM UTC (30d6438)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/1402/50679f5...agl:30d6438.html" title="Last updated on Jun 5, 2020, 11:03 PM UTC (30d6438)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -221,6 +221,7 @@ spec: FIDO-CTAP; urlPrefix: https://fidoalliance.org/specs/fido-v2.0-ps-20190130
     type: dfn
         text: CTAP2 canonical CBOR encoding form; url: ctap2-canonical-cbor-encoding-form
         text: §6.2. Responses; url: responses
+        text: large, per-credential blobs; url: large-blob
 
 spec: FIDO-APPID; urlPrefix: https://fidoalliance.org/specs/fido-v2.0-id-20180227/fido-appid-and-facets-v2.0-id-20180227.html
     type: dfn
@@ -5839,6 +5840,8 @@ Note: [=[RPS]=] can assume that the opaque data will be compressed when being wr
 Since a certificate system needs to sign over the public key of the credential, and that public key is only available after creation, this extension is not defined in the [=registration extension|registration=] context.
 
 Since certificates are sizable relative to the storage capabilities of typical authenticators, user agents SHOULD consider what indications and confirmations are suitable to best guide the user in allocating this limited resource and prevent abuse.
+
+Note: In order to interoperate, user agents storing large blobs on authenticators using [[!FIDO-CTAP]] are expected to use the provisions detailed in that specification for storing [=large, per-credential blobs=].
 
 : Extension identifier
 :: `largeBlob`
--- a/index.bs
+++ b/index.bs
@@ -5848,7 +5848,7 @@ This [=client extension|client=] [=registration extension=] and [=authentication
     };
     </xmp>
 
-    <div dfn-type="dict-member" dfn-for="CredentialPropertiesOutput">
+    <div dfn-type="dict-member" dfn-for="AuthenticationExtensionsClientInputs">
         :   <dfn>blobWrite</dfn>
         ::  An opaque byte string that the [=[WRP]=] wishes to store with the freshly created credential (in the [=registration extension=] case) or an existing credential (in the case of an [=authentication extension=]).
 
@@ -5857,7 +5857,7 @@ This [=client extension|client=] [=registration extension=] and [=authentication
     </div>
 
 : Client extension processing ([=registration extension|registration=])
-:: If {{AuthenticationExtensionsClientInputs/blobWrite}} is present, attempt to store its contents on the [=authenticator=], associated with the freshly created credential. Set {{AuthenticationExtensionsClientInputs/blobWritten}} to `true` if successful and `false` otherwise.
+:: If {{AuthenticationExtensionsClientInputs/blobWrite}} is present, attempt to store its contents on the [=authenticator=], associated with the freshly created credential. Set {{AuthenticationExtensionsClientOutputs}}.{{AuthenticationExtensionsClientOutputs/blobWritten}} to `true` if successful and `false` otherwise.
 
 : Client extension processing ([=authentication extension|authentication=])
 ::
@@ -5876,12 +5876,12 @@ This [=client extension|client=] [=registration extension=] and [=authentication
     };
     </xmp>
 
-    <div dfn-type="dict-member" dfn-for="CredentialPropertiesOutput">
+    <div dfn-type="dict-member" dfn-for="AuthenticationExtensionsClientOutputs">
         :   <dfn>blob</dfn>
         ::  The opaque byte string that was associated with the credential identified by {{PublicKeyCredential/rawId}}. Only valid in the [=authentication extension=] case and only present if {{AuthenticationExtensionsClientInputs/blobRead}} was `true`.
 
         :   <dfn>blobWritten</dfn>
-        ::  A boolean that indicates that the contents of {{AuthenticationExtensionsClientInputs/blobWrite}} were successfully store on the [=authenticator=], associated with the specified extension.
+        ::  A boolean that indicates that the contents of {{AuthenticationExtensionsClientInputs/blobWrite}} were successfully stored on the [=authenticator=], associated with the specified extension.
     </div>
 
 
--- a/index.bs
+++ b/index.bs
@@ -5834,6 +5834,8 @@ At this time, one [=credential property=] is defined: the [=resident key credent
 
 This [=client extension|client=] [=authentication extension=] allows a [=[RP]=] to store opaque data associated with a credential. Since [=authenticators=] can only store small amounts of data, and most [=[RPS]=] are online services that can store arbitrary amounts of state for a user, this is only useful in specific cases. For example, the [=[RP]=] might wish to issue certificates rather than run a centralised authentication service.
 
+Note: [=[RPS]=] can assume that the opaque data will be compressed when being written to a space-limited device and so need not compress it themselves.
+
 Since a certificate system needs to sign over the public key of the credential, and that public key is only available after creation, this extension is not defined in the [=registration extension|registration=] context.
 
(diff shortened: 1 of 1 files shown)


agl:
Since we're removing so many extensions we clearly have some room :)

gmandyam:
I guess this issue was just ahead of its time:  https://github.com/w3c/webauthn/issues/98

index.bs
```diff
@@ -5830,6 +5830,71 @@ At this time, one [=credential property=] is defined: the [=resident key credent
...
+
+: Client extension input
+::  <xmp class="idl">
+    partial dictionary AuthenticationExtensionsClientInputs {
+        ArrayBuffer blobWrite;
+        boolean blobRead;
+    };
+    </xmp>
+
+    <div dfn-type="dict-member" dfn-for="CredentialPropertiesOutput">
```
equalsJeffH, line 5851: helps fix the broken build: 
```suggestion
    <div dfn-type="dict-member" dfn-for="AuthenticationExtensionsClientInputs">
```

index.bs
```diff
@@ -5830,6 +5830,71 @@ At this time, one [=credential property=] is defined: the [=resident key credent
...
+
+: Client extension output
+:: <xmp class="idl">
+    partial dictionary AuthenticationExtensionsClientOutputs {
+        ArrayBuffer blob;
+        boolean blobWritten;
+    };
+    </xmp>
+
+    <div dfn-type="dict-member" dfn-for="CredentialPropertiesOutput">
```
equalsJeffH, line 5879: helps fix the broken build: 
```suggestion
    <div dfn-type="dict-member" dfn-for="AuthenticationExtensionsClientOutputs">
```

agl:
Applied suggestions from the future, but closing this for now as it may be premature.

agl:
Reopening this PR since the corresponding CTAP2 changes have matured.

index.bs
```diff
@@ -5830,6 +5830,71 @@ At this time, one [=credential property=] is defined: the [=resident key credent
...
+
+: Client extension processing ([=registration extension|registration=])
+:: If {{AuthenticationExtensionsClientInputs/blobWrite}} is present, attempt to store its contents on the [=authenticator=], associated with the freshly created credential. Set {{AuthenticationExtensionsClientInputs/blobWritten}} to `true` if successful and `false` otherwise.
+
+: Client extension processing ([=authentication extension|authentication=])
+::
+       1. If {{AuthenticationExtensionsClientInputs/blobWrite}} is present and {{PublicKeyCredentialRequestOptions/allowCredentials}} contains exactly one element:
+           1. If the [[#sctn-getAssertion|assertion]] operation is successful, attempt to store the contents of {{AuthenticationExtensionsClientInputs/blobWrite}} on the [=authenticator=], associated with the indicated credential.
+           1. Set {{AuthenticationExtensionsClientOutputs/blobWritten}} to `true` if successful and `false` otherwise.
+       1. If {{AuthenticationExtensionsClientInputs/blobRead}} is present and has the value `true`:
```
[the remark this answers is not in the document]
agl: (Have banned concurrent read and write to eliminate the ambiguity.)

equalsJeffH:
see also:  https://en.wikipedia.org/wiki/The_Blob

index.bs
```diff
@@ -5830,6 +5830,74 @@ At this time, one [=credential property=] is defined: the [=resident key credent
...
+
+
+: Authenticator extension input
+:: None.
+
+: Authenticator extension processing
+:: None.
+
+: Authenticator extension output
+:: None.
```
emlun, line 5898: But the blob would be stored in the authenticator, right? If so, I think it's worthwhile to point out that that is the case but the procedure is specified elsewhere (i.e., CTAP), like in #1424.

index.bs
```diff
@@ -5830,6 +5830,74 @@ At this time, one [=credential property=] is defined: the [=resident key credent
...
+
+: Operation applicability
+:: [=authentication extension|Authentication=]
+
+: Client extension input
+::  <xmp class="idl">
+    partial dictionary AuthenticationExtensionsClientInputs {
+        ArrayBuffer largeBlobWrite;
+        boolean largeBlobRead;
+    };
```
emlun, line 5854: These should be wrapped in a single member `largeBlob` of a new dictionary type (see also #1430).

index.bs
```diff
@@ -5830,6 +5830,74 @@ At this time, one [=credential property=] is defined: the [=resident key credent
...
+       1. If {{AuthenticationExtensionsClientInputs/largeBlobWrite}} is present, {{PublicKeyCredentialRequestOptions/allowCredentials}} contains exactly one element, and {{AuthenticationExtensionsClientInputs/largeBlobRead}} is not present, or has the value `false`:
+           1. If the [[#sctn-getAssertion|assertion]] operation is successful, attempt to store the contents of {{AuthenticationExtensionsClientInputs/largeBlobWrite}} on the [=authenticator=], associated with the indicated credential.
+           1. Set {{AuthenticationExtensionsClientOutputs/largeBlobWritten}} to `true` if successful and `false` otherwise.
+
+: Client extension output
+:: <xmp class="idl">
+    partial dictionary AuthenticationExtensionsClientOutputs {
+        ArrayBuffer largeBlob;
+        boolean largeBlobWritten;
+    };
```
emlun, line 5879: Ditto

index.bs
```diff
@@ -5830,6 +5830,74 @@ At this time, one [=credential property=] is defined: the [=resident key credent
...
+    </xmp>
+
+    <div dfn-type="dict-member" dfn-for="AuthenticationExtensionsClientOutputs">
+        :   <dfn>largeBlob</dfn>
+        ::  The opaque byte string that was associated with the credential identified by {{PublicKeyCredential/rawId}}. Only valid in the [=authentication extension=] case and only present if {{AuthenticationExtensionsClientInputs/largeBlobRead}} was `true`.
```
emlun, line 5884: >Only valid in the authentication extension case

This seems redundant since the extension is already defined as an authentication extension only.

index.bs
```diff
@@ -5830,6 +5830,74 @@ At this time, one [=credential property=] is defined: the [=resident key credent
...
+
+        :   <dfn>largeBlobWritten</dfn>
+        ::  A boolean that indicates that the contents of {{AuthenticationExtensionsClientInputs/largeBlobWrite}} were successfully stored on the [=authenticator=], associated with the specified extension.
```
emlun, line 5887: Did you mean this?
```suggestion
        ::  A boolean that indicates that the contents of {{AuthenticationExtensionsClientInputs/largeBlobWrite}} were successfully stored on the [=authenticator=], associated with the specified credential.
```

index.bs
agl: Agreed. Removed.

index.bs
agl: Yep, thanks.

index.bs
agl: Done in a way similar to #1424.

index.bs
agl: We haven't been following that  as noted! I guess we can start… Done.

index.bs

agl:
@ve7jtb brought up the point that this interface is suitable for cases where the browser will show a UI and handle questions like “what if the blob doesn't fit?” and “should I delete other blobs to make it fit?”. However, platform libraries may wish to have applications handle that UI, so this interface might be unsuitable for them.

After thinking about it, I think that's ok? I think the interface for UI vs no-UI cases are different, and that browsers will always want to be in the UI case.

Platforms may be general purpose operating systems that run applications with a large degree of implicit trust and may wish to make a different decision. That's fine, and that implies that the interface would be different.

agl:
> This looks mostly good to me, but I worry that it's currently impossible for clients to implement this interoperably without additional outside knowledge. I assume the corresponding CTAP extension will be the primary back-end, so an explicit reference to CTAP seems appropriate.

I think quite a lot of WebAuthn requires additional knowledge of how to map it correctly to CTAP2 in practice :) Still, no objection to providing an explicit link, although FIDO will need to rev the public version of CTAP2, and WebAuthn will have to update the base URL, for it to resolve correctly. (Done in latest commit.)

emlun:
Heheh, fair point. :)

index.bs
```diff
@@ -5830,6 +5831,78 @@ At this time, one [=credential property=] is defined: the [=resident key credent
...
+
+Note: [=[RPS]=] can assume that the opaque data will be compressed when being written to a space-limited device and so need not compress it themselves.
+
+Since a certificate system needs to sign over the public key of the credential, and that public key is only available after creation, this extension is not defined in the [=registration extension|registration=] context.
+
+Since certificates are sizable relative to the storage capabilities of typical authenticators, user agents SHOULD consider what indications and confirmations are suitable to best guide the user in allocating this limited resource and prevent abuse.
+
+Note: In order to interoperate, user agents storing large blobs on authenticators using [[!FIDO-CTAP]] are expected to use the provisions detailed in that specification for storing [=large, per-credential blobs=].
```
bdewater, line 5844: The last bit currently links to https://fidoalliance.org/specs/fido-v2.0-ps-20190130/fido-client-to-authenticator-protocol-v2.0-ps-20190130.html#large-blob on the editor's draft, which has no information on this extension?

agl: Correct, it's prospective:

> FIDO will need to rev the public version of CTAP2, and WebAuthn will have to update the base URL, for it to resolve correctly.

ve7jtb: It should be pointing to CTAP 2.1.   Unfortunately the current public Review draft is also a bit old.
https://fidoalliance.org/specs/fido2/fido-client-to-authenticator-protocol-v2.1-rd-20191217.html

The goal is to have a new public RD out shortly.

bdewater: @agl missed that part - thanks!

PR #1424, opened 2020-05-25
PRF extension.

agl:
Some applications such as password managers have requested the ability
to associate a symmetric key with a credential. The CTAP2 `hmac-secret`
extension allows something very like this, and is already widely
deployed. The limitation is that it's not possible to get an HMAC output
during registration because the extension only provides outputs for
assertions and it requires user presence. That gave me pause and we
could, instead, use the new credBlob extension. But I think the utility
of being able to rotate keys, and having existing hardware support, is
compelling enough and we'll have to see whether RPs can tolerate needing
two touches to setup.

***
<a href="https://pr-preview.s3.amazonaws.com/agl/webauthn/pull/1424.html" title="Last updated on Jun 29, 2020, 9:21 PM UTC (b036d14)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/1424/f10427d...agl:b036d14.html" title="Last updated on Jun 29, 2020, 9:21 PM UTC (b036d14)">Diff</a>


index.bs
```diff
@@ -5638,6 +5638,93 @@ At this time, one [=credential property=] is defined: the [=resident key credent
...
+: Client extension input ([=registration extension|registration=])
+::  <xmp class="idl">
+    partial dictionary AuthenticationExtensionsClientInputs {
+        boolean prfEnable;
+    };
+    </xmp>
+
+    <div dfn-type="dict-member" dfn-for="AuthenticationExtensionsClientInputs">
+        :   <dfn>prfEnable</dfn>
+        ::  If true, specifies that two PRFs should be associated with the created credential.
```
emlun, line 5668:
```suggestion
        ::  If [TRUE], specifies that two PRFs should be associated with the created credential.
```

index.bs
```diff
@@ -5638,6 +5638,93 @@ At this time, one [=credential property=] is defined: the [=resident key credent
...
+    </div>
+
+: Client extension processing ([=registration extension|registration=])
+::
+       1. If {{AuthenticationExtensionsClientInputs/prfEnable}} is present and has the value `true`:
+           1. Set `hmac-secret` to `true` in the authenticator extensions input.
+           1. Set {{AuthenticationExtensionsClientOutputs/prfEnabled}} to the value of `hmac-secret` in the authenticator extensions output. If not present, set {{AuthenticationExtensionsClientOutputs/prfEnabled}} to false.
+       1. Else:
+           1. Set {{AuthenticationExtensionsClientOutputs/prfEnabled}} to false.
```
emlun, line 5677:
```suggestion
       1. If {{AuthenticationExtensionsClientInputs/prfEnable}} is present and has the value [TRUE]:
           1. Set `hmac-secret` to [TRUE] in the authenticator extensions input.
           1. Set {{AuthenticationExtensionsClientOutputs/prfEnabled}} to the value of `hmac-secret` in the authenticator extensions output. If not present, set {{AuthenticationExtensionsClientOutputs/prfEnabled}} to [FALSE].
       1. Else:
           1. Set {{AuthenticationExtensionsClientOutputs/prfEnabled}} to [FALSE].
```

index.bs
```diff
@@ -5638,6 +5638,93 @@ At this time, one [=credential property=] is defined: the [=resident key credent
...
+: Client extension output ([=registration extension|registration=])
+:: <xmp class="idl">
+    partial dictionary AuthenticationExtensionsClientOutputs {
+        boolean prfEnabled;
+    };
+    </xmp>
+
+    <div dfn-type="dict-member" dfn-for="AuthenticationExtensionsClientOutputs">
+        :   <dfn>prfEnabled</dfn>
+        ::  True iff the two PRFs are available for use with the created credential.
```
emlun, line 5688:
```suggestion
        ::  [TRUE] iff the two PRFs are available for use with the created credential.
```

index.bs
```diff
@@ -5638,6 +5638,93 @@ At this time, one [=credential property=] is defined: the [=resident key credent
...
+: Client extension input ([=authentication extension|authentication=])
+::  <xmp class="idl">
+    partial dictionary AuthenticationExtensionsClientInputs {
+        sequence<DOMString> prfEval;
+    };
+    </xmp>
+
+    <div dfn-type="dict-member" dfn-for="AuthenticationExtensionsClientInputs">
+        :   <dfn>prfEnable</dfn>
+        ::  A sequence of one or two strings that are inputs on which to evaluate the PRF.
```
emlun, line 5702: How about encoding that in the type? This is also how CTAP does it in `hmac-secret`.

```
    partial dictionary AuthenticationExtensionsClientInputs {
        required DOMString prfEval1;
        DOMString prfEval2;
    };
```

index.bs
```diff
@@ -5638,6 +5638,93 @@ At this time, one [=credential property=] is defined: the [=resident key credent
...
+    </div>
+
+: Client extension processing ([=authentication extension|authentication=])
+::
+       1. If {{AuthenticationExtensionsClientInputs/prfEval}} is present and non-empty:
+           1. If the length of {{AuthenticationExtensionsClientInputs/prfEval}} is greater than two, return a {{DOMException}} whose name is “{{NotSupportedError}}”.
+           1. Let `salt1` be the value of <kbd>SHA-256(UTF8Encode("WebAuthn PRF\x00" + {{AuthenticationExtensionsClientInputs/prfEval}}[0]))</kbd>.
```
emlun, line 5709: I'm guessing this is meant to express `SHA256(0x576562417574686e2050524600 || prfEval)`, but the fact the `\x00` is inside `UTF8Encode` might be a bit confusing? I'd probably write this as `SHA-256(UTF8Encode("WebAuthn PRF") || 0x00 || prfEval[0]))` just to be safe.

Also, why `<kdb>` instead of `<code>`?

index.bs
```diff
@@ -5638,6 +5638,93 @@ At this time, one [=credential property=] is defined: the [=resident key credent
...
+           1. If the length of {{AuthenticationExtensionsClientInputs/prfEval}} is two, let `salt2` be the value of <kbd>SHA-256(UTF8Encode("WebAuthn PRF\x00" + {{AuthenticationExtensionsClientInputs/prfEval}}[1]))</kbd>.
+           1. Send an `hmac-secret` extension to the [=authenticator=] using the values of `salt1` and, if set, `salt2` as the parameters of the same name in that process.
+           1. Decrypt the extension result and set {{AuthenticationExtensionsClientOutputs/prfOutputs}} to be a one- or two-element sequence containing the result(s).
+
+: Client extension output ([=authentication extension|authentication=])
+:: <xmp class="idl">
+    partial dictionary AuthenticationExtensionsClientOutputs {
+        sequence<ArrayBuffer> prfOutputs;
```
emlun, line 5717: The above comment about types applies here too.

akshayku:
User verification choice is important in hmac-secret extension. I think it needs more clarification regarding implications of it.

If an RP uses this extension with UV Discouraged for now and then in future tries to go with UV Required, that transition is tricky. First they have to call webauthn get() using UV discouraged, decrypt their state, then call another webauthn get() call with UV required so that they can encrypt it properly. 

RPs should not do UV Preferred as that is dependent on authenticator. 

Probably, It's better that RPs make up their mind regarding UV before using it. If they always want to remain one way all the time, no issues. 

If use cases are with UV required only, may be it is easiest to have this extension applicable only to UVRequired case?

index.bs
agl: Done.

index.bs
agl: Done.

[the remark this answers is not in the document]
agl: If, in the future, we did v2 of this, one might otherwise do "WebAuthn PRF2" + input. However, that's then potentially ambiguous with an input that previously started with a 2. The NUL byte ensures that doesn't happen.

index.bs
```diff
@@ -5638,6 +5638,93 @@ At this time, one [=credential property=] is defined: the [=resident key credent
 : Authenticator extension output
 :: None.
 
+## Pseudo-random function extension (<dfn>prf</dfn>) ## {#prf-extension}
+
+This [=client extension|client=] [=registration extension=] and [=authentication extension=] allows a [=[RP]=] to evaluate outputs from one of two secret, pseudo-random functions (PRFs) associated with a [=credential=]. A pseudo-random function, F, is such that, given x<sub>1</sub>, x<sub>2</sub>, &hellip; x<sub>n</sub>, and F(x<sub>1</sub>), F(x<sub>2</sub>), &hellip; F(x<sub>n</sub>), no adversary can predict any bit of F(x<sub>m</sub>) for any m &gt; n. The PRFs provided by this extension map from {{DOMString}}s of any length to 32-byte {{ArrayBuffer}}s.
+
+If a [=credential=] is created on an [=authenticator=] that supports this extension then two fresh PRFs are associated with it. One is evaluated when [=user verification=] is performed and the other is evaluated when [=user verification=] is not performed. It is the responsibility of the [=[RP]=] to set the {{PublicKeyCredentialRequestOptions/userVerification}} parameter accordingly.
```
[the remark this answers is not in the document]
agl: Integrated, but have changed some wording. PTAL.

And thanks for https://github.com/fido-alliance/fido-2-specs/issues/873. I didn't notice that CTAP2.0 was different and I've updated the wording here accordingly.

index.bs
```diff
@@ -5638,6 +5638,93 @@ At this time, one [=credential property=] is defined: the [=resident key credent
...
+
+As a motivating example, PRF outputs could be used as symmetric keys to encrypt user data. Such encrypted data would be inaccessible without the ability to get assertions from the associated [=credential=]. By using the provision below to evaluate the PRF at two points in a single [=assertion=] operation, the encryption key could be periodically rotated during [=assertions=] without additional user interaction. If the evaluation inputs were made unpredictable then even an attacker who could satisfy [=user verification=], and who had time-limited access to the authenticator, could not learn the encryption key without also knowing the correct PRF input.
```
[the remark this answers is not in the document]
agl: I think the quoted text is accurate? I'm not sure what you think is wrong, or is it just too terse?

index.bs
```diff
@@ -5638,6 +5638,93 @@ At this time, one [=credential property=] is defined: the [=resident key credent
...
+
+This extension is implemented on top of the [[FIDO-CTAP]] `hmac-secret` extension. It is a separate [=client extension=] because `hmac-secret` requires that inputs and outputs be encrypted in a manner that only the user agent can perform, and to provide domain separation between uses by WebAuthn and any uses by the underlying platform. The latter is achieved by hashing the provided PRF inputs with a context string to prevent evaluation of the PRFs for arbitrary inputs.
```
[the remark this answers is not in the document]
agl: Have deleted “domain” to avoid confusion with the other meaning used several times in the spec.

index.bs
```diff
@@ -5638,6 +5638,93 @@ At this time, one [=credential property=] is defined: the [=resident key credent
 : Authenticator extension output
 :: None.
 
+## Pseudo-random function extension (<dfn>prf</dfn>) ## {#prf-extension}
+
+This [=client extension|client=] [=registration extension=] and [=authentication extension=] allows a [=[RP]=] to evaluate outputs from one of two secret, pseudo-random functions (PRFs) associated with a [=credential=]. A pseudo-random function, F, is such that, given x<sub>1</sub>, x<sub>2</sub>, &hellip; x<sub>n</sub>, and F(x<sub>1</sub>), F(x<sub>2</sub>), &hellip; F(x<sub>n</sub>), no adversary can predict any bit of F(x<sub>m</sub>) for any m &gt; n. The PRFs provided by this extension map from {{DOMString}}s of any length to 32-byte {{ArrayBuffer}}s.
```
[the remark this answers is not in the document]
agl: Have deleted the definition of a PRF.

index.bs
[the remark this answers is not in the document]
equalsJeffH: hm, perhaps too terse.  I suppose I do not understand what you mean by "...evaluate the PRF at two points in a single [=assertion=] operation..."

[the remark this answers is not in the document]
agl: Have expanded on that wording a little in latest commit.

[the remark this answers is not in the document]
equalsJeffH: thx, changing "two points" to "two inputs" did the trick.

index.bs
```diff
@@ -5638,6 +5639,107 @@ At this time, one [=credential property=] is defined: the [=resident key credent
...
+       1. If {{AuthenticationExtensionsClientInputs/prfEnable}} is present and has the value [TRUE]:
+           1. If {{AuthenticatorSelectionCriteria/userVerification}} has the value {{UserVerificationRequirement/preferred}}:
+               1. Return a {{DOMException}} whose name is “{{NotSupportedError}}”
+           1. Set `hmac-secret` to [TRUE] in the authenticator extensions input.
+           1. If {{AuthenticationExtensionsClientInputs/prfEval}} is present and a future extension to [[FIDO-CTAP]] permits evaluation of the PRF at creation time, configure `hmac-secret` inputs accordingly:
+               * Let `salt1` be the value of <code>SHA-256(UTF8Encode("WebAuthn PRF") || 0x00 || UTF8Encode({{AuthenticationExtensionsPRFInputs/input1}}))</code>.
+               * If {{AuthenticationExtensionsPRFInputs/input2}} is present, let `salt2` be the value of <code>SHA-256(UTF8Encode("WebAuthn PRF" || 0x00 || UTF8Encode({{AuthenticationExtensionsPRFInputs/input2}}))</code>.
```
emlun, line 5695: A parenthesis went missing. :) (that may have been my fault)

```suggestion
               * If {{AuthenticationExtensionsPRFInputs/input2}} is present, let `salt2` be the value of <code>SHA-256(UTF8Encode("WebAuthn PRF") || 0x00 || UTF8Encode({{AuthenticationExtensionsPRFInputs/input2}}))</code>.
```

index.bs
```diff
@@ -5638,6 +5639,107 @@ At this time, one [=credential property=] is defined: the [=resident key credent
...
+
+        :   <dfn>prfOutputs</dfn>
+        ::  The results of evaluating the PRF for the inputs given in {{AuthenticationExtensionsClientInputs/prfEval}}. Outputs may not be available during [=registration=]; see comments in {{AuthenticationExtensionsClientInputs/prfEval}}.
+    </div>
+
+: Client extension input ([=authentication extension|authentication=])
+::  <xmp class="idl">
+    partial dictionary AuthenticationExtensionsClientInputs {
+        record<USVString, AuthenticationExtensionsPRFInputs> prfEvalByCredential;
+    };
```
emlun, line 5719: This member also needs to be named just `prf`. I think it can remain a `record`, but maybe there shouldn't be both this and the registration extension input with the same name but different types? @selfissued what do you think?

index.bs
```diff
@@ -5638,6 +5639,107 @@ At this time, one [=credential property=] is defined: the [=resident key credent
...
+        ::  A record mapping [=credential IDs=] to PRF inputs to evaluate for that credential.
+    </div>
+
+
+: Client extension processing ([=authentication extension|authentication=])
+::
+       1. If {{AuthenticationExtensionsClientInputs/prfEvalByCredential}} is present and {{PublicKeyCredentialRequestOptions/allowCredentials}} is empty, return a {{DOMException}} whose name is “{{NotSupportedError}}”.
+       1. If either {{AuthenticationExtensionsClientInputs/prfEvalByCredential}} is present or {{AuthenticationExtensionsClientInputs/prfEval}} is present, and {{PublicKeyCredentialRequestOptions/userVerification}} has the value {{UserVerificationRequirement/preferred}}, return a {{DOMException}} whose name is “{{NotSupportedError}}”.
+       1. Find any applicable PRF input(s):
+           1. If {{AuthenticationExtensionsClientInputs/prfEvalByCredential}} is present and contains an entry whose key is the [=base64url encoding=] of the [=credential ID=] that will be returned, use the value of that entry.
```
emlun, line 5733: >If `prfEvalByCredential` [...] contains an entry whose key is the [...] credential ID that will be returned [...]

Does the client know this at this time? Or will the client issue `authenticatorGetCredential(up=0)` requests until it finds one (or exhausts the options)?

emlun, line 5733: >[...] whose key is the [=base64url encoding=] of the [=credential ID=] [...]

We use `BufferSource`s for credential IDs elsewhere, so I'm thinking we should perhaps do that here too for consistency. `BufferSource` isn't a valid type for [record](https://heycam.github.io/webidl/#idl-record) keys, though, but [`ByteString`](https://heycam.github.io/webidl/#idl-ByteString) is. I don't know whether that's good enough.

index.bs
```diff
@@ -5638,6 +5639,107 @@ At this time, one [=credential property=] is defined: the [=resident key credent
...
+
+As a motivating example, PRF outputs could be used as symmetric keys to encrypt user data. Such encrypted data would be inaccessible without the ability to get assertions from the associated [=credential=]. By using the provision below to evaluate the PRF at two inputs in a single [=assertion=] operation, the encryption key could be periodically rotated during [=assertions=] by choosing a fresh, random input and reencrypting under the new output. If the evaluation inputs are unpredictable then even an attacker who could satisfy [=user verification=], and who had time-limited access to the authenticator, could not learn the encryption key without also knowing the correct PRF input.
+
+This extension is implemented on top of the [[FIDO-CTAP]] `hmac-secret` extension. It is a separate [=client extension=] because `hmac-secret` requires that inputs and outputs be encrypted in a manner that only the user agent can perform, and to provide separation between uses by WebAuthn and any uses by the underlying platform. The latter is achieved by hashing the provided PRF inputs with a context string to prevent evaluation of the PRFs for arbitrary inputs.
```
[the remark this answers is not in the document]
agl: Done.

index.bs
```diff
@@ -5638,6 +5639,107 @@ At this time, one [=credential property=] is defined: the [=resident key credent
...
+
+: Extension identifier
+:: `prf`
+
+: Operation applicability
+:: [=registration extension|Registration=] and [=authentication extension|authentication=]
+
+: Client extension input ([=registration extension|registration=])
+::  <xmp class="idl">
+    dictionary AuthenticationExtensionsPRFInputs {
```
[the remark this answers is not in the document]
agl: I had look over other WebIDL and there seem to be quite a lot of cases of upper-case acronyms in dictionary names: GPU, MIDI, USB, RTC, IIR, NDEF, CSS, UI, … Given that, I think upper-case is the WebIDL style.

index.bs
```diff
@@ -5638,6 +5639,107 @@ At this time, one [=credential property=] is defined: the [=resident key credent
...
+
+    dictionary AuthenticationExtensionsPRFOutputs {
+        required ArrayBuffer output1;
+        ArrayBuffer output2;
+    };
+
+    partial dictionary AuthenticationExtensionsClientInputs {
+        boolean prfEnable;
+        AuthenticationExtensionsPRFInputs prfEval;
+    };
```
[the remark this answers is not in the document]
agl: Done.

index.bs
```diff
@@ -5638,6 +5639,107 @@ At this time, one [=credential property=] is defined: the [=resident key credent
...
+               * If {{AuthenticationExtensionsPRFInputs/input2}} is present, let `salt2` be the value of <code>SHA-256(UTF8Encode("WebAuthn PRF" || 0x00 || UTF8Encode({{AuthenticationExtensionsPRFInputs/input2}}))</code>.
+           1. Set {{AuthenticationExtensionsClientOutputs/prfEnabled}} to the value of `hmac-secret` in the authenticator extensions output. If not present, set {{AuthenticationExtensionsClientOutputs/prfEnabled}} to [FALSE].
+           1. Set {{AuthenticationExtensionsClientOutputs/prfOutputs}} to the PRF outputs, if any.
+
+: Client extension output ([=registration extension|registration=])
+:: <xmp class="idl">
+    partial dictionary AuthenticationExtensionsClientOutputs {
+        boolean prfEnabled;
+        AuthenticationExtensionsPRFOutputs prfOutputs;
+    };
```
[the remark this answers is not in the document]
agl: Done.

index.bs
agl: Have defined  a dictionary, AuthenticationExtensionsPRFInputs, to hold the inputs. The member of AuthenticationExtensionsClientInputs is now just called `prf` and is an instance of that dictionary.

index.bs
agl: The client will need to batch up=0 requests in order to find the credential that will be asserted first. I think clients already need to support this in practice so I don't anticipate it being a problem.

agl: I pondered this and went with using base64url-encoded IDs as the keys, although I don't have strong opinions. As you note, we can't have BufferSources as keys. ByteStrings work, but then we're adding a 3rd representation of credential IDs, while we already have the base64url ones in the interface.

index.bs
```diff
@@ -5638,6 +5639,97 @@ At this time, one [=credential property=] is defined: the [=resident key credent
...
+               1. Return a {{DOMException}} whose name is “{{NotSupportedError}}”.
+           1. Set `hmac-secret` to [TRUE] in the authenticator extensions input.
+           1. If {{AuthenticationExtensionsPRFInputs/eval}} is present and a future extension to [[FIDO-CTAP]] permits evaluation of the PRF at creation time, configure `hmac-secret` inputs accordingly:
+               * Let `salt1` be the value of <code>SHA-256(UTF8Encode("WebAuthn PRF") || 0x00 || {{AuthenticationExtensionsPRFValues/first}})</code>.
+               * If {{AuthenticationExtensionsPRFValues/second}} is present, let `salt2` be the value of <code>SHA-256(UTF8Encode("WebAuthn PRF") || 0x00 || {{AuthenticationExtensionsPRFValues/second}})</code>.
+           1. Set {{AuthenticationExtensionsPRFOutputs/enabled}} to the value of `hmac-secret` in the authenticator extensions output. If not present, set {{AuthenticationExtensionsPRFOutputs/enabled}} to [FALSE].
+           1. Set {{AuthenticationExtensionsPRFOutputs/results}} to the PRF outputs, if any.
+
+: Client extension output ([=registration extension|registration=])
```
emlun, line 5698:
```suggestion
: Client extension output
```
Maybe also move this section to after both "client extension processing" sections?

index.bs
```diff
@@ -5638,6 +5639,97 @@ At this time, one [=credential property=] is defined: the [=resident key credent
...
+: Client extension processing ([=registration extension|registration=])
+::
+       1. If {{AuthenticationExtensionsClientInputs/prf}} is present:
+           1. If {{AuthenticationExtensionsPRFInputs/evalByCredential}} is present, return a {{DOMException}} whose name is “{{NotSupportedError}}”.
+           1. If {{AuthenticatorSelectionCriteria/userVerification}} has the value {{UserVerificationRequirement/preferred}}:
+               1. Return a {{DOMException}} whose name is “{{NotSupportedError}}”.
+           1. Set `hmac-secret` to [TRUE] in the authenticator extensions input.
+           1. If {{AuthenticationExtensionsPRFInputs/eval}} is present and a future extension to [[FIDO-CTAP]] permits evaluation of the PRF at creation time, configure `hmac-secret` inputs accordingly:
+               * Let `salt1` be the value of <code>SHA-256(UTF8Encode("WebAuthn PRF") || 0x00 || {{AuthenticationExtensionsPRFValues/first}})</code>.
+               * If {{AuthenticationExtensionsPRFValues/second}} is present, let `salt2` be the value of <code>SHA-256(UTF8Encode("WebAuthn PRF") || 0x00 || {{AuthenticationExtensionsPRFValues/second}})</code>.
```
emlun, line 5694: Should probably use the `eval.` prefix to be precise?
```suggestion
               * Let `salt1` be the value of <code>SHA-256(UTF8Encode("WebAuthn PRF") || 0x00 || {{AuthenticationExtensionsPRFInputs/eval}}.{{AuthenticationExtensionsPRFValues/first}})</code>.
               * If <code>{{AuthenticationExtensionsPRFInputs/eval}}.{{AuthenticationExtensionsPRFValues/second}}</code> is present, let `salt2` be the value of <code>SHA-256(UTF8Encode("WebAuthn PRF") || 0x00 || {{AuthenticationExtensionsPRFInputs/eval}}.{{AuthenticationExtensionsPRFValues/second}})</code>.
```

index.bs
```diff
@@ -5638,6 +5639,97 @@ At this time, one [=credential property=] is defined: the [=resident key credent
...
+       1. If either {{AuthenticationExtensionsPRFInputs/evalByCredential}} is present or {{AuthenticationExtensionsPRFInputs/eval}} is present, and {{PublicKeyCredentialRequestOptions/userVerification}} has the value {{UserVerificationRequirement/preferred}}, return a {{DOMException}} whose name is “{{NotSupportedError}}”.
+       1. Find any applicable PRF input(s):
+           1. If {{AuthenticationExtensionsPRFInputs/evalByCredential}} is present and contains an entry whose key is the [=base64url encoding=] of the [=credential ID=] that will be returned, use the value of that entry.
+           1. If no input(s) were found in the previous step, and if {{AuthenticationExtensionsPRFInputs/eval}} is present, then use that value.
+       1. If PRF input(s) were found:
+           1. Let `salt1` be the value of <code>SHA-256(UTF8Encode("WebAuthn PRF") || 0x00 || {{AuthenticationExtensionsPRFValues/first}})</code>.
+           1. If {{AuthenticationExtensionsPRFValues/second}} is present, let `salt2` be the value of <code>SHA-256(UTF8Encode("WebAuthn PRF") || 0x00 || {{AuthenticationExtensionsPRFValues/second}})</code>.
```
emlun, line 5727: I think a prefix to precisely disambiguate between `eval` and `evalByCredential` could be good here too. There are also a few map keywords we can reference here.

```suggestion
       1. Let |ev| be null, and try to find any applicable PRF input(s):
           1. If {{AuthenticationExtensionsPRFInputs/evalByCredential}} is present and [=map/exists|contains=] an [=map/entry=] whose [=map/key=] is the [=base64url encoding=] of the [=credential ID=] that will be returned, let |ev| be the [=map/value=] of that entry.
           1. If |ev| is null and {{AuthenticationExtensionsPRFInputs/eval}} is present, then let |ev| be the value of {{AuthenticationExtensionsPRFInputs/eval}}.
       1. If |ev| is not null:
           1. Let `salt1` be the value of <code>SHA-256(UTF8Encode("WebAuthn PRF") || 0x00 || |ev|.{{AuthenticationExtensionsPRFValues/first}})</code>.
           1. If <code>|ev|.{{AuthenticationExtensionsPRFValues/second}}</code> is present, let `salt2` be the value of <code>SHA-256(UTF8Encode("WebAuthn PRF") || 0x00 || |ev|.{{AuthenticationExtensionsPRFValues/second}})</code>.
```

index.bs
```diff
@@ -5638,6 +5639,97 @@ At this time, one [=credential property=] is defined: the [=resident key credent
...
+    </div>
+
+: Client extension processing ([=authentication extension|authentication=])
+::
+       1. If {{AuthenticationExtensionsPRFInputs/evalByCredential}} is present and {{PublicKeyCredentialRequestOptions/allowCredentials}} is empty, return a {{DOMException}} whose name is “{{NotSupportedError}}”.
+       1. If either {{AuthenticationExtensionsPRFInputs/evalByCredential}} is present or {{AuthenticationExtensionsPRFInputs/eval}} is present, and {{PublicKeyCredentialRequestOptions/userVerification}} has the value {{UserVerificationRequirement/preferred}}, return a {{DOMException}} whose name is “{{NotSupportedError}}”.
+       1. Find any applicable PRF input(s):
+           1. If {{AuthenticationExtensionsPRFInputs/evalByCredential}} is present and contains an entry whose key is the [=base64url encoding=] of the [=credential ID=] that will be returned, use the value of that entry.
+           1. If no input(s) were found in the previous step, and if {{AuthenticationExtensionsPRFInputs/eval}} is present, then use that value.
+       1. If PRF input(s) were found:
```
emlun, line 5725: I think we should explicitly define what should happen if no inputs were found. I read this to mean the client should then return an empty map as the extension output, but I imagine it could also be interpreted as to omit the extension output if it would be empty.

index.bs
```diff
@@ -5638,6 +5639,97 @@ At this time, one [=credential property=] is defined: the [=resident key credent
...
+    };
+
+    partial dictionary AuthenticationExtensionsClientOutputs {
+        AuthenticationExtensionsPRFOutputs prf;
+    };
+    </xmp>
+
+    <div dfn-type="dict-member" dfn-for="AuthenticationExtensionsPRFOutputs">
+        :   <dfn>enabled</dfn>
+        ::  [TRUE] if, and only if, the one or two PRFs are available for use with the created credential.
```
emlun, line 5712: Maybe worth calling out that this is only present in registration ceremonies?

agl: Done.

index.bs
agl: Good point, thanks. I've altered things to be explicit that not finding an input gives `prf: {}` in the output extensions. (I.e. the RP can tell the difference between the PRF extension not being supported, and a suitable input not being found.)

index.bs
```diff
@@ -5638,6 +5639,100 @@ At this time, one [=credential property=] is defined: the [=resident key credent
...
+    </div>
+
+: Client extension processing ([=authentication extension|authentication=])
+::
+       1. If {{AuthenticationExtensionsPRFInputs/evalByCredential}} is present and {{PublicKeyCredentialRequestOptions/allowCredentials}} is empty, return a {{DOMException}} whose name is “{{NotSupportedError}}”.
+       1. If either {{AuthenticationExtensionsPRFInputs/evalByCredential}} is present or {{AuthenticationExtensionsPRFInputs/eval}} is present, and {{PublicKeyCredentialRequestOptions/userVerification}} has the value {{UserVerificationRequirement/preferred}}, return a {{DOMException}} whose name is “{{NotSupportedError}}”.
+       1. Let |ev| be null, and try to find any applicable PRF input(s):
+           1. If {{AuthenticationExtensionsPRFInputs/evalByCredential}} is present and [=map/exists|contains=] an [=map/entry=] whose [=map/key=] is the [=base64url encoding=] of the [=credential ID=] that will be returned, let |ev| be the [=map/value=] of that entry.
+           1. If |ev| is null and {{AuthenticationExtensionsPRFInputs/eval}} is present, then let |ev| be the value of {{AuthenticationExtensionsPRFInputs/eval}}.
```
akshayku, line 5724: For `AuthenticationExtensionsPRFInputs/eval`, RP will be using the same nonce/salts for all its users?

agl: The purpose of `eval` is to serve RPs that are using an empty allowlist. In that case, if they want single-touch operation, then they'll have to use the same salts for every user. I think that's a reasonable thing to do since they'll be doing UV too? That was my thinking anyway.

akshayku: For different credentials for same RP on that same authenticator, PRF value will be different as credRandoms are different for those credentials. So that should not be a linkability issue. 

If attacker gets hold of users authenticator, and if attacker also knows the PIN, then PRF secret is revealed. Which is not the case with per-credential nonces/salts. But, user's account is already compromised in that case and user should revoke that credential asap. RP has to figure out what to do with material encrypted with the secret though. 

That's what I was thinking. Things will work. It just needs more thought.

index.bs
```diff
@@ -5638,6 +5639,99 @@ At this time, one [=credential property=] is defined: the [=resident key credent
...
+        :   <dfn>eval</dfn>
+        ::  One or two inputs on which to evaluate PRF. Not all [=authenticators=] support evaluating the PRFs during credential creation so outputs may, or may not, be provided. If not, then an [=assertion=] is needed in order to obtain the outputs.
+
+        :   <dfn>evalByCredential</dfn>
+        ::  A record mapping [=base64url encoding|base64url encoded=] [=credential IDs=] to PRF inputs to evaluate for that credential. Only applicable during [=assertions=] when {{PublicKeyCredentialRequestOptions/allowCredentials}} is not empty.
+    </div>
+
+: Client extension processing ([=registration extension|registration=])
+::
+       1. If {{AuthenticationExtensionsClientInputs/prf}} is present:
```
emlun, line 5687: I think this check is redundant, because you would never begin extension processing if `prf` is not present?

index.bs
```diff
@@ -5638,6 +5639,99 @@ At this time, one [=credential property=] is defined: the [=resident key credent
...
+        :   <dfn>enabled</dfn>
+        ::  [TRUE] if, and only if, the one or two PRFs are available for use with the created credential. This is only reported during [=registration=] and is not present in the case of [=authentication=].
+
+        :   <dfn>results</dfn>
+        ::  The results of evaluating the PRF for the inputs given in {{AuthenticationExtensionsPRFInputs/eval}} or {{AuthenticationExtensionsPRFInputs/evalByCredential}}. Outputs may not be available during [=registration=]; see comments in {{AuthenticationExtensionsPRFInputs/eval}}.
+    </div>
+
+: Client extension processing ([=authentication extension|authentication=])
+::
+       1. If {{AuthenticationExtensionsClientInputs/prf}} is present and either {{AuthenticationExtensionsPRFInputs/eval}} is present or {{AuthenticationExtensionsPRFInputs/evalByCredential}} is not empty:
```
emlun, line 5720: Like above, I think the check on `prf` here is also redundant.

index.bs
```diff
@@ -5638,6 +5639,99 @@ At this time, one [=credential property=] is defined: the [=resident key credent
...
+           1. If {{PublicKeyCredentialRequestOptions/userVerification}} has the value {{UserVerificationRequirement/preferred}}, return a {{DOMException}} whose name is “{{NotSupportedError}}”.
+           1. If {{AuthenticationExtensionsPRFInputs/evalByCredential}} is not empty but {{PublicKeyCredentialRequestOptions/allowCredentials}} is empty, return a {{DOMException}} whose name is “{{NotSupportedError}}”.
+           1. Set {{AuthenticationExtensionsClientOutputs/prf}} to an empty dictionary.
```
emlun, line 5723: Hm. This `prf` and the one above (although it might get removed) appear identical unless you inspect the links, so a disambiguation might be good. Might also be good to hint that the contents will likely change in the next few steps? So maybe something like this?

```suggestion
           1. Initialize the {{AuthenticationExtensionsClientOutputs/prf}} extension output to an empty dictionary.
```

index.bs
```diff
@@ -5638,6 +5639,99 @@ At this time, one [=credential property=] is defined: the [=resident key credent
...
+               * If <code>{{AuthenticationExtensionsPRFInputs/eval}}.{{AuthenticationExtensionsPRFValues/second}}</code> is present, let `salt2` be the value of <code>SHA-256(UTF8Encode("WebAuthn PRF") || 0x00 || {{AuthenticationExtensionsPRFInputs/eval}}.{{AuthenticationExtensionsPRFValues/second}})</code>.
+           1. Set {{AuthenticationExtensionsPRFOutputs/enabled}} to the value of `hmac-secret` in the authenticator extensions output. If not present, set {{AuthenticationExtensionsPRFOutputs/enabled}} to [FALSE].
+           1. Set {{AuthenticationExtensionsPRFOutputs/results}} to the decrypted PRF result(s), if any.
+
+Note: If PRF results are obtained during [=registration=] then [=[RPS]=] MUST inspect the [=UV=] bit in the [=flags=] of the response in order to determine the correct value of {{PublicKeyCredentialRequestOptions/userVerification}} for future [=assertions=]. Otherwise results from [=assertions=] may be inconsistent with those from the [=registration=].
+
+: Client extension output
```
emlun, line 5698: Please move this section to after the **Client extension processing (authentication)** section.

index.bs
```diff
@@ -5638,6 +5639,99 @@ At this time, one [=credential property=] is defined: the [=resident key credent
...
+               * Let `salt1` be the value of <code>SHA-256(UTF8Encode("WebAuthn PRF") || 0x00 || {{AuthenticationExtensionsPRFInputs/eval}}.{{AuthenticationExtensionsPRFValues/first}})</code>.
+               * If <code>{{AuthenticationExtensionsPRFInputs/eval}}.{{AuthenticationExtensionsPRFValues/second}}</code> is present, let `salt2` be the value of <code>SHA-256(UTF8Encode("WebAuthn PRF") || 0x00 || {{AuthenticationExtensionsPRFInputs/eval}}.{{AuthenticationExtensionsPRFValues/second}})</code>.
+           1. Set {{AuthenticationExtensionsPRFOutputs/enabled}} to the value of `hmac-secret` in the authenticator extensions output. If not present, set {{AuthenticationExtensionsPRFOutputs/enabled}} to [FALSE].
+           1. Set {{AuthenticationExtensionsPRFOutputs/results}} to the decrypted PRF result(s), if any.
+
+Note: If PRF results are obtained during [=registration=] then [=[RPS]=] MUST inspect the [=UV=] bit in the [=flags=] of the response in order to determine the correct value of {{PublicKeyCredentialRequestOptions/userVerification}} for future [=assertions=]. Otherwise results from [=assertions=] may be inconsistent with those from the [=registration=].
```
emlun, line 5696: Very minor nit - you're welcome to ignore it if you like.

```suggestion
Note: If PRF results are obtained during [=registration=] then the [=[RP]=] MUST inspect the [=UV=] bit in the [=flags=] of the response in order to determine the correct value of {{PublicKeyCredentialRequestOptions/userVerification}} for future [=assertions=]. Otherwise results from [=assertions=] may be inconsistent with those from the [=registration=].
```

index.bs
agl: Done.

index.bs
agl: I don't mind being more specific but it is the case that other extensions assume this check. Thus have removed it in both cases to align.

index.bs
agl: (Removed as above.)

index.bs
```diff
@@ -5638,6 +5639,98 @@ At this time, one [=credential property=] is defined: the [=resident key credent
...
+
+Note: If PRF results are obtained during [=registration=] then the [=[RP]=] MUST inspect the [=UV=] bit in the [=flags=] of the response in order to determine the correct value of {{PublicKeyCredentialRequestOptions/userVerification}} for future [=assertions=]. Otherwise results from [=assertions=] may be inconsistent with those from the [=registration=].
+
+: Client extension processing ([=authentication extension|authentication=])
+::
+       1. If either {{AuthenticationExtensionsPRFInputs/eval}} is present or {{AuthenticationExtensionsPRFInputs/evalByCredential}} is not empty:
+           1. If {{PublicKeyCredentialRequestOptions/userVerification}} has the value {{UserVerificationRequirement/preferred}}, return a {{DOMException}} whose name is “{{NotSupportedError}}”.
+           1. If {{AuthenticationExtensionsPRFInputs/evalByCredential}} is not empty but {{PublicKeyCredentialRequestOptions/allowCredentials}} is empty, return a {{DOMException}} whose name is “{{NotSupportedError}}”.
+           1. Initialize the {{AuthenticationExtensionsClientOutputs/prf}} extension output to an empty dictionary.
```
emlun, line 5702: Hm, this step was meant to always run, wasn't it? In that case it would have to move to before the parent step. But on the other hand, I think that check ("If either eval is present or evalByCredential is not empty") is now also unnecessary? With the new logic below I think that if `eval` is absent and `evalByCredential` is empty, then _ev_ will just end up null and the output will remain empty (but present).

agl: This was deliberate: echoing an extension at assertion time means that the user agent processed the extension, tried, but wasn't able to get a response. (Maybe the authenticator doesn't support hmac_secret.)

But if the browser didn't find any inputs then that's not true. I.e. it's an RP error, and echoing the extension like everything is ok would be misleading.

emlun: Ok, but in that case, shouldn't some "invalid input" error be raised instead? [§9.2. Defining Extensions](https://w3c.github.io/webauthn/#sctn-extension-specification) reads:

>Any client extension that is processed by the client MUST return a client extension output value so that the WebAuthn Relying Party knows that the extension was honored by the client.

So if the extension output is missing, that generally means the extension was ignored or not supported by either the browser or authenticator, which would not be the case here. The `appidExclude` extension goes against this, but we should probably change that since the spec is not internally consistent as is...

agl: > Ok, but in that case, shouldn't some "invalid input" error be raised instead?

That would be fine too, but it would preclude adding anything else in this extension in the future unless it always comes with inputs in the existing members. I'm ok with that, I just didn't see a reason to do it.

> So if the extension output is missing, that generally means the extension was ignored or not supported by either the browser or authenticator, which would not be the case here.

But that _is_ the case here. The RP provided no inputs so the extension was ignored.

> The appidExclude extension goes against this, but we should probably change that

appidExclude shouldn't have an output because there's nothing the RP can do with that information. I'm sure that paragraph in 9.2 made rough sense at the time, but it hasn't proven to be true and I think we should delete it.

emlun: >But that is the case here. The RP provided no inputs so the extension was ignored.

Ok, I see what you mean. Fair enough.

Let's track the appidExclude thing in #1445.

index.bs
```diff
@@ -5638,6 +5639,97 @@ At this time, one [=credential property=] is defined: the [=resident key credent
...
+
+As a motivating example, PRF outputs could be used as symmetric keys to encrypt user data. Such encrypted data would be inaccessible without the ability to get assertions from the associated [=credential=]. By using the provision below to evaluate the PRF at two inputs in a single [=assertion=] operation, the encryption key could be periodically rotated during [=assertions=] by choosing a fresh, random input and reencrypting under the new output. If the evaluation inputs are unpredictable then even an attacker who could satisfy [=user verification=], and who had time-limited access to the authenticator, could not learn the encryption key without also knowing the correct PRF input.
```
Firstyear, line 5648: I think that there should be a clarification about the exact process to perform a key rotation as this section is hard to parse.

index.bs
```diff
@@ -5638,6 +5639,97 @@ At this time, one [=credential property=] is defined: the [=resident key credent
...
+
+When requested, if a [=credential=] is created on an [=authenticator=] that supports this extension then either one or two freshly-seeded PRFs are associated with it. If two PRFs are associated then, when outputs are requested, one PRF is evaluated if [=user verification=] is performed and the other is evaluated if only a [=test of user presence|user presence test=] is performed. It is the responsibility of the [=[RP]=] to set the {{PublicKeyCredentialRequestOptions/userVerification}} parameter accordingly: [=[RPS]=] SHOULD *consistently* use either the {{UserVerificationRequirement/required}} or {{UserVerificationRequirement/discouraged}} values of {{UserVerificationRequirement}} when using this extension, otherwise the outputs may vary for a given input. To avoid mistakes, the default value of {{UserVerificationRequirement/preferred}} is prohibited when using this extension during [=assertion=] because that could cause the PRF used to vary between operations if [=user verification=] is later enabled on an [=authenticator=].
```
Firstyear, line 5646: It's not clear if the "first" or "second" prf is the one returned during this exension. This must be clarified to state which of "first" or "second" are returned, and in which userverification scenarios related to the registration process.

index.bs
```diff
@@ -5638,6 +5639,97 @@ At this time, one [=credential property=] is defined: the [=resident key credent
 : Authenticator extension output
 :: None.
 
+## Pseudo-random function extension (<dfn>prf</dfn>) ## {#prf-extension}
+
+This [=client extension|client=] [=registration extension=] and [=authentication extension=] allows a [=[RP]=] to evaluate outputs from a pseudo-random function (PRF) associated with a [=credential=]. The PRFs provided by this extension map from {{USVString}}s of any length to 32-byte {{ArrayBuffer}}s.
```
Firstyear, line 5644: This introduction needs clarification. It does not clearly or accurately describe *what* this extension does. It appears to read as a "generate a random number on this device". It's also unclear if the operation is yielding the prf content, or is actually signing with hmac in the device, and I think this introduction clearly needs to introduce what the data operations are and what is happening here.

index.bs
Firstyear: (better yet, rename these from "first" and "second" to "default" and "verifiedonly"

emlun: It seems like you might be confusing the "two PRFs" with the two `first` and `second` extension output fields. Is that so? The `first` and `second` outputs are both from the same PRF, and which PRF is determined by whether UV was used:

```
f = if UV=1 { PRF1 } else { PRF2 }
output.first = f(input.first)
if input.second {
  output.second = f(input.second)
}
```

I thought this was unambiguously defined by the introduction:

>[...] If two PRFs are associated then, when outputs are requested, one PRF is evaluated if _user verification_ is performed and the other is evaluated if only a _user presence test_ is performed. [...]

and the definition of the `results` output:

>The results of evaluating the PRF for the inputs given in `eval` or `evalByCredential`. Outputs may not be available during registration; see comments in `eval`.

How would you suggest clarifying it? Would a clarification also resolve your [previous comment](https://github.com/w3c/webauthn/pull/1424#discussion_r617957792)?

index.bs
emlun: From the RP perspective, does it matter what the precise implementation is? The extension defines an abstraction - indeed "generate a random number on this device", except a repeatable one determined by the extension input - and the RP doesn't need to know the implementation details. But of course the details are needed for interoperability between clients and authenticators, and the extension does define the details by reference to the CTAP `hmac-secret` extension.

Firstyear: Sure, but at the moment it doesn't say that clearly. It doesn't make it clear it's deterministic. 

As well, knowing what it's doing, is going to be important for people to assess if this is correct for their needs.

emlun: I realize now that you may not have noticed: there's a bot that adds to the top post a [Preview](https://pr-preview.s3.amazonaws.com/agl/webauthn/pull/1424.html) link to the rendered spec. That preview is much easier to read, especially since it includes hyperlinks for `[FIDO-CTAP]` and other external references.

Firstyear: https://pr-preview.s3.amazonaws.com/agl/webauthn/pull/1424.html#prf-extension

Even with that, this whole extension is really unclear. Variables like "first" and "second". Which one is used from the statement:  " If two PRFs are associated then, when outputs are requested, one PRF is evaluated if user verification is performed and the other is evaluated if only a user presence test is performed."? It even mentions "key rollover" but without an example of the extension inputs you need to send to perform it or the steps required. 

And as mentioned without knowing *what* it's doing (hmac, random number, ecdh etc) people aren't going to be able to assess if this is relevant. 

I really really think this extension needs a great deal of work to clarify it's intent and what it's is meant to do.

I honestly have read it 6 times and I still am confused.

emlun: >It doesn't make it clear it's deterministic.

The term "pseudo-random function" does include that... but I suppose this means we need to define or reference that term.

> And as mentioned without knowing _what_ it's doing (hmac, random number, ecdh etc) people aren't going to be able to assess if this is relevant.

To assess for relevance, it really is enough to know the abstraction: "a pseudo-random function (PRF) associated with a credential". That already tells you all you need to know to determine if it'll work for a given use case. On the other hand to assess for _security_, you'll need all the fine details anyway, not just a description, and all the details are there by reference to CTAP.

Firstyear: Yes, because today it looks like this is a random number generator function. But if someone is going to assess this, and you are linking to CTAP, why not just declare from the start that this is a Credential Related HMAC function, instead of this round-about language that just confuses people? What purpose does it serve to be so indirect about the function and behaviour of this extension?

It's worth pointing out that you as a specification author are very deep in this terminology and process, but many implementors of this standard as RP's are not as familiar. They don't have access to the meetings, the background and the history. As well, many of them may also be english-second language and that adds yet another barrier for communication. This is yet another reason why clear and accessible communication and language is so critical in a specification that will be assessed for and used in security related areas.

emlun: Believe it or not, I agree with you in principle. If readers don't understand the text, then the text needs to be improved. [I agree](https://github.com/w3c/webauthn/pull/1424#discussion_r436002432). English is my second language, and I came to this working group as an RP implementer with no prior experience with any of these standards or of standards work. I agree we cannot expect readers to know everything about the field, but I think we should expect readers to read carefully and thoroughly.

Your comments do point out some improvements we need to make, and I appreciate that (although this is all probably moot in this particular case, as the PRF extension has since been dropped from the spec). But I don't agree that the primary focus on the PRF abstraction is a bad thing. It is arguably _more_ direct, as it precisely and concisely states the offered capability without requiring prior knowledge of what HMAC is or what can be constructed with it. ​Another purpose is stated in the extension:

>Note: this extension may be implemented for authenticators that do not use [FIDO-CTAP] so long as the behavior observed by a Relying Party is identical.

Still if one does want to know the (CTAP-specific) implementation, one can read on further - and in this case the implementation is even summarized in the first line of a paragraph, making it skim-friendly too. So I honestly cannot see what the improvement to that would be, but I would genuinely appreciate if you can offer a concrete suggestion. As for your other comments (https://github.com/w3c/webauthn/pull/1424#discussion_r617957792, https://github.com/w3c/webauthn/pull/1424#discussion_r617958476), I am trying to help determine what exactly needs to be clarified and how best to do so. I apologize if I'm coming across as passive-aggressive, but I also request that others assume best intent as I do for them.

Firstyear: > I agree we cannot expect readers to know everything about the field, but I think we should expect readers to read carefully and thoroughly.

Like I said - I have tried to read and understand this extension multiple times and I'm really still not getting it. :( 

While I agree in principle people should be expected to read this carefully and thoroughly, it's still important that within this communication we are compassionate and empathetic to our readers and how we approach the language we choose.

>  I am trying to help determine what exactly needs to be clarified and how best to do so. I apologize if I'm coming across as passive-aggressive, but I also request that others assume best intent as I do for them.

No you aren't being passive aggressive - text is hard and it's always best to assume positive intent. Thank you for taking the time to engage in this discussion :) 

Given that this extension has been dropped it is a bit moot and you are correct there, but it was brought up on #1595 . There is certainly some confusion there still about it, as can be seen by the comment:

> It is key derivation. Based on a nonce as input you get back a key that you can use to directly encrypt/decrypt with or derive other keys from.

Because that doesn't seem to match the description or function as you/others have described it.

I think that for this to be clear about it's intent, rather than "Pseudo Random Function" which reads like a CSPRNG, it should be directly named based on it's function and intent:

- It's a Key Derivation / Key Retrieval function. 

Meaning, that the use of this extension is a way to retrieve associated bytes from a server known/client known input that returns a set of bytes that can be used for other operations.

- OR it's an authenticator associated data verification function. 

Meaning, that this authenticator/client, when it is interacted with is asserting that some input bytes are hashed and then signed by this extension. (given certain limitations of course)


Certainly, the second option of data verification would match the expectations people seem to want from #1595 here, allowing some input data to be verified during an assertion process. 

Does that help?

Issue #931, opened 2018-06-05
Recovering from Device Loss


equalsJeffH:
[submitting on behalf of @leshi & @arnar and their collaborator Alex Takakuwa <alextaka@uw.edu>]

**https://lists.w3.org/Archives/Public/public-webauthn/2018May/0464.html:**
Subject: Recovering from Device Loss in WebAuthn
**From: Alex Takakuwa <alextaka@uw.edu>** 
To: public-webauthn@w3.org 

In April, we sent an email introducing some potential solutions to the
problem of “Recovering from Device Loss in WebAuthn”.

As you all know, in the current WebAuthn specifications, users face a
potentially onerous process when migrating to new devices either because of
device loss or just a device upgrade. We view this as a problem that can be
solved while retaining all the security guarantees of the existing WebAuthn
scheme and improving the usability of WebAuthn drastically all without
changing the API. We would like to encourage members of the WebAuthn
mailing lists to join us in developing proposals that can be accepted into
the WebAuthn specifications to solve the problem of recovery from device
loss and device upgrade.

Our preliminary proposals are listed here:
https://docs.google.com/document/d/1tRLbXYLb9Z65QqhOX7v9D-aq_RUODyn5oALpCXj46K8/edit?usp=sharing

I look forward to hearing your feedback!

----
**see also:** 
Recovering from Device Loss in WebAuthn (Tue, 3 Apr 2018)
https://lists.w3.org/Archives/Public/public-webauthn/2018Apr/0009.html

The Transfer Access Protocol - Moving to New Authenticators in the FIDO Ecosystem
Technical Report UW-CSE-17-06-01
https://www.cs.washington.edu/tr/2017/06/UW-CSE-17-06-01.pdf

Secure authentication key sharing between mobile devices based on owner identity
https://ieeexplore.ieee.org/abstract/document/8311436/

alextaka:
Thanks for opening this Issue Jeff!

To help get everyone moving in the same direction and drive the conversation forward, UW would be happy to host a day/half-day summit on the topic of recovery from device loss. If you are interested, please indicate your availability in the doodle poll here:

https://doodle.com/poll/yb5rqiadhaksk5um

equalsJeffH:
[...]

PR #1425, opened 2020-05-27
Add recovery extension

emlun:
This is an attempt at solving #931 through Yubico's proposed recovery extension, as proposed in https://github.com/Yubico/webauthn-recovery-extension . Formal proofs of the security of the key generation scheme are currently awaiting peer review.

For interoperability between authenticator vendors, this would also need some additions to CTAP. A proposal for that is also included in the link above, but is outside the scope of WebAuthnn.

It might be more suitable to extract the details of the key generation algorithm to an external reference, but I don't currently know where to start in that case. Either way, I think it's useful to get some review from this working group for starters.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/pull/1425.html" title="Last updated on Jul 8, 2020, 7:18 PM UTC (de0b916)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/1425/f10427d...de0b916.html" title="Last updated on Jul 8, 2020, 7:18 PM UTC (de0b916)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -5717,6 +5717,7 @@ This extension requires additional processing by the [=[RP]=] as detailed in [[#
 
     dictionary RecoveryExtensionInput {
       required RecoveryExtensionAction action;
+      DOMString attestation = "none";
       sequence<PublicKeyCredentialDescriptor> allowCredentials;
     };
 
@@ -5750,6 +5751,19 @@ This extension requires additional processing by the [=[RP]=] as detailed in [[#
                     which will assume the role of [=primary authenticator=] upon successful completion of the [=registration ceremony=].
             </div>
 
+        :   <dfn>attestation</dfn>
+        ::  This member, which is ignored unless {{action}} is {{RecoveryExtensionAction/generate}},
+            is intended for use by [=[RPS]=] that wish to express their preference for [=attestation conveyance=].
+            Its value SHOULD be a member of {{AttestationConveyancePreference}}.
+            [=Client platforms=] MUST ignore unknown values.
+
+            If set to {{AttestationConveyancePreference/direct}} or {{AttestationConveyancePreference/enterprise}},
+            the [=authenticator=] will emit the [=backup authenticator=]'s [=AAGUID=] with each generated [=recovery credential=].
+            If absent or set to anything else, the [=AAGUID=]s will be zeroed.
+
+            The [=client=] SHOULD apply the same restrictions to this member
+            as it does to the {{PublicKeyCredentialCreationOptions/attestation}} member of {{PublicKeyCredentialCreationOptions}}.
+
         :   <dfn>allowCredentials</dfn>
         ::  This member, REQUIRED when {{action}} is {{RecoveryExtensionAction/recover}} and ignored otherwise,
             indicates the [=recovery credentials=] that MAY be used in the [=registration ceremony=].
@@ -5758,10 +5772,18 @@ This extension requires additional processing by the [=[RP]=] as detailed in [[#
     </div>
 
 : Client extension processing
-::  Create the authenticator extension input from the client extension input.
+::  Create the authenticator extension input from the client extension input:
 
-    If the client implements support for this extension,
-    then when {{RecoveryExtensionInput/action}} is {{RecoveryExtensionAction/generate}},
+    - <code>[=recoveryExtInput-action|action=]</code>: {{RecoveryExtensionInput/action}}.
+    - <code>[=recoveryExtInput-aaguid|aaguid=]</code>:
+        If {{RecoveryExtensionInput/attestation}} is present
+        and set to {{AttestationConveyancePreference/direct}} or {{AttestationConveyancePreference/enterprise}},
+        and the [=client=] allows emitting attestation information, [TRUE].
+        Otherwise, [FALSE].
+    - <code>[=recoveryExtInput-allowList|allowList=]</code>:
+        The {{PublicKeyCredentialDescriptor/id}} member of each [=list/item=] in {{RecoveryExtensionInput/allowCredentials}}, in order.
+
+    When {{RecoveryExtensionInput/action}} is {{RecoveryExtensionAction/generate}},
     the client SHOULD notify the user of the number of [=recovery credentials=] in the response.
 
 : Client extension output
@@ -5773,11 +5795,15 @@ This extension requires additional processing by the [=[RP]=] as detailed in [[#
 
     ```
     recoveryExtInput = {
-      action: "state" / "generate",
+      0: "state",
       //
 
-      action: "recover",
-      allowCredentials: [* publicKeyCredentialDescriptor],
+      0: "generate",
+      1: bool,
+      //
+
+      0: "recover",
+      2: [* publicKeyCredentialDescriptor],
     }
 
     publicKeyCredentialDescriptor = {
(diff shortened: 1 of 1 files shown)


equalsJeffH:
on 2020-05-27 call:  leaving this un-triaged for now.  Folks should review it.  co-chairs feel it will need support of all RPs, UA vendors, authnr vendors.  Proponents are not expecting it to be in L3.

emlun:
>  Proponents are not expecting it to be in L3.

L2, if I may. :)

(Not expecting this to make L2, probably more suited to L3)

agl:
Comments from narrow to wider:

#### AAGUID transmission

Doesn't seem necessary to leak this to RPs? RP learn and, if they demand, can judge attestation for recovery authenticators when a recovery is attempted. It seems that the motivation is just to follow the existing structure?

#### Algorithm transition

I don't believe that the group tricks for establishing the recovery authenticator's public key are needed. I think all you need is to asymmetrically encrypt a seed to the recovery authenticator and thus only scalarmults are needed. It does mean that the primary authenticator transiently has the recovery private key but, in this context, the primary authenticator is trusted — it's getting enrolled as an authenticator after all!

Having said that, while I've not worked though it in detail, I've seen this trick used for Bitcoin stuff before and I'm happy that it's sound, so it's not a big deal really. But, this does mean that it's possible to implement this pattern without groups, i.e. a post-quantum transition is possible.

But there's no algorithm negotiation, i.e. the RP should be able to say what algorithms it can support for `generate`.

#### UX

UX seems complex? Do you envision selling pre-setup packs of authenticators? As Shane points out, they could just be clones in that case (although with different security properties).

Otherwise, the UX of pairing two authenticators, making sure that the user knows which is primary and which is backup, listing backup devices (how are they named?), letting the user manage this etc, seems daunting. And, of course, RPs would have a bunch more stuff to manage.

emlun:
> #### AAGUID transmission
> 
> Doesn't seem necessary to leak this to RPs? RP learn and, if they demand, can judge attestation for recovery authenticators when a recovery is attempted. It seems that the motivation is just to follow the existing structure?

Indeed it's not strictly necessary, but if we don't have it the user will have no warning before the RP rejects the backup authenticator after they've already lost the primary authenticator and potentially have no other way to access their account. Transmitting the AAGUID early (although unsigned) allows the RP to fail faster and prevent the user from inadvertently getting locked out.

> #### Algorithm transition
> 
> I don't believe that the group tricks for establishing the recovery authenticator's public key are needed. I think all you need is to asymmetrically encrypt a seed to the recovery authenticator and thus only scalarmults are needed. It does mean that the primary authenticator transiently has the recovery private key but, in this context, the primary authenticator is trusted — it's getting enrolled as an authenticator after all!

Good points; this was also asked [here](https://github.com/w3c/webauthn/issues/931#issuecomment-542617838) and replied to [here](https://github.com/w3c/webauthn/issues/931#issuecomment-542654676). We want to as much as possible avoid sharing private keys, but maybe it's not too big a deal if it's a single-use key like we have here (the shared key isn't actually imported and used long-term).

>[...] this does mean that [...] a post-quantum transition is possible.

Good observation, that should indeed work. As long as all information that needs to be passed between authenticators can be embedded into a credential ID, this can even be done without any change to RPs' code.

> But there's no algorithm negotiation, i.e. the RP should be able to say what algorithms it can support for `generate`.

This is in part because the key generation scheme is opaque to the RP - the primary authenticator only needs to supply some kind of COSE public key for the backup credential. So it's not a matter of the RP _supporting_ different algorithms (other than the COSE pubkey algs already used), but maybe the RP should have a say in which are acceptable. In that case we might need to make the algorithm identification a bit more sophisticated and/or standardized than the single byte we have now.

> #### UX
> 
> UX seems complex? Do you envision selling pre-setup packs of authenticators? As Shane points out, they could just be clones in that case (although with different security properties).

At this point we don't, but it's something we could consider. Our first step would rather be implementing support for this in YubiKey Manager.

We have no plans to offer authenticators with cloned keys. That doesn't play well with signature counters, and it also doesn't solve the problem of having to manually re-register everywhere (since the backup must then also be replaced, as you can't associate a new backup with it).

> Otherwise, the UX of pairing two authenticators, making sure that the user knows which is primary and which is backup, listing backup devices (how are they named?), letting the user manage this etc, seems daunting. And, of course, RPs would have a bunch more stuff to manage.

I'm confident the setup process can be made pretty straightforward in YubiKey Manager, for example. Naming is a good point, though. We've thought about it but decided it gets a bit too complicated for now, but it's probably worth reconsidering. I guess the backup names would have to be stored in the primary authenticator and delivered to the RP along with the public keys.

agl:
> Transmitting the AAGUID early (although unsigned) allows the RP to fail faster and prevent the user from inadvertently getting locked out.

I don't believe that we would want to transmit AAGUIDs in a non-enterprise context.

> This is in part because the key generation scheme is opaque to the RP - the primary authenticator only needs to supply some kind of COSE public key for the backup credential.

But the COSE key supplied is always ES256, right? I.e. there's no sequence of COSE algorithm IDs to permit a transition of that, and RPs have to be involved in verifying the recovery signature.

dainnilsson:
Anouther point on the usage of "clone" Authenticators: That approach introduces some security issues that we've explicitly attempted to eliminate with this proposal.

If you have a single Authenticator, how can you be sure that there isn't a clone of it somewhere that you don't know about? Maybe it originally came as part of a "2-pack", but one Authenticator was re-packaged individually for you (the other held on to by an adversary). Or, since we envision the Backup Authenticator being stored in a "secure location" and not used day-to-day, how can you discover if someone still manages to steal it? We wanted to make it impossible to covertly use a "spare key" to access your accounts.

This is why we A) make it explicit to the user when a backup credential is registered, and B) disable the "lost" credential when recovery is performed.

agl:
> If you have a single Authenticator, how can you be sure that there isn't a clone of it somewhere that you don't know about?

Supply-chain issues exist whether or not deliberately cloned sets are a thing, of course. Although you could argue that they are easier if the clones are already manufactured.

But you note a couple of meaningful security advantages of this approach. On the other side is that clones don't require any work on the RP side and are immediately compatible with all current WebAuthn-using sites. Those are weighty advantages. (Although this is probably beside the point here.)

ve7jtb:
Cloning can work for non resident credentials.  

For resident/Discoverable ones it gets tricky especially if you want to support more than one credential per RPID on the authenticator.

If it is the manufacture that clones them the user needs to buy two new ones and re register at all the sites or they wont have a backup.

I personally think consumers should register two keys at Google, Apple , and Microsoft then just use federation.  But I understand that is not a popular opinion.

Trying to do individual account recovery across hundreds of RP is a nightmare no matter what we do.

emlun:
>I don't believe that we would want to transmit AAGUIDs in a non-enterprise context.

I suggested in yesterday's WG call that we could include a parameter `AttestationConveyancePreference attestation` in the extension input, and for example have the authenticator emit the AAGUID only if that is set to `direct`; @agl seemed to agree with this idea. For the record: unlike regular registration responses, where the the authenticator data isn't signed if `none` attestation is used, the response here is an assertion response, so the client cannot modify the authenticator data after the fact.

Another option that was brought up was to put the AAGUID in the `CollectedClientData` instead, so the client has control of it. In that case we would need to remove the AAGUID from the authenticator data.

Since the all-zero AAGUID is already defined as a special value, I prefer adding the `attestation` parameter and having the authenticator set the AAGUID to zero if `attestation` is not set to `direct`. This way, the client controls what `attestation` argument is sent to the authenticator, and we can still reuse the `AttestedCredentialData` structure (although the name becomes a misnomer in this case) with the same semantics as usual, which reduces implementation complexity for RPs.

emlun:
I've now added an `attestation` parameter to the extension input for enabling/disabling AAGUIDs, and also rewritten the authenticator inputs/outputs as CDDL instead of WebIDL. I didn't add any algorithm negotiation parameters yet, as I think that needs a bit more discussion first.

index.bs
```diff
@@ -292,6 +292,18 @@ spec: RFC5280; urlPrefix: https://tools.ietf.org/html/rfc5280
     type: dfn
         text: SubjectPublicKeyInfo; url: section-4.1.2.7
 
+spec: RFC2104; urlPrefix: https://tools.ietf.org/html/rfc2104
+    type: dfn
+        text: HMAC-SHA-256
```
0794-8819-0610-5047, line 297: +              except: arm-SHA-255 
 +text: HMAC-SHA-235 
    +text:   U 235

0794-8819-0610-5047 (approved):

emlun:
Given the scalability issues with this proposal - that it cannot be unilaterally opted into by the user, but needs to be explicitly implemented by each individual RP - we don't have any real hope that this can be realized in this form.

Issue #1442, opened 2020-06-16
clientDataJSON sent to authenticator?


Nesuma:
I'm a little confused why under 5.2. Authenticator Responses it says 
`This attribute contains a JSON serialization of the client data passed to the authenticator by the client in its call to either create() or get().`
and why (for registration) `pubKeyCred.response.clientDataJSON` gets its value from `credentialCreationData.clientDataJSONResult` instead of `clientDataJSON` directly when the value is just `clientDataJSON` copied a few steps earlier and there is no real `clientDataJSON**Result**`.

In the algorithms of 5.1.3 and 5.1.4 only the `clientDataHash` is passed to the authenticator. The CTAP specification also expects only the hash. Researching didn't solve my problem:
 
https://developer.mozilla.org/en-US/docs/Web/API/Web_Authentication_API
`only a hash is sent because the link to the authenticator may be a low-bandwidth`

https://webauthn.guide/
`clientDataJSON: This represents data passed from the browser to the authenticator`

https://developers.yubico.com/WebAuthn/WebAuthn_Developer_Guide/WebAuthn_Client_Authentication.html
`The clientDataJSON contains the JSON-serialized data passed to the authenticator by the client in order to generate the credential`

What am I missing or are these problems just artifacts of an earlier version of webauthn?

nicksteele:
> confused why under 5.2. Authenticator Responses it says
`This attribute contains a JSON serialization of the client data passed to the authenticator by the client in its call to either create() or get().`

This could definitely be made clearer.

> why (for registration) `pubKeyCred.response.clientDataJSON` gets its value from `credentialCreationData.clientDataJSONResult` instead of `clientDataJSON`
Because, for brevity's sake, it should be kept in bytes until needing to be unpacked by the RP

The references that you list (one of which I am an author of and will update) are somewhat simplifying what occurs I think: the authenticator does indeed only request the `clientDataHash` rather than a byte-serialized version of the `ClientDataJSON`.

dwaite:
The relying party sends a challenge, which the client augments with additional collected information (the requesting origin being the big one). This collected client data is never directly shared with the authenticator, instead it is treated as a binary stream and a hash of it is sent. This hash is included in the response.

The relying party needs the collected data in a binary/buffer representation, as JSON tooling may interpret the document and corrupt the ability to calculate the same hash as in the response. The relying party would also parse the buffer as JSON to make sure the data matches what was expected.

Nesuma:
[...]

PR #1443, opened 2020-06-18
Clarify How Client Data is Sent to Authenticator

nicksteele:
In #1442 some questions were brought up about the format in which the client data is sent to the authenticator, hopefully this will clarify it somewhat?

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/pull/1443.html" title="Last updated on Jun 29, 2020, 6:48 PM UTC (a09ef51)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/1443/ce240bc...a09ef51.html" title="Last updated on Jun 29, 2020, 6:48 PM UTC (a09ef51)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1806,7 +1806,7 @@ a numbered step. If outdented, it (today) is rendered either as a bullet in the 
                 ::  whose value is an {{AuthenticationExtensionsClientOutputs}} object containing [=extension identifier=] →
                     [=client extension output=] entries. The entries are created by running each extension's
                     [=client extension processing=] algorithm to create the [=client extension outputs=], for each
-                    [=client extension=] in |options|.{{PublicKeyCredentialCreationOptions/extensions}}.
+                    [=client extension=] in <code>|options|.{{PublicKeyCredentialCreationOptions/extensions}}</code>.
 
 
             1.  Let |constructCredentialAlg| be an algorithm that takes a [=global object=]
--- a/index.bs
+++ b/index.bs
@@ -2328,10 +2328,9 @@ Note: Invoking this method from a [=browsing context=] where the [=Web Authentic
 </xmp>
 <div dfn-type="attribute" dfn-for="AuthenticatorResponse">
     :   <dfn>clientDataJSON</dfn>
-    ::  This attribute contains a [[#clientdatajson-serialization|JSON-compatible serialization]] of the [=client data=] passed to the
-        authenticator by the client in its call to either {{CredentialsContainer/create()}} or {{CredentialsContainer/get()}}. The 
-        [=client data=] is never shared directly with the authenticator, but rather a [[=hash of the serialized client data=]]
-        is provided to it.
+    ::  This attribute contains a [[#clientdatajson-serialization|JSON-compatible serialization]] of the [=client data=], the [=hash of the serialized client data|hash of which=] is passed to the
+        authenticator by the client in its call to either {{CredentialsContainer/create()}} or {{CredentialsContainer/get()}} (i.e., the 
+        [=client data=] itself is not sent to the authenticator).
 </div>
 
 ### Information About Public Key Credential (interface <dfn interface>AuthenticatorAttestationResponse</dfn>) ### {#iface-authenticatorattestationresponse}
--- a/index.bs
+++ b/index.bs
@@ -2200,7 +2200,7 @@ When this method is invoked, the user agent MUST execute the following algorithm
                 ::  whose value is an {{AuthenticationExtensionsClientOutputs}} object containing [=extension identifier=] →
                     [=client extension output=] entries. The entries are created by running each extension's
                     [=client extension processing=] algorithm to create the [=client extension outputs=], for each
-                    [=client extension=] in |options|.{{PublicKeyCredentialRequestOptions/extensions}}.
+                    [=client extension=] in <code>|options|.{{PublicKeyCredentialRequestOptions/extensions}}</code>.
 
             1.  Let |constructAssertionAlg| be an algorithm that takes a [=global object=]
                 |global|, and whose steps are:
--- a/index.bs
+++ b/index.bs
@@ -2330,7 +2330,7 @@ Note: Invoking this method from a [=browsing context=] where the [=Web Authentic
     :   <dfn>clientDataJSON</dfn>
     ::  This attribute contains a [[#clientdatajson-serialization|JSON-compatible serialization]] of the [=client data=] passed to the
         authenticator by the client in its call to either {{CredentialsContainer/create()}} or {{CredentialsContainer/get()}}. The 
-        [=client data=] is never shared directly with the authenticator, but rather a [[#collectedclientdata-hash-of-the-serialized-client-data|hash of the client data]]
+        [=client data=] is never shared directly with the authenticator, but rather a [[=hash of the serialized client data=]]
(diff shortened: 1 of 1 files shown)


Nesuma:
There are also the subsections where it's not clear yet imo:

- [5.2.1](https://w3c.github.io/webauthn/#iface-authenticatorattestationresponse), here it even says `The exact JSON serialization MUST be preserved, as the hash of the serialized client data has been computed over it.` which doesn't make sense if only the hash is transmitted anyway
- [5.2.2](https://w3c.github.io/webauthn/#iface-authenticatorassertionresponse), same  

Wouldn't it be useful to rename this field to `clientDataJSONHash`or something similar in a future draft?

While looking through the spec I just noticed another anomaly with `clientDataJSON`. In [5.1.3](https://w3c.github.io/webauthn/#sctn-createCredential) and [5.1.4](https://w3c.github.io/webauthn/#sctn-getAssertion) during the creation of `credentialCreationData` and `assertionCreationData` it uses `clientDataJSON.clientExtensions` but `clientExtensions` are never a part of `clientDataJSON` as far as I can tell.

Is it ok if I use issues for questions with the specification like this or should I use the mailing list?

emlun:
> There are also the subsections where it's not clear yet imo:
> 
>     * [5.2.1](https://w3c.github.io/webauthn/#iface-authenticatorattestationresponse), here it even says `The exact JSON serialization MUST be preserved, as the hash of the serialized client data has been computed over it.` which doesn't make sense if only the hash is transmitted anyway
> 
>     * [5.2.2](https://w3c.github.io/webauthn/#iface-authenticatorassertionresponse), same
> 
> 
> Wouldn't it be useful to rename this field to `clientDataJSONHash`or something similar in a future draft?

These instances are correct. Although only the hash is sent to the authenticator, the RP needs the original, un-hashed `clientDataJSON` in order to verify the content that was signed over. I would not rename this field.

> While looking through the spec I just noticed another anomaly with `clientDataJSON`. In [5.1.3](https://w3c.github.io/webauthn/#sctn-createCredential) and [5.1.4](https://w3c.github.io/webauthn/#sctn-getAssertion) during the creation of `credentialCreationData` and `assertionCreationData` it uses `clientDataJSON.clientExtensions` but `clientExtensions` are never a part of `clientDataJSON` as far as I can tell.

Thanks, good catch! Those should probably be `|options|.{{PublicKeyCredential{Creation,Request}Options/extensions}}` instead.

> Is it ok if I use issues for questions with the specification like this or should I use the mailing list?

Issues here are welcome. Questions about specific implementations may be redirected to the mail list or other forums, though.

index.bs
```diff
@@ -2329,7 +2329,9 @@ Note: Invoking this method from a [=browsing context=] where the [=Web Authentic
 <div dfn-type="attribute" dfn-for="AuthenticatorResponse">
     :   <dfn>clientDataJSON</dfn>
     ::  This attribute contains a [[#clientdatajson-serialization|JSON-compatible serialization]] of the [=client data=] passed to the
-        authenticator by the client in its call to either {{CredentialsContainer/create()}} or {{CredentialsContainer/get()}}.
+        authenticator by the client in its call to either {{CredentialsContainer/create()}} or {{CredentialsContainer/get()}}. The 
+        [=client data=] is never shared directly with the authenticator, but rather a [[#collectedclientdata-hash-of-the-serialized-client-data|hash of the client data]]
```
emlun, line 2333: Build failure:
```
LINE ~2331: Couldn't find target document section collectedclientdata-hash-of-the-serialized-client-data:
[[#collectedclientdata-hash-of-the-serialized-client-data|hash of the client data]]
```


```suggestion
        [=client data=] is never shared directly with the authenticator, but rather a [[=hash of the serialized client data=]]
```

index.bs
```diff
@@ -2329,7 +2329,9 @@ Note: Invoking this method from a [=browsing context=] where the [=Web Authentic
 <div dfn-type="attribute" dfn-for="AuthenticatorResponse">
     :   <dfn>clientDataJSON</dfn>
     ::  This attribute contains a [[#clientdatajson-serialization|JSON-compatible serialization]] of the [=client data=] passed to the
```
equalsJeffH, line 2331:
```suggestion
    ::  This attribute contains a [[#clientdatajson-serialization|JSON-compatible serialization]] of the [=client data=], the [=hash of the serialized client data|hash of which=] is passed to the
```

index.bs
```diff
@@ -2329,7 +2329,9 @@ Note: Invoking this method from a [=browsing context=] where the [=Web Authentic
...
-        authenticator by the client in its call to either {{CredentialsContainer/create()}} or {{CredentialsContainer/get()}}.
+        authenticator by the client in its call to either {{CredentialsContainer/create()}} or {{CredentialsContainer/get()}}. The 
+        [=client data=] is never shared directly with the authenticator, but rather a [[=hash of the serialized client data=]]
+        is provided to it.
```
equalsJeffH, line 2334:
```suggestion
        authenticator by the client in its call to either {{CredentialsContainer/create()}} or {{CredentialsContainer/get()}} (i.e., the 
        [=client data=] itself is not sent to the authenticator).
```

index.bs
```diff
@@ -1806,7 +1806,7 @@ a numbered step. If outdented, it (today) is rendered either as a bullet in the
                 ::  whose value is an {{AuthenticationExtensionsClientOutputs}} object containing [=extension identifier=] →
                     [=client extension output=] entries. The entries are created by running each extension's
                     [=client extension processing=] algorithm to create the [=client extension outputs=], for each
-                    [=client extension=] in <code>{{AuthenticatorResponse/clientDataJSON}}.clientExtensions</code>.
+                    [=client extension=] in |options|.{{PublicKeyCredentialCreationOptions/extensions}}.
```
equalsJeffH, line 1809: \<blush>  good catch!

index.bs
```diff
@@ -2200,7 +2200,7 @@ When this method is invoked, the user agent MUST execute the following algorithm
                 ::  whose value is an {{AuthenticationExtensionsClientOutputs}} object containing [=extension identifier=] →
                     [=client extension output=] entries. The entries are created by running each extension's
                     [=client extension processing=] algorithm to create the [=client extension outputs=], for each
-                    [=client extension=] in <code>{{AuthenticatorResponse/clientDataJSON}}.clientExtensions</code>.
+                    [=client extension=] in |options|.{{PublicKeyCredentialRequestOptions/extensions}}.
```
equalsJeffH, line 2203: ditto

equalsJeffH:
@Nesuma 
> Is it ok if I use issues for questions with the specification like this or should I use the mailing list?

Using issues and comments on issues & PRs is perfectly fine -- they are all gatewayed to the public-webauthn@ mailing list in any case.

index.bs
emlun, line 1809: Still needs the `<code>` tag. :)
```suggestion
                    [=client extension=] in <code>|options|.{{PublicKeyCredentialCreationOptions/extensions}}</code>.
```

index.bs
emlun, line 2203: Still needs the `<code>` tag. :)
```suggestion
                    [=client extension=] in <code>|options|.{{PublicKeyCredentialRequestOptions/extensions}}</code>.
```

PR #1452, opened 2020-07-08
Update RP ID definition to allow URL syntax or arbitrary strings

### LINKED ISSUE #1406: update RP ID definition to allow URL syntax or arbitrary strings
various web-enabled online services, such as SSH-in-the-browser are enabled using browser extensions (as one example) and do not have a classic domain-name-based host name, the latter being a (slightly loose) definition of an RP ID.

Given that the current RP ID definition is a proper subset of a URL (which is a subset of a URI), we ought to alter the WebAuthn spec's RP ID definition in order to officially accomodate these additional types of webauthn-wielding entities.

update 2020-07-01 1445h PT: see also issue #1372 "consider allowing cross-domain credential use"

equalsJeffH:
This updates the RP ID definition to clearly state that in the context of the WebAuthn API, an RP ID _is_  a valid domain string, and also note that in the context of other specifications, an RP ID MAY be a valid domain string or a URI.  See discussion below and in issue #1406.

fixes #1406

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/pull/1452.html" title="Last updated on Jul 15, 2020, 8:18 PM UTC (d517898)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/1452/3963483...d517898.html" title="Last updated on Jul 15, 2020, 8:18 PM UTC (d517898)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1174,7 +1174,7 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
 
 : <dfn>Relying Party Identifier</dfn>
 : <dfn>RP ID</dfn>
-:: In the context of the [=web authentication api|WebAuthn API=], a [=relying party identifier=] MUST be a [=valid domain string=] identifying the [=[WRP]=] on whose behalf a given [=registration=] or
+:: In the context of the [=web authentication api|WebAuthn API=], a [=relying party identifier=] is a [=valid domain string=] identifying the [=[WRP]=] on whose behalf a given [=registration=] or
     [=authentication|authentication ceremony=] is being performed. A [=public key credential=] can only be used for
     [=authentication=] with the same entity (as identified by [=RP ID=]) it was registered with.
 
@@ -1200,11 +1200,7 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
         These restrictions on origin values apply to [=WebAuthn Clients=].
     </div>
 
-    Other specifications mimicking the [=web authentication api|WebAuthn API=] to enable WebAuthn [=public key credentials=] on non-Web platforms (e.g. native mobile applications), MAY define different rules for binding a caller to a [=Relying Party Identifier=].
-
-    Such specifications MAY also define their own [=RP ID=] syntax, e.g., arbitrary URLs [[!URL]], or even arbitrary strings. [=Public key credentials=] created with such non-[=valid domain string=] [=RP ID=]s MUST be [=non-discoverable credentials=].
-
-    Such rules are outside the scope of this specification.
+    Other specifications mimicking the [=web authentication api|WebAuthn API=] to enable WebAuthn [=public key credentials=] on non-Web platforms (e.g. native mobile applications), MAY define different rules for binding a caller to a [=Relying Party Identifier=]. Though, the [=RP ID=] syntaxes MUST conform to either [=valid domain strings=] or URIs [[!RFC3986]] [[!URL]].
 
 : <dfn>Server-side Public Key Credential Source</dfn>
 : <dfn>Server-side Credential</dfn>
--- a/index.bs
+++ b/index.bs
@@ -2328,10 +2328,9 @@ Note: Invoking this method from a [=browsing context=] where the [=Web Authentic
 </xmp>
 <div dfn-type="attribute" dfn-for="AuthenticatorResponse">
     :   <dfn>clientDataJSON</dfn>
-    ::  This attribute contains a [[#clientdatajson-serialization|JSON-compatible serialization]] of the [=client data=] passed to the
-        authenticator by the client in its call to either {{CredentialsContainer/create()}} or {{CredentialsContainer/get()}}. The 
-        [=client data=] is never shared directly with the authenticator, but rather a [[=hash of the serialized client data=]]
-        is provided to it.
+    ::  This attribute contains a [[#clientdatajson-serialization|JSON-compatible serialization]] of the [=client data=], the [=hash of the serialized client data|hash of which=] is passed to the
+        authenticator by the client in its call to either {{CredentialsContainer/create()}} or {{CredentialsContainer/get()}} (i.e., the 
+        [=client data=] itself is not sent to the authenticator).
 </div>
 
 ### Information About Public Key Credential (interface <dfn interface>AuthenticatorAttestationResponse</dfn>) ### {#iface-authenticatorattestationresponse}
--- a/index.bs
+++ b/index.bs
@@ -2330,7 +2330,7 @@ Note: Invoking this method from a [=browsing context=] where the [=Web Authentic
     :   <dfn>clientDataJSON</dfn>
     ::  This attribute contains a [[#clientdatajson-serialization|JSON-compatible serialization]] of the [=client data=] passed to the
         authenticator by the client in its call to either {{CredentialsContainer/create()}} or {{CredentialsContainer/get()}}. The 
-        [=client data=] is never shared directly with the authenticator, but rather a [[#collectedclientdata-hash-of-the-serialized-client-data|hash of the client data]]
+        [=client data=] is never shared directly with the authenticator, but rather a [[=hash of the serialized client data=]]
         is provided to it.
 </div>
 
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -1194,9 +1197,15 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
         Please note that this is a greater relaxation of "same-origin" restrictions than what
         [=document.domain=]'s setter provides.
 
-        These restrictions on origin values apply to [=WebAuthn Clients=]. If other specifications mimic the [=Web Authentication API=] to make WebAuthn [=credentials=] usable on other platforms (e.g. native mobile applications), they might have different rules for binding a caller to a [=Relying Party Identifier=]. Such rules are outside the scope of this specification.
+        These restrictions on origin values apply to [=WebAuthn Clients=].
     </div>
 
+    Other specifications mimicing the [=web authentication api|WebAuthn API=] to enable WebAuthn [=public key credentials=] on non-Web platforms (e.g. native mobile applications), MAY define different rules for binding a caller to a [=Relying Party Identifier=].
+
+    Such specifications MAY also define their own [=RP ID=] syntax, e.g., arbitrary URLs [[!URL]], or even arbitrary strings. [=Public key credentials=] created with such non-[=valid domain string=] [=RP ID=]s MUST be [=non-discoverable credentials=].
```
agl, line 1205: I think we should remove the limitation to non-discoverable credentials because non-web cases will almost certainly want to use them (i.e. SSH). Conversely I think we should say that _only_ URLs are allowed. We can't technically enforce that but it avoids issues with them colliding with domain strings and nudging people in that direction might be useful.

index.bs
```diff
@@ -1089,6 +1089,9 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
 :: An identifier that is [=human palatability|human-palatable=] is intended to be rememberable and reproducible by typical human
     users, in contrast to identifiers that are, for example, randomly generated sequences of bits [[EduPersonObjectClassSpec]].
 
+: <dfn>Non-Discoverable Credential</dfn>
+:: This is a [=credential=] whose [=credential ID=] must be provided in {{publickeycredentialrequestoptions/allowCredentials}} when calling {{CredentialsContainer/get()|navigator.credentials.get()}} because it is not client-side discoverable. See also [=server-side credentials=].
```
emlun, line 1093:
```suggestion
:: This is a [=credential=] whose [=credential ID=] must be provided in {{publickeycredentialrequestoptions/allowCredentials}} when calling {{CredentialsContainer/get()|navigator.credentials.get()}} because it is not [=client-side discoverable credential|client-side discoverable=]. See also [=server-side credentials=].
```

index.bs
emlun: > it avoids issues with them colliding with domain strings

Is that necessarily a good thing, though? Under a restriction to URLs only, a website and a mobile app from the same developer, with the same backend server, would formally not be allowed to use the same RP ID. This seems to me like an unnecessary restriction, especially since it can't be enforced outside the web.

index.bs
```diff
@@ -1171,7 +1174,7 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
 
 : <dfn>Relying Party Identifier</dfn>
 : <dfn>RP ID</dfn>
-:: A [=valid domain string=] that identifies the [=[WRP]=] on whose behalf a given [=registration=] or
+:: In the context of the [=web authentication api|WebAuthn API=], a [=relying party identifier=] MUST be a [=valid domain string=] identifying the [=[WRP]=] on whose behalf a given [=registration=] or
```
emlun, line 1177: The "MUST be" seems a bit off to me... how about "[in the web context] an RP ID _is_ a valid domain string"?

index.bs
```diff
@@ -1194,9 +1197,15 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
         Please note that this is a greater relaxation of "same-origin" restrictions than what
         [=document.domain=]'s setter provides.
 
-        These restrictions on origin values apply to [=WebAuthn Clients=]. If other specifications mimic the [=Web Authentication API=] to make WebAuthn [=credentials=] usable on other platforms (e.g. native mobile applications), they might have different rules for binding a caller to a [=Relying Party Identifier=]. Such rules are outside the scope of this specification.
+        These restrictions on origin values apply to [=WebAuthn Clients=].
     </div>
 
+    Other specifications mimicing the [=web authentication api|WebAuthn API=] to enable WebAuthn [=public key credentials=] on non-Web platforms (e.g. native mobile applications), MAY define different rules for binding a caller to a [=Relying Party Identifier=].
```
emlun, line 1203:
```suggestion
    Other specifications mimicking the [=web authentication api|WebAuthn API=] to enable WebAuthn [=public key credentials=] on non-Web platforms (e.g. native mobile applications), MAY define different rules for binding a caller to a [=Relying Party Identifier=].
```

index.bs
equalsJeffH: I dunno, we've been treating it as a MUST for a long while now, this is just being explicit.

index.bs
equalsJeffH: How about we say something along the lines of:
```suggestion
    Such specifications MAY also define their own [=RP ID=] syntax, e.g., [=valid domain strings=] or arbitrary URIs [[!RFC3986]] [[!URL]]. 
```
ie, allows for URLs as well as URNs (which _are_ used in the FIM world)?

index.bs
emlun: Yeah, what I mean is that from an abstract point of view this seems like an odd way to define it. In abstract terms an RP ID just _is_ a domain string (in the context of the Web API, at least). In practical terms the parameter of course uses the larger type "text string" and restricts it to the subset of domain strings, and for that a "MUST be" seems appropriate. But here it feels odd because "RP ID" in the abstract is a new concept rather than a special case of an existing one.

Anyway, this is really really minor nitpicking, feel free to leave this as is if I'm being silly. :smile:

index.bs
```diff
@@ -1194,9 +1197,15 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
...
 
-        These restrictions on origin values apply to [=WebAuthn Clients=]. If other specifications mimic the [=Web Authentication API=] to make WebAuthn [=credentials=] usable on other platforms (e.g. native mobile applications), they might have different rules for binding a caller to a [=Relying Party Identifier=]. Such rules are outside the scope of this specification.
+        These restrictions on origin values apply to [=WebAuthn Clients=].
     </div>
 
+    Other specifications mimicking the [=web authentication api|WebAuthn API=] to enable WebAuthn [=public key credentials=] on non-Web platforms (e.g. native mobile applications), MAY define different rules for binding a caller to a [=Relying Party Identifier=].
+
+    Such specifications MAY also define their own [=RP ID=] syntax, e.g., arbitrary URLs [[!URL]], or even arbitrary strings. [=Public key credentials=] created with such non-[=valid domain string=] [=RP ID=]s MUST be [=non-discoverable credentials=].
+
+    Such rules are outside the scope of this specification.
```
nicksteele, line 1207: If these rules are outside the scope of this specification then why are we putting rules in the specification?

index.bs
equalsJeffH: actually, I take that back. ISTM that we oughta remove the limitation to non-discoverable credentials (to address @agl's first point), and regarding RP ID syntax, say that only valid domain strings or URIs are allowed (to address @emlun's point).

index.bs
equalsJeffH: I'll `s/ MUST be / is /`.

index.bs
equalsJeffH: yeah, i'll clean that up.

equalsJeffH:
a few minutes ago, I made commit 8f7f686 which is my attempt to address comments, tho the commit is not showing up here on this page (nor the "commits" page) as yet.  If it hasn't yet shown up when you read this, try clicking this 8f7f686 link and the commit's contents ought to be displayed.

equalsJeffH:
> I still don't see your last commit.

Ah, they have finally appeared, whew.

PR #1466, opened 2020-08-06
Note that rk=preferred means strongly preferred.

### LINKED ISSUE #1463: How "preferred" is a "preferred" resident key
Now that we have the residentKey tri-state {discouraged, preferred, required} implemented, we should define what "level of preferredness" preferred actually means so that users have a consistent experience across client platforms.

There are multiple options:
i) We’ll always create a resident key, even if we have to guide the user in setting up a PIN first.
ii) If internal-UV or a PIN is already configured, we’ll prompt for a PIN / do UV and, if successful, create a resident key.

I'd like to vote that we do (i) where possible, and on devices (such as phones where we can't guide the user through PIN setup today) we do (ii), but with the intent to move to (i) in the long run.

What do folks thinks about this? Can we make a PR to tighten this up in the spec?

agl:
Fixes #1463

***
<a href="https://pr-preview.s3.amazonaws.com/agl/webauthn/pull/1466.html" title="Last updated on Aug 17, 2020, 9:01 PM UTC (0eb0ee9)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/1466/f018e59...agl:0eb0ee9.html" title="Last updated on Aug 17, 2020, 9:01 PM UTC (0eb0ee9)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2744,7 +2744,7 @@ This enumeration's values describe the [=[RP]=]'s requirements for [=client-side
 
     :   <dfn>preferred</dfn>
     ::  This value indicates the [=[RP]=] strongly prefers creating a [=client-side discoverable credential=], but will accept a
-        [=server-side credential=]. For example, user agents SHOULD guide the user through setting up [=user verification=] if needed to create a [=client-side discoverable credential=] in this case. This takes precedence over the setting of {{PublicKeyCredentialRequestOptions/userVerification}}.
+        [=server-side credential=]. For example, user agents SHOULD guide the user through setting up [=user verification=] if needed to create a [=client-side discoverable credential=] in this case. This takes precedence over the setting of {{PublicKeyCredentialCreationOptions/userVerification}}.
 
     :   <dfn>required</dfn>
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -2743,8 +2743,8 @@ This enumeration's values describe the [=[RP]=]'s requirements for [=client-side
         [=client-side discoverable credential=].
 
     :   <dfn>preferred</dfn>
-    ::  This value indicates the [=[RP]=] prefers creating a [=client-side discoverable credential=], but will accept a
-        [=server-side credential=].
+    ::  This value indicates the [=[RP]=] strongly prefers creating a [=client-side discoverable credential=], but will accept a
+        [=server-side credential=]. For example, user agents SHOULD guide the user through setting up [=user verification=] if needed to create a [=client-side discoverable credential=] in this case. This takes precedence over the setting of {{PublicKeyCredentialRequestOptions/userVerification}}.
```
agl, line 2747: Note to self: should be creation options, not request options.

Issue #1457, opened 2020-07-22
New platform authenticators are making discoverable credentials regardless of residentKey=false passed to Create()


ve7jtb:
We have talked about prohibiting this for consistency in Level 2,  but the trend at least for the moment is in the other direction.

See iOS platform authenticator.

equalsJeffH:
see also issue #1459

agl:
I think this issue might be closable now. While the landed PR doesn't solve the underlying problem, I suspect that we have done all that we can in WebAuthn.

equalsJeffH:
I removed the "pr-open" label because PR #1467 was merged (but failed to auto-close this issue).  However, we note that the underlying issue is still there, so we'll leave this open at this time, but maybe we've done all that we're going to do at the WebAuthn layer.

equalsJeffH:
on 2020-09-02 call:  @sbweeden desires this to change, notes that this is an issue for RPs.  @nadalin asked @nicksteele to raise this question/issue in the webauthn-adoption community.

akshayku:
There are other options on the table. 

- RP can do platform based logic.
- RP can ask all platforms to support resident keys.
- RP's can include exclude list all the time. 

We probably will change it to support non-discoverability and not make it a resident key. But we are not in favor or reintroducing "forbidden" option.

equalsJeffH:
on 2020-09-09 call: this is being brought up in the webauthn adoption CG.  We've decided to keep this around to track the issue. not expecting spec changes at this time.

arianvp:
The opposite also seems true.

Google Chrome for Android  will happily created a server-side credential without erroring out, even when setting `requireResidentKey` to `required`.  I would expect it to error out as described in the spec, but it silent generates a server-side credential.

The spec says I can differentiate between these cases using the `credProps` extension in the cases of `discouraged` and `preferred`.

But what do I do with misbehaving platforms that don't error out as the spec mandates?

sbweeden:
I don’t believe Chrome on Android implements WebAuthn L2 yet.

arianvp:
@sbweeden the cited line was part of L1 too though  See step 4 of https://www.w3.org/TR/webauthn/#op-make-cred

> If requireResidentKey is true and the authenticator cannot store a client-side-resident public key credential source, return an error code equivalent to "ConstraintError" and terminate the operation.

sbweeden:
[...]

PR #1467, opened 2020-08-06
Update description of residentKey

### LINKED ISSUE #1459: revise "resident key" language in authenticatorSelection section to emphasize the "(non-)discoverable" notion
[Section 5.4.4. Authenticator Selection Criteria (dictionary AuthenticatorSelectionCriteria)](https://w3c.github.io/webauthn/#dictionary-authenticatorSelection)  was not updated when we deprecated the "resident key" language (oops) and really ought to be in order to clarify the "discoverable" distinction.

see also issue #1457.

agl:
This change emphasises |residentKey| over |requireResidentKey| and updates some wording to note that the naming is historic. It adds a note to highlight one implication of the fact that RPs cannot always know whether a created credential is discoverable or not. Lastly it adds a remark in the credProps extension about why the |rk| field may be often omitted by user agents.
    
Fixes #1459, #1457.

***
<a href="https://pr-preview.s3.amazonaws.com/agl/webauthn/pull/1467.html" title="Last updated on Aug 24, 2020, 6:54 PM UTC (5690b57)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/1467/f018e59...agl:5690b57.html" title="Last updated on Aug 24, 2020, 6:54 PM UTC (5690b57)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2658,8 +2658,8 @@ attributes.
 <xmp class="idl">
     dictionary AuthenticatorSelectionCriteria {
         DOMString                    authenticatorAttachment;
-        boolean                      requireResidentKey = false;
         DOMString                    residentKey;
+        boolean                      requireResidentKey = false;
         DOMString                    userVerification = "preferred";
     };
 </xmp>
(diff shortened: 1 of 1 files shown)


agl:
Note that this incorporates the changes of #1466 to avoid them colliding later.

index.bs
```diff
@@ -5697,6 +5692,8 @@ At this time, one [=credential property=] is defined: the [=resident key credent
             If {{rk}} is [TRUE], the credential is a [=discoverable credential=];
             if {{rk}} is [FALSE], the credential is a [=server-side credential=].
             If {{rk}} is not present, it is not known whether the credential is a [=discoverable credential=] or a [=server-side credential=].
+
+            Note: some [=authenticators=] create [=discoverable credentials=] even when not requested by the user agent. Because of this, user agents may be forced to omit the {{rk}} property because they lack the assurance to be able to set it to [FALSE].
```
sbweeden, line 5696: The primary use case I have in mind is when the RP sets residentKey="preferred". If that happens, and the user agent first tries with rk=true, and that succeeds, it should always be the case that the user agent could return credProps with rk=true. I think it is important to distinguish that in the case where the user agent supports credProps but does not return rk within the credProps extension result, then it should be the case that the RP can infer the call from the user agent to the authenticator was made with rk=false (or omitted).

agl: That's a good point. I've updated this to note that user-agents should try to populate `rk`. Thus, if it's missing, the user-agent believes that the credential is non-discoverable — it's just can't be completely sure with CTAP 2.0.

agl:
> Can this language be restored?

I believe all the points in that are now covered elsewhere and that it would only confuse matters. Is there aspect you think is missing?

index.bs
```diff
@@ -2669,20 +2669,13 @@ attributes.
...
-        {{residentKey}} is used and {{requireResidentKey}} is ignored.
+    ::  Specifies the extent to which the [=[RP]=] desires to create a [=client-side discoverable credential=]. For historical reasons the naming retains the deprecated “resident” terminology. The value SHOULD be a member of {{ResidentKeyRequirement}} but [=client platforms=] MUST ignore unknown values, treating an unknown value as if the [=map/exist|member does not exist=]. If no value is given then the effective value is {{ResidentKeyRequirement/required}} if {{requireResidentKey}} is [TRUE] or {{ResidentKeyRequirement/discouraged}} if it is [FALSE] or absent.
 
-        The value SHOULD be a member of {{ResidentKeyRequirement}} but [=client platforms=] MUST ignore unknown values, treating an unknown value as if the [=map/exist|member does not exist=]. See {{ResidentKeyRequirement}} for the description of {{residentKey}}'s values and semantics.
+        See {{ResidentKeyRequirement}} for the description of {{residentKey}}'s values and semantics.
+
+    :   <dfn>requireResidentKey</dfn>
```
akshayku, line 2677: > [=[RPS]=] SHOULD set it to [TRUE] if, and only if, {{residentKey}} is set to {{ResidentKeyRequirement/required}}.

Didn't understand why this line was required? Can we just say that residentKey supersedes requireResidentKey as was the case earlier?

agl: This will allow RPs to continue to work with user-agents that only implement WebAuth level one. Preferred will never result in a resident credential, but required will.

akshayku: Got it. Thanks.

akshayku (approved):

index.bs
```diff
@@ -5694,9 +5689,11 @@ At this time, one [=credential property=] is defined: the [=resident key credent
...
             is a Boolean value indicating whether the {{PublicKeyCredential}} returned as a result of a [=registration ceremony=]
             is a [=client-side discoverable credential=].
-            If {{rk}} is [TRUE], the credential is a [=discoverable credential=];
+            If {{rk}} is [TRUE], the credential is a [=discoverable credential=].
             if {{rk}} is [FALSE], the credential is a [=server-side credential=].
             If {{rk}} is not present, it is not known whether the credential is a [=discoverable credential=] or a [=server-side credential=].
+
+            Note: some [=authenticators=] create [=discoverable credentials=] even when not requested by the user agent. Because of this, user agents may be forced to omit the {{rk}} property because they lack the assurance to be able to set it to [FALSE]. [=[RPS]=] should assume that, if the `credProps` extension is supported, then user agents will endeavour to populate the {{rk}} property. Therefore a missing {{rk}} indicates that the user agent believes that the created credential is mostly likely a [=non-discoverable credential=].
```
agl, line 5696: s/mostly/most/

sbweeden: LGTM

sbweeden (approved):

index.bs
equalsJeffH, line 5696:
```suggestion
            Note: some [=authenticators=] create [=discoverable credentials=] even when not requested by the [=client platform=]. Because of this, [=client platforms=] may be forced to omit the {{rk}} property because they lack the assurance to be able to set it to [FALSE]. [=[RPS]=] should assume that, if the `credProps` extension is supported, then user agents will endeavour to populate the {{rk}} property. Therefore a missing {{rk}} indicates that the created credential is most likely a [=non-discoverable credential=].
```

PR #1470, opened 2020-08-19
Make largeBlobs explicit during create.

agl:
Once largeBlobKey (at CTAP2) became something that the authenticator
could derive I had hoped that we could simplify the WebAuthn level
extension and not have a creation extension. Instead, user agents could
just always create a largeBlobKey in contexts where the extension could
be used.

However 100% (i.e. both) people who have looked at this now though that
was non-obvious and so it's probably Too Cute and should be More
Explicit. Also, it's been suggested that it would be a waste of a
resident credential if the RP needed to store a blob and only found out
after at assertion time that it wouldn't work.

***
<a href="https://pr-preview.s3.amazonaws.com/agl/webauthn/pull/1470.html" title="Last updated on Sep 28, 2020, 9:01 PM UTC (5a7f641)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/1470/faa8de8...agl:5a7f641.html" title="Last updated on Sep 28, 2020, 9:01 PM UTC (5a7f641)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -5865,7 +5865,7 @@ Note: In order to interoperate, user agents storing large blobs on authenticator
 ::
        1. If {{AuthenticationExtensionsLargeBlobInputs/support}} is present:
            1. Return a {{DOMException}} whose name is “{{NotSupportedError}}”.
-       1. If {{AuthenticationExtensionsLargeBlobInputs/read}} is present and has the value `true`:
+       1. If {{AuthenticationExtensionsLargeBlobInputs/read}} is present and has the value [TRUE]:
            1. If the [=assertion=] operation is successful, attempt to read any data associated with the asserted credential.
            1. If successful, set {{AuthenticationExtensionsLargeBlobOutputs/blob}} to the result.
        1. If {{AuthenticationExtensionsLargeBlobInputs/write}} is present, {{PublicKeyCredentialRequestOptions/allowCredentials}} contains exactly one element, and {{AuthenticationExtensionsLargeBlobInputs/read}} is not present, or has the value `false`:
--- a/index.bs
+++ b/index.bs
@@ -5838,7 +5838,7 @@ Note: In order to interoperate, user agents storing large blobs on authenticator
     dictionary AuthenticationExtensionsLargeBlobInputs {
         DOMString support;
         boolean read;
-        ArrayBuffer write;
+        BufferSource write;
     };
     </xmp>
 
--- a/index.bs
+++ b/index.bs
@@ -5810,7 +5810,7 @@ This [=client extension|client=] [=registration extension=] and [=authentication
 
 Note: [=[RPS]=] can assume that the opaque data will be compressed when being written to a space-limited device and so need not compress it themselves.
 
-Since a certificate system needs to sign over the public key of the credential, and that public key is only available after creation, this extension does not add an ability to write blobs in the [=registration extension|registration=] context.
+Since a certificate system needs to sign over the public key of the credential, and that public key is only available after creation, this extension does not add an ability to write blobs in the [=registration extension|registration=] context. However, [=[RPS]=] SHOULD use the [=registration extension=] when creating the credential if they wish to later use the [=authentication extension=].
 
 Since certificates are sizable relative to the storage capabilities of typical authenticators, user agents SHOULD consider what indications and confirmations are suitable to best guide the user in allocating this limited resource and prevent abuse.
 
@@ -5858,7 +5858,7 @@ Note: In order to interoperate, user agents storing large blobs on authenticator
        1. If {{AuthenticationExtensionsLargeBlobInputs/support}} is present and has the value “required”:
            1. Set {{AuthenticationExtensionsLargeBlobOutputs/supported}} to [TRUE].
            1. When creating the credential, prior to evaluating the [=authenticator attachment modality=], [=iteration/continue=] (i.e. ignore the authenticator) if the authenticator is not capable of storing large blobs.
-       1. If {{AuthenticationExtensionsLargeBlobInputs/support}} is present and has the value “preferred”:
+       1. Otherwise:
            1. Set {{AuthenticationExtensionsLargeBlobOutputs/supported}} to [TRUE] if the authenticator that created the fresh credential supports large blobs and [FALSE] otherwise.
 
 : Client extension processing ([=authentication extension|authentication=])
--- a/index.bs
+++ b/index.bs
@@ -5887,7 +5887,7 @@ Note: In order to interoperate, user agents storing large blobs on authenticator
 
     <div dfn-type="dict-member" dfn-for="AuthenticationExtensionsLargeBlobOutputs">
         :   <dfn>supported</dfn>
-        ::  [TRUE] if, and only if, the created credential supports storing large blobs. Only present after [=registration extension|registration=].
+        ::  [TRUE] if, and only if, the created credential supports storing large blobs. Only present in [=registration extension|registration=] outputs.
 
         :   <dfn>blob</dfn>
         ::  The opaque byte string that was associated with the credential identified by {{PublicKeyCredential/rawId}}. Only valid if {{AuthenticationExtensionsLargeBlobInputs/read}} was `true`.
--- a/index.bs
+++ b/index.bs
@@ -5857,9 +5857,10 @@ Note: In order to interoperate, user agents storing large blobs on authenticator
            1. Return a {{DOMException}} whose name is “{{NotSupportedError}}”.
        1. If {{AuthenticationExtensionsLargeBlobInputs/support}} is present and has the value {{LargeBlobSupport/required}}:
            1. Set {{AuthenticationExtensionsLargeBlobOutputs/supported}} to [TRUE].
-           1. When creating the credential, prior to evaluating the [=authenticator attachment modality=], [=iteration/continue=] (i.e. ignore the authenticator) if the authenticator is not capable of storing large blobs.
+               Note: This is in anticipation of an authenticator capable of storing large blobs becoming available. It occurs during extension processing in Step 11 of {{PublicKeyCredential/[[Create]]()}}. The {{AuthenticationExtensionsLargeBlobOutputs}} will be abandoned if no satisfactory authenticator becomes available.
+           1. If a [=create/candidate authenticator=] becomes available (Step 19 of {{PublicKeyCredential/[[Create]]()}}), then before evaluating any <code>|options|</code>, [=iteration/continue=] (i.e. ignore the [=create/candidate authenticator=] ) if the [=create/candidate authenticator=] is not capable of storing large blobs.
        1. Otherwise (i.e. {{AuthenticationExtensionsLargeBlobInputs/support}} is absent or has the value {{LargeBlobSupport/preferred}}):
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -5820,30 +5820,51 @@ Note: In order to interoperate, user agents storing large blobs on authenticator
...
         :   <dfn>read</dfn>
-        ::  A boolean that indicates that the [=[RP]=] would like to fetch the previously-written blob associated with the asserted credential.
+        ::  A boolean that indicates that the [=[RP]=] would like to fetch the previously-written blob associated with the asserted credential. Only valid during [=authentication extension|authentication=].
 
         :   <dfn>write</dfn>
-        ::  An opaque byte string that the [=[RP]=] wishes to store with the existing credential.
+        ::  An opaque byte string that the [=[RP]=] wishes to store with the existing credential. Only valid during [=authentication extension|authentication=].
     </div>
 
+: Client extension processing ([=registration extension|registration=])
```
nsatragno, line 5854: We should indicate that creating a credential with the `largeBlobs` extension is required in order to store and read a blob on assertion later regardless of authenticator support. Otherwise, we still have to be Cute (tm) and always attempt to create a largeBlobKey.

index.bs
```diff
@@ -5820,30 +5820,51 @@ Note: In order to interoperate, user agents storing large blobs on authenticator
...
+::
+       1. If {{AuthenticationExtensionsLargeBlobInputs/read}} or {{AuthenticationExtensionsLargeBlobInputs/write}} is present:
+           1. Return a {{DOMException}} whose name is “{{NotSupportedError}}”.
+       1. If {{AuthenticationExtensionsLargeBlobInputs/support}} is present and has the value “required”:
```
nsatragno, line 5858: It seems to me that there is no point in using the registration extension with a `AuthenticationExtensionsLargeBlobInputs/support` that is not present. Perhaps we should default to `preferred` or throw an error.

index.bs
agl: Done.

index.bs
agl: Done. (Defaulting to “preferred”.)

akshayku (approved):

index.bs
equalsJeffH: I do not see where we "default to 'preferred'" ?

index.bs
```diff
@@ -5820,30 +5820,51 @@ Note: In order to interoperate, user agents storing large blobs on authenticator
...
+           1. Set {{AuthenticationExtensionsLargeBlobOutputs/supported}} to [TRUE].
+           1. When creating the credential, prior to evaluating the [=authenticator attachment modality=], [=iteration/continue=] (i.e. ignore the authenticator) if the authenticator is not capable of storing large blobs.
```
equalsJeffH, line 5860: what happens if no authenticator is capable of storing large blobs ?

index.bs
```diff
@@ -5820,30 +5820,51 @@ Note: In order to interoperate, user agents storing large blobs on authenticator
...
+       1. Otherwise:
+           1. Set {{AuthenticationExtensionsLargeBlobOutputs/supported}} to [TRUE] if the authenticator that created the fresh credential supports large blobs and [FALSE] otherwise.
```
equalsJeffH, line 5862: perhaps..
```suggestion
           1. Set {{AuthenticationExtensionsLargeBlobOutputs/supported}} to [TRUE] if the [=create/selected authenticator=] supports large blobs and [FALSE] otherwise.
```
..and up in the [`[[Create]]()` alg](https://w3c.github.io/webauthn/#sctn-createCredential) step 19, case "If any authenticator indicates success," step 1, which says:
>        1. Remove authenticator from issuedRequests.

..change it to be:
>        1. Remove authenticator from issuedRequests. This authenticator is now the <dfn for="create">selected authenticator</dfn>.

...?

index.bs
```diff
@@ -5820,30 +5820,51 @@ Note: In order to interoperate, user agents storing large blobs on authenticator
...
     </div>
 
+: Client extension processing ([=registration extension|registration=])
+::
+       1. If {{AuthenticationExtensionsLargeBlobInputs/read}} or {{AuthenticationExtensionsLargeBlobInputs/write}} is present:
+           1. Return a {{DOMException}} whose name is “{{NotSupportedError}}”.
+       1. If {{AuthenticationExtensionsLargeBlobInputs/support}} is present and has the value “required”:
+           1. Set {{AuthenticationExtensionsLargeBlobOutputs/supported}} to [TRUE].
+           1. When creating the credential, prior to evaluating the [=authenticator attachment modality=], [=iteration/continue=] (i.e. ignore the authenticator) if the authenticator is not capable of storing large blobs.
+       1. Otherwise:
```
equalsJeffH, line 5861: `support` is absent or has value "preferred" ?

index.bs
```diff
@@ -5820,30 +5820,51 @@ Note: In order to interoperate, user agents storing large blobs on authenticator
...
+           1. Set {{AuthenticationExtensionsLargeBlobOutputs/supported}} to [TRUE] if the authenticator that created the fresh credential supports large blobs and [FALSE] otherwise.
+
 : Client extension processing ([=authentication extension|authentication=])
 ::
+       1. If {{AuthenticationExtensionsLargeBlobInputs/support}} is present:
+           1. Return a {{DOMException}} whose name is “{{NotSupportedError}}”.
        1. If {{AuthenticationExtensionsLargeBlobInputs/read}} is present and has the value `true`:
            1. If the [[#sctn-getAssertion|assertion]] operation is successful, attempt to read any data associated with the asserted credential.
```
equalsJeffH, line 5869:
```suggestion
           1. If the [=assertion=] operation is successful, attempt to read any data associated with the asserted credential.
```
What happens if for some reason read is not successful?

index.bs
```diff
@@ -5820,30 +5820,51 @@ Note: In order to interoperate, user agents storing large blobs on authenticator
...
+           1. Set {{AuthenticationExtensionsLargeBlobOutputs/supported}} to [TRUE].
+           1. When creating the credential, prior to evaluating the [=authenticator attachment modality=], [=iteration/continue=] (i.e. ignore the authenticator) if the authenticator is not capable of storing large blobs.
+       1. Otherwise:
+           1. Set {{AuthenticationExtensionsLargeBlobOutputs/supported}} to [TRUE] if the authenticator that created the fresh credential supports large blobs and [FALSE] otherwise.
+
 : Client extension processing ([=authentication extension|authentication=])
 ::
+       1. If {{AuthenticationExtensionsLargeBlobInputs/support}} is present:
+           1. Return a {{DOMException}} whose name is “{{NotSupportedError}}”.
        1. If {{AuthenticationExtensionsLargeBlobInputs/read}} is present and has the value `true`:
```
equalsJeffH, line 5868:
```suggestion
       1. If {{AuthenticationExtensionsLargeBlobInputs/read}} is present and has the value [TRUE]:
```

equalsJeffH (changes requested):
It looks to me like one can include both `read` and `write` in a largeBlob extension, and the `read` may succeed (tho there's no signal if it does not), and the `write` will be silently ignored.  is that what's intended?  further modest comments below...

index.bs
agl: The process would continue to wait for an applicable authenticator. I.e. same as if rk=required and no existing authenticator supported resident keys.

index.bs
emlun, line 5858: I suggest linking this:

```suggestion
       1. If {{AuthenticationExtensionsLargeBlobInputs/support}} is present and has the value {{LargeBlobSupport/required}}:
```

index.bs
```diff
@@ -5858,12 +5879,16 @@ Note: In order to interoperate, user agents storing large blobs on authenticator
...
     dictionary AuthenticationExtensionsLargeBlobOutputs {
+        boolean supported;
         ArrayBuffer blob;
         boolean written;
     };
     </xmp>
 
     <div dfn-type="dict-member" dfn-for="AuthenticationExtensionsLargeBlobOutputs">
+        :   <dfn>supported</dfn>
+        ::  [TRUE] if, and only if, the created credential supports storing large blobs. Only present after [=registration extension|registration=].
```
emlun, line 5890: >Only present after registration.

I first read this as "present in ceremonies subsequent to a registration ceremony", which is clearly incorrect if you stop to think about it, but perhaps it could be a wee bit clearer. Perhaps something like this?

```suggestion
        ::  [TRUE] if, and only if, the created credential supports storing large blobs. Only present in [=registration extension|registration=] outputs.
```

agl:
> It looks to me like one can include both read and write in a largeBlob extension, and the read may succeed (tho there's no signal if it does not), and the write will be silently ignored. is that what's intended? further modest comments below...

Have explicitly banned that. Have also clarified how a read failure is indicated.

index.bs

nsatragno (approved):
PR still looks good to me

index.bs
```diff
@@ -5820,34 +5820,63 @@ Note: In order to interoperate, user agents storing large blobs on authenticator
...
 ::
-       1. If {{AuthenticationExtensionsLargeBlobInputs/read}} is present and has the value `true`:
-           1. If the [[#sctn-getAssertion|assertion]] operation is successful, attempt to read any data associated with the asserted credential.
+       1. If {{AuthenticationExtensionsLargeBlobInputs/support}} is present:
+           1. Return a {{DOMException}} whose name is “{{NotSupportedError}}”.
+       1. If both {{AuthenticationExtensionsLargeBlobInputs/read}} and {{AuthenticationExtensionsLargeBlobInputs/write}} are present:
+           1. Return a {{DOMException}} whose name is “{{NotSupportedError}}”.
+       1. If {{AuthenticationExtensionsLargeBlobInputs/read}} is present and has the value [TRUE]:
+           1. Initialize the [=client extension output=], {{AuthenticationExtensionsClientOutputs/largeBlob}}.
+           1. If the [=assertion=] operation is successful, attempt to read any data associated with the asserted credential.
```
equalsJeffH, line 5872: perhaps this is more clear where in [[DiscoverFromExt...]]() this occurs...
```suggestion
           1. If any authenticator indicates success (in {{PublicKeyCredential/[[DiscoverFromExternalSource]]()}}), attempt to read any largeBlob data associated with the asserted credential.
```

equalsJeffH (approved):
Looks overall good to me, thx, tho have some comments for your consideration.

emlun (approved):

index.bs
agl: Thanks. Have made the suggested updates.

index.bs
```diff
@@ -5820,34 +5822,66 @@ Note: In order to interoperate, user agents storing large blobs on authenticator
...
 
+    enum LargeBlobSupport {
+      "required",
+      "preferred",
+    };
+
     dictionary AuthenticationExtensionsLargeBlobInputs {
+        DOMString support;
         boolean read;
         ArrayBuffer write;
```
nsatragno, line 5841: Semi unrelated to this PR, but should this be a `BufferSource` instead? That's what we're using elsewhere for byte sequences.

agl: Yep! Thanks. Done.

equalsJeffH (approved):
LGTM, thx @agl!

agl:
From the call of 2020-09-30: clear to merge by the end of the week unless John has comments.

ve7jtb (approved):
LGTM

PR #1472, opened 2020-08-24
Virtual authenticators extension support

### LINKED ISSUE #1471: Virtual authenticator support for extensions
At the moment, the [User Agent Automation section](https://w3c.github.io/webauthn/#sctn-automation) does not say anything about treatment of extensions by virtual authenticators. As user agents implement the extensions, we should strive to support them through the virtual authenticators so that relying parties can write automated tests and we can ensure interoperability with web platform tests.

My proposal is to add
- for every defined extension, a capability with the name `webauthn:$extension_identifier` signalling virtual authenticator support for the extension, and
- an `extensions` array to the `Authenticator Configuration` object, indicating which extensions must be supported by the specific virtual authenticator. Extensions not present in the array should not be supported (this is to test that a given relying party or user agent does the right thing when the underlying authenticator does not support an extension).
- when defining an extension, the text
  > User agents implementing $extension SHOULD support the `webauthn:$extension_identifier` webdriver capability

nsatragno:
Add support for authenticator extensions on WebDriver virtual authenticators.

This comes in the form of:
* A capability for each extension signalling support.
* An option to toggle which extensions are supported by a given virtual authenticator.

Fixes #1471

***
<a href="https://pr-preview.s3.amazonaws.com/nsatragno/webauthn/pull/1472.html" title="Last updated on Sep 2, 2020, 3:52 PM UTC (ea9ad16)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/1472/ad8d1d0...nsatragno:ea9ad16.html" title="Last updated on Sep 2, 2020, 3:52 PM UTC (ea9ad16)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -5469,7 +5469,7 @@ Instead, in step three, the comparison on the host is relaxed to accept hosts on
 : Authenticator extension output
 :: None.
 
-## FIDO <dfn>AppID Exclusion</dfn> Extension (appidExclude) ## {#sctn-appid-exclude-extension}
+## FIDO AppID Exclusion Extension (appidExclude) ## {#sctn-appid-exclude-extension}
 
 This registration extension allows [=[WRPS]=] to exclude authenticators that contain specified credentials that were created with the legacy FIDO U2F JavaScript API [[FIDOU2FJavaScriptAPI]].
 
@@ -5933,38 +5933,20 @@ Additionally, [=extension capabilities=] are defined for every [=authenticator e
         </thead>
         <tbody>
             <tr>
-                <td>FIDO AppID Extension Support</td>
-                <td>`"webauthn:appid"`</td>
-                <td>boolean</td>
-                <td>Indicates whether the [=endpoint node=] WebAuthn WebDriver implementation supports the [=AppID=] extension.</td>
-            </tr>
-            <tr>
-                <td>FIDO AppID Exclusion Extension Support</td>
-                <td>`"webauthn:appidExclude"`</td>
-                <td>boolean</td>
-                <td>Indicates whether the [=endpoint node=] WebAuthn WebDriver implementation supports the [=AppID Exclusion=] extension.</td>
-            </tr>
-            <tr>
                 <td>User Verification Method Extension Support</td>
-                <td>`"webauthn:uvm"`</td>
+                <td>`"webauthn:extension:uvm"`</td>
                 <td>boolean</td>
                 <td>Indicates whether the [=endpoint node=] WebAuthn WebDriver implementation supports the [=User Verification Method=] extension.</td>
             </tr>
             <tr>
-                <td>Credential Properties Extension Support</td>
-                <td>`"webauthn:credProps"`</td>
-                <td>boolean</td>
-                <td>Indicates whether the [=endpoint node=] WebAuthn WebDriver implementation supports the [=credProps=] extension.</td>
-            </tr>
-            <tr>
                 <td>Pseudo-Random Function Extension Support</td>
-                <td>`"webauthn:prf"`</td>
+                <td>`"webauthn:extension:prf"`</td>
                 <td>boolean</td>
                 <td>Indicates whether the [=endpoint node=] WebAuthn WebDriver implementation supports the [=prf=] extension.</td>
             </tr>
             <tr>
                 <td>Large Blob Storage Extension Support</td>
-                <td>`"webauthn:largeBlob"`</td>
+                <td>`"webauthn:extension:largeBlob"`</td>
                 <td>boolean</td>
                 <td>Indicates whether the [=endpoint node=] WebAuthn WebDriver implementation supports the [=largeBlob=] extension.</td>
             </tr>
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -5943,6 +6012,11 @@ Each stored [=virtual authenticator=] has the following properties:
     [=User Verification=] will always succeed. If set to [FALSE], it will fail.
 
     Note: This property has no effect if |hasUserVerification| is set to [FALSE].
+: |extensions|
+:: A string array containing the [=extension identifiers=] supported by the [=Virtual Authenticator=].
+
+[=Virtual authenticators=] MUST support all [=authenticator extensions=] present in its |extensions| array.
+They MUST NOT support any [=authenticator extension=] not present in its |extensions| array.
```
equalsJeffH, line 6019: If the virtual authnr does not support any extensions, is this property's value an empty array, or is the property absent? Ah, I see below in "Add Virtual Authenticator" that the default is an empty array.

emlun, line 6019: There's a mix of plural and singular here. I would suggest:

```suggestion
A [=virtual authenticators=] MUST support all [=authenticator extensions=] present in its |extensions| array.
It MUST NOT support any [=authenticator extension=] not present in its |extensions| array.
```

index.bs
```diff
@@ -5916,6 +5917,74 @@ When [=matching capabilities=], the extension-specific steps to match `"webauthn
...
+                <th>Capability</th>
+                <th>Key</th>
+                <th>Value Type</th>
+                <th>Description</th>
+            </tr>
+        </thead>
+        <tbody>
+            <tr>
+                <td>FIDO AppID Extension Support</td>
+                <td>`"webauthn:appid"`</td>
```
emlun, line 5937: How about using a second namespace level for extensions, like `webauthn:extension:appid` or `webauthn:ext:appid`? I'm thinking capabilities like `webauthn:pin-userVerification` or something like that might be useful at some point, maybe for features added in the future, and it seems appropriate to separate the extensions from other features.

index.bs
emlun: Might also be appropriate to indent this as a second paragraph of the `|extensions|` definition? I'm not sure.

emlun (commented):
Maybe this should also include a new step in the remote end steps for "add credential" and "get credentials"?

index.bs
```diff
@@ -5916,6 +5917,56 @@ When [=matching capabilities=], the extension-specific steps to match `"webauthn
         the match is unsuccessful.
     2. Otherwise, the match is successful.
 
+### <dfn>Authenticator Extension Capabilities</dfn> ### {#sctn-authenticator-extension-capabilities}
+
+Additionally, [=extension capabilities=] are defined for every [=authenticator extension=] defined in this specification:
```
emlun, line 5922: This "every" is no longer accurate after the latest update. :)

index.bs
```diff
@@ -5943,6 +5994,15 @@ Each stored [=virtual authenticator=] has the following properties:
...
     Note: This property has no effect if |hasUserVerification| is set to [FALSE].
+: |extensions|
+:: A string array containing the [=extension identifiers=] supported by the [=Virtual Authenticator=].
+
+    A [=Virtual authenticator=] MUST support all [=authenticator extensions=] present in its |extensions| array.
+    It MUST NOT support any [=authenticator extension=] not present in its |extensions| array.
+: |uvm|
+:: A {{UvmEntries}} array to be set as the [=authenticator extension output=] when processing the [=User Verification Method=] extension.
+
+   Note: This property has no effect if the [=Virtual Authenticator=] does not support the [=User Verification Method=] extension.
```
emlun, line 6005: Heh, this needs one more space of indentation to render as intended.

```suggestion
    Note: This property has no effect if the [=Virtual Authenticator=] does not support the [=User Verification Method=] extension.
```

nsatragno: Something something meaningful indentation is evil. Done!

index.bs
nsatragno: I think it's still true.

> An extension can also be an *authenticator extension*, meaning that the extension involves communication with and processing by the authenticator.

The extensions I left here are precisely those that involve processing by the authenticator, i.e. those for which "Authenticator extension processing" is not empty.

--

I agree this text can be confusing so I added a clarification.

emlun: Oh right, sorry, this was indeed accurate. So I don't think the clarification is really needed, but I don't mind keeping it in either.

emlun (approved):
LGTM, thanks @nsatragno!

jcjones:
So I just want to make sure I understand properly - the idea here is that _if_ the WPTs can expect the Virtual Authenticator to support a given extension, _then_ the capabilities reported via WebDriver to the test should indicate that extension?

nsatragno:
> So I just want to make sure I understand properly - the idea here is that if the WPTs can expect the Virtual Authenticator to support a given extension, then the capabilities reported via WebDriver to the test should indicate that extension?

Yes, that is correct.

jcjones (approved):

akshayku (approved):

Issue #1453, opened 2020-07-08
add additional attestation format(s) from Apple


equalsJeffH:
Apple has indicated they have their own [attestation statement format(s)](https://www.iana.org/assignments/webauthn/webauthn.xhtml) that they'd like to have included in the spec. this is just a placeholder to keep it in mind.

equalsJeffH:
on 2020-08-26 call:  jiewan is still awaiting for internal reply, it'll be only one attestation.

mackie1001:
As of iOS 14 Beta 6 it appears the new attestation format is now in play. Based on conversations you've had is this subject to change or are we just waiting on official documentation? 
I see that some have started looking into is using the DeviceCheck/AppAttest docs as a starting point but do you know of any working reference implementations?

equalsJeffH:
we are waiting apple to provide a specification of their [attestation statement format(s)](https://www.iana.org/assignments/webauthn/webauthn.xhtml)

mpeng-okta:
Quick question @equalsJeffH: I see that this has been added to the [WebAuthn Level 2 editor draft](https://w3c.github.io/webauthn/#sctn-apple-anonymous-attestation). Are there any plans on adding the attestation format  the existing WebAuthn Level 1 spec as well, and if not, does this mean an implementation of the Level 1 spec should ignore the `apple` attestation format?

I also see that it hasn't been registered yet [on IANA's page for attestation format identifiers](https://www.iana.org/assignments/webauthn/webauthn.xhtml). Would this have any implications for RP developers, or would it be fine to begin building support for this attestation format -- i.e. we could consider the IANA registration just a procedural step/formality?

PR #1474, opened 2020-08-27
Abstracting the concept of Privacy CA/Attestation CA into Anonymization CA

### LINKED ISSUE #1422: more fully delineate "privacy ca", "attestation ca", "anonymization ca"
the "privacy ca", "attestation ca", "anonymization ca" terms appear in the spec in various places and the Note in [14.4.1. Attestation Privacy](https://w3c.github.io/webauthn/#sctn-attestation-privacy) attempts to tease them apart, but could do a better job.  Am imagining appending a bulleted list summarizing each's particular characteristics, or re-writing the Note to more clearly delineate between them.

alanwaketan:
This patch defines Anonymous CA which only gathers abstract common facts from Privacy CA/Attestation CA like technology such that different authenticator vendors can refer to it without refering to the TCG definition.

This PR is to fix #1422.

***
<a href="https://pr-preview.s3.amazonaws.com/alanwaketan/webauthn/pull/1474.html" title="Last updated on Oct 28, 2020, 7:08 PM UTC (702abf6)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/1474/3d38cab...alanwaketan:702abf6.html" title="Last updated on Oct 28, 2020, 7:08 PM UTC (702abf6)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -4272,7 +4272,7 @@ under [=[RP]=] policy.
 :: In this case, the [=authenticator=] works with a cloud-operated [=Anonymization CA=] owned by its manufacturer to dynamically generate per-[=credential=] [=attestation certificates=] on the CA such that no identification information of the [=authenticator=] will be revealed to [=[RPS]=] in the [=attestation statement=].
 
     Note: [=Attestation statements=] conveying [=attestations=] of [=attestation type|type=] [=AttCA=] or [=AnonCA=] use the same data structure
-        as [=attestation statements=] conveying [=attestations=] of [=attestation type|type=] [=Basic=], so the three attestation types
+        as those of [=attestation type|type=] [=Basic=], so the three attestation types
         are, in general, distinguishable only with externally provided knowledge regarding the contents of the [=attestation
         certificates=] conveyed in the [=attestation statement=].
 
--- a/index.bs
+++ b/index.bs
@@ -2778,7 +2778,7 @@ Note: The {{AttestationConveyancePreference}} enumeration is deliberately not re
     :   <dfn>none</dfn>
     ::  This value indicates that the [=[RP]=] is not interested in [=authenticator=] [=attestation=]. For example, in order to
         potentially avoid having to obtain [=user consent=] to relay identifying information to the [=[RP]=], or to save a
-        roundtrip to an [=Attestation CA=].
+        roundtrip to an [=Attestation CA=] or [=Anonymization CA=].
 
         This is the default value.
 
@@ -4264,10 +4264,13 @@ calling {{CredentialsContainer/create()|navigator.credentials.create()}} they se
     Note: This concept typically leads to multiple attestation certificates. The attestation certificate requested most recently
         is called "active".
 
-    Note: [=Attestation statements=] conveying [=attestations=] of [=attestation type|type=] [=AttCA=] use the same data structure
-    as [=attestation statements=] conveying [=attestations=] of [=attestation type|type=] [=Basic=], so the two attestation types
-    are, in general, distinguishable only with externally provided knowledge regarding the contents of the [=attestation
-    certificates=] conveyed in the [=attestation statement=].
+: <dfn>Anonymization CA</dfn> (<dfn>AnonCA</dfn>)
+:: In this case, the [=authenticator=] works with a cloud-operated [=Anonymization CA=] owned by its manufacturer to dynamically generate per-[=credential=] [=attestation certificates=] on the CA such that no identification information of the [=authenticator=] will be revealed to [=[RPS]=] in the [=attestation statement=].
+
+    Note: [=Attestation statements=] conveying [=attestations=] of [=attestation type|type=] [=AttCA=] or [=AnonCA=] use the same data structure
+        as [=attestation statements=] conveying [=attestations=] of [=attestation type|type=] [=Basic=], so the three attestation types
+        are, in general, distinguishable only with externally provided knowledge regarding the contents of the [=attestation
+        certificates=] conveyed in the [=attestation statement=].
 
 : No attestation statement (<dfn>None</dfn>)
 :: In this case, no attestation information is available. See also [[#sctn-none-attestation]].
@@ -6717,7 +6720,7 @@ instead of revealing the biometric data itself to the [=[RP]=].
 
 ### Attestation Privacy ### {#sctn-attestation-privacy}
 
-[=Attestation certificates=] and [=attestation key pairs=] and can be used to track users
+[=Attestation certificates=] and [=attestation key pairs=] can be used to track users
 or link various online identities of the same user together.
 This can be mitigated in several ways, including:
 
@@ -6733,16 +6736,14 @@ This can be mitigated in several ways, including:
     sufficiently large groups. This may serve as guidance about suitable batch sizes.
 
 - A [=[WAA]=] may be capable of dynamically generating different [=attestation key pairs=] (and requesting related
-    [=attestation certificate|certificates=]) per-[=origin=]
-    (similar to the [=Attestation CA=] approach). For example, an [=authenticator=] can ship with a
+    [=attestation certificate|certificates=]) per-[=credential=] as described in the [=Anonymization CA=] approach. For example, an [=authenticator=] can ship with a
     master [=attestation private key=] (and [=attestation certificate|certificate=]),
-    and combined with a cloud-operated <dfn>Anonymization CA</dfn>,
-    can dynamically generate per-[=origin=] [=attestation key pairs=] and [=attestation certificates=].
+    and combined with a cloud-operated [=Anonymization CA=],
+    can dynamically generate per-[=credential=] [=attestation key pairs=] and [=attestation certificates=].
 
     Note: In various places outside this specification, the term "Privacy CA" is used to refer to what is termed here
         as an [=Anonymization CA=]. Because the Trusted Computing Group (TCG) also used the term "Privacy CA" to refer to what
-        the TCG now refers to as an [=Attestation CA=] (ACA) [[!TCG-CMCProfile-AIKCertEnroll]], and the envisioned functionality
-        of an [=Anonymization CA=] is not firmly established, we are using the term [=Anonymization CA=] here to try to mitigate
+        the TCG now refers to as an [=Attestation CA=] (ACA) [[!TCG-CMCProfile-AIKCertEnroll]], we are using the term [=Anonymization CA=] here to try to mitigate
         confusion in the specific context of this specification.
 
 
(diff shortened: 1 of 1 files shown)


alanwaketan:
Should I add myself to the contributor list manually?

emlun:
> you are generalizing the TCG's [Attestation CA](https://w3c.github.io/webauthn/#attestation-ca) into the [Anonymization CA notion](https://w3c.github.io/webauthn/#anonymization-ca). [...] and denote [Attestation CA] as a particular instance of an Anonymization CA.

I think the categories make more sense the other way around: that an "anonymization CA" is one subtype of the broader class "attestation CA". Any solution of this kind is one where some other CA (whether hosted in the cloud or on the client device) than the authenticator is used to perform attestation - that attestation _might_ anonymize the attestation, but in theory it could just as well generate a uniquely identifiable attestation. Or it might preserve much the same information as the authenticator's original attestation statement, but add additional information like, say, various certification levels.

So if the goal is to abstract or generalize the concept, then replacing the term "attestation CA" with "anonymization CA" seems counterproductive, because the latter is less general than the former.

equalsJeffH:
> I think the categories make more sense the other way around: that an "anonymization CA" is one subtype of the broader class "attestation CA"

TCG has co-opted the "attestation CA" term, hence the present terminology.

alanwaketan:
> In [section 6.5.3](https://w3c.github.io/webauthn/#sctn-attestation-types) you are generalizing the TCG's [Attestation CA](https://w3c.github.io/webauthn/#attestation-ca) into the [Anonymization CA notion](https://w3c.github.io/webauthn/#anonymization-ca). We will have to think about that, it _might_ be ok to do so, but only if we retain the present Attestation CA text elsewhere in the spec (perhaps as a subsubsection of [section 6.5.3](https://w3c.github.io/webauthn/#sctn-attestation-types)), and denote it as a particular instance of an Anonymization CA.
> 

Sounds good to me.

> I am guessing that once y'all create a PR addressing issue #1453 (which, ISTM, may add a new specific instance of an Anonymization CA, and may add a new [attestation statement format](https://www.iana.org/assignments/webauthn/webauthn.xhtml#webauthn-attestation-statement-format-ids)) we will be able to properly complete this PR.

I don't foresee myself describing Apple Anonymous Attestation (AAA) in a such detailed way TCG does. In fact, besides the description putting here, I don't think I will write anything else to outline the functionality of AAA.

alanwaketan:
> > you are generalizing the TCG's [Attestation CA](https://w3c.github.io/webauthn/#attestation-ca) into the [Anonymization CA notion](https://w3c.github.io/webauthn/#anonymization-ca). [...] and denote [Attestation CA] as a particular instance of an Anonymization CA.
> 
> I think the categories make more sense the other way around: that an "anonymization CA" is one subtype of the broader class "attestation CA". Any solution of this kind is one where some other CA (whether hosted in the cloud or on the client device) than the authenticator is used to perform attestation - that attestation _might_ anonymize the attestation, but in theory it could just as well generate a uniquely identifiable attestation. Or it might preserve much the same information as the authenticator's original attestation statement, but add additional information like, say, various certification levels.
> 
> So if the goal is to abstract or generalize the concept, then replacing the term "attestation CA" with "anonymization CA" seems counterproductive, because the latter is less general than the former.

Sounds like TCG's attestation CA is on device? Weird, it sounds like the intermediate CA's certificate is the tracking vector.

emlun:
> TCG has co-opted the "attestation CA" term, hence the present terminology.

Ah, I see. Well, I recall that we were at some point talking about "attestation proxy" or "proxy attestation" as an attestation type. If we do need a more abstract term, maybe we could use something like that?

equalsJeffH:
on 2020-08-09 call:  @alanwaketan is waiting for permission to upload their attestation stmt format, then we can proceed on this as appropriate

equalsJeffH:
on 2020-10-7 call:
we have new PR #1491 from @alanwaketan (thx!), so let's refine that and then refine this one as appropriate and then land them as appropriate.

alanwaketan:
After revisiting this issue, I think it could be hard to unite the concept of TCG's attestation CA with Apple Anonymous Attestation. The only common thing here is that a CA is involved but the usage of the CA is quite different.

sbweeden:
My suggestion here is to leave AttCA as-is. Leave TPM attestation assigned as using AttCA (again as-is). Add AnonymizationCA if desired to differentiate what Apple is doing. I think the description proposed for AnonymizationCA is good, and worth adding. It doesn't need to necessarily alter what AttCA means. 

Further, I would then add an explanatory note (which is useful to RP developers) indicating that effectively Basic, AttCA and AnonymizationCA, if being validated by an RP, all result in an X5C certificate chain verification process against a discovered root (e.g. with MDS as currently described).

equalsJeffH:
@sbweeden's above suggestion https://github.com/w3c/webauthn/pull/1474#issuecomment-706889314 sounds good to me -- who is able to concoct a PR effecting all that?

sbweeden:
If a new AnonymizationCA concept is introduced, and if there were to ever be an Apple FIDO2 MDS document published (rather than just the trust root CA), it would make sense that the MDS document would be able to contain a legitimate value for the attestationTypes attribute (see https://fidoalliance.org/specs/fido-v2.0-id-20180227/fido-metadata-statement-v2.0-id-20180227.html#idl-def-MetadataStatement). As a result, and just an FYI for this PR, I have opened a corresponding issue to discuss in FIDO (https://github.com/fido-alliance/common-specs/issues/164).

alanwaketan:
> @sbweeden's above suggestion [#1474 (comment)](https://github.com/w3c/webauthn/pull/1474#issuecomment-706889314) sounds good to me -- who is able to concoct a PR effecting all that?

I can update the PR to reflect that and will coordinate with Shane.

sbweeden:
@equalsJeffH @emlun - This PR has been worked by @alanwaketan and myself this week, and should be ready for review. Giving advance notice so that we can close on any gaps asap for L2.

index.bs
```diff
@@ -4264,10 +4268,13 @@ calling {{CredentialsContainer/create()|navigator.credentials.create()}} they se
...
 
-    Note: [=Attestation statements=] conveying [=attestations=] of [=attestation type|type=] [=AttCA=] use the same data structure
-    as [=attestation statements=] conveying [=attestations=] of [=attestation type|type=] [=Basic=], so the two attestation types
-    are, in general, distinguishable only with externally provided knowledge regarding the contents of the [=attestation
-    certificates=] conveyed in the [=attestation statement=].
+: <dfn>Anonymization CA</dfn> (<dfn>AnonCA</dfn>)
+:: In this case, the [=authenticator=] works with a cloud-operated [=Anonymization CA=] owned by its manufacturer to dynamically generate per-[=credential=] [=attestation certificates=] on the CA such that no identification information of the [=authenticator=] will be revealed to [=[RPS]=] in the [=attestation statement=].
+
+    Note: [=Attestation statements=] conveying [=attestations=] of [=attestation type|type=] [=AttCA=] or [=AnonCA=] use the same data structure
+        as [=attestation statements=] conveying [=attestations=] of [=attestation type|type=] [=Basic=], so the three attestation types
```
emlun, line 4275: Minor suggestion, feel free to ignore:

```suggestion
        as those of [=attestation type|type=] [=Basic=], so the three attestation types
```

emlun (approved):
LGTM, thanks!

emlun (approved):

alanwaketan:
@akshayku What do you think now?

index.bs
```diff
@@ -4269,7 +4269,7 @@ under [=[RP]=] policy.
         is called "active".
 
 : <dfn>Anonymization CA</dfn> (<dfn>AnonCA</dfn>)
-:: In this case, the [=authenticator=] works with a cloud-operated [=Anonymization CA=] owned by its manufacturer to dynamically generate per-[=credential=] [=attestation certificates=] on the CA such that no identification information of the [=authenticator=] will be revealed to [=[RPS]=] in the [=attestation statement=].
+:: In this case, the [=authenticator=] works with a cloud-operated [=Anonymization CA=] owned by its manufacturer to dynamically generate per-[=credential=] [=attestation certificates=] on the CA such that no identification information of an individual [=authenticator=] will be revealed to [=[RPS]=] in the [=attestation statement=].
```
timcappalli, line 4272: "cloud-operated" sounds a bit weird. Should it be "cloud-hosted" instead?

index.bs
```diff
@@ -4264,10 +4268,13 @@ calling {{CredentialsContainer/create()|navigator.credentials.create()}} they se
     Note: This concept typically leads to multiple attestation certificates. The attestation certificate requested most recently
         is called "active".
 
-    Note: [=Attestation statements=] conveying [=attestations=] of [=attestation type|type=] [=AttCA=] use the same data structure
-    as [=attestation statements=] conveying [=attestations=] of [=attestation type|type=] [=Basic=], so the two attestation types
-    are, in general, distinguishable only with externally provided knowledge regarding the contents of the [=attestation
-    certificates=] conveyed in the [=attestation statement=].
+: <dfn>Anonymization CA</dfn> (<dfn>AnonCA</dfn>)
+:: In this case, the [=authenticator=] works with a cloud-operated [=Anonymization CA=] owned by its manufacturer to dynamically generate per-[=credential=] [=attestation certificates=] on the CA such that no identification information of an individual [=authenticator=] will be revealed to [=[RPS]=] in the [=attestation statement=].
```
akshayku, line 4272: Does it needs to be owned by the manufacturer?

ve7jtb: Perhaps owned or operated on behalf of or in cooperation with the manufacturer may cover more cases.
In the two cases I can think of as I consider Android saftyNet to be a very similar thing. They are both operated by the OS/platform manufacturer, and not neccicarily atleast in the android case the device manufacturer.

ve7jtb, line 4272:
```suggestion
:: In this case, the [=authenticator=] works with a cloud-operated [=Anonymization CA=] owned by, or operated on behalf of its manufacturer to dynamically generate per-[=credential=] [=attestation certificates=] on the CA such that no identification information of an individual [=authenticator=] will be revealed to [=[RPS]=] in the [=attestation statement=].
```

ve7jtb (approved):
Approved with one suggestion.

index.bs
equalsJeffH: I suggest re-wording this like so:
```suggestion
:: In this case, the [=authenticator=] uses an [=Anonymization CA=] which dynamically generates per-[=credential=] [=attestation certificates=] such that the [=attestation statements=] presented to [=[RPS]=] do not provide uniquely identifiable information, e.g., that might be used for tracking purposes.
```

equalsJeffH (changes requested):
overall this is ok, but I'd re-word the AnonCA description...

index.bs
akshayku: Another question: Why does this needs to be cloud operated?

akshayku (approved):

equalsJeffH (approved):
lgtm! ;-)

PR #1480, opened 2020-09-07
Add security consideration on detecting non-WebAuthn accounts

### LINKED ISSUE #1475: The risk of attacker may can identify whether if the account support FIDO or not
Hi this is Keiko Itakura from Rakuten which is member of both W3C and FIDO Alliance Japan.

We(FIDO Japan WG) wonder there is a security risk in the following in WebAuthn.  I would like to know what W3C thinks of this, and also if there is a plan to reflect changes in the WebAuthn specs about this possible risk.

- Isn't there a risk that the attacker can identify whether or not the account supports FIDO?

There is a possibility the CredentialID/Username pair is exposed in the case of NRK(Non resident key) .  
For example,
- Let's say an attacker sends Usernames that can guess e-mail addresses etc. to the RP. 
- Then he can see the repllies from the RP servers for each.
- If a CredentialID-like thing is returned, and if it shows whether or not the account supports FIDO(The reply is different from the case where FIDO is not supported. )
- then the attacker can select the accounts that do not support FIDO as target.
If they use this, they can see that unsupported ones have weak security and are easy to attack. Do you not think this is a security issue?  And if it is a problem, how can this problem be addressed?

emlun:
Fixes #1475.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/pull/1480.html" title="Last updated on Sep 10, 2020, 5:21 PM UTC (84cf69e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/1480/b4148f2...84cf69e.html" title="Last updated on Sep 10, 2020, 5:21 PM UTC (84cf69e)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -6562,6 +6562,27 @@ devices=].
 different [=public key credential|credentials=] are [=bound credential|bound=] to different [=authenticators=].
 
 
+### Unprotected account detection ### {#sctn-unprotected-account-detection}
+
+[INFORMATIVE]
+
+This security consideration applies to [=[RPS]=] that support [=authentication ceremonies=]
+with a non-[=list/empty=] {{PublicKeyCredentialRequestOptions/allowCredentials}} argument as the first authentication step.
+For example, if using authentication with [=server-side credentials=] as the first authentication step.
+
+In this case the {{PublicKeyCredentialRequestOptions/allowCredentials}} argument risks leaking information
+about which user accounts have WebAuthn credentials registered and which do not,
+which may be a signal of account protection strength.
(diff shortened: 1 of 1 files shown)


equalsJeffH (approved):
LGTM, thanks @emlun!

equalsJeffH:
on 2020-09-09 call:  @ve7jtb notes that he's seen articles discussing this means of differentiating accounts to attack based on this approach.
@agl suggests a clarification to apply and @emlun will apply and merge.

ve7jtb (approved):
OK with the change to password from empty allow list

PR #1483, opened 2020-09-14
Add note about fixed/empty user handles

agl:
See fido-alliance/fido-2-specs#951

***
<a href="https://pr-preview.s3.amazonaws.com/agl/webauthn/pull/1483.html" title="Last updated on Sep 16, 2020, 7:11 PM UTC (ebe2e4a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/1483/faa8de8...agl:ebe2e4a.html" title="Last updated on Sep 16, 2020, 7:11 PM UTC (ebe2e4a)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2630,7 +2630,7 @@ credential.
         The [=user handle=] MUST NOT contain [PII] about the user, such as a username or e-mail address;
         see [[#sctn-user-handle-privacy]] for details.
 
-        Note: the [=user handle=] should not be a constant value across different accounts, even for [=non-discoverable credentials=], because some authenticators always create [=discoverable credentials=]. Thus a constant [=user handle=] would prevent a user from using such an authenticator with more than one account at the [=[RP]=]. Additionally, the empty value is known to have problems in practice so is a bad choice, even if a constant value is desired.
+        Note: the [=user handle=] should not be a constant value across different accounts, even for [=non-discoverable credentials=], because some authenticators always create [=discoverable credentials=]. Thus a constant [=user handle=] would prevent a user from using such an authenticator with more than one account at the [=[RP]=]. Additionally,  a [=user handle=] having an empty value is known to be problematic in practice so is a bad choice, even if a constant value is desired.
 
     :   <dfn>displayName</dfn>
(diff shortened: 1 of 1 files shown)


lgarron:
As someone seeing this paragraph without context, it's rather unclear. Would it be possible to specify what the username is "fixed" across here, or rather in which cases it would cause trouble? For example:

- a given registration (attempt)
- a given browser profile
- a given user for a given service
- all users of a given service

agl:
> As someone seeing this paragraph without context, it's rather unclear. Would it be possible to specify what the username is "fixed" across here

Good point. Let me try again. How's it now?

lgarron:
> > As someone seeing this paragraph without context, it's rather unclear. Would it be possible to specify what the username is "fixed" across here
> 
> Good point. Let me try again. How's it now?

Definitely more clear, thanks for the revision! :-D

index.bs
```diff
@@ -2630,6 +2630,8 @@ credential.
         The [=user handle=] MUST NOT contain [PII] about the user, such as a username or e-mail address;
         see [[#sctn-user-handle-privacy]] for details.
 
+        Note: the [=user handle=] should not be a constant value across different accounts, even for [=non-discoverable credentials=], because some authenticators always create [=discoverable credentials=]. Thus a constant [=user handle=] would prevent a user from using such an authenticator with more than one account at the [=[RP]=]. Additionally, the empty value is known to have problems in practice so is a bad choice, even if a constant value is desired.
```
equalsJeffH, line 2633: maybe?
```suggestion
        Note: the [=user handle=] should not be a constant value across different accounts, even for [=non-discoverable credentials=], because some authenticators always create [=discoverable credentials=]. Thus a constant [=user handle=] would prevent a user from using such an authenticator with more than one account at the [=[RP]=]. Additionally,  a [=user handle=] having an empty value is known to be problematic in practice so is a bad choice, even if a constant value is desired.
```

equalsJeffH (approved):
LGTM, thx @agl!

akshayku (approved):

jcjones (approved):

PR #1488, opened 2020-09-30
Clarify signature format for packed attestations et al

equalsJeffH:
This makes the existing format recommendation normative, clarifies that the existing sig format definitions are MUSTs, and notes that these are examples of how to handle sig algorithm values that are not presently listed here.

fix #1441

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/pull/1488.html" title="Last updated on Sep 30, 2020, 6:42 PM UTC (fcf2c13)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/1488/6156ba8...fcf2c13.html" title="Last updated on Sep 30, 2020, 6:42 PM UTC (fcf2c13)">Diff</a>

PR #1491, opened 2020-10-07
Adding Apple Anonymous Attestation Statement Format

### LINKED ISSUE #1453: add additional attestation format(s) from Apple

alanwaketan:
This patch adds a new section to illustrate how Apple Anonymous Attestation works in general.

This PR is to fix #1453.

***
<a href="https://pr-preview.s3.amazonaws.com/alanwaketan/webauthn/pull/1491.html" title="Last updated on Oct 21, 2020, 8:01 PM UTC (f36e826)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/1491/3d38cab...alanwaketan:f36e826.html" title="Last updated on Oct 21, 2020, 8:01 PM UTC (f36e826)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -5183,7 +5183,7 @@ This attestation statement format is exclusively used by Apple for certain types
     1. Verify that |attStmt| is valid CBOR conforming to the syntax defined above and perform CBOR decoding on it to extract the contained fields.
     2. Concatenate |authenticatorData| and |clientDataHash| to form |nonceToHash|.
     3. Perform SHA-256 hash of |nonceToHash| to produce |nonce|.
-    4. Verify |nonce| matches the value of the extension with OID ( 1.2.840.113635.100.8.2 ) in |credCert|.
+    4. Verify that |nonce| equals the value of the extension with OID ( 1.2.840.113635.100.8.2 ) in |credCert|.
     5. Verify credential public key matches the Subject Public Key of |credCert|.
     6. If successful, return implementation-specific values representing attestation type [=Anonymization CA=] and attestation trust path |x5c|.
 
--- a/index.bs
+++ b/index.bs
@@ -5184,7 +5184,7 @@ This attestation statement format is exclusively used by Apple for certain types
     2. Concatenate |authenticatorData| and |clientDataHash| to form |nonceToHash|.
     3. Perform SHA-256 hash of |nonceToHash| to produce |nonce|.
     4. Verify that |nonce| equals the value of the extension with OID ( 1.2.840.113635.100.8.2 ) in |credCert|.
-    5. Verify credential public key matches the Subject Public Key of |credCert|.
+    5. Verify that the [=credential public key=] equals the Subject Public Key of |credCert|.
     6. If successful, return implementation-specific values representing attestation type [=Anonymization CA=] and attestation trust path |x5c|.
 
 # <dfn>WebAuthn Extensions</dfn> # {#sctn-extensions}
--- a/index.bs
+++ b/index.bs
@@ -45,6 +45,7 @@ Former Editor: Angelo Liao, w3cid 94342, Microsoft, huliao@microsoft.com
 !Contributors: <a href="mailto:agl@google.com">Adam Langley</a> (Google)
 !Contributors: <a href="mailto:mandyam@qti.qualcomm.com">Giridhar Mandyam</a> (Qualcomm)
 !Contributors: <a href="mailto:nsatragno@google.com">Nina Satragno</a> (Google)
+!Contributors: <a href="mailto:jiewen_tan@apple.com">Jiewen Tan</a> (Apple)
 !Contributors: <a href="mailto:sweeden@au1.ibm.com">Shane Weeden</a> (IBM)
 !Contributors: <a href="mailto:mkwst@google.com">Mike West</a> (Google)
 !Contributors: <a href="mailto:jyasskin@google.com">Jeffrey Yasskin</a> (Google)
@@ -5137,6 +5138,55 @@ if the [=authenticator=] does not support [=attestation=].
 : Verification procedure
 :: Return implementation-specific values representing [=attestation type=] [=None=] and an empty [=attestation trust path=].
 
+## Apple Anonymous Attestation Statement Format ## {#sctn-apple-anonymous-attestation}
+
+This attestation statement format is exclusively used by Apple for certain types of Apple devices that support WebAuthn.
+
+: Attestation statement format identifier
+:: apple
+
+: Attestation types supported
+:: [=Anonymization CA=]
+
+: Syntax
+:: The syntax of an Apple attestation statement is defined as follows:
+
+    ```
+    $$attStmtType //= (
+                          fmt: "apple",
+                          attStmt: appleStmtFormat
+                      )
+
+    appleStmtFormat = {
+                          x5c: [ credCert: bytes, * (caCert: bytes) ]
+                      }
+    ```
+
+    The semantics of the above fields are as follows:
+
+    : x5c
+    :: |credCert| followed by its certificate chain, each encoded in X.509 format.
+
+    : credCert
+    :: The credential public key certificate used for attestation, encoded in X.509 format.
+
+: Signing procedure
+::
+    1. Let |authenticatorData| denote the authenticator data for the attestation, and let |clientDataHash| denote the [=hash of the serialized client data=].
+    2. Concatenate |authenticatorData| and |clientDataHash| to form |nonceToHash|.
+    3. Perform SHA-256 hash of |nonceToHash| to produce |nonce|.
+    4. Let Apple anonymous attestation CA generate an X.509 certificate for the [=credential public key=] and include the |nonce| as a certificate extension with OID ( 1.2.840.113635.100.8.2 ). |credCert| denotes this certificate. The |credCert| thus serves as a proof of the attestation, and the included |nonce| proves the attestation is live. In addition to that, the |nonce| also protects the integrity of the |authenticatorData| and [=client data=].
+    5. Set |x5c| to |credCert| followed by its certificate chain.
+
+: Verification procedure
+:: Given the verification procedure inputs |attStmt|, |authenticatorData| and |clientDataHash|, the verification procedure is as follows:
+    1. Verify that |attStmt| is valid CBOR conforming to the syntax defined above and perform CBOR decoding on it to extract the contained fields.
+    2. Concatenate |authenticatorData| and |clientDataHash| to form |nonceToHash|.
+    3. Perform SHA-256 hash of |nonceToHash| to produce |nonce|.
+    4. Verify |nonce| matches the value of the extension with OID ( 1.2.840.113635.100.8.2 ) in |credCert|.
+    5. Verify credential public key matches the Subject Public Key of |credCert|.
+    6. If successful, return implementation-specific values representing attestation type [=Anonymization CA=] and attestation trust path |x5c|.
+
 # <dfn>WebAuthn Extensions</dfn> # {#sctn-extensions}
 
 The mechanism for generating [=public key credentials=], as well as requesting and generating Authentication assertions, as
@@ -6385,6 +6435,10 @@ IANA "WebAuthn Attestation Statement Format Identifiers" registry [[!IANA-WebAut
 - WebAuthn Attestation Statement Format Identifier: fido-u2f
 - Description: Used with FIDO U2F authenticators
 - Specification Document: Section [[#sctn-fido-u2f-attestation]] of this specification
+    <br/><br/>
+- WebAuthn Attestation Statement Format Identifier: apple
+- Description: Used with Apple devices' [=platform authenticators=]
+- Specification Document: Section [[#sctn-apple-anonymous-attestation]] of this specification
 
 ## WebAuthn Extension Identifier Registrations ## {#sctn-extensions-reg}
 


index.bs
```diff
@@ -5129,6 +5130,56 @@ if the [=authenticator=] does not support [=attestation=].
...
+    1. Let |authenticatorData| denote the authenticator data for the attestation, and let |clientDataHash| denote the hash of the serialized client data.
+    2. Concatenate |authenticatorData| and |clientDataHash| to form |nonceToHash|.
+    3. Perform SHA-256 hash of |nonceToHash| to produce |nonce|.
+    4. Let Apple anonymous attestation CA generate a certificate for the credential public key and include the |nonce| as an extension with OID ( 1.2.840.113635.100.8.2 ). |credCert| denotes this certificate.
+    5. Set |x5c| to |credCert| followed by its certificate chain.
+
+: Verification procedure
+:: Given the verification procedure inputs |attStmt|, |authenticatorData| and |clientDataHash|, the verification procedure is as follows:
+    1. Verify that |attStmt| is valid CBOR conforming to the syntax defined above and perform CBOR decoding on it to extract the contained fields.
+    2. Verify |x5c| is a valid certificate chain starting from the |credCert| to the Apple WebAuthn root certificate.
```
MasterKale, line 5176: Would it be possible to include a link to where the Apple WebAuthn root certificate can be downloaded?

MasterKale: It seems this is the URL to the root cert: https://www.apple.com/certificateauthority/Apple_WebAuthn_Root_CA.pem (as linked from this page I stumbled upon: https://www.apple.com/certificateauthority/private/)

index.bs
```diff
@@ -5129,6 +5130,56 @@ if the [=authenticator=] does not support [=attestation=].
...
+    3. Concatenate |authenticatorData| and |clientDataHash| to form |nonceToHash|.
+    4. Perform SHA-256 hash of |nonceToHash| to produce |nonce|.
+    5. Verify |nonce| matches the value of the extension with OID ( 1.2.840.113635.100.8.2 ) in |credCert|.
```
MasterKale, line 5179: Is there an additional value that needs to be prepended to `nonceToHash` before hashing? In two attestations now from an iPhone X running iOS 14.0.1 there's been a constant `3024a1220420` in the "1.2.840.113635.100.8.2" extension that appears before the hash that matches what I get from hashing `nonceToHash`:

## Attestation 1
**nonce (from hashing authData + clientDataHash):**
```
e6ccfd879f384b14b19a5d633e1ed1d6567d1b40d171b689eddbe7bca5668696
```

**credCert extension value:**
```
3024a1220420e6ccfd879f384b14b19a5d633e1ed1d6567d1b40d171b689eddbe7bca5668696
```

## Attestation 2
**nonce (from hashing authData + clientDataHash):**
```
fbb0475110bdc0ad60245a8672568acab3e9860b1005c7d7aa745976cb7d8ead
```

**credCert extension value:**
```
3024a1220420fbb0475110bdc0ad60245a8672568acab3e9860b1005c7d7aa745976cb7d8ead
```

MasterKale: Please disregard, this is operator error. I'm not diving deep enough into the ASN.1 structure to get to the OCTET STRING 😅

index.bs
herrjemand: Maybe we could do the same for the safety and android key?

bdewater: The spec seems to have been intentionally written to be non-prescriptive about this. All other attestation formats flow like this:
- [Section 7.1 step 14](https://www.w3.org/TR/webauthn-1/#registering-a-new-credential) kicks things off:
  > Verify that _attStmt_ is a correct attestation statement, conveying a valid attestation signature, by using the attestation statement format _fmt_’s verification procedure given _attStmt_, _authData_ and the hash of the serialized client data computed in step 7.
- In [section 8](https://www.w3.org/TR/webauthn-1/#defined-attestation-formats), find the section of the attestation format in question. In general the final step of the happy path looks something like this:
  > If successful, return implementation-specific values representing attestation type <type> and attestation trust path _x5c_.
- Section 7.1 step 15 then continues with the above, but this makes finding the root of trust generic instead of format specific: 
  > If validation is successful, obtain a list of acceptable trust anchors (attestation root certificates or ECDAA-Issuer public keys) for that attestation type and attestation statement format _fmt_, from a trusted source or from policy. For example, the FIDO Metadata Service provides one way to obtain such information, using the `aaguid` in the `attestedCredentialData` in authData.

Initially I found this annoying as an implementer, the MDS was pretty lacking and actually getting the stuff I needed was hard. Getting TPM root certificates anyone? However since I've come to realize it was probably written this way to allow for:
- vendors choosing not to publish to the Fido MDS for whatever reason (although this would make my life a lot easier if they do)
- the [Fido conformance tool](https://fidoalliance.org/certification/functional-certification/conformance/) to provide their own root certificates for testing purposes, this can't ship with the real Apple cert for obvious reasons
- other less obvious use cases I can't think of right now

So this line should be changed to follow the rest of the spec (or less likely: the rest of the spec should be changed) IMO.

nicksteele:
I think this looks alright but agree that it would be beneficial to include [a link to Apple's root certificate](https://www.apple.com/certificateauthority/Apple_WebAuthn_Root_CA.pem). Will wait to see if there's any feedback on  today's call before moving forward.

equalsJeffH:
this really needs a merge-from-master -- which I've done in local clone -- i can push it up here if you wish.

nicksteele:
That would be great @equalsJeffH much appreciated

equalsJeffH:
ah, it's in @alanwaketan's fork.  It merged-from-master with no conflicts, so he ought to just do (in his local clone):
```
git remote update
git co master
git pull
git co attestation
git merge -s recursive -Xpatience master
git push origin attestation
```

index.bs
MasterKale: Hmm, in that case perhaps the WACG can take this on and maintain links to the various root certs in our "how to WebAuthn" website. It'd allow the spec to remain implementation-agnostic while giving devs one place to go to for links to the latest root certs 🤔

equalsJeffH: > Would it be possible to include a link to where the Apple WebAuthn root certificate can be downloaded?

I do not think that we'd place that sort of link directly in the spec.

> in that case perhaps the WACG can take this on and maintain links to the various root certs in our "how to WebAuthn" website.

Yes, doing something like that is much more plausible.

index.bs
alanwaketan: Good!

index.bs
```diff
@@ -5129,6 +5130,56 @@ if the [=authenticator=] does not support [=attestation=].
...
+
+    : x5c
+    :: |credCert| followed by its certificate chain, each encoded in X.509 format.
+
+    : credCert
+    :: The credential public key certificate used for attestation, encoded in X.509 format.
+
+: Signing procedure
+::
+    1. Let |authenticatorData| denote the authenticator data for the attestation, and let |clientDataHash| denote the hash of the serialized client data.
```
emlun, line 5167: This can be linked to the defined term. :)

```suggestion
    1. Let |authenticatorData| denote the authenticator data for the attestation, and let |clientDataHash| denote the [=hash of the serialized client data=].
```

index.bs
equalsJeffH: > Would it be possible to include a link to where the Apple WebAuthn root certificate can be downloaded?

I do not think that we'd place that sort of link directly in the spec.

> in that case perhaps the WACG can take this on and maintain links to the various root certs in our "how to WebAuthn" website.

Yes, doing something like that is much more plausible.

ve7jtb (approved):
lgtm

index.bs
```diff
@@ -5129,6 +5130,56 @@ if the [=authenticator=] does not support [=attestation=].
...
+
+: Verification procedure
+:: Given the verification procedure inputs |attStmt|, |authenticatorData| and |clientDataHash|, the verification procedure is as follows:
+    1. Verify that |attStmt| is valid CBOR conforming to the syntax defined above and perform CBOR decoding on it to extract the contained fields.
+    2. Verify |x5c| is a valid certificate chain starting from the |credCert| to the Apple WebAuthn root certificate.
+    3. Concatenate |authenticatorData| and |clientDataHash| to form |nonceToHash|.
+    4. Perform SHA-256 hash of |nonceToHash| to produce |nonce|.
+    5. Verify |nonce| matches the value of the extension with OID ( 1.2.840.113635.100.8.2 ) in |credCert|.
+    6. Verify credential public key matches the Subject Public Key of |credCert|.
+    7. If successful, return implementation-specific values representing attestation type Anonymous CA and attestation trust path |x5c|.
```
emlun, line 5181: As discussed on the WG call, we don't have "Anonymous CA" as a defined attestation type. Where to go on that (see also #1488) is not yet decided...

index.bs
```diff
@@ -5129,6 +5130,56 @@ if the [=authenticator=] does not support [=attestation=].
...
+    1. Let |authenticatorData| denote the authenticator data for the attestation, and let |clientDataHash| denote the hash of the serialized client data.
+    2. Concatenate |authenticatorData| and |clientDataHash| to form |nonceToHash|.
+    3. Perform SHA-256 hash of |nonceToHash| to produce |nonce|.
+    4. Let Apple anonymous attestation CA generate an X.509 certificate for the [=credential public key=] and include the |nonce| as a certificate extension with OID ( 1.2.840.113635.100.8.2 ). |credCert| denotes this certificate.
+    5. Set |x5c| to |credCert| followed by its certificate chain.
+
+: Verification procedure
+:: Given the verification procedure inputs |attStmt|, |authenticatorData| and |clientDataHash|, the verification procedure is as follows:
+    1. Verify that |attStmt| is valid CBOR conforming to the syntax defined above and perform CBOR decoding on it to extract the contained fields.
+    2. Verify |x5c| is a valid certificate chain starting from the |credCert| to the Apple WebAuthn root certificate.
```
sbweeden, line 5176: Similar feedback to comment from @bdewater. I don't think step 2 of the verification procedure which currently suggests performing the x5c validation belongs in this section. It is covered by the steps that follow more generically in [section 7.1 of the spec](https://w3c.github.io/webauthn/#sctn-registering-a-new-credential) (currently steps 20/21 in the editors draft).

index.bs
sbweeden: Agree that this comment needs reconciliation with the attestation type PR. Ultimately though it makes little difference what the attestation type is (so long as it is not "self" or "none") because all other current attestation types require x5c validation to follow.

index.bs
```diff
@@ -5129,6 +5130,56 @@ if the [=authenticator=] does not support [=attestation=].
...
+
+: Attestation types supported
+:: [=Anonymous CA=]
+
+: Syntax
+:: The syntax of an Apple attestation statement is defined as follows:
+
+    ```
+    $$attStmtType //= (
+                          fmt: "apple",
```
sbweeden, line 5148: Somewhat orthogonal (but related) to this PR and the new fmt value, the IANA registry should be updated. For references see step 19 of [section 7.1 Registering a new Credential](https://w3c.github.io/webauthn/#sctn-registering-a-new-credential), which in turn points to [IANA WebAuthn Registry](https://www.iana.org/assignments/webauthn/webauthn.xhtml). It may be best just to open a new issue for this to be done at present, then let that take place after this PR is merged.

sbweeden (changes requested):

index.bs
nicksteele: Created #1495 to track that

index.bs
alanwaketan: Fixed.

index.bs
```diff
@@ -5129,6 +5130,56 @@ if the [=authenticator=] does not support [=attestation=].
...
+
+    : credCert
+    :: The credential public key certificate used for attestation, encoded in X.509 format.
+
+: Signing procedure
+::
+    1. Let |authenticatorData| denote the authenticator data for the attestation, and let |clientDataHash| denote the hash of the serialized client data.
+    2. Concatenate |authenticatorData| and |clientDataHash| to form |nonceToHash|.
+    3. Perform SHA-256 hash of |nonceToHash| to produce |nonce|.
+    4. Let Apple anonymous attestation CA generate a certificate for the credential public key and include the |nonce| as an extension with OID ( 1.2.840.113635.100.8.2 ). |credCert| denotes this certificate.
```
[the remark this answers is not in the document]
emlun: As discussed on the call, I don't think this qualifies as self attestation, as the CA does generate a signature chain that can be linked to a trusted root cert.

index.bs
sbweeden: Will resolve this review comment with the understanding that #1495 has been opened to address separately.

alanwaketan:
> In my testing with iPhone on iOS 14, there is an additional "alg" parameter provided within the attStmt body. Is that missing from this specification?

Oops, I believe it is a bug in the software. The current attestation process should not utilize that parameter at all. The X5C itself should have suggested the algorithm to validate the chain. There was an old/not shipping version that utilize this field. I will remove it. Thanks for catching this bug!

alanwaketan:
> > In my testing with iPhone on iOS 14, there is an additional "alg" parameter provided within the attStmt body. Is that missing from this specification?
> 
> Oops, I believe it is a bug in the software. The current attestation process should not utilize that parameter at all. The X5C itself should have suggested the algorithm to validate the chain. There was an old/not shipping version that utilize this field. I will remove it. Thanks for catching this bug!

Here is the WebKit issue for your reference: https://bugs.webkit.org/show_bug.cgi?id=217720.

sbweeden (approved):

nicksteele (approved):
Waiting for resolution of #1474 but this should be ready to go. As per the weekly call, the `alg` parameter in `AttStmt` will be removed from Apple's response payload.

equalsJeffH:
pr #1491 should also update the https://w3c.github.io/webauthn/#sctn-att-fmt-reg section appropriately.

index.bs
```diff
@@ -5137,6 +5138,55 @@ if the [=authenticator=] does not support [=attestation=].
...
+    5. Set |x5c| to |credCert| followed by its certificate chain.
+
+: Verification procedure
+:: Given the verification procedure inputs |attStmt|, |authenticatorData| and |clientDataHash|, the verification procedure is as follows:
+    1. Verify that |attStmt| is valid CBOR conforming to the syntax defined above and perform CBOR decoding on it to extract the contained fields.
+    2. Concatenate |authenticatorData| and |clientDataHash| to form |nonceToHash|.
+    3. Perform SHA-256 hash of |nonceToHash| to produce |nonce|.
+    4. Verify |nonce| matches the value of the extension with OID ( 1.2.840.113635.100.8.2 ) in |credCert|.
+    5. Verify credential public key matches the Subject Public Key of |credCert|.
+    6. If successful, return implementation-specific values representing attestation type Anonymization CA and attestation trust path |x5c|.
```
sbweeden, line 5188: Do you want to make this reference to Anonymization CA like:
`[=Anonymization CA=]`

alanwaketan: Fixed.

sbweeden:
I believe this PR is ready for further/full review now. I have been working with @alanwaketan this week on same. One thing to note for reviewers is that this PR contains references to `[=Anonymization CA=]`. The definition for that will be updated once #1474 lands however because there is an existing `<dfn>Anonymization CA</dfn>` in the master branch this has the fortunate side effect that it won't really matter in which order #1491 and #1474 land, though it probably makes sense to bring them both in at the same time.

index.bs
```diff
@@ -5137,6 +5138,55 @@ if the [=authenticator=] does not support [=attestation=].
...
+
+    : credCert
+    :: The credential public key certificate used for attestation, encoded in X.509 format.
+
+: Signing procedure
+::
+    1. Let |authenticatorData| denote the authenticator data for the attestation, and let |clientDataHash| denote the [=hash of the serialized client data=].
+    2. Concatenate |authenticatorData| and |clientDataHash| to form |nonceToHash|.
+    3. Perform SHA-256 hash of |nonceToHash| to produce |nonce|.
+    4. Let Apple anonymous attestation CA generate an X.509 certificate for the [=credential public key=] and include the |nonce| as a certificate extension with OID ( 1.2.840.113635.100.8.2 ). |credCert| denotes this certificate. The |credCert| thus serves as a proof of the attestation, and the included |nonce| proves the attestation is live. In addition to that, the |nonce| also protects the integrity of the |authenticatorData| and [=client data=].
```
emlun, line 5178: Is there a name for this OID? (Like "OID 1.3.6.1.4.1.45724.1.1.4 (`id-fido-gen-ce-aaguid`)" in §8.2.1. Packed Attestation Statement Certificate Requirements)

index.bs
```diff
@@ -5137,6 +5138,55 @@ if the [=authenticator=] does not support [=attestation=].
...
+    5. Set |x5c| to |credCert| followed by its certificate chain.
+
+: Verification procedure
+:: Given the verification procedure inputs |attStmt|, |authenticatorData| and |clientDataHash|, the verification procedure is as follows:
+    1. Verify that |attStmt| is valid CBOR conforming to the syntax defined above and perform CBOR decoding on it to extract the contained fields.
+    2. Concatenate |authenticatorData| and |clientDataHash| to form |nonceToHash|.
+    3. Perform SHA-256 hash of |nonceToHash| to produce |nonce|.
+    4. Verify |nonce| matches the value of the extension with OID ( 1.2.840.113635.100.8.2 ) in |credCert|.
```
emlun, line 5186: "Matches" could imply some looser comparison like a pattern or prefix match, so I would say "equals" here for clarity (assuming that is what is meant).
```suggestion
    4. Verify that |nonce| equals the value of the extension with OID ( 1.2.840.113635.100.8.2 ) in |credCert|.
```

index.bs
```diff
@@ -5137,6 +5138,55 @@ if the [=authenticator=] does not support [=attestation=].
...
+    5. Verify credential public key matches the Subject Public Key of |credCert|.
```
emlun, line 5187: Same here, and we can also link the "credential public key" term.
```suggestion
    5. Verify that the [=credential public key=] equals the Subject Public Key of |credCert|.
```

emlun (commented):
Looks good to me, just a couple suggestions for clarity...

index.bs
emlun: From 2020-10-21: no such name is defined at this time.

alanwaketan:
Can I merge this now?

akshayku (approved):

equalsJeffH (approved):
LGTM, thx @alanwaketan !

emlun (approved):

PR #1499, opened 2020-10-15
define "self-signed basic attestation type" "SSBasic"

### LINKED ISSUE #1498: ensure basic attestation type allows for self-signed attestation certs
some authenticators in the wild return, on registration, an attestation certificate that is signed by the private key associated with the public key in the cert. Thus, the attestation certificate is _essentially_ the "(CA) root certificate".  Verifying such a certificate can be accomplished by, e.g.: 
```
$ cd /some-directory    # where the cert files reside...
$ openssl verify -CAfile ./self-signed-cert.pem -CApath /some-directory ./self-signed-cert.pem
./self-signed-cert.pem: OK
```
...though the value of verifying the signature is debatable.  If the RP has obtained the putative "root certificate" via out-of-band means (e.g., via the FIDO Alliance Metadata Service), they can also "verfiy" the returned attestation cert against the putative "root certificate" by a simple byte-by-byte comparison (once the certificates are both represented in the same format, e.g., DER or PEM).

This key pair is used only for attestation (unlike the present "self-attestation" attstn type, where the user's per-RP key pair is used to construct a one-off per-RP attestation cert) and thus the same attestation cert is returned to all RPs. Thus we ought to define a new attestation type for this.

Since this is fairly similar to Basic Attestation, I suggest:  "Self-signed Basic Attestation"

equalsJeffH:
some authenticators in the wild return, on registration, an attestation certificate that is signed by the private key associated with the public key in the cert. Thus, the attestation certificate is essentially the "(CA) root certificate".

this change defines this sort of attestation as "self-signed basic attestation", "SSBasic" for short.

fixes #1498

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/pull/1499.html" title="Last updated on Oct 15, 2020, 3:11 PM UTC (eecb55e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/1499/3d38cab...eecb55e.html" title="Last updated on Oct 15, 2020, 3:11 PM UTC (eecb55e)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -4247,6 +4247,9 @@ calling {{CredentialsContainer/create()|navigator.credentials.create()}} they se
 
     [=Basic attestation=] is sometimes also called <dfn>batch attestation</dfn>.
 
+: <dfn>Self-Signed Basic Attestation</dfn> (<dfn>SSBasic</dfn>)
+:: This is attestation where, similar to the [=Basic=] attestation type, the authenticator's [=attestation key pair=] is specific to an authenticator model or a series of models. Though, [=SSBasic=]'s difference is that the attestation certificate is signed by the private key associated with the public key in the certificate, thus the attestation certificate is essentially the "(CA) root certificate". Verifying such an attestation certificate can include verifying the certificate's signature, although this may be of little value to an [=[RP]=]. If the RP has obtained the putative "root certificate" via out-of-band means (e.g., via the FIDO Alliance Metadata Service), they can also verfiy the returned attestation cert against the putative "root certificate" by a simple byte-by-byte comparison (once the certificates are both represented in the same format, e.g., DER or PEM). Note also that [[=RFC5280=]] defines specific requirements for various fields of a "self-signed" certificate. This attestation type does not require a [=SSBasic=] attestation certificate to adhere to those latter requirements.
+
 : <dfn>Self Attestation</dfn> (<dfn>Self</dfn>)
 :: In the case of [=self attestation=], also known as surrogate basic attestation [[UAFProtocol]], the Authenticator does not have
     any specific [=attestation key pair=]. Instead it uses the [=credential private key=] to create the [=attestation signature=].
@@ -4651,7 +4654,7 @@ implementable by [=authenticators=] with limited resources (e.g., secure element
 :: packed
 
 : Attestation types supported
-:: [=Basic=], [=Self=], [=AttCA=]
+:: [=Basic=], [=SSBasic=], [=Self=], [=AttCA=]
 
 : Syntax
 :: The syntax of a Packed Attestation statement is defined by the following CDDL:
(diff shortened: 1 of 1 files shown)


herrjemand:
Need to have a proper impact analysis and collect feedback from certified servers before this change is approved

ve7jtb (commented):
I think this is reasonable.  
My only question is if we should have guidance for creating such a self signed certificate going forward.   There are probably some things like making sure the certificate chains to itself, that would be useful for backwards compatibility.

ve7jtb:
> Need to have a proper impact analysis and collect feedback from certified servers before this change is approved

I am not sure that WebAuthn needs to do that.  In principal we can support new attestation formats like apple attestation, without first checking what existing Fido servers do.

As long as this behavior fails in a safe way in servers, eg the worst that happens in existing servers is that the the attestations are considers invalid, I don't see a problem.   New servers will support it like the apple attestation.

RaeHayward:
> > Need to have a proper impact analysis and collect feedback from certified servers before this change is approved
> 
> I am not sure that WebAuthn needs to do that. In principal we can support new attestation formats like apple attestation, without first checking what existing Fido servers do.
> 
> As long as this behavior fails in a safe way in servers, eg the worst that happens in existing servers is that the the attestations are considers invalid, I don't see a problem. New servers will support it like the apple attestation.

Please note that FIDO Alliance is in no way trying to dictate the work of this group and the comment is not our official stance on this topic. 
From my perspective I am working to coordinate on our end, as it relates to certification, for a more cohesive ecosystem and understand the changes that we would potentially need to incorporate in order to support that end goal.

emlun (commented):
I'm not sure how this is qualitatively different from any other Basic attestation. If the RP has the self-signed "root" certificate in its trust store, then it can already just verify the signature as usual, no? Presumably the trust store is indexed by X.509 subject name and the self-signed certificate has identical subject and issuer names, so the RP would just look up a second copy of the same certificate and see a valid signature by the trusted copy. Conversely if the self-signed cert is not in the trust store, then the RP would fail to look it up and thus not trust the signature. So why does this special case need its own category?

equalsJeffH:
> I'm not sure how this is qualitatively different from any other Basic attestation.  ...  So why does this special case need its own category?

Yes, it may not merit its own attestation type.  Perhaps clarifying that an authenticator's "attestation root certificate" (as made available by the authnr's manufacturer) may be the same as the authnr-returned attestation cert, and can be "verified" by either a byte-by-byte comparison or verifying the signature, is sufficient.

As it turns out, the FIDO Metadata Statement spec [allows for such self-signed attestation certificates](https://fidoalliance.org/specs/fido-u2f-v1.2-ps-20170411/fido-metadata-statement-v1.2-ps-20170411.html#h_note_21:~:text=It%20might%20be%20the%20actual%20certificate%20presented%20by%20the%20authenticator).

jcjones (changes requested):
We shouldn't need a new attestation type here. Self-signed should be usable with Basic as-is.

equalsJeffH:
am closing this given https://github.com/w3c/webauthn/pull/1499#pullrequestreview-512862975, https://github.com/w3c/webauthn/pull/1499#pullrequestreview-514135067, and https://github.com/w3c/webauthn/pull/1499#issuecomment-713801566.

PR #1509, opened 2020-10-28
ensure basic attestation type allows for self-signed attestation certs

### LINKED ISSUE #1498: ensure basic attestation type allows for self-signed attestation certs

equalsJeffH:
This PR simply cleans up the language in [8.2. Packed Attestation Statement Format](https://w3c.github.io/webauthn/#sctn-packed-attestation) such that self-signed attestation certs are implicitly allowed.

fixes #1498

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/pull/1509.html" title="Last updated on Nov 9, 2020, 6:32 PM UTC (ff28800)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/1509/da36403...ff28800.html" title="Last updated on Nov 9, 2020, 6:32 PM UTC (ff28800)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -4249,10 +4249,10 @@ under [=[RP]=] policy.
 
 : <dfn>Basic Attestation</dfn> (<dfn>Basic</dfn>)
 :: In the case of basic attestation [[UAFProtocol]], the authenticator's [=attestation key pair=] is specific to an
-    authenticator model.  Thus, authenticators of the same model often share the same [=attestation key pair=]. See
+    authenticator "model", i.e., a "batch" of authenticators.  Thus, authenticators of the same, or similar, model often share the same [=attestation key pair=]. See
     [[#sctn-attestation-privacy]] for further information.
 
-    [=Basic attestation=] is sometimes also called <dfn>batch attestation</dfn>.
+    [=Basic attestation=] is also referred to as <dfn>batch attestation</dfn>.
 
 : <dfn>Self Attestation</dfn> (<dfn>Self</dfn>)
 :: In the case of [=self attestation=], also known as surrogate basic attestation [[UAFProtocol]], the Authenticator does not have
@@ -4697,7 +4697,7 @@ implementable by [=authenticators=] with limited resources (e.g., secure element
     :: A byte string containing the [=attestation signature=].
 
     : x5c
-    :: The elements of this array contain |attestnCert| and its certificate chain, each encoded in X.509 format. The attestation
+    :: The elements of this array contain |attestnCert| and its certificate chain (if any), each encoded in X.509 format. The attestation
         certificate |attestnCert| MUST be the first element in the array.
 
     : attestnCert
@@ -4711,7 +4711,7 @@ implementable by [=authenticators=] with limited resources (e.g., secure element
 
         1. If [=Basic=] or [=AttCA=] [=attestation=] is in use, the authenticator produces the |sig| by concatenating |authenticatorData| and
             |clientDataHash|, and signing the result using an [=attestation private key=] selected through an authenticator-specific
-            mechanism. It sets |x5c| to the certificate chain of the [=attestation public key=] and |alg| to the algorithm of the
+            mechanism. It sets |x5c|'s first element to |attestnCert|'s value, and includes the related certificate chain (if any) in further |x5c| elelment. It sets |alg| to the algorithm of the
             attestation private key.
 
         1. If [=self attestation=] is in use, the authenticator produces |sig| by concatenating |authenticatorData| and |clientDataHash|,
--- a/index.bs
+++ b/index.bs
@@ -4459,6 +4459,8 @@ In order to perform a [=registration ceremony=], the [=[RP]=] MUST proceed as fo
         - Otherwise, use the X.509 certificates returned as the [=attestation trust path=] from the [=verification procedure=]
             to verify that the attestation public key correctly chains up to an acceptable root certificate.
 
+            Note: The attestation certificate and the root certificate obtained above in step 20 can be the same certificate.
+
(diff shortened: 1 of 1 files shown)


sbweeden (approved):
LGTM

ve7jtb (approved):
LGTM

index.bs
```diff
@@ -4711,7 +4711,7 @@ implementable by [=authenticators=] with limited resources (e.g., secure element
 
         1. If [=Basic=] or [=AttCA=] [=attestation=] is in use, the authenticator produces the |sig| by concatenating |authenticatorData| and
             |clientDataHash|, and signing the result using an [=attestation private key=] selected through an authenticator-specific
-            mechanism. It sets |x5c| to the certificate chain of the [=attestation public key=] and |alg| to the algorithm of the
+            mechanism. It sets |x5c|'s first element to |attestnCert|'s value, and includes the related certificate chain (if any) in further |x5c| elelment. It sets |alg| to the algorithm of the
```
akshayku, line 4714:
```suggestion
            mechanism. It sets |x5c|'s first element to |attestnCert|'s value, and includes the related certificate chain (if any) in further |x5c| element. It sets |alg| to the algorithm of the
```

akshayku (approved):

akshayku (approved):

ve7jtb (approved):
I think that note clarifies the self signed batch certificate case.

index.bs
```diff
@@ -4459,6 +4459,8 @@ In order to perform a [=registration ceremony=], the [=[RP]=] MUST proceed as fo
         - Otherwise, use the X.509 certificates returned as the [=attestation trust path=] from the [=verification procedure=]
             to verify that the attestation public key correctly chains up to an acceptable root certificate.
 
+            Note: The attestation certificate and the root certificate obtained above in step 20 can be the same certificate.
```
jcjones, line 4462: I think maybe this note is not so helpful at this point. We should just modify line 4460 to change how we discuss validating the certificate. See other comment.

```suggestion
```

index.bs
```diff
@@ -4459,6 +4459,8 @@ In order to perform a [=registration ceremony=], the [=[RP]=] MUST proceed as fo
         - Otherwise, use the X.509 certificates returned as the [=attestation trust path=] from the [=verification procedure=]
             to verify that the attestation public key correctly chains up to an acceptable root certificate.
```
jcjones, line 4460:
```suggestion
            to verify that the attestation public key either correctly chains up to an acceptable root certificate, or is itself an acceptable certificate.
```

jcjones (changes requested):
I'm not unbiased here, but my proposed changes are how I would expect this situation to be discussed for the Web PKI. In either case though, I think it belongs in the algorithm text. We could keep the Note too, but I want to have the algorithm text updated as well.

index.bs
equalsJeffH: @jcjones - While I support your suggestion for L4460 (thx!), I am unsure whether it is sufficiently clear to folks who are not as steeped in PKI-ness as you are. 

Perhaps we further update L4460 like so:
> ...verify that the attestation certificate either correctly chains up to an acceptable root certificate, or is itself an acceptable certificate (i.e., it and the root certificate obtained in step 20 may be the same).

..?

jcjones: OK! Works for me.

equalsJeffH:
@jcjones' feedback is applied & embellished per discussion above, please re-review :)

akshayku (approved):

index.bs
```diff
@@ -4711,7 +4712,7 @@ implementable by [=authenticators=] with limited resources (e.g., secure element
 
         1. If [=Basic=] or [=AttCA=] [=attestation=] is in use, the authenticator produces the |sig| by concatenating |authenticatorData| and
             |clientDataHash|, and signing the result using an [=attestation private key=] selected through an authenticator-specific
-            mechanism. It sets |x5c| to the certificate chain of the [=attestation public key=] and |alg| to the algorithm of the
+            mechanism. It sets |x5c|'s first element to |attestnCert|'s value, and includes the related certificate chain (if any) in further |x5c| element. It sets |alg| to the algorithm of the
```
emlun, line 4715: How about this?

```suggestion
            mechanism. It sets |x5c| to |attestnCert| followed by the related certificate chain (if any). It sets |alg| to the algorithm of the
```

ve7jtb (approved):
Yes JC's changes are better.
Looks good now.

jcjones (approved):
LGTM thanks!

emlun (approved):
LGTM!

sbweeden (approved):
This is fine by me.

PR #1537, opened 2020-12-09
Ban empty user.id aka user handle

### LINKED ISSUE #1536: ban empty user ID / user handle
The spec already notes: "...a user handle having an empty value is known to be problematic in practice..."

However, in testing, its been discovered that:
Returning a `user` in GetAssertion with an empty ID: Windows 10 fails.
Same response, but omitting `user`: Windows 10 works.
([Japanese blog post about this workaround.](https://qiita.com/gebo/items/612373fc58395b2927e7))
"OpenSK users reported Windows 10 interoperability issues."

@akshayku has noted:  I would prefer RP always sending a non-zero userID or browser erroring out when it receives a zero length userID or browser/platform not setting a zero length userID in makeCredential even if RP sets it empty as clearly RP does not care about userID in this scenario.

Thus, we suggest the webauthn spec states that user.id MUST not be empty, and if an RP wishes a constant value for whatever reason, they pick something innocuous such as a single space char.

equalsJeffH:
fixes #1536 (see the latter for details)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/pull/1537.html" title="Last updated on Dec 9, 2020, 5:24 PM UTC (bd668d1)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/1537/68aa633...bd668d1.html" title="Last updated on Dec 9, 2020, 5:24 PM UTC (bd668d1)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2624,9 +2624,9 @@ credential.
         {{PublicKeyCredentialUserEntity/displayName}} nor {{PublicKeyCredentialEntity/name}} members. See Section 6.1 of [[!RFC8266]].
 
         The [=user handle=] MUST NOT contain [PII] about the user, such as a username or e-mail address;
-        see [[#sctn-user-handle-privacy]] for details.
+        see [[#sctn-user-handle-privacy]] for details. The [=user handle=] MUST NOT be empty, though it MAY be null.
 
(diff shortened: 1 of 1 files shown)


akshayku (approved):

agl (approved):

ve7jtb (approved):
LGTM

PR #1539, opened 2020-12-09
Add algorithm step to enforce specified limits on user.id.

### LINKED ISSUE #1302: Length restriction on user handles has no processing model associated with it
https://w3c.github.io/webauthn/#dom-publickeycredentialuserentity-id says:

> A user handle is an opaque byte sequence with a maximum size of 64 bytes

but nothing says where or whether that maximum size is enforced.  Based on code inspection, Firefox checks for this in code that gets called from `CredentialsContainer.create` and throws an exception (a DOMException with name "TypeError") if the user handle is longer than 64 bytes.  Also based on code inspection it looks like Chrome throws a JS TypeError in this situation (with a comment about <https://www.w3.org/TR/webauthn/#user-handle>).  

Note that the lack of clear processing model leads to lack of interop here: the two browsers are throwing totally different kinds of exceptions...

@jcjones

agl:
Fixes #1302.

***
<a href="https://pr-preview.s3.amazonaws.com/agl/webauthn/pull/1539.html" title="Last updated on Dec 10, 2020, 3:48 PM UTC (f887013)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/1539/0bc3737...agl:f887013.html" title="Last updated on Dec 10, 2020, 3:48 PM UTC (f887013)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1511,6 +1511,8 @@ When this method is invoked, the user agent MUST execute the following algorithm
 
         Note: The user agent should take cognitive guidelines into considerations regarding timeout for users with special needs.
 
+1. If the length of <code>|options|.{{PublicKeyCredentialCreationOptions/user}}.{{PublicKeyCredentialUserEntity/id}}</code> is not between 1 and 64 bytes (inclusive) then return a {{TypeError}}.
+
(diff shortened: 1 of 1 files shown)


equalsJeffH (approved):
On L1731:  s/ Step 7 / Step 8 /

otherwise, LGTM

ve7jtb (approved):
LGTM

akshayku (approved):

Issue #1510, opened 2020-11-02
User verification policy leads to ambiguous usage situations.


Firstyear:
Hi there,

I have noticed a potential issue in the construction of UserVerificationPolicy with regard to PublicKeyCredentialCreationOptions and PublicKeyCredentialRequestOptions.

During the generation of a credential, this uses the UserVerificationPolicy (UVP) given to the AuthenticatorSelectionCriteria and embedded into PublicKeyCredentialCreationOptions. This creates a relationship between the UVP and the registered credential.

However, during an authentication, PublicKeyCredentialRequestOptions embeds the UVP in the top level dictionary, and the set of allowCredentials contains the set of credential ids that can proceed. This creates a relationship where the UVP is related to the authentication operation.

This can lead to a situation where a credential that was registered with verification, can then be used as an unverified credential in future operations. Additionally, one could register a credential with UVP::Required, and another with UVP::Discouraged, but in a future authentication this creates an ambiguity - do you require UVP::Required, and have the credential with discouraged now require verification? Or do we do the authentication with UVP::Discouraged, making the verified credential "un-verified". In the situation you use UVP::Preferred, this still creates ambiguity such as with CTAP1/2 where the credential can now become verified in some browsers, and remains unverified in others.

A possible solution is to treat the UserVerificationPolicy as "per user" and applies blanket to all tokens associated to the user, but this also reduces flexibility. Consider that an U2F token which was UVP::Discouraged at registration may be used with a password, but the user may also have a verified token that use used a sole source of MFA. The authentication flow should be that the user conducts a webauthn authentication first, and then based on that result and which credential was used we choose to request the password or not. But if we use UVP::Preferred, we now are verifying the U2F token adding an unnecessary step to the workflow. 

I'm sure that it may be difficult to achieve now, but I would suggest that UVP should be moved to PublicKeyCredentialDescriptor so that in the set of allowedCredentials the browser/client, is able to determine based on which credential we are selecting the method to proceed for verification of that individual token. 

Alternately, it would be potentially beneficial to consider this situation and how a diverse array of webauthn tokens can be associated to a single account, and how that interacts with the UVP and to provide recommendations accordingly. As webauthn exists in conjunction with other authentication mechanisms it should be considered how this is assembled into complete MFA systems such as a UVP::Required Token for a user, and a UVP::Discouraged Token + Password that can co-exist and have policies able to express this.

Thanks, 

[...]

PR #1547, opened 2021-01-20
Draft changes for allowCredentials to indicate per-credential verification

Firstyear:
This is a rough draft of changes to the webauthn specification to allow the service to indicate to the browser per-authenticator verification policies.

The current definition of userVerification is focussed around the concept of per-ceremony verification requirements. However, this is not compatible with identity management systems which tend to have per-authenticator requirements, or even per-identity requirements. 

An example of this is today if an IDM service existed that allowed a user to have multiple credentials, and those credentials could be either verified (MFA devices, as per https://pages.nist.gov/800-63-3/sp800-63b.html 5.1.9) or the user may have a password and an unverified single factor authenticator (SFA https://pages.nist.gov/800-63-3/sp800-63b.html 5.1.7). In the webauthn specification this is not possible to express in `PublicKeyCredentialRequestOptions.userVerification`. If this is set to "required" we exclude the SFA device. If this is set to "discouraged" we exclude the MFA device. If set to preferred, we can cause the SFA device to upgrade to "verified" in some browsers (IE CTAP2 vs CTAP1). We also then run the risk that the user may tamper with the `PublicKeyCredentialRequestOptions.userVerification` to set it to discouraged, causing our MFA device to downgrade to an SFA device, as this field is not part of the hashed collected client data. Since we issued the request as 'preferred' we must accept unverified credentials, which may be an authentication bypass or downgrade attack depending on the server implementation.

This means to satisfy this scenario would would require a "lead in" page to request from the user which of their devices/authentication methods they wish to use, so that we can issue the correct `PublicKeyCredentialRequestOptions.userVerification`, but this may be an information disclosure which may be considered a risk to some organisations.

As webauthn intends to be a device authentication for users, and those users must exist in an identity management system with requirements and policy around authentication, then webauthn should have a method to represent this in it's exchange.

This PR extends the PublicKeyCredentialDescriptor field with an optional boolean `verificationRequired` which defines per-authenticator verification policy that the server can issue to support mixed SFA/MFA devices in a single webauthn operation. The browser is then able to offer the correct verification steps for each device as required, and the server can verify this with it's per-authenticator knowledge of the requirements. If the browser does not understand this field (older browser etc) then it will continue to use the userVerification flag from PublicKeyCredentialRequestOptions.

This also changes the default userVerification from preferred to discouraged. This is because in authentication, since preferred is not part of the signed clientdatajson and may be altered by the client, and the server *must* accept unverified credentials by design, then preferred is equivalent to discouraged. It would then be implementation specific on how to act in this scenario, but it is very likely this represents a verification bypass. Preferred still has value in credential registration however, allowing the client to indicate if the device is a SFA or MFA device, and if tampered with at this stage to downgrade to discouraged, the service is able to make other appropriate policy decisions to assume the device is not verified IE then continuing to enforce a password as another factor. 

fixes: https://github.com/w3c/webauthn/issues/1510

see also: https://fy.blackhats.net.au/blog/html/2020/11/21/webauthn_userverificationpolicy_curiosities.html

***
<a href="https://pr-preview.s3.amazonaws.com/Firstyear/webauthn/pull/1547.html" title="Last updated on Jan 20, 2021, 1:33 AM UTC (49576f9)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/1547/a1f9066...Firstyear:49576f9.html" title="Last updated on Jan 20, 2021, 1:33 AM UTC (49576f9)">Diff</a>


equalsJeffH:
on 2021-01-27 call:
@sbweeden suggests closing this PR, he met with the submitter yesterday.
see issue #1510

sbweeden:
After discussion with author and working group, this PR to be closed and some suggestions for non-normative changes that would assist RP developers to be added to issue #1510 for consideration in a future PR to L3.

Issue #692, opened 2017-11-17
#createCredential alg: "if options.authenticatorSelection is present..." may be incorrect


equalsJeffH:
The current step 18 of [#createCredential alg](https://w3c.github.io/webauthn/#createCredential) says:
> If `options.authenticatorSelection` is present, for each authenticator...

However, if `options.authenticatorSelection` is not present, we will not invoke any authenticators, which will likely make the user sad. 

Given that we need to pass `options.authenticatorSelection.requireResidentKey` and `options.authenticatorSelection.requireUserVerification` to `authenticatorMakeCredential()`, perhaps we should make `authenticatorSelection` a `required` member of [`MakePublicKeyCredentialOptions`](MakePublicKeyCredentialOptions) ?   (unless I'm misunderstanding things...?)

emlun:
See the first commit in #687

emlun:
I made a separate PR with just that first commit: #696

emlun:
I support making `options.authenticatorSelection` required and default to `{}`. Then we can also get rid of the "If `authenticatorSelection` is present" branch, and the descriptions of `requireResidentKey` and `requireUserVerification` in [authenticatorMakeCredential][make] won't describe possibly undefined behaviour. I suppose it would technically be a breaking change to how `credentials.create()` is implemented, but it wouldn't at all affect observable behaviour or the API on either end.

[make]: https://w3c.github.io/webauthn/#op-make-cred

emlun:
The bikeshed issue appeared in bikeshed commit [aa80a167c54650c1c8f2382945f3894ee5b80bb4](https://github.com/tabatkins/bikeshed/commit/aa80a167c54650c1c8f2382945f3894ee5b80bb4).

equalsJeffH:
thx @emlun, see https://github.com/tabatkins/bikeshed/issues/1139

emlun:
Fixed by #696

Issue #155, opened 2016-07-28
add references to IANA WebAuthn Attestation Formats and Extension Identifiers Registries


equalsJeffH:
We have an in-progress Internet-Draft (I-D) that creates two distinct, but related, IANA registries, one for "WebAuthn Attestation Formats" and one for "WebAuthn Extension Identifiers".  

the initial draft of this I-D is here..
> draft-hodges-webauthn-registries-00
> https://lists.w3.org/Archives/Public/public-webauthn/2016Jun/0097.html

..and a subsequent message clarifying its purpose and applicability to the webauthn spec is here..

> [webauthn] Attestation type identifiers lack formal definition and matching rules
> https://lists.w3.org/Archives/Public/public-webauthn/2016Jul/0070.html

..as well as this message clarifying the relationship between registries and specifications..

> wrt registries (was: Spec status)
> https://lists.w3.org/Archives/Public/public-webauthn/2016Jul/0308.html

We can and should at this time add appropriate placeholder references to the WebAuthn spec draft referencing these to-be-created  registries, and noting the existence of draft-hodges-webauthn-registries-00 -- this will hopefully mitigate misunderstandings among us regarding whether we are creating registries or not. 

A related but different issue is whether or not to have the actual specification of the presently-known attestation formats within the webauthn spec or in a separate specification. See #108

equalsJeffH:
updating issue title to reflect the change terminology from 'attestation type' to 'attestation format'

equalsJeffH:
fixed by commit: https://github.com/w3c/webauthn/commit/b695628aeca3834a4dfa1348a72fce2190503ccc which was merged to master

PR #1576, opened 2021-02-18
conditional UI via mediation

equalsJeffH:
conditional UI flow rework
    
see also https://github.com/w3c/webappsec-credential-management/pull/155
    
and [Explainer: WebAuthn Conditional/Hinted UI](https://docs.google.com/document/d/11hWpUPAnblPtkn1f7AIQW0ujoiu_BAKzlMVhZKQPiW8/)


index.bs
```diff
@@ -1911,14 +1911,16 @@ to indicate what [=credential sources=] are acceptable to it. The [=client platf
...
 should be available without [=user mediation=] (roughly, this specification's [=authorization gesture=]), and if it does not find
-exactly one of those, it then calls <code>PublicKeyCredential.{{PublicKeyCredential/[[DiscoverFromExternalSource]]()}}</code> to have
-the user select a [=credential source=].
+exactly one of those, it then calls <code>PublicKeyCredential.{{PublicKeyCredential/[[DiscoverFromExternalSource]]()}}</code> to have the user select a [=credential source=].
+
+In any case, the user agent SHOULD show some UI to the user to guide them in selecting and
+authorizing an authenticator with which to complete the operation. [=[RPS]=] can provide a hint that a prominent, modal UI <em>should not</em> be used for this process by setting <code>|options|.{{CredentialRequestOptions/mediation}}</code> to "conditional". This is known as a <dfn>Conditional UI Flow</dfn>. [=[RP]=] script SHOULD first check that <code>navigator.credentials.{{CredentialsContainer/conditionalMediationSupported}}</code> is [TRUE] in order to avoid the possiblity of causing a user-visible error to be returned if the user agent does not support [=conditional UI flow=].
```
sbweeden, line 1923: Where is there a definition of navigator.credentials.CredentialsContainer/conditionalMediationSupported ? 

I have not heard of this before, or the use cases around the `mediation` property of CredentialRequestOptions, and would like to understand more how it is intended to apply to a practical WebAuthn use case.

equalsJeffH: see https://github.com/w3c/webappsec-credential-management/pull/155 and [Explainer: WebAuthn Conditional/Hinted UI](https://docs.google.com/document/d/11hWpUPAnblPtkn1f7AIQW0ujoiu_BAKzlMVhZKQPiW8/)

MasterKale, line 1923: I just caught myself up on this proposal. On the whole I think this is a great idea. However I suggest we consider the following alternative to the naming of the boolean `conditionalMediationSupported` :

```
navigator.credentials.isConditionalMediationAvailable()
```

I suggest this because we already have the existing helper method `isUserVerifyingPlatformAuthenticatorAvailable()` that establishes an `is...Available()` naming convention for these kinds of helper methods. I'm also suggesting this becomes a **method** instead of a **boolean** because I believe it will eliminate confusion over whether a particular feature check in the WebAuthn API is a simple boolean, or something that needs to be invoked. If they're all invoked methods that return booleans then there's less mental overhead for RP developers working with this API.

emlun: Thanks @MasterKale, looks like this feedback belongs also in https://github.com/w3c/webappsec-credential-management/pull/155 which is what would add that attribute.

MasterKale: Thank you for the nudge, I went ahead and cross-posted. Honestly I feel like this suggestion is less compelling over there without something like `isUserVerifyingPlatformAuthenticatorAvailable()` in the credential management API (which there isn't)...

If conditional mediation goes through with `conditionalMediationSupported`, is there any value in trying to update `PublicKeyCredential` with a new static method `isConditionalMediationAvailable()` that returns the value of `navigator.credentials.conditionalMediationSupported`? 😛

lgarron: A drive-by thought: Is it certain that this kind of support will always be a single boolean? The function leaves open the possibility of arguments in the future, in case the level of support depends on certain properties (e.g. of the registration parameters) in a non-trivial way.

equalsJeffH: ISTM @MasterKale makes good suggestions wrt naming and it being a method rather than a simple boolean.

equalsJeffH (commented):
Hi @nsatragno, here's some high-level comments on the latest commit 2c09264...
1. We do not need to define `isConditionalMediationAvailable()` in this spec (in the way you have it presently) -- _**unless**_ you wish to move that method from the `Credential` interface to the `PublicKeyCredential interface` -- because we are defining it as part of `Credential` base class (or whatever the webidl term is) in [Credential Management PR #155](https://github.com/w3c/webappsec-credential-management/pull/155). Again, to override `Credential`'s `isConditionalMediationAvailable()`'s default _implementation_, we define an _implementation_ here, rather than attempt to override it using webidl's "partial interface". See the very last paragraph of [section 5.1. PublicKeyCredential Interface](https://w3c.github.io/webauthn/#iface-pkcredential): "PublicKeyCredential's interface object inherits Credential's implementation of ... and defines its own implementation of ... "  (yeah, it's subtle/obtuse). E.g., note how WebAuthn overrides`[[Store]]`'s implementation.  Similarly, `[[Create]]` and `[[DiscoverFromExternalSource]]` in webauthn are overriding their default impls in credman.  I can write this if it'd be helpful.

**update 2021-10-28** wrt (1):  I equalsJeffH am effectively wrong here (apologies!), we can do this as @nsatragno proposes. Credman PR #155 defines  `isConditionalMediationAvailable()` as a method (on the `Credential` interface), and as @nsatragno points out, the [WebIDL spec says in 2.2. Interfaces](https://webidl.spec.whatwg.org/#idl-interfaces): "Interfaces may specify an interface member that has the same name as one from an inherited interface. Objects that implement the derived interface will expose the member on the derived interface. It is language binding specific whether the overridden member can be accessed on the object."  We are assuming that this will work fine for JS's language binding. (@nsatragno to test)


2. The `silentCredentialDiscovery` operation simply gathers discoverable creds per a given RP from an authenticator?  If so I'm thinking we can say and spec that more straightforwardly. I have some ideas.

3. we ought to be more careful/precise with how we are using credential sources -- i think we are just using some of the items they contain, and "entire" cred sources ought not be passed in-whole to the client platform because technically doing so reveals the cred private key.  See also new issue  #1678.

index.bs
equalsJeffH: @emlun, @MasterKale (and other reviewers!): 
WRT @nsatragno's https://github.com/w3c/webauthn/pull/1576#discussion_r736620812 immediately above regarding  `isConditionalMediationAvailable()` being usable by devs as simply `PublicKeyCredential.isConditionalMediationAvailable()`, please also see the current state of w3c/webappsec-credential-management#155 -- We're feeling it is getting very close to being baked.

equalsJeffH (commented):
Hi @nsatragno -- this is looking pretty good overall, thanks!  there's a bunch of detail-level comments mostly having to do with maintaining the spec's alg style and use of the Infra spec.  Also a few suggestions for adding context for readers.

nsatragno:
@w3c/webauthn contributors interested in reviewing the credential management side of this patch can do so at [w3c/webappsec-credential-management/155](https://github.com/w3c/webappsec-credential-management/pull/155). I would also suggest considering joining the webappsec wg since that spec lives under that umbrella.

index.bs
```diff
@@ -2096,103 +2234,56 @@ When this method is invoked, the user agent MUST execute the following algorithm
         ::  [=set/For each=] |authenticator| in |issuedRequests| invoke the [=authenticatorCancel=] operation on |authenticator|
             and [=set/remove=] |authenticator| from |issuedRequests|. Return a {{DOMException}} whose name is "{{NotAllowedError}}".
 
+        :   If <code>|options|.{{CredentialRequestOptions/mediation}}</code> is {{CredentialMediationRequirement/conditional}}
+            and the user interacts with an [=input=] form control with an [=autocomplete=] attribute whose value
+            contains a `"webauthn"` [=autofill detail token=],
+        ::  1. Prompt the user to optionally select a [=DiscoverableCredentialMetadata=] from |silentlyDiscoveredCredentials|.
+
```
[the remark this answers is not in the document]
nsatragno: Done.

index.bs
```diff
@@ -1986,24 +2112,34 @@ When this method is invoked, the user agent MUST execute the following algorithm
...
-                ::  Recommended range: 30000 milliseconds to 180000 milliseconds.
-                ::  Recommended default value: 120000 milliseconds (2 minutes).
+    1. If <code>|options|.{{PublicKeyCredentialRequestOptions/allowCredentials}}</code> is not [=list/empty=],
+        return a {{DOMException}} whose name is "{{NotAllowedError}}", and terminate this algorithm.
 
-                :   is set to {{UserVerificationRequirement/required}} or {{UserVerificationRequirement/preferred}}
-                ::  Recommended range: 30000 milliseconds to 600000 milliseconds.
-                ::  Recommended default value: 300000 milliseconds (5 minutes).
-            </dl>
+    1. Set a timer |lifetimeTimer| to a value of infinity.
```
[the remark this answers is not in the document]
nsatragno: It's only partly because of that (if that was the only reason, we could keep the timer). The main reason is we don't want the credentials to suddenly stop appearing on the autofill prompt after some time has passed. That would be extremely confusing for the user. I'll add a note explaining this.

index.bs
```diff
@@ -1947,6 +1973,106 @@ This
 {{CredentialsContainer/get()|navigator.credentials.get()}} operation can be aborted by leveraging the {{AbortController}};
 see [[dom#abortcontroller-api-integration]] for detailed instructions.
 
+#### <dfn for="PublicKeyCredential" algorithm="Issuing a request to an authenticator">Issuing a request to an authenticator</dfn> #### {#sctn-issuing-request-to-authenticator}
```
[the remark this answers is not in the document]
nsatragno: > I'm also thinking that this alg actually ought to follow [[DiscoverFromExternalSource]] rather than precede it, because the opening paragraphs of section 5.1.4 serve to introduce [[DiscoverFromExternalSource]] per se

[[DiscoverFromExternalSource]] calls this algorithm though, so maybe that's an argument to keep the order like this?

I added an introductory paragraph.

index.bs
```diff
@@ -2096,103 +2234,56 @@ When this method is invoked, the user agent MUST execute the following algorithm
         ::  [=set/For each=] |authenticator| in |issuedRequests| invoke the [=authenticatorCancel=] operation on |authenticator|
             and [=set/remove=] |authenticator| from |issuedRequests|. Return a {{DOMException}} whose name is "{{NotAllowedError}}".
 
+        :   If <code>|options|.{{CredentialRequestOptions/mediation}}</code> is {{CredentialMediationRequirement/conditional}}
+            and the user interacts with an [=input=] form control with an [=autocomplete=] attribute whose value
+            contains a `"webauthn"` [=autofill detail token=],
+        ::  1. Prompt the user to optionally select a [=DiscoverableCredentialMetadata=] from |silentlyDiscoveredCredentials|.
```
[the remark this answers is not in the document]
nsatragno: Guarded this section under `if |credentialMetadata| is not [=list/empty=]`...

index.bs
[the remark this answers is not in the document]
equalsJeffH: > [[DiscoverFromExternalSource]] calls this algorithm though, so maybe that's an argument to keep the order like this?

I suppose. I'm just thinking from perspective of reader who reads the section 5.1.4 intro text, which intros `[[DiscoverFromExternalSource]]()`, but then the next subsection is an internal function of the latter.  I suppose with all the hyperlinks that it does not matter terribly.

index.bs
```diff
@@ -2081,6 +2226,8 @@ When this method is invoked, the user agent MUST execute the following algorithm
     [=authenticators=] can be <a href="https://en.wikipedia.org/w/index.php?title=Hot_plug">hot-plugged</a> into (e.g., via USB)
     or discovered (e.g., via NFC or Bluetooth) by the [=client=] by various mechanisms, or permanently built into the [=client=].
 
+1. Let |silentlyDiscoveredCredentials| be a new [=map/is empty|empty] [=map=] whose [=map/entry|entries=] are of the form: [=DiscoverableCredentialMetadata=] → [=authenticator=].
```
equalsJeffH, line 2229: Ah, we are missing a `=` here (sorry):
```suggestion
1. Let |silentlyDiscoveredCredentials| be a new [=map/is empty|empty=] [=map=] whose [=map/entry|entries=] are of the form: [=DiscoverableCredentialMetadata=] → [=authenticator=].
```

index.bs
```diff
@@ -2096,103 +2243,62 @@ When this method is invoked, the user agent MUST execute the following algorithm
...
-
-            Note: One way a [=client platform=] can determine that no |authenticator| will become available is by examining the <code>{{transports}}</code> members of the present <code>{{PublicKeyCredentialDescriptor}}</code> [=list/items=] of <code>|options|.{{PublicKeyCredentialRequestOptions/allowCredentials}}</code>, if any. For example, if all <code>{{PublicKeyCredentialDescriptor}}</code> [=list/items=] list only <code>{{AuthenticatorTransport/internal}}</code>, but all [=platform authenticator|platform=] |authenticator|s have been tried, then there is no possibility of satisfying the request. Alternatively, all <code>{{PublicKeyCredentialDescriptor}}</code> [=list/items=] may list <code>{{transports}}</code> that the [=client platform=] does not support.
+        :   If <code>|options|.{{CredentialRequestOptions/mediation}}</code> is {{CredentialMediationRequirement/conditional}}
+            and the user interacts with an [=input=] form control with an [=autocomplete=] attribute whose value
+            contains a `"webauthn"` [=autofill detail token=],
+        ::  1. If |credentialMetadata| is not [=list/empty=]:
```
equalsJeffH, line 2249: should this actually be:
```suggestion
        ::  1. If |silentlyDiscoveredCredentials| is not empty=]:
```
..?

index.bs
```diff
@@ -4027,6 +4131,52 @@ This operation is ignored if it is invoked in an [=authenticator session=] which
...
+        ::  Other information used by the [=authenticator=] to inform its UI.
+    </dl>
+
+1. [=map/For each=] [=public key credential source=] |credSource| of |authenticator|'s [=credentials map=]:
+
+    1. If |credSource| is not a [=client-side discoverable credential=], [=iteration/continue=].
+
+    1. If |credSource|.[=public key credential source/rpId=] is not |rpId|, [=iteration/continue=].
+
+    1. Let |discoveredCredentialMetadata| be a new [=DiscoverableCredentialMetadata] [=struct=] whose [=struct/items=]
```
equalsJeffH, line 4171: missing  `=`:
```suggestion
    1. Let |discoveredCredentialMetadata| be a new [=DiscoverableCredentialMetadata=] [=struct=] whose [=struct/items=]
```

equalsJeffH (commented):
thanks @nsatragno!  I have one material question and there's a couple of missing `=`s in intended `=]` bikeshed link closures.

equalsJeffH:
Further considerations upon reflection: wrt the proposed `silentCredentialDiscovery` operation, how will that interact with credentials' "[credential protection policy](https://fidoalliance.org/specs/fido-v2.1-ps-20210615/fido-client-to-authenticator-protocol-v2.1-ps-20210615.html#:~:text=The%20list%20of%20possible%20values%20for%20credProtect%20is)" (aka "[credProtect](https://fidoalliance.org/specs/fido-v2.1-ps-20210615/fido-client-to-authenticator-protocol-v2.1-ps-20210615.html#sctn-credProtect-extension)") ?  
 
( apologies if someone has already explained this and I've forgotten... )

 AFAiCT, in the present conditional media formulation (up thru ff37db6) the user does not even have to "interact" with page before the client platform invokes the `silentCredentialDiscovery` operation.  If an  authenticator becomes available and it supports the `silentCredentialDiscovery` operation, discoverable creds mapped to the present RP ID are enumerated, and user verification has not yet occurred (IIUC).  What happens if those creds were created at credProtect levels 2 or 3 ?  Presently, the latter creds would not be discovered.

RPs may override any platform-imposed credProtect policy --- are we implicitly assuming that RPs creating new user creds and wishing to employ conditional mediation need to also explicitly stipulate a credProtect policy of 1 (userVerificationOptional) when creating (discoverable) creds ?

In any case, IIUC, we probably ought to Note here in the WebAuthn spec credProtect implications wrt conditional mediation and provide some appropriate guidance for RPs ...?

index.bs
nsatragno: Whoops! Good catch.

equalsJeffH:
This PR is no longer a draft!  Please review it, as well as https://github.com/w3c/webappsec-credential-management/pull/155

thanks!
@equalsJeffH & @nsatragno

index.bs
```diff
@@ -1372,6 +1378,7 @@ that are returned to the caller when a new credential is created, or a new asser
         [SameObject] readonly attribute AuthenticatorResponse    response;
         [SameObject] readonly attribute DOMString?               authenticatorAttachment;
         AuthenticationExtensionsClientOutputs getClientExtensionResults();
+        static boolean isConditionalMediationAvailable();
```
emlun, line 1381: `isUserVerifyingPlatformAuthenticatorAvailable` returns a `Promise<Boolean>`, should we continue that pattern here? (Question, not a proposal)

index.bs
```diff
@@ -2290,8 +2396,6 @@ When this method is invoked, the user agent MUST execute the following algorithm
     user without [=user consent|consent=], this step MUST NOT be executed before |lifetimeTimer| has expired. See
     [[#sctn-assertion-privacy]] for details.
 
-During the above process, the user agent SHOULD show some UI to the user to guide them in the process of selecting and
-authorizing an authenticator with which to complete the operation.
```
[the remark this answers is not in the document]
emlun: Ah, I see now that it moved to the new paragraph in §5.1.4. Use an Existing Credential to Make an Assertion. I'll leave these comments here in case anyone else wonders the same thing.

emlun (dismissed):

index.bs
```diff
@@ -2096,103 +2243,62 @@ When this method is invoked, the user agent MUST execute the following algorithm
...
 
-                        1. If |allowCredentialDescriptorList| [=list/is empty=], [=continue=].
+        :   If an |authenticator| becomes available on this [=client device=],
+        ::  Note:  This includes the case where an |authenticator| was available upon |lifetimeTimer| initiation.
 
-                        1. Let |distinctTransports| be a new [=ordered set=].
+            1. If If <code>|options|.{{CredentialRequestOptions/mediation}}</code> is {{CredentialMediationRequirement/conditional}}
```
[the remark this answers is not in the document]
equalsJeffH: oops, thx!

index.bs
```diff
@@ -1947,6 +1973,111 @@ This
...
+                    and
+                    <code>|options|.{{PublicKeyCredentialRequestOptions/allowCredentials}}.{{PublicKeyCredentialDescriptor/type}}</code>.
+                    Set |allowCredentialDescriptorList| to this filtered list.
+
+                1. If |allowCredentialDescriptorList| [=list/is empty=], [=continue=].
+
+                1. Let |distinctTransports| be a new [=ordered set=].
+
+                1. If |allowCredentialDescriptorList| has exactly one value, set
+                    <code>|savedCredentialIds|[|authenticator|]</code> to <code>|allowCredentialDescriptorList|[0].id</code>'s
```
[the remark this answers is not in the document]
equalsJeffH: > maybe we should just state that this is a sub-algorithm rather than a standalone one?

The introductory paragraph to section 5.1.4.1. Issuing a Credential Request to an Authenticator attempts to state that, but without using a term like "sub-algorithm".  It could probably be more clear and perhaps use that latter term.

I also continue to feel that having this "Issuing a Credential Request" section be after the "[[DiscoverFromExternalSource]]" section would be better, [as discussed in a prior comment thread](https://github.com/w3c/webauthn/pull/1576#discussion_r739866766) (and in following comments there). Perhaps this will help convince @nsatragno? ;-)

index.bs
```diff
@@ -2096,103 +2243,62 @@ When this method is invoked, the user agent MUST execute the following algorithm
...
-                                    |clientDataHash|, |allowCredentialDescriptorList|, |userVerification|, and
-                                    |authenticatorExtensions| as parameters.
-                            </dl>
+                    Note: A request will be issued to this authenticator upon user selection of a credential.
 
-                    :   [=list/is empty=]
-                    ::  Using local configuration knowledge of the appropriate transport to use with |authenticator|, invoke the
-                        [=authenticatorGetAssertion=] operation on |authenticator| with |rpId|, |clientDataHash|,
-                        |userVerification| and |authenticatorExtensions| as parameters.
+            1. Else:
```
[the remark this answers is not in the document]
equalsJeffH: Good catch, thanks.

I suspect that the below is more along the lines of what we want here in this "If an authenticator becomes available on this client device," case of the "switch". 

In the below, step 1 and its substeps are unchanged, and step 2 is now an "else if", and a new step 3 clarifies that in the case of conditional mediation, we continue to wait for an authenticator that supports the silentCredentialDiscovery operation to become available.

WDTY @nsatragno?

```
1. If options.mediation is conditional and the authenticator supports 
    the silentCredentialDiscovery operation:
    1. < collect silentlyDiscoveredCredentials for later >

2. Else if options.mediation is not conditional:
    1. Execute the issuing a credential request to an authenticator 
       algorithm with authenticator and options.

3. Else:
    1. Assert: options.mediation is conditional and the authenticator does not 
       support the silentCredentialDiscovery operation.
    2. Terminate these substeps and continue (the while loop).
```

index.bs
```diff
@@ -4027,6 +4131,52 @@ This operation is ignored if it is invoked in an [=authenticator session=] which
...
+
+This is an OPTIONAL operation authenticators MAY support to enable {{CredentialMediationRequirement/conditional}}
+[=user mediation=].
+
+It takes the following input parameter:
+
+: |rpId|
+:: The caller's [=RP ID=], as <a href='#GetAssn-DetermineRpId'>determined</a> by the user agent and the client.
+
+If [=silentCredentialDiscovery=] is not supported, the operation always returns an empty [=list=].
```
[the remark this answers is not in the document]
equalsJeffH: Well, I think this is there to handle the case where the client requests silentCredentialDiscovery even though the authenticator does not support it. 

A well behaved client using an authenticator API that is able to indicate whether the authnr supports the operation ought not do that. 

I think the intention is to not throw an user-visible error.

index.bs
equalsJeffH: Good question. IIRC @nsatragno has some reasoning as to why it is not cast as a Promise. I _think_ it might be that this method is a query for the client itself and can be instantly answered without having to, for example, assess the platform's capabilities (which might take some time).

index.bs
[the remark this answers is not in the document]
emlun: That could perhaps be even clearer as:

```
1. If options.mediation is conditional:
    1. If the authenticator supports the silentCredentialDiscovery operation:
        1. < collect silentlyDiscoveredCredentials for later >
    2. Else:
        1. Terminate these substeps and continue (the while loop).

2. Else (i.e., if options.mediation is not conditional):
    1. Execute the issuing a credential request to an authenticator 
       algorithm with authenticator and options.
```

index.bs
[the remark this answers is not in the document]
nsatragno: Alright, I have moved this algorithm below [[DiscoverFromExternalSource]], passed |savedCredentialIds| as a parameter to avoid referencing it out of nowhere, and declared it as a subalgorithm of [[DiscoverFromExternalSource]]. PTAL.

index.bs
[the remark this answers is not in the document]
nsatragno: This is on purpose. A user agent should allow use of authenticators that don't support `silentCredentialDiscovery` through the "traditional" way (perhaps under some specific action by the user).

I have added a note to make the intent more clear.

index.bs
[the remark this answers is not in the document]
nsatragno: > If the authenticator doesn't support the operation at all, wouldn't it be the client that performs the fallback?

Yeah and we're already prescribing this on the [[DiscoverFromExternalSource]] algorithm.

> A well behaved client using an authenticator API that is able to indicate whether the authnr supports the operation ought not do that.

Yeah, and there is no scenario I can think of where this would malfunction.

Removed.

index.bs
```diff
@@ -2096,103 +2138,66 @@ When this method is invoked, the user agent MUST execute the following algorithm
...
-                            value (see [here](#authenticatorGetAssertion-return-values) in [[#sctn-op-get-assertion]] for more information).
+                1. [=list/For each=] |credentialMetadata| of |collectedDiscoveredCredentialMetadata|:
 
-                        1. [=list/For each=] credential descriptor |C| in |allowCredentialDescriptorList|,
-                            [=set/append=] each value, if any, of <code>|C|.{{transports}}</code> to |distinctTransports|.
+                    1. [=map/Set=] |silentlyDiscoveredCredentials|[|credentialMetadata|] to |authenticator|.
 
-                            Note: This will aggregate only distinct values of {{transports}} (for this [=authenticator=]) in
-                                |distinctTransports| due to the properties of [=ordered sets=].
+                    Note: A request will be issued to this authenticator upon user selection of a credential.
```
equalsJeffH, line 2192: Per our discussion, suggesting we expand this note and link to the switch case where the request is issued due to "...particular user context" -- I could not come up with a reasonable way to condense **"input** form control with an autocomplete attribute whose value contains a "webauthn" autofill detail token")

( I also indented it another level):
```suggestion
                        Note: A request will be issued to this authenticator upon user selection of a credential via
                        interaction with a particular UI context
                        (see [here](#GetAssn-ConditionalMediation-Interact-FormControl) for details).
```

index.bs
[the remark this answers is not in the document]
equalsJeffH: Ah, yes, that's subtle, thanks for adding the Note!

index.bs
```diff
@@ -2096,103 +2138,66 @@ When this method is invoked, the user agent MUST execute the following algorithm
...
-                        |userVerification| and |authenticatorExtensions| as parameters.
+                1. Execute the [=issuing a credential request to an authenticator=] algorithm with |authenticator|, |savedCredentialIds| and |options|.
 
-                        Note: In this case, the [=[RP]=] did not supply a list of acceptable credential descriptors. Thus, the
-                            authenticator is being asked to exercise any credential it may possess that is [=scoped=] to
-                            the [=[RP]=], as identified by |rpId|.
-                </dl>
+                    Note: This branch is taken if <code>|options|.{{CredentialRequestOptions/mediation}}</code> is {{CredentialMediationRequirement/conditional}}
+                    and the |authenticator| does not support the [=silentCredentialDiscovery=] operation to allow use of such authenticators during a
+                    {{CredentialMediationRequirement/conditional}} [=user mediation=] request.
```
equalsJeffH, line 2200: Nit:
```suggestion
                    Note: This branch is taken if <code>|options|.{{CredentialRequestOptions/mediation}}</code> is {{CredentialMediationRequirement/conditional}}
                    and the |authenticator| does not support the [=silentCredentialDiscovery=] operation. This allows the use of such authenticators during a
                    {{CredentialMediationRequirement/conditional}} [=user mediation=] request.
```

equalsJeffH (commented):
Hi @nsatragno, thanks for your considered update and clarifications. Per our discussion I'm making some (hopefully) clarifying-for-the-reader suggestions below. There's also a question about what might be just a latent spec bug.  I tested the new anchor and link the comments below add, hopefully they Just Work.

index.bs
emlun: Agreed, sounds good to me.

equalsJeffH (commented):
This is getting close, thx @nsatragno!
The only semi-substantive comment I have on the PR's current state at  9e23ab0 is: upon re-reading the intro to section 5.1.4 with refreshed eyes, it seems that the new 3d paragraph beginning with "In general, the user agent SHOULD show some UI..." splits up two paragraphs that ought to remain together. I suggest moving the new "In general..." 3d paragraph down one paragraph so those two existing paragraphs remain together.

index.bs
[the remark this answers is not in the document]
equalsJeffH: LGTM, thanks for doing all that.

index.bs
```diff
@@ -2096,103 +2132,72 @@ When this method is invoked, the user agent MUST execute the following algorithm
...
 
             1. [=set/Append=] |authenticator| to |issuedRequests|.
```
emlun, line 2202: Looks like this should be indented into the preceding `Else` block? (Since the earlier "If options.mediation is conditional and the user interacts with an input form control with an autocomplete attribute whose value contains a "webauthn" autofill detail token" case appends to _issuedRequests_ at that time instead)

equalsJeffH: By "this", if you are referring to L2202 "3. Append authenticator to issuedRequests.", then I am thinking you are correct.

index.bs
```diff
@@ -1986,24 +2011,37 @@ When this method is invoked, the user agent MUST execute the following algorithm
...
+    {{CredentialMediationRequirement/conditional}}:
 
-        Recommended ranges and defaults for the {{PublicKeyCredentialRequestOptions/timeout}} member of |options| are as follows.
-          If <code>|options|.{{PublicKeyCredentialRequestOptions/userVerification}}</code>
-            <dl class="switch">
-                :   is set to {{UserVerificationRequirement/discouraged}}
-                ::  Recommended range: 30000 milliseconds to 180000 milliseconds.
-                ::  Recommended default value: 120000 milliseconds (2 minutes).
+    1. If <code>|options|.{{PublicKeyCredentialRequestOptions/allowCredentials}}</code> is not [=list/empty=],
+        return a {{DOMException}} whose name is "{{NotAllowedError}}", and terminate this algorithm.
```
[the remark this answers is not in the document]
nsatragno: Agreed, this should be NotSupportedError. Done.

index.bs
nsatragno: Good catch, you're right. Done.

nsatragno (commented):
Addressed emlun's comments.

> it seems that the new 3d paragraph beginning with "In general, the user agent SHOULD show some UI..." splits up two paragraphs that ought to remain together. I suggest moving the new "In general..." 3d paragraph down one paragraph so those two existing paragraphs remain together.

Done!

equalsJeffH:
Note: the ["Credential Management: Conditional Mediation" W3C TAG review request #692](https://github.com/w3ctag/design-reviews/issues/692) made by @nsatragno back on 25-Nov-2021 is now [approved](https://github.com/w3ctag/design-reviews/issues/692#issuecomment-1026638121) (yay!)

equalsJeffH:
Cleaning up a couple of old to-do items: Where `options.mediation` is tested for the value of "conditional", update the Note to link to the step further below where the "...user interacts with an \<input> form control...", and also mention that the `get()` simply not resolving rather than throwing exceptions is handled in Credman.

Additionally, in earlier discussion with @nsatragno, she emphasized that we "have to ensure we do not return exceptions when doing conditional UI, even if the user's authnr or other things are not right.  However, if the webdev has made an error of some form, then we should return an error right away."  I think an example of the latter sort of webdev error would be if they use the "webauthn" autofill detail token on their form control but do not set `options.mediation` to "conditional".

Also, we have been discussing the actual name of the autofill detail token.  Some candidates are: "webauthn", "credential", "conditional", and is currently not decided.  The requisite changes to the HTML spec have not yet been proposed, and this will be a discussion item once the latter proposal (as a PR on the HTML spec) is put forward.

equalsJeffH:
A further note: 
there is consideration of having the conditional UI autofill detail token be a sort of modifier to the 
the "username" and "current-password" and "new-password" (or maybe a subset) autofill field names. The webdev would do somenthing like `autocomplete="webauthn current-password"` on their form control.

Also, the username" and "current-password" and "new-password" autofill field names can be used on these form controls: `<input>, <textarea>, <select>`.  Conditional UI ought to nominally work with `<input>, <textarea>` or maybe all three. 

These are things that need to be sorted out in the context of the requisite HTML spec changes (ie, PR) -- with WebAuthn folk providing input. 



are usable with these form control elements:  <input>, <textarea>, <select>

MasterKale:
> there is consideration of having the conditional UI autofill detail token be a sort of modifier to the
> the "username" and "current-password" and "new-password" (or maybe a subset) autofill field names. The webdev would do somenthing like `autocomplete="webauthn current-password"` on their form control.

What would get autofilled into the `current-password` input in this scenario? Is the idea here that a user clicking a credential in response to conditional mediation when the Password input gains focus instead somehow leads to submission of the credential? And could this support JavaScript-less submission of a `<form>` for authenticating with a WebAuthn credential?

equalsJeffH:
> What would get autofilled into the current-password input in this scenario? Is the idea here that a user clicking a credential in response to conditional mediation when the Password input gains focus instead somehow leads to submission of the credential?

[ I changed the example [in my earlier post](https://github.com/w3c/webauthn/pull/1576#issuecomment-1078430998) to be `autocomplete="webauthn username"` because it's more simple ]

The idea is (roughly): upon page load, if `isConditionalMediationAvailable()` is true, then JS does an async call of `nav.creds.get()` with `options.mediation="conditional"` (i.e., "in the background"), and the page's HTML is:
```html
<label for="name">Username:</label>
<input type="text" name="name" autofill="webauthn username">
```
Then, the conditional UI non-modal cred-picker dialog is displayed (positioned near/at the form control field the webauthn autofill detail token is associated with) listing any available webauthn/fido creds for this user at this RP (if there's none, perhaps still display the dialog listing only the "Sign with another device..." item):
image
If the user clicks outside the dialog, the dialog disappears and autofill occurs as normal for that form control field (in this case, it would be the username that is autofilled), and would continue with the legacy flow (e.g. username/password). In this case the `get()` call would never resolve.

However, if the user selects a cred displayed in the non-modal cred-picker dialog, then the page's background `get()` call proceeds using the selected credential and the user is prompted as usual by the client platform for appropriate UV gesture/PIN/etc, and the user is signed-in using webauthn.

See also the Conditional UI explainer.

MasterKale:
I made a point to re-read the explainer before asking my question and still had some uncertainty of how it's all intended to work.

After reading your response what I think the explainer needs is a `.then()` on the `.get()` with some pseudo code explaining that that's when you submit the response to the RP.

dwaite:
I would not expect username or password to be auto-filled based on a mediated selection of a PublicKeyCredential - I would expect the promise to complete and the site to determine how it wanted to update its UX based on that completion.

None of the values in the [PublicKeyUserCredentialEntity](https://www.w3.org/TR/webauthn-2/#dictdef-publickeycredentialuserentity) are appropriate for a username or email field, and we do not have a value which would be appropriate for a password field.

akshayku (approved):

nsatragno:
Since the last review:
* Merged the AbortController changes in.
* Had `isConditionalMediationAvailable()` return a promise instead of a raw boolean, see issue 1745
* Removed a TODO for the html spec change: https://github.com/whatwg/html/pull/8006

This is ready to land from my side

emlun:
> * Had `isConditionalMediationAvailable()` return a promise instead of a raw boolean, see issue 1745

Hm, on closer inspection... can we actually do this? The [function definition in `Credential`](https://w3c.github.io/webappsec-credential-management/#dom-credential-isconditionalmediationavailable) returns a raw `boolean`, so I don't actually think we can override it with a different return type, can we?

emlun (changes requested):

index.bs
```diff
@@ -2338,10 +2341,116 @@ When this method is invoked, the user agent MUST execute the following algorithm
...
+    :   <dfn>authenticator</dfn>
+    ::  A [=client platform=]-specific handle identifying an [=authenticator=] presently available on this [=client platform=].
+
+    :   <dfn>savedCredentialIds</dfn>
+    ::  A [=map=] containing [=authenticator=] → [=credential ID=]. This argument will be modified in this algorithm.
+
+    :   <dfn>options</dfn>
+    ::  This argument is a {{CredentialRequestOptions}} object whose
+        <code>|options|.{{CredentialRequestOptions/publicKey}}</code> member contains a {{PublicKeyCredentialRequestOptions}}
+        object specifying the desired attributes of the [=public key credential=] to discover.
```
emlun, line 2365: This parameter is used below as if it is just the `PublicKeyCredentialRequestOptions` part, not the whole `CredentialRequestOptions` object. There is a related, pre-existing issue in the parent `get()` (and `create()`) algorithms too, see: #1752

emlun (changes requested):

index.bs
nsatragno: Alright, let's fix it here only for this sub-algorithm in this PR to avoid piling up more changes on this. I had the caller pass in the `options.publicKey` instead.

index.bs
```diff
@@ -2037,24 +2058,37 @@ When this method is invoked, the user agent MUST execute the following algorithm
...
+    {{CredentialMediationRequirement/conditional}}:
 
-        Recommended ranges and defaults for the {{PublicKeyCredentialRequestOptions/timeout}} member of |options| are as follows.
-          If <code>|options|.{{PublicKeyCredentialRequestOptions/userVerification}}</code>
-            <dl class="switch">
-                :   is set to {{UserVerificationRequirement/discouraged}}
-                ::  Recommended range: 30000 milliseconds to 180000 milliseconds.
-                ::  Recommended default value: 120000 milliseconds (2 minutes).
+    1. If <code>|options|.{{PublicKeyCredentialRequestOptions/allowCredentials}}</code> is not [=list/empty=],
+        return a {{DOMException}} whose name is "{{NotSupportedError}}", and terminate this algorithm.
```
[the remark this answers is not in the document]
emlun: Ah, I see. Indeed stateless credentials would be hard to integrate in an account picker flow. Supporting a non-empty `allowCredentials` would enable CTAP2 creds with `credProtect` level 2 to work with this flow, though I guess that's a bit niche since the main use case for this is to use discoverable credentials without first identifying the user. And you're definitely right that it's easier to add the feature later than to remove it once introduced.

emlun (changes requested):

wseltzer:
Note: @equalsJeffH was a member at the time of the PR, so the IPR flag can be overridden.

ve7jtb (approved):
LGTM

sbweeden (approved):

nsatragno (approved):

agl:
From the call of 2022-06-29: we're happy to merge this once the conflict has been resolved. The IPR bot might object to Jeff, but Wendy or others can override because Wendy is happy that Jeff was a member when he did the work.

PR #1588, opened 2021-03-25
Explicitly restrict NONE aaguid to none attestation only

herrjemand:
AAGUID 00000000-0000-0000-0000-000000000000 must not be reused by any other attestation.

Apple has recently fixed that (Thank you Jiewen), so we should be explicit with intent of that AAGUID.

Ref:
https://github.com/WebKit/WebKit/blob/b133f3082ab5f5c409915ef3aafaa2ee15196d6d/Source/WebKit/UIProcess/WebAuthentication/Cocoa/LocalAuthenticator.mm
https://bugs.webkit.org/show_bug.cgi?id=217945

cc: @sbweeden

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/pull/1588.html" title="Last updated on Mar 25, 2021, 9:06 AM UTC (8e2194b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/1588/9ce4288...8e2194b.html" title="Last updated on Mar 25, 2021, 9:06 AM UTC (8e2194b)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -3326,7 +3326,8 @@ Each authenticator stores a <dfn for=authenticator>credentials map</dfn>, a [=ma
 Additionally, each authenticator has an AAGUID, which is a 128-bit identifier indicating the type (e.g. make and model) of the
 authenticator. The AAGUID MUST be chosen by the manufacturer to be identical across all substantially identical authenticators
 made by that manufacturer, and different (with high probability) from the AAGUIDs of all other types of authenticators.
-The AAGUID for a given type of authenticator SHOULD be randomly generated to ensure this. The [=[RP]=] MAY use the AAGUID to infer certain
+The AAGUID for a given type of authenticator SHOULD be randomly generated to ensure this. The NONE AAGUID(00000000-0000-0000-0000-000000000000
+) is reserved for NONE attestation, and SHALL not be used for other types of attestation. The [=[RP]=] MAY use the AAGUID to infer certain
 properties of the authenticator, such as certification level and strength of key protection, using information from other sources.
 
 The primary function of the authenticator is to provide [=WebAuthn signatures=], which are bound to various contextual data. These


herrjemand:
p.s.  We already implying that anyway, so this is just clarification:

> different (with high probability) from the AAGUIDs of all other types of authenticators. The AAGUID for a given type of authenticator SHOULD be randomly generated to ensure this.

emlun:
There's also this in [step 20 of §5.1.3. Create a New Credential](https://w3c.github.io/webauthn/#ref-for-aaguid%E2%91%A0):

>If `credentialCreationData.attestationConveyancePreferenceOption`’s value is
>- **"none"**
>  Replace potentially uniquely identifying information with non-identifying versions of the same:
>   1. If the AAGUID in the attested credential data is 16 zero bytes, `credentialCreationData.attestationObjectResult.fmt` is "packed", and "x5c" is absent from `credentialCreationData.attestationObjectResult`, then self attestation is being used and no further action is needed.
>
>  [...]

So it looks like at least in theory, the zero AAGUID is also valid for "packed" self attestation, but I don't know if any authenticators are actually producing such attestation statements. Either way, there's no corresponding instruction about the AAGUID in the ["packed" attestation signing procedure](https://w3c.github.io/webauthn/#ref-for-self-attestation%E2%91%A0%E2%91%A4):

>3. If self attestation is in use, the authenticator produces _sig_ by concatenating _authenticatorData_ and _clientDataHash_, and signing the result using the credential private key. It sets _alg_ to the algorithm of the credential private key and omits the other fields.

herrjemand:
That is weird. At FIDO we don't allow SELF(aka SURROGATE) authenticators to have ZERO AAGUID. All authenticators must have AAGUID, and in case of SELF attestation, in none attestation scenario, AAGUID will be returned, because it can not be used for correlation.

Kieun:
As far as I remember, U2F authenticators have zero AAGUID within the authenticator data structure. So, basic attestation or self attestation with U2F authenticators is allowed to have zero AAGUID.  

@herrjemand 
> and in case of SELF attestation, in none attestation scenario, AAGUID will be returned, because it can not be used for correlation.

If the attestation conveyance is set to "none" and the authenticator is a type of self (surrogate) authenticators, the client also needs to stripped off all the identifiable data (AAGUID and attestationObject). If not, the RP may correlate users with this AAGUID if it is non-zero.

So my thinking is that the zero AAGUID is only allowed to **None** attestation format except **FIDO U2F** format.

emlun:
Right, I thought I remembered something about the [`fido-u2f` attestation statement format](https://www.w3.org/TR/2021/PR-webauthn-2-20210225/#sctn-fido-u2f-attestation) also using zero AAGUID but couldn't find it in the spec. We should probably elaborate in the signing procedure that this particular format usually involves the client rearranging the raw U2F response into the WebAuthn attestation object format. A U2F authenticator doesn't return a zero AAGUID, rather the client inserts that as part of constructing the _attested credential data_ structure.

MasterKale:
> A U2F authenticator doesn't return a zero AAGUID, rather the client inserts that as part of constructing the attested credential data structure.

I'm glad you mention this because I was reimplementing FIDO-U2F verification a couple of weeks back and couldn't figure out _why_ a zero AAGUID check was in the list of steps. I did some searching around and could only find some off-handed comment that led me to believe it was something FIDO-related but that was it. I'd support including something like this bit in those verification steps for some additional context to help make some more sense of it.

herrjemand:
Yes... So maybe say none and legacy attestations?

MasterKale:
Ah, it turns out I'd left a note for myself! I found where the part of the FIDO spec that defines "all-zero AAGUID" for U2F devices wishing to use the CTAP2 `authenticatorMakeCredential` Command:

https://fidoalliance.org/specs/fido-v2.1-rd-20191217/fido-client-to-authenticator-protocol-v2.1-rd-20191217.html#u2f-authenticatorMakeCredential-interoperability

Screen Shot 2021-03-25 at 9 44 15 AM

The reason this tidbit of knowledge ever became known to me was because I bothered with FIDO Conformance Testing - it specifically requires you to check that AAGUID is all zeroes for FIDO-U2F attestations. If it weren't for that I never would have known to enforce that based on the WebAuthn spec alone.

herrjemand:
Yes, this is correct assumption @MasterKale

herrjemand:
Per decision of the group

emlun:
From 2021-03-31 WG call: @agl objects that there seems to be little reason to ban the zero AAGUID for all future implementations, and that doing so would likely lead to divergences as different platforms choose different solutions for how to set a nonzero AAGUID when an authenticator does not provide one. WebAuthn encompasses more than just FIDO authenticators, so the WebAuthn API can be more permissive than FIDO certification requirements are.

PR #1600, opened 2021-04-21
Remove user handle "MAY be null" statement

### LINKED ISSUE #1598: "The user handle MUST NOT be empty, though it MAY be null" - but only in responses?
In [§5.4.3 User Account Parameters for Credential Generation (dictionary PublicKeyCredentialUserEntity)](https://www.w3.org/TR/2021/REC-webauthn-2-20210408/#dom-publickeycredentialuserentity-id), the description of the `id` attribute reads:

>[...] The user handle MUST NOT be empty, though it MAY be null.

However the definition of the same attribute marks it `required`, and Chrome 89 returns an error if you attempt to set it to null in a `navigator.credentials.create()` call.

I think this was supposed to mean that the user handle _parameter_ [`publicKey.user.id`](https://www.w3.org/TR/2021/REC-webauthn-2-20210408/#dom-publickeycredentialuserentity-id) MUST NOT be empty and MUST NOT be null, but the user handle _return value_ [`PublicKeyCredential.response.userHandle`](https://www.w3.org/TR/2021/REC-webauthn-2-20210408/#dom-authenticatorassertionresponse-userhandle) MAY be null (for example when using a U2F authenticator or a non-discoverable credential). Is that right?

equalsJeffH:
As far as we can tell, the statement "...though it MAY be null" clause of the statement:
> [...] The user handle MUST NOT be empty, though it MAY be null.

..is incorrect and ought to be removed.

cf. https://www.w3.org/TR/2021/REC-webauthn-2-20210408/#dom-publickeycredentialuserentity-id

fixes #1598

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/pull/1600.html" title="Last updated on Apr 21, 2021, 9:59 PM UTC (965a18f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/1600/d3b1124...965a18f.html" title="Last updated on Apr 21, 2021, 9:59 PM UTC (965a18f)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2638,7 +2638,7 @@ credential.
         {{PublicKeyCredentialUserEntity/displayName}} nor {{PublicKeyCredentialEntity/name}} members. See Section 6.1 of [[!RFC8266]].
 
         The [=user handle=] MUST NOT contain [PII] about the user, such as a username or e-mail address;
-        see [[#sctn-user-handle-privacy]] for details. The [=user handle=] MUST NOT be empty, though it MAY be null.
+        see [[#sctn-user-handle-privacy]] for details. The [=user handle=] MUST NOT be empty.
 
         Note: the [=user handle=] <i>ought not</i> be a constant value across different accounts, even for [=non-discoverable credentials=], because some authenticators always create [=discoverable credentials=]. Thus a constant [=user handle=] would prevent a user from using such an authenticator with more than one account at the [=[RP]=].
 


agl (approved):

emlun (approved):
LGTM, thanks!

akshayku (approved):

PR #1654, opened 2021-07-27
Fix inconsistency in RP directions for credential transports

### LINKED ISSUE #1587: Inconsistent RP directions for handling credential transports
The definition of [`AuthenticatorAttestationResponse.getTransports()`](https://www.w3.org/TR/2021/PR-webauthn-2-20210225/#dom-authenticatorattestationresponse-transports-slot) reads:

>[...] The values SHOULD be members of AuthenticatorTransport but Relying Parties MUST ignore unknown values.

However [§7.1. Registering a New Credential, step 23](https://www.w3.org/TR/2021/PR-webauthn-2-20210225/#ref-for-dom-authenticatorattestationresponse-gettransports%E2%91%A3) reads (emphasis added):

>[...]
>It is RECOMMENDED to also:
>- Associate the credentialId with the transport hints returned by calling credential.response.getTransports(). **This value SHOULD NOT be modified** before or after storing it. [...]

These directions seem to be in conflict, and I think the latter is more correct. The transports value is primarily for the client to consume while the RP just needs to echo it back later, so the RP shouldn't need to care about what it contains. So I think we should edit the former to read:

>The values SHOULD be members of AuthenticatorTransport but Relying Parties MUST accept unknown values.

or perhaps:

>The values SHOULD be members of AuthenticatorTransport but Relying Parties SHOULD accept unknown values.

emlun:
Fixes #1587.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/pull/1654.html" title="Last updated on Jul 27, 2021, 4:11 PM UTC (a36e38c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/1654/33b0215...a36e38c.html" title="Last updated on Jul 27, 2021, 4:11 PM UTC (a36e38c)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2399,7 +2399,7 @@ during registration.
     ::  This operation returns the {{COSEAlgorithmIdentifier}} of the new credential. See [[#sctn-public-key-easy]].
 
     :   <dfn>\[[transports]]</dfn>
-    ::  This [=internal slot=] contains a sequence of zero or more unique {{DOMString}}s in lexicographical order. These values are the transports that the [=authenticator=] is believed to support, or an empty sequence if the information is unavailable. The values SHOULD be members of {{AuthenticatorTransport}} but [=[RPS]=] MUST ignore unknown values.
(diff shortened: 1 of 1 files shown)


akshayku (approved):

ve7jtb (approved):
LGTM

agl (approved):

sbweeden (approved):

Issue #1659, opened 2021-08-06
Choosing An Attestation Statement Format?


DanielSanchezDiaz:
Hello! This is a really well written specification, super helpful! I do have one question however. Is there a way to convey a preference for an attestation statement format to the authenticator? From the spec: "It is expected that most authenticators will support a small number of attestation types and attestation statement formats, while Relying Parties will decide what attestation types are acceptable to them by policy." This part made it seem like relying parties only had control over the attestation type but not the attestation statement format fmt. Since authenticators might be able to support different types of attestation statement formats, I'd like to know if there is a way to convey that you would prefer say "packed" over "fido-u2f"
Thanks so much!

sbweeden:
To answer your question, there is not. 

In practical use, the RP doesn't get to decide and every authenticator I've seen supports just one type for mainstream WebAuthn use (e.g. packed-basic). Regarding the example you used, I've only ever seen fido-u2f attestations when using authenticators or clients that don't support CTAP2. 

Do you have a specific use case where you think an individual authenticator *should* support more than one attestation statement format?

arshadnoor:
You could continue to be "lazy" if you used an open-source, FIDO 
Certified server that supports most, if not all, attestation formats. 
You could even get "lazier" by not having to write any FIDO code and 
just focus on the business application. :-D

https://sourceforge.net/projects/strongkeyfido
https://github.com/strongkey/fido2

Arshad Noor
StrongKey

equalsJeffH:
this is addressed by PR #1660, closing.

PR #1660, opened 2021-08-06
Explicitly state that RPs cannot in general choose attestation type/format

emlun:
Fixes issue #1659.

Marking this as draft for now because I'm not entirely sure this addition is worth its weight. I'd like to hear what some others think.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/pull/1660.html" title="Last updated on Aug 20, 2021, 2:45 PM UTC (d24439d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/1660/38ad84d...d24439d.html" title="Last updated on Aug 20, 2021, 2:45 PM UTC (d24439d)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -4097,9 +4097,8 @@ The privacy, security and operational characteristics of [=attestation=] depend 
 - The characteristics of the individual [=authenticator=], such as its construction, whether part or all of it runs in a secure
     operating environment, and so on.
 
-The [=attestation type=] and [=attestation statement format=] is chosen by the [=authenticator=];
-[=[RPS]=] may signal some limited preferences via the <code>{{PublicKeyCredentialCreationOptions/attestation}}</code> parameter,
-but cannot otherwise influence this choice.
+The [=attestation type=] and [=attestation statement format=] is chosen by the [=authenticator=] and [=client=];
+[=[RPS]=] can only signal limited [=attestation conveyance=] preferences during [=registration=].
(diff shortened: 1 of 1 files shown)


agl (approved):

equalsJeffH (commented):
Hm. I agree we ought to make it explicitly clear that the authenticator chooses the attestation type&format. However, the proposed statement confuses attestation type&format with attestation conveyance, which is a different thing.

equalsJeffH (commented):
thx @emlun, nominally looks OK to me, but I wonder about the "and client" portion of the clause:
> [=attestation statement format=] is chosen by the [=authenticator=] and [=client=]
...is that referring to the "none" attestation conveyance case where the client modifies the returned attestation statement?  If so, I'm inclined to excise the "and [=client=]" and then it LGTM.

emlun:
That and the `"indirect"` attestation conveyance case. But yeah, we can probably just leave it at saying just the authenticator chooses. Any reader who wants to know more about attestation will easily find the exceptions to that anyway.

agl (approved):

sbweeden (approved):
LGTM

emlun:
Since @equalsJeffH commented:

>nominally looks OK
>[...]
>I'm inclined to excise the "and [=client=]" and then it LGTM.

it seems like this is ready to merge. Thanks everyone!

PR #1664, opened 2021-08-23
Define 1023 bytes to be the maximum credential ID length.

### LINKED ISSUE #1617: further explain, for RPs, lack of stipulation of maximum Credential ID length
the Note at https://www.w3.org/TR/webauthn/#dom-publickeycredential-identifier-slot says:
> This API does not constrain the format or length of this identifier, except that it MUST be sufficient for the authenticator to uniquely select a key. For example, an authenticator without on-board storage may create identifiers containing a credential private key wrapped with a symmetric key that is burned into the authenticator.

We did not specify a maximum Credential ID length for various reasons (that are not well-documented, sorry. See https://github.com/w3c/webauthn/issues/40#issuecomment-219035185 for the apparent summary decision:
> RPs should write defensive code that blocks credential IDs that are ridiculously large.

Though, we _could_ add another paragraph to the above Note along the lines of:
> RPs need to be prepared to accept credential IDs of varying length, because the credential ID's size is determined by the authenticator and can vary. For example, they may defensively decline to register credentials having credential IDs greater than some maximum length the RP decides. A possibly reasonable maximum value is 1024 bytes.

( I just pulled 1024 bytes out of the air. Has anyone been logging the credential ID lengths they've seen? )

agl:
The value 1024 has been picked to be a nice round number that seems
larger than any reasonable credential ID should need to be.

(I think several of the step references were _previously_ incorrect and
have attempted to fix them here. Thus step number changes are not
expected to be simply updates reflecting the new step.)

Fixes #1617

***
<a href="https://pr-preview.s3.amazonaws.com/agl/webauthn/pull/1664.html" title="Last updated on Sep 29, 2021, 11:29 PM UTC (7888d10)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/1664/38f9f72...agl:7888d10.html" title="Last updated on Sep 29, 2021, 11:29 PM UTC (7888d10)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -4123,7 +4123,7 @@ object=] for a given credential. Its format is shown in <a href="#table-attested
         <tr>
             <td><dfn>credentialIdLength</dfn></td>
             <td>2</td>
-            <td>Byte length <strong>L</strong> of [=credentialId=], 16-bit unsigned big-endian integer. Value MUST be &lt; 1024.</td>
+            <td>Byte length <strong>L</strong> of [=credentialId=], 16-bit unsigned big-endian integer. Value MUST be &le; 1023.</td>
         </tr>
         <tr>
             <td><dfn>credentialId</dfn></td>


sbweeden (approved):
LGTM

index.bs
```diff
@@ -4481,8 +4481,9 @@ In order to perform a [=registration ceremony=], the [=[RP]=] MUST proceed as fo
         - If [=None|no attestation=] was provided, verify that [=None=] attestation is acceptable under [=[RP]=] policy.
         - If [=self attestation=] was used, verify that [=self attestation=] is acceptable under [=[RP]=] policy.
         - Otherwise, use the X.509 certificates returned as the [=attestation trust path=] from the [=verification procedure=]
-            to verify that the attestation public key either correctly chains up to an acceptable root certificate, or is itself an acceptable certificate (i.e., it and the root certificate obtained in Step 20 may be the same).
+            to verify that the attestation public key either correctly chains up to an acceptable root certificate, or is itself an acceptable certificate (i.e., it and the root certificate obtained in Step 19 may be the same).
 
+1. Check that the <code>[=credentialId=]</code> is &le; 1024 bytes. Credential IDs larger than 1024 bytes SHOULD cause the RP to fail this [=registration ceremony=].
```
emlun, line 4486: Nit:
```suggestion
1. Check that the <code>[=credentialId=]</code> is at most 1024 bytes long. Credential IDs longer than 1024 bytes SHOULD cause the RP to fail this [=registration ceremony=].
```

emlun (approved):
Thanks! LGTM modulo one minor language nit.

If we're recommending that RPs should enforce compliance, might it make sense to also allow (or even recommend) clients to do the same?

lgarron:
A lot of power-of-2-based limits are defined so that the values exactly fill the state space of a bit string of a certain length>

To that end, would it make sense to make the limit `< 1024`, i.e. 1023?

Or is this so high-level that there isn't a point in trying to account for bit-level details?

emlun:
@lgarron I don't follow... what would be the benefit of that?

selfissued:
Per discussion on the 8-Sep-21 call, I think that it should be < 1024.

MasterKale:
I brought this up at the latest WG meeting, but are we sure 1024 is even sufficient? Some authenticators wrap a bunch of data into credential ID and so my gut says _maybe_ it's possible to end up needing to handle more than 1024 bytes.

emlun:
My question was also answered on the call: the thing that "exactly fill[s] the state space of a bit string of a certain length" is the _length_ of the value - so limiting the length to 2<sup>N</sup>-1 means the _length_ encoding doesn't need one more bit just to be able to encode the value 2<sup>N</sup>.

emlun:
@MasterKale Let's estimate an upper limit. A [Public Key Credential Source](https://www.w3.org/TR/2021/WD-webauthn-3-20210427/#public-key-credential-source) consists of:

- **type** which can probably be encoded as a single byte, or not at all.
- **id** whose max size is being discussed here.
- **privateKey** whose size depends on the algorithm. As far as I know, 4096-bit RSA keys are the largest keys currently seeing widespread use, so keys larger than 512 bytes seem unlikely in practice.
- **rpId** for which I'm not sure there is a theoretical max length, but if you're embedding all this data into the credential ID it's probably not a discoverable key. For non-discoverable keys you really only need to be able to _verify_ the the `rpId`, which you can instead do with a MAC since the `rpId` will be passed as a parameter at authentication time. HMAC-SHA256 generates 32 bytes long MACs, which should be more than enough entropy for the purpose. And you should already have a MAC anyway to make the credential ID as a whole immutable.
- **userHandle** which has a defined max length of 64 bytes.
- **otherUI** which is technically open-ended, but likely contents are `rp.name`, `user.name` and `user.displayName`, each of which MAY be truncated to 64 bytes.

So all in all that would be 1 + 512 + 32 + 64 + 3*64 = **801 bytes** if the authenticator chooses to encode _everything_ into the credential ID (apart from the credential ID itself). So a 1023 limit still leaves a spare 222 bytes for longer `otherUI` if desired, a longer MAC, or key sizes up to 5872 bits. And with a smaller key, like a 256-bit elliptic curve key, you'll only need **321 bytes**, with 702 spare bytes left over.

(Maybe you'll also need a few more bytes to encode the key algorithm and some value lengths, but that's mostly insignificant in comparison.)

MasterKale:
@emlun Thank you for breaking that down! 1023 bytes seems more than adequate now, with plenty of room to spare.

index.bs
```diff
@@ -4481,8 +4481,9 @@ In order to perform a [=registration ceremony=], the [=[RP]=] MUST proceed as fo
         - If [=None|no attestation=] was provided, verify that [=None=] attestation is acceptable under [=[RP]=] policy.
         - If [=self attestation=] was used, verify that [=self attestation=] is acceptable under [=[RP]=] policy.
         - Otherwise, use the X.509 certificates returned as the [=attestation trust path=] from the [=verification procedure=]
-            to verify that the attestation public key either correctly chains up to an acceptable root certificate, or is itself an acceptable certificate (i.e., it and the root certificate obtained in Step 20 may be the same).
+            to verify that the attestation public key either correctly chains up to an acceptable root certificate, or is itself an acceptable certificate (i.e., it and the root certificate obtained in Step 19 may be the same).
 
+1. Check that the <code>[=credentialId=]</code> is &lt; 1024 bytes. Credential IDs larger than this bytes SHOULD cause the RP to fail this [=registration ceremony=].
```
lgarron, line 4486: Nit: Typo?
```suggestion
1. Check that the <code>[=credentialId=]</code> is &lt; 1024 bytes. Credential IDs larger than this many bytes SHOULD cause the RP to fail this [=registration ceremony=].
```

(Or "this number of bytes"?)

agl: Thanks!

emlun (changes requested):

equalsJeffH (approved):
LGTM, thx @agl!  fwiw, I concur with @emlun's suggestions.

lgarron:
Glad to see this updated to ≤1023 :-D

Would it be worth making that more clear in the PR title as well?

emlun (approved):

index.bs
```diff
@@ -4481,8 +4481,9 @@ In order to perform a [=registration ceremony=], the [=[RP]=] MUST proceed as fo
         - If [=None|no attestation=] was provided, verify that [=None=] attestation is acceptable under [=[RP]=] policy.
         - If [=self attestation=] was used, verify that [=self attestation=] is acceptable under [=[RP]=] policy.
         - Otherwise, use the X.509 certificates returned as the [=attestation trust path=] from the [=verification procedure=]
-            to verify that the attestation public key either correctly chains up to an acceptable root certificate, or is itself an acceptable certificate (i.e., it and the root certificate obtained in Step 20 may be the same).
+            to verify that the attestation public key either correctly chains up to an acceptable root certificate, or is itself an acceptable certificate (i.e., it and the root certificate obtained in Step 19 may be the same).
 
+1. Check that the <code>[=credentialId=]</code> is &lt; 1024 bytes. Credential IDs larger than this many bytes SHOULD cause the RP to fail this [=registration ceremony=].
```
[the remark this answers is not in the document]
emlun: Heheh :smile:

As mentioned on today's call, I don't really have a preference between "< 1024" and "≤ 1023", it's just that the above suggestion seemed less obtuse than changing the latter half to "Credential IDs with a length larger than or equal to this many bytes"... :slightly_smiling_face:

index.bs
```diff
@@ -4123,7 +4123,7 @@ object=] for a given credential. Its format is shown in <a href="#table-attested
         <tr>
             <td><dfn>credentialIdLength</dfn></td>
             <td>2</td>
-            <td>Byte length <strong>L</strong> of Credential ID, 16-bit unsigned big-endian integer.</td>
+            <td>Byte length <strong>L</strong> of [=credentialId=], 16-bit unsigned big-endian integer. Value MUST be &lt; 1024.</td>
```
equalsJeffH, line 4126: do we want to say `&le; 1023` here also?

equalsJeffH (approved):
A question below, otherwise LGTM.

index.bs
agl: (Done.)

emlun (approved):

index.bs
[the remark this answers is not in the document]
mikel-misha: > Some day you can ask me about the court case related to this :)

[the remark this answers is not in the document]
mikel-misha: > > Some day you can ask me about the court case related to this :)
> 
>

Issue #1666, opened 2021-08-25
Authenticator Attachment in Public Key Credential 


z11h:
In order for a site to know whether a local platform authenticator was used, or whether the user used another device and thus might want to register a local platform authenticator, this is a proposal that an authenticator attachment field be added to the public key credential response on both registration success & authentication success.

Thus, if the proposed authenticator attachment field of the attestation/assertion is “cross-platform”, and isUVPAA (i.e a user-verifying platform authenticator is available [1]) returns true, then sites should have the ability to offer to the user to register the current device's platform authenticator.

(In order to avoid superfluously re-registering devices if the user happened to use a phone or security key to sign in, even though the platform authenticator is already registered, the site may wish to track the registration status of the platform authenticator in local state. If the authenticator attachment was “internal” then the local state should be set to reflect that.)

Reference: https://github.com/w3c/webauthn/issues/1637
[1] https://w3c.github.io/webauthn/#user-verifying-platform-authenticator

saara09:
https://w3c.github.io/webauthn/#user-verifying-platform-authenticator

Issue #1637, opened 2021-07-01
Possible experiences in a future WebAuthn


agl:
Dear WG members,

With the publication of WebAuthn Level Two and CTAP 2.1, the WebAuthn standard is now reasonably mature. Google has been talking with some other parties about how to build upon this to help sites fully migrate their users to a new default credential built on WebAuthn that can broadly replace the username+password pattern.

Some components of these ideas have been discussed for some time, or have been recently made public:
* Apple [discussed](https://developer.apple.com/videos/play/wwdc2021/10106/) syncing of platform credentials at WWDC a few weeks ago. ([Summary]( https://www.cnet.com/news/apple-says-its-new-logon-tech-is-as-easy-as-passwords-but-far-more-secure/))
* Google has been talking about caBLE and phones as security keys for quite a while.
* Microsoft have [supported](https://www.microsoft.com/en-us/microsoft-365/blog/2018/11/20/sign-in-to-your-microsoft-account-without-a-password-using-windows-hello-or-a-security-key/) 1st-factor sign-in via WebAuthn for some years.
* Issue #1545 sketches out ideas for how WebAuthn UI might be more hospitable to RPs that have a population of people used to passwords (see PR #1576 for spec implementation. Conditional UI requires additional support in the underlying Credential Management spec. This was done in [Credman PR #155](https://github.com/w3c/webappsec-credential-management/pull/155) (landed)).
* Issue #1546 raised the possibility of associating a second, device-bound key with a credential,  Issue #1658 provides a concrete design, and #1663 defines the `devicePubKey` extension (implementing the design).

With this message (which is really an abuse of the GitHub issue system to post something!) we would like to draw these threads together.

(Also, we would like to briefly acknowledge that password managers and federation provide some of the same benefits as WebAuthn and we wish to note that nothing in this message signals a reduction in investment in these areas by Google.)

### Possible experiences in a future WebAuthn

In order to motivate discussions about what WebAuthn changes may be needed to support wider adoption we would like to sketch some user experiences that we might want to enable:

When signing up on a new website, rather than being forced to choose a password, a user is invited to register a WebAuthn authenticator. They choose to register their phone, using a technology like caBLE. This creates a discoverable credential on the phone and that credential is encrypted and synced so that loss of the phone doesn’t lock the user out.

[...]

PR #1668, opened 2021-09-08
Add transport used during authentication to assertion payload

z11h:
This spec change adds the transport used during authentication to the assertion payload returned from the browser to the RP.

Fixes: #1666 

Also discussed in #1637

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2207,6 +2207,10 @@ When this method is invoked, the user agent MUST execute the following algorithm
                 :   <code><dfn for="assertionCreationData">signatureResult</dfn></code>
                 ::  whose value is the bytes of the signature value returned by the [=authenticator=].
 
+                :   <code><dfn for="assertionCreationData">transportResult</dfn></code>
+                ::  whose value is the transport used to communicate to the [=authenticator=]. Values SHOULD be members of {{AuthenticatorTransport}}. If the user agent does not have any transport information, set the value to 
+                    null.
+
                 :   <code><dfn for="assertionCreationData">userHandleResult</dfn></code>
                 ::  If the [=authenticator=] returned a [=user handle=], set the value of [=userHandleResult=] to be the bytes of
                     the returned [=user handle=]. Otherwise, set the value of [=userHandleResult=] to null.
@@ -2241,6 +2245,11 @@ When this method is invoked, the user agent MUST execute the following algorithm
                         :   {{AuthenticatorAssertionResponse/signature}}
                         ::  A new {{ArrayBuffer}}, created using |global|'s [=%ArrayBuffer%=], containing the bytes of
                             <code>|assertionCreationData|.[=assertionCreationData/signatureResult=]</code>.
+
+                        :   {{AuthenticatorAssertionResponse/transport}}
+                        ::  If <code>|assertionCreationData|.[=assertionCreationData/transportResult=]</code> is null, set this
+                            field to null. Otherwise, set this field to a new {{DOMString}}, containing the value of
+                            <code>|assertionCreationData|.[=assertionCreationData/transportResult=]</code>.
 
                         :   {{AuthenticatorAssertionResponse/userHandle}}
                         ::  If <code>|assertionCreationData|.[=assertionCreationData/userHandleResult=]</code> is null, set this
@@ -2429,6 +2438,7 @@ optionally evidence of [=user consent=] to a specific transaction.
     interface AuthenticatorAssertionResponse : AuthenticatorResponse {
         [SameObject] readonly attribute ArrayBuffer      authenticatorData;
         [SameObject] readonly attribute ArrayBuffer      signature;
+        [SameObject] readonly attribute DOMString?       transport;
         [SameObject] readonly attribute ArrayBuffer?     userHandle;
     };
 </xmp>
@@ -2444,6 +2454,10 @@ optionally evidence of [=user consent=] to a specific transaction.
 
     :   <dfn>signature</dfn>
     ::  This attribute contains the raw signature returned from the authenticator. See [[#sctn-op-get-assertion]].
+
+    :   <dfn>transport</dfn>
+    ::  This attribute contains the transport used to communicate to the [=authenticator=], or `null` if the user agent does not have any transport information.
+        See [[#sctn-op-get-assertion]].
 
     :   <dfn>userHandle</dfn>
     ::  This attribute contains the [=user handle=] returned from the authenticator, or null if the authenticator did not return a


nsatragno (approved):
lgtm

Firstyear:
This field isn't part of the signed collected client data, so can't this be tampered with during the response? This may lead to failse assumptions of security/validity of this data. There is already a signed extension for this purpose, so how is this an improvement?

agl:
> This field isn't part of the signed collected client data, so can't this be tampered with during the response?

The user-agent could "tamper" with it, yes. As with several other fields (including the transport list in the registration).

> There is already a signed extension for this purpose, so how is this an improvement?

There's no extension defined for this but, even if we defined one, it wouldn't have any support from authenticators. An extension isn't needed for the motivating use here, which is allow a website to figure out whether to offer to register a platform authenticator.

Firstyear:
> There's no extension defined for this but, even if we defined one, it wouldn't have any support from authenticators. An extension isn't needed for the motivating use here, which is allow a website to figure out whether to offer to register a platform authenticator.

If this is the case, it should be called a transportSelectionHint to make it extremely clear that it is not an enforceable security property. There have already been multiple instances of RP's implementing webauthn that have incorrectly assumed that criteria are security properties rather than UX hints, and I think we should improve this in the language we use for these values.

agl:
> If this is the case, it should be called a transportSelectionHint to make it extremely clear that it is not an enforceable security property. There have already been multiple instances of RP's implementing webauthn that have incorrectly assumed that criteria are security properties rather than UX hints, and I think we should improve this in the language we use for these values.

I don't think we can save such RPs. We can't rename everything that exists and, if we start calling things transportAsReportedByTheBrowser then it suggests that the other values _are_ coming from the authenticator!

agl:
Another point from the above: might RPs do silly things with this like decided that assertions over USB are ok but NFC aren't? If we're saying that the point is to determine platform vs not, might be spell this `attachment=platform` or `attachment=cross-platform`?

index.bs
```diff
@@ -2445,6 +2455,10 @@ optionally evidence of [=user consent=] to a specific transaction.
     :   <dfn>signature</dfn>
     ::  This attribute contains the raw signature returned from the authenticator. See [[#sctn-op-get-assertion]].
 
+    :   <dfn>transport</dfn>
+    ::  This attribute contains the transport used to communicate to the [=authenticator=], or `null` if the user agent does not have any transport information.
+        See [[#sctn-op-get-assertion]].
```
sbweeden, line 2460: Why have this reference to section _6.3.3. The authenticatorGetAssertion Operation_ as part of the description for `transport` when the value is the responsibility of the user agent and not the authenticator?

Perhaps just remove this reference?

sbweeden: I also really feel there should be some description of the intended use of this new field. When I was first developing an RP implementation, my focus was on section 7, and for assertion ceremonies 7.2. Perhaps here in 5.2.2 is the right place for an explanatory note, or perhaps in section 7.2, either way there should be an indication that this value is a hint to the RP as to how the client communicated with the authenticator, for possible consideration in what might happen next in a user experience.

index.bs
```diff
@@ -2429,6 +2438,7 @@ optionally evidence of [=user consent=] to a specific transaction.
     interface AuthenticatorAssertionResponse : AuthenticatorResponse {
         [SameObject] readonly attribute ArrayBuffer      authenticatorData;
         [SameObject] readonly attribute ArrayBuffer      signature;
+        [SameObject] readonly attribute DOMString?       transport;
```
sbweeden, line 2441: I think we should at least consider also having this field in the _AuthenticatorAttestationResponse_
I'd like to see that idea discussed on a future WebAuthn call before merging this PR.

My thinking here is that getTransports(), which is a method that is also part of the  _AuthenticatorAttestationResponse_ interface, returns a list and in cases where the authenticator supports more than one transport it is still useful to know which transport was used during a registration ceremony.

sbweeden (changes requested):

index.bs
emlun: In that case, perhaps the property should be on `PublicKeyCredential` instead of on the `AuthenticatorResponse`? That would also help separate between `transport` and `transports` fields - having both on the same object seems like asking for bugs to happen.

sbweeden: Yes - that would be a good place for it.

z11h: Yup, I agree, having both transport and transports fields could get confusing, which is why instead of exposing the actual transport, it'll now be exposed as platform attachment, which is either platform or cross-platform as suggested by Adam above. Does that resolve the issue with having similar names?

sbweeden: I would like to see and review the new intended PR as it's not quite clear to me what is now being proposed for API changes.

emlun: This also relates to a discussion in https://github.com/w3c/webauthn/pull/1621#discussion_r709721573 .

MasterKale: Keeping `transport`/`attachment`/etc... on `PublicKeyCredential` would also help developers track where that's returned (_if_ it gets returned, anyway), since it'd be a top-level property in the same location no matter if it's a registration or authentication response:

**Registration:**
```js
{
  id: '...',
  rawId: '...',
  response: {
    attestationObject: '...',
    clientDataJSON: '...',
  },
  type: 'public-key',
  transports: ['usb', '...'],  // <---
}
```

**Authentication:**
```js
{
  id: '...',
  rawId: '...',
  response: {
    authenticatorData: '...',
    clientDataJSON: '...',
    signature:
      '...',
  },
  type: 'public-key',
  transport: 'internal',   // <---
  attachment: 'platform',  // <---
}
```

z11h: @sbweeden, I've made updates to the API, have a look when you can.

also, @MasterKale, you're suggesting to move transports in attestation response during registration to PublicKeyCredential out of response? The transports fit more inside the attestationResponse, and there'll be only an attachment field (not transport & attachment) for both Registration & Authentication.

index.bs
```diff
@@ -1388,6 +1389,12 @@ that are returned to the caller when a new credential is created, or a new asser
         {{CredentialsContainer/create()}}, this attribute's value will be an {{AuthenticatorAttestationResponse}}, otherwise,
         the {{PublicKeyCredential}} was created in response to {{CredentialsContainer/get()}}, and this attribute's value
         will be an {{AuthenticatorAssertionResponse}}.
+    
+    :   <dfn>authenticatorAttachment</dfn>
+    ::  This attribute contains the authenticatorAttachment used to communicate to the [=authenticator=], or `null` if
```
nsatragno, line 1394: Can you make authenticatorAttachment a reference to the AuthenticatorAttachment enum? I believe

```suggestion
    ::  This attribute contains the {{AuthenticatorAttachment}} used to communicate to the [=authenticator=], or `null` if 
```

should do the trick.

index.bs
```diff
@@ -1388,6 +1389,12 @@ that are returned to the caller when a new credential is created, or a new asser
         {{CredentialsContainer/create()}}, this attribute's value will be an {{AuthenticatorAttestationResponse}}, otherwise,
         the {{PublicKeyCredential}} was created in response to {{CredentialsContainer/get()}}, and this attribute's value
         will be an {{AuthenticatorAssertionResponse}}.
+    
+    :   <dfn>authenticatorAttachment</dfn>
+    ::  This attribute contains the authenticatorAttachment used to communicate to the [=authenticator=], or `null` if 
+        the user agent does not have any attachment information. This attribute should be used by RPs in conjecture 
+        with isUVPAA to prompt user to register platform authenticator if it is availible and user is currently being authenticated
```
nsatragno, line 1396:
```suggestion
        with `isUserVerifyingPlatformAuthenticatorAvailable()` to prompt user to register a platform authenticator if it is available and the user is currently being authenticated 
```

index.bs
```diff
@@ -1388,6 +1389,12 @@ that are returned to the caller when a new credential is created, or a new asser
         {{CredentialsContainer/create()}}, this attribute's value will be an {{AuthenticatorAttestationResponse}}, otherwise,
         the {{PublicKeyCredential}} was created in response to {{CredentialsContainer/get()}}, and this attribute's value
         will be an {{AuthenticatorAssertionResponse}}.
+    
+    :   <dfn>authenticatorAttachment</dfn>
+    ::  This attribute contains the authenticatorAttachment used to communicate to the [=authenticator=], or `null` if 
+        the user agent does not have any attachment information. This attribute should be used by RPs in conjecture
```
nsatragno, line 1395:
```suggestion
        the user agent does not have any attachment information. This attribute may be used by RPs in conjecture 
```

index.bs
```diff
@@ -2207,6 +2214,10 @@ When this method is invoked, the user agent MUST execute the following algorithm
                 :   <code><dfn for="assertionCreationData">signatureResult</dfn></code>
                 ::  whose value is the bytes of the signature value returned by the [=authenticator=].
 
+                :   <code><dfn for="assertionCreationData">authenticatorAttachmentResult</dfn></code>
+                ::  whose value is the authenticator attachment type used to communicate to the [=authenticator=]. Values SHOULD be either "platform" or "cross-platform". If the user agent does not have any authenticator attachment information,
```
nsatragno, line 2218:
```suggestion
                ::  whose value is the authenticator attachment type used to communicate to the [=authenticator=]. Values SHOULD be members of the {{AuthenticatorAttachment}} enum. If the user agent does not have any authenticator attachment information, 
```

index.bs
```diff
@@ -2242,6 +2253,11 @@ When this method is invoked, the user agent MUST execute the following algorithm
                         ::  A new {{ArrayBuffer}}, created using |global|'s [=%ArrayBuffer%=], containing the bytes of
                             <code>|assertionCreationData|.[=assertionCreationData/signatureResult=]</code>.
 
+                        :   {{AuthenticatorAssertionResponse/authenticatorAttachment}}
+                        ::  If <code>|assertionCreationData|.[=assertionCreationData/authenticatorAttachmentResult=]</code> is null, set this
+                            field to null. Otherwise, set this field to a new {{DOMString}}, containing the value of
+                            <code>|assertionCreationData|.[=assertionCreationData/authenticatorAttachmentResult=]</code>.
+
```
nsatragno, line 2260: You need to add a similar piece of text to Create, i.e. section 5.1.3, step 20. "If any authenticator indicates success".

index.bs
```diff
@@ -2207,6 +2214,10 @@ When this method is invoked, the user agent MUST execute the following algorithm
                 :   <code><dfn for="assertionCreationData">signatureResult</dfn></code>
                 ::  whose value is the bytes of the signature value returned by the [=authenticator=].
 
+                :   <code><dfn for="assertionCreationData">authenticatorAttachmentResult</dfn></code>
```
nsatragno, line 2217: I don't think this needs to be part of `assertionCreationData` since the attachment doesn't come from the authenticator. Move this section directly to the bit below where it gets added to `PublicKeyCredentialResponse` (in this patch, that's line 2256).

index.bs
```diff
@@ -3962,6 +3978,10 @@ When this method is invoked, the [=authenticator=] MUST perform the following pr
 1. If any error occurred while generating the [=assertion signature=], return an error code equivalent to "{{UnknownError}}" and
     terminate the operation.
 
+1. Let |authenticatorAttachment| be <code>platform</code> or <code>cross-platform</code> based on the transport returned by authenticator following successful authentication.
```
nsatragno, line 3981: It's the User Agent's responsibility to tell the attachment, not the authenticator's. Remove this section.

nsatragno (changes requested):

index.bs
MasterKale: Forgive me, I mis-remembered something I do in one of my libraries as being part of the spec. I forgot that transports are only available as `AuthenticatorAttestationResponse.getTransports()`. I am _not_ advocating for moving anything out of `AuthenticatorAttestationResponse`.

What I _am_ pushing for is something like `AuthenticatorAssertionResponse.getAttachment()`, to keep things consistent with `AuthenticatorAttestationResponse.getTransports()`. I'm trying to keep track of everything being discussed in this PR and if my understanding is correct I think I agree with @sbweeden and disagree with @emlun...

> I would like to see and review the new intended PR as it's not quite clear to me what is now being proposed for API changes.

I think I agree with @sbweeden on this one, if this PR is moving away from exposing the exact transport used during authentication for authenticator attachment instead then it might make sense to close this out and prep a new PR.

index.bs
sbweeden: `conjecture`? or `conjunction`

index.bs
sbweeden: The reference to isUserVerifyingPlatformAuthenticatorAvailable should be a spec link

index.bs
z11h: yup, conjunction. fixing.

index.bs
emlun: This should also incorporate a "The value SHOULD be a member of AuthenticatorAttachment" part like other enum-ish fields. I'm not sure what the recommendation should be for unknown values, but maybe something like:

```
The value SHOULD be a member of {{AuthenticatorAttachment}}, and Relying Parties SHOULD treat unknown values as if the value were null.
```

index.bs
emlun: I agree this is the user agents responsibility, but I think parts of this could be merged into the section discussed in https://github.com/w3c/webauthn/pull/1668/files#r712433782 (which would in turn be moved to line 2256 as noted there). The merged text should also use the `{{AuthenticatorAttachment/platform}}` and `{{AuthenticatorAttachment/cross-platform}}` references.

emlun (changes requested):

index.bs
emlun: I think the idea now is to add the attachment property to both registration and authentication responses, since `transports` communicates a different (though related) piece of information. That would be the reason to put it on `PublicKeyCredential` instead, if it's going to be on both response types anyway.

z11h: @MasterKale, I agree with @emlun,I'm not seeing any reason for consistency between AuthenticatorAttestationResponse.getTransports() & AuthenticatorAssertionResponse.getAttachment(). They have different use cases, and for the main reason the attachment will be used, I'm pushing to keep the attachment on the top level of PublicKey Credential so its exposed both during registration (attestation) as well as authentication (assertion)

> I think I agree with @sbweeden on this one, if this PR is moving away from exposing the exact transport used during authentication for authenticator attachment instead then it might make sense to close this out and prep a new PR.

This PR has been updated with the change from transport -> authenticator attachment, and has good discussions within it, so I believe i'm going to keep this PR and not open up a new one.

MasterKale: Alright, I'm all caught up after re-reviewing the latest content of the PR. I'm onboard now with putting it on PublicKeyCredential :v:

nsatragno (approved):
lgtm

akshayku (approved):

equalsJeffH:
on 2021-09-22 call:
various folks to re-review

sbweeden (approved):

sbweeden (approved):

equalsJeffH (changes requested):
Thanks for this PR @z11h, this will be helpful in achieving our more-polished UX goals as outlined in issue #1637.  While this PR is broadly fine, it needs detail-level polish such that it meshes properly with how we've spec'd things. I've attempted to craft drop-in builds on your original text which accomplish this.

equalsJeffH (changes requested):
I've reviewed and caught one link mistake I made (apologies).  Also, these changes link to `[[#sctn-attachments-from-transports]]`  and assume the suggestions I made in PR #1670 are accepted <grin type="sheepish">, but it seems @emlun hasn't had a chance to review that as yet.

So we can fix the link below, and then either (a) wait for @emlun to review #1670 and proceed per however that works out, or (b) merge this PR and then be prepared to revise things as necessary depending on how #1670 shakes out.

What do folks wish to do?

z11h:
@equalsJeffH: Since @emlun is in a different time zone, I'm fine waiting until he has a look at your comments on #1670 and we can sit on this PR until that PR shakes out and we'll go ahead and merge this, to prevent any dead links.

index.bs
```diff
@@ -1388,6 +1389,12 @@ that are returned to the caller when a new credential is created, or a new asser
         {{CredentialsContainer/create()}}, this attribute's value will be an {{AuthenticatorAttestationResponse}}, otherwise,
         the {{PublicKeyCredential}} was created in response to {{CredentialsContainer/get()}}, and this attribute's value
         will be an {{AuthenticatorAssertionResponse}}.
+    
+    :   <dfn>authenticatorAttachment</dfn>
+    ::  This attribute reports the [=authenticator attachment modality=] in effect at the time this {{PublicKeyCredential}} was created.
```
ve7jtb, line 1394: In the case of an authentication ceremony, I don't think we want to report the attachment modality when the credential was created.  We want the current attachment modality.

This mostly impacts caBLE authenticators.  where a credential created as platform can later be used as roaming.

In that case I don't think the client could know via CTAP2 how the credential was creates and would probably in an implementation report it as cross-platform in practice.    So I think this is more a problem with the wording of the descryption.

timcappalli: Agreed. Current attachment modality is what is important.

z11h: you're correct, resolving.

equalsJeffH: hm, y'all are correct, I messed that wording up. dunno what I was thinking. @z11h will be proposing a fix.

emlun:
Thanks for your patience everyone! #1670 has now evolved to be mutually dependent on this PR, so I think it makes more sense to merge it into here instead of into #1621. I couldn't re-target the PR base branch though, because this branch is in https://github.com/z11h/webauthn, so I instead replaced #1670 with https://github.com/z11h/webauthn/pull/3.

z11h:
@emlun, merged your branch onto my fork.

all comments have been resolved, PR is ready for another look :)

index.bs
```diff
@@ -2417,6 +2432,26 @@ To remove the need to parse CBOR at all in many cases, {{AuthenticatorAttestatio
 
 Note: {{AuthenticatorAttestationResponse/getPublicKey()}} and {{AuthenticatorAttestationResponse/getAuthenticatorData()}} were only added in level two of this spec. [=[RPS]=] SHOULD use feature detection before using these functions by testing the value of `'getPublicKey' in AuthenticatorAttestationResponse.prototype`. [=[RPS]=] that require this function to exist may not interoperate with older user-agents.
 
+
+#### Determining Authenticator Attachment from Credential Transports #### {#sctn-attachments-from-transports}
```
equalsJeffH, line 2436:
```suggestion
### Determining Authenticator Attachment from Authenticator Transports ### {#sctn-attachments-from-transports}
```
...and additionally move this section up into section 5.1, perhaps after  `sctn-isUserVerifyingPlatformAuthenticatorAvailable`.

index.bs
```diff
@@ -2417,6 +2432,26 @@ To remove the need to parse CBOR at all in many cases, {{AuthenticatorAttestatio
...
+
+As a part of creating a {{PublicKeyCredential}} object, as the result of a  [=registration ceremony|registration=] or [=authentication ceremony=], the [=client platform=] determines the [=authenticator attachment modality=] and reports it in {{PublicKeyCredential/authenticatorAttachment}}. This section defines how to make that determination.
```
equalsJeffH, line 2438: Nit (correcting my poor diction):
```suggestion
When creating a {{PublicKeyCredential}} object, as the result of a  [=registration ceremony|registration=] or [=authentication ceremony=], the [=client platform=] determines the [=authenticator attachment modality=] and reports it in {{PublicKeyCredential/authenticatorAttachment}}. This section defines how the [=client platform=] makes that determination.
```

equalsJeffH (changes requested):
This is looking good @z11h & @emlun, thanks for merging PR #1670 into here.  Upon review it seems having the new `#sctn-attachments-from-transports` section embedded in section 5.2, which is about authenticator responses, is odd because it describes client platform behavior that sets an attribute on the `PublicKeyCredential` interface. I suggest we move the new `#sctn-attachments-from-transports` section up into section 5.1, say as a peer to and after  `sctn-isUserVerifyingPlatformAuthenticatorAvailable` (it'd now have only three hashs in its title).
WDYT?

Also, I suggest we use the existing term "authenticator transport" rather than introducing a new term "credential transport" (suggestions and a nit fix below).

Plus, we ought to have at least @nsatragno review the normative language we're adding here.

index.bs
nsatragno: nit:

```suggestion
As a part of creating a {{PublicKeyCredential}} object, as the result of a [=registration ceremony|registration=] or [=authentication ceremony=], the [=client platform=] determines the [=authenticator attachment modality=] and reports it in {{PublicKeyCredential/authenticatorAttachment}}. This section defines how to make that determination.
```

nsatragno (approved):
Still LGTM with a minor nit.

z11h:
@nsatragno resolved.

@equalsJeffH great comments, I've reordered the sections, and agree that authenticator transport is a more fitting term.

equalsJeffH (approved):

equalsJeffH:
thanks @z11h, LGTM.  If @emlun approves upon re-review we should be ready to merge :)

index.bs
```diff
@@ -1388,6 +1389,15 @@ that are returned to the caller when a new credential is created, or a new asser
...
         will be an {{AuthenticatorAssertionResponse}}.
+    
+    :   <dfn>authenticatorAttachment</dfn>
+    ::  This attribute reports the [=authenticator attachment modality=] in effect at the time the
+        {{CredentialsContainer/create()|navigator.credentials.create()}} or
+        {{CredentialsContainer/get()|navigator.credentials.get()}} methods successfully complete.
+        The attribute’s value SHOULD be a member of {{AuthenticatorAttachment}}. [=[RPS]=] SHOULD treat unknown values
+        as if the value were null.
+
+        Note: If, as the result of a [=registration ceremony|registration=] or [=authentication ceremony=], {{PublicKeyCredential/authenticatorAttachment}}'s value is "cross-platform" and concurrently {{isUserVerifyingPlatformAuthenticatorAvailable}} returns [TRUE], then the user employed a [=roaming authenticator=] for this [=ceremony=] while there is an available [=platform authenticator=]. Thus the [=[RP]=] has the opportunity to prompt the user to register the available [=platform authenticator=], which may enable more streamlined user experience flows.
```
emlun, line 1400: We can move the `Note:` from the new section to here instead.

```suggestion
        Note: If, as the result of a [=registration ceremony|registration=] or [=authentication ceremony=], {{PublicKeyCredential/authenticatorAttachment}}'s value is "cross-platform" and concurrently {{isUserVerifyingPlatformAuthenticatorAvailable}} returns [TRUE], then the user employed a [=roaming authenticator=] for this [=ceremony=] while there is an available [=platform authenticator=]. Thus the [=[RP]=] has the opportunity to prompt the user to register the available [=platform authenticator=], which may enable more streamlined user experience flows.

Note: An [=authenticator's=] [=authenticator attachment modality|attachment modality=] could change over time.
For example, a mobile phone might at one time only support [=platform attachment=]
but later receive updates to support [=cross-platform attachment=] as well.
```
```suggestion
        Note: If, as the result of a [=registration ceremony|registration=] or [=authentication ceremony=], {{PublicKeyCredential/authenticatorAttachment}}'s value is "cross-platform" and concurrently {{isUserVerifyingPlatformAuthenticatorAvailable}} returns [TRUE], then the user employed a [=roaming authenticator=] for this [=ceremony=] while there is an available [=platform authenticator=]. Thus the [=[RP]=] has the opportunity to prompt the user to register the available [=platform authenticator=], which may enable more streamlined user experience flows.
```

index.bs
```diff
@@ -1857,6 +1867,9 @@ a numbered step. If outdented, it (today) is rendered either as a bullet in the
                     :   {{PublicKeyCredential/[[identifier]]}}
                     ::  |id|
 
+                    :   {{PublicKeyCredential/authenticatorAttachment}}
+                    ::  A {{DOMString}} whose value SHOULD be a member of {{AuthenticatorAttachment}}, based on the [[#enum-transport|transport]] used by the [=authenticator=] during the [=registration ceremony=], as determined per [[#sctn-attachments-from-transports]].
```
emlun, line 1871: Since [AuthenticatorAttachment](https://w3c.github.io/webauthn/#enum-attachment) already defines a mapping between attachments and DOMStrings, we can probably use that here.

```suggestion
                    ::  The {{AuthenticatorAttachment}} value matching the current [=authenticator attachment modality=] of |authenticator|.
```

index.bs
```diff
@@ -2223,10 +2236,12 @@ When this method is invoked, the user agent MUST execute the following algorithm
                 1.  Let |pubKeyCred| be a new {{PublicKeyCredential}} object associated with |global| whose fields are:
 
                     :   {{PublicKeyCredential/[[identifier]]}}
-
                     ::  A new {{ArrayBuffer}}, created using |global|'s [=%ArrayBuffer%=], containing the bytes of
                         <code>|assertionCreationData|.[=credentialIdResult=]</code>.
 
+                    :   {{PublicKeyCredential/authenticatorAttachment}}
+                    ::  A {{DOMString}} whose value SHOULD be a member of {{AuthenticatorAttachment}}, based on the [[#enum-transport|transport]] used by the [=authenticator=] during the [=authentication ceremony=], as determined per [[#sctn-attachments-from-transports]].
```
emlun, line 2243: Since [AuthenticatorAttachment](https://w3c.github.io/webauthn/#enum-attachment) already defines a mapping between attachments and DOMStrings, we can probably use that here.

```suggestion
                    ::  The {{AuthenticatorAttachment}} value matching the current [=authenticator attachment modality=] of |authenticator|.
```

index.bs
```diff
@@ -2330,6 +2345,25 @@ Note: Invoking this method from a [=browsing context=] where the [=Web Authentic
...
+    : used an {{AuthenticatorTransport/internal}} transport, then
+    :: {{PublicKeyCredential/authenticatorAttachment}} MUST be set to "platform".
+
+    : used a {{transports|transport}} that is a member of {{transports}} other than {{AuthenticatorTransport/internal}},
+    :: {{PublicKeyCredential/authenticatorAttachment}} MUST be set to "cross-platform".
+
+Note: An [=authenticator's=] [=authenticator attachment modality|attachment modality=] could change over time.
+For example, a mobile phone might at one time only support [=platform attachment=]
+but later receive updates to support [=cross-platform attachment=] as well.
+
```
emlun, line 2366: As mentioned above, this section doesn't really make much sense in this new context. I think we should just delete it.

index.bs
```diff
@@ -1388,6 +1389,15 @@ that are returned to the caller when a new credential is created, or a new asser
...
+    ::  This attribute reports the [=authenticator attachment modality=] in effect at the time the
+        {{CredentialsContainer/create()|navigator.credentials.create()}} or
+        {{CredentialsContainer/get()|navigator.credentials.get()}} methods successfully complete.
+        The attribute’s value SHOULD be a member of {{AuthenticatorAttachment}}. [=[RPS]=] SHOULD treat unknown values
+        as if the value were null.
+
+        Note: If, as the result of a [=registration ceremony|registration=] or [=authentication ceremony=], {{PublicKeyCredential/authenticatorAttachment}}'s value is "cross-platform" and concurrently {{isUserVerifyingPlatformAuthenticatorAvailable}} returns [TRUE], then the user employed a [=roaming authenticator=] for this [=ceremony=] while there is an available [=platform authenticator=]. Thus the [=[RP]=] has the opportunity to prompt the user to register the available [=platform authenticator=], which may enable more streamlined user experience flows.
 
     :   {{PublicKeyCredential/getClientExtensionResults()}}
     ::  This operation returns the value of {{PublicKeyCredential/[[clientExtensionsResults]]}}, which is a [=map=] containing
```
emlun, line 1403: Below here, on [approximately line 1665](https://github.com/w3c/webauthn/pull/1668/files#diff-5e793325cd2bfc452e268a4aa2f02b4024dd9584bd1db3c2595f61f1ecf7b985R1665-R1666), there's a reference that is now becoming ambiguous. We need to make this change:

```diff
diff --git a/index.bs b/index.bs
index 139a97a..8c59fd3 100644
--- a/index.bs
+++ b/index.bs
@@ -1662,7 +1666,7 @@ a numbered step. If outdented, it (today) is rendered as a bullet in the midst o

             1. If <code>|options|.{{PublicKeyCredentialCreationOptions/authenticatorSelection}}</code> is present:

-                  1. If <code>|options|.{{PublicKeyCredentialCreationOptions/authenticatorSelection}}.{{authenticatorAttachment}}</code> is
+                  1. If <code>|options|.{{PublicKeyCredentialCreationOptions/authenticatorSelection}}.{{AuthenticatorSelectionCriteria/authenticatorAttachment}}</code> is
                     present and its value is not equal to |authenticator|'s [=authenticator attachment modality=], [=iteration/continue=].

                   1. If <code>|options|.{{PublicKeyCredentialCreationOptions/authenticatorSelection}}.{{residentKey}}</code>
```

index.bs
```diff
@@ -2330,6 +2345,25 @@ Note: Invoking this method from a [=browsing context=] where the [=Web Authentic
...
+    : used a {{transports|transport}} that is a member of {{transports}} other than {{AuthenticatorTransport/internal}},
+    :: {{PublicKeyCredential/authenticatorAttachment}} MUST be set to "cross-platform".
+
+Note: An [=authenticator's=] [=authenticator attachment modality|attachment modality=] could change over time.
+For example, a mobile phone might at one time only support [=platform attachment=]
+but later receive updates to support [=cross-platform attachment=] as well.
+
 ## Authenticator Responses (interface <dfn interface>AuthenticatorResponse</dfn>) ## {#iface-authenticatorresponse}
 
 [=Authenticators=] respond to [=[RP]=] requests by returning an object derived from the
```
emlun, line 2369: Below here, around approximately [lines 2735-2738](https://github.com/w3c/webauthn/pull/1668/files#diff-5e793325cd2bfc452e268a4aa2f02b4024dd9584bd1db3c2595f61f1ecf7b985L2701-L2704), we should add that the `AuthenticatorAttachment` enum is gaining a second role.

```diff
diff --git a/index.bs b/index.bs
index 139a97a..8c59fd3 100644
--- a/index.bs
+++ b/index.bs
@@ -2735,7 +2720,9 @@ attributes.
 This enumeration's values describe [=authenticators=]' [=authenticator attachment modality|attachment modalities=].
 [=[RPS]=] use this to express a preferred [=authenticator attachment modality=]
 when calling {{CredentialsContainer/create()|navigator.credentials.create()}}
-to [[#sctn-createCredential|create a credential]].
+to [[#sctn-createCredential|create a credential]],
+and [=clients=] use this to report the [=authenticator attachment modality=]
+used to complete a [=registration ceremony|registration=] or [=authentication ceremony=].

 <xmp class="idl">
     enum AuthenticatorAttachment {
```

emlun (changes requested):
Sorry for the back-and-forth, but seeing it in its new context I don't think the new section (`#sctn-attachments-from-transports`) makes much sense anymore - the client (unlike an RP) doesn't need to "determine the attachment from the transports", it already knows the attachment. So rewriting the section for this new purpose, the "algorithm" would become something like:

>If the authenticator
>- is using [platform attachment][plattach]
>    - set the value of `authenticatorAttachment` to [platform][plat].
>- is using [cross-platform attachment][xplattach],
>    - set the value of `authenticatorAttachment` to [cross-platform][xplat].

which seems pretty redundant since the [§5.4.5. Authenticator Attachment Enumeration (enum AuthenticatorAttachment)][enum] section already defines the same mapping. So I think we can get rid of the new section entirely and just refer back to [AuthenticatorAttachment][enum]. Here are some suggestions for that.

[enum]: https://w3c.github.io/webauthn/#enum-attachment
[plattach]: https://w3c.github.io/webauthn/#platform-attachment
[plat]: https://w3c.github.io/webauthn/#dom-authenticatorattachment-platform
[xplattach]: https://w3c.github.io/webauthn/#cross-platform-attachment
[xplat]: https://w3c.github.io/webauthn/#dom-authenticatorattachment-cross-platform

z11h (commented):
@emlun, proposal updated. :)

index.bs
```diff
@@ -1388,6 +1389,19 @@ that are returned to the caller when a new credential is created, or a new asser
...
+        {{CredentialsContainer/create()|navigator.credentials.create()}} or
+        {{CredentialsContainer/get()|navigator.credentials.get()}} methods successfully complete.
+        The attribute’s value SHOULD be a member of {{AuthenticatorAttachment}}. [=[RPS]=] SHOULD treat unknown values
+        as if the value were null.
+
+        Note: If, as the result of a [=registration ceremony|registration=] or [=authentication ceremony=], {{PublicKeyCredential/authenticatorAttachment}}'s value is "cross-platform" and concurrently {{isUserVerifyingPlatformAuthenticatorAvailable}} returns [TRUE], then the user employed a [=roaming authenticator=] for this [=ceremony=] while there is an available [=platform authenticator=]. Thus the [=[RP]=] has the opportunity to prompt the user to register the available [=platform authenticator=], which may enable more streamlined user experience flows.
+
+Note: An [=authenticator's=] [=authenticator attachment modality|attachment modality=] could change over time.
+For example, a mobile phone might at one time only support [=platform attachment=]
+but later receive updates to support [=cross-platform attachment=] as well.
```
emlun, line 1404: Sorry, my mistake - this needs to be indented to match the `Note:` before it.

index.bs
```diff
@@ -2326,10 +2345,6 @@ This method has no arguments and returns a Boolean value.
     };
 </xmp>
 
-Note: Invoking this method from a [=browsing context=] where the [=Web Authentication API=] is "disabled" according to the [=allowed to use=] algorithm&mdash;i.e., by a [=permissions policy=]&mdash;will result in the promise being rejected with a {{DOMException}} whose name is "{{NotAllowedError}}". See also [[#sctn-permissions-policy]].
-
-</div>
-
```
emlun, line 2332: Got a bit too much deleted in commit [0cd8e11b](https://github.com/w3c/webauthn/pull/1668/commits/0cd8e11b2edab596645c839afcd3231905aabed5#diff-5e793325cd2bfc452e268a4aa2f02b4024dd9584bd1db3c2595f61f1ecf7b985L2348-L2351) :)

emlun (changes requested):

PR #1680, opened 2021-11-05
Tighten requirements for rejecting duplicate credential IDs.

### LINKED ISSUE #1679: Does signing the credential public key with the attestation private key prove to the RP that the user owns the credential private key?
https://www.w3.org/TR/webauthn/#attestation-certificate says

> At registration time, the authenticator uses the attestation private key to sign the Relying Party-specific credential public key (and additional data) that it generates and returns via the authenticatorMakeCredential operation. Relying Parties use the attestation public key conveyed in the attestation certificate to verify the attestation signature

but in self attestation

> the Authenticator does not have any specific attestation key pair. Instead it uses the credential private key to create the attestation signature

To me, self attestation directly proves to the RP that the user has the credential private key since the credential public key is used to verify the attestation signature.

If webauthn was used as a usernameless registration / login, with self attestation, I would feel confident that after registering, they can also be signed in immediately without any further action. With other kinds of attestation (basic, none, etc), I feel that after registering, I would need to go through the login flow to get proof that the user has the private key. Am I misunderstanding?

agl:
The existing wording suggests rejecting registrations with duplicate
credential IDs, but says that sites may replace the record if they wish.
But accidential duplicate credential IDs aren't worth worrying about and
it's safer to always reject duplicates.

Include a note with the reasoning so that sites who don't want to do
this check can at least think about the implications.

Fixes #1679

***
<a href="https://pr-preview.s3.amazonaws.com/agl/webauthn/pull/1680.html" title="Last updated on Dec 1, 2021, 8:13 PM UTC (eae134f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/1680/a4b80ad...agl:eae134f.html" title="Last updated on Dec 1, 2021, 8:13 PM UTC (eae134f)">Diff</a>


emlun (approved):
Looks good to me!

index.bs
```diff
@@ -4514,9 +4514,9 @@ In order to perform a [=registration ceremony=], the [=[RP]=] MUST proceed as fo
...
 
-1. Check that the <code>[=credentialId=]</code> is not yet registered to any other user. If registration
-    is requested for a credential that is already registered to a different user, the [=[RP]=] SHOULD
-    fail this [=registration ceremony=], or it MAY decide to accept the registration, e.g. while deleting the older registration.
+1. Check that the <code>[=credentialId=]</code> is not yet registered for any user. If it is, the [=[RP]=] MUST fail this [=registration ceremony=].
+
+    NOTE: [=credential IDs=] contain sufficient entropy that the case of accidential duplication can be ignored. However, several [=attestation=] formats do not include a self-signature and thus an attacker, with access to a site's credential database, could attempt to register a victim's credential as their own. If the [=[RP]=] accepted this and replaced the existing registration, and [=discoverable credentials=] were used, then the victim could be forced to sign into the attackers account at their next attempt. Data saved by the victim in that state would then be available to the attacker. Thus duplicate credential IDs are rejected in this step.
```
lgarron, line 4519: It took me a few readings to understand that the first sentence wasn't negating the "MUST". What about something like that?
```suggestion
    NOTE: [=credential IDs=] contain sufficient entropy that an accidental ID collision is unlikely. However, several [=attestation=] formats do not include a self-signature and thus an attacker, with access to a site's credential database, could attempt to register a victim's credential as their own. If the [=[RP]=] accepted this and replaced the existing registration, and [=discoverable credentials=] were used, then the victim could be forced to sign into the attackers account at their next attempt. Data saved by the victim in that state would then be available to the attacker. Thus duplicate credential IDs are rejected in this step.
```
(Also, is this completely true, given randomness issues with low-power devices / virtual environments / OS bugs?)

equalsJeffH: I prefer @lgarron's suggested first sentence re-write.

equalsJeffH: Some modest suggested clarifications:
1. "several [=attestation=] formats" --- I think what you mean here is "[attestation types](https://www.w3.org/TR/webauthn/#sctn-attestation-types)".  IIUC, all attestation types other than "self attestation" do not include a self-signature by the cred private key.
2. an attacker would not necessarily need to have "access to a site's credential database", rather they would just need to have somehow obtained the victim's site-specific credential ID(s), which could e.g. also be done via infecting the user's browser in some fashion.
3. I'd state up front that this is rationale for RPs to simply reject duplicate cred IDs.
4. I would not want this potential attack exposition to scare off RPs from using WebAuthn.   Am not sure whether or how to additionally say something along the lines of "obtaining user cred IDs, eg by site account dbase compromise or browser compromise, does not give an attacker _direct_ access to victims' accounts, unlike the case of obtaining users' username+passwords."

This likely needs further polishing (and @lgarron's first sentence fix from above):
```suggestion
    NOTE: The rationale for [=[RPS]=] rejecting duplicate [=credential IDs=] is as follows: [=credential IDs=] contain sufficient entropy that the case of accidental duplication can be ignored. However, [=attestation types=] other than [=self attestation=] do not include a self-signature explicitly proving possession of the [=credential private key=] at [=registration=] time. Thus an attacker who has managed to obtain a user's [=credential ID=] for a site (this could be potentially accomplished in various ways), could attempt to register a victim's credential as their own at that site. If the [=[RP]=] accepts this new registration and replaces the victim's existing credential registration, and the [=discoverable credentials|credentials are discoverable=], then the victim could be forced to sign into the attacker's account at their next attempt. Data saved to the site by the victim in that state would then be available to the attacker.
```

equalsJeffH (commented):
Thx @agl, I've some modest suggestions below....

index.bs
emlun: I'd add to @equalsJeffH's suggestion that the attacker would also need the victim's public key:

```
[...] Thus an attacker who has managed to obtain a user's [=credential ID=] and [=credential public key=] for a site  [...]
```

agl: Have applied Jeff's suggestion and updated it with Emil's note that the public key is also needed.

I'm not sure if Lucas was proposing an explicit change for the first line but I've rewritten it to avoid the negative potentially applying to the MUST. (I also changed it to a SHOULD because we're giving the reasoning and thus kinda suggesting to the RP why it might be optional.)

index.bs
```diff
@@ -4514,9 +4514,9 @@ In order to perform a [=registration ceremony=], the [=[RP]=] MUST proceed as fo
 
 1. Check that the <code>[=credentialId=]</code> is &le; 1023 bytes. Credential IDs larger than this many bytes SHOULD cause the RP to fail this [=registration ceremony=].
```
equalsJeffH, line 4515: pedantic nit:  we use "verify that" in all the other steps:
```suggestion
1. Verify that the <code>[=credentialId=]</code> is &le; 1023 bytes. Credential IDs larger than this many bytes SHOULD cause the RP to fail this [=registration ceremony=].
```

index.bs
```diff
@@ -4514,9 +4514,9 @@ In order to perform a [=registration ceremony=], the [=[RP]=] MUST proceed as fo
...
 
-1. Check that the <code>[=credentialId=]</code> is not yet registered to any other user. If registration
-    is requested for a credential that is already registered to a different user, the [=[RP]=] SHOULD
-    fail this [=registration ceremony=], or it MAY decide to accept the registration, e.g. while deleting the older registration.
+1. Check that the <code>[=credentialId=]</code> is not yet registered for any user. If the <code>[=credentialId=]</code> is already known then the [=[RP]=] SHOULD fail this [=registration ceremony=].
```
equalsJeffH, line 4517: pedantic nit:  we use "verify that" in all the other steps:
```suggestion
1. Verify that the <code>[=credentialId=]</code> is not yet registered for any user. If the <code>[=credentialId=]</code> is already known then the [=[RP]=] SHOULD fail this [=registration ceremony=].
```

equalsJeffH (approved):
LGTM, thx @agl! (bonus pedantic nits included ;-)

akshayku (approved):

PR #1693, opened 2022-01-25
Change "create any credentials" to "create any assertions" in get() description

### LINKED ISSUE #1667: Cross origin authentication without iframes (accommodating SPC in WebAuthn)
On WebAuthn WG call on 2021-08-25, a proposal concerning [secure payment confirmation (SPC)](https://github.com/w3c/secure-payment-confirmation) was discussed which probably originated from web payments groups. Attaching for more details: 
[[WebAuthn WG August 2021] Secure Payment Confirmation.pdf](https://github.com/w3c/webauthn/files/7051084/WebAuthn.WG.August.2021.Secure.Payment.Confirmation.pdf)

In above proposal, a mini proposal is to open up authentication to cross origin without iframes. Opening up this issue to hopefully have the concentrated conversation around cross origin authentication without iframes and want to talk about security, privacy and user experience and potential misuse.  

So at the moment, we don't allow authentication outside of origin in webauthn. We allowed it in iframes in L2, but don't allow it for any other website to be able to invoke .get() call credentials without iframes. 

Recently, we discussed to open up .create() call inside iframes. Till this point RP is in control. If it chooses to, like all of non-payment scenarios, it can choose to not enable these calls in iframes. And if they do, they know who the parent is so that they can potentially filter out origins that they trust. 

I see three kind of RPs with different levels of control. 

1. RP who never wants to do .create() or .get() calls outside of their origin. Current webauthn spec protections allow that which really helps in phishing resistant property of WebAuthn. 
    - Most RPs live in this space
2. RP who is OK with iframes but only want to do in iframes because they are still in control of who can invoke webauthn ceremonies. 
3. RP who wants .create() and .get() calls outside of origin. Like merchants, who wants to do on-behalf of banks. 
   - The above Web Payments proposal is not proposing .create() call outside of iframes at the moment. But I heard @christiaanbrand saying that he wants to do that in future. 

The most concern I have is around the third category of RPs/Scenarios which will negatively impact Category  1 and also category 2. 

IIUC, the proposal is to only allow credentials who are marked as SCP compliant to be opened up for now for authentication in cross origin case. 
- I don't understand, how these credentials will be marked as such and be temper proof by the calling website. I think the proposal is that merchant will have all the credentialIDs and will invoke it .get() call with a new field set which says that this credential is marked as SCP compliant.
- What is the cryptographic proof that says that this credential is created with SCP in mind? How will browser determine that one credential is allowed vs another credential is not allowed. 

If we cannot determine this without tampering, then we have privacy issues. A phishing website will get the credentialID and invoke it in their origin, user will be prompted with the dialog, which most likely they don't much read, and user approves and website knows that user has credential for a website which they have no right to enquire. 

In above context, let's get one thing out of the way. CredentialIDs are public and that's why we don't store PII information in that. In webauthn designs, we allow with username flows, where given a username, credentialIDs being returned for passwordless authentication silently. Ofcourse, RP can add fake one, but they need to also provide the real ones.  Even outside webauthn, credentialIDs are kind of treated as semi public info. And till now this was not an issue because no other origin can really use them. So credentialIDs will be known to many parties. 

Another point I heard was around it is only applicable for non-discoverable credentials. I don't understand why. I can understand that this design only works if used in allowlist context, which can be satisfied by both non-discoverable as well discoverable credentials. And restricing it to only with allowlist, I am not sure what we are getting. As mentioned above, credential IDs are not really protected in any full secure way by the RPs. 

Third argument I heard that any websites ability to popup authentication, even if we restrict it SCP credentials only is only a nuisance. I argue that it is a privacy issue and a possibly a phishing vector. It gives ability for a random RP to almost know for certainty that a credential exist for a website. And that is why, the second category above which says, I have cases around cross-origin authentication, but I only want it to be inside an iframe. And that should be RP's choice. Even if this is nuisance, it seriously hampers the webauthn brand.  

Till now, I am only thinking from RP's persctive. I haven't thought through, where the user choice in all of this. 

Overall, it seems that this is dangerous and in pursuit of solving one desire to not have iframes (which I don't disagree with per say, but just recently we rejected other legitimate use cases around raw signatures), we are messing up with core fundamentals too much here. Ofcourse, I don't have the full design and I may be wrong here. 

@agl @christiaanbrand @equalsJeffH @ve7jtb  for above questions and clarifications if I misunderstood things.

### LINKED ISSUE #1687: Incorrect "to create" phrase used in get() section's introduction?
There these two paragraphs in section [5.1.4. Use an Existing Credential to **Make an Assertion** - PublicKeyCredential’s \[\[Get\]\]\(options) Method](https://www.w3.org/TR/2021/REC-webauthn-2-20210408/#sctn-getAssertion);
> The get() implementation [CREDENTIAL-MANAGEMENT-1] calls PublicKeyCredential.\[\[CollectFromCredentialStore]]() to collect any credentials that should be available without user mediation (roughly, this specification’s authorization gesture), and if it does not find exactly one of those, it then calls PublicKeyCredential.\[\[DiscoverFromExternalSource]]() to have the user select a credential source.

> Since this specification requires an authorization gesture **to create** any credentials, the PublicKeyCredential.[[CollectFromCredentialStore]](origin, options, sameOriginWithAncestors) internal method inherits the default behavior of Credential.\[\[CollectFromCredentialStore]](), of returning an empty set.  [emphasis added]

When I recently re-read that section of the spec and encountered that "to create" phrase, I immediately had these questions: 

Why is the phrase "to create" used here?  Ought it be "to use" or "to exercise" ?  

I.e., I do not understand how a gesture at cred creation time would factor in here?

emlun:
Fixes #1687.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/pull/1693.html" title="Last updated on Jan 26, 2022, 4:45 PM UTC (36c7f9b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/1693/bc87207...36c7f9b.html" title="Last updated on Jan 26, 2022, 4:45 PM UTC (36c7f9b)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1961,7 +1961,7 @@ to indicate what [=credential sources=] are acceptable to it. The [=client platf
 matching the specified criteria, and guides the user to pick one that the script will be allowed to use. The user may choose to
 decline the entire interaction even if a [=credential source=] is present, for example to maintain privacy. If the user picks a
 [=credential source=], the user agent then uses
-[[#sctn-op-get-assertion]] to sign a [=[RP]=]-provided challenge and other collected data into an assertion, which is used as a
+[[#sctn-op-get-assertion]] to sign a [=[RP]=]-provided challenge and other collected data into an [=authentication assertion=], which is used as a
 [=credential=].
 
 The {{CredentialsContainer/get()}} implementation [[!CREDENTIAL-MANAGEMENT-1]] calls
@@ -1970,7 +1970,7 @@ should be available without [=user mediation=] (roughly, this specification's [=
 exactly one of those, it then calls <code>PublicKeyCredential.{{PublicKeyCredential/[[DiscoverFromExternalSource]]()}}</code> to have
 the user select a [=credential source=].
 
-Since this specification requires an [=authorization gesture=] to create any [=credentials=], the <code>PublicKeyCredential.<dfn
+Since this specification requires an [=authorization gesture=] to create any [=assertions=], the <code>PublicKeyCredential.<dfn
 for="PublicKeyCredential" method>\[[CollectFromCredentialStore]](origin, options, sameOriginWithAncestors)</dfn></code> [=internal method=] inherits the default behavior of
 {{Credential/[[CollectFromCredentialStore]]()|Credential.[[CollectFromCredentialStore]]()}}, of returning an empty set.
 


equalsJeffH (approved):
LGTM, thx @emlun !

ve7jtb (approved):
lgtm

Issue #1665, opened 2021-08-24
Synced/multi-device user Credentials


equalsJeffH:
Issue #1637 introduces possible experiences in a future WebAuthn, various aspects of which are enabled by "syncing platform credentials" via platform providers' sync fabrics. The spec will need updating to explicitly accommodate and explain the "synced credential" notion. This will accommodate other spec updates, e.g. PR #1663, which need to make reference to the "synced credential" notion. 

Note that the spec mostly _implies_ that [public key credential sources](https://www.w3.org/TR/webauthn/#public-key-credential-source) are (presently) hardware-bound by default (though the Cred Private Key definition is explicit,  see [below](https://github.com/w3c/webauthn/issues/1665#issuecomment-904781736)). These various notions will all need to be made explicit.

[ Note: synced credentials are a crucial enabler of the "passkey" concept. ]

emlun:
The [Credential Private Key](https://www.w3.org/TR/2021/REC-webauthn-2-20210408/#credential-private-key) definition will probably need a tweak too:

>[...]
>The credential private key is bound to a particular authenticator - its managing authenticator - and is expected to never be exposed to any other party, not even to the owner of the authenticator.
>[...]

equalsJeffH:
For associated discussion on terminology, see also:

* https://github.com/w3c/webauthn/issues/1691#issuecomment-1019036337
* https://github.com/w3c/webauthn/issues/1691#issuecomment-1019295694

For associated discussion regarding the webauthn spec sections that (likely amongst others) will require updating, see also:

* https://github.com/w3c/webauthn/issues/1691#issuecomment-1019514936

timcappalli:
We should avoid using the term "sync" in the WebAuthn spec. Sync is one specific mechanism of moving/copying/migrating credentials. Other terms include:

- "copy"
- "move"
- "backup"
- "export"
- "peer-to-peer transfer"

We're really talking about "multi-device WebAuthn credentials" and "single-device WebAuthn credentials". I would propose using these two phrases instead.

MasterKale:
> We're really talking about "multi-device WebAuthn credentials" and "single-device WebAuthn credentials". I would propose using these two phrases instead.

Can the spec include mention of passkeys, even if it's just in Terminology? If we can't use "passkeys" directly in the spec then I think it'd be beneficial to help readers understand the following:

- "multi-device WebAuthn credential" === "passkey"
- "single-device WebAuthn credential" === "single-device passkey"

[...]

PR #1663, opened 2021-08-21
device public key extension

equalsJeffH:
The resolves #1658 by defining the `devicePubKey` extension et al.  It is admittedly rough and will need further work, thus am casting it as a "draft" PR.

update 4-Mar-2022: @ve7jtb has submitted issue https://github.com/w3c/webauthn/issues/1701 --- this PR needs to be updated to address it.

update 19-Mar-2022:  commit f0fe8f2 is a rough start at adding an authenticator-generated nonce to `attObjForDevicePublicKey`: fixes #1701 

update 23-Mar-2022:  there's now commits beyond f0fe8f2 attempting to further refine the RP usage and extension output verification procedures. Though, see also issue #1711 and https://github.com/w3c/webauthn/pull/1663#issuecomment-1077868106: **issue #1711 really needs to be addressed** as a part of the `devicePubKey` effort.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/pull/1663.html" title="Last updated on Oct 7, 2022, 7:34 PM UTC (6ae32a0)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/1663/8a6daec...6ae32a0.html" title="Last updated on Oct 7, 2022, 7:34 PM UTC (6ae32a0)">Diff</a>

PR #1695, opened 2022-02-08
backup states in authenticator data

### LINKED ISSUE #1692: Backup state of credentials
With recent multi-device passkeys concept for backup and recovery purposes and the desire to remove passwords from the account, we need to provide signal to the RP that it can go ahead and remove passwords. 

This is a refinement of original proposal of "durable" flags in https://github.com/w3c/webauthn/wiki/Explainer:-broadening-the-user-base-of-WebAuthn#durable-signal 
(which is derived from: https://github.com/w3c/webauthn/issues/1637#issue-935187400)

`dpk`, as per current definition in the WebAuthn PR https://github.com/w3c/webauthn/pull/1663 is that it is just another key which is device bound. It does not tell whether primary credential is device bound or not. 

Windows current thinking is that if and when we implement multi-device backup, user will be given a choice of whether they want these credentials to be backed-up or not at registration time. User may also have ability to choose sync/backup state in the future depending on various factors like access to sync/backup fabric etc. and it's properties. Which means that backup state of the credential can change over time. And RP should be able to determine what is happening from the authenticator responses. 

There are two possible solution around this. 
1. ### Extension: "backup-state"
    - The extension tells RP about the backup state the credential.
    - ### Need
        - Existing devices which does not support backup and newer devices which may support backup needs to coexist.
        - RP also needs to distinguish between device bound key vs non-device bound key in certain situations.
        - RP needs to know in which state the credential is in so that it can offer removal of password from the account.
        - User will be offered choice on certain platforms and authenticators where user may transition from unbacked-up state to backup-state or vice-versa.
        - Hence the value is dynamic in nature if backup is supported by the authenticator.
    - ### Applicability
        - Both at credential.create() as well credential.get()
    - ### Values
        - BACKUP_NOT_APPLICABLE
           - Credential created is device bound and does not support backup now or in future.
           - All existing credentials created are in this state.
           - Device-public-key extension is not returned in this state.
           - Default state if extension is not returned
        - BACKUP_NOT_CONFIGURED
           - Credential is not configured to be backed-up at this time but may be configured in future.
           - Device-public-key extension should be returned in this state.
        - BACKUP_IN_PROGRESS
           - Device-public-key extension should be returned in this state.
           - Typically in credential.create() stage where credential has just been created and not yet synced but should be synced shortly.
        - BACKUP_COMPLETED
           - Device-public-key extension should be returned in this state.

2. ### Authenticator Flags (as proposed by @ve7jtb)
    - Single Device 0 / multi-device capable 1
    - Not backed up 0 / backed up 1

cc @agl / @equalsJeffH / @timcappalli

timcappalli:
This PR adds backup state bits to authenticator data as described in #1692

WIP

fixes #1692
improves #1665

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/pull/1695.html" title="Last updated on Apr 22, 2022, 1:34 PM UTC (03b829e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/1695/dcd3190...03b829e.html" title="Last updated on Apr 22, 2022, 1:34 PM UTC (03b829e)">Diff</a>


index.bs
```diff
@@ -3657,7 +3664,86 @@ This is because the first 37 bytes of the signed data in a FIDO U2F authenticati
...
+        [=Authenticator data=] bits 3 and 4 combinations 
+    </figcaption>
+</figure>
+
+It is recommmended that [=Relying Party|Relying Parties=] store the last value of these bits with the user account for future evaluation.
+
+The following is a non-normative, non-exhaustive list of how [=Relying Party|Relying Parties=] may utilize these bits:
+
+    - Requiring additional [=authenticators=]:
+        When bit 3 is set to `0`, the [=authenticator=] will never allow the credential to transition from a single-device credential to
```
MasterKale, line 3721: It seems weird to me that earlier it's claimed that "the values of bits 3 and 4 may change over time...", but then here it's stated that, "when bit 3 is set to `0`, the authenticator will **never** allow the credential to transition...". But by saying that bit 3 can change over time then _technically_ when bit 3 is `0` it just means that the credential is not allowed to be backed up _right now_, right?

index.bs
```diff
@@ -3544,7 +3545,13 @@ laid out as shown in <a href="#table-authData">Table <span class="table-ref-foll
                 - Bit 2: [=User Verified=] ([=UV=]) result.
                     - `1` means the user is [=user verified|verified=].
                     - `0` means the user is not [=user verified|verified=].
-                - Bits 3-5: Reserved for future use (`RFU2`).
+                - Bit 3:
+                    - `1` means the [=Public Key Credential Source|credential source=] is allowed to be backed up
```
agl, line 3549: This formulation leaves an "impossible" state: xxx10xxx would mean that it _is_ backed up, but it's not allowed to be. I liked Akshay's four states:

00 &rarr; will never be backed up
01 &rarr; could be backed up, but backup is not currently happening.
10 &rarr; will soon be backed up.
11 &rarr; is backed up.

This text merges the 01 and 10 states, but I think it's worth having the difference since we're consuming limited flag space.

index.bs
timcappalli: fixed here: https://github.com/w3c/webauthn/pull/1695/commits/aa68aba8b656ca9bb1595f9b24408410949ff945

index.bs
akshayku: We can go either way on this.

Couple of reasons for current PR proposal
- Till now all the bits had individual meanings. And @ve7jtb proposal in #1692 seemed cleaner. PR just refines the wording of it. 
    - Single Device 0 / multi-device capable 1
    - Not backed up 0 / backed up 1
- "will soon be backed up" and "could be backed up, but backup is not currently configured." have same action items for a relying party. So "will soon be backed up" seemed unnecessary.

timcappalli:
TODO: Add validation process to section 7.2

MasterKale:
@timcappalli Thank you for kicking off this effort to make use of these extra bits. We here at Duo (and the greater Cisco) are eager for the extra insight into passkey utilization as we continue to champion WebAuthn adoption for passwordless auth. Once passkeys launch it's just a matter of time before customers start asking us for greater control over the use of passkeys by their users. If we can have the data in these bits on Day 1 RPs like us can better adapt our product to respond to these requests, instilling confidence in our users that WebAuthn is the preferred way to achieve a more secure authentication experience.

index.bs
```diff
@@ -3657,7 +3664,91 @@ This is because the first 37 bytes of the signed data in a FIDO U2F authenticati
 <code>[=authDataExtensions|extensions=]</code> are never present. FIDO U2F authentication signatures can therefore be verified by
 the same procedure as other [=assertion signatures=] generated by the [=authenticatorGetAssertion=] operation.
 
+### Credential Backup State ### {#sctn-credential-backup}
 
+Credential backup eligibility and current backup state is conveyed by [=Authenticator data|authenticator data=] bits `3` and `4` as 
+defined in <a href="#table-authData">Table <span class="table-ref-following"/></a>.
+
+The value of bit 3 is set during [=authenticatorMakeCredential=] operation and MUST NOT change.
```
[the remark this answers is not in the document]
timcappalli: > It would be good to elaborate on or refer to a section where it is explained how the authenticator determines the value of that bit in authenticatorMakeCredential.
> For example, refer to DPK extension, or similar.

@rlin1 the authenticator is ultimately responsible for the lifecycle management and state of the credential. I'm not sure how we would put the internal business logic of an authenticator into the spec. Do you have any suggested text?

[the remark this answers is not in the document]
rlin1: I was not referring to business logic.  But in PR#1663 it is kind of obvious that the RP could add the DPK extension to a getAssertion call in order to ask for a non-synchronizable key.
In this Issue 1692 - when using the flags approach it is not clear (to me) how the RP could ask for a singleDeviceKey.
So we should add an explanation to this PR1695 how the RP could do that - e.g. using the  credProps or using some extension in the request - or clarify that the RP cannot (if that is the intention)

index.bs
```diff
@@ -3657,7 +3664,86 @@ This is because the first 37 bytes of the signed data in a FIDO U2F authenticati
...
+            <td>`0`</td>
+            <td>
+                The credential is a single-device credential and cannot become a multi-device credential
+            </td>
+        </tr>
+        <tr>
+            <td>`0`</td>
+            <td>`1`</td>
+            <td>
+                This combination is not allowed
```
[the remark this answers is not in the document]
Firstyear: The fact that the bits can "change over time" is a red flag, since authenticator data is only sent at registration, not auth, which means when the bits change, the RP won't know they changed.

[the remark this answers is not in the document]
timcappalli: > The fact that the bits can "change over time" is a red flag, since authenticator data is only sent at registration, not auth, which means when the bits change, the RP won't know they changed. 

authData is sent during a getAssertion and the RP can evaluate any change to backup state.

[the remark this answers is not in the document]
Firstyear: Yes, my mistake, I thought this was in the attestation. 

In this case, I think the issue is that there is no advie from the standard that an RP should persist the authdata flags to detect changes (this has previously caused CVE's related UV anyway ...)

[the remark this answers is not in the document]
timcappalli: > Yes, my mistake, I thought this was in the attestation.
> 
> In this case, I think the issue is that there is no advie from the standard that an RP should persist the authdata flags to detect changes (this has previously caused CVE's related UV anyway ...)

Covered in section 7 with the other verification procedures.

index.bs
```diff
@@ -4747,6 +4855,9 @@ a numbered step. If outdented, it (today) is rendered as a bullet in the midst o
 1. If [=user verification=] is required for this assertion, verify that the [=User Verified=] bit of the <code>[=flags=]</code> in
     |authData| is set.
 
+1. If the credential [=backup state=] is used as part of Relying Party business logic or policy, compare the previously stored
```
Firstyear, line 4858: There needs to be advice on what do if the value has changed. Or that at least if the value changes then the RP should make a reasonably policy decision.

timcappalli: The use cases are defined. We are not going to make recommendations beyond that. These are hints, not security properties.

Firstyear: There is no reference to the word "hint" in this change at all. This means that implementors will read this and WILL believe it IS a security property given it's in a signed part of the attestation/authentication.  This change is completely unclear that it's a hint. This needs correction.

Firstyear: PS: There is already evidence that the "hint" about authenticator attachement and UV "hints" have already be assumed to be security properties by RP's which has led to CVE's, so there is established precedent that we MUST be clear about what is a hint, and what is a security property, and this change is currently not achieving that.

timcappalli: Sounds like your concern is bigger than this PR. Please create a new issue with your feedback or a PR.

Firstyear: No, my concern absolutely matters to this PR. If you are asserting this change is a "hint" and not a security property, then say that explicitly in this PR.

index.bs
```diff
@@ -4747,6 +4855,9 @@ a numbered step. If outdented, it (today) is rendered as a bullet in the midst o
 1. If [=user verification=] is required for this assertion, verify that the [=User Verified=] bit of the <code>[=flags=]</code> in
     |authData| is set.
 
+1. If the credential [=backup state=] is used as part of Relying Party business logic or policy, compare the previously stored 
+    value with the [=backup state=] bit of the <code>[=flags=]</code> in |authData|, perform evaluation, and then store the new value.
```
emlun, line 4859:
```suggestion
1. If the credential [=backup state=] is used as part of [=[RP]=] business logic or policy, compare the previously stored 
    value with the [=backup state=] bit of the <code>[=flags=]</code> in |authData|, perform evaluation, and then store the new value.
```

emlun (changes requested):
Here's a mixed bag of editorial fixes, wordsmithing ideas and open questions... :slightly_smiling_face:

index.bs
equalsJeffH: I think `{#sctn-credential-backup}}` should be `[[#sctn-credential-backup]]` ?

index.bs
```diff
@@ -3544,7 +3555,13 @@ laid out as shown in <a href="#table-authData">Table <span class="table-ref-foll
...
                     - `1` means the user is [=user verified|verified=].
                     - `0` means the user is not [=user verified|verified=].
-                - Bits 3-5: Reserved for future use (`RFU2`).
+                - Bit 3: [=Backup Eligibility=].
+                    - `1` means the [=Public Key Credential Source|credential source=] is allowed to be backed up 
+                    - `0` means the [=Public Key Credential Source|credential source=] is not allowed to be backed up
+                - Bit 4: [=Backup State=].
+                    - `1` means the [=Public Key Credential Source|credential source=] is currently backed up
+                    - `0` means the [=Public Key Credential Source|credential source=] is not currently backed up
+                - Bits 5: Reserved for future use (`RFU2`).
```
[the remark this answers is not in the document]
equalsJeffH: I would also make these changes:
s/ [=Public Key Credential Source|credential source=] / [=public key credential source=] /
s/ backed up / [=backed up=] /

index.bs
```diff
@@ -3657,7 +3674,91 @@ This is because the first 37 bytes of the signed data in a FIDO U2F authenticati
...
 
+### Credential Backup State ### {#sctn-credential-backup}
+
+Credential backup eligibility and current backup state is conveyed by [=Authenticator data|authenticator data=] bits `3` and `4` as 
+defined in <a href="#table-authData">Table <span class="table-ref-following"/></a>.
+
+The value of bit 3 is set during [=authenticatorMakeCredential=] operation and MUST NOT change. 
+
+The value of bit 4 may change over time based on the current state of the [=Public Key Credential Source|credential source=]. <a href="#table-backupStates">Table <span class="table-ref-following"/></a> below defines
+valid combinations and their meaning.
```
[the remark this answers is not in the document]
equalsJeffH: I would also make these changes:
s/ backup eligibility / [=backup eligibility=] /
s/ backup state / [=backup state=] /

index.bs
```diff
@@ -3657,7 +3674,91 @@ This is because the first 37 bytes of the signed data in a FIDO U2F authenticati
...
+        A single-device credential is not durable and cannot survive single device loss. [=Relying Party|Relying Parties=]
+        should ensure that an account has additional [=authenticators=] enrolled and/or an account recovery process in place.
+
+        For example, the user could be prompted to set up a single-device credential on a [=roaming authenticator=], like a 
+        security key, or another [=authenticator=] that is capable of holding multi-device credentials.
+    
+    - Upgrading a user to a password-free account:
+
+        When bit 4 changes from `0` to `1`, the [=authenticator=] is signaling that the [=Public Key Credential Source|credential source=]
+        is durable (it has been backed up and is protected from single device loss). This is often referred to as a multi-device credential.
```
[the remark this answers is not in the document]
equalsJeffH: agreed, this definitional conflict needs to be sorted out.

index.bs
equalsJeffH: Also, up on line 3728 we recommend the RP storing both flags/bits, but here in the RP ops sections we are recommending storing only the "backup state" and not "backup eligibility".

index.bs
```diff
@@ -4580,6 +4681,13 @@ In order to perform a [=registration ceremony=], the [=[RP]=] MUST proceed as fo
 1. If [=user verification=] is required for this registration, verify that the [=User Verified=] bit of the <code>[=flags=]</code>
     in |authData| is set.
 
+1. If the credential [=backup eligibility=] is used as part of Relying Party business logic or policy, evaluate the 
+    [=backup eligibility=] bit of the <code>[=flags=]</code> in |authData|.
```
equalsJeffH, line 4685: In addition to @emlun's comments, I note we recommend (way back up on line 3728) the RP storing both backup flags/bits, but here in the RP ops sections we are recommending storing only the "backup state" and not "backup eligibility".  An RP may wish to store both values, e.g., to inform their UX flows?  (use of various forms of the term "... flow(s)" in various places in the spec is established)

Additionally, by "business logic" do you mean UX flows?  We have not used the term "business logic" in the spec as yet. these cred backup-driven additions to the RP ops are the only places to do so.  

Perhaps:
```suggestion
1. If the [=[RP]=] uses the credential's [=backup eligibility=] to inform its user experience flows and/or policies, evaluate the 
    [=backup eligibility=] bit of the <code>[=flags=]</code> in |authData|.
```

index.bs
```diff
@@ -4580,6 +4681,13 @@ In order to perform a [=registration ceremony=], the [=[RP]=] MUST proceed as fo
...
+
+1. If the credential [=backup state=] is used as part of  Relying Party business logic or policy, evaluate the [=backup state=] 
+    bit of the <code>[=flags=]</code> in |authData|, and then store the value for evaluation in future 
+    [=authentication ceremony|authentication ceremonies=].
```
equalsJeffH, line 4689: similar comment here as the one on lines +4684 to +4685

equalsJeffH (dismissed):
Here's some comments/suggestions building on @emlun's review... :smile:

index.bs
Firstyear: This still doesn't resolve that it is a *HINT* and not a strict property of the device. If this bit is a hint we must call it as such and call it "backup state hint". Otherwise people will assume it is a security property.

index.bs
timcappalli: Business logic is the traditional way to refer to this, but I have no issue with your proposed text.

index.bs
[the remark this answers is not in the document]
timcappalli: I used Jeff's proposed text. They convey the same thing.

index.bs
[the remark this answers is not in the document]
timcappalli: I used Jeff's proposed text. They convey the same thing.

index.bs
```diff
@@ -3657,7 +3674,91 @@ This is because the first 37 bytes of the signed data in a FIDO U2F authenticati
...
+            <td>`1`</td>
+            <td>
+                The credential is a multi-device credential
+            </td>
+        </tr>
+    </table>
+    <figcaption>
+        [=Authenticator data=] bits 3 and 4 combinations 
+    </figcaption>
+</figure>
```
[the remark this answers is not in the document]
timcappalli: Added. thanks!

index.bs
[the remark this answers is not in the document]
timcappalli: @emlun are you saying I should add "([=BE=])" and "([=BS=])" after each name, similar to UP and UV?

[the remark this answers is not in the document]
emlun: Not in this particular review thread, but in [another thread](https://github.com/w3c/webauthn/pull/1695#discussion_r823685874), yes.

index.bs
emlun: @timcappalli Could you elaborate on what you mean by the statement "These are hints, not security properties"? Apart from the provision that the state bit may change over time, I don't see any room for the flags to be vague. In what way are they hints (non-compliant and unattested authenticators notwithstanding)?

Firstyear: My external assumption related to this situation is that apple passkeys already exist and can be tested and *do not* provide these flags. So the absence or presence of this bit may not be a true reflection of the state (IE apple passkeys don't set the bit today, but are multi device).

index.bs
```diff
@@ -3555,10 +3569,10 @@ laid out as shown in <a href="#table-authData">Table <span class="table-ref-foll
                 - Bit 2: [=User Verified=] ([=UV=]) result.
                     - `1` means the user is [=user verified|verified=].
                     - `0` means the user is not [=user verified|verified=].
-                - Bit 3: [=Backup Eligibility=].
+                - Bit 3: [=Backup Eligibility=] ([=BE=].
```
emlun, line 3572: :slightly_smiling_face:
```suggestion
                - Bit 3: [=Backup Eligibility=] ([=BE=]).
```

index.bs
```diff
@@ -3555,10 +3569,10 @@ laid out as shown in <a href="#table-authData">Table <span class="table-ref-foll
...
                     - `1` means the [=Public Key Credential Source|credential source=] is allowed to be backed up 
                     - `0` means the [=Public Key Credential Source|credential source=] is not allowed to be backed up
-                - Bit 4: [=Backup State=].
+                - Bit 4: [=Backup State=] ([=BS=].
```
emlun, line 3575:
```suggestion
                - Bit 4: [=Backup State=] ([=BS=]).
```

index.bs
```diff
@@ -3676,85 +3690,85 @@ the same procedure as other [=assertion signatures=] generated by the [=authenti
...
-                This combination is not allowed
+                This combination is not allowed.
             </td>
         </tr>
         <tr>
             <td>`1`</td>
             <td>`0`</td>
             <td>
-                The credential is currently a single-device credential but may become a multi-device credential in the future
+                The credential is currently a [=single-device credential=] but may become a [=multi-device credential=] in the future.
```
emlun, line 3726: This is still inconsistent with the Backup Eligibility definition: single- vs. multi-device-ness is defined there to be an immutable property of a credential.

index.bs
```diff
@@ -3676,85 +3690,85 @@ the same procedure as other [=assertion signatures=] generated by the [=authenti
...
 
-        When bit 3 is set to `0`, the [=authenticator=] will never allow the credential to transition from a single-device credential to
-        a multi-device credential. 
+        When `BE` [=flag=] is set to `0`, the [=generating authenticator=] will never allow the credential to transition from a 
+        [=single-device credential=] to a [=multi-device credential=]. 
         
-        A single-device credential is not durable and cannot survive single device loss. [=Relying Party|Relying Parties=]
-        should ensure that an account has additional [=authenticators=] enrolled and/or an account recovery process in place.
+        A [=single-device credential=] is not durable and cannot survive single device loss. [=[RPS]=] should ensure that a [=user account=] 
+        has additional [=authenticators=] [=registration ceremony|registered=] and/or an account recovery process in place.
```
emlun, line 3752: This "should" looks normative, right?
```suggestion
        A [=single-device credential=] is not durable and cannot survive single device loss. [=[RPS]=] SHOULD ensure that a [=user account=] 
        has additional [=authenticators=] [=registration ceremony|registered=] and/or an account recovery process in place.
```

emlun, line 3752: We haven't defined what "durable" is supposed to mean in this context, and I've seen the term frowned upon as it implies that non-syncing authenticators are "not durable". I think it's better to slightly rephrase this.

```suggestion
        A [=single-device credential=] is not resilient to single device loss. [=[RPS]=] should ensure that a [=user account=] 
        has additional [=authenticators=] [=registration ceremony|registered=] and/or an account recovery process in place.
```

index.bs
```diff
@@ -3676,85 +3690,85 @@ the same procedure as other [=assertion signatures=] generated by the [=authenti
...
-        security key, or another [=authenticator=] that is capable of holding multi-device credentials.
+        For example, the user could be prompted to set up an additional [=authenticator=], such as a [=roaming authenticator=] or an 
+        [=authenticator=] that is capable of [=multi-device credentials=].
     
     - Upgrading a user to a password-free account:
 
-        When bit 4 changes from `0` to `1`, the [=authenticator=] is signaling that the [=Public Key Credential Source|credential source=]
-        is durable (it has been backed up and is protected from single device loss). This is often referred to as a multi-device credential.
+        When the `BS` [=flag=] changes from `0` to `1`, the [=authenticator=] is signaling that the [=credential=] is backed up and is protected from single device loss. 
+        This is often referred to as a [=multi-device credential=]
```
emlun, line 3760: This is inconsistent with the Backup Eligibility definition: the credential is already a _multi-device credential_ even when the BS flag is still 0.

index.bs
```diff
@@ -4681,12 +4695,11 @@ In order to perform a [=registration ceremony=], the [=[RP]=] MUST proceed as fo
 1. If [=user verification=] is required for this registration, verify that the [=User Verified=] bit of the <code>[=flags=]</code>
     in |authData| is set.
 
-1. If the credential [=backup eligibility=] is used as part of Relying Party business logic or policy, evaluate the 
+1. If the [=[RP]=] uses the credential's [=backup eligibility=] to inform its user experience flows and/or policies, evaluate the 
     [=backup eligibility=] bit of the <code>[=flags=]</code> in |authData|.
```
emlun, line 4699: Adding the abbreviated name here makes the connection to the flags table clearer.

```suggestion
    [=backup eligibility=] (BE) bit of the <code>[=flags=]</code> in |authData|.
```

index.bs
```diff
@@ -4681,12 +4695,11 @@ In order to perform a [=registration ceremony=], the [=[RP]=] MUST proceed as fo
...
 
-1. If the credential [=backup state=] is used as part of  Relying Party business logic or policy, evaluate the [=backup state=] 
-    bit of the <code>[=flags=]</code> in |authData|, and then store the value for evaluation in future 
-    [=authentication ceremony|authentication ceremonies=].
+1. If the [=[RP]=] uses the credential's [=backup state=] to inform its user experience flows and/or policies, evaluate the [=backup state=] 
+    bit of the <code>[=flags=]</code> in |authData|, and then store the value for evaluation in future [=authentication ceremony|authentication ceremonies=].
```
emlun, line 4702: Adding the abbreviated name here makes the connection to the flags table clearer. Also, Bikeshed can automatically singularize the "authentication ceremonies" term.

```suggestion
1. If the [=[RP]=] uses the credential's [=backup state=] to inform its user experience flows and/or policies, evaluate the [=backup state=] (BS)
    bit of the <code>[=flags=]</code> in |authData|, and then store the value for evaluation in future [=authentication ceremonies=].
```

index.bs
```diff
@@ -3676,85 +3690,85 @@ the same procedure as other [=assertion signatures=] generated by the [=authenti
...
     - Adding an additional factor after a state change:
 
-        When bit 4 changes from `1` to `0`, the [=authenticator=] is signaling that the [=Public Key Credential Source|credential source=]
-        is no longer durable (not backed up and is not protected from single device loss). This could be the result of the user deleting the 
-        credential, deleting a backup account, an issue with the backup service, or another undefined error condition.
+        When the `BS` [=flag=] changes from `1` to `0`, the [=authenticator=] is signaling that the [=credential=] is no longer backed up, 
+        and no longer protected from single device loss. This could be the result of the user actions, such as disabling the backup service, 
+        or errors, such as issues with the backup service.
 
         When this transition occurs, the credential is no longer durable and a [=Relying Party=] should guide the user through a process to
```
emlun, line 3770: This SHOULD looks normative, right?

```suggestion
        When this transition occurs, the credential is no longer durable and a [=Relying Party=] SHOULD guide the user through a process to 
```

index.bs
```diff
@@ -3676,85 +3690,85 @@ the same procedure as other [=assertion signatures=] generated by the [=authenti
...
-        When bit 4 changes from `1` to `0`, the [=authenticator=] is signaling that the [=Public Key Credential Source|credential source=]
-        is no longer durable (not backed up and is not protected from single device loss). This could be the result of the user deleting the 
-        credential, deleting a backup account, an issue with the backup service, or another undefined error condition.
+        When the `BS` [=flag=] changes from `1` to `0`, the [=authenticator=] is signaling that the [=credential=] is no longer backed up, 
+        and no longer protected from single device loss. This could be the result of the user actions, such as disabling the backup service, 
+        or errors, such as issues with the backup service.
 
         When this transition occurs, the credential is no longer durable and a [=Relying Party=] should guide the user through a process to 
-        validate their other sign in factors. If the user does not have another durable credential for their account, they should be guided 
+        validate their other sign in factors. If the user does not have another credential for their account, they should be guided
```
emlun, line 3771: This SHOULD looks normative, right?

```suggestion
        validate their other sign in factors. If the user does not have another credential for their account, they SHOULD be guided 
```

index.bs
emlun, line 3770: Same thing about "durable", and we've already stated that the credential is no longer backed up.

```suggestion
        When this transition occurs, the [=Relying Party=] should guide the user through a process to 
```

index.bs
```diff
@@ -3676,85 +3690,85 @@ the same procedure as other [=assertion signatures=] generated by the [=authenti
...
+        When the `BS` [=flag=] changes from `1` to `0`, the [=authenticator=] is signaling that the [=credential=] is no longer backed up, 
+        and no longer protected from single device loss. This could be the result of the user actions, such as disabling the backup service, 
+        or errors, such as issues with the backup service.
 
         When this transition occurs, the credential is no longer durable and a [=Relying Party=] should guide the user through a process to 
-        validate their other sign in factors. If the user does not have another durable credential for their account, they should be guided 
+        validate their other sign in factors. If the user does not have another credential for their account, they should be guided 
         through adding an additional authentication factor to ensure they do not lose access to their account. An example would be prompting
         the user to set up a single-device credential on a [=roaming authenticator=], like a security key, or another [=authenticator=] that 
         is capable of holding multi-device credentials.
```
emlun, line 3774: Looks like I missed this. This should probably use the same wording as above.

```suggestion
        through adding an additional authentication factor to ensure they do not lose access to their account.
        For example, the user could be prompted to set up an additional [=authenticator=], such as a [=roaming authenticator=] or an 
        [=authenticator=] that is capable of [=multi-device credentials=].
```

index.bs
```diff
@@ -972,15 +972,23 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
...
+
 : <dfn>Backup Eligibility</dfn>
-:: The [=managing authenticator=] controls whether a [=Public Key Credential Source|credential source=] is allowed to be backed up.
-   A credential that is eligible for backup is referred to as a <dfn>multi-device credential</dfn> whereas a credential that is not
-   eligible for backup is referred to as a <dfn>single-device credential</dfn>. Backup can occur via mechanisms including but not 
-   limited to peer-to-peer sync, cloud sync, local network sync, and manual import/export.
+: <dfn>Backup Eligible</dfn>
+:: A [=Public Key Credential Source=]'s [=generating authenticator=] determines at creation time whether the [=public key credential source=] 
+   is allowed to be [=backed up=]. [=Backup eligibility=] is signaled in [=authenticator data=]'s [=flags=] along with the current [=backup state=]. 
+   [=Backup eligibility=] is a [=credential property=] and is permanent for a given [=public key credential source=].A [=backup eligible=] [=public key credential source=]
```
emlun, line 985: Missing period, might as well make it a line break while we're at it.
```suggestion
   [=Backup eligibility=] is a [=credential property=] and is permanent for a given [=public key credential source=].
   A [=backup eligible=] [=public key credential source=] 
```

emlun (changes requested):
Thanks for the edits and patience! Here's some more...

index.bs
Firstyear: Why did you make this resolved? You still haven't clarified if this is a hint or a security property? You say it's a hint, but then then spec is written like it's a hard security property. What is it?

Firstyear (changes requested):
You need to clarify, from your own words, if this is a hint or a security property.

index.bs
emlun: I'm guessing that the contention is that they are "hard security properties", but under the assumption that the authenticator is compliant. Without attestation there is no guarantee that the authenticator is compliant - and there might not be with attestation either (e.g., Apple passkeys today are not compliant with this proposal, but remember this is still a long time away from proposed standard maturity) - so in that sense they are "hints" because you can't know for sure unless the attestation tells you the authenticator is compliant and trustworthy. So which of the two it is depends only on how much you trust or don't trust the authenticator.

But again, the above is true for the whole spec, not just these flags. The spec generally assumes that authenticators and clients are compliant, but there's no guarantee that they are. Attestation is the only way to get any firm guarantees; without it the only thing you can be completely sure of is that valid assertion signatures can only be generated by an authenticator with possession of the credential private key.

Firstyear: If anything, since passkeys from apple are not compliant then the bit flags need to be swapped - 0 means "can be backed up" and we have no idea about the credentials state, and 1 is a hard "yes, we are guaranteed thes credential is not transferable backed up". It's the only way to actually draw a correct view here.

Because if the bit is 0 (IE all devices today) you have to check the attestation to then determine if it's backed up or not, but if it's a 1, then you know the manufacturer is compliant.

emlun: > but if it's a 1, then you know the manufacturer is compliant.

No you don't. You can only know that if you trust the attestation.

But this did bring up a related point: if we're changing the meaning of 0, then already existing credentials that can be backed up should be allowed to change their backup eligibility (BE) bit from 0 to 1, so we can't say that the BE bit must never change. I'll raise this point in the top-level discussion.

Firstyear: > No you don't. You can only know that if you trust the attestation.

I know this. My point is that if the bit is a 1 then the manufacturer is making a token after this documents existing, and is saying "yep, this credential is locked to the device subject to attestation". I'm excruciatingly aware of how this works. 

Regardless, like I said, the question you should be asking is "even with attestation, can an RP assert that this credential is locked to a single cryptographic processor". 

In the current for of this backup state PR the answer is NO an RP can not achieve this. For the primary reason as mentioned with apple. 0 means "it might be backed up" and 1 means "it certainly is". This means that unless apple creates a whole second attestation CA stack in parallel which "guarantees something is not backed up", then any RP has to assume every apple device is a passkey, forever more, meaning they can't be used in pretty much any enterprise RP context. 

So the bit's MUST be swapped, else this PR is practically useless to any RP in a more controlled environment.

emlun:
This came up in [an above discussion thread](https://github.com/w3c/webauthn/pull/1695#discussion_r823287805):

If a credential exists today and supports backup, then its backup eligibility (BE) flag (currently RFU2) will be set to 0 as currently specified. After we change the meaning of that bit, this credential should have its BE flag set to 1 instead. But the current proposal defines that the BE bit is permanent and MUST NOT change. This makes it impossible for the existing credential to remain compliant with the spec after this change.

How do we resolve this? I can see a couple of options:

 1. Allow the BE flag to change from 0 to 1.

    This would effectively redefine BE=0 from "will never be backup eligible" to "is not currently configured to be backed up". Would it also be allowed to change from 1 to 0?

 2. Add an informal note informing RPs that some few credentials' BE flags may change from 0 to 1 during a transitional period as they implement the new requirement, but it's expected that the "will never be backup eligible" promise will become accurate as the feature matures.

 3. @Firstyear [proposes above](https://github.com/w3c/webauthn/pull/1695#discussion_r850033652) to invert the bit: 0 would mean "might support backup" and 1 would mean "will never support backup".

    This would cover the case of already-existing backup eligible credentials, but it would worse represent the vast majority of existing credentials that are not and never will be backup eligible. Those credentials would then be better represented by BE=1, but most hardware authenticators would not be able to make that change.

I'm not sure whether I favour either of (1) and (2). I am against (3) for the reason noted above.

Firstyear:
I don't really see how (1) solves the problem either.

(3) might be annoying, but at least it communicates clearly the state of what's happened. Apple jumped the gun, solved a real user experience problem, but did it in a way that will mess up RP's ability to make security decisions so now we have to assume everything is a passkey. 

(2) means that the BE flags then "cant really be trusted" if they "might change over time". Suddenly the security properties you asserted over a credential now changed? That's not okay from an RP's view. About the only way I'd see this being okay is if there is also advice in the spec that says:

- Transition from 0 -> 1 is valid, 1 -> 0 is not, and must result in the credential being invalidated.
- On transition from 0 -> 1, then the RP MAY make a policy decision about if they invalidate the credential or not.

I think that's the only way I'd really accept 2 here, is if we help guide RP's on what MIGHT be happening and what they may want to do. Backup of a credential should be a one way street, where once it goes to the backed up state, we must assume it's no longer HW bound.

Firstyear:
Worth pointing out there is no way for an RP to signal in credential creation options that we want a NON backed up credential from what I can see either ...

index.bs
emlun: > Remember as an RP the more important question is "Is this credential locked firmly to this single device" rather than "is this credential backed up or transfered elsewhere?".

This is an overgeneralization. Many RPs will care a lot about the latter question and not at all about the former.

> And in those contexts they can use attestation to validate the manufacturer (IE yubikey with flag 0 is not backed up) where as when apple eventually becomes compliant they can have a 1 in this field.

Right. So you _can_ confirm that a credential is not backed up with the flag as currently proposed. And remember that Apple has no obligation to set the bit to 1 before we reach Proposed Standard maturity with this change.

> Because apple today don't have a seperate attestation cert chain for passkey vs resident key, meaning that in the current form of this proposal, with apple it will be impossible for us to determine if the credential is NOT backed up without apple changing their entire provisioning infrastructure to have two seperate attestation chains.

I'm afraid you'll have to direct that complaint to Apple, it is not a fault of this proposal.

MasterKale:
> This makes it impossible for the existing credential to remain compliant with the spec after this change.
We shouldn't ignore the fact that these types of multi-device credentials are currently opt-in; they're not representative of functionality that's GA.

I believe we should keep the immutability of the BE flag, and continue to allow BE of 0 to mean "not eligible for backup" because it will accurately represent more credentials than changing plans so BE of 1 means "hardware backed"; the latter leaves existing security keys in a weird state (to @emlun's point).

What if instead we start suggesting ways in which RP's can detect these "beta" credentials to retire or accept updated flags for accordingly. For example:

> If a credential with `"internal"` transport is successfully used for auth in Safari on macOS after X date (whenever Apple's passkeys go GA), and it happens to come back with BE of 1 and BS of 1, then it's likely Apple successfully updated their implementation of the spec and it should be safe to update the RP's knowledge of the credential's flags accordingly. In all other browsers and OS's, this should result in a failure because no one else is demonstrating multi-device credentials.

There's undeniably going to be a transition period because of how easy it's been for nerds like me to opt in and how impossible it's been for RP's to know a multi-device credential from a single-device credential. If someone opts into a beta technology they have to expect some rough spots along the way. Why not lean into that a little before we make a spec decision that'll make the spec even trickier to consume?

serianox:
We have to factor in that this change will take at least a couple of years to reach *new* security keys. So if we invert the meaning of the BE flag now (0 -> might support backup) , we could even get L3+ security keys that gets rejected because they "may" support backup.

ve7jtb:
If there is no attestation or an attestation with uncertified meta-data then the RP has no idea if the credential is multi-device or not.

Current multi-device credentials do not have attestations.  

There is no magic that will let a RP know that a a BE=0 credential is not multi-device or the UV=1 is really the result of multifactor authentication without checking the meta-data for the authenticator.

You may be able to assume that an authenticator returning BE=1 is probably true because there is no real reason to lie.

The BE flag helps when you have a certified authenticator with a single AAGUID that may do backup on a credential by credential basis.  (none do at this point) 

All other authenticators tend to be an all-or-nothing proposition and that can again be determined by looking in the meta-data.

If an RP is not checking meta-data now they probably don't need to in the future.   Any credential they get muli or single device should be good enough.    If they want to use the flags to determine if it is safe to remove the password because account recovery is taken care of then they can do that with the flags without checking meta-data.

Anyone who wants to restrict authenticators needs to check meta-data.   That is true now and will continue to be after multi-device credentials roll out.

Apple platform credentials with an attestation are currently not multi device.  
Apple platform credentials with no attestation are probably multi-device.
In future apple can implement the flaga and reintroduce attestations.

I don't think the BE flag in current credentials really needs to change state for the small number of credentials we are talking about.

If Apple wants to reflag them they can as their authenticator is not Fido certified so they can do anything they like.  We don't need some special carve out to deal with their beta program.   I don't think anyone is going to complain too much if they do something sensable.

The immutable rule really only meaningfully applies to Fido certified authenticators.

Firstyear:
> Current multi-device credentials do not have attestations.

Incorrect. Apple passkeys are currently multi device and attested. They share the same attestation CA chain as the single-device credentials.

timcappalli:
> > Current multi-device credentials do not have attestations.
> 
> Incorrect. Apple passkeys are currently multi device and attested. They share the same attestation CA chain as the single-device credentials.

This is not the place to discuss pre-standard vendor implementations. There are no generally available implementations of multi-device credentials which is why this PR is written the way it is.

Firstyear:
Duplicated:

Someone from apple contacted me directly to discuss, and it turns out there was a second bug in safari related to attestation and platform syncing, where my device was behaving incorrectly and creating a device-bound key with attestation even when sync was requested. That's what led to the confusion as it appears apple have an undocumented behaviour where passkeys do NOT respond to any attestation requests, but if sync is disabled, they will then allow attestation to proceed.

As a result, my concerns about currently existing passkeys are invalid, meaning that the proposal "as is" is good to go then. Sorry for the pain @timcappalli :(

index.bs
```diff
@@ -971,6 +972,24 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
     consent=] for (i.e., <em>authorizes</em>) a [=ceremony=] to proceed. This MAY involve [=user verification=] if the
     employed [=authenticator=] is capable, or it MAY involve a simple [=test of user presence=].
 
+: <dfn>Backup</dfn>
+: <dfn>Backed Up</dfn>
+:: [=Public Key Credential Sources=] may be [=backed up=] in some fashion such that they may become present on an authenticator other
```
agl, line 977: (nit, not blocking: these definitions link to themselves several times. Seems superfluous!)

index.bs
```diff
@@ -3657,7 +3688,89 @@ This is because the first 37 bytes of the signed data in a FIDO U2F authenticati
...
+        </tr>
+    </table>
+    <figcaption>
+        `BE` and `BS` [=flag=] combinations
+    </figcaption>
+</figure>
+
+It is RECOMMENDED that [=[RPS]=] store the most recent value of these [=flags=] with the [=user account=] for future evaluation.
+
+The following is a non-normative, non-exhaustive list of how [=[RPS]=] might utilize these [=flags=]:
```
agl, line 3744: (nit, not blocking: s/utilize/use/)

index.bs
```diff
@@ -3657,7 +3688,89 @@ This is because the first 37 bytes of the signed data in a FIDO U2F authenticati
...
+
+    - Requiring additional [=authenticators=]:
 
+        When `BE` [=flag=] is set to `0`, the [=generating authenticator=] will never allow the credential to transition from a
```
agl, line 3748: (not blocking) Does any credential &ldquo;transition from a [=single-device credential=] to a [=multi-device credential=]&rdquo;? The definition above says that a credential is exactly one of those things, fixed at creation time, unchanging. I fear that this paragraph suggests otherwise.

index.bs
```diff
@@ -3657,7 +3688,89 @@ This is because the first 37 bytes of the signed data in a FIDO U2F authenticati
...
+    
+    - Upgrading a user to a password-free account:
+
+        When the `BS` [=flag=] changes from `0` to `1`, the [=authenticator=] is signaling that the [=credential=] is backed up and is protected from single device loss.
+        
+        A [=Relying Party=] may decide to prompt the user to upgrade their account security and remove their password.
+    
+    - Adding an additional factor after a state change:
+
+        When the `BS` [=flag=] changes from `1` to `0`, the [=authenticator=] is signaling that the [=credential=] is no longer backed up,
```
agl, line 3765: (non blocking) In my mind, I thought this transition was going to be banned for simplicity, but I haven't been following all the discussions on this.

agl (approved):

Kieun:
The platforms might decide whether the new device is capable of restoring the backup credential? Is there any policy for this?
RP might want to enforce 2FA (with UV) and so it requires UV for the registration/authentication. If the generated credential is BE (backup eligible) and then restored from the new device, then still we can make sure that the credential is protected by UV? Or, is it possible that the credential is restored but the UP is only supported on the new device?

Firstyear:
> The platforms might decide whether the new device is capable of restoring the backup credential? Is there any policy for this? RP might want to enforce 2FA (with UV) and so it requires UV for the registration/authentication. If the generated credential is BE (backup eligible) and then restored from the new device, then still we can make sure that the credential is protected by UV? Or, is it possible that the credential is restored but the UP is only supported on the new device?

Wouldn't this also imply a change in the attestation chain of the credential when it's restored since it's on a different piece of hardware that has different UV capabilities?

akshayku (approved):

emlun:
> Wouldn't this also imply a change in the attestation chain of the credential when it's restored since it's on a different piece of hardware that has different UV capabilities?

I believe this question is what the [`devicePubKey` extension](https://github.com/w3c/webauthn/pull/1663) is intended to answer - that the attestation for the backed-up key can be "weak", but the `devicePubKey` attestation on each device can give stronger guarantees about the device-bound key.

But it's a good point that we should probably include some guidance or disclaimers about how this interacts with UV. Presumably, a credential restored from backup onto a new authenticator would use whatever UV option(s) available on the new authenticator, unrelated to any UV configured on the original authenticator. So for example, if the original authenticator had a PIN configured, then after restoring the credential onto a new authenticator, the new authenticator may have a different PIN configured or no PIN at all.

Whether my presumption above is accurate or not, I think we should call out in the [User Verification](https://w3c.github.io/webauthn/#user-verification) definition how credential backup is expected to interact with UV. This might relate to `devicePubKey` as well - for example, if an RP is keeping track of which credentials have sent `UV=1` in the past, then the RP may want to track that per device public key for multi-device credentials (because `UV=1` is meaningless the first time on a new device, but meaningful from the second time on).

ve7jtb:
In practice, the three platform authenticators only support credential release via UV if the RP requests it or not.

However, from a specification point of view, those are just current implementation choices.  

Unless we say explicitly that multi-device credentials need to be protected with some form of user verification, then we need to remind RP that a credential might always be returned without UV even if it was created with UV.   That is also the case now as the UV flag can be tampered with in the request.  

Chrome and Windows will prompt a user to add a pin for roaming authenticators if you send UV required in the request.  
One possibility is requiring authenticators to register a UV method if UV=required in the request and none is configured.  Though that is typically an implementation detail and not in the spec.

Kieun:
At least, during registration process, RP has a way to specify their preference or policy for the authentication. But, if the restored credential has different properties, it will cause unexpected situation. So, my thought is that the backup credential needs to keep the basic properties from the original one.

agl:
Agreed to merge at face-to-face meeting.

emlun (approved):

index.bs
```diff
@@ -3657,7 +3688,89 @@ This is because the first 37 bytes of the signed data in a FIDO U2F authenticati
...
+        </tr>
+    </table>
+    <figcaption>
+        `BE` and `BS` [=flag=] combinations
+    </figcaption>
+</figure>
+
+It is RECOMMENDED that [=[RPS]=] store the most recent value of these [=flags=] with the [=user account=] for future evaluation.
+
+The following is a non-normative, non-exhaustive list of how [=[RPS]=] might use these [=flags=]:
```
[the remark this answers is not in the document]
emlun: Moved to #1740.

Issue #1683, opened 2021-11-15
Provide request deserialization, response serialization


dwaite:
The addition of `ArrayBuffer` in the WebAuthn APIs has been an often-cited challenge for relying party developers, as JavaScript does not have integration for converting such buffers to text for serialization via JSON. JSON itself is not necessarily the best format because of the binary content, including for items like extensions.

I'd like to propose deserialization request forms of PublicKeyCredential as long as serialization of the response forms of PublicKeyCredential.

To do this, I would propose a serialization based on a schema (such as CBOR and CDDL) covering the options valid for a PublicKeyCredential. CBOR data is preferred because of the need to preserve the binary signed messages, and due to several internal values being declared as binary identifiers or as CBOR. 

Additionally, convenience methods to work in terms of a base64-encoded form of the CBOR data would be desirable for use in text-based API.

As deserialization would result in a PublicKeyCredential object, Javascript would still have full fidelity to inspect and possibly manipulate before making a create() or get() request. Likewise, Javascript would have the ability to inspect the response object before serializing it to share with a remote service.

Deserialization for responses and serialization of requests would both be more difficult due to the possibility of arbitrary manipulation by the user and responses having methods in addition to data.

MasterKale:
Thank you for submitting this ticket, I had intended to file something similar ages ago but never managed to make time for it. In my mind I totally think `PublicKeyCredential` needs to be expanded in this direction.

I encourage us to think about how devs work with native JS capabilities and build the spec _for_ them, not build the spec for us and hope JS devs will follow. To get some of my own thoughts out about this:

- In my mind base64url would be the preferable encoding/decoding scheme because it's [specifically called out as a dependency](https://www.w3.org/TR/webauthn-2/#sctn-dependencies).
- I think making it easier for developers to consume WebAuthn API via JSON inputs and outputs would go a long way in encouraging adoption without forcing developers to A) figure out WebAuthn serialization for themselves, or B) incorporate _another_ dependency like Github's [webauthn-json](https://github.com/github/webauthn-json) or my [@simplewebauthn/browser](https://github.com/MasterKale/SimpleWebAuthn).
[...]

PR #1703, opened 2022-02-23
New PublicKeyCredential methods for JSON (de)serialization

MasterKale:
This PR introduces new methods to `PublicKeyCredential` that lean on the client to assist with JSON deserialization of `PublicKeyCredentialCreationOptions` and `PublicKeyCredentialRequestOptions`, and JSON serialization of `PublicKeyCredential` output from `navigator.credentials.create()` and `navigator.credentials.get()`.

See Issue #1683 for more context.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/pull/1703.html" title="Last updated on Jun 20, 2022, 7:49 PM UTC (12e5071)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/1703/dcd3190...12e5071.html" title="Last updated on Jun 20, 2022, 7:49 PM UTC (12e5071)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1452,10 +1452,13 @@ that are returned to the caller when a new credential is created, or a new asser
         [=client extension processing=].
 
     :   {{PublicKeyCredential/toJSON()}}
-    ::  This operation returns {{RegistrationResponseJSON}} or {{AuthenticationResponseJSON}}, a JSON-serialized representation of the instance of
-        {{PublicKeyCredential}} suitable for submission to a [=[RP]=] server as a payload with `application/json` mimetype. The [=client=] is in
-        charge of encoding any {{ArrayBuffer}} values using [=base64url encoding=] to generate strings compatible with JSON syntax. {{ArrayBuffer}}
-        values to be encoded are identified according to the detected type of the instance's `response` property.
+    ::  This operation returns {{RegistrationResponseJSON}} or {{AuthenticationResponseJSON}}, that is a JSON-serialized
+        representation of the instance of {{PublicKeyCredential}} suitable for submission to a [=[RP]=] server as a DOMString
+        payload with `application/json` mimetype. The [=client=] is in charge of
+        [serializing values to JSON as usual](https://webidl.spec.whatwg.org/#idl-tojson-operation), but must take additional
+        steps to encode any {{ArrayBuffer}} values to {{DOMString}} values using [=base64url encoding=] to ensure compatibility
+        with JSON syntax. {{ArrayBuffer}} values to be encoded are identified according to the detected type of the instance's
+        `response` property.
 
         <dl dfn-for="">
             <xmp class="idl">
--- a/index.bs
+++ b/index.bs
@@ -1407,6 +1407,7 @@ that are returned to the caller when a new credential is created, or a new asser
         [SameObject] readonly attribute AuthenticatorResponse    response;
         [SameObject] readonly attribute DOMString?               authenticatorAttachment;
         AuthenticationExtensionsClientOutputs getClientExtensionResults();
+        TBD toJSON();
     };
 </xmp>
 <dl dfn-type="attribute" dfn-for="PublicKeyCredential">
@@ -1449,6 +1450,13 @@ that are returned to the caller when a new credential is created, or a new asser
     ::  This operation returns the value of {{PublicKeyCredential/[[clientExtensionsResults]]}}, which is a [=map=] containing
         [=extension identifier=] → [=client extension output=] entries produced by the extension's
         [=client extension processing=].
+
+    :   {{PublicKeyCredential/toJSON()}}
+    ::  This operation returns a JSON-serialized representation of the instance of {{PublicKeyCredential}} suitable for submission
+        to a [=[RP]=] server as a payload with `application/json` mimetype. The [=client=] returns an object with the same structure
+        as the instance of the {{PublicKeyCredential}} with {{ArrayBuffer}} values encoded with [=base64url encoding=]. {{ArrayBuffer}}
+        values to be encoded are identified according to the detected type of the instance's `response` property, which will either be
+        {{AuthenticatorAttestationResponse}} or {{AuthenticatorAssertionResponse}}.
 
     :   <dfn>\[[type]]</dfn>
     ::  The {{PublicKeyCredential}} [=interface object=]'s {{Credential/[[type]]}} [=internal slot=]'s value is the string
@@ -2396,18 +2404,6 @@ Note: Invoking this method from a [=browsing context=] where the [=Web Authentic
 This method is intended to make WebAuthn API consumption tenable without requiring the use of third-party libraries.
 
 Upon invocation, the [=client=] takes the provided options and returns an identically-structured object with specific strings encoded with [=base64url encoding=] decoded to {{ArrayBuffer}} values according to the corresponding type definition.
-
-</div>
-
-### Serialize Registration and Authentication ceremony responses - PublicKeyCredential's `toJSON()` Method ### {#sctn-toJSON}
-
-<div link-for-hint="WebAuthentication/toJSON">
-
-[=[WRPS]=] use this method to convert the instance of {{PublicKeyCredential}} returned from {{CredentialsContainer/create()|navigator.credentials.create()}} or {{CredentialsContainer/get()|navigator.credentials.get()}} to a JSON-serialized representations suitable for submission to a [=[RP]=] server as a payload with `application/json` mimetype.
-
-This method is intended to make WebAuthn API consumption tenable without requiring the use of third-party libraries.
-
-Upon invocation, the [=client=] returns an object with the same structure as the instance of the {{PublicKeyCredential}} with {{ArrayBuffer}} values encoded with [=base64url encoding=]. {{ArrayBuffer}} values to be encoded are identified according to the detected type of the instance's `response` property, which will either be {{AuthenticatorAttestationResponse}} or {{AuthenticatorAssertionResponse}}.
 
 </div>
 
--- a/index.bs
+++ b/index.bs
@@ -1453,7 +1453,7 @@ that are returned to the caller when a new credential is created, or a new asser
 
     :   {{PublicKeyCredential/toJSON()}}
     ::  This operation returns {{RegistrationResponseJSON}} or {{AuthenticationResponseJSON}},
-        which are [=JSON types=] mirroring {{PublicKeyCredential}}, suitable for submission to a
+        which are [=JSON type=] representations mirroring {{PublicKeyCredential}}, suitable for submission to a
         [=[RP]=] server as an `application/json` payload. The [=client=] is in charge of
         [serializing values to JSON types as usual](https://webidl.spec.whatwg.org/#idl-tojson-operation),
         but MUST take additional steps to first encode any {{ArrayBuffer}} values to {{DOMString}} values
@@ -2468,6 +2468,10 @@ also apply to any [=client extension inputs=] processed by the [=client=].
 "WebAuthn Extension Identifiers" registry [[!IANA-WebAuthn-Registries]] but not defined in
 [[#sctn-extensions]].
 
+If the [=client=] encounters any issues parsing any of the [=JSON type=] representations then it
+MUST return an error code equivalent to "{{EncodingError}}" with a description of the incompatible
+value and terminate the operation.
+
 <xmp class="idl">
     partial interface PublicKeyCredential {
         static PublicKeyCredentialCreationOptions parseCreationOptionsFromJSON(PublicKeyCredentialCreationOptionsJSON options);
@@ -2521,6 +2525,10 @@ also apply to any [=client extension inputs=] processed by the [=client=].
 {{AuthenticationExtensionsClientInputsJSON}} MAY include extensions registered in the IANA
 "WebAuthn Extension Identifiers" registry [[!IANA-WebAuthn-Registries]] but not defined in
 [[#sctn-extensions]].
+
+If the [=client=] encounters any issues parsing any of the [=JSON type=] representations then it
+MUST return an error code equivalent to "{{EncodingError}}" with a description of the incompatible
+value and terminate the operation.
 
 <xmp class="idl">
     partial interface PublicKeyCredential {
--- a/index.bs
+++ b/index.bs
@@ -2437,21 +2437,11 @@ Upon invocation, the [=client=] takes the provided options and returns an identi
 
 <xmp class="idl">
     partial interface PublicKeyCredential {
-        static PublicKeyCredentialCreationOptions optionsFromJSON(CreationOptionsJSON creationOptions);
+        static PublicKeyCredentialCreationOptions getCreationOptionsFromJSON(PublicKeyCredentialCreationOptionsJSON options);
     };
 
     partial interface PublicKeyCredential {
-        static PublicKeyCredentialRequestOptions optionsFromJSON(RequestOptionsJSON requestOptions);
-    };
-
-    dictionary CreationOptionsJSON {
-        required DOMString                                   method = "create";
-        required PublicKeyCredentialCreationOptionsJSON      options;
-    };
-
-    dictionary RequestOptionsJSON {
-        required DOMString                                   method = "get";
-        required PublicKeyCredentialRequestOptionsJSON       options;
+        static PublicKeyCredentialRequestOptions getRequestOptionsFromJSON(PublicKeyCredentialRequestOptionsJSON options);
     };
 
     dictionary PublicKeyCredentialCreationOptionsJSON {
--- a/index.bs
+++ b/index.bs
@@ -2444,16 +2444,16 @@ Note: Invoking this method from a [=browsing context=] where the [=Web Authentic
 
 </div>
 
-### Deserialize Registration ceremony options - PublicKeyCredential's `getCreationOptionsFromJSON()` Method ### {#sctn-getCreationOptionsFromJSON}
+### Deserialize Registration ceremony options - PublicKeyCredential's `parseCreationOptionsFromJSON()` Method ### {#sctn-parseCreationOptionsFromJSON}
 
-<div link-for-hint="WebAuthentication/getCreationOptionsFromJSON">
+<div link-for-hint="WebAuthentication/parseCreationOptionsFromJSON">
 
 [=[WRPS]=] use this method to convert [=JSON type=] representations of options for
 {{CredentialsContainer/create()|navigator.credentials.create()}} into
 {{PublicKeyCredentialCreationOptions}}.
 
 Upon invocation, the [=client=] MUST convert the
-{{PublicKeyCredential/getCreationOptionsFromJSON(options)/options}} argument into a new,
+{{PublicKeyCredential/parseCreationOptionsFromJSON(options)/options}} argument into a new,
 identically-structured {{PublicKeyCredentialCreationOptions}} object, using [=base64url encoding=]
 to decode any {{DOMString}} attributes in {{PublicKeyCredentialCreationOptionsJSON}} that correspond
 to [=buffer source type=] attributes in {{PublicKeyCredentialCreationOptions}}. This conversion MUST
@@ -2465,7 +2465,7 @@ extend to any [=client extension inputs=] processed by the [=client=].
 
 <xmp class="idl">
     partial interface PublicKeyCredential {
-        static PublicKeyCredentialCreationOptions getCreationOptionsFromJSON(PublicKeyCredentialCreationOptionsJSON options);
+        static PublicKeyCredentialCreationOptions parseCreationOptionsFromJSON(PublicKeyCredentialCreationOptionsJSON options);
     };
 
     dictionary PublicKeyCredentialCreationOptionsJSON {
@@ -2498,16 +2498,16 @@ extend to any [=client extension inputs=] processed by the [=client=].
 
 </div>
 
-### Deserialize Authentication ceremony options - PublicKeyCredential's `getRequestOptionsFromJSON()` Methods ### {#sctn-getRequestOptionsFromJSON}
+### Deserialize Authentication ceremony options - PublicKeyCredential's `parseRequestOptionsFromJSON()` Methods ### {#sctn-parseRequestOptionsFromJSON}
 
-<div link-for-hint="WebAuthentication/getRequestOptionsFromJSON">
+<div link-for-hint="WebAuthentication/parseRequestOptionsFromJSON">
 
 [=[WRPS]=] use this method to convert [=JSON type=] representations of options for
 {{CredentialsContainer/get()|navigator.credentials.get()}} into
 {{PublicKeyCredentialRequestOptions}}.
 
 Upon invocation, the [=client=] MUST convert the
-{{PublicKeyCredential/getRequestOptionsFromJSON(options)/options}} argument into a new,
+{{PublicKeyCredential/parseRequestOptionsFromJSON(options)/options}} argument into a new,
 identically-structured {{PublicKeyCredentialRequestOptions}} object, using [=base64url encoding=]
 to decode any {{DOMString}} attributes in {{PublicKeyCredentialRequestOptionsJSON}} that correspond
 to [=buffer source type=] attributes in {{PublicKeyCredentialRequestOptions}}. This conversion MUST
@@ -2519,7 +2519,7 @@ extend to any [=client extension inputs=] processed by the [=client=].
 
 <xmp class="idl">
     partial interface PublicKeyCredential {
-        static PublicKeyCredentialRequestOptions getRequestOptionsFromJSON(PublicKeyCredentialRequestOptionsJSON options);
+        static PublicKeyCredentialRequestOptions parseRequestOptionsFromJSON(PublicKeyCredentialRequestOptionsJSON options);
     };
 
     dictionary PublicKeyCredentialRequestOptionsJSON {
--- a/index.bs
+++ b/index.bs
@@ -2453,7 +2453,7 @@ Upon invocation, the [=client=] takes the provided options and returns an identi
         required Base64URLString                                challenge;
         required sequence<PublicKeyCredentialParameters>        pubKeyCredParams;
         unsigned long                                           timeout;
-        required sequence<PublicKeyCredentialDescriptorJSON>    excludeCredentials = [];
+        sequence<PublicKeyCredentialDescriptorJSON>    excludeCredentials = [];
         AuthenticatorSelectionCriteria                          authenticatorSelection;
         DOMString                                               attestation = "none";
         AuthenticationExtensionsClientInputsJSON                extensions;
--- a/index.bs
+++ b/index.bs
@@ -2453,6 +2453,11 @@ Upon invocation, the [=client=] must take the provided options and return an ide
 making sure to decode to {{ArrayBuffer}} those {{DOMString}} properties that are encoded with [=base64url encoding=]
 but must be {{ArrayBuffer}} values according to the definition of {{PublicKeyCredentialCreationOptions}}.
 
+{{AuthenticationExtensionsClientInputsJSON}} may include extensions registered in the IANA
+"WebAuthn Extension Identifiers" registry [[!IANA-WebAuthn-Registries]] but not defined in [[#sctn-extensions]].
+In either case the [=client=] must decode to {{ArrayBuffer}} those {{DOMString}} values that are encoded with
+[=base64url encoding=] but must be {{ArrayBuffer}} values according to the extension's definition.
+
 <xmp class="idl">
     partial interface PublicKeyCredential {
         static PublicKeyCredentialCreationOptions getCreationOptionsFromJSON(PublicKeyCredentialCreationOptionsJSON options);
@@ -2483,17 +2488,6 @@ but must be {{ArrayBuffer}} values according to the definition of {{PublicKeyCre
     };
 
     dictionary AuthenticationExtensionsClientInputsJSON {
-        DOMString                                       appid;
-        DOMString                                       appidExclude;
-        boolean                                         uvm;
-        boolean                                         credProps;
-        AuthenticationExtensionsLargeBlobInputsJSON     largeBlob;
-    };
-
-    dictionary AuthenticationExtensionsLargeBlobInputsJSON {
-        DOMString                       support;
-        boolean                         read;
-        Base64URLString                 write;
     };
 </xmp>
 
--- a/index.bs
+++ b/index.bs
@@ -2387,6 +2387,30 @@ Note: Invoking this method from a [=browsing context=] where the [=Web Authentic
 
 </div>
 
+### Deserialize Registration and Authentication ceremony options - PublicKeyCredential's `optionsFromJSON()` Method ### {#sctn-optionsFromJSON}
+
+<div link-for-hint="WebAuthentication/optionsFromJSON">
+
+[=[WRPS]=] use this method to convert JSON-compatible serialized representations of options for {{CredentialsContainer/create()|navigator.credentials.create()}} to {{PublicKeyCredentialCreationOptions}}, or JSON-compatible serialized representations of options for {{CredentialsContainer/get()|navigator.credentials.get()}} to {{PublicKeyCredentialRequestOptions}}.
+
+This method is intended to make WebAuthn API consumption tenable without requiring the use of third-party libraries.
+
+Upon invocation, the [=client=] takes the provided options and returns an identically-structured object with specific strings encoded with [=base64url encoding=] decoded to {{ArrayBuffer}} values according to the corresponding type definition.
+
+</div>
+
+### Serialize Registration and Authentication ceremony responses - PublicKeyCredential's `toJSON()` Method ### {#sctn-toJSON}
+
+<div link-for-hint="WebAuthentication/toJSON">
+
+[=[WRPS]=] use this method to convert the instance of {{PublicKeyCredential}} returned from {{CredentialsContainer/create()|navigator.credentials.create()}} or {{CredentialsContainer/get()|navigator.credentials.get()}} to a JSON-serialized representations suitable for submission to a [=[RP]=] server as a payload with `application/json` mimetype.
+
+This method is intended to make WebAuthn API consumption tenable without requiring the use of third-party libraries.
+
+Upon invocation, the [=client=] returns an object with the same structure as the instance of the {{PublicKeyCredential}} with {{ArrayBuffer}} values encoded with [=base64url encoding=]. {{ArrayBuffer}} values to be encoded are identified according to the detected type of the instance's `response` property, which will either be {{AuthenticatorAttestationResponse}} or {{AuthenticatorAssertionResponse}}.
+
+</div>
+
 ## Authenticator Responses (interface <dfn interface>AuthenticatorResponse</dfn>) ## {#iface-authenticatorresponse}
 
 [=Authenticators=] respond to [=[RP]=] requests by returning an object derived from the
--- a/index.bs
+++ b/index.bs
@@ -2407,6 +2407,12 @@ Note: Invoking this method from a [=browsing context=] where the [=Web Authentic
 
 Upon invocation, the [=client=] takes the provided options and returns an identically-structured object with specific strings encoded with [=base64url encoding=] decoded to {{ArrayBuffer}} values according to the corresponding type definition.
 
+<xmp class="idl">
+    partial interface PublicKeyCredential {
+        static TBD optionsFromJSON();
+    };
+</xmp>
+
 Note: This method is intended to make WebAuthn API consumption tenable without requiring the use of third-party libraries to handle this
 deseraili
 
--- a/index.bs
+++ b/index.bs
@@ -1407,7 +1407,7 @@ that are returned to the caller when a new credential is created, or a new asser
         [SameObject] readonly attribute AuthenticatorResponse    response;
         [SameObject] readonly attribute DOMString?               authenticatorAttachment;
         AuthenticationExtensionsClientOutputs getClientExtensionResults();
-        TBD toJSON();
+        PublicKeyCredentialJSON toJSON();
     };
 </xmp>
 <dl dfn-type="attribute" dfn-for="PublicKeyCredential">
@@ -1457,6 +1457,37 @@ that are returned to the caller when a new credential is created, or a new asser
         as the instance of the {{PublicKeyCredential}} with {{ArrayBuffer}} values encoded with [=base64url encoding=]. {{ArrayBuffer}}
         values to be encoded are identified according to the detected type of the instance's `response` property, which will either be
         {{AuthenticatorAttestationResponse}} or {{AuthenticatorAssertionResponse}}.
+
+        <xmp class="idl">
+            typedef DOMString Base64URLString;
+            typedef (RegistrationResponseJSON or AuthenticationResponseJSON) PublicKeyCredentialJSON;
+
+            dictionary RegistrationResponseJSON {
+                Base64URLString id;
+                Base64URLString rawId;
+                AuthenticatorAttestationResponseJSON response;
+                DOMString type;
+            };
+
+            dictionary AuthenticationResponseJSON {
+                Base64URLString id;
+                Base64URLString rawId;
+                AuthenticatorAssertionResponseJSON response;
+                DOMString type;
+            };
+
+            dictionary AuthenticatorAttestationResponseJSON {
+                Base64URLString clientDataJSON;
+                Base64URLString attestationObject;
+            };
+
+            dictionary AuthenticatorAssertionResponseJSON {
+                Base64URLString clientDataJSON;
+                Base64URLString authenticatorData;
+                Base64URLString signature;
+                Base64URLString? userHandle;
+            };
+        </xmp>
 
         <div class="note">
         Note: This method is intended to make WebAuthn API consumption tenable without requiring the use of third-party libraries.
--- a/index.bs
+++ b/index.bs
@@ -2455,34 +2455,37 @@ Upon invocation, the [=client=] takes the provided options and returns an identi
     };
 
     dictionary PublicKeyCredentialCreationOptionsJSON {
-        PublicKeyCredentialRpEntity                     rp;
-        PublicKeyCredentialUserEntityJSON               user;
-        Base64URLString                                 challenge;
-        sequence<PublicKeyCredentialDescriptorJSON>     excludeCredentials = [];
-        AuthenticatorSelectionCriteria                  authenticatorSelection;
-        DOMString                                       attestation = "none";
-        AuthenticationExtensionsClientInputsJSON        extensions;
+        required PublicKeyCredentialRpEntity                    rp;
+        required PublicKeyCredentialUserEntityJSON              user;
+        required Base64URLString                                challenge;
+        required sequence<PublicKeyCredentialParameters>        pubKeyCredParams;
+        unsigned long                                           timeout;
+        required sequence<PublicKeyCredentialDescriptorJSON>    excludeCredentials = [];
+        AuthenticatorSelectionCriteria                          authenticatorSelection;
+        DOMString                                               attestation = "none";
+        AuthenticationExtensionsClientInputsJSON                extensions;
     };
 
     dictionary PublicKeyCredentialRequestOptionsJSON {
-        Base64URLString challenge;
-        unsigned long                                   timeout;
-        DOMString                                       rpId;
-        sequence<PublicKeyCredentialDescriptorJSON>     allowCredentials = [];
-        DOMString                                       userVerification = "preferred";
-        AuthenticationExtensionsClientInputsJSON        extensions;
+        required Base64URLString                                challenge;
+        unsigned long                                           timeout;
+        DOMString                                               rpId;
+        sequence<PublicKeyCredentialDescriptorJSON>             allowCredentials = [];
+        DOMString                                               userVerification = "preferred";
+        AuthenticationExtensionsClientInputsJSON                extensions;
     };
 
     dictionary PublicKeyCredentialUserEntityJSON {
-        Base64URLString         id;
-        DOMString               name;
-        DOMString               displayName;
+        required Base64URLString        id;
+        required DOMString              name;
+        required DOMString              displayName;
     };
 
     dictionary PublicKeyCredentialDescriptorJSON {
-        Base64URLString         id;
-        DOMString               type;
-        sequence<DOMString>     transports;
+        required Base64URLString        id;
+        required DOMString              type;
+        sequence<DOMString>             transports;
+    };
 
     dictionary AuthenticationExtensionsClientInputsJSON {
         DOMString                                       appid;
--- a/index.bs
+++ b/index.bs
@@ -1452,11 +1452,10 @@ that are returned to the caller when a new credential is created, or a new asser
         [=client extension processing=].
 
     :   {{PublicKeyCredential/toJSON()}}
-    ::  This operation returns a JSON-serialized representation of the instance of {{PublicKeyCredential}} suitable for submission
-        to a [=[RP]=] server as a payload with `application/json` mimetype. The [=client=] returns an object with the same structure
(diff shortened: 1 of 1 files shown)


MasterKale:
I apologize for the roughness of these initial drafts, this is my first time attempting to submit something substantial to the spec. @nicksteele has graciously volunteered for a bit of hand-holding as I work on this, but I look forward to your comments and suggestions as we work to make this a reality.

index.bs
```diff
@@ -2387,6 +2387,30 @@ Note: Invoking this method from a [=browsing context=] where the [=Web Authentic
...
+
+[=[WRPS]=] use this method to convert JSON-compatible serialized representations of options for {{CredentialsContainer/create()|navigator.credentials.create()}} to {{PublicKeyCredentialCreationOptions}}, or JSON-compatible serialized representations of options for {{CredentialsContainer/get()|navigator.credentials.get()}} to {{PublicKeyCredentialRequestOptions}}.
+
+This method is intended to make WebAuthn API consumption tenable without requiring the use of third-party libraries.
+
+Upon invocation, the [=client=] takes the provided options and returns an identically-structured object with specific strings encoded with [=base64url encoding=] decoded to {{ArrayBuffer}} values according to the corresponding type definition.
+
+</div>
+
+### Serialize Registration and Authentication ceremony responses - PublicKeyCredential's `toJSON()` Method ### {#sctn-toJSON}
```
emlun, line 2402: This should probably be presented as an "instance method" in §5.1, like [`getClientExtensionResults()`](https://pr-preview.s3.amazonaws.com/w3c/webauthn/pull/1703.html#dom-publickeycredential-getclientextensionresults). I'm not sure if we also need to be a bit more specific about how the conversion is done; if so, the section on [converting extension inputs to CBOR](https://pr-preview.s3.amazonaws.com/w3c/webauthn/pull/1703.html#ref-for-client%E2%91%A6%E2%91%A8) could be a good example to try to follow.

index.bs
```diff
@@ -2387,6 +2387,30 @@ Note: Invoking this method from a [=browsing context=] where the [=Web Authentic
 
 </div>
 
+### Deserialize Registration and Authentication ceremony options - PublicKeyCredential's `optionsFromJSON()` Method ### {#sctn-optionsFromJSON}
```
emlun, line 2390: This section should include an IDL definition like in the `isUserVerifyingPlatformAuthenticatorAvailable()` section, and declare the parameter like the other sibling sections do.

emlun (changes requested):
Thanks for taking this on! For starters, some comments on the high-level disposition...

index.bs
MasterKale: Absolutely! I've been working with @dwaite behind the scenes to prepare IDL definitions for these new methods; now that a PR exists we can start that work in earnest.

index.bs
MasterKale: I went ahead and changed `toJSON()` to an instance method, and took a stab at defining some IDLs to help capture what should be returned.

MasterKale:
Alright, I've got new types defined for everything so I'm flipping this to Ready for Review.

One thing I need help with (aside from review of the IDL definitions) is figuring out the proper way to address "FATAL ERROR"s like this:

```
FATAL ERROR: 'dictionary' definitions don't use a 'for' attribute, but this one claims it's for 'PublicKeyCredential' (perhaps inherited from an ancestor). This is probably a markup error.
```

I wasn't able to intuit a solution from existing definitions.

I'm also uncertain how my intended API for `optionsFromJSON()` is best defined as WebIDL. I want to define two possible options blobs, as documented in https://docs.google.com/document/d/e/2PACX-1vTEyAjhn6a3Rqz2KLKcPg7NwoCGO31Lz7E_2zYt8J6Kzey8UUYycv5iukUos5waD4gsml-aEOEs1it0/pub, instead of using basic unions that make it ambiguous which input generates which output.

index.bs
```diff
@@ -1450,6 +1451,47 @@ that are returned to the caller when a new credential is created, or a new asser
         [=extension identifier=] → [=client extension output=] entries produced by the extension's
         [=client extension processing=].
 
+    :   {{PublicKeyCredential/toJSON()}}
+    ::  This operation returns {{RegistrationResponseJSON}} or {{AuthenticationResponseJSON}}, a JSON-serialized representation of the instance of
+        {{PublicKeyCredential}} suitable for submission to a [=[RP]=] server as a payload with `application/json` mimetype. The [=client=] is in
+        charge of encoding any {{ArrayBuffer}} values using [=base64url encoding=] to generate strings compatible with JSON syntax. {{ArrayBuffer}}
```
Firstyear, line 1457: You need to specify if it's padded or unpadded since there are two variants of base64url.

index.bs
```diff
@@ -1450,6 +1451,47 @@ that are returned to the caller when a new credential is created, or a new asser
...
+            dictionary AuthenticatorAssertionResponseJSON {
+                Base64URLString clientDataJSON;
+                Base64URLString authenticatorData;
+                Base64URLString signature;
+                Base64URLString? userHandle;
+            };
+        </xmp>
+
+        <div class="note">
+        Note: This method is intended to make WebAuthn API consumption tenable without requiring the use of third-party libraries.
```
Firstyear, line 1492: Might be better to clarify that it's about preventing client side JS needing to "muck about" with the internals of the types instead I think.

Firstyear (changes requested):

index.bs
MasterKale: This links to the existing ["Base64url encoding"](https://www.w3.org/TR/webauthn-2/#sctn-dependencies) entry in the Dependencies section of the spec. It states the following:

> The term Base64url Encoding refers to the base64 encoding using the URL- and filename-safe character set defined in Section 5 of [[RFC4648]](https://www.w3.org/TR/webauthn-2/#biblio-rfc4648), with all trailing '=' characters omitted (as permitted by Section 3.2) and without the inclusion of any line breaks, whitespace, or other additional characters.

Might this entry instead be updated to explicitly state "unpadded", rather than implying it with "with all trailing '=' characters omitted'"? Or is padded/unpadded not related to the trailing =?

Firstyear: The trailing = represents the padding, yes. So this means it's base64url with no padding vs base64url with padding (which has the ==). But even the inline references to "base64url" generally the accepted default is "padding is default" meaning that if I was to read this, I would interpret this as base64url with padding (unless I checked the references). Many developers and readers will not follow this reference and will assume they know what base64url encoding is here! It's much better to be explicit in the text (base64url without padding) and that links to the section you mention.

index.bs
```diff
@@ -2387,6 +2429,70 @@ Note: Invoking this method from a [=browsing context=] where the [=Web Authentic
...
+    };
+
+    dictionary PublicKeyCredentialCreationOptionsJSON {
+        PublicKeyCredentialRpEntity                     rp;
+        PublicKeyCredentialUserEntityJSON               user;
+        Base64URLString                                 challenge;
+        sequence<PublicKeyCredentialDescriptorJSON>     excludeCredentials = [];
+        AuthenticatorSelectionCriteria                  authenticatorSelection;
+        DOMString                                       attestation = "none";
+        AuthenticationExtensionsClientInputs            extensions;
```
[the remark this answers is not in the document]
MasterKale: Oh, good call, I thought all extensions used JSON-compatible primitives, but I forgot about `largeBlob` using BufferSource/ArrayBuffer for its `write` input and `blob` output.

I'll define a new "-JSON" dictionary for these.

index.bs
emlun: I disagree with @Firstyear here, the reference is fine as is. Defined terms exist to enhance precision, and readers of any specification should be prepared to look up what the terms mean in that context - especially when uses of the terms are hyperlinked to their definitions. And even if we set that aside, WebAuthn already consistently uses base64url _without_ padding for everything else.

index.bs
emlun: Agreed. I also think this note isn't really needed and could just be deleted if we want.

emlun:
Right now the method descriptions don't quite agree with the WebIDL definitions - the descriptions say that they take in and put out JSON, which would be DOMString values, but the definitions work with JavaScript objects. We need to clarify which of the two it is - probably the latter, since there's already the `JSON.parse()` and `.stringify()` methods to transform to/from JSON?

I'm not sure whether WebIDL allows the overloaded signature for `optionsFromJSON`, but either way I think the API is a bit confusing. Clearly the `method` property is meant for the client to decide which method variant is being called, but it's specified with default values, which would probably work against that? It might be better to split this function into two `PublicKeyCredentialCreationOptions.fromJson()` and `PublicKeyCredentialRequestOptions.fromJson()` functions, to clearly separate them.

---

> One thing I need help with (aside from review of the IDL definitions) is figuring out the proper way to address "FATAL ERROR"s like this:

These errors are because the WebIDL block is enclosed in the `<dl dfn-type="attribute" dfn-for="PublicKeyCredential">` tag with the member definitions, so Bikeshed looks for definitions like `PublicKeyCredential.Base64URLString` and fails to find them. I tried a couple things:

- Move the `toJSON()` block out of the parent `<dl>`. This fixes the errors and otherwise seems to still work, but it seems a bit strange to have `toJSON()` outside the definition list where it does belong (it's just the inline type definitions clashing a bit).
- Enclose the `<xmp>` with `<dl dfn-for="">`. This also seems to fix the errors and otherwise work, but I don't know how correct it is.

dwaite:
@emlun The `optionsFromJSON` method is defined on `PublicKeyCredential` rather than  e.g. `PublicKeyCredentialRequestOptions` as the latter is a dictionary rather than an interface.

I'd lean toward having the method take JSON text - this would both increase symmetry with the `toJSON()`method,  and allow us to declare the operation of the method in terms of prose rather than IDL.

dwaite:
> > I'd lean toward having the method take JSON text - this would both increase symmetry with the toJSON()method, and allow us to declare the operation of the method in terms of prose rather than IDL.
> 
> Huh, so if the input is JSON then you don't need WebIDL definitions for it? I approached it as though I was writing something akin to TypeScript definitions for these methods.

You're right - toJSON() emits Objects not Strings, so the current IDL is proper.

MasterKale:
In JS dev land it's easy to forget that "JSON" is actually a `string` with `Content-Type: application/json`. Most of the time when we're "working with JSON" we're talking about the `Object` you get out of `JSON.parse()` when you pass that `string` into it. 

It seems when I mention "JSON" here in the context of this spec I'm evoking thoughts of passing `string`'s around, not JavaScript `Object`'s. Perhaps this is where I'm getting confused. Can someone confirm this for me so I have a chance at getting into the right mindset for some of this feedback?

index.bs
MasterKale: I'll just leave things as-is here and lean on the definition to continue communicating that Base64URL is unpadded.

MasterKale:
Based on some discussion during the WG meeting today, it may be feasible to update the `.toJSON()` method here to reference https://webidl.spec.whatwg.org/#idl-tojson-operation, and then in plain language describe that the client has the responsibility to return an **object** with `ArrayBuffer`'s encoded to strings using Base64URL.

@kreichgauer Did I understand this correctly? Or would we still need to wrangle IDL definitions to declare this behavior?

dwaite:
> Based on some discussion during the WG meeting today, it may be feasible to update the .toJSON() method here to reference https://webidl.spec.whatwg.org/#idl-tojson-operation, and then in plain language describe that the client has the responsibility to return an object with ArrayBuffer's encoded to strings using Base64URL.

I've been researching if this is appropriate to do indicate with a custom WebIDL extended attribute (for example `[Base64URLSerialization]`) which would also provide a hint usable by tooling.

kreichgauer:
> Based on some discussion during the WG meeting today, it may be feasible to update the `.toJSON()` method here to reference https://webidl.spec.whatwg.org/#idl-tojson-operation, and then in plain language describe that the client has the responsibility to return an **object** with `ArrayBuffer`'s encoded to strings using Base64URL.
> 
> @kreichgauer Did I understand this correctly? Or would we still need to wrangle IDL definitions to declare this behavior?

That's how I interpret the toJSON section of the WebIDL spec at least.

I did a little digging, and Web Payments seems to want to define [PaymentResponse.toJSON()](https://www.w3.org/TR/payment-request/#paymentresponse-interface)
 this way.  Though they just refer to the default method steps of the regular operation, which we can't.

The Push API is a counter example. In [PushSubscription.toJSON()](https://www.w3.org/TR/push-api/#dom-pushsubscription-tojson) it defines a separate PushSubscriptionJSON dictionary. But it defines its own processing rules.

RyannaArline:
How do I properly respond

MasterKale:
I gained a bit of insight into WebIDL that convinced me, for sake of WebIDL definitions, it'd be best to split up the static `optionsFromJSON()` into two new static methods instead:

- `getCreationOptionsFromJSON()`
- `getRequestOptionsFromJSON()`

I think it's a little more verbose, but is probably worth the tradeoff if it makes it easy for browser vendors to implement this new functionality.

I also took another stab at refining the description of the `toJSON()` method, this time referring to WebIDL's default instructions for converting an object to a JSON representation with the added responsibility of ArrayBuffer encoding.

index.bs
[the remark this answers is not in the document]
MasterKale: I went ahead and defined this here:

https://github.com/w3c/webauthn/pull/1703/files#diff-5e793325cd2bfc452e268a4aa2f02b4024dd9584bd1db3c2595f61f1ecf7b985R2483

index.bs
MasterKale: These notes have been removed completely.

lgarron:
> * `getCreationOptionsFromJSON()`
> * `getRequestOptionsFromJSON()`

This sounds reasonable on the surface, but I think `get` has the potential to be a *really* confusing prefix, given that exactly one out of two of these results is passed to a function called `get`. 😳

MasterKale:
> > * `getCreationOptionsFromJSON()`
> > * `getRequestOptionsFromJSON()`
> 
> This sounds reasonable on the surface, but I think `get` has the potential to be a _really_ confusing prefix, given that exactly one out of two of these results is passed to a function called `get`. 😳

Thank you for the feedback. Thinking about it from that perspective, I couldn't help but think about how the method would be consumed:

```js
const options = PublicKeyCredential.getRequestOptionsFromJSON({
  'rpId': 'localhost',
  'challenge': 'Ecue...5ZDE',
  'allowCredentials': [
    {
      'id': 'ASdG...om6A',
      'type': 'public-key',
      'transports': ['internal'],
    }
  ],
});
const resp = await navigator.credentials.get(options);
```

This seems fine to me, though. If `navigator.credentials.get()` were actually `PublicKeyCredential.get()` then I think I'd agree more strongly with you.

I'm trying _really_ hard not to end up with another `isUserVerifyingPlatformAuthenticatorAvailable()` method name. I already sacrificed the simple "`optionsFromJSON()`" name I originally wanted. If I don't use `get` as a prefix, though, I fear the next logical prefix would be something like `generate` or the even more "correct" `deserialize`...

- `PublicKeyCredential.deserializeCreationOptionsFromJSON()`
- `PublicKeyCredential.deserializeRequestOptionsFromJSON()`

It's happening again... 🙈

Maybe `parse`?

- `PublicKeyCredential.parseCreationOptionsFromJSON()`
- `PublicKeyCredential.parseRequestOptionsFromJSON()`

```js
const options = PublicKeyCredential.parseRequestOptionsFromJSON({
  'rpId': 'localhost',
  'challenge': 'Ecue...5ZDE',
  'allowCredentials': [
    {
      'id': 'ASdG...om6A',
      'type': 'public-key',
      'transports': ['internal'],
    }
  ],
});
const resp = await navigator.credentials.get(options);
```

index.bs
```diff
@@ -2387,6 +2430,90 @@ Note: Invoking this method from a [=browsing context=] where the [=Web Authentic
...
+### Deserialize Authentication ceremony options - PublicKeyCredential's `getRequestOptionsFromJSON()` Methods ### {#sctn-getRequestOptionsFromJSON}
+
+<div link-for-hint="WebAuthentication/getRequestOptionsFromJSON">
+
+[=[WRPS]=] use this method to convert JSON-compatible serialized representations of options for {{CredentialsContainer/get()|navigator.credentials.get()}} to {{PublicKeyCredentialRequestOptions}}.
+
+Upon invocation, the [=client=] must take the provided options and return an identically-structured object,
+making sure to decode to {{ArrayBuffer}} those {{DOMString}} properties that are encoded with [=base64url encoding=]
+but must be {{ArrayBuffer}} values according to the definition of {{PublicKeyCredentialRequestOptions}}.
+
```
sbweeden, line 2499: Maybe I missed it somewhere else, but is there any description of how errors should be handled if, for example, the provided JSON is deemed invalid by the client? I expect they would just return via exception, however without it being defined we are open to different browser implementation decisions when an RP developer wants consistency.

The same comment applies to the `getRequestOptionsFromJSON()` method.

index.bs
```diff
@@ -2387,6 +2430,90 @@ Note: Invoking this method from a [=browsing context=] where the [=Web Authentic
...
+        required DOMString              type;
+        sequence<DOMString>             transports;
+    };
+
+    dictionary AuthenticationExtensionsClientInputsJSON {
+        DOMString                                       appid;
+        DOMString                                       appidExclude;
+        boolean                                         uvm;
+        boolean                                         credProps;
+        AuthenticationExtensionsLargeBlobInputsJSON     largeBlob;
```
sbweeden, line 2478: A challenge here is related to the handling of extensions that may not be defined within WebAuthn itself. One of the characteristics of the current specification is that extensions might be defined outside of WebAuthn, as has already been done for the [credProtect extension in CTAP2](https://fidoalliance.org/specs/fido-v2.1-ps-20210615/fido-client-to-authenticator-protocol-v2.1-ps-20210615.html#sctn-credProtect-extension), and which is not recognised or present in this IDL, yet which an RP can include in credential creation options. How do you propose to handle that?

emlun:
> I'm trying _really_ hard not to end up with another `isUserVerifyingPlatformAuthenticatorAvailable()` method name. I already sacrificed the simple "`optionsFromJSON()`" name I originally wanted. [...]
> 
> * `PublicKeyCredential.deserializeCreationOptionsFromJSON()`
> * `PublicKeyCredential.deserializeRequestOptionsFromJSON()`

How about:

- `PublicKeyCredentialCreationOptions.fromJSON()`
- `PublicKeyCredentialRequestOptions.fromJSON()`

...?

Or maybe `.parseJSON()` or `.parseFromJSON()` if any of those seems more appropriate. So a usage example would look like:

```js
navigator.credentials.create({ publicKey: PublicKeyCredentialCreationOptions.fromJSON(/* ... */) })
```

index.bs
emlun: I agree, this is especially tricky since the spec is written with the assumption that clients might just pass extension inputs through without any explicit extension processing. It seems to me that it's impossible for the `fromJSON` methods to tell which fields would need to be converted to `BufferSource`s without explicit support for the particular extension, so I'm not sure the `fromJSON` methods even _can_ be made compatible with the assumed pass-through mechanism.

index.bs
```diff
@@ -1450,6 +1451,48 @@ that are returned to the caller when a new credential is created, or a new asser
         [=extension identifier=] → [=client extension output=] entries produced by the extension's
         [=client extension processing=].
 
+    :   {{PublicKeyCredential/toJSON()}}
```
emlun, line 1454: The `toJSON()` result should also include at least `PublicKeyCredential.authenticatorAttachment` and `AuthenticatorAttestationResponse.getTransports()`, and probably also `PublicKeyCredential.getClientExtensionResults()`.

An open question is: should it also include `getAuthenticatorData()`, `getPublicKey()` and `getPublicKeyAlgorithm()` from `AuthenticatorAttestationResponse`? One can argue either way: some RPs (likely those that don't care about attestation) may prefer them, but it's unnecessary to include them for other RPs that do not need them.

index.bs
sbweeden: Maybe we should change the pass thru semantics of extensions since I don't believe there is any client that even honours that.

emlun (changes requested):

index.bs
emlun: Maybe, but I think it's been brought up before and met with objections. It's certainly far too big a decision to make here in this PR, we'd have to bring it up to the rest of the WG first, maybe with its own dedicated issue.

MasterKale: Huh, I guess I never hit send on the reply I'd typed up here. In short I'm of the opinion that we only support extensions specifically defined in this spec - anything outside of the spec, like `credProtect`, can be "manually managed by the RP". 

At least that's one way I imagined we handle the existence of extensions defined outside this spec.

emlun: I'm afraid that's not the case; [§9. WebAuthn Extensions](https://www.w3.org/TR/2021/REC-webauthn-2-20210408/#sctn-extensions) explicitly references an external registry of extensions:

>The IANA "WebAuthn Extension Identifiers" registry [[IANA-WebAuthn-Registries]](https://www.w3.org/TR/2021/REC-webauthn-2-20210408/#biblio-iana-webauthn-registries) established by [[RFC8809]](https://www.w3.org/TR/2021/REC-webauthn-2-20210408/#biblio-rfc8809) can be consulted for an up-to-date list of registered [WebAuthn Extensions](https://www.w3.org/TR/2021/REC-webauthn-2-20210408/#webauthn-extensions).

And an earlier paragraph states

>Clients wishing to support the widest possible range of extensions MAY choose to pass through any extensions that they do not recognize to authenticators, generating the [authenticator extension input](https://www.w3.org/TR/2021/REC-webauthn-2-20210408/#authenticator-extension-input) by simply encoding the [client extension input](https://www.w3.org/TR/2021/REC-webauthn-2-20210408/#client-extension-input) in CBOR. [...]

which is a normative declaration that this is a valid implementation choice supported by the spec. So I'd say we need to either be compatible with that provision or remove it, and removing it would require consensus from the rest of the working group.

dwaite:
> How about:
> 
> * `PublicKeyCredentialCreationOptions.fromJSON()`
> * `PublicKeyCredentialRequestOptions.fromJSON()`

These two types are dictionaries and not interfaces; as such they don't have an exposed name and can't define operations. 

If they had been defined as interfaces this would indeed have been the logical place. I don't know if there are ways to mitigate the compatibility breakage of such a change at this point.

index.bs
MasterKale: Good call, I've added `authenticatorAttachment`, `transports`, and `clientExtensionResults` to the output from this method.

MasterKale: As for those other `get...()` methods I'm biased for _not_ including them because I'm of the opinion that validation libraries would prefer to parse the whole `PublicKeyCredential`, so including those would be superfluous.

index.bs
```diff
@@ -1450,6 +1451,60 @@ that are returned to the caller when a new credential is created, or a new asser
...
+        representation of the instance of {{PublicKeyCredential}} suitable for submission to a [=[RP]=] server as a DOMString
+        payload with `application/json` mimetype. The [=client=] is in charge of
+        [serializing values to JSON as usual](https://webidl.spec.whatwg.org/#idl-tojson-operation), but must take additional
+        steps to encode any {{ArrayBuffer}} values to {{DOMString}} values using [=base64url encoding=] to ensure compatibility
+        with JSON syntax. {{ArrayBuffer}} values to be encoded are identified according to the detected type of the instance's
+        `response` property.
+    ::  |clientExtensionResults| below should be the output of {{PublicKeyCredential/getClientExtensionResults()}}, with any
+        {{ArrayBuffer}} values encoded to {{DOMString}} values using [=base64url encoding=]. This may include {{ArrayBuffer}}
+        values from extensions registered in the IANA "WebAuthn Extension Identifiers" registry [[!IANA-WebAuthn-Registries]]
+        but not defined in [[#sctn-extensions]].
```
MasterKale, line 1465: I tried defining this to support any extension that might be returned, but I'm uncertain if clients are expected to be aware of all registered extensions and their various values. What should happen, for example, if a client wasn't aware of `largeBlob` and received its `ArrayBuffer` output? Should it drop the extension value? Return the `ArrayBuffer`?

dwaite: The `largeBlob` extension isn't the best example since there is additional client work needed to create a response.

There are two cases defined for client extension results:

1. An extension understood by the client. This should arguably be the only option, see below
2. An authenticator extension output based on an authenticator extension input that was not understood, but instead processed by the [limited heuristic](https://www.w3.org/TR/webauthn-2/#sctn-extensions). This client extension output might be anything, but is 'suggested' to be a pure JSON version of the authenticator extension output, with no binary conversion allowed.

We may separately want to strike the heuristic for output from level 3. This is not the behavior of the extensions defined after level 1, and AFAIK only `uvm` remains with that behavior defined. Such duplication creates a second source of information for server processing which is not properly protected by the credential or any attestation.

index.bs
```diff
@@ -2387,6 +2442,84 @@ Note: Invoking this method from a [=browsing context=] where the [=Web Authentic
 
 </div>
 
+### Deserialize Registration ceremony options - PublicKeyCredential's `getCreationOptionsFromJSON()` Method ### {#sctn-getCreationOptionsFromJSON}
```
kreichgauer, line 2445: IMHO, `PublicKeyCredential.publicKeyCredentialCreationOptionsFromJSON()` would be the right choice here. (And `publicKeyCredentialRequestOptionsFromJSON` for the assertion case.) I saw the comments on the main thread, and I agree it's slightly wordy. But if nothing else, referencing the exact name makes it more obvious what this method is supposed to do, and it aids grep-ability. And most web app code bases will call this method at most once, so the benefit of a shorter name isn't that huge. :-)

index.bs
```diff
@@ -1450,6 +1451,60 @@ that are returned to the caller when a new credential is created, or a new asser
         [=extension identifier=] → [=client extension output=] entries produced by the extension's
         [=client extension processing=].
 
+    :   {{PublicKeyCredential/toJSON()}}
+    ::  This operation returns {{RegistrationResponseJSON}} or {{AuthenticationResponseJSON}}, that is a JSON-serialized
```
kreichgauer, line 1455: I'm not sure if this is accurate. JSON-serialized implies string, right? Perhaps the entire sentence could be deleted?

index.bs
kreichgauer: I think I would lean towards including the values of `getAuthenticatorData`, `getPublicKey`, and `getPublicKeyAlgorithm`.

The motivation behind these utility methods was to spare RPs from having to extract the values from the [attestation object CBOR map](https://w3c.github.io/webauthn/#attestation-object), which means they would be able to process WebAuthn responses server-side without a CBOR parser dependency. So if we didn't include these values, we would either reintroduce the CBOR dependency, or RPs would need to combine the JSON with the return values from the utility methods. I think that would be quite an unfortunate rough edge. Just including these values would probably be simpler for RPs.

Firstyear: @kreichgauer That's not possible. cbor is internal to the attestation cert formats, and needs to be "as is" for signature validation. You can't do the signature processing in JS and decode on the client side because then the JS of the page can tamper with the results and invalidate the trust chain.

So a cbor dependency will always be required for an RP.

kreichgauer: I should have said RPs that don't request attestation during the registration process. You are correct that processing attestation statements requires a CBOR parser.

index.bs
```diff
@@ -1450,6 +1451,60 @@ that are returned to the caller when a new credential is created, or a new asser
...
+        representation of the instance of {{PublicKeyCredential}} suitable for submission to a [=[RP]=] server as a DOMString
+        payload with `application/json` mimetype. The [=client=] is in charge of
+        [serializing values to JSON as usual](https://webidl.spec.whatwg.org/#idl-tojson-operation), but must take additional
+        steps to encode any {{ArrayBuffer}} values to {{DOMString}} values using [=base64url encoding=] to ensure compatibility
```
kreichgauer, line 1459: I assume we don't *actually* want Base64URL here but are using it for consistency only. I.e. We're not expecting developers want to put these json strings into URL parameters, right?

Would it be worth to use regular Base64 in order to simplify use of standard server-side JSON parsers? Go's encoding/json for example automatically encodes byte slices into base64 (not base64url) strings ([see docs here](https://pkg.go.dev/encoding/json#:~:text=Array%20and%20slice%20values%20encode%20as%20JSON%20arrays%2C%20except%20that%20%5B%5Dbyte%20encodes%20as%20a%20base64%2Dencoded%20string%2C%20and%20a%20nil%20slice%20encodes%20as%20the%20null%20JSON%20value.). So going with Base64url could potentially be a minor annoyance for server developers (extra code to write; potential source for subtle bugs).

The obvious drawback is that the spec already uses Base64URL elsewhere, so mixing the two could be mildly confusing. Though we could call out the difference and explain the reasoning behind it if we wanted.

index.bs
MasterKale: > JSON-serialized implies string, right?

Does it? If you tell a JS dev "this outputs JSON" I'll bet they're going to think they'll get back an `object` instead of a `string`. If what you're saying is true then I'll need to adjust this wording to reflect that the output should be an `object`. Something like:

> This operation returns...an `object` that can be fed to `JSON.stringify()` to get back a string that can be sent to the RP as `"application/json"` without mutilating any `ArrayBuffer`/`Uint8Array`'s that might be present in the original `object`.

index.bs
```diff
@@ -1450,6 +1451,60 @@ that are returned to the caller when a new credential is created, or a new asser
...
+        DOMString type;
+    };
+
+    dictionary AuthenticatorAttestationResponseJSON {
+        Base64URLString clientDataJSON;
+        Base64URLString attestationObject;
+        sequence<DOMString> transports;
+    };
+
+    dictionary AuthenticationResponseJSON {
```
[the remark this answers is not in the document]
MasterKale: It's frustrating to me that `PublicKeyCredential` has been overloaded to mean two possible data structures, since `response` can be one of two types. I think this is a current developer pain point, that there aren't subclasses of `PublicKeyCredential` coming out of `.create()` and `.get()` to be uniquely typed. That's why I've been trying to emphasize the idea of "`RegistrationResponse`" and "`AuthenticationResponse`" here, to get to a 1:1 mapping between `navigator.credentials` method and return type.

To the suggestion that a `PublicKeyCredentialJSON` gets defined as a union type, is it a compromise to make it easier for the client to implement this feature? I personally want to see these additions err on the side of better _developer_ usability, and necessarily see this as hoisting some of this complexity onto the client. The aim is to make it _easier_ for devs to consume WebAuthn, and to achieve this I think we should be more straightforward/explicit with return types.

index.bs
MasterKale: > The obvious drawback is that the spec already uses Base64URL elsewhere, so mixing the two could be mildly confusing.

This is exactly why I've chosen to use Base64URL across multiple implementations despite Base64 being workable without external dependencies or custom code, because [the spec lists it as a dependency](https://www.w3.org/TR/webauthn-2/#sctn-dependencies).

> We're not expecting developers want to put these json strings into URL parameters, right?

Definitely not, I'm almost certain the output would be conveyed in the body of a POST request.

> Though we could call out the difference and explain the reasoning behind it if we wanted.

I think it's asking for a bad time if we mix-and-match encodings, and the ship has probably sailed for switching WebAuthn to use Base64 instead (I agree with you that more languages support Base64 out of the box and probably would have advocated for Base64 had I been fortunate to be involved in those early days.) I'd argue requiring RP's to keep track of when to use Base64 vs Base64URL to encoding or decoding would push this feature the other way, towards making WebAuthn more challenging to use.

index.bs
kreichgauer: So the `toJSON` spec says the method returns a "JSON type" (object, array, string, number, etc.) and calls the output of JSON.stringify a "JSON string". Perhaps we can reuse the established language? 

> This operation converts the `PublicKeyCredential` into a `PublicKeyCredentialJSON`, which is a [JSON type](https://webidl.spec.whatwg.org/#dfn-json-types). In ECMAScript, instances of `PublicKeyCredentialJSON` can be converted into a JSON string with the [JSON.stringify() method](https://262.ecma-international.org/12.0/#sec-json.stringify).

index.bs
emlun: I've opened an issue to discuss dropping generic extension pass-through: #1730

But on second thought, a possible work-around to this issue is to specify a way to flag binary values. For example, MongoDB Extended JSON encodes binary values as `{ "$binary": "<base64>" }`. Or we could say that any value with a particular prefix in the property name be interpreted as binary - for example, `$BufferSource_base64url_challenge` would be base64url-decoded and renamed to `challenge` in the output.

It feels quite hacky and it's not well standardized, but it could enable this to be compatible with generic extension pass-through.

index.bs
```diff
@@ -2387,6 +2442,84 @@ Note: Invoking this method from a [=browsing context=] where the [=Web Authentic
...
+{{CredentialsContainer/create()|navigator.credentials.create()}} to {{PublicKeyCredentialCreationOptions}}.
+
+Upon invocation, the [=client=] must take the provided options and return an identically-structured object,
+making sure to decode to {{ArrayBuffer}} those {{DOMString}} properties that are encoded with [=base64url encoding=]
+but must be {{ArrayBuffer}} values according to the definition of {{PublicKeyCredentialCreationOptions}}.
+
+{{AuthenticationExtensionsClientInputsJSON}} may include extensions registered in the IANA
+"WebAuthn Extension Identifiers" registry [[!IANA-WebAuthn-Registries]] but not defined in [[#sctn-extensions]].
+In either case the [=client=] must decode to {{ArrayBuffer}} those {{DOMString}} values that are encoded with
+[=base64url encoding=] but must be {{ArrayBuffer}} values according to the extension's definition.
```
emlun, line 2459: I think this could be a bit more rigorous, how about this?

```suggestion
[=[WRPS]=] use this method to convert [=JSON type=] representations of options for
{{CredentialsContainer/create()|navigator.credentials.create()}} into {{PublicKeyCredentialCreationOptions}}.

Upon invocation, the [=client=] MUST convert the {{PublicKeyCredential/getCreationOptionsFromJSON(options)/options}} argument
into a new, identically-structured {{PublicKeyCredentialCreationOptions}} object,
using [=base64url encoding=] to decode any {{DOMString}} attributes in {{PublicKeyCredentialCreationOptionsJSON}}
that correspond to [=buffer source type=] attributes in {{PublicKeyCredentialCreationOptions}}.
This conversion MUST extend to any [=client extension inputs=] processed by the [=client=].

{{AuthenticationExtensionsClientInputsJSON}} MAY include extensions registered in the IANA
"WebAuthn Extension Identifiers" registry [[!IANA-WebAuthn-Registries]] but not defined in [[#sctn-extensions]].
```

This honestly still feels a bit too hand-wavy to me, but it could be a step in the right direction?

index.bs
```diff
@@ -2387,6 +2442,84 @@ Note: Invoking this method from a [=browsing context=] where the [=Web Authentic
...
+
+### Deserialize Authentication ceremony options - PublicKeyCredential's `getRequestOptionsFromJSON()` Methods ### {#sctn-getRequestOptionsFromJSON}
+
+<div link-for-hint="WebAuthentication/getRequestOptionsFromJSON">
+
+[=[WRPS]=] use this method to convert JSON-compatible serialized representations of options for {{CredentialsContainer/get()|navigator.credentials.get()}} to {{PublicKeyCredentialRequestOptions}}.
+
+Upon invocation, the [=client=] must take the provided options and return an identically-structured object,
+making sure to decode to {{ArrayBuffer}} those {{DOMString}} properties that are encoded with [=base64url encoding=]
+but must be {{ArrayBuffer}} values according to the definition of {{PublicKeyCredentialRequestOptions}}.
```
emlun, line 2504: Ditto here:

```suggestion
[=[WRPS]=] use this method to convert [=JSON type=] representations of options for
{{CredentialsContainer/get()|navigator.credentials.get()}} into {{PublicKeyCredentialRequestOptions}}.

Upon invocation, the [=client=] MUST convert the {{PublicKeyCredential/getRequestOptionsFromJSON(options)/options}} argument
into a new, identically-structured {{PublicKeyCredentialRequestOptions}} object,
using [=base64url encoding=] to decode any {{DOMString}} attributes in {{PublicKeyCredentialRequestOptionsJSON}}
that correspond to [=buffer source type=] attributes in {{PublicKeyCredentialRequestOptions}}.
This conversion MUST extend to any [=client extension inputs=] processed by the [=client=].

{{AuthenticationExtensionsClientInputsJSON}} MAY include extensions registered in the IANA
"WebAuthn Extension Identifiers" registry [[!IANA-WebAuthn-Registries]] but not defined in [[#sctn-extensions]].
```

emlun (changes requested):

MasterKale:
During the F2F: #1730 may be a blocker to this, the trick is how do we handle inputs for extensions not defined in the spec? Perhaps around this new functionality we link to the extensions registry and define instructions for how to massage JSON input for those extensions into their appropriate data structures.

index.bs
MasterKale: @emlun helped craft verbiage below that reflected @kreichgauer's suggestion to match established language.

index.bs
MasterKale: This was resolved by defining `AuthenticationExtensionsClientInputsJSON` as an empty dictionary, and then leaving instructions for clients on how to handle any present values (this is for `.create()`, something similar is currently defined for `.get()` too):

> Upon invocation, the [client](http://localhost:8000/#client) MUST convert the [options](http://localhost:8000/#dom-publickeycredential-parsecreationoptionsfromjson-options-options) argument into a new, identically-structured [PublicKeyCredentialCreationOptions](http://localhost:8000/#dictdef-publickeycredentialcreationoptions) object, using [base64url encoding](http://localhost:8000/#base64url-encoding) to decode any [DOMString](https://webidl.spec.whatwg.org/#idl-DOMString) attributes in [PublicKeyCredentialCreationOptionsJSON](http://localhost:8000/#dictdef-publickeycredentialcreationoptionsjson) that correspond to [buffer source type](https://webidl.spec.whatwg.org/#dfn-buffer-source-type) attributes in [PublicKeyCredentialCreationOptions](http://localhost:8000/#dictdef-publickeycredentialcreationoptions). This conversion MUST extend to any [client extension inputs](http://localhost:8000/#client-extension-input) processed by the [client](http://localhost:8000/#client).
> 
> [AuthenticationExtensionsClientInputsJSON](http://localhost:8000/#dictdef-authenticationextensionsclientinputsjson) MAY include extensions registered in the IANA "WebAuthn Extension Identifiers" registry [[IANA-WebAuthn-Registries]](http://localhost:8000/#biblio-iana-webauthn-registries) but not defined in [§ 9 WebAuthn Extensions](http://localhost:8000/#sctn-extensions).

index.bs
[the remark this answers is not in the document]
MasterKale: > To the suggestion that a `PublicKeyCredentialJSON` gets defined as a union type, is it a compromise to make it easier for the client to implement this feature? I personally want to see these additions err on the side of better _developer_ usability, and necessarily see this as hoisting some of this complexity onto the client.

@kreichgauer It's been a few weeks, what do you currently think about my response above to your `PublicKeyCredentialJSON` suggestion?

index.bs
```diff
@@ -1450,6 +1451,62 @@ that are returned to the caller when a new credential is created, or a new asser
...
+        [serializing values to JSON as usual](https://webidl.spec.whatwg.org/#idl-tojson-operation),
+        but MUST take additional steps to first encode any {{ArrayBuffer}} values to {{DOMString}} values
+        using [=base64url encoding=] to ensure compatibility with JSON syntax.
+        {{RegistrationResponseJSON/clientExtensionResults|RegistrationResponseJSON.clientExtensionResults}} or
+        {{AuthenticationResponseJSON/clientExtensionResults|AuthenticationResponseJSON.clientExtensionResults}}
+        members MUST be set to the output of {{PublicKeyCredential/getClientExtensionResults()}},
+        with any {{ArrayBuffer}} values encoded to {{DOMString}} values using
+        [=base64url encoding=]. This MAY include {{ArrayBuffer}} values from extensions registered
+        in the IANA "WebAuthn Extension Identifiers" registry [[!IANA-WebAuthn-Registries]] but not
+        defined in [[#sctn-extensions]].
```
emlun, line 1467: I think we also need the same instruction for `transports`? Then let's put both that and the above "The clientExtensionResult member ..." each in its own paragraph to make it easier to read.

```suggestion
        defined in [[#sctn-extensions]].

        The {{AuthenticatorAttestationResponseJSON/transports|AuthenticatorAttestationResponseJSON.transports}}
        member MUST be set to the output of {{AuthenticatorAttestationResponse/getTransports()}}.
```

MasterKale:
@emlun Thank you for the verbiage tweaks! Those are all incorporated now, one step closer to landing this thing 💪

index.bs
MasterKale: I added the following to both sections, if we hash it out here I'll make the same changes in the other section:

> If the [=client=] encounters any issues parsing any of the [=JSON type=] representations then it
MUST return an error code equivalent to "{{EncodingError}}" with a description of the incompatible
value and terminate the operation.

dagnelies:
This is already a great step forward. 

As far as I understand, it is converting *all* ArrayBuffer objects into base64url. I wonder if it would make sense to add some more nuanced conversions.

**For `credentials.create`**


In particular, I would view 2 exceptions:

- The `response.clientDataJSON` contains a raw JSON encoded string. So it would be better to parse this ArrayBuffer directly (like `{"type":"webauthn.create","challenge":"ZmFrZS1pZC05ZDFqazgwa251NA","origin":"http://localhost:63342","crossOrigin":false}` instead or re-encoding this string into base64url.
- The `response.userHandle` contains the plain text "user.id" like `my-john-doe-id` and here also the raw value holds more value than its base64url encoded variant.

dwaite:
> * During both creation and authentication, the `response.clientDataJSON` contains a raw JSON encoded string. So it would be better to parse this ArrayBuffer directly (like `{"type":"webauthn.create","challenge":"ZmFrZS1pZC05ZDFqazgwa251NA","origin":"http://localhost:63342","crossOrigin":false}` instead or re-encoding this string into base64url.

Speaking to this one, the `clientDataJSON` is part of the signed response, so any changes to white space, ordering, or Unicode canonicalization will ruin the response. Hence it is sent as binary rather than as a JSON structure or stringified JSON text.

dagnelies:
@dwaite I'm aware of that. I was indeed referring to a raw JSON String, not the parsed object. I think it would be rather trivial to utf-8 encode it on the server side to get the correct byte array for signature verification. It just makes things more readable IMHO.

dwaite:
> @dwaite I'm aware of that. I was indeed referring to a raw JSON String, not the parsed object. I think it would be rather trivial to utf-8 encode it on the server side (instead of base64 decoding it) to get the correct byte array for signature verification.

Unicode normalization forms would be the issue here, not utf-8 encoding (I would expect the JSON to already be UTF-8 encoded).

The only way this works AFAIK would be if the client and the server (either through normalization step or intermediary transport guarantees) represented that string in Unicode normalization form C, for the purposes of re-creating the byte stream accurately from the unicode text.

I personally don't think such a requirement is worth saving a few bytes on the wire.

dagnelies:
Hmmm... I did not expect unicode normalization to be a potential issue. I expected that what you send from the browser and what you obtain server side to always be identical. But you are right, in case of doubts and to avoid middleware issues, it might be better to send base64. That seems more bullet proof. ... So everything base64 it shall be😉 ... The single leftover exception is not worth it either.

Firstyear:
> Speaking to this one, the `clientDataJSON` is part of the signed response, so any changes to white space, ordering, or Unicode canonicalization will ruin the response. Hence it is sent as binary rather than as a JSON structure or stringified JSON text.

But it's sent / received into a Uint8Array which is the problem, so when you change it between bytes to base64 you aren't actually affecting or changing any of these things.

emlun:
> * During authentication, the `response.userHandle` contains the plain text "user.id" like `my-john-doe-id` and here also the raw value holds more value than its base64url encoded variant.

The spec states that "The [user handle](https://www.w3.org/TR/2021/REC-webauthn-2-20210408/#user-handle) MUST NOT contain personally identifying information about the user, such as a username or e-mail address" and "It is RECOMMENDED to let the [user handle](https://www.w3.org/TR/2021/REC-webauthn-2-20210408/#user-handle) be 64 random bytes". We should not add features to support a use case we actively recommend against.

I think `clientDataJSON` should also stay base64url-encoded to keep the emphasis that it needs to be conveyed byte-identically to the value that was signed. More generally I think this JSON variant of the API should mirror the BufferSource variant as closely as possible and not make any structural or semantic changes to the data, otherwise the mismatch will be confusing.

I guess that could be an argument against splitting the root object into two separate types for registration and authentication, but I think that one is fair enough since that is likely how most developers already think of it anyway.

dagnelies:
Agreed

MasterKale:
I haven't chimed in because @dwaite and @emlun beat me to my responses. I'm glad to see my intuition was right about how best to handle those values 🙂

Most of the feedback has been handled on my end. The one outstanding change suggestion that I know about right now is the suggestion to include directions for what to do if bad data is passed into these objects, namely throwing an `EncodingError`. Can someone please look over my addition that attempts to address that? This thing is so close to being ready, I'm very excited 😅

emlun (changes requested):
Apart from a couple more nits I think this seems good to go for now. I would like to be a bit more rigorous, but I can't see how to do it without essentially spelling out the entire conversions, which seems too obtuse and fragile. This is probably unambiguous enough, we can revise later if we think of a better way to express it. :slightly_smiling_face:

emlun (approved):
Thanks a lot @MasterKale for taking this on!

index.bs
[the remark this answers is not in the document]
MasterKale: > I guess it depends on whether you think the hypothetical WebAuthn developer would first look at the PublicKeyCredential interface IDL or not? If they did, comparing the two and realizing that they're different in weird subtle ways (beyond the binary->base64url conversion) I think could be a bit puzzling? If on the other hand, we'd assume the hypothetical developer would only ever see the JSON IDL, I agree what you propose would be simpler to grok.

@kreichgauer Can we leave comments in IDL definitions? I want to believe a developer who doesn't want to get too deep into things will jump to these JSON methods, but it's probably more realistic that they'd stumble on `PublicKeyCredential` first and then the JSON IDL. If we could leave comments in the JSON IDL explaining why there are split types then perhaps we can have the best of both worlds.

MasterKale:
I like to resolve all comments in my PR's before the changes land, so I'm highlighting the last unresolved comment on this PR. It's a conversation @kreichgauer and I are having on a comment made back in April (we're nearly 100 comments on this 😱): https://github.com/w3c/webauthn/pull/1703#discussion_r902152703

Martin mentioned this:

> I agree with the desire to aid developers as much as we can, and I agree that the structure of the API with create() and get() returning the same interface type but encapsulating two different response dictionary types is…awkward. (It's a side effect of having WebAuthn bolted on top of Credential Management, but our developers/users shouldn't care about that.) However, I'm kind of on the fence whether I find the separate inheritance hierarchy for JSON serialization more or less confusing. I guess it depends on whether you think the hypothetical WebAuthn developer would first look at the PublicKeyCredential interface IDL or not? If they did, comparing the two and realizing that they're different in weird subtle ways (beyond the binary->base64url conversion) I think could be a bit puzzling? If on the other hand, we'd assume the hypothetical developer would only ever see the JSON IDL, I agree what you propose would be simpler to grok.

And I responded: 

> Can we leave comments in IDL definitions? I want to believe a developer who doesn't want to get too deep into things will jump to these JSON methods, but it's probably more realistic that they'd stumble on `PublicKeyCredential` first and then the JSON IDL. If we could leave comments in the JSON IDL explaining why there are split types then perhaps we can have the best of both worlds.

I'm interested in seeing what others might think about what we're discussing. It might be more useful to continue hashing this out down here, though, since it's easier to find. When some kind of decision is reached I can address whatever might need to change, then go back up into the comment and resolve it.

index.bs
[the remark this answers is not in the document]
emlun: Yes, [IDL can contain comments](https://webidl.spec.whatwg.org/#prod-comment). The syntax is the same as in JavaScript.

sbweeden (approved):

kreichgauer (approved):

PR #1704, opened 2022-02-24
Change UV check step from passive to active voice

### LINKED ISSUE #1699: Clarification needed regarding meaning of "user verification is required"
Step 15 of [Registering a new Credential](https://www.w3.org/TR/webauthn-2/#sctn-registering-a-new-credential) mentions

> 15. If [user verification](https://www.w3.org/TR/webauthn-2/#user-verification) is required for this registration, verify that the [User Verified](https://www.w3.org/TR/webauthn-2/#concept-user-verified) bit of the [flags](https://www.w3.org/TR/webauthn-2/#flags) in authData is set.

this should maybe be changed to this instead

> 15. If options.[authenticatorSelection](https://www.w3.org/TR/webauthn-2/#dom-publickeycredentialcreationoptions-authenticatorselection).[userVerification](https://www.w3.org/TR/webauthn-2/#dom-authenticatorselectioncriteria-userverification) is set to [required](https://www.w3.org/TR/webauthn-2/#dom-userverificationrequirement-required), verify that the [User Verified](https://www.w3.org/TR/webauthn-2/#concept-user-verified) bit of the [flags](https://www.w3.org/TR/webauthn-2/#flags) in authData is set.

---

Similarly with step 17 of [Verifying an Authentication Assertion](https://www.w3.org/TR/webauthn-2/#sctn-verifying-assertion):

> 17. If [user verification](https://www.w3.org/TR/webauthn-2/#user-verification) is required for this assertion, verify that the [User Verified](https://www.w3.org/TR/webauthn-2/#concept-user-verified) bit of the [flags](https://www.w3.org/TR/webauthn-2/#flags) in authData is set.

should maybe be this instead:

> 17. If options.[userVerification](https://www.w3.org/TR/webauthn-2/#dom-publickeycredentialrequestoptions-userverification) is set to [required](https://www.w3.org/TR/webauthn-2/#dom-userverificationrequirement-required), verify that the [User Verified](https://www.w3.org/TR/webauthn-2/#concept-user-verified) bit of the [flags](https://www.w3.org/TR/webauthn-2/#flags) in authData is set.

--- 

And if that shouldn't be done, how is "user verification is required" different from the respective `userVerification` being set to `required`?

emlun:
Fixes #1699.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/pull/1704.html" title="Last updated on Feb 24, 2022, 4:09 PM UTC (43aeea6)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/1704/dcd3190...43aeea6.html" title="Last updated on Feb 24, 2022, 4:09 PM UTC (43aeea6)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -4577,8 +4577,8 @@ In order to perform a [=registration ceremony=], the [=[RP]=] MUST proceed as fo
 
 1. Verify that the [=User Present=] bit of the <code>[=flags=]</code> in |authData| is set.
 
-1. If [=user verification=] is required for this registration, verify that the [=User Verified=] bit of the <code>[=flags=]</code>
-    in |authData| is set.
+1. If the [=[RP]=] requires [=user verification=] for this registration,
+    verify that the [=User Verified=] bit of the <code>[=flags=]</code> in |authData| is set.
 
 1. Verify that the "alg" parameter in the [=credentialPublicKey|credential public key=] in |authData|
     matches the {{PublicKeyCredentialParameters/alg}} attribute of one of the [=list/items=] in
@@ -4744,8 +4744,8 @@ a numbered step. If outdented, it (today) is rendered as a bullet in the midst o
 
 1. Verify that the [=User Present=] bit of the <code>[=flags=]</code> in |authData| is set.
 
-1. If [=user verification=] is required for this assertion, verify that the [=User Verified=] bit of the <code>[=flags=]</code> in
-    |authData| is set.
+1. If the [=[RP]=] requires [=user verification=] for this assertion,
+    verify that the [=User Verified=] bit of the <code>[=flags=]</code> in |authData| is set.
 
 1. Verify that the values of the [=client extension outputs=] in |clientExtensionResults| and the [=authenticator extension
     outputs=] in the <code>[=authdataextensions|extensions=]</code> in |authData| are as expected, considering the [=client


equalsJeffH (approved):
LGTM

Infinisil:
I'm still not sure what difference there is between "the RP requires user verification" and `userVerification` being set to "required". As a Relying Party implementor I'd still interpret the former as the latter.

In which scenario is it possible that these two are different?
- The RP requires user verification but doesn't set `userVerification` to "required": That seems unreasonable, the RP shouldn't reject a response without user verification if it didn't ask for it
- The RP doesn't require user verification but sets `userVerification` to "required": This sounds like the RP really should've set `userVerification` to "preferred" and not "required" then.

Is there a use case that I'm not seeing? If there really is a reasonable one then I think it makes sense to point that out in the spec somewhere, so that library implementors don't equate these things.

sbweeden:
> I'm still not sure what difference there is between "the RP requires user verification" and `userVerification` being set to "required". As a Relying Party implementor I'd still interpret the former as the latter.

Consider that the RP has determined that for this creation or authentication that it wants UV.  It sends a page to the browser with JS calling WebAuthn, and uv=required in the options. The RP cannot be certain at this point that the client (i.e. browser) actually then uses these options unmodified in the call to the authenticator. There can be reasons for this, such as:
 - a nefarious user who has acquired a victims authenticator but not the PIN might try to use it by intercepting the call in a browser debugger and modifying the JS.
 - The client might have a browser plugin that does same

As a result, the RP should never "trust" that the options it sent to the browser were actually used in communications with the authenticator. Instead it should keep a "copy of its policy", which might be a local copy of the options it intended the client to use, and always verify the response from the authenticator against that. 


> In which scenario is it possible that these two are different?

When the options are modified at the browser, after being transmitted in a page with JS from the RP, but before being used in communications with the authenticator.

> * The RP requires user verification but doesn't set `userVerification` to "required": That seems unreasonable, the RP shouldn't reject a response without user verification if it didn't ask for it
> * The RP doesn't require user verification but sets `userVerification` to "required": This sounds like the RP really should've set `userVerification` to "preferred" and not "required" then.
> 
> Is there a use case that I'm not seeing? If there really is a reasonable one then I think it makes sense to point that out in the spec somewhere, so that library implementors don't equate these things.

I think the important thing here is that what the RP thinks it sets as options needs to be policy/state stored at the RP and that responses are verified against that. How the RP maintains that policy is likely going to be a copy of the options sent to the client, but doesn't have to be.

This is why the proposed language is used - to indicate the intent of the RP, rather than specifically saying what options were used in the actual WebAuthn call at the browser.

Given this view of things, is your concern now addressed, or would you propose different text?

Infinisil:
> As a result, the RP should never "trust" that the options it sent to the browser were actually used in communications with the authenticator. Instead it should keep a "copy of its policy", which might be a local copy of the options it intended the client to use, and always verify the response from the authenticator against that.

That makes a lot of sense, but the current specification already presumes that `options` were stored by the server, because it uses them for verification e.g. in https://w3c.github.io/webauthn/#sctn-registering-a-new-credential:

> 8. Verify that the value of C.[`challenge`](https://w3c.github.io/webauthn/#dom-collectedclientdata-challenge) equals the base64url encoding of options.[`challenge`](https://w3c.github.io/webauthn/#dom-publickeycredentialcreationoptions-challenge).

Or

> 15. Verify that the "alg" parameter in the [credential public key](https://w3c.github.io/webauthn/#credentialpublickey) in authData matches the [alg](https://w3c.github.io/webauthn/#dom-publickeycredentialparameters-alg) attribute of one of the [items](https://infra.spec.whatwg.org/#list-item) in options.[pubKeyCredParams](https://w3c.github.io/webauthn/#dom-publickeycredentialcreationoptions-pubkeycredparams).

If these `options` really can't be trusted, then all the steps in [WebAuthn Relying Party Operations](https://w3c.github.io/webauthn/#sctn-rp-operations) making use of `options` for verification should be rewritten to not use it, and instead use the same language as it already does for user verification.

On the other hand, if we can assume that the `options` referenced in the specification can be trusted (maybe by pointing out that it should be temporarily stored by the server for later verification, but I think this is already clear when wanting to implement the spec), then I think the current wording of these steps makes sense, but then I also believe the user verification step should be rewritten to reference `options`:

> 14. If options.[authenticatorSelection](https://w3c.github.io/webauthn/#dom-publickeycredentialcreationoptions-authenticatorselection).[userVerification](https://w3c.github.io/webauthn/#dom-authenticatorselectioncriteria-userverification) is set to [required](https://w3c.github.io/webauthn/#dom-userverificationrequirement-required), verify that the [User Verified](https://w3c.github.io/webauthn/#concept-user-verified) bit of the [flags](https://w3c.github.io/webauthn/#flags) in authData is set.

sbweeden (approved):
LGTM

sbweeden:
Agreed to merge on 2022-04-06 WG call.

sbweeden:
The consensus on the WG call was that the PR was acceptable. You may always propose another PR via the existing process.

Issue #1577, opened 2021-02-24
Support for remote desktops


agl:
Initially filing this as a placeholder for the level three charter:

As WebAuthn becomes more common (yay!) the need to support remote desktop products becomes more salient. This is delicate because a remote desktop violates the proximity assurances of WebAuthn, but we'll struggle to save the world from passwords if we don't support them.

Thus in level three we may wish to consider things like additional CollectedClientData fields for this and I wouldn't want charter questions to exclude that.

MasterKale:
This'll be an interesting issue to solve. I can't imagine a browser running in a VM is going to have any insight into the fact that USB passthrough is being used to expose a roaming authenticator (plugged into the client running remote desktop) to it for it to report that back within `clientDataJSON`...

agl:
The rough design that we have in mind would avoid those issues, although we're only flagging it for the charter for now.

serianox:
I think that this is already working out of the box on Windows 10 for NFC authenticators (see https://docs.microsoft.com/en-us/windows/security/identity-protection/smart-cards/smart-card-and-remote-desktop-services).

kzu:
Definitely not working with Okta/webauthn via Remote Desktop (never even get the prompt, have both USB yubikey + windows hello configured for the account):

![image](https://user-images.githubusercontent.com/169707/142264231-1679b132-2dc4-4602-aad6-091d5e435474.png)

serianox:
Tested, and working out of the box with remote desktop in NFC. In HID, required to configure forwarding of the USB device first. Both local and remote station were running Windows 10.

In case the remote desktop protocol doesn't forward PCSC, I suppose that forwarding the CCID is enough to get NFC working, in the same fashion as forwarding USB for HID.
![Capture](https://user-images.githubusercontent.com/3913502/142506167-e3b1111b-f7df-4e7b-b290-e04de0d3415c.PNG)

seism0saurus:
Hi @serianox,
Could you explain how you did it with USB HID?
I use RDP and tried redirection of the FIDO token device with RemoteFX, Donglify and USB over Ethernet. All three make the token visible in the device manager and I can access it in the Windows security token configuration. But every website gives me the same error as kzu showed.

> Tested, and working out of the box with remote desktop in NFC. In HID, required to configure forwarding of the USB device first. Both local and remote station were running Windows 10.
> 
[...]

PR #1717, opened 2022-04-05
Add explainer for remote desktop support

kreichgauer:
This is meant to serve as an [explainer](https://tag.w3.org/explainers/) for WebAuthn remote desktop support that we intend to experiment with in Chrome.

(Remote desktop support is already being tracked as #1577.)


kreichgauer:
agl points out our other explainers are in the [wiki

docs/remote_desktop_explainer.md
```diff
@@ -0,0 +1,53 @@
...
+        // behalf of https://accounts.example.com.
+        remoteDesktopClientOverride: {
+            origin: 'https://accounts.example.com',
+            sameOriginWithAncestor: false,
+        },
+    },
+}});
+```
+
+The two extension members, `origin` and `sameOriginWithAncestors`, replace the arguments of the same names when invoking `PublicKeyCredential`'s `[[[DiscoverFromExternalSource]]` internal method](https://w3c.github.io/webauthn/#dom-publickeycredential-discoverfromexternalsource-slot). This causes the user agent to match the scope of the RP ID against the overridden origin (https://accounts.example.com), rather than the origin of the Relying Party (https://myrdc.example). The supplied values are also used to assemble the `[CollectedClientData` dictionary](https://w3c.github.io/webauthn/#dictionary-client-data) that the authenticator signs over when generating an assertion. (The remaining `CollectedClientData` members can be inferred from context and therefore don't need to be injected.) To the Relying Party, the response should be indistinguishable from one that would have been generated without forwarding.
```
emlun, line 41: A few square brackets are misaligned...
```suggestion
The two extension members, `origin` and `sameOriginWithAncestors`, replace the arguments of the same names when invoking [`PublicKeyCredential`'s `[[[DiscoverFromExternalSource]]` internal method](https://w3c.github.io/webauthn/#dom-publickeycredential-discoverfromexternalsource-slot). This causes the user agent to match the scope of the RP ID against the overridden origin (https://accounts.example.com), rather than the origin of the Relying Party (https://myrdc.example). The supplied values are also used to assemble the [`CollectedClientData` dictionary](https://w3c.github.io/webauthn/#dictionary-client-data) that the authenticator signs over when generating an assertion. (The remaining `CollectedClientData` members can be inferred from context and therefore don't need to be injected.) To the Relying Party, the response should be indistinguishable from one that would have been generated without forwarding.
```

docs/remote_desktop_explainer.md
```diff
@@ -0,0 +1,53 @@
...
+
+Extension processing would be analogous for WebAuthn create() calls: Presence of the extension causes the equivalent arguments to the `[[[Create]]` internal method](https://w3c.github.io/webauthn/#sctn-createCredential) to be overridden, such that the request is processed as if it had been made by the Relying Party origin on the remote host.
```
emlun, line 43:
```suggestion
Extension processing would be analogous for WebAuthn create() calls: Presence of the extension causes the equivalent arguments to the [`[[Create]]` internal method](https://w3c.github.io/webauthn/#sctn-createCredential) to be overridden, such that the request is processed as if it had been made by the Relying Party origin on the remote host.
```

kreichgauer:
WG call on Apr 6 decided that we close this issue and move the explainer to the wiki.

Issue #1618, opened 2021-05-31
Identify which items in creation and assertion options are client UI/UX hints


Firstyear:
A large number of items in the creation and assertion options exist to help a client identify and a select an authenticator for usage. However, many of these items are expressed using language that indicates to a user that these are enforced policy or a requirement. This leads to situations where people may rely on these options, and a client may freely alter or tamper with them without the RP knowing.

The options that should be clearly identified as UX hints, and not enforced policies are:

-- Creation

* PublicKeyCredentialParameters (verified in level 2)
* excludeCredentials (intended for the client to allow filtering of available credentials, reads like it is a "deny list")
* AuthenticatorSelectionCriteria (criteria implies a strict requirement of what authenticators can be selected by the client, even though it's a hint for what authenticators could be used)
* All members of AuthenticatorSelectionCriteria 

-- Assertion

* userVerification (this is a client UX hint, and may not correlate to the authenticators uv state) 

Relates #1615

emlun:
[§2. Conformance][conf] states that "A User Agent MUST behave as described by § 5 Web Authentication API in order to be considered conformant", so many of these behaviours are in fact REQUIRED of conformant clients and not just optional hints. But this could motivate some expanded discussion of the implications when the client or authenticator does not conform to those requirements, and in which cases nonconformance can be detected. I'm not sure whether that would belong in [§2. Conformance][conf], [§13. Security Considerations][sec-cons] or inline with each parameter definition etc.

Either way, I agree that many of these descriptions can be tightened up to better describe how to properly use them and/or what they will actually do (assuming the client and authenticator are conformant).

[conf]: https://w3c.github.io/webauthn/#sctn-conformance
[sec-cons]: https://w3c.github.io/webauthn/#sctn-security-considerations

Firstyear:
When I say they are hints, the javascript on the page if compromised or modified can alter the content of this structure before it is sent to the clients navigator.credentials api. From the "output" of the navigator.credentials call it is not possible to assert what was the state of the input that went into that api was, and if it was or was not tampered with, and so it's not possible to know if these requirements were upheld. However, to the RP, the language *looks* like the client *must* enforce these options, even though there is a possibility of alteration.

For example, on one IDM as a service site, they incorrectly use a combination of discouraged/preferred UV in registration/login, so your authenticator will request UV even though it is not actually validated, so I have a local js that disables that and sets it to always be "discouraged". 

[...]

Issue #1556, opened 2021-01-28
Note to encourage storage of registered credential details


Firstyear:
during registration today, there are a number of parameters related to the credential that *could* be stored for future verification and policy assertion, but it has been up to RP implementors to interpret and discover this. A note should exist in registration that encourages the storage of:

* uv bit
* attestation data (if performed) so that the model/transport of the credential is known
* extension data

This allows RP's to then extend the authentication ceremony such that a credential known to set a uv bit in registration with "preferred" can have the next authentication performed with "userVerification required" as it is know it can provide that. Similar, this data can be used for validation of other credential properties that may matter to policy for RP's. 

By having a note encouraging this, this allows RP's to understand that this is how certain policies can be implemented. 

Similar notes should exist around authenticatorAttachement that these are *hints* to the client on how to work with the authenticator and are not the foundation of security policy - it is based on the credential used and the RP's knowledge of that credential from registartion to enforce any policy.

lgarron:
Is it possible to make this easier for RPs to do correctly, by having the API return a single/field object with all the stuff you "should" store?

To some extent, saving the attestation data used to be a sufficiently effective way to store "stuff you might need later". But I imagine that many RPs will not realize they should have stored more info, unless it is the simple path. Given that the extension data is not returned by default (although [we always ask for it](https://github.com/github/webauthn-json)), it seems very easy to overlook. Bundling it with the attestation data/UV bit/anything else important into an "attestation++" object would make this more ergonomic.

dwaite:
The other way of thinking about this is that you have your credential creation request and response, and you are choosing what information _not_ to save.

Sure, the request message you sent or the client credential data in the response might be something you won't use normally at runtime, but if you save them you can re-run your security evaluation on the initial credential registration at any point in the future, such as after a security fix or change to your attestation policy.

Once you decide to pick and choose what data you save, there aren't a lot of values that MUST be saved across all the different kinds of implementations. In fact, I believe a limited functionality implementation only MUST save the public key, although you would probably still _want_ to save the credential handle with it.

Trying to define what information people SHOULD save would likely require defining use cases that go along with that recommendation. We may also wish to extend such recommendations to extensions, and suggest any third party extension also give such guidance.

Firstyear:
I think that having the object with things you should store sounds like the best solution here. This way the RP can store "everything" they *might* need so that if in the future they change internal policy and require extra data, they can do so retroactively without breaking older tokens etc.

[...]

PR #1718, opened 2022-04-07
Refer to options for the user verification check

infinisil:
Fixes https://github.com/w3c/webauthn/issues/1699, see also the discussion in https://github.com/w3c/webauthn/pull/1704

Ping @emlun @sbweeden

***
<a href="https://pr-preview.s3.amazonaws.com/tweag/w3c-webauthn/pull/1718.html" title="Last updated on Apr 7, 2022, 1:44 PM UTC (5d42923)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/1718/331188b...tweag:5d42923.html" title="Last updated on Apr 7, 2022, 1:44 PM UTC (5d42923)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -4577,8 +4577,9 @@ In order to perform a [=registration ceremony=], the [=[RP]=] MUST proceed as fo
 
 1. Verify that the [=User Present=] bit of the <code>[=flags=]</code> in |authData| is set.
 
-1. If the [=[RP]=] requires [=user verification=] for this registration,
-    verify that the [=User Verified=] bit of the <code>[=flags=]</code> in |authData| is set.
+1. If <code>|options|.{{PublicKeyCredentialCreationOptions/authenticatorSelection}}.{{AuthenticatorSelectionCriteria/userVerification}}</code>
+    is set to {{UserVerificationRequirement/required}}, verify that the [=User Verified=]
+    bit of the <code>[=flags=]</code> in |authData| is set.
 
 1. Verify that the "alg" parameter in the [=credentialPublicKey|credential public key=] in |authData|
     matches the {{PublicKeyCredentialParameters/alg}} attribute of one of the [=list/items=] in
@@ -4744,8 +4745,9 @@ a numbered step. If outdented, it (today) is rendered as a bullet in the midst o
 
 1. Verify that the [=User Present=] bit of the <code>[=flags=]</code> in |authData| is set.
 
-1. If the [=[RP]=] requires [=user verification=] for this assertion,
-    verify that the [=User Verified=] bit of the <code>[=flags=]</code> in |authData| is set.
+1. If <code>|options|.{{PublicKeyCredentialRequestOptions/userVerification}}</code>
+    is set to {{UserVerificationRequirement/required}}, verify that the [=User Verified=]
+    bit of the <code>[=flags=]</code> in |authData| is set.
 
 1. Verify that the values of the [=client extension outputs=] in |clientExtensionResults| and the [=authenticator extension
     outputs=] in the <code>[=authdataextensions|extensions=]</code> in |authData| are as expected, considering the [=client


infinisil:
@agl Ah I see now, so a potential attack here is a non-conformant client, pretending that the authenticator doesn't support user verification, when it really does, in case of `userVerification` set to `"preferred"`. A conformant client would always ask for user verification in such a case, as specified in the client's [Create](https://www.w3.org/TR/webauthn-2/#sctn-createCredential) and [Get](https://www.w3.org/TR/webauthn-2/#sctn-getAssertion) specifications.

In this case I believe the description of [`"preferred"`](https://www.w3.org/TR/webauthn-2/#dom-userverificationrequirement-preferred) should be adjusted, because it currently reads:

> `preferred`: This value indicates that the [Relying Party](https://www.w3.org/TR/webauthn-2/#relying-party) prefers [user verification](https://www.w3.org/TR/webauthn-2/#user-verification) for the operation if possible, but will not fail the operation if the response does not have the [UV](https://www.w3.org/TR/webauthn-2/#uv) [flag](https://www.w3.org/TR/webauthn-2/#flags) set.

I believe the `will not fail` should be changed to `may not fail`. A mention of client policy would also be good here.

Additionally I believe that the [Registering a New Credential](https://www.w3.org/TR/2021/REC-webauthn-2-20210408/#sctn-registering-a-new-credential) and [Verifying an Authentication Assertion](https://www.w3.org/TR/2021/REC-webauthn-2-20210408/#sctn-verifying-assertion) relying party operations should be expanded with a step to determine whether user verification is _effectively_ required (similar to the `Create`/`Get` operations). This would look something like this:

> 14. Let `userVerification` be the effective user verification requirement for credential creation as follows:
>
>     If `options.userVerification`
>     - is set to `"required"`: Let `userVerification` be `true`
>     - is set to `"preferred"`: Let `userVerification` be determined by client policy, taking into account further information gathered from the response [be more specific, what information can the relying party use here?]
>     - is set to `"discouraged"`: Let `userVerification` be `false`
> 15. If `userVerification` is `true`, verify that the [User Verified](https://www.w3.org/TR/2021/REC-webauthn-2-20210408/#concept-user-verified) bit of the [flags](https://www.w3.org/TR/2021/REC-webauthn-2-20210408/#flags) in authData is set.

This also makes it clear that it's relying party policy, but only in the `"preferred"` case.

If that sounds okay then I can propose a PR.

Firstyear:
> This also makes it clear that it's relying party policy, but only in the `"preferred"` case.

There is no text in the RP section about what do do in the "I asked for preferred but got back UV=false" case. Only "UV=true is verify when required was went".

So this section also needs to be amended to include that if the RP asks for preferred, they can make a choice based on RP data to verify the UV bit.

emlun:
@agl's comment above still reflects the WG's position on this - the proposed text is too specific and doesn't capture that an RP may have many reasons to require UV or not. We think the current spec text is appropriately abstract here.

Firstyear:
I think you mis-understand the issue, and the fact that there are literally CVE's that exist because of this defect. It has affected Azure AD, Okta, nextcloud and more. UV preferred is misleading because RP's do not validate it when requested, and RP's *believe* that it is required.

There is a blindingly obvious gap in this specification for "what happens when I ask for UV preferred" during registration and authentication.

emlun:
That is a different issue which has been discussed and closed in #1618.

Firstyear:
Sure, but you still never resolved the issue either in #1618.

infinisil:
I'm a bit disappointed by W3C regarding this.
- I saw something in the specification that I didn't understand how to implement in a library, so I opened [an issue](https://github.com/w3c/webauthn/issues/1699) asking for clarification.
- A small [PR](https://github.com/w3c/webauthn/pull/1704) with a supposed fix was opened, however I [clearly stated](https://github.com/w3c/webauthn/pull/1704#issuecomment-1050043574) that it still didn't make any sense to me, but the PR was merged without addressing my points and the issue was deemed resolved.
- I was told that I can propose another PR, which I did here. While the changes themselves aren't that important, I finally [understood](https://github.com/w3c/webauthn/pull/1718#issuecomment-1128023604) the real problem with the specification and made some clear suggestions for how it could be improved.
- However again, it seems like the WG doesn't deem my concerns here valid and closes the PR, considering the issue resolved. At this point the specification is still as unclear as it was from the beginning, and I'm glad for @Firstyear backing me up here.

So what should I do here? Leave it be and hope that everybody in the future figures out what the specification means themselves? Or should I try to open another PR, hoping that my concerns are taken seriously then?

emlun:
I'm sorry we let you down @infinisil. Opening issues for things I didn't understand was how I, too, got involved with the WG, and I want that opportunity to be open to you too.

As we've said, we think the language proposed here is too specific, but your and @Firstyear's input have given me some other ideas for how we could improve this. I've outlined some of these thoughts in https://github.com/w3c/webauthn/issues/1510#issuecomment-1151576857. In principle I wouldn't mind some help drafting it, but it will likely also tie into #1556 and #1740 - these all relate to and/or would benefit from a more formalized representation of a "credential record" on the server side - so it's going to be a bit tricky to get it all to fit together and to work within Bikeshed. But you (and @Firstyear) are welcome to give it a try if you want, though I understand if you feel disheartened. I've pushed some half-baked thoughs on a [branch in my fork](https://github.com/emlun/webauthn/tree/wip/credential-record) where you can at least see the general direction I'm going with some of it. But note that it's all very experimental, perhaps none of it will make it into the spec in the end.

Firstyear:
> though I understand if you feel disheartened.

I certainly feel this, since it has been a bit more than a year (especially around this issue), and I feel that the concerns I have an as RP aren't really taken seriously - from both a public RP viewpoint, and an enterprise one I can see lots of rough patches that RP's will run into. But I'm glad to see some of these things start to change, because I do want to contribute, and I do care about these things.

PR #1732, opened 2022-05-18
Re-add the PRF extension.

agl:
This reverts commit 9e634a8a4e3b30a768cda642452c862cc836a65d.

***
<a href="https://pr-preview.s3.amazonaws.com/agl/webauthn/pull/1732.html" title="Last updated on Jun 2, 2022, 8:56 AM UTC (baa8169)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/1732/0678b0d...agl:baa8169.html" title="Last updated on Jun 2, 2022, 8:56 AM UTC (baa8169)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -5941,6 +5941,97 @@ At this time, one [=credential property=] is defined: the [=resident key credent
 : Authenticator extension output
 :: None.
 
+## Pseudo-random function extension (<dfn>prf</dfn>) ## {#prf-extension}
+
+This [=client extension|client=] [=registration extension=] and [=authentication extension=] allows a [=[RP]=] to evaluate outputs from a pseudo-random function (PRF) associated with a [=credential=]. The PRFs provided by this extension map from {{USVString}}s of any length to 32-byte {{ArrayBuffer}}s.
+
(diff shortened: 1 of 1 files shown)


Firstyear:
IMO we shouldn't include random extensions like this to the standard. Currently between the 5 models of test authenticator I have from various vendors, no webauthn extension is supported by them.

I think these extensions should be left to specific device manufacturers to define and publish, so that at the least, RP's know *which* manufacturers might support them, rather than misleading that "every" webauthn device "might" support this. 

Alternately, there should be a declaration of which vendor's devices support that extension so that it can be tested and checked by RP's when they go to implement the standard.

nuno0529:
> IMO we shouldn't include random extensions like this to the standard. Currently between the 5 models of test authenticator I have from various vendors, no webauthn extension is supported by them.

Just update the relative information for the point that the extension `hmac-secret` in CTAP2 is the one of [MUST to implement features as Microsoft-compatible security key](https://docs.microsoft.com/nb-NO/windows/security/identity-protection/hello-for-business/microsoft-compatible-security-key). So most authenticators from below vendors should already support it. https://docs.microsoft.com/en-us/azure/active-directory/authentication/concept-authentication-passwordless#fido2-security-key-providers
Besides, this optional feature is also marked as mandatory feature for fido2.1 security key, https://fidoalliance.org/specs/fido-v2.1-ps-20210615/fido-client-to-authenticator-protocol-v2.1-ps-20210615.html#mandatory-features

Firstyear:
> > IMO we shouldn't include random extensions like this to the standard. Currently between the 5 models of test authenticator I have from various vendors, no webauthn extension is supported by them.
> 
> Just update the relative information for the point that the extension `hmac-secret` in CTAP2 is the one of [MUST to implement features as Microsoft-compatible security key](https://docs.microsoft.com/nb-NO/windows/security/identity-protection/hello-for-business/microsoft-compatible-security-key). So most authenticators from below vendors should already support it. https://docs.microsoft.com/en-us/azure/active-directory/authentication/concept-authentication-passwordless#fido2-security-key-providers Besides, this optional feature is also marked as mandatory feature for fido2.1 security key, https://fidoalliance.org/specs/fido-v2.1-ps-20210615/fido-client-to-authenticator-protocol-v2.1-ps-20210615.html#mandatory-features

That's for microsoft to decide what they want to require as an extension, and how they communicate that to users about their authenticator compatibility. We shouldn't throw things into webauthn just because one corporate platform wants it.

agl:
I doesn't look like I can @dveditz in this repo so I'll email him directly about this.

lgarron:
Nit: I think this title is meant to be "Re-add"/"Readd" rather than "Read"?

ve7jtb (approved):
LGTM

timcappalli (approved):
LGTM

emlun (approved):

Issue #1730, opened 2022-05-10
Drop generic client extension processing?


emlun:
This discussion came up in PR #1703. Currently, [§9. WebAuthn Extensions](https://www.w3.org/TR/2021/REC-webauthn-2-20210408/#sctn-extensions) states:

>[...]
>Clients wishing to support the widest possible range of extensions MAY choose to pass through any extensions that they do not recognize to authenticators, generating the [authenticator extension input](https://www.w3.org/TR/2021/REC-webauthn-2-20210408/#authenticator-extension-input) by simply encoding the [client extension input](https://www.w3.org/TR/2021/REC-webauthn-2-20210408/#client-extension-input) in CBOR. [...]

and as such for the spec to be internally consistent, it would be expected that all other parts of the spec should be compatible with this provision.

#1703 proposes adding utilities for decoding [`PublicKeyCredentialCreationOptions`](https://www.w3.org/TR/2021/REC-webauthn-2-20210408/#dictdef-publickeycredentialcreationoptions) and [`PublicKeyCredentialRequestOptions`](https://www.w3.org/TR/2021/REC-webauthn-2-20210408/#dictdef-publickeycredentialrequestoptions) objects from JSON, so RPs don't need to reach deep into the data structures to convert binary values from textual representations before passing them into `navigator.credentials.create()` and `.get()`.

Unfortunately, this seems to be [incompatible with the generic extension pass-through provision](https://github.com/w3c/webauthn/pull/1703#discussion_r845941977), since for unknown extensions there's no way for the client to know which extension input values need to be converted to `BufferSource` values, and JSON has no standardized way to flag binary values. In practice all clients represented in the working group have chosen to not implement generic extension pass-through, so we're opening for discussion:

**Should we keep the generic extension pass-through, or drop it in L3?**

MasterKale:
The py_webauthn project recently received a PR to add in [support for the `credProtect` extension](https://github.com/duo-labs/py_webauthn/pull/128). I initially declined to land it, but after @emlun's comments in #1703 that _technically_ there's nothing preventing us from passing through unrecognized extensions, I reconsidered and currently plan on allowing it at some point.

[...]

PR #1737, opened 2022-06-01
Remove Generic Passthrough

nicksteele:
This is to address #1730. As discussed in the working group meeting, platforms do not currently support generic extension passthroughs there are no plans by platforms to support it, nor interest in retaining it. This will remove the generic passthrough processing provisioning from L3.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/pull/1737.html" title="Last updated on Jun 1, 2022, 6:57 PM UTC (f1f8b5a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/1737/0678b0d...f1f8b5a.html" title="Last updated on Jun 1, 2022, 6:57 PM UTC (f1f8b5a)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -5409,50 +5409,14 @@ ignored by the client browser or OS and not passed to the authenticator at all, 
 Ignoring an extension is never considered a failure in WebAuthn API processing, so when [=[RPS]=] include extensions with any
 API calls, they MUST be prepared to handle cases where some or all of those extensions are ignored.
 
-Clients wishing to support the widest possible range of extensions MAY choose to pass through any extensions that they do not
-recognize to authenticators, generating the [=authenticator extension input=] by simply encoding the [=client extension input=]
-in CBOR. All [=WebAuthn Extensions=] MUST be defined in such a way that this implementation choice does not endanger the user's
+All [=WebAuthn Extensions=] MUST be defined in such a way that this implementation choice does not endanger the user's
 security or privacy. For instance, if an extension requires client processing, it could be defined in a manner that ensures such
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -5409,51 +5409,15 @@ ignored by the client browser or OS and not passed to the authenticator at all,
 Ignoring an extension is never considered a failure in WebAuthn API processing, so when [=[RPS]=] include extensions with any
 API calls, they MUST be prepared to handle cases where some or all of those extensions are ignored.
 
-Clients wishing to support the widest possible range of extensions MAY choose to pass through any extensions that they do not
-recognize to authenticators, generating the [=authenticator extension input=] by simply encoding the [=client extension input=]
-in CBOR. All [=WebAuthn Extensions=] MUST be defined in such a way that this implementation choice does not endanger the user's
+All [=WebAuthn Extensions=] MUST be defined in such a way that this implementation choice does not endanger the user's
 security or privacy. For instance, if an extension requires client processing, it could be defined in a manner that ensures such
```
emlun, line 5413: >Ignoring an extension is never considered a failure in WebAuthn API processing, so when [=[RPS]=] include extensions with any API calls, they MUST be prepared to handle cases where some or all of those extensions are ignored.
>
>All [=WebAuthn Extensions=] MUST be defined in such a way that this implementation choice does not endanger the user's security or privacy.

The start of the second paragraph ends up a non-sequitur without the text that was removed. I'll look over it and open a follow-up PR to fix it.

emlun: Fixed in PR #1768.

Issue #1739, opened 2022-06-09
Discussing mechanisms for enterprise RP's to enforce bound properties of credentials


MasterKale:
This week at WWDC Apple demonstrated the ability of one iCloud user with a registered passkey to share that passkey with another individual on a separate iCloud account. The implications of this in the enterprise space are dire as it means that RP's like Duo can no longer ensure that a platform authenticator response matching a stored public key corresponds 1:1 to the user we originally associated that credential with.

The current state of the spec lacks much configurability to support enterprise RP use cases. Shareable credentials threaten to make it untenable for RP's to continue to leverage WebAuthn as platform vendors evolve authenticator capabilities more quickly than the API evolves.

We should start collecting additional RP concerns and discussing how WebAuthn might evolve to address enterprise RP concerns now that many of the consumer use cases are being addressed with Apple's, Google's, and Microsoft's upcoming passkey support.

MasterKale:
To help clear things up, I'm referring to Apple's demonstration of shareable passkeys in this WWDC session: https://developer.apple.com/videos/play/wwdc2022/10092/

sbweeden:
I believe this is closely related to #1688

emlun:
See also discussion in #1714

sbweeden:
Perhaps wild suggestion. If an RP requests direct attestation, why wouldn't the platforms offer a device-bound credential in this case?

MasterKale:
Particular asks for supporting use of WebAuthn by enterprise RP's:

- Ability to request a device-bound credential and always get back direct attestation (@sbweeden)
- ~~Ability to request a credential be permanently associated to the registering user (@master_kale)~~ (opens the door to too many "bad" outcomes)
- <conversation goes way off the rails>

sbweeden:
I'll try articulate in my own words the requirement I raised in the F2F:

As an enterprise RP, I would like a way to signal to the browser during a registration ceremony, that a device-bound key is required. This offers the client browser an opportunity to tailor the UX around this requirement. I'm explicitly avoiding saying "how" this is signalled (although it has to be some property(s) of the registration request), but ultimately my RP policy is that I wish to know that a credential is bound to a device, and then subsequently be able to ensure this remains true for subsequent authentications.

emlun:
> Perhaps wild suggestion. If an RP requests direct attestation, why wouldn't the platforms offer a device-bound credential in this case?

[...]

Issue #1714, opened 2022-03-30
Provide an explicit way to opt out of multi-device syncing/backups


lxgr:
Sorry in advance if I missed the most recent state of the discussion on multi-device credentials, but if I understand the current proposals correctly,

- This property will be indicated as part of authenticator data (#1692, #1695),
- There will be an opportunity to additionally create per-device "linked/bound" keys (#1658), but
- There _won't_ be an explicit way for an RP to indicate that it wants to opt out of backups/multi-device syncing.

There might be a roundabout way to accomplish this (e.g. through always requesting a device-bound key per #1658), but am I understanding it correctly that there will be no "easy" way to do so, other than effectively only relying on device-bound keys and ignoring/discarding the "actual" key?

Is this intentional? At least for some scenarios, account takeover/phishing might be a large enough concern that RPs might decide to not accept certain (probably mostly host-based) authenticator models' attestation keys anymore for their service, even though they might otherwise be satisfied with the authenticator's security policies and implementation.

serianox:
Hi @lxgr,

Yes, your understanding of multi-device credentials is correct. RP that _do_ care about veting device syncing can currently opt-out with device attestation.

> There won't be an explicit way for an RP to indicate that it wants to opt out of backups/multi-device syncing.

Since this is not UX friendly, because the error happen after te registration, I assume that we will soon add another preference selector for registration to specifically request devices without syncing, or an option to generate a key that will never be synced.

lxgr:
This might unfortunately disqualify the use of platform authenticators in areas where two-factor authentication is explicitly mandated by a regulator, e.g. the EEA in the domain of secure cardholder authentication. A risk decision such as this:

> The RP needs to make a decision if their account recovery is stronger than sending a password reset email to the person's apple or google account

is explicitly prohibited by law in some scenarios there (and password reset emails are not an allowable alternative).

> The RP can cookie the browser

Do we really want to encourage the use of cookies as a form of device binding? If anything, RPs should be using DPKs for that use case, right?

On a related note, as a WebAuthN user, I explicitly don't want some of my credentials to be synced to my keychain under any circumstances (in particular the ones relating to financial transaction authorizations). Maybe there should be a parallel discussion around whether users should be provided with a way to opt out of key backups/syncing on a per-key basis? This is a different concern than the one of RPs opting out, though.

timcappalli:
@ve7jtb wrote:
[...]

Issue #1720, opened 2022-04-11
Update top level use cases to account for multi-device credentials


timcappalli:
The current top level use cases (`sctn-use-cases`) were written prior to the multi-device credential effort and should be rewritten to include this new topic.

Be sure to include: https://github.com/w3c/webauthn/issues/1735

Firstyear:
Will these use cases also discuss when RP's want to *exclude* multi-device credentials?

timcappalli:
> Will these use cases also discuss when RP's want to *exclude* multi-device credentials? 

Proposed text and/or PRs are welcome!

Firstyear:
Sure, I was in the middle of writing these up for a blog anyway, so I'll submit some text and and overview of this from the view of an RP/Enterprise and public site that would want this.

Firstyear:
So, some draft use cases. Especially as someone who implements an RP, and works with a lot of people deploying theirs, I think this boils down to a set of five possible scenarioes and workflows. 

Security Token (Public)
^^^^^^^^^^^^^^^^^^^^^^^

In this use case, we want our authenticator to be a single factor to compliment an existing password.
Instead of an authenticator, a TOTP scheme could alternately be used where either the TOTP or authenticator
plus the password is sufficient to grant access.

Generally in this use case, most identity providers do not care about attestation of the authenticator,
what is more important is that some kind of non-password authentication exists and is present.

Step wise this would appear as:

Registration:

1. The user indicates they wish to enroll a security token
2. The identity provider issues a challenge
3. The browser lists which authenticators attached to the device *could* be registered
4. The user interacts with the authenticator (*note* a pin should not be requested, but fingerprint is okay since it's "transparent")
5. The authenticator releases the signed public key
6. The authenticator is added to the users account

Authentication:

1. The user enters their username
2. The user provides their password and it is validated (*note* we could do this after webauthn)
3. The user indicates they wish to use a security token
2. The identity provider issues a webauthn challenge, limited by the list of authenticators and transports we know are valid for the authenticators associated.
5. The browser offers the list of authenticators that can proceed
6. The user interacts with the authenticator (*note* a pin should not be requested, but fingerprint is okay since it's "transparent")
7. The authenticator releases the signature

Security Token (Corporate)
^^^^^^^^^^^^^^^^^^^^^^^^^^

This is the same as the public use case, except that in many corporations we may want to define a list
of trusted providers of tokens. It's important to us here that these tokens have a vetted or audited
[...]

PR #1744, opened 2022-06-09
DRAFT: Backup eligibility parameter during registration

emlun:
Brainstorm idea for #1739. This is meant to facilitate discussion, not a fully-formed proposal.

Something like this would enable the client to optimize the user interaction to increase the chance that the registration completes successfully. I note in https://github.com/w3c/webauthn/issues/1714#issuecomment-1084473966 that since we now have the `BS` and `BE` flags in the authenticator data, that to me signals that this is a significant credential property that there should perhaps be a feature toggle for.

However the argument against (again, see https://github.com/w3c/webauthn/issues/1714#issuecomment-1084473966) is that we don't want RPs to see this as a "make it more secure" parameter and just set it to `"forbidden"` without further consideration. So in order to respect the interests of the user, this proposal allows the client to let the user override the RP's preference if desired.

Perhaps something like this could be a reasonable middle-ground? Is the risk of ecosystem fragmentation still too great? Is it not powerful enough to be useful to RPs? Discuss!

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/pull/1744.html" title="Last updated on Jun 9, 2022, 9:58 PM UTC (f6db880)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/1744/9622388...f6db880.html" title="Last updated on Jun 9, 2022, 9:58 PM UTC (f6db880)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2769,6 +2769,7 @@ attributes.
         DOMString                    residentKey;
         boolean                      requireResidentKey = false;
         DOMString                    userVerification = "preferred";
+        DOMString                    backupEligible = "any";
     };
 </xmp>
 
@@ -2798,6 +2799,12 @@ attributes.
         The value SHOULD be a member of {{UserVerificationRequirement}} but [=client platforms=] MUST ignore unknown values, treating an unknown value as if the [=map/exist|member does not exist=].
 
         See {{UserVerificationRequirement}} for the description of {{AuthenticatorSelectionCriteria/userVerification}}'s values and semantics.
+
+    :   <dfn>backupEligible</dfn>
+    ::  Specifies the extent to which the [=[RP]=] desires to create a [=backup eligible=] [=credential=].
+        The value SHOULD be a member of {{BackupEligibilityRequirement}} but [=client platforms=] MUST ignore unknown values, treating an unknown value as if the [=map/exist|member does not exist=].
+
+        See {{BackupEligibilityRequirement}} for the description of {{AuthenticatorSelectionCriteria/backupEligible}}'s values and semantics.
 </div>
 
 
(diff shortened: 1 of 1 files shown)


sbweeden:
This type of signal definitely addresses the semantics of what I was asking for in #1739. Attested DPK might also, so I'm not ruling that out just yet. In fact a request with the flag in this proposal could be satisfied by returning the device-bound component in an attested DPK extension response.

index.bs
```diff
@@ -3414,6 +3421,63 @@ Note: The {{UserVerificationRequirement}} enumeration is deliberately not refere
 </div>
 
 
+### Backup Eligibility Requirement Enumeration (enum <dfn enum>BackupEligibilityRequirement</dfn>) ### {#enum-backupEligibilityRequirement}
+
+<xmp class="idl">
+    enum BackupEligibilityRequirement {
+        "forbidden",
+        "discouraged",
```
Firstyear, line 3429: I think discouraged and preferred are a bit useless here, since they are wishy-washy. Why would I "discourage" it and how is that different to "any"? I think forbidden, any and required are the only three options we need. 'any' being "let the user decide".

Firstyear:
This needs strong use cases related to understand what RP's want here with relation to not just backup elligibility, but also attestation. See https://github.com/w3c/webauthn/issues/1720#issuecomment-1102126655

index.bs
emlun: Yep, that's certainly one of the things to discuss here. But before we dive into details on the design, I'd like to hear whether browsers would be interested in supporting something like this at all.

Firstyear: I don't see why browser should dictate a feature that is required for RP's to actually work and enforce policy ....

Firstyear: That doesn't mean we can't change this to have a better user experience. For a user to "register" a device and be sent "errors" about their device being ineligible is not a positive experience. Instead we should create a scenario where we guide the UX to only show authenticators that are "very likely" to succeed.

PR #1755, opened 2022-06-24
Add "hybrid" as a transport.

agl:
FIDO has chosen the name Hybrid for the transport formerly known as
caBLE. This change adds it as a known transport in WebAuthn. Since the
`AuthenticatorTransport` enum is illustrative and not actually
referenced anywhere in the IDL, this is technically a non-normative
change. Also, RPs only store and regurgitate these strings so no change
in RP processing is needed. Still, it's useful to have these strings
documented.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/pull/1755.html" title="Last updated on Jul 14, 2022, 10:52 PM UTC (aaf8ebc)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/1755/9622388...aaf8ebc.html" title="Last updated on Jul 14, 2022, 10:52 PM UTC (aaf8ebc)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -3358,7 +3358,7 @@ Note: The {{AuthenticatorTransport}} enumeration is deliberately not referenced,
     ::  Indicates the respective [=authenticator=] can be contacted over Bluetooth Smart (Bluetooth Low Energy / BLE).
 
     :   <dfn>hybrid</dfn>
-    ::  Indicates the respective [=authenticator=] can be contacted over the Hybrid transport that combines (often separate) data-transport and proximity mechanisms. This is most commonly used to communicate with phones.
+    ::  Indicates the respective [=authenticator=] can be contacted using a combination of (often separate) data-transport and proximity mechanisms. This supports, for example, authentication on a desktop computer using a smartphone.
 
     :   <dfn>internal</dfn>
     ::  Indicates the respective [=authenticator=] is contacted using a [=client device=]-specific transport,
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -3356,6 +3357,9 @@ Note: The {{AuthenticatorTransport}} enumeration is deliberately not referenced,
     :   <dfn>ble</dfn>
     ::  Indicates the respective [=authenticator=] can be contacted over Bluetooth Smart (Bluetooth Low Energy / BLE).
 
+    :   <dfn>hybrid</dfn>
+    ::  Indicates the respective [=authenticator=] can be contacted over the Hybrid transport that combines (often separate) data-transport and proximity mechanisms. This is most commonly used to communicate with phones.
```
MasterKale, line 3361: > ...can be contacted over the Hybrid transport that combines (often separate) data-transport and proximity mechanisms. This is most commonly used to communicate with phones.

How about something like this instead?

> ...can be contacted using a combination of (often separate) data-transport and proximity mechanisms. This supports, for example, authentication on a desktop using a previously registered smartphone.

I think it's important to leave examples here open-ended because we've seen Apple demonstrate caBLE QR code display on an iPhone, ostensibly for cross-device auth (I haven't tested it myself but that was the impression the WWDC session left on me)

MasterKale (approved):

emlun (approved):

akshayku (approved):

PR #1771, opened 2022-07-11
Namespace authData dfns under authData/ and flags under authData/flags/

emlun:
This is a step towards resolving #1740 and #1510, but I think it's substantive enough to warrant review on its own. This is a prerequisite of PR #1772, PR #1773 and PR #1774.

This repurposes the definitions of the **UP** and **UV** terms from referring to the abstract procedure to the corresponding authenticator data flags instead, because we were already using the terms only to refer to the flags. Also added are formal definitions for the other authData flags. These definitions along with those for other authData components are also moved into the `authData/` namespace for better consistency with properties of other interfaces and dictionaries.

This will break some link anchors:

| Anchor before | Anchor after
| ------------- | ------------
| [`#rpidhash`](https://w3c.github.io/webauthn/#rpidhash) | `#authdata-rpidhash`
| [`#flags`](https://w3c.github.io/webauthn/#flags) | `#authdata-flags`
| [`#signcount`](https://w3c.github.io/webauthn/#signcount) | `#authdata-signcount`
| [`#attestedcredentialdata`](https://w3c.github.io/webauthn/#attestedcredentialdata) | `#authdata-attestedcredentialdata`
| [`#authdataextensions`](https://w3c.github.io/webauthn/#authdataextensions) | `#authdata-extensions`
| [`#up`](https://w3c.github.io/webauthn/#up) | [`#authdata-flags-up`](https://pr-preview.s3.amazonaws.com/w3c/webauthn/pull/1771.html#authdata-flags-up)
| [`#uv`](https://w3c.github.io/webauthn/#uv) | [`#authdata-flags-uv`](https://pr-preview.s3.amazonaws.com/w3c/webauthn/pull/1771.html#authdata-flags-uv)
| [`#aaguid`](https://w3c.github.io/webauthn/#aaguid) | `#authdata-attestedcredentialdata-aaguid`
| [`#credentialidlength`](https://w3c.github.io/webauthn/#credentialidlength) | `#authdata-attestedcredentialdata-credentialidlength`
| [`#credentialid`](https://w3c.github.io/webauthn/#credentialid) | `#authdata-attestedcredentialdata-credentialid`
| [`#credentialpublickey`](https://w3c.github.io/webauthn/#credentialpublickey) | `#authdata-attestedcredentialdata-credentialpublickey`

***
[...]

PR #1773, opened 2022-07-11
Extract Credential Record abstraction

emlun:
This is a step towards resolving #1510 and #1556, but is substantive enough to warrant review on its own. This depends on PR #1771 and is a prerequisite of PR #1774.

I open this as a "draft" PR because it has PR #1771 as its base branch, to isolate the diff to the changes unique to this PR. I will re-target this PR to the main branch before merging.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/pull/1773.html" title="Last updated on Sep 22, 2022, 2:18 PM UTC (f6479ef)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/1773/573b1c2...f6479ef.html" title="Last updated on Sep 22, 2022, 2:18 PM UTC (f6479ef)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1157,7 +1157,8 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
             from any [=ceremony=] using the [=public key credential source=].
     </dl>
 
-    [=[RPS]=] MAY include any additional [=struct/items=] as needed,
+    [=WebAuthn extensions=] MAY define additional [=struct/items=] needed to process the extension.
+    [=[RPS]=] MAY also include any additional [=struct/items=] as needed,
     and MAY omit any [=struct/items=] not needed for their implementation.
 
     The <dfn>credential descriptor for a credential record</dfn> is a {{PublicKeyCredentialDescriptor}} value with the contents:
--- a/index.bs
+++ b/index.bs
@@ -1157,6 +1157,21 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
             from any [=ceremony=] using the [=public key credential source=].
     </dl>
 
+    The following [=struct/items=] are OPTIONAL:
+
+    <dl dfn-for="credential record">
+        :   <dfn>attestationObject</dfn>
+        ::  The value of the {{AuthenticatorAttestationResponse/attestationObject}} attribute
+            when the [=public key credential source=] was [=registration|registered=].
+            Storing this enables the [=[RP]=] to reference the credential's [=attestation statement=] at a later time.
+
+        :   <dfn>attestationClientDataJSON</dfn>
+        ::  The value of the {{AuthenticatorResponse/clientDataJSON}} attribute
+            when the [=public key credential source=] was [=registration|registered=].
+            Storing this in combination with the above [=credential record/attestationObject=] [=struct/item=]
+            enables the [=[RP]=] to re-verify the [=attestation signature=] at a later time.
+    </dl>
+
     [=WebAuthn extensions=] MAY define additional [=struct/items=] needed to process the extension.
     [=[RPS]=] MAY also include any additional [=struct/items=] as needed,
     and MAY omit any [=struct/items=] not needed for their implementation.
@@ -5160,6 +5175,16 @@ In order to perform a [=registration ceremony=], the [=[RP]=] MUST proceed as fo
 
         :   [=credential record/BS=]
         ::  The value of the [=authData/flags/BS=] [=flag=] in |authData|.
+    </dl>
+
+    The new [=credential record=] MAY also include the following OPTIONAL contents:
+
+    <dl>
+        :   [=credential record/attestationObject=]
+        ::  <code>|response|.{{AuthenticatorAttestationResponse/attestationObject}}</code>.
+
+        :   [=credential record/attestationClientDataJSON=]
+        ::  <code>|response|.{{AuthenticatorResponse/clientDataJSON}}</code>.
     </dl>
 
 1. If the attestation statement |attStmt| successfully verified but is not trustworthy per step 20 above, the [=[RP]=] SHOULD fail
--- a/index.bs
+++ b/index.bs
@@ -5138,6 +5138,9 @@ In order to perform a [=registration ceremony=], the [=[RP]=] MUST proceed as fo
     with the following contents:
 
     <dl>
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -1118,6 +1118,60 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
...
+
+        :   <dfn>BE</dfn>
+        ::  The value of the [=authData/flags/BE=] [=flag=] when the [=public key credential source=] was created.
+
+        :   <dfn>BS</dfn>
+        ::  The latest value of the [=authData/flags/BS=] [=flag=] in the [=authenticator data=]
+            from any [=ceremony=] using the [=public key credential source=].
+    </dl>
+
+    [=[RPS]=] MAY include any additional [=struct/items=] as needed,
```
timcappalli, line 1160: Should we recommend that if the device public key is requested and returned, it should be stored with or with a reference to the credential record?

emlun: Hm, good point. I'm not sure it's appropriate to call out specific extensions here, but we should at least point out that extensions might add items that need to be stored. Perhaps as an extension point similar to how extensions add fields to `partial dictionary AuthenticationExtensionsClientOutputs` etc?

index.bs
```diff
@@ -1118,6 +1118,60 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
...
+        :   <dfn>signCount</dfn>
+        ::  The latest value of the [=authData/signCount|signature counter=] in the [=authenticator data=]
+            from any [=ceremony=] using the [=public key credential source=].
+
+        :   <dfn>transports</dfn>
+        ::  The value returned from {{AuthenticatorAttestationResponse/getTransports()}}
+            when the [=public key credential source=] was [=registration|registered=].
+
+            Note: Modifying or removing [=list/items=] from the value returned from {{AuthenticatorAttestationResponse/getTransports()}}
+            could negatively impact user experience, or even prevent use of the corresponding credential.
```
Firstyear, line 1150: Reality seems to be the opposite, seems to be that chrome doesn't properly assign all the transports for a credential during registration with getTransports, so removing (clearing) this value is actually safe 😅

index.bs
Firstyear: we probably also need to store the attestation ca chain or related attestation metadata to assist with DPK here. We already store this in webauthn-rs, here: https://github.com/kanidm/webauthn-rs/blob/master/webauthn-rs-core/src/interface.rs#L245

index.bs
agl: Chrome had a bug where it was overconfident that it knew a complete set of transports. Chrome 105 will report an empty set unless it has transports from the getInfo or attestation certificate.

index.bs
agl: (Irrespective of DPK, some enterprise RPs certainly like to store the attestation for record keeping.)

index.bs
```diff
@@ -5078,23 +5132,31 @@ In order to perform a [=registration ceremony=], the [=[RP]=] MUST proceed as fo
...
 
-1. If the attestation statement |attStmt| verified successfully and is found to be trustworthy, then register the new
-    credential with the [=user account=] that was denoted in <code>|options|.{{PublicKeyCredentialCreationOptions/user}}</code>:
+1. If the attestation statement |attStmt| verified successfully and is found to be trustworthy,
+    then create and store a new [=credential record=] in the [=user account=]
+    that was denoted in <code>|options|.{{PublicKeyCredentialCreationOptions/user}}</code>,
+    with the following contents:
+
+    <dl>
+        :   [=credential record/id=]
```
agl, line 5141: (The definition, above, also includes a `type` member.)

agl (approved):
LGTM with nits.

index.bs
emlun: Setting `transports` to empty in `PublicKeyCredentialDescriptor` is equivalent to omitting it, so yes, that should always work in the sense of not preventing use of a credential. This could make the user experience slightly worse if the user only has `internal`-only credentials, though, as they might get prompted to connect an external authenticator they don't have.

However if you remove items from `transports` but it remains non-empty, then that could prevent using a credential if the client assumes the credential requires an unavailable transport (say, if `getTransports()` returns `["usb", "nfc"]` and you remove `"usb"`, then pass `transports: ["nfc"]` on a device with no NFC antenna).

akshayku (approved):

selfissued (approved):
LGTM

timcappalli (approved):
lgtm

index.bs
emlun: Should be fixed by a5710f6, please re-review!

emlun:
On the 2022-09-21 WG call we said to merge this after I'd fixed @agl's [review comment](https://github.com/w3c/webauthn/pull/1773#discussion_r950739605), but I'd forgotten there was another [unresolved comment thread](https://github.com/w3c/webauthn/pull/1773#discussion_r918219321). That led to a slightly more substantial change, so I'll give @timcappalli an opportunity to re-review before we merge this.

index.bs
```diff
@@ -1118,6 +1118,76 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
...
+        ::  The value of the {{AuthenticatorAttestationResponse/attestationObject}} attribute
+            when the [=public key credential source=] was [=registration|registered=].
+            Storing this enables the [=[RP]=] to reference the credential's [=attestation statement=] at a later time.
+
+        :   <dfn>attestationClientDataJSON</dfn>
+        ::  The value of the {{AuthenticatorResponse/clientDataJSON}} attribute
+            when the [=public key credential source=] was [=registration|registered=].
+            Storing this in combination with the above [=credential record/attestationObject=] [=struct/item=]
+            enables the [=[RP]=] to re-verify the [=attestation signature=] at a later time.
+    </dl>
```
timcappalli, line 1173: @emlun now that attestation is possible on getAssertion, do we need additional language here to address that?

emlun: Yes, but I'd like to tie that into #1663 instead since that's where that change will be introduced. I intend to open a meta-PR to update #1663 to use the new credential record abstraction once this is merged (unless #1663 gets merged before I can get to it :smile:).

timcappalli:
changes lgtm

emlun:
Thanks @timcappalli (and everyone else)!

Issue #1656, opened 2021-07-27
Cross-origin credential creation in iframes


agl:
In level two we supported cross-origin assertions (when [allowed](https://www.w3.org/TR/2021/REC-webauthn-2-20210408/#sctn-iframe-guidance) by the top-level) but omitted cross-origin creation because there wasn't anyone with a use-case.

We would like to revisit that and allow cross-origin creation along the same lines as assertion. We believe this will be useful in a payments context.

When making a payment there are three (or four) parties involved. The customer seeks to authorise the payment. They are on the merchant's site. A bank needs to approve the payment, and there might be a payment processor between the merchant and the bank. If the bank can make a cross-origin assertion on the merchant page then that can greatly improve their confidence that the transaction is genuine. However, that assumes that the bank _has_ a credential for the user, and that would be aided by being able to enroll users inline. Thus the desire for cross-origin creation.

arshadnoor:
Hi Adam (@agl ),

I concur with the use-case: that it is essential for Banks to receive digital signatures from Consumers directly when they're confirming a payment transaction at a Merchant site. But I humbly disagree with the premise that the optimal way to register Consumers is with cross-origin credential creation. I think Banks and browser companies are making a mistake and forcing the FIDO protocol/implementations to become a lot more complex than it need become - at least for this use-case.

Every Bank, IMO, is going to have to contend with FIDO sooner or later whether they like it or not. Windows 11 and Apple Passkey will ensure that. If they have to implement FIDO for their own users to interact with them directly, why would they not simply require that the Consumer register with them _first_ and associate all their registered Public Keys for whatever payment instruments they have acquired from the Bank?

By doing so, here are all the benefits that accrue to the ecosystem:

- Consumers only have to register with the Bank once (for each unique Authenticator they choose to register);
- Consumers do not have to be distracted with additional friction during the Checkout process (to register an Authenticator);
- Consumers do not have to be worried that they are being attacked by someone while shopping at a site ("What are these additional things the Merchant wants me to do?", "How do I know I can trust what I'm seeing here?", "Perhaps the Merchant has been compromised?");
- Merchants do NOT have to create complex Checkout workflows for registration of Authenticators. When the Consumer chooses a specific payment instrument during Checkout, and the Merchant sends that information to the Bank, the Bank recognizes the instrument, finds the associated Public Key(s) and a unique challenge comes back as a response for the Consumer to sign;
[...]

PR #1801, opened 2022-09-14
Allow for credential creation in a cross-origin iframe

stephenmcgruer:
See https://github.com/w3c/webauthn/issues/1656

This PR adds the ability to call navigator.credentials.create() in a cross-origin
iframe (more exactly, a frame that is not same-origin with its ancestor chain),
as long as the following conditions are met:

- The iframe has a publickey-credentials-create-feature permission policy
- The iframe has a [user activation](https://html.spec.whatwg.org/multipage/interaction.html#tracking-user-activation)

It has a corresponding PR on the credential management spec; see
https://github.com/w3c/webappsec-credential-management/pull/209

***
<a href="https://pr-preview.s3.amazonaws.com/stephenmcgruer/webauthn/pull/1801.html" title="Last updated on Feb 27, 2023, 12:52 PM UTC (cc256f3)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/1801/d3270c5...stephenmcgruer:cc256f3.html" title="Last updated on Feb 27, 2023, 12:52 PM UTC (cc256f3)">Diff</a>


stephenmcgruer:
cc @ve7jtb , who I believe wished to make sure that this was compatible with the previous L2 discussion on what create() in a cross-origin iframe would look like (particularly the permission policy name, I believe).

stephenmcgruer:
Hi folks; I had hoped to bring this PR up on today's call, but since it is cancelled pinging here instead. We are still waiting on a response here from Apple, and I would love to have that for next WebAuthn WG meeting if possible (2022/11/16). If there is no answer by the 16th (or during the call), I will be interested to know how the WebAuthn WG wants to proceed at that time (i.e., what the method is to resolve this PR one way or another).

agl (approved):
I might even skip the TODO. I'm not sure that we'll be in a better place to resolve it in the future and it's clear that implementations have managed to figure out user activation in these contexts already.

stephenmcgruer:
> I might even skip the TODO. I'm not sure that we'll be in a better place to resolve it in the future and it's clear that implementations have managed to figure out user activation in these contexts already.

Ack - done, removed the TODO in https://github.com/w3c/webauthn/pull/1801/commits/eb32a68f2a3088aa573f03ce20858f0bfe43d4db

rlin1:
Lgtm

stephenmcgruer:
Post-vacations ping to @akshayku @timcappalli @ve7jtb - could you review this before this week's WebAuthn sync? (I.e. before the 11th). I'd like to get this merged.

ve7jtb (commented):
It looks good but I think the containing page info should probably be added to client data.

stephenmcgruer:
@timcappalli @ve7jtb - I've gone ahead and pushed two commits to this PR that add a note on the client making it clear to the user if the credential creation is in a cross-origin iframe, and also added a `topOrigin` member to convey some sense of the containing page info (though one could have a chain of different origins of course, and this wouldn't be represented).

Hope this is somewhat along the lines of what you were each thinking - but if not no worries, happy to revise to whatever feedback you have :)

index.bs
```diff
@@ -3590,6 +3606,11 @@ Note: The {{CollectedClientData}} may be extended in the future. Therefore it's
     ::  This member contains the fully qualified [=origin=] of the requester, as provided to the authenticator by the client, in
         the syntax defined by [[!RFC6454]].
 
+    :   <dfn>topOrigin</dfn>
+    ::  This member contains the fully qualified [=top-level origin=] of the requester, as provided to the authenticator by the client, in
```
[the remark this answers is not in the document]
stephenmcgruer: Great point! Must admit I just copied the description from `origin`, and presumed that all of CollectedClientData was passed to the authenticator. That is not true of course 🤣 - only a hash of the JSON is, and `origin` is passed indirectly I believe (via `rpId`, which is checked against `origin` and the algorithm only continues if they match).

I'm happy to reword if you or @ve7jtb have thoughts here; perhaps just drop the authenticator mention and say:

`This member contains the fully qualified [=top-level origin=] of the requester, in the syntax defined by [[!RFC6454]]. If {{CollectedClientData/crossOrigin}} is `false`, this member should be identical to {{CollectedClientData/origin}}.`

?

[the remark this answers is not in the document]
timcappalli: Would a client be expected to populate topOrigin in a traditional, top level call (no iframe)?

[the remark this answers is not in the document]
stephenmcgruer: Great question! As written, the PR does say a client should populate `topOrigin` in a traditional, top level call, and populate it identically to `origin`. However it might be reasonable to say that `topOrigin` is empty (or null?) in those situations. I don't think I have a preference in either direction.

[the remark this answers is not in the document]
timcappalli: I think empty or not present makes sense for non-iframe scenarios, but required for iframe.

@akshayku @ve7jtb thoughts?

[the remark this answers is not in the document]
akshayku: Empty or not present in non-iframe scenario makes sense to me too. And it should be present in iframe scenarios.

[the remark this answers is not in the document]
stephenmcgruer: Thanks. I've updated the PR to be contingent on `sameOriginWithAncestors` being `false`. Note that this means `topOrigin` would not be set in the following scenario:

- relying-party.example
  - relying-party.example <-- a call from this sub-frame would not set `topOrigin`

If we would prefer it be set for *all* iframe scenarios, I can do that too.

index.bs
```diff
@@ -3644,12 +3668,15 @@ The serialization algorithm works by appending successive byte strings to an, in
 1. Append [=CCDToString=]({{CollectedClientData/challenge}}) to |result|.
 1. Append 0x2c226f726967696e223a (`,"origin":`) to |result|.
 1. Append [=CCDToString=]({{CollectedClientData/origin}}) to |result|.
+1. If {{CollectedClientData/topOrigin}} is present:
+    1. Append 0x2c22746f704f726967696e223a (`,"topOrigin":`) to |result|.
+    1. Append [=CCDToString=]({{CollectedClientData/topOrigin}}) to |result|.
```
emlun, line 3673: This has to go after `crossOrigin`, as inserting it before that would be to "change the order in which those fields are serialized" which is prohibited in the "Future works" section below. Right, @agl?

index.bs
```diff
@@ -3690,6 +3717,7 @@ Verifiers may use the following algorithm to verify an encoded {{CollectedClient
     1. A string, |type|, that contains the expected {{CollectedClientData/type}}.
     1. A byte string, |challenge|, that contains the challenge byte string that was given in the {{PublicKeyCredentialRequestOptions}} or {{PublicKeyCredentialCreationOptions}}.
     1. A string, |origin|, that contains the expected {{CollectedClientData/origin}} that issued the request to the user agent.
+    1. An optional string, |topOrigin|, that contains the expected {{CollectedClientData/topOrigin}} of the frame that issued the request to the user agent.
```
emlun, line 3720: Again, this needs to go after `crossOrigin`.

index.bs
```diff
@@ -3699,6 +3727,9 @@ Verifiers may use the following algorithm to verify an encoded {{CollectedClient
 1. Append [=CCDToString=](|challengeBase64|) to |expected|.
 1. Append 0x2c226f726967696e223a (`,"origin":`) to |expected|.
 1. Append [=CCDToString=](|origin|) to |expected|.
+1. If |topOrigin| is specified:
+    1. Append 0x2c22746f704f726967696e223a (`,"topOrigin":`) to |expected|.
+    1. Append [=CCDToString=](|topOrigin|) to |expected|.
```
emlun, line 3732: Again, this needs to go after `crossOrigin`.

index.bs
```diff
@@ -3720,7 +3751,7 @@ Verifiers may use the following algorithm to verify an encoded {{CollectedClient
 
 #### Future development #### {#clientdatajson-development}
 
-In order to remain compatible with the [limited verification algorithm](#clientdatajson-verification), future versions of this specification must not remove any of the fields {{CollectedClientData/type}}, {{CollectedClientData/challenge}}, {{CollectedClientData/origin}}, or {{CollectedClientData/crossOrigin}} from {{CollectedClientData}}. They also must not change the [serialization algorithm](#clientdatajson-verification) to change the order in which those fields are serialized.
+In order to remain compatible with the [limited verification algorithm](#clientdatajson-verification), future versions of this specification must not remove any of the fields {{CollectedClientData/type}}, {{CollectedClientData/challenge}}, {{CollectedClientData/origin}}, {{CollectedClientData/topOrigin}}, or {{CollectedClientData/crossOrigin}} from {{CollectedClientData}}. They also must not change the [serialization algorithm](#clientdatajson-verification) to change the order in which those fields are serialized.
```
emlun, line 3754: Perhaps we should also explicitly mention that inserting a new field counts as changing the order?

```suggestion
In order to remain compatible with the [limited verification algorithm](#clientdatajson-verification), future versions of this specification must not remove any of the fields {{CollectedClientData/type}}, {{CollectedClientData/challenge}}, {{CollectedClientData/origin}}, {{CollectedClientData/crossOrigin}}, or {{CollectedClientData/topOrigin}} from {{CollectedClientData}}. They also must not change the [serialization algorithm](#clientdatajson-verification) to change the order in which those fields are serialized, or insert new fields between them.
```

emlun (changes requested):
I see some issues with the limited verification algorithm, but they should be straightforward to fix.

stephenmcgruer:
Thanks @emlun - you're definitely right and I was uncomfortable with my changes to the limited verification algorithm as-is. I'm going to address the easier errors you pointed out first, and then come back and think harder about the overall structure!

index.bs
emlun: I also thought about whether this algorithm would be simpler if `topOrigin` is always present even when equal to `origin`, but I think that it ultimately doesn't matter since L2 browser implementations would still emit client data without `topOrigin` even if L3 always requires it.

index.bs
stephenmcgruer: Fixed the order; see other comment for ongoing bigger-picture questions.

index.bs
stephenmcgruer: Incorporated this suggested change, thanks!

index.bs
stephenmcgruer: Fixed.

index.bs
stephenmcgruer: Apologies for the delay here - I've been pondering on this on and off over the week. I think there's an underlying question about whether the limited verification algorithm assumes a correct L3 implementation, or if it tries to 'catch' incorrect implementations.

Assuming that the underlying implementation is correct, then the 2x2 combination of inputs seems to be:

(a) crossOrigin=yes, topOrigin=null - verifier expected a crossOrigin case but does not care about the topOrigin
(b) crossOrigin=yes, topOrigin=foo.org - verifier expected a crossOrigin case and expect it to be from foo.org **if** that information is available (aka if the implementation is L3)
(c) crossOrigin=no, topOrigin=foo.org - invalid input (should specify this probably)
(d) crossOrigin=no, topOrigin=null - verifier expected a same-origin case

For that, this is what I have currently:

13. If _expected_ is not a prefix of _clientDataJSON_ then verification has failed
14. Let _remainingClientDataJSON_ be the result of removing the prefix _expected_ from _clientDataJSON_
15. If _topOrigin_ is specified and ",topOrigin:" is a prefix of _remainingClientDataJSON_ then:
    1. Let _expectedTopOrigin_ be an empty byte string
    2. Append ",topOrigin:" to _expectedTopOrigin_
    3. Append _topOrigin_ to _expectedTopOrigin_
    4. if _expectedTopOrigin_ is not a prefix of _remainingClientDataJSON_, fail
    5. let _remainingClientDataJSON_ be the result of removing the prefix _expectedTopOrigin_ from _clientDataJSON_
16. If _remainingClientDataJSON_ is not at least one byte long, fail
17. If the first byte of _remainingClientDataJSON_ is:
    - 0x7d => success
    - 0x2c => success
    - otherwise => fail

(a) is tested by step 13 alone, and then we skip step 15 because the verifier doesn't care
(b) is tested by step 13 and step 15 for L3 implementations, or just by step 13 for L2 implementations
(c) should be disallowed before the algorithm begins
(d) is tested by step 13 alone, and then we skip step 15 because any correct implementation will not have specified topOrigin


Does that make sense? My assumptions and/or algorithm may each be independently incorrect, but I'd like to validate my assumptions about what we want the verifier to be able to assume and to test first!

emlun: That looks fair to me, I think. I'm generally in favour of strictness, so I reckon an explicit check of (c) is probably good. But I'd like for @agl to weigh in on this too.

agl: I would suggest keeping this algorithm simple and not worrying too much about it. There was a particular case that absolutely wouldn't accept parsing JSON and thus this exists, but I don't think the spec should put too much effort into handling every case without a suggestion that it would actually be useful for someone. Perhaps `topOrigin` could just be ignored in this algorithm?

ve7jtb (approved):
LGTM

timcappalli (approved):
changes LGTM

akshayku (approved):

index.bs
stephenmcgruer: As per discussion in today's WebAuthn WG sync, I have removed any modification to the LVA, pending its potential deletion in the future.

stephenmcgruer:
Ping @emlun - your review is still considered outstanding by Github which blocks merging. If you can either re-review or just [clear your requested change](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/dismissing-a-pull-request-review), that would be great thanks!

index.bs
```diff
@@ -1712,6 +1712,9 @@ This [=internal method=] accepts three arguments:
     ::  This argument is a Boolean value which is [TRUE] if and only if the caller's [=environment settings object=] is
         [=same-origin with its ancestors=]. It is [FALSE] if caller is cross-origin.
 
+        Note: Invocation of this [=internal method=] indicates that it was allowed by
+        [=permissions policy=], which is evaluated at the [[!CREDENTIAL-MANAGEMENT-1]] level.
+        See [[#sctn-permissions-policy]].
```
emlun, line 1717: This is intended so it appears related to `sameOriginWithAncestors`, is that intentional?

index.bs
```diff
@@ -3649,7 +3673,10 @@ The serialization algorithm works by appending successive byte strings to an, in
     1. Append 0x66616c7365 (`false`) to |result|.
```
emlun, line 3673: In the introduction of this section (emphasis added):

>This verification algorithm requires only [...] and **three** conditional checks [...]

The "three" needs to either be deleted or updated to four. Something like "simple conditional checks" could work too.

index.bs
```diff
@@ -3649,7 +3673,10 @@ The serialization algorithm works by appending successive byte strings to an, in
...
 1. Otherwise:
     1. Append 0x74727565 (`true`) to |result|.
-1. Create a temporary copy of the {{CollectedClientData}} and remove the fields {{CollectedClientData/type}}, {{CollectedClientData/challenge}}, {{CollectedClientData/origin}}, and {{CollectedClientData/crossOrigin}} (if present).
+1. If {{CollectedClientData/topOrigin}} is present:
+    1. Append 0x2c22746f704f726967696e223a (`,"topOrigin":`) to |result|.
+    1. Append [=CCDToString=]({{CollectedClientData/topOrigin}}) to |result|.
+1. Create a temporary copy of the {{CollectedClientData}} and remove the fields {{CollectedClientData/type}}, {{CollectedClientData/challenge}}, {{CollectedClientData/origin}}, {{CollectedClientData/topOrigin}} (if present), and {{CollectedClientData/crossOrigin}} (if present).
```
emlun, line 3679: For consistency:
```suggestion
1. Create a temporary copy of the {{CollectedClientData}} and remove the fields {{CollectedClientData/type}}, {{CollectedClientData/challenge}}, {{CollectedClientData/origin}}, {{CollectedClientData/crossOrigin}} (if present), and {{CollectedClientData/topOrigin}} (if present).
```

index.bs
```diff
@@ -5220,6 +5248,14 @@ In order to perform a [=registration ceremony=], the [=[RP]=] MUST proceed as fo
...
 1. Verify that the value of <code>|C|.{{CollectedClientData/origin}}</code> matches the [=[RP]=]'s [=origin=].
 
+1. If <code>|C|.{{CollectedClientData/topOrigin}}</code> is present:
+
+    1. Verify that the [=[RP]=] expects that this credential would have been created within an iframe that is
+        not [=same-origin with its ancestors=].
+
+    1. Verify that the value of <code>|C|.{{CollectedClientData/topOrigin}}</code> matches the [=origin=] of a page
+        that the [=[RP]=] expects to be sub-framed within.
+
```
emlun, line 5258: This check should also be added to §7.2. Verifying an Authentication Assertion, no?

emlun (changes requested):
Sorry for the delay! Just a few simple tweaks remaining now as far as I can tell.

index.bs
stephenmcgruer: I was matching `5.1.4.1. PublicKeyCredential’s [[DiscoverFromExternalSource]](origin, options, sameOriginWithAncestors) Method`, though I agree that the indentation is strange. Let me know if you'd like it changed in both places, otherwise I will leave it as-is for now.

index.bs
stephenmcgruer: Fixed, thanks!

index.bs
stephenmcgruer: Done.

index.bs
stephenmcgruer: `topOrigin` is not yet set for get() - @timcappalli will be doing that follow-up work in https://github.com/w3c/webauthn/issues/1842 .

stephenmcgruer (commented):
Thanks for the review!

index.bs
emlun: Ah, I see. Thanks!

index.bs
emlun: Ah, I see. Let's handle those both together later then, if at all.

emlun (approved):
Thanks @stephenmcgruer!

nsatragno:
(Merging as this is all green & blocking https://github.com/w3c/webappsec-credential-management/pull/209 and some other work, feel free to open an issue for outstanding issues)

index.bs
```diff
@@ -5220,6 +5248,14 @@ In order to perform a [=registration ceremony=], the [=[RP]=] MUST proceed as fo
 
 1. Verify that the value of <code>|C|.{{CollectedClientData/origin}}</code> matches the [=[RP]=]'s [=origin=].
 
+1. If <code>|C|.{{CollectedClientData/topOrigin}}</code> is present:
```
ndpar, line 5251: Multiple references to step 19 and 20 below in paragraph 7.1 should be (manually?) updated.

emlun: Thanks! Moved to new issue: #1861

prnk28:
LFG 🚀

PR #1805, opened 2022-09-22
Fix incorrect use of options variable: rename to pkOptions

### LINKED ISSUE #1752: Incorrect use of `options` variable in `create()` and `get()` definitions
[§5.1.3. Create a New Credential](https://w3c.github.io/webauthn/#sctn-createCredential) and [§5.1.4. Use an Existing Credential to Make an Assertion](https://w3c.github.io/webauthn/#sctn-discover-from-external-source) both declare their **options** parameter as the `Credential[Creation|Request]Options` object inherited from CredMan:

>**options**
>This argument is a [`CredentialCreationOptions`](https://w3c.github.io/webappsec-credential-management/#dictdef-credentialcreationoptions) object whose _options_.[`publicKey`](https://w3c.github.io/webauthn/#dom-credentialcreationoptions-publickey) member contains a [`PublicKeyCredentialCreationOptions`](https://w3c.github.io/webauthn/#dictdef-publickeycredentialcreationoptions) object [...]

Both also re-assign the _options_ variable:

>Let _options_ be the value of _options_.[`publicKey`](https://w3c.github.io/webauthn/#dom-credentialrequestoptions-publickey).

But both then also reference _options_.`signal`, which is a member of `Credential[Creation|Request]Options` but not `PublicKeyCredential[Creation|Request]Options`:

>If _options_.[`signal`](https://w3c.github.io/webappsec-credential-management/#dom-credentialrequestoptions-signal) is present and [aborted](https://dom.spec.whatwg.org/#abortsignal-aborted), throw the _options_.[`signal`](https://w3c.github.io/webappsec-credential-management/#dom-credentialrequestoptions-signal)’s [abort reason](https://dom.spec.whatwg.org/#abortsignal-abort-reason).

For correctness, we need to keep a reference to the _options_ inherited from CredMan and either introduce a new variable for its _options_.`pubKey` attribute, or simply reference the WebAuthn-specific parameters using the _options_.`pubKey` prefix.

emlun:
Fixes #1752. This is mutually exclusive with PR #1806. I couldn't decide which approach is better, but I'm slightly favouring this one over #1806.

This fixes the issue by introducing a new variable _pkOptions_ instead of re-assigning the existing variable _options_, so that _options_ can keep its original value.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/pull/1805.html" title="Last updated on Sep 22, 2022, 5:27 PM UTC (3a543c4)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/1805/0bfc0d0...3a543c4.html" title="Last updated on Sep 22, 2022, 5:27 PM UTC (3a543c4)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1648,15 +1648,15 @@ When this method is invoked, the user agent MUST execute the following algorithm
 
     Note: This "sameOriginWithAncestors" restriction aims to address a tracking concern raised in [Issue #1336](https://github.com/w3c/webauthn/issues/1336). This may be revised in future versions of this specification.
 
-1. Let |options| be the value of <code>|options|.{{CredentialCreationOptions/publicKey}}</code>.
+1. Let |pkOptions| be the value of <code>|options|.{{CredentialCreationOptions/publicKey}}</code>.
 
-1. If the {{PublicKeyCredentialCreationOptions/timeout}} member of |options| is present, check if its value lies within a
+1. If <code>|pkOptions|.{{PublicKeyCredentialCreationOptions/timeout}}</code> is present, check if its value lies within a
     reasonable range as defined by the [=client=] and if not, correct it to the closest value lying within that range. Set a timer
-    |lifetimeTimer| to this adjusted value. If the {{PublicKeyCredentialCreationOptions/timeout}} member of |options| is not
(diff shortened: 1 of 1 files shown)


agl (approved):
I like this one better too.

Issue #1711, opened 2022-03-23
RP operations: some extension processing may assume that the encompassing signature is valid


equalsJeffH:
In both of the [RP Operations subsections](https://www.w3.org/TR/webauthn/#sctn-rp-operations) (Registering a new cred, and verifying an authn assertion), the step for verifying/processing of extension outputs is placed _before_ the step for verifying the signature value over "authenticator data".

This is fine for idempotent extensions that simply marshall data for eventual return to the RP as a part of the operation's response.

However, extensions such as `devicePubKey` / "DPK" (PR #1663) call for the RP to associate and store extension-generated data with the user's account. The RP probably should only do so if the overall credential creation or authentication results themselves validate correctly, which notably includes verifying the "encompassing" signature over "authenticator data".

Perhaps the step for verifying extension outputs should be moved to after the step(s) for verifying the "encompassing" signature over "authenticator data" in both of the registering a new cred, and verifying an authn assertion sections.

emlun:
Related: #1064

PR #1807, opened 2022-09-22
Move state updates to last in RP verification steps

emlun:
This relates to (but does not fix) issue #1711, and will be useful for applying the new "credential record" (see #1773) abstraction to PR #1663. The stored state should be updated only after performing all validation and verification steps.

This also changes the definition type (and therefore autolink syntax) for the credential record struct members from `dfn` to `abstract-op`. This will help avoid conflicts with existing definitions, including `[=scope=]`, as we introduce a struct for devicePubKey records as well.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/pull/1807.html" title="Last updated on Sep 26, 2022, 8:30 AM UTC (05fe54d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/1807/6c823f1...05fe54d.html" title="Last updated on Sep 26, 2022, 8:30 AM UTC (05fe54d)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1127,7 +1127,7 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
     The following [=struct/items=] are RECOMMENDED in order to implement all steps of
     [[#sctn-registering-a-new-credential]] and [[#sctn-verifying-assertion]] as defined:
 
-    <dl dfn-for="credential record">
+    <dl dfn-for="credential record" dfn-type="abstract-op">
         :   <dfn>type</dfn>
         ::  The [=public key credential source/type=] of the [=public key credential source=].
 
@@ -1158,7 +1158,7 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
 
     The following [=struct/items=] are OPTIONAL:
 
-    <dl dfn-for="credential record">
+    <dl dfn-for="credential record" dfn-type="abstract-op">
         :   <dfn>attestationObject</dfn>
         ::  The value of the {{AuthenticatorAttestationResponse/attestationObject}} attribute
             when the [=public key credential source=] was [=registration|registered=].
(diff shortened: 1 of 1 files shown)


emlun:
(You can also see a sneak peek of the draft for devicePubKey records here: https://github.com/w3c/webauthn/compare/jeffh-fix-1658-device-bound-key-extension...emlun:webauthn:dpk-credential-record#diff-5e793325cd2bfc452e268a4aa2f02b4024dd9584bd1db3c2595f61f1ecf7b985R7079)

Firstyear (approved):
Looks really good, makes sense to me.

agl (approved):

index.bs
```diff
@@ -5157,36 +5157,36 @@ In order to perform a [=registration ceremony=], the [=[RP]=] MUST proceed as fo
...
-        :   [=credential record/signCount=]
+        :   [$credential record/signCount$]
         ::  <code>|authData|.[=authData/signCount=]</code>.
 
-        :   [=credential record/transports=]
+        :   [$credential record/transports$]
         ::  The value returned from <code>|response|.{{AuthenticatorAttestationResponse/getTransports()}}</code>.
 
-        :   [=credential record/BE=]
+        :   [$credential record/BE$]
```
MasterKale, line 5176: Any reason not to make this a bit more readable and call the column `backupEligible` instead? It'll save people a click who don't know to equate "BE" to the flag's full name.

index.bs
```diff
@@ -5157,36 +5157,36 @@ In order to perform a [=registration ceremony=], the [=[RP]=] MUST proceed as fo
...
         ::  The value of the [=authData/flags/BE=] [=flag=] in |authData|.
 
-        :   [=credential record/BS=]
+        :   [$credential record/BS$]
```
MasterKale, line 5179: Any reason not to make this a bit more readable and call the column `backupStatus` instead? It'll save people a click who don't know to equate "BS" to the flag's full name.

index.bs
emlun: Good point, thanks! But unrelated to this change, so let's tackle that in issue #1811.

index.bs
emlun: Good point, thanks! But unrelated to this change, so let's tackle that in issue #1811.

timcappalli (approved):
LGTM

emlun:
@MasterKale I'll consider your current comments resolved, but you're welcome to add any more comments either here (even after merging) or in #1811.

PR #1810, opened 2022-09-30
Conditional request allowList credential filtering

### LINKED ISSUE #1793: Support Filtering by allow list in Conditional UI
## Description

It is common for sites to re-confirm an already authenticated user's password to perform sensitive operations in an application - such as changing a username, two-factor auth setting or any other security related data point. Replacing this flow with user verifying authenticator credentials is required to fully eliminate password use.

As I worked on implementing Conditional UI, I realized that while the confirming credentials can only be for the currently authenticated user, Conditional UI provides no way to filter those credentials. Choosing the credentials of any user except the currently authenticated one will always fail. 

Screenshot 2022-09-05 at 8 31 21 AM

While the Conditional UI explainer [explicitly requires an empty allowCredentials list](https://github.com/w3c/webauthn/wiki/Explainer:-WebAuthn-Conditional-UI#empty-allowcredentials), it seems like this use case was not considered. In traditional password based flows, a hidden field with the username is utilized to hint to password managers which credential is being requested.

Conditional UI needs a method to filter or at least hint which user's credentials are acceptable for this use case.

## Related Links

Without a Conditional UI hint, implementers will be forced to rely on some sort of browser state to prevent a negative user interaction which will incur all of the original problems leading to the development of the Conditional UI: https://github.com/w3c/webauthn/issues/1356

nsatragno:
Allow filtering credentials during conditional requests by passing an `allowList` to `navigator.credentials.get`. This allows relying parties who know who the user attempting to authenticate is (e.g. because they entered their username) to have autofill only show credentials for that user. See the bug for a description of use-cases.

Fixes #1793


index.bs
```diff
@@ -2190,8 +2190,11 @@ When this method is invoked, the user agent MUST execute the following algorithm
 1. If <code>|options|.{{CredentialRequestOptions/mediation}}</code> is present with the value
     {{CredentialMediationRequirement/conditional}}:
 
-    1. If <code>|options|.{{PublicKeyCredentialRequestOptions/allowCredentials}}</code> is not [=list/empty=],
-        return a {{DOMException}} whose name is "{{NotSupportedError}}", and terminate this algorithm.
+    1. Let |credentialIdFilter| be <code>|options|.{{CredentialRequestOptions/publicKey}}.{{PublicKeyCredentialRequestOptions/allowCredentials}}</code>.
+
+    1. Set <code>|options|.{{CredentialRequestOptions/publicKey}}.{{PublicKeyCredentialRequestOptions/allowCredentials}}</code> to [=list/empty=].
+
+        Note: This prevents [=non-discoverable credentials=] from being used during {{CredentialMediationRequirement/conditional}} requests.
```
emlun, line 2197: I recall that we said at TPAC and in https://github.com/w3c/webauthn/issues/1793#issuecomment-1244413340 that conditional UI wouldn't support non-discoverable credentials, but I don't recall the motivation. Why add that restriction?

nsatragno: Essentially the same reasoning as in the [explainer](https://github.com/w3c/webauthn/wiki/Explainer:-WebAuthn-Conditional-UI#empty-allowcredentials): non discoverable credentials don't usually store `PublicKeyCredentialUserEntity` information, so we wouldn't be able to display them on the autofill surface. Silent discovery would not be available for non discoverable credentials on platform authenticators.

(I need to go and update the explainer with this new filtering)

agl (approved):

index.bs
emlun: Ah, I see. And I suppose that restricting to only discoverable keys prevents any confusion around why some particular authenticator's non-discoverable keys work with conditional UI (because it happens to store those values anyway) while another's doesn't. Fair enough, then. Thanks!

index.bs
```diff
@@ -2364,11 +2369,13 @@ When this method is invoked, the user agent MUST execute the following algorithm
 
                 1. [=list/For each=] |credentialMetadata| of |collectedDiscoveredCredentialMetadata|:
 
-                    1. [=map/Set=] |silentlyDiscoveredCredentials|[|credentialMetadata|] to |authenticator|.
+                    1. If |credentialIdFilter| [=list/is empty=] or |credentialIdFilter| contains an item whose
```
emlun, line 2372: I guess?
```suggestion
                    1. If |credentialIdFilter| [=list/is empty=] or |credentialIdFilter| [=list/contains=] an [=list/item=] whose
```

index.bs
```diff
@@ -2364,11 +2369,13 @@ When this method is invoked, the user agent MUST execute the following algorithm
...
+                        {{PublicKeyCredentialDescriptor/id}}'s value is set to |credentialMetadata|'s [=DiscoverableCredentialMetadata/id=],
```
emlun, line 2373: I think this could read slightly better this way:
```suggestion
                        {{PublicKeyCredentialDescriptor/id}}'s value equals <code>|credentialMetadata|.[=DiscoverableCredentialMetadata/id=]</code>,
```

index.bs
```diff
@@ -2428,6 +2435,9 @@ When this method is invoked, the user agent MUST execute the following algorithm
                     [=client extension processing=] algorithm to create the [=client extension outputs=], for each
                     [=client extension=] in <code>|options|.{{PublicKeyCredentialRequestOptions/extensions}}</code>.
 
+            1. If |credentialIdFilter| [=list/is not empty=] and |credentialIdFilter| does not contain an item whose
```
emlun, line 2438: I guess?

```suggestion
            1. If |credentialIdFilter| [=list/is not empty=] and |credentialIdFilter| does not [=list/contain=] an [=list/item=] whose
```

index.bs
```diff
@@ -2428,6 +2435,9 @@ When this method is invoked, the user agent MUST execute the following algorithm
...
+                {{PublicKeyCredentialDescriptor/id}}'s value is set to the value of [=credentialIdResult=], [=continue=].
```
emlun, line 2439: This should be prefixed with _assertionCreationData_, and I think this could read slightly better this way:

```suggestion
                {{PublicKeyCredentialDescriptor/id}}'s value equals <code>|assertionCreationData|.[=credentialIdResult=]</code>, [=continue=].
```

emlun (approved):
LGTM, with a few minor nits.

ve7jtb (approved):
lgtm

pascoej (approved):

nsatragno:
From the call: Nina to merge after fixing the merge errors.

PR #1836, opened 2023-01-04
Only expose the UV PRF

### LINKED ISSUE #1851: PRF inputs should be BufferSource instead of ArrayBuffer
Throughout the WebAuthn API we use `BufferSource` for binary input parameters (e.g., [`PublicKeyCredentialCreationOptions.challenge`](https://w3c.github.io/webauthn/#dom-publickeycredentialcreationoptions-challenge), [`PublicKeyCredentialUserEntity.id`](https://w3c.github.io/webauthn/#dom-publickeycredentialuserentity-id)) and `ArrayBuffer` for binary return values (e.g., [`PublicKeyCredential.rawId`](https://w3c.github.io/webauthn/#dom-publickeycredential-rawid), [`AuthenticatorAttestationResponse.attestationObject`](https://w3c.github.io/webauthn/#dom-authenticatorattestationresponse-attestationobject)). However the `prf` extension uses `ArrayBuffer` for both input parameters and output return values. This means that this code example:

```javascript
var credid = null;
    publicKey: {
        challenge: new Uint8Array([1, 2, 3, 4]),
        pubKeyCredParams: [{type: 'public-key', alg: -7}],
        rp: { name: 'Test' },
        user: { id: new Uint8Array([5, 6, 7, 8]), name: 'test', displayName: 'Test' },
        extensions: {
            prf: {
                eval: {
                    first: new Uint8Array([1, 2, 3, 4]),
                },
            },
        },
    },
}).then(cred => {
    console.log(cred);
    console.log(cred.getClientExtensionResults());
    credid = cred.rawId;
})
```
generates the following error in Chrome Canary (112.0.5580.0):
```
VM1538:2 Uncaught (in promise) TypeError: Failed to execute 'create' on 'CredentialsContainer': Failed to read the 'publicKey' property from 'CredentialCreationOptions': Failed to read the 'extensions' property from 'PublicKeyCredentialCreationOptions': Failed to read the 'prf' property from 'AuthenticationExtensionsClientInputs': Failed to read the 'eval' property from 'AuthenticationExtensionsPRFInputs': Failed to read the 'first' property from 'AuthenticationExtensionsPRFValues': Failed to convert value to 'ArrayBuffer'.
    at <anonymous>:2:23
```

This can be worked around using `new Uint8Array(...).buffer`, but is not in line with how the rest of the API works.

## Proposed Change

- Split [`AuthenticationExtensionsPRFValues`](https://w3c.github.io/webauthn/#dictdef-authenticationextensionsprfvalues) into two versions: one for input and one for output.
- Change `ArrayBuffer` to `BufferSource` in the one used in client extension inputs.

agl:
Authenticators may have different PRFs for the UV and non-UV case. Thus setting uv=preferred during an assertion is fraught: it doesn't fully specify which PRF to use.

However, while implementing this, I ended up feeling that the prohibition on using uv=preferred was too strong. Sites may reasonably want to use uv=preferred and to take advantage of available PRF results. If the evaluation points are global then this isn't so silly as to justify a prohibition, I suspect.

Fixes #1851.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/pull/1836.html" title="Last updated on Mar 16, 2023, 4:44 PM UTC (6478874)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/1836/33e7507...6478874.html" title="Last updated on Mar 16, 2023, 4:44 PM UTC (6478874)">Diff</a>


Firstyear:
I think the only concern is that it relies on all browsers to implement uv=preferred consistently across device classes. Chrome and Safari are certainly here where uv=preferred is more often than not uv=required, but firefox is really absent here. 

I think the stronger wording was fine because I think RP's should know there are risks of changing between the UV types, and so consistency is important.

Perhaps if you want to make this wording softer, could it be worth saying that when an RP gets the PRF outputs back they also need to check uv=true/false to determine if the correct PRF was used?

index.bs
```diff
@@ -6547,7 +6547,7 @@ At this time, one [=credential property=] is defined: the [=resident key credent
...
...ser presence|user presence test=] is performed. It is the responsibility of the [=[RP]=] to set the {{PublicKeyCredentialRequestOptions/userVerification}} parameter accordingly: [=[RPS]=] SHOULD *consistently* use either the {{UserVerificationRequirement/required}} or {{UserVerificationRequirement/discouraged}} values of {{UserVerificationRequirement}} when using this extension, otherwise the outputs may vary for a given input. For example, if this extension is used with {{PublicKeyCredentialRequestOptions/userVerification}} set to {{UserVerificationRequirement/preferred}} and an authenticator that doesn't have [=user verification=] configured, then later configuring user verification on that authenticator may cause the PRF outputs to change. Because of this, when creating a credential with this extension where {{PublicKeyCredentialRequestOptions/userVerification}} set to {{UserVerificationRequirement/preferred}}, the [=client platform=] SHOULD configure the authenticator for [=user verification=] if it is capable of it as that avoids the PRF changing in the future. At assertion time, [=[RPS]=] SHOULD check the UV bit in the [=flags=] to confirm that the expected PRF was evaluated.
```
emlun, line 6550: Two suggestions:

- Fix a typo: "where `userverification` _is_ set to `preferred`"
- Rewrite "UV bit in the flags" as "UV flag" and autolink both "UV" and "flag"

```suggestion
When requested, if a [=credential=] is created on an [=authenticator=] that supports this extension then either one or two freshly-seeded PRFs are associated with it. If two PRFs are associated then, when outputs are requested, one PRF is evaluated if [=user verification=] is performed and the other is evaluated if only a [=test of user presence|user presence test=] is performed. It is the responsibility of the [=[RP]=] to set the {{PublicKeyCredentialRequestOptions/userVerification}} parameter accordingly: [=[RPS]=] SHOULD *consistently* use either the {{UserVerificationRequirement/required}} or {{UserVerificationRequirement/discouraged}} values of {{UserVerificationRequirement}} when using this extension, otherwise the outputs may vary for a given input. For example, if this extension is used with {{PublicKeyCredentialRequestOptions/userVerification}} is set to {{UserVerificationRequirement/preferred}} and an authenticator that doesn't have [=user verification=] configured, then later configuring user verification on that authenticator may cause the PRF outputs to change. Because of this, when creating a credential with this extension where {{PublicKeyCredentialRequestOptions/userVerification}} set to {{UserVerificationRequirement/preferred}}, the [=client platform=] SHOULD configure the authenticator for [=user verification=] if it is capable of it as that avoids the PRF changing in the future. At assertion time, [=[RPS]=] SHOULD check the [=authData/flags/UV=] [=flag=] to confirm that the expected PRF was evaluated.
```

index.bs
```diff
@@ -6600,8 +6600,8 @@ Note: If PRF results are obtained during [=registration=] then the [=[RP]=] MUST
 
 : Client extension processing ([=authentication extension|authentication=])
 ::
-      1. If {{PublicKeyCredentialRequestOptions/userVerification}} has the value {{UserVerificationRequirement/preferred}}, return a {{DOMException}} whose name is “{{NotSupportedError}}”.
       1. If {{AuthenticationExtensionsPRFInputs/evalByCredential}} is not empty but {{PublicKeyCredentialRequestOptions/allowCredentials}} is empty, return a {{DOMException}} whose name is “{{NotSupportedError}}”.
+      1. If any [=map/key=] in {{AuthenticationExtensionsPRFInputs/evalByCredential}} is the empty string, or is not a valid [=base64url encoding=], return a {{DOMException}} whose name is “{{NotSupportedError}}”.
```
emlun, line 6604: Is `NotSupportedError` the most suitable error name here? I'm thinking something more like [`SyntaxError`](https://webidl.spec.whatwg.org/#syntaxerror) or [`DataError`](https://webidl.spec.whatwg.org/#dataerror) might be a better fit (at least for the second line, probably both lines?). The Web IDL spec even includes an example that sounds a lot like what we're doing here:

>"[SyntaxError](https://webidl.spec.whatwg.org/#syntaxerror)" [DOMException](https://webidl.spec.whatwg.org/#idl-DOMException) is used to report parsing errors in web APIs, for example when parsing selectors

...although I suppose that's probably referring to CSS selectors, so maybe not quite as close a match.

emlun (changes requested):

index.bs
agl: (I editing this suggestion because I think the typo fix was done in the wrong place. Please search for "is set to" in the result and check that it makes sense.)

index.bs
agl: Thanks. I think you're right that SyntaxError makes more sense here. Done.

index.bs
emlun: Sorry, yes - good catch!

MasterKale:
> I think the stronger wording was fine because I think RP's should know there are risks of changing between the UV types, and so consistency is important.

After dabbling with `prf` last week I think I have to second this notion. Most of the hype around this extension are people wanting to implement client-side E2EE. It feels like a foot gun to leave the door open for RP's to build out functionality that sometimes fails (catastrophically, in the case of encryption) because a browser decides that `"preferred"` actually means "evaluate the non-UV prf because UV isn't configured", missing the nuance being introduced here that the UV prf should be evaluated "...if it is capable of it..."

`"preferred"` is already fraught with inconsistent behavior, as recent conversations in the WG have highlighted. Is it in our best interest to continue to encourage ambiguous implementations of WebAuthn around this value, when practically speaking UV should almost always be required or discouraged depending on the RP's authentication model?

akshayku:
Looks like there are corner cases with `preferred` which will make deploying PRF difficult for a relying party. 

Some corner cases:
- For security keys supporting user verification but PIN has not been setup yet:
  - Same Browser: 
    - Browser not setting up PIN which resulted in `preferred` evaluating to "without UV" at one point. Then user setup user verification out of band on the authenticator some other place. And next time, same evaluation resulting in "With UV" 
  - Different Browser
    - Different browser having different behavior about whether it is setting PIN or not
- Platform Authenticator
  - User moves from one device to another where `preferred` meaning has changed due to device properties in case of platform authenticator in case of syncing credential.

index.bs
```diff
@@ -6545,14 +6545,14 @@ At this time, one [=credential property=] is defined: the [=resident key credent
...
 
 This extension is implemented on top of the [[FIDO-CTAP]] `hmac-secret` extension. It is a separate [=client extension=] because `hmac-secret` requires that inputs and outputs be encrypted in a manner that only the user agent can perform, and to provide separation between uses by WebAuthn and any uses by the underlying platform. This separation is achieved by hashing the provided PRF inputs with a context string to prevent evaluation of the PRFs for arbitrary inputs.
 
+The `hmac-secret` extension provides two PRFs per credential: one which is used for requests where [=user verification=] is performed and another for all other requests. This extension only exposes a single PRF per credential and, when implementing on top of `hmac-secret`, that PRF MUST be the one used for when [=user verification=] is performed. This overrides the {{UserVerificationRequirement}} if neccessary.
```
Firstyear, line 6554: I think this is a good idea to only use the UV version. Rather than the extension overriding the uv param, should the extension be ignored if UV wasn't returned / set?

ve7jtb (approved):
lgtm

MasterKale (approved):

index.bs
agl: Discussed on the call of 2023-02-22:

When the uv and prf requests conflict, one of them has to win. My feeling is that prf should win because:

1. The RP may not be able to function without PRF results, but an extra UV is just an inconvenience for the user.
2. Platforms already upgrade UV automatically in a number of cases, such as discoverable-credential requests.

So I (and the group on the call) are still leaning towards giving priority to the prf request.

index.bs
```diff
@@ -6545,14 +6545,14 @@ At this time, one [=credential property=] is defined: the [=resident key credent
...
+This [=client extension|client=] [=registration extension=] and [=authentication extension=] allows a [=[RP]=] to evaluate outputs from a pseudo-random function (PRF) associated with a [=credential=]. The PRFs provided by this extension map from {{BufferSource}}s of any length to 32-byte {{ArrayBuffer}}s.
```
emlun, line 6548:
```suggestion
This [=client extension|client=] [=registration extension=] and [=authentication extension=] allows a [=[RP]=] to evaluate outputs from a pseudo-random function (PRF) associated with a [=credential=]. The PRFs provided by this extension map from {{BufferSource}}s of any length to 32-byte {{BufferSource}}s.
```

index.bs
```diff
@@ -6600,8 +6600,8 @@ Note: If PRF results are obtained during [=registration=] then the [=[RP]=] MUST
 
 : Client extension processing ([=authentication extension|authentication=])
 ::
-      1. If {{PublicKeyCredentialRequestOptions/userVerification}} has the value {{UserVerificationRequirement/preferred}}, return a {{DOMException}} whose name is “{{NotSupportedError}}”.
       1. If {{AuthenticationExtensionsPRFInputs/evalByCredential}} is not empty but {{PublicKeyCredentialRequestOptions/allowCredentials}} is empty, return a {{DOMException}} whose name is “{{NotSupportedError}}”.
+      1. If any [=map/key=] in {{AuthenticationExtensionsPRFInputs/evalByCredential}} is the empty string, or is not a valid [=base64url encoding=], or does not match an element of {{PublicKeyCredentialRequestOptions/allowCredentials}} after performing [=base64url encoding|base64url decoding=], then return a {{DOMException}} whose name is “{{SyntaxError}}”.
```
emlun, line 6604: To be a bit more formally precise (feel free to ignore this if it seems redundant):

```suggestion
      1. If any [=map/key=] in {{AuthenticationExtensionsPRFInputs/evalByCredential}} is the empty string, or is not a valid [=base64url encoding=], or does not equal the {{PublicKeyCredentialDescriptor/id}} of some element of {{PublicKeyCredentialRequestOptions/allowCredentials}} after performing [=base64url encoding|base64url decoding=], then return a {{DOMException}} whose name is “{{SyntaxError}}”.
```

emlun (approved):
LGTM with a few minor nits.

emlun (approved):

akshayku (approved):

pascoej (approved):

PR #1843, opened 2023-01-25
Define discrete list of pubKeyCredParams algs

### LINKED ISSUE #1757: Which "pubKeyCredParams" to use?
Hi,

I noticed that during `credentials.create(...)`, if the list does not contain what the authenticator can provide, the authenticator will not be included in the list of authenticators to choose from. For example, if you don't include `"alg":-257`, Windows Hello won't work.

Now, as a relying party this all sounds a bit like unknown mysteries. 

- the specification says "pick your algorithms" from a [huge list](https://www.iana.org/assignments/cose/cose.xhtml#algorithms)!
- no idea which algos the authenticators support
- no idea which algos you really have to support as an RP

In practice, using this list restricts your choice to a subset of authenticators available... if you manage to find out which algo is needed. Also, most RPs are not deeply knowledgeable about which crypto algorithms is better suited or not.

So ...are all common authenticators covered by RS256 and ES256? Or should you as an RP add some more to cover most authenticators? Which ones?

MasterKale:
This PR adds a suggestion to RP's to define three algorithm ID's in `pubKeyCredParams` for the widest possible authenticator support. [Quoting myself from comments I made in #1757](https://github.com/w3c/webauthn/issues/1757#issuecomment-1321203084):

> I'd put `-8` at the top of the list to encourage adoption of Ed25519, since authenticators that can support it will choose to use it if it's first on the list (in my experience this is basically YubiKey 5's, but I think SoloKeys support it too). After that `-7` and `-257` are the two that practically cover almost 100% of existing authenticators.

Fixes #1757.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/pull/1843.html" title="Last updated on Jan 26, 2023, 5:26 PM UTC (4670540)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/1843/d3270c5...4670540.html" title="Last updated on Jan 26, 2023, 5:26 PM UTC (4670540)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -3036,6 +3036,12 @@ optionally evidence of [=user consent=] to a specific transaction.
         The [=client=] and [=authenticator=] make a best-effort to create a credential of the most preferred type possible.
         If none of the listed types can be created, the {{CredentialsContainer/create()}} operation fails.
 
+        [=[RP]=]'s that wish to support a wide range of [=authenticators=] SHOULD define entries for the following discrete list of signature algorithms:
+
+        * -8 (Ed25519)
+        * -7 (ES256)
+        * -257 (RS256)
+
     :   <dfn>timeout</dfn>
     ::  This OPTIONAL member specifies a time, in milliseconds, that the [=[RP]=] is willing to wait for the call to complete. This is
         treated as a hint, and MAY be overridden by the [=client=].
--- a/index.bs
+++ b/index.bs
@@ -3036,7 +3036,7 @@ optionally evidence of [=user consent=] to a specific transaction.
         The [=client=] and [=authenticator=] make a best-effort to create a credential of the most preferred type possible.
         If none of the listed types can be created, the {{CredentialsContainer/create()}} operation fails.
 
-        [=[RP]=]'s that wish to support a wide range of [=authenticators=] SHOULD define entries for the following discrete list of signature algorithms:
+        [=[RPS]=] that wish to support a wide range of [=authenticators=] SHOULD include at least the following {{COSEAlgorithmIdentifier}} values:
 
         * -8 (Ed25519)
         * -7 (ES256)
--- a/index.bs
+++ b/index.bs
@@ -3042,6 +3042,8 @@ optionally evidence of [=user consent=] to a specific transaction.
         * -7 (ES256)
         * -257 (RS256)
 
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -3036,6 +3036,14 @@ optionally evidence of [=user consent=] to a specific transaction.
         The [=client=] and [=authenticator=] make a best-effort to create a credential of the most preferred type possible.
         If none of the listed types can be created, the {{CredentialsContainer/create()}} operation fails.
 
+        [=[RP]=]'s that wish to support a wide range of [=authenticators=] SHOULD define entries for the following discrete list of signature algorithms:
```
MasterKale, line 3039: I considered adding an "in this order" to this list, but above it mentions "ordered from most preferred to least preferred" so maybe it'd be superfluous?

emlun: Yeah, I agree that sounds redundant.

emlun, line 3039: `[=[RP]=]'s` would render as `Relying Party's`. Use the `[RPS]` macro instead for the plural.

```suggestion
        [=[RPS]=] that wish to support a wide range of [=authenticators=] SHOULD define entries for the following discrete list of signature algorithms:
```

emlun, line 3039: In what way is this list "discrete"? I'm not sure what that word is supposed to mean here, I've only seen it in contexts like "discrete vs. continuous" or "discrete components vs. integrated circuits".

emlun, line 3039: "SHOULD define entries for" reads a bit awkward to me. Perhaps something like this would read better?

`SHOULD include at least the following algorithm identifiers:` (or even `the following {{COSEAlgorithmIdentifier}} values:`)

With that we could probably remove the trailing "Additional signature algorithms can be included as needed" as well.

MasterKale: I tried using "discrete" to communicate "a finite list" instead of "all possible algorithm IDs now and in the future."

MasterKale: Great suggestions, thank you.

MasterKale: So...many...macros...but an easy enough change so I went ahead with it.

emlun: Ok. Either way, I guess we can close this as the term no longer appears in the latest version. :slightly_smiling_face:

emlun (approved):
LGTM, thanks!

PR #1850, opened 2023-02-05
Create SECURITY.md

carlena3497:


emlun:
Thanks, but the suggested file is just a template with no substance. If you would like to propose an improvement to the spec or its accompanying resources, please [open an issue](https://github.com/w3c/webauthn/issues/new/choose).

PR #1855, opened 2023-02-20
Recommend duration of challenge validity

### LINKED ISSUE #1848: How long the relying party should maintain the challenge and related information?
In section [13.4.3. Cryptographic Challenges](https://w3c.github.io/webauthn/#sctn-cryptographic-challenges), there are some information about such requirements.
> As a cryptographic protocol, Web Authentication is dependent upon randomized challenges to avoid replay attacks. Therefore, the values of both [PublicKeyCredentialCreationOptions](https://w3c.github.io/webauthn/#dictdef-publickeycredentialcreationoptions).[challenge](https://w3c.github.io/webauthn/#dom-publickeycredentialcreationoptions-challenge) and [PublicKeyCredentialRequestOptions](https://w3c.github.io/webauthn/#dictdef-publickeycredentialrequestoptions).[challenge](https://w3c.github.io/webauthn/#dom-publickeycredentialrequestoptions-challenge) MUST be randomly generated by [Relying Parties](https://w3c.github.io/webauthn/#relying-party) in an environment they trust (e.g., on the server-side), and the returned [challenge](https://w3c.github.io/webauthn/#dom-collectedclientdata-challenge) value in the client’s response MUST match what was generated. This SHOULD be done in a fashion that does not rely upon a client’s behavior, e.g., the [Relying Party](https://w3c.github.io/webauthn/#relying-party) SHOULD store the challenge temporarily until the operation is complete. Tolerating a mismatch will compromise the security of the protocol.

Following sentence is quite unclear to me.
> This SHOULD be done in a fashion that does not rely upon a client’s behavior, e.g., the [Relying Party](https://w3c.github.io/webauthn/#relying-party) SHOULD store the challenge temporarily until the operation is complete.

It seems that the RP server should have a kind of timeout to invalidate the challenge in order not to rely on the client's behavior.
RP might set some timeout values by considering the expected user action timeout and indicate such expectations in the options. Our current implementation follows such behavior.

One of issues if the RP have similar logics as follows if the conditional UI (autofill) is used.
1. User loads the login page (w/ conditional UI) and fetches the webauthn.get options.
2. User does not perform webauthn login process.  e.g., moves to other tabs and stays there until the server's challenge is timeout
3. User moves to the original tab and performs webauthn login with Auto-Fill UI
4. _User sees an error since the challenge on the RP server side is already timeout(or invalidated)_

**In above case, if the RP server does not have any timeout for the issued challenge, it would create some security risks.**

What's the recommendation of this? Or, is it better to respect preferred timeout value from the option in the browser' auto-fill implementation?

emlun:
Fixes #1848.

Additions I also considered but decided against:

- Mentioning that the RP MAY enforce a shorter timeout if a shorter `options.timeout` is requested. I think the recommendation of a "similar" timeout probably communicates that this is a vague limit with a lot of wiggle room for such things.
- Recommending that challenges SHOULD NOT be valid past the recommended duration. This is softly implied, and doesn't seem necessary as this section comes from the angle of how long the RP _needs_ to store the challenge, so the RP probably prefers a short time already.
- Recommending allowing only a single verification attempt per challenge/ceremony. While probably a good idea, it shouldn't matter much in practice, and there might be reasons to allow retrying in some cases. It's also perhaps not entirely clear what "once per ceremony" means - once per `navigator.credentials.{create,get}()` call or once per session? - so it might be more confusing than helpful.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/pull/1855.html" title="Last updated on Jun 28, 2023, 10:11 AM UTC (c36459f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/1855/6dfbdba...c36459f.html" title="Last updated on Jun 28, 2023, 10:11 AM UTC (c36459f)">Diff</a>


MasterKale (approved):
I thought to suggest we frame "discouraged" vs "preferred"/"required" as "second-factor" vs "passwordless"/"passkeys", but maybe that framing is too "practical" to be included in the spec?

Anyway I think this looks good so I'm going ahead and approving.

emlun:
Yes, I think that's a bit too much indirection. The reader has to do more work to relate those terms to the `UserVerificationRequirement` values, when we can just refer to them directly instead. Especially since this new section is also referenced in the client algorithms, I already think that "the [=[RP]=]'s [=user verification=] preference for the ceremony" is almost too informal as it doesn't explicitly and unambiguously say what defines that preference. But I think it's probably clear enough with the reference to the `UserVerificationRequirement` values.

emlun:
2023-03-22 WG call: conditional mediation makes this a bit complicated. We might [change the API a bit](https://github.com/w3c/webauthn/issues/1856#issuecomment-1443822151) to better support conditional mediation, holding off on this one until that's decided.

index.bs
```diff
@@ -8501,8 +8488,25 @@ as discussed in [[#sctn-username-enumeration]].
...
+
+## Recommended Ranges for Ceremony Timeouts ## {#sctn-timeout-recommended-ranges}
+
 [=Ceremonies=] relying on timing, e.g., a [=registration ceremony=] (see {{PublicKeyCredentialCreationOptions/timeout}}) or an [=authentication ceremony=] (see {{PublicKeyCredentialRequestOptions/timeout}}), ought to follow [[!WCAG21]]'s [Guideline 2.2 Enough Time](https://www.w3.org/TR/WCAG21/#enough-time). If a [=client platform=] determines that a [=[RP]=]-supplied timeout does not appropriately adhere to the latter [[!WCAG21]] guidelines, then the [=client platform=] MAY adjust the timeout accordingly.
 
+The <dfn>recommended ranges and defaults for a WebAuthn ceremony timeout</dfn> are as follows:
+
+If the [=[RP]=]'s [=user verification=] preference for the ceremony is
+
+  <dl class="switch">
```
akshayku, line 8500: See also https://github.com/w3c/webauthn/pull/1885 where I am proposing adjusting these ranges.

emlun: Updated to account for the changes in #1885.

emlun:
@akshayku @sbweeden This is now updated to account for the changes in #1885, and ready for re-review.

Kieun:
I was thinking that this issue is blocked until #1856 is resolved.
Regarding original questions for conditional-mediation, the RP client needs to abort request and only allow modal request if the user click some buttons?

emlun:
Oh right, I forgot. This should probably stay blocked then. This issue is on the agenda for today's WG meeting, we'll decide on that then.

sbweeden (approved):

ve7jtb (approved):
LGTM

akshayku (approved):

PR #1880, opened 2023-04-21
Add authenticatorDisplayName to credProps

emlun:
In today's WG meeting, there was a desire expressed for some way for the RP to receive a user-friendly name for the new credential. For example, in order to instruct the user to "use your passkey saved in Password Manager X". To that end, this would add a new credential property to the `credProps` extension, possibly configurable by the user and with the possibility of a fallback value set by the authenticator vendor.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/pull/1880.html" title="Last updated on May 3, 2023, 7:26 PM UTC (8dd7190)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/1880/54e0962...8dd7190.html" title="Last updated on May 3, 2023, 7:26 PM UTC (8dd7190)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1168,7 +1168,7 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
             Storing this in combination with the above [$credential record/attestationObject$] [=struct/item=]
             enables the [=[RP]=] to re-verify the [=attestation signature=] at a later time.
 
-        :   <dfn>displayName</dfn>
+        :   <dfn>authenticatorDisplayName</dfn>
         ::  A [=human palatability|human-palatable=] description of the [=public key credential source=].
 
             If used, the [=[RP]=] SHOULD use this to describe the [=credential record=] in the user's account settings.
@@ -6632,9 +6632,9 @@ This [=client extension|client=] [=registration extension=] facilitates reportin
             The [=client=] MAY consider a user-configured response chosen by the user,
             and SHOULD allow the user to modify a vendor-provided default response.
 
-            If the [=[RP]=] includes a <code>[$credential record/displayName$]</code> [=struct/item=] in [=credential records=],
+            If the [=[RP]=] includes an <code>[$credential record/authenticatorDisplayName$]</code> [=struct/item=] in [=credential records=],
             the [=[RP]=] MAY offer this value, if present,
-            as a default value for the <code>[$credential record/displayName$]</code> of the new [=credential record=].
+            as a default value for the <code>[$credential record/authenticatorDisplayName$]</code> of the new [=credential record=].
     </div>
 
 
--- a/index.bs
+++ b/index.bs
@@ -1167,6 +1167,17 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
             when the [=public key credential source=] was [=registration|registered=].
             Storing this in combination with the above [$credential record/attestationObject$] [=struct/item=]
             enables the [=[RP]=] to re-verify the [=attestation signature=] at a later time.
+
+        :   <dfn>displayName</dfn>
+        ::  A [=human palatability|human-palatable=] description of the [=public key credential source=].
+
+            If used, the [=[RP]=] SHOULD use this to describe the [=credential record=] in the user's account settings.
+            The [=[RP]=] SHOULD allow the user to choose this value, and MAY allow the user to modify it at will.
+
+            The [=credProps|Credential Properties Extension=] defines two [=credential properties=],
+            {{CredentialPropertiesOutput/userDisplayName}} and {{CredentialPropertiesOutput/vendorDisplayName}},
+            which, when available, MAY be offered as defaults for this value.
+            The [=[RP]=] MAY also derive a default value from the authenticator's [=attestation statement=], if any.
     </dl>
 
     [=WebAuthn extensions=] MAY define additional [=struct/items=] needed to process the extension.
@@ -6564,9 +6575,6 @@ During a transition from the FIDO U2F JavaScript API, a [=[RP]=] may have a popu
 
 This [=client extension|client=] [=registration extension=] facilitates reporting certain [=credential properties=] known by the [=client=] to the requesting [=[WRP]=] upon creation of a [=public key credential source=] as a result of a [=registration ceremony=].
 
-At this time, one [=credential property=] is defined: the [=resident key credential property=]
-(i.e., [=client-side discoverable credential property=]).
-
 : Extension identifier
 :: `credProps`
 
@@ -6590,6 +6598,8 @@ At this time, one [=credential property=] is defined: the [=resident key credent
     <xmp class="idl">
     dictionary CredentialPropertiesOutput {
         boolean rk;
+        USVString userDisplayName;
+        USVString vendorDisplayName;
     };
 
     partial dictionary AuthenticationExtensionsClientOutputs {
@@ -6608,6 +6618,33 @@ At this time, one [=credential property=] is defined: the [=resident key credent
             If {{rk}} is not present, it is not known whether the credential is a [=discoverable credential=] or a [=server-side credential=].
 
             Note: some [=authenticators=] create [=discoverable credentials=] even when not requested by the [=client platform=]. Because of this, [=client platforms=] may be forced to omit the {{rk}} property because they lack the assurance to be able to set it to [FALSE]. [=[RPS]=] should assume that, if the `credProps` extension is supported, then [=client platforms=] will endeavour to populate the {{rk}} property. Therefore a missing {{rk}} indicates that the created credential is most likely a [=non-discoverable credential=].
+
+        :   <dfn>userDisplayName</dfn>
+        ::  This OPTIONAL property is a [=human palatability|human-palatable=] description of the credential,
+            chosen by the user.
+
+            The [=client=] MUST allow the user to choose this value,
+            MAY or MAY not present that choice during [=registration ceremonies=],
+            and MAY reuse the same value for multiple credentials with the same [=managing authenticator=] across multiple [=[RPS]=].
+            The [=client=] MAY offer the user a default for this value.
+            The [=client=] MAY query the [=authenticator=], by some unspecified mechanism, for a suitable default value.
+
+            If the [=[RP]=] includes a <code>[$credential record/displayName$]</code> [=struct/item=] in [=credential records=],
+            the [=[RP]=] MAY offer this value, if present,
+            as a default value for the <code>[$credential record/displayName$]</code> of the new [=credential record=].
+
+        :   <dfn>vendorDisplayName</dfn>
+        ::  This OPTIONAL property is a [=human palatability|human-palatable=] description of the credential,
+            chosen by the [=authenticator=] vendor.
+
+            The [=client=] MAY query the [=authenticator=], by some unspecified mechanism, for this value,
+            MAY allow the user to override that value, if any,
+            and MAY reuse the same value for multiple credentials across multiple [=[RPS]=].
+
+            If the [=[RP]=] includes a <code>[$credential record/displayName$]</code> [=struct/item=] in [=credential records=],
+            the [=[RP]=] MAY offer this value, if present,
+            as a default value for the <code>[$credential record/displayName$]</code> of the new [=credential record=].
+            {{userDisplayName}} SHOULD take precedence over {{vendorDisplayName}} if present.
     </div>
 
 
--- a/index.bs
+++ b/index.bs
@@ -1174,9 +1174,9 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
             If used, the [=[RP]=] SHOULD use this to describe the [=credential record=] in the user's account settings.
             The [=[RP]=] SHOULD allow the user to choose this value, and MAY allow the user to modify it at will.
 
-            The [=credProps|Credential Properties Extension=] defines two [=credential properties=],
-            {{CredentialPropertiesOutput/userDisplayName}} and {{CredentialPropertiesOutput/vendorDisplayName}},
-            which, when available, MAY be offered as defaults for this value.
+            The [=credProps|Credential Properties Extension=] defines the [=credential property=]
+            {{CredentialPropertiesOutput/authenticatorDisplayName}}
+            which, when available, MAY be offered as a default for this value.
             The [=[RP]=] MAY also derive a default value from the authenticator's [=attestation statement=], if any.
     </dl>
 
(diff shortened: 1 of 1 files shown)


emlun:
(And just to be abundantly clear: this is a proposal for discussion. "Let's not do this, close with no further action" is a perfectly acceptable conclusion.)

index.bs
```diff
@@ -1167,6 +1167,17 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
             when the [=public key credential source=] was [=registration|registered=].
             Storing this in combination with the above [$credential record/attestationObject$] [=struct/item=]
             enables the [=[RP]=] to re-verify the [=attestation signature=] at a later time.
+
+        :   <dfn>displayName</dfn>
```
ndpar, line 1171: Shouldn't the initial value of this field always be equal to [user.displayName](https://w3c.github.io/webauthn/#dictdef-publickeycredentialuserentityjson)? And shouldn't credProps.userDisplayName (always) return the original user.displayName value?

ndpar: I see. In this case `user.displayName` sounds redundant, because I can distinguish my accounts by `user.name`, but this is a different story. Thanks for clarification.

ndpar:
I like the idea to give a user a mean to name the credential during the registration ceremony, and store it in the credential record. But I don't fully understand the envisioned UX. Will the browser show an input field and pass the value in the extension output? Typically, RP has a custom UI to choose the name.

index.bs
arnar: Maybe a different name makes it clearer on first read? E.g. credentialStoreNickname or something in that vein?

emlun: Additional feedback during today's WG meeting was that we should return a single value to the RP; I've merged both properties and renamed it to `authenticatorDisplayName`.

emlun:
Additional feedback during today's WG meeting was that we should return a single value to the RP, so I've merged both properties and renamed it to `authenticatorDisplayName`.

emlun:
@ndpar My envisioned UX is that the browser may have configuration settings for the default nickname for new credentials. Perhaps one for the platform authenticator, if any, and one per external security key. Similarly, passkey providers and external security keys could have such a configuration setting as well, which the browser can check if the user hasn't configured the browser setting(s).

If the user hasn't configured any such setting, the browser may choose to prompt the user to set it, or simply omit the credential nickname. Once set, the user shouldn't need to be prompted again for that authenticator.

RPs would still have custom UI to set credential nicknames, but could set the initial value to this extension output when available.

index.bs
Firstyear, line 1171: I think there is possible confusion here because displayName generally refers to the visible name of the account holder.  I think this needs to be named something more explicit to make it completely clear exactly what it refers to because otherwise RP's may confuse this and end up with dialogs that say "User your credential for site X from Password Manager Steve Bobington".

Firstyear (changes requested):

index.bs
emlun: Yes, my bad - I renamed `{{CredentialPropertiesOutput/displayName}}` but forgot to also rename `[$credential record/displayName$]`. Both are renamed to `authenticatorDisplayName` now.

MasterKale:
> E.g., if I set the displayName as "A" with Browser X for the authenticator and I set the displayName as "B" with Browser Y for the same authenticator; I would get "A" and "B" for the credential generated by the same authenticator with different browsers.

The flow that we imagined didn't involve the browser storing this string, rather it would flow through to the RP so that the RP had something it could nickname the credential. The client displaying this new value would strictly be for the user to optionally change the name that makes it back to the RP, and if the client handled the "enter a name" step of credential naming as part of the tail end of the registration ceremony then it would mean RP's wouldn't all have to implement similar functionality in their sites.

And to be clear I don't think anyone suggested that the client or authenticator store whatever value the user ended up setting for the credential. The authenticator display name would be a default suggested name for the credential that the user could change one way, as it propagates through to the user.

index.bs
Firstyear: Thank you!

Firstyear:
Might be good @emlun to list a string length limit here.

hpsin:
As an RP currently testing passkeys internally, this would solve significant pain points our testers reported. Namely, they didn't really "get" where their passkeys would be usable, and we are reluctant to use heuristics to attempt to brand the passkey ourself. What we end up with eventually is "Touch ID on iPhone" being syncable to one of Chrome or Mac (or maybe both, but you don't know how). Users indicated they had no way of predicting if their passkey would work elsewhere.

Providing a reliable, understandable name and vendor to start would give users confidence in the usability and restrictions on their passkey.

emlun:
@Firstyear I don't think a length limit is appropriate at this layer, since the data flow is directed from authenticator to RP and the RP is already recommended to allow the user to modify the value - so if the authenticator were to return a 2 kB long string the RP is not willing to store, the RP can just tell the user to shorten it before saving it. I wouldn't be against pointing that case out as a recommendation to RPs, though.

A length limit would be appropriate for a new API for configuring this value on the client-to-authenticator layer, but that is not currently part of the extension definition.

Firstyear:
I think we should recommend it then.

Kieun:
Please correct me if I'm wrong.

What I'm understanding is that browsers may offer the way for the user to configure the authenticator name.
This configuration process by the user is not performed during the registration ceremony. If user sets certain name for the authenticator, the browser should have an ability to store such information so that such configuration keep surviving.
If anything isn't configured, the browsers might return default name for the authenticator.
If this is the case, as I've mentioned, such configured information is stored in the browser side and which causes the inconsistency between the names from the different browsers with same authenticator.

As a RP, we would like to get meaningful consistent authenticator display name if the user creates credentials with the same authenticator regardless of browsers which the user runs.

emlun:
Yes, the proposal as currently written would allow that. But presumably, browser implementers would probably agree that it's a bad idea to configure and store a nickname for an external authenticator unless you can store it on that authenticator, for the reasons you describe. Perhaps that is enough, perhaps we'll add an explicit recommendation against it, perhaps there are reasons to do it anyway. But this proposal will also apply to platform authenticators and rich third party passkey provider apps, where that concern doesn't apply, at least not in the same way.

None of this is final, though. I expect the details of the proposal will change as people take some more time to think about it and sketch out some implementation scenarios. In particular we'll have to consider how this should work, if at all, for multi-device passkeys.

emlun:
By the way, to help illustrate the imagined UX flow, see for example https://demo.yubico.com/playground . After you create an account and register a WebAuthn credential (called "security key" or "internal authenticator" in this app; we're working on updated terminology), the app asks you to set a nickname for the credential. The nickname field is prefilled with a default value derived from attestation if available, or otherwise user agent sniffing in the case of platform credentials.

Google and GitHub, and probably many others, also have similar nickname prompts during registration, but don't attempt to prefill the nickname field.

The UX flow with this extension would most likely be the same in all those scenarios, except one could prefill the nickname field with this extension output if available instead of having to fall back on less reliable heuristics.

irew:
Sounds good! We've already seen a few deployments fallback to unreliable means of naming the item in the UI.

Firstyear:
> > Google and GitHub, and probably many others, also have similar nickname prompts during registration, but don't attempt to prefill the nickname field.
> 
> Speaking for GitHub.com, we do attempt to propose a nickname, based on browser and OS, for all security keys today. However, for passkeys we will drop this behavior since e.g. "Chrome on MacOS" is an actively bad nickname for your Dashlane passkey, and will be quite a poor UX when it comes to lifecycle management as the user attempts to understand what passkey they are unregistering. I don't believe users will come up with a better nickname on their own though, since they are unlikely to understand the source of their passkey provider.

There is also a strong likelihood that users will be confused by the UI/UX presented by chrome that may mislead them to misname the key itself. About the best an RP could do here with this nickname field is to fill it in with a nice display name of the site itself. 

Has any user testing or research actually been done on this feature with real humans to see how they will use / misuse this feature proposition?

sbweeden:
If authenticators returned an AAGUID with the credential during registration, even without attestation, wouldn't that also address this same use case? I believe that's what some RP's (including Google) have been asking for.

MasterKale:
> If authenticators returned an AAGUID with the credential during registration, even without attestation, wouldn't that also address this same use case? I believe that's what some RP's (including Google) have been asking for.

Unattested AAGUID came up at the F2F as an acceptable way to solve this problem too, but at the time I don't think many of us thought AAGUIDs without attestation would catch on...

I agree that if we could get authenticator vendors to buy into the idea of returning a valid AAGUID under more circumstances (i.e. no attestation) it would solve this problem and a couple other "UX optimization" problems that RP's can have when trying to communicate passkey identity to users. The kind of problems that can be solved without needing a signature over the values because it's metadata that presents a low risk of weakening RP security posture if someone "spoofs" authenticator identity.

Kieun:
As far as I understand, this requirement is not for identifying each authenticator but for offering user-friendly identifiable name of authenticator. Without integrating with FIDO MDS, the RP could simply get such name.

If the RP is willing to identify each authenticator and do some processing with identifier (e.g., AAGUID), then it would rely on external trust source such as FIDO MDS.
I'm thinking that this is another requirement from the RPs considering more security.

timcappalli:
> If authenticators returned an AAGUID with the credential during registration, even without attestation, wouldn't that also address this same use case? I believe that's what some RP's (including Google) have been asking for.

@shawna-slh this provides more flexibility. For example, being able to say "Tim's 1Password Family Vault" vs "Tim's 1Password Private Vault", etc which is super helpful for users.

index.bs
```diff
@@ -6608,6 +6617,24 @@ At this time, one [=credential property=] is defined: the [=resident key credent
...
+
+            The [=client=] MUST allow the user to choose this value,
+            MAY or MAY not present that choice during [=registration ceremonies=],
+            and MAY reuse the same value for multiple credentials with the same [=managing authenticator=] across multiple [=[RPS]=].
+
+            The [=client=] MAY query the [=authenticator=], by some unspecified mechanism, for this value.
+            The [=authenticator=] MAY allow the user to configure the response to such a query.
+            The [=authenticator=] vendor MAY provide a default response to such a query.
+            The [=client=] MAY consider a user-configured response chosen by the user,
+            and SHOULD allow the user to modify a vendor-provided default response.
```
MasterKale, line 6633: I'm wondering how these lines here compliment the three lines above:

```
The [=client=] MUST allow the user to choose this value,
MAY or MAY not present that choice during [=registration ceremonies=],
and MAY reuse the same value for multiple credentials with the same [=managing authenticator=] across multiple [=[RPS]=].
```

Is the idea to push browsers to implementing UI mid-registration that allows the user to confirm or change an autofill'd input box with a value the authenticator or provider provides? But then there's the aspect of a user customizing the value, and the client needing to remember the change for a given provider across multiple sites...it's hard to understand where the source of truth would be.

sbweeden (approved):
I am fine with these changes. It would provide a more standardized way of communicating back a nickname to the RP, which might be pre-populated by the client or authenticator, or when they do not provide a value, the RP might prompt for it.

index.bs
emlun: The idea was to give browsers the option, but not obligation, to implement mid-registration UI for this. I tried to write this in a way that an input box somewhere in browser (or password manager) settings would qualify as "allowing the user to choose this value", and the browser could simply omit the value if that setting is not configured.

The above applies mostly to platform authenticators. For external authenticators, the value should presumably be stored on the authenticator and made available to clients by some standardized mechanism such as `authenticatorGetInfo()` in CTAP. In this case, "allow the user to choose" could consist of offering a way to configure the value stored on the authenticator when possible, again with the option but not obligation to do so during the registration.

index.bs
```diff
@@ -6608,6 +6617,24 @@ At this time, one [=credential property=] is defined: the [=resident key credent
             If {{rk}} is not present, it is not known whether the credential is a [=discoverable credential=] or a [=server-side credential=].
 
             Note: some [=authenticators=] create [=discoverable credentials=] even when not requested by the [=client platform=]. Because of this, [=client platforms=] may be forced to omit the {{rk}} property because they lack the assurance to be able to set it to [FALSE]. [=[RPS]=] should assume that, if the `credProps` extension is supported, then [=client platforms=] will endeavour to populate the {{rk}} property. Therefore a missing {{rk}} indicates that the created credential is most likely a [=non-discoverable credential=].
+
+        :   <dfn>authenticatorDisplayName</dfn>
+        ::  This OPTIONAL property is a [=human palatability|human-palatable=] description of the credential's [=managing authenticator=],
+            chosen by the user.
+
+            The [=client=] MUST allow the user to choose this value,
+            MAY or MAY not present that choice during [=registration ceremonies=],
```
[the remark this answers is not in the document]
emlun: I considered that, but RFC2119 doesn't define "MAY NOT" as a key word. Strictly speaking "MAY or MAY not" is also equivalent to just "MAY", so perhaps we should simply say that.

emlun:
> or when they do not provide a value, the RP might prompt for it.

This got me thinking - could it make sense to define an [autofill detail token](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#autofill-detail-tokens) like `webauthn-authenticator-nickname` to go with this? I'm thinking perhaps that could be a way to both get the value into an input field easily and also unobtrusively offer the user to save the value to the authenticator configuration after the fact.

Though I can see one complication already: in case there's not exactly one authenticator present, how does the client decide which one's nickname to autofill and optionally update? Intuitively it should of course be "the one that was just used", but that's not so easily defined when the user may wait an indefinite length of time before submitting the nickname, may have performed subsequent WebAuthn ceremonies in other tabs, etc. Perhaps you could associate by credential ID somehow, but I've a feeling this gets much too complicated for such a minor feature.

ndpar (approved):
Looks good to me

arnar:
(I won't be able to review until mid September. Feel free to sub me with someone if needed earlier.)

Firstyear (approved):

MasterKale (approved):

index.bs
```diff
@@ -6608,6 +6617,24 @@ At this time, one [=credential property=] is defined: the [=resident key credent
...
+            and MAY reuse the same value for multiple credentials with the same [=managing authenticator=] across multiple [=[RPS]=].
+
+            The [=client=] MAY query the [=authenticator=], by some unspecified mechanism, for this value.
+            The [=authenticator=] MAY allow the user to configure the response to such a query.
+            The [=authenticator=] vendor MAY provide a default response to such a query.
+            The [=client=] MAY consider a user-configured response chosen by the user,
+            and SHOULD allow the user to modify a vendor-provided default response.
+
+            If the [=[RP]=] includes an <code>[$credential record/authenticatorDisplayName$]</code> [=struct/item=] in [=credential records=],
+            the [=[RP]=] MAY offer this value, if present,
```
MasterKale, line 6636: Wait, this looks off:

> If the [=[RP]=] includes an <code>[$credential record/authenticatorDisplayName$]</code> [=struct/item=] in [=credential records=]

I think this should say "If the authenticator includes", right?

MasterKale (changes requested):

index.bs
emlun: No, this refers to a property of a _credential record_, which is a data structure internal to the RP.

But I do notice that it's ambiguous here whether "this value" refers to the credential record property or the extension output property. I'll try to disambiguate that.

MasterKale (approved):

pascoej (approved):

index.bs
emlun: Oops, looks like this slipped through unaddressed. Fixed in PR #2005.

PR #1887, opened 2023-05-04
Include the "Easily accessing credential data" fields in JSON.

agl:
The WebAuthn API provides accessors to get the SPKI-formatted public key and authenticator data without needing to parse CBOR or handle COSE. However, the JSON structures, prior to this change, didn't include these values giving users an unfortunate choice: either use the accessors and do the JSON encoding yourself, or use the provided `toJSON` function.

But we can have both!

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/pull/1887.html" title="Last updated on May 17, 2023, 7:12 PM UTC (5d62f33)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/1887/dbd8df4...5d62f33.html" title="Last updated on May 17, 2023, 7:12 PM UTC (5d62f33)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1623,8 +1623,8 @@ that are returned to the caller when a new credential is created, or a new asser
         // negotiate a public-key algorithm that the user agent doesn't
         // understand. (See section “Easily accessing credential data” for a
         // list of which algorithms user agents must support.) If using such an
-        // algorithm then the public key must be parsed from attestationObject
-        // directly.
+        // algorithm then the public key must be parsed directly from
+        // attestationObject or authenticatorData.
         Base64URLString publicKey;
         required long long publicKeyAlgorithm;
         // This value contains copies of some of the fields above. See
--- a/index.bs
+++ b/index.bs
@@ -1617,8 +1617,21 @@ that are returned to the caller when a new credential is created, or a new asser
 
     dictionary AuthenticatorAttestationResponseJSON {
         required Base64URLString clientDataJSON;
-        required Base64URLString attestationObject;
+        required Base64URLString authenticatorData;
+        required long long publicKeyAlgorithm;
         required sequence<DOMString> transports;
+
+        // This field is missing if pubKeyCredParams was used to negotiate
+        // a public-key algorithm that the user agent doesn't understand.
+        // (See section “Easily accessing credential data” for a list of
+        // which algorithms user agents must support.) If using such an
+        // algorithm then the public key must be parsed from
+        // attestationObject directly.
+        Base64URLString publicKey;
+
+        // This value contains copies of some of the fields above. See
+        // section “Easily accessing credential data”.
+        required Base64URLString attestationObject;
     };
 
     dictionary AuthenticationResponseJSON {
--- a/index.bs
+++ b/index.bs
@@ -1618,17 +1618,15 @@ that are returned to the caller when a new credential is created, or a new asser
     dictionary AuthenticatorAttestationResponseJSON {
         required Base64URLString clientDataJSON;
         required Base64URLString authenticatorData;
-        required long long publicKeyAlgorithm;
         required sequence<DOMString> transports;
-
-        // This field is missing if pubKeyCredParams was used to negotiate
-        // a public-key algorithm that the user agent doesn't understand.
-        // (See section “Easily accessing credential data” for a list of
-        // which algorithms user agents must support.) If using such an
-        // algorithm then the public key must be parsed from
-        // attestationObject directly.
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -1617,8 +1617,21 @@ that are returned to the caller when a new credential is created, or a new asser
 
     dictionary AuthenticatorAttestationResponseJSON {
         required Base64URLString clientDataJSON;
-        required Base64URLString attestationObject;
+        required Base64URLString authenticatorData;
+        required long long publicKeyAlgorithm;
```
MasterKale, line 1621: Can `publicKeyAlgorithm` be positioned after the new `publicKey` below? I think it'd help to group the two together since they're related.

And reading the comment below for `publicKey`, would `publicKeyAlgorithm` also be missing if "pubKeyCredParams was used to negotiate a public-key algorithm that the user agent doesn't understand"? If so it might help emphasize that point by putting the two values togethere here.

index.bs
```diff
@@ -1617,8 +1617,21 @@ that are returned to the caller when a new credential is created, or a new asser
...
         required sequence<DOMString> transports;
+
```
MasterKale, line 1623: Does this `dictionary` look too cluttered if you remove the empty line breaks?

index.bs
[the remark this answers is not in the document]
agl: Yes, [long long](https://webidl.spec.whatwg.org/#idl-long-long) is WebIDL for int64, which is the closest reasonable mirror of CBOR's 65-bit integers.

index.bs
```diff
@@ -1617,8 +1617,21 @@ that are returned to the caller when a new credential is created, or a new asser
...
+        // a public-key algorithm that the user agent doesn't understand.
+        // (See section “Easily accessing credential data” for a list of
+        // which algorithms user agents must support.) If using such an
+        // algorithm then the public key must be parsed from
+        // attestationObject directly.
+        Base64URLString publicKey;
+
+        // This value contains copies of some of the fields above. See
+        // section “Easily accessing credential data”.
+        required Base64URLString attestationObject;
```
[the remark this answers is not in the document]
agl: I put at the bottom since most RPs shouldn't need it—with the new fields, it's only needed if doing attestation. I hoped that putting it at the bottom would lead most implementators to get the needed information from the easy fields, while obviously keeping the full attestation object for those that need it.

index.bs
agl: I can go either way. I've removed the line breaks; tell me if it's too cluttered now :)

index.bs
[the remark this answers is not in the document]
MasterKale: I understand better your intent in moving `attestationObject` down, and it makes sense with how much we've diminished the importance of attestation for consumer use cases that I'm okay with this now. I'm going to mark this resolved.

index.bs
agl: > The only thing that needs tweaking now is the indentation on the first few lines of comments:

Thanks. (Tabs vs spaces. Need to configure vi on this machine.)

MasterKale (approved):
lgmt :shipit:

index.bs
```diff
@@ -1617,8 +1617,19 @@ that are returned to the caller when a new credential is created, or a new asser
...
         required Base64URLString clientDataJSON;
-        required Base64URLString attestationObject;
+        required Base64URLString authenticatorData;
         required sequence<DOMString> transports;
+        // The publicKey field will be missing if pubKeyCredParams was used to
+        // negotiate a public-key algorithm that the user agent doesn't
+        // understand. (See section “Easily accessing credential data” for a
+        // list of which algorithms user agents must support.) If using such an
+        // algorithm then the public key must be parsed from attestationObject
+        // directly.
```
emlun, line 1627: This can be `authenticatorData` too, right?
```suggestion
        // algorithm then the public key must be parsed directly from
        // attestationObject or authenticatorData.
```

emlun (approved):
LGTM, with or without the below suggestion.

agl:
(from the call of 2023-05-17: good to merge when checks are happy.)

emlun:
I realize belatedly that we probably also need instructions on how to populate these new fields, like we have for `clientExtensionResults` and `transports`:
https://github.com/w3c/webauthn/blob/6dfbdbaad8425140030afacb7de27eb13e7872ce/index.bs#L1592-L1602

I'll open a follow-up PR.

emlun:
I'll fix the [above issue](https://github.com/w3c/webauthn/pull/1887#issuecomment-1591886120) in PR #1906.

PR #1923, opened 2023-07-18
Add new getClientCapabilities method

timcappalli:
This PR adds a new method, `getClientCapabilities()` and a matching enum for the response (currently `hybrid-transport` and `client-pin-entry`).

These capabilities can help an RP craft a sign in or registration experience based on client capabilities. For example, an RP that requires user verification may not want to offer security key registration on a client where PIN entry is not supported. Another example would be a smart TV or kiosk where only hybrid is available (no local platform authenticator). In this scenario, offering a username field doesn't make a lot of sense and a "Sign in with a passkey" button would offer a better user experience.

***

### :boom: Error: connect ETIMEDOUT 173.230.149.95:443 :boom: ###

[PR Preview](https://github.com/tobie/pr-preview#pr-preview) failed to build. _(Last tried on Jan 9, 2024, 6:30 PM UTC)_.

<details>
<summary>More</summary>

:rotating_light: [CSS Spec Preprocessor](https://api.csswg.org/bikeshed/) - CSS Spec Preprocessor is the web service used to build Bikeshed specs.

:link: [Related URL](https://api.csswg.org/bikeshed/?url=https%3A%2F%2Fraw.githubusercontent.com%2Fw3c%2Fwebauthn%2Fef54380aebf78db9bcadb1b5751c1158e1d802d7%2Findex.bs&force=1&md-warning=not%20ready)

_If you don't have enough information above to solve the error by yourself (or to understand to which web service the error is related to, if any), please [file an issue](https://github.com/tobie/pr-preview/issues/new?title=Error%20not%20surfaced%20properly&body=See%20w3c/webauthn%231923.)._
</details>


emlun:
Would this make these expressions (ignoring `Promise` indirection) equivalent?

- ```
  isPasskeyPlatformAuthenticatorAvailable()
  ```
- ```
  isUserVerifyingPlatformAuthenticatorAvailable()
    || getClientCapabilities().includes("hybrid-transport")
  ```

timcappalli:
@emlun logically, yes. isPPAA() is for the masses (basic use cases), whereas gCC() is for more advanced use cases, and can easily be expanded.

index.bs
```diff
@@ -2802,6 +2802,26 @@ Note: Invoking this method from a [=browsing context=] where the [=Web Authentic
...
+
+[=[WRPS]=] use this method to determine the availability of a limited set of [=WebAuthn Client|client=] capabilities to offer certain workflows and experiences to users. For example, an RP may offer a sign in button on clients where only {{AuthenticatorTransport/hybrid}} transport is available (instead of showing a username field), or offer security key enrollment on [=WebAuthn Client|clients=] that support client PIN entry.
+
+Upon invocation, the [=client=] employs a [=client platform=]-specific procedure to discover availablity of these capabilities.
+
+This method has no arguments and returns a set of strings.
+
+<xmp class="idl">
+    partial interface PublicKeyCredential {
+        static Promise<ClientCapabilities> getClientCapabilities();
```
agl, line 2817: Should this return a promise for a _sequence_ of `ClientCapabilities`? (Or, actually, of `DOMString` given [2.1.1](https://www.w3.org/TR/webauthn-2/#sct-domstring-backwards-compatibility).)

index.bs
```diff
@@ -3967,6 +3987,24 @@ Note: The {{UserVerificationRequirement}} enumeration is deliberately not refere
         interest of minimizing disruption to the user interaction flow).
 </div>
 
+### Client Capabilities Enumeration (enum <dfn enum>ClientCapabilities</dfn>) ### {#enum-clientCapabilities}
+
+<xmp class="idl">
+    enum ClientCapabilities {
```
agl, line 3993: All other enum names in the spec are singular, so perhaps `ClientCapability`?

index.bs
```diff
@@ -3967,6 +3987,24 @@ Note: The {{UserVerificationRequirement}} enumeration is deliberately not refere
...
+        "hybrid-transport",
+        "client-pin-entry"
+    };
+</xmp>
+
+This enumeration defines a limited set of client capabilities which a [=[WRP]=] may evaluate to offer certain workflows and experiences to users.
+
+<div dfn-type="enum-value" dfn-for="ClientCapabilities">
+    :   <dfn>hybrid-transport</dfn>
+    ::  The [=WebAuthn Client=] supports {{AuthenticatorTransport/hybrid}} transport.
```
agl, line 4003:
```suggestion
    ::  The [=WebAuthn Client=] supports the {{AuthenticatorTransport/hybrid}} transport.
```

index.bs
```diff
@@ -3967,6 +3987,24 @@ Note: The {{UserVerificationRequirement}} enumeration is deliberately not refere
...
+
+<xmp class="idl">
+    enum ClientCapabilities {
+        "hybrid-transport",
+        "client-pin-entry"
+    };
+</xmp>
+
+This enumeration defines a limited set of client capabilities which a [=[WRP]=] may evaluate to offer certain workflows and experiences to users.
+
```
agl, line 4000: Add a "Note:" about the enum not being referenced, as done for other enums? ([Example](https://www.w3.org/TR/webauthn-2/#enum-attachment))

agl (approved):

index.bs
```diff
@@ -3967,6 +3987,26 @@ Note: The {{UserVerificationRequirement}} enumeration is deliberately not refere
...
+Note: The {{ClientCapability}} enumeration is deliberately not referenced, see [[#sct-domstring-backwards-compatibility]].
+
+<div dfn-type="enum-value" dfn-for="ClientCapability">
+    :   <dfn>hybrid-transport</dfn>
+    ::  The [=WebAuthn Client=] supports the {{AuthenticatorTransport/hybrid}} transport.
```
MasterKale, line 4005: I wonder if "supports" is strong enough a signal. Yes, in absolute terms Chrome "supports" hybrid even when it lacks the OS-level Bluetooth permission. But in relative terms Chrome cannot "support" hybrid because it lacks access to Bluetooth. Which nuance is this intending to convey to an RP?

akshayku (approved):

index.bs
```diff
@@ -3967,6 +3987,26 @@ Note: The {{UserVerificationRequirement}} enumeration is deliberately not refere
...
+
+    :   <dfn>client-pin-entry</dfn>
+    ::  The [=WebAuthn Client=] supports PIN entry for [=user verification=] or other [=authenticator=] operations.
```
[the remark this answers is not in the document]
MasterKale: There are some existing platforms that can't do UV because PIN entry isn't supported so this already won't necessarily be `true.

index.bs
MasterKale: This is more of an "absolute support": a client can confirm all the pieces are there for use of hybrid, even though it may require user remediation to make it possible to successfully complete a ceremony over hybrid.

MasterKale (approved):

MasterKale:
I think it would be great if we added an equivalent capability enums for `isUVPAA()` to this as well, and maybe `isPPAA()`? Maybe the conditional mediation check too? The idea is future capabilities checks get added as enums to this, and _maybe_ there are convenience methods added too that make an appropriate call to this method behind the scenes.

timcappalli:
> I think it would be great if we added an equivalent capability enums for `isUVPAA()`

user-verifying platform authenticator is not a client capability

emlun:
Some discussion points from the 2023-07-26 WG call:

- (Fixed) `client-pin-entry` capability is too blunt; some external authenticators support internal UV, but the absence of this capability would make it seem like UV is not supported for those authenticators.
- Given that [this together with `isUVPAA()` duplicates `isPPAA()`](https://github.com/w3c/webauthn/pull/1923#issuecomment-1650281617), I don't think this and `isPPAA()` should coexist. I prefer this `getClientCapabilities()` approach over `isPPAA()`, since the former is extensible and more explicit. If we merge this, I think we should revert #1901. (sorry Tim, no offense!)

emlun:
>it would be great if we added an equivalent capability enums for [...] Maybe the conditional mediation check too?

Note that [`isConditionalMediationAvailable()`](https://w3c.github.io/webauthn/#dom-publickeycredential-isconditionalmediationavailable) is defined in the Credential Management spec, and WebAuthn only overrides it, so unlike `isPPAA()` we cannot delete `isConditionalMediationAvailable()` in favour of `getClientCapabilities()`. So again in the interest of de-duplication, I think we should keep the conditional mediation check separate.

> user-verifying platform authenticator is not a client capability

How is it not?

But perhaps that's a moot point anyway. Again, I don't think there should be both a `uvpaa` capability and the `isUVPAA()` method, and removing `isUVPAA()` at this point would be a breaking change. Making it deprecated isn't very appealing either since we probably can't ever actually follow through with deleting it.

timcappalli:
> Given that https://github.com/w3c/webauthn/pull/1923#issuecomment-1650281617, I don't think this and isPPAA() should coexist. I prefer this getClientCapabilities() approach over isPPAA(), since the former is extensible and more explicit. If we merge this, I think we should revert https://github.com/w3c/webauthn/pull/1901. (sorry Tim, no offense!)

@emlun I think I'd rather withdraw this PR than lose isPPAA() as that new method is more valuable to a larger number of developers.

emlun:
Ok, let's hear some more opinions before we decide either way (keep `isPPAA()`, keep `getCC()` or keep both).

In the meantime, this just occurred to me: perhaps instead of a list of values, this should be structured as object attributes like the [`credProps` output](https://www.w3.org/TR/2021/REC-webauthn-2-20210408/#dictdef-credentialpropertiesoutput)?

Say that we add a new capability value for a feature that already exists in some clients. Some clients immediately update and start reporting the new enum value, but some clients are slower to update and for a while don't report the new value - even though they do in fact support the feature. This leads RPs to incorrectly conclude that the feature isn't available.

So perhaps it would be better to structure the data model like this?

```
    static Promise<ClientCapabilities> getClientCapabilities();
};
dictionary ClientCapabilities {
    boolean hybridTransport;
    // Additional properties can be added here
};
```

so then, if `cap = await getClientCapabilities()`:

- `cap["feature"] === true`: `feature` is known to be supported
- `cap["feature"] === false`: `feature` is known to be not supported
- `cap["feature"] === undefined`: `feature` availability is unknown

MasterKale:
> Note that [`isConditionalMediationAvailable()`](https://w3c.github.io/webauthn/#dom-publickeycredential-isconditionalmediationavailable) is defined in the Credential Management spec, and WebAuthn only overrides it, so unlike `isPPAA()` we cannot delete `isConditionalMediationAvailable()` in favour of `getClientCapabilities()`. So again in the interest of de-duplication, I think we should keep the conditional mediation check separate.

In response to this point, I don't want [a point I made earlier](https://github.com/w3c/webauthn/pull/1923#issuecomment-1652394315) to get glossed over:

> ...The idea is future capabilities checks get added as enums to this, and maybe there are convenience methods added too that make an appropriate call to this method behind the scenes.

I don't see `getCC()` replacing methods like `isPPAA()`. Rather I see there being value in a method `isPPAA()` essentially calling `getCC(['uvpaa', 'hybrid'])` behind the scenes. There's value in having "convenience methods" like `isPPAA()` to make it easier to use (and provide guidance on using a la https://passkeys.dev) WebAuthn, while still enabling RP's with more advanced use cases to drop down a level and call `getCC()` directly.

Firstyear:
> > ...The idea is future capabilities checks get added as enums to this, and maybe there are convenience methods added too that make an appropriate call to this method behind the scenes.
> 
> I don't see `getCC()` replacing methods like `isPPAA()`. Rather I see there being value in a method `isPPAA()` essentially calling `getCC(['uvpaa', 'hybrid'])` behind the scenes. There's a developer usability win in having "convenience methods" like `isPPAA()` to make it easier to use (and provide guidance on using a la https://passkeys.dev) WebAuthn, while still enabling RP's with more advanced use cases to drop down a level and call `getCC()` directly.

I am worried we are getting into "too complex" territory. It's always the case that with standards like webauth, oauth2, oidc, ldap, or anything else, that while the standard may support 100 different options, what evolves is a small subset that is actually used day to day by implementors. For example, LDAP spans many rfc's with complex search options and extensions, and in reality most people just use search and bind.

I'm worried we are heading down this path - most RP's will never use 90% of what is being created here like getCC or isPPAA because these features end up causing UI/UX issues rather than solving them. 

For example, isUVPAA() exists and is now being forgotten by the wayside because we keep chasing these micro UI/UX tweaks that no one seems to need. isPPAA is probably going to end up the same, as RP's realise that security keys like yubikeys exist and they can't just force users to use their phones with extremely slow and awkward cable workflows (did you know in latency it takes ~15 seconds to use phone for cable in AU? And this doesn't even include all the time to get the camera open and dive through all the choose your own adventure menus chrome gives you) . 

Even getCC querying for hybrid begs a lot of questions to what you are trying to achieve since the browser itself is already going to offer cable as an option without you needing to tweak anything. I don't think many rp's really even need this capability, since the *user* should be the one choosing what they want. 

I think you need to expand on what you are trying to actually achieve here and how that works for the positive and inverse cases, because currently this seems like an internal MS detail or plan leaking into the specification rather than a change that I can broadly see RP's embracing.

agl:
> Ok, let's hear some more opinions before we decide either way (keep isPPAA(), keep getCC() or keep both).

Chrome doesn't believe that existing functions should be removed, but a) we are against adding more and more static methods to PublicKeyCredential in the future, and b) favour having this function return values that duplicate the return values of the existing feature-detection functions so that sites could just call this for all their needs.

arianvp:
I suggested something similar to this PR as an alternative to isPPAA in the original discussion about merging isPPAA so I'm in favor of this change.

I still think merging isPPAA was a mistake though

https://github.com/w3c/webauthn/pull/1901#issuecomment-1593179039

timcappalli:
See issue #1937 for context on the most recent commit prior to reviewing this updated PR.

index.bs
```diff
@@ -3967,6 +3987,33 @@ Note: The {{UserVerificationRequirement}} enumeration is deliberately not refere
         interest of minimizing disruption to the user interaction flow).
 </div>
 
+### Client Capability Set (dictionary <dfn dictionary>ClientCapability</dfn>) ### {#dictionary-client-capability}
+
+<xmp class="idl">
+    dictionary ClientCapability {
```
MasterKale, line 3993: I wonder if there's not a reason to pluralize this, to `ClientCapabilities` instead. It's not an Enum anymore so I'd say the plural form is alright.

index.bs
```diff
@@ -3967,6 +3987,33 @@ Note: The {{UserVerificationRequirement}} enumeration is deliberately not refere
         interest of minimizing disruption to the user interaction flow).
 </div>
 
+### Client Capabilities Set (dictionary <dfn dictionary>ClientCapabilities</dfn>) ### {#dictionary-client-capabilities}
+
+<xmp class="idl">
+    dictionary ClientCapabilities {
```
agl, line 3993: Encoding the possible features as IDL defeats the point. We might as well have static functions in that case!

agl (changes requested):

index.bs
timcappalli: > Encoding the possible features as IDL defeats the point. We might as well have static functions in that case!

@emlun / @MasterKale while the dictionary model is a bit cleaner, using an enum provides a stronger long term solution as previously defined here: https://w3c.github.io/webauthn/#sct-domstring-backwards-compatibility

I think this should be reverted back to an enum w/ DOMStrings.

MasterKale: I'm not entirely sure how it'd map to IDL, but I do think we should define the return value in a way that would allow for devs to destructure the return value and easily know which capabilities are available (with `undefined` communicating that the client doesn't know how to determine support for that capability):

For example

```ts
type ClientCapabilities = {
  userVerifyingPlatformAuthenticator?: boolean; // true | false | undefined
  hybridTransport?: boolean;
  passkeyPlatformAuthenticator?: boolean;
  conditionalMediation?: boolean;
}

type getClientCapabilities = () => Promise<ClientCapabilities>;
```

And then in the real world devs could do this:

```js
const {
  userVerifyingPlatformAuthenticator,
  hybridTransport,
  passkeyPlatformAuthenticator,
  conditionalMediation,
} = await getClientCapabilities();
```

I think that'd make for a pleasantly usable method.

emlun: I still think it seems significant to communicate a difference between a capability being positively unsupported vs. unknown, though. How should we do that, then? I can imagine a few options:

 1. Return two lists: features supported and features unsupported:
    ```js
    { capable: ["hybrid-transport", "conditional-mediation"], incapable: ["user-verifying-platform-authenticator"] }
    ```
 2. Return two lists: features supported and list of all known features:
    ```js
    { capable: ["conditional-mediation"], all: ["conditional-mediation", "hybrid-transport", "passkey-platform-authenticator", "user-verifying-platform-authenticator"] }
    ```
 3. Return a [`maplike<DOMString, boolean>`](https://webidl.spec.whatwg.org/#idl-maplike) instead of a `dictionary`
 4. Do none of the above, and instead forbid future additions of `ClientCapability` values that are already supported in browsers.

Of these I think I like (3) best. Thoughts, other ideas? Or should we let it be ambiguous whether a capability is unsupported or unknown?

MasterKale: @emlun The problem of getting TS to understand when new capabilities drop before the official TypeScript types get updated isn't so bad. For example, in your example we could define a type that extends the official one and coerce TS into understanding:

```ts
// Platform library (e.g., Node) providing TypeScript type declaration converted from WebIDL
export type ClientCapabilities = {
  hybridTransport: boolean;
}

export type ClientCapabilitiesFuture = ClientCapabilities & {
  conditionalMediation: boolean;
};

// Application code below
const foo = await PublicKeyCredential.getClientCapabilities() as ClientCapabilitiesFuture;

// Assume browsers now support a `conditionalMediation` property but the above definition library hasn't yet been updated
console.log(foo.conditionalMediation);
```

Then projects could, on a case-by-case basis integrate additional capabilities that browsers might independently start adding return values for. But could we define a data structure in IDL that would define the four capabilities we'd include in the spec, but open the door for "but also there might be some other values in there based on browsers expanding this method's insights"?

> 3\. Return a [`maplike<DOMString, boolean>`](https://webidl.spec.whatwg.org/#idl-maplike) instead of a `dictionary`:
>     ```webidl
>     interface ClientCapabilities {
>       maplike<DOMString, boolean>;
>     }
>     ```

If we can't have the hybrid I proposed above, I suppose I'd be fine with this. The downside is that every RP using the method and TypeScript would have to independently define a "`KnownClientCapabilities`" type to case the return value to with the known values. And it'd be harder to catch misspellings...this is a tough one 🤔

agl: I don't know enough IDL to know when to use `maplike` or `record`, but `Promise<record<...>>` exists in Web APIs today, but `Promise<maplike<...>>` does not. (I think if we're returning an object, we can define an interface that is `maplike`, and with additional properties, or we can just record a `record` object.)

index.bs
```diff
@@ -2779,25 +2779,33 @@ Note: Invoking this method from a [=browsing context=] where the [=Web Authentic
...
+    interface GetClientCapabilitiesResult {
+        readonly maplike<DOMString, boolean>;
     };
 </xmp>
 
+Keys in {{GetClientCapabilitiesResult}} SHOULD be members of {{ClientCapability}}.
+
+When the value for a given capability is [TRUE], the feature is known to be currently supported by the client.
+When the value for a given capability is [FALSE], the feature is known to be not currently supported by the client.
+When a given capability is `undefined`, the availibility of the client feature is not known.
```
MasterKale, line 2807:
```suggestion
When a given capability is `undefined`, the availability of the client feature is not known.
```

emlun (commented):
LGTM aside from some minor issues.

index.bs
```diff
@@ -3967,6 +3976,38 @@ Note: The {{UserVerificationRequirement}} enumeration is deliberately not refere
...
+
+<div dfn-type="enum-value" dfn-for="ClientCapability">
+    :   <dfn>user-verifying-platform-authenticator</dfn>
+    ::  The [=WebAuthn Client=] supports usage of a [=user-verifying platform authenticator=].
+
+    :   <dfn>hybrid-transport</dfn>
+    ::  The [=WebAuthn Client=] supports usage of the {{AuthenticatorTransport/hybrid}} transport.
+
+    :   <dfn>passkey-platform-authenticator</dfn>
+    ::  The [=WebAuthn Client=] supports usage of a [=passkey platform authenticator=], locally and/or via {{AuthenticatorTransport/hybrid}} transport.
```
emlun, line 4003: This description doesn't really align with our definition of [platform authenticator](https://w3c.github.io/webauthn/#platform-attachment). In [§6.2.1. Authenticator Attachment Modality](https://w3c.github.io/webauthn/#sctn-authenticator-attachment-modality), following the definition of _platform authenticator_, we state (emphasis added):

>Some [platform authenticators](https://w3c.github.io/webauthn/#platform-authenticators) could possibly also act as [roaming authenticators](https://w3c.github.io/webauthn/#roaming-authenticators) depending on context. For example, a [platform authenticator](https://w3c.github.io/webauthn/#platform-authenticators) integrated into a mobile device could make itself available as a [roaming authenticator](https://w3c.github.io/webauthn/#roaming-authenticators) via Bluetooth. In this case [clients](https://w3c.github.io/webauthn/#client) running on the mobile device would recognise the authenticator as a [platform authenticator](https://w3c.github.io/webauthn/#platform-authenticators), while **[clients](https://w3c.github.io/webauthn/#client) running on a different [client device](https://w3c.github.io/webauthn/#client-device) and communicating with the same authenticator via Bluetooth would recognize it as a [roaming authenticator](https://w3c.github.io/webauthn/#roaming-authenticators)**.

So the sentence "using a platform authenticator via hybrid transport" doesn't make sense with these definitions, because an authenticator reached via hybrid transport is by definition not a platform authenticator from the perspective of the client (even if it is a platform authenticator from the perspective of a different client device). Likewise, the definition of [passkey platform authenticator](https://w3c.github.io/webauthn/#passkey-platform-authenticator) is similarly inconsistent with its own name.

I don't think we currently have a term to describe this distinction between a roaming authenticator embedded in a different client device and a distinct roaming authenticator not embedded in any client device. I think we need a better name for this than "passkey platform authenticator", but that issue may be out of scope for this PR.

index.bs
timcappalli: I think this stream of comments is resolved?

emlun: Looking at this again, I think @agl is right that `record` is probably more appropriate for this. But I'd say we could handle that in a separate PR if you want to land this one @timcappalli.

index.bs
```diff
@@ -2779,25 +2779,34 @@ Note: Invoking this method from a [=browsing context=] where the [=Web Authentic
...
+This method has no arguments and returns a map of capability keys to Boolean values. 
 
 <xmp class="idl">
     partial interface PublicKeyCredential {
-        static Promise<boolean> isPasskeyPlatformAuthenticatorAvailable();
+        static Promise<GetClientCapabilitiesResult> getClientCapabilities();
+    };
+
+    [SecureContext, Exposed=Window]
+    interface GetClientCapabilitiesResult {
```
agl, line 2798: This is in a global namespace so I think would be better named `PublicKeyCredentialClientCapabilities`.

index.bs
agl: I'm not sure how to implement this enum value unless it's just `hybrid-transport` OR `user-verifying-platform-authenticator`. Is it? If so, it that worthwhile?

agl (approved):
I've done a quick implementation and the IDL looks good.

index.bs
MasterKale: It's an either/or, from the PR that added `isPPAA()`:

image

As for the value of this, it's intended to be convenient for an RP to use this one value than to have to know to write the either/or check themself. They could achieve the same thing using the uvpa and hybrid capabilities.

akshayku (approved):

index.bs
MasterKale: For posterity I pulled this from https://github.com/w3c/webauthn/pull/1901

index.bs
```diff
@@ -3967,6 +3976,38 @@ Note: The {{UserVerificationRequirement}} enumeration is deliberately not refere
         interest of minimizing disruption to the user interaction flow).
 </div>
 
+### Client Capability Enumeration (enum <dfn enum>ClientCapability</dfn>) ### {#enum-clientCapability}
+
+<xmp class="idl">
+    enum ClientCapability {
+        "user-verifying-platform-authenticator",
+        "hybrid-transport",
```
nsatragno, line 3984: Would this be different from [`navigator.bluetooth.getAvailability()`](https://webbluetoothcg.github.io/web-bluetooth/#dom-bluetooth-getavailability)? We can always recommend people call that method. And that way, we would not have to reimplement any privacy guards (see step 2, if the user denied permission previously the ua can return false).

As discussed in our meeting, user agents should not indicate bluetooth powered-on state to rps silently.

jschanck:
I also implemented a sketch of this, and I didn't run into any problems.

I'd like to see some discussion of fingerprinting vectors (particularly the `user-verifying-platform-authenticator` and `hybrid-transport` capabilities) added to the Privacy Considerations section. I'm surprised that there is not already an entry on `isUserVerifyingPlatformAuthenticatorAvailable()`.

jschanck (approved):

MasterKale:
As an exercise to make sure I understand how the `passkey-platform-authenticator` capability will XOR `user-verifying-platform-authenticator` and `hybrid-transport` capabilities:

```
uvpaa | hybrid |  ppa
------|--------|-------
  y   |   y    |   y
  y   |   n    |   y
  y   |   ?    |   y
  n   |   y    |   y
  n   |   n    |   n
  n   |   ?    |   n
  ?   |   y    |   y
  ?   |   n    |   n
  ?   |   ?    |   n
```

Assuming I didn't miss something, I wonder if anything here might be considered as a net new fingerprinting vector 🤔

emlun:
> As an exercise to make sure I understand how the `passkey-platform-authenticator` capability will encode `user-verifying-platform-authenticator` and `hybrid-transport` capabilities:

The discussion (on the 2023-11-15 WG call) centered around how the client would have to handle these flags in order to not inadvertently leak fingerprinting information. The motivating example being that Bluetooth may or may not be available on Windows devices, so a client may not want to report a `hybrid` capability in order to not leak that as a fingerprinting bit. In that case, the client would _also_ need to consider that `passkey-platform-authenticator` is equivalent to `user-verifying-platform-authenticator OR hybrid`, so if `uvpaa: false, ppa: true`, then the RP can conclude `hybrid: true` (and also `uvpaa: false, ppa: false => hybrid: false`).

So I would rather say that `passkey-platform-authenticator` would need to be undefined if either of `user-verifying-platform-authenticator` or `hybrid` is undefined (or, equivalently, the former should be present if and only if BOTH of the latter are present):

```
uvpaa | hybrid |  ppa
------|--------|-------
  y   |   y    |   y
  y   |   n    |   y
  y   |   ?    |   ?
  n   |   y    |   y
  n   |   n    |   n
  n   |   ?    |   ?
  ?   |   y    |   ?
  ?   |   n    |   ?
  ?   |   ?    |   ?
```

emlun:
Actually, no - to be more precise, `passkey-platform-authenticator` would have to be undefined in cases where its value would uniquely determine the true value of an undefined `hybrid` value. Assuming that `uvpaa` is always defined, because the method still exists, the only case `ppa` would leak information is when `uvpaa` is `true`. So in order to not leak a bluetooth fingerprinting bit, the combinations that could be emitted are these:

|  uvpaa  | hybrid  |   ppa   |
|:-------:|:-------:|:-------:|
| `true`  | ?       | `true`  |
| `true`  | `true`  | `true`  |
| `true`  | `false` | `true`  |
| `false` | ?       | ?       |
| `false` | `true`  | `true`  |
| `false` | `false` | `false` |

timcappalli:
@emlun @jschanck how deep do we want to go in the privacy considerations? I added some initial text in [ecc19d4](https://github.com/w3c/webauthn/pull/1923/commits/ecc19d46c98fcb9cfa185da04ae5b0901ada6d79). It is very high level but brings attention to the concerns. 

Do we want to be recommending a certain behavior and spelling out specifics (e.g. your table above)?

jschanck:
I'm happy with that level of detail. I would more explicitly call out the fingerprinting risk, and I would also note that lack of support can be a fingerprinting risk. So I would phrase your second paragraph as:

> The client's support or lack of support of a WebAuthn capability may pose a fingerprinting risk. Client implementations MAY wish to limit capability disclosures based on client policy and/or user consent.

timcappalli:
> I'm happy with that level of detail. I would more explicitly call out the fingerprinting risk, and I would also note that lack of support can be a fingerprinting risk.

Thanks @jschanck. addressed in 594e338.

index.bs
nsatragno: (this got addressed below)

nsatragno (approved):
Approved, thank you!

MasterKale (changes requested):
Hmm, I'm checking out Safari Tech Preview 185's implementation of `getClientCapabilities()` and am wondering if we need to revisit this PR's definition of the shape of the return value (see [here](https://github.com/w3c/webauthn/pull/1923#discussion_r1283627219) for past discussion on this).

The return value is currently defined as such:

```
[SecureContext, Exposed=Window]
interface PublicKeyCredentialClientCapabilities {
    readonly maplike<DOMString, boolean>;
};
```

We're seeing now a client interpret this to return a brand new `PublicKeyCredentialClientCapabilities` that looks like a `Map` in all but name. This means developers have to use `.get()` on the returned `capabilities` to pull anything out:

```js
const capabilities = await PublicKeyCredential.getClientCapabilities();
console.log(capabilities['user-verifying-platform-authenticator']); // undefined
console.log(capabilities.get('user-verifying-platform-authenticator')); // true
```

Additionally, it's impossible to destructure this object, especially since entries in `ClientCapability` are kebab-cased:

```js
const { 'user-verifying-platform-authenticator' } = capabilities;
// SyntaxError: Unexpected token '}'. Expected a ':' prior to a named destructuring property.
```

Trying to preview entries in a `Map`-like is not intuitive either:

Screenshot 2023-12-21 at 9 34 34 AM

In light of this I'd like to pull out the idea I had for the shape of this method and ask we reconsider how we define the output of `getClientCapabilities()`:

> I'm not entirely sure how it'd map to IDL, but I do think we should define the return value in a way that would allow for devs to destructure the return value and easily know which capabilities are available (with undefined communicating that the client doesn't know how to determine support for that capability):
> 
> For example
> 
> ```ts
> type ClientCapabilities = {
>   userVerifyingPlatformAuthenticator?: boolean; // true | false | undefined
>   hybridTransport?: boolean;
>   passkeyPlatformAuthenticator?: boolean;
>   conditionalMediation?: boolean;
> }
> 
> 
> type getClientCapabilities = () => Promise<ClientCapabilities>;
> ```
> 
> And then in the real world devs could do this:
> 
> ```ts
> const {
>   userVerifyingPlatformAuthenticator,
>   hybridTransport,
>   passkeyPlatformAuthenticator,
>   conditionalMediation,
> } = await getClientCapabilities();
> ```
> 
> I think that'd make for a pleasantly usable method.

Use of kebab-case here actually hurts the developer ergonomics of this method, and I think we can do better before we finalize this.

timcappalli:
Switching the values to camel case is fine with me if it improves the developer experience. This is a fairly minor change. 

Any concerns if I make that change @agl, @emlun, @nsatragno, @pascoej, @jschanck?

MasterKale:
> Switching the values to camel case is fine with me if it improves the developer experience. This is a fairly minor change.

To be clear I'm also wanting us to potentially change the `maplike` here...

```
[SecureContext, Exposed=Window]
interface PublicKeyCredentialClientCapabilities {
    readonly maplike<DOMString, boolean>;
};
```

...to something else that encourages clients to return basic, barebones objects. I think `maplike` gives the impression we want a `Map` used here when I think we should aim to return a populated `{}`.

agl:
I've no problem with changing how things are spelt, but I did an implementation in Chromium of this and I don't recall it being terribly awkward. But please poke me about this after the holidays when I can dig that code up and check.

agl:
Yep, `maplike` does require using `get` etc.

`record<DOMString, string>` might be better, coupled with camelCase?

Screenshot 2023-12-26 at 3 45 19 PM

MasterKale (approved):
Approving to clear up my own request for change since I made the changes I requested.

MasterKale:
Tagging @emlun @agl @pascoej @jschanck to confirm the change to `record<DOMString, boolean>` with camelCased values. Given that next week seems to be our last chance to get this merged to meet some browser vendor deadlines I'm trying to get your eyes on this to finalize any additional changes async, so that next week's meeting can involve an uneventful merge.

index.bs
```diff
@@ -2793,25 +2793,31 @@ Note: Invoking this method from a [=browsing context=] where the [=Web Authentic
...
 <xmp class="idl">
     partial interface PublicKeyCredential {
-        static Promise<boolean> isPasskeyPlatformAuthenticatorAvailable();
+        static Promise<PublicKeyCredentialClientCapabilities> getClientCapabilities();
     };
+
+    typedef record<DOMString, boolean> PublicKeyCredentialClientCapabilities;
 </xmp>
 
+[=map/Keys=] in {{PublicKeyCredentialClientCapabilities}} MUST be sorted in ascending lexicographical order.
```
agl, line 2814: I think this comment about sorting is moot with a `record`? We're now in the world of Javascript object order which is complex and not something that this should have an opinion on, I think.

MasterKale: Makes sense to me, and I'd hazard to assume that no developer would go looking to make sure this is true. I've removed it.

emlun: We added this [for the sake of minimizing fingerprinting opportunities](https://github.com/w3c/webauthn/pull/1923#discussion_r1341194415), not as something for developers to use - just like `getTransports()` is required to return a sorted list for the same reason. [Record types](https://webidl.spec.whatwg.org/#idl-record) are also defined to be [ordered maps](https://infra.spec.whatwg.org/#ordered-map) just like `maplike`s, so I don't think the requirement is moot. I think the order requirement should stay, but I don't feel very strongly about it.

index.bs
```diff
@@ -2793,25 +2793,30 @@ Note: Invoking this method from a [=browsing context=] where the [=Web Authentic
...
     };
+
+    typedef record<DOMString, boolean> PublicKeyCredentialClientCapabilities;
 </xmp>
 
+The set of [=map/keys=] SHOULD equal the set of [=enumeration values=] of {{ClientCapability}}.
+
+When the value for a given capability is [TRUE], the feature is known to be currently supported by the client.
+When the value for a given capability is [FALSE], the feature is known to be not currently supported by the client.
+When a capability does not [=map/exist=] as a key, the availability of the client feature is not known.
```
emlun, line 2818: We should link to `{#sctn-disclosing-client-capabilities}` somewhere, maybe here?

```suggestion
When a capability does not [=map/exist=] as a key, the availability of the client feature is not known.
See also [[#sctn-disclosing-client-capabilities]].
```

emlun (commented):
Changing `maplike` to `record` looks good to me, thanks for catching that!

index.bs
MasterKale: @emlun I'm even less opinionated here, I didn't consider order of keys being something you could use for fingerprinting. I'm happy to add it back in if it prevents us from having to revisit the user privacy discussion (which, to be clear, I thought was good and ended with us in a good spot.)

@agl when you mention that "JavaScript object order...is complex" was that from an implementation perspective? Does it put an undue burden on client developers if the sorting requirement is restored?

index.bs
MasterKale: Oh, you know what, I think I found a good place for that link to the Disclosing Client Capabilities section, how about this?

Screenshot 2024-01-08 at 9 48 33 AM

emlun: Hmyeah, good point - right now it is a bit ambiguous which of "The set of keys SHOULD equal the set of enumeration values" and "Client implementations MAY wish to limit capability disclosures" takes precedence. I don't think `null` or `undefined` are legal values of a `record<DOMString, boolean>`, so we'd need to tweak the type a bit in order to support your suggestion. It seems to me like both `record<DOMString, boolean | undefined>` and `record<DOMString, boolean?>` (nullable boolean) are legal, so I think we could choose either `undefined` or `null` for supported but undisclosed features.

But on the other hand, I think it's better to rephrase the "The set of keys SHOULD equal the set of enumeration values" requirement instead of complicating the type:

- `undefined` vs. `null`: A key being absent vs. its value being `undefined` are largely equivalent in JavaScript implementations (one difference being that `Object.keys({foo: undefined}) != Object.keys({})`), so between `undefined` and `null` I think we should choose the former so we don't end up with _4_ possible values.

- Key absent vs. value `undefined`: The API and its consumers needs to be ready to handle different browsers with different sets of feature flags, so there will always be the possibility that a feature flag is defined in the spec but absent from the set of keys. So you could get `{ newFeature: false, hybridTransport: true }` in one browser and `{ hybridTransport: true }` in another browser (or older version of the same browser), even if both browsers follow the recommendation to include keys for all known features.

  So then there are again 4 different possibilities: the key can be present or absent, and the value can be `false`, `true` or `undefined`. Is it useful to know that the browser knows about the existence of a feature identifier, but doesn't wish to disclose whether it's supported? (There doesn't seem to be a JS binding to obtain the list of enumeration values, but `Object.keys()` would let you iterate over the keys present.)

  Actually, I guess that too could be a fingerprinting opportunity and thus a privacy concern. I would err on the side of caution and have keys be absent if the feature is unknown or if the browser doesn't wish to disclose it. This also reduces the number of possible outputs to 3 per feature.

So, how about this? :smile:

```
The set of [=map/keys=] SHOULD equal the set of [=enumeration values=] of {{ClientCapability}},
but the client MAY omit keys as it deems necessary; see [[#sctn-disclosing-client-capabilities]].
```

index.bs
MasterKale: In the name of maintaining the fingerprinting minimization aspects of this API, and because `record` and `maplike` are both ordered types I restored this line. I also alphabetized the `ClientCapability` enums to subconsciously encourage clients to lexicographically sort these values when they support this API 😏

index.bs
MasterKale: Thanks for thinking out loud here, I follow your train of thought. I agree this can turn into a real headache if we try to require defining unknown capability keys as `undefined` instead of allowing those keys to be omitted. I'll add in your MAY addition above.

index.bs
agl: I thought that Javascript had defined the iteration order these days and that we shouldn't override that. Indeed [it has](https://tc39.es/ecma262/#sec-ordinaryownpropertykeys), but the defined order is partially based on the insertion order. So, yep, this should stay.

agl (approved):

emlun (approved):

ve7jtb (approved):
lgtm

lgarron:
I 100% agree with the camelCase convention reasons, but just a technical note:

> ```js
> const { 'user-verifying-platform-authenticator' } = capabilities;
> // SyntaxError: Unexpected token '}'. Expected a ':' prior to a named
> // destructuring property.
> ```

This is actually possible to do using a `:` (as suggested by the error message):

```js
const { 'user-verifying-platform-authenticator': userVerifyingPlatformAuthenticator } = capabilities;
```

However, this is definitely not as ergonomic, more code to ship to the end-user for no good reason, and a less debuggable than if the exact same string is used to refer this the destructured value throughout the JSON and the code.

Issue #1743, opened 2022-06-09
Spec abstract is out of date on the eve of multi-device credentials and cross-device auth


MasterKale:
This sentence in the spec's abstract is now out of date with the upcoming launch of multi-device credentials across all three major platform vendors:

> Conceptually, one or more [public key credentials](https://www.w3.org/TR/webauthn-2/#public-key-credential), each [scoped](https://www.w3.org/TR/webauthn-2/#scope) to a given [WebAuthn Relying Party](https://www.w3.org/TR/webauthn-2/#webauthn-relying-party), are created by and [bound](https://www.w3.org/TR/webauthn-2/#bound-credential) to [authenticators](https://www.w3.org/TR/webauthn-2/#authenticator) as requested by the web application.

This needs to be updated to reflect the reality that A) backup of credentials mean credentials are no longer device-bound, and B) capabilities like Apple's shareable multi-device credentials means credentials are potentially no longer user-bound either.

dagnelies:
I learned about this recently, and I must admit that the concept of sharing private keys in the "cloud" does not sound like the epitome of security to me. If some day in the future this central credentials storage gets leaked/breached/hacked/eavesdropped/whatever, it would be a disaster. I'm pretty sure there are all kind of security measures in place, but it wouldn't be the first time something like this happens.

Until now, the private key being bound to the device gave me a certain sense of security. Not anymore though with "synced-in-the-cloud-secrets" aka multi-device credentials. It makes the whole reasoning also a bit more difficult.

In this light, I think too it is crucial to update the abstract of this specification to highlight that these "private keys" are not device-bound anymore, but can be synced/shared. The concept is fundamentally altered because of this, with implications for usage, security and privacy.

vonDubenshire:
Lot of documentation here will calm you.

> Until now, the private key being bound to the device gave me a certain sense of security. Not anymore though with "synced-in-the-cloud-secrets" aka multi-device credentials. It makes the whole reasoning also a bit more difficult.
> 
> In this light, I think too it is crucial to update the abstract of this specification to highlight that these "private keys" are not device-bound anymore, but can be synced/shared. The concept is fundamentally altered because of this, with implications for usage, security and privacy.

"Passkeys in the Google Password Manager
[...]

PR #1931, opened 2023-07-26
Update Authenticator Definition

nicksteele:
This addresses #1743 by updating the authenticator definition to denote that an authenticator does not have to be wholly local without going into more detail on concepts like an identity fabric.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/pull/1931.html" title="Last updated on Jul 26, 2023, 9:52 PM UTC (aff97c2)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/1931/5bd3dd1...aff97c2.html" title="Last updated on Jul 26, 2023, 9:52 PM UTC (aff97c2)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -952,13 +952,13 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
 : <dfn>[WAA]</dfn>
 :: A cryptographic entity, existing in hardware or software, that can [=registration|register=] a user with a given [=[RP]=]
     and later [=Authentication Assertion|assert possession=] of the registered [=public key credential=], and optionally
-    [=user verification|verify the user=] when requested by the [=[RP]=]. [=Authenticators=] can report information
+    [=user verification|verify the user=] to the [=[RP]=]. [=Authenticators=] can report information
     regarding their [=authenticator types|type=] and security characteristics via [=attestation=] during [=registration=]
     and [=assertion=].
 
     A [=[WAA]=] could be a [=roaming authenticator=], a dedicated hardware subsystem integrated into the [=client device=],
     or a software component of the [=client=] or [=client device=]. A [=[WAA]=] is not necessarily confined to operating in 
-    a local context, and can generate or store a [=credential key pair=] in a server outside of the [=client device=].
+    a local context, and can generate or store a [=credential key pair=] in a server outside of [=client-side=] hardware.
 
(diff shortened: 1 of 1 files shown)


timcappalli:
> a local context, and can generate or store a [=credential key pair=] in a server outside of the [=client device=].

In the case of hybrid, the authenticator is not part of the client device. I think this gets into the authenticator boundary discussion. I don't have a suggestion as to what to put instead of client device, but I don't think it should be "client device".

Anyone else have suggestions?

index.bs
```diff
@@ -952,11 +952,13 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
 : <dfn>[WAA]</dfn>
 :: A cryptographic entity, existing in hardware or software, that can [=registration|register=] a user with a given [=[RP]=]
     and later [=Authentication Assertion|assert possession=] of the registered [=public key credential=], and optionally
-    [=user verification|verify the user=], when requested by the [=[RP]=]. [=Authenticators=] can report information
```
emlun, line 955: Sorry for this extreme nitpicking, but I disagree with deleting this comma. Some authenticators require user verification even when not requested by the RP; this can even be a preference configured by the user in some cases.

emlun:
> In the case of hybrid, the authenticator is not part of the client device. I think this gets into the authenticator boundary discussion. I don't have a suggestion as to what to put instead of client device, but I don't think it should be "client device".
> 
> Anyone else have suggestions?

Perhaps something like `outside of [=client-side=] hardware`?

>[**Client-Side**](https://www.w3.org/TR/2021/REC-webauthn-2-20210408/#client-side)
>This refers in general to the combination of the user’s [client platform](https://www.w3.org/TR/2021/REC-webauthn-2-20210408/#client-platform), [authenticators](https://www.w3.org/TR/2021/REC-webauthn-2-20210408/#authenticator), and everything gluing it all together.

MasterKale (approved):

index.bs
nicksteele: I would say that the better solution would be to replace 
> verify the user, when requested by the RP" 

with 

> " verify the user to the RP"

the current format with the comma makes an already run-on sentence have improper syntax.

emlun: Ok, that too seems fine to me.

emlun (approved):

akshayku (approved):

PR #1932, opened 2023-07-26
Clarify distinction between PublicKeyCredentialUserEntity name and displayName

### LINKED ISSUE #1852: Emphasize use of `user.name` for RP's to help users distinguish credentials
## Context

RP's, like us at Cisco, use WebAuthn with a broad RP ID scoped in a way to allow for credential reuse across multiple "applications" on unique subdomains. Unfortunately, we've discovered that current, nascent passkeys management UIs emphasize displaying a credential's RP ID and username, which are often 

**Chrome:**
Screenshot 2023-02-13 at 6 14 57 PM

**1Password:**
Screenshot 2023-02-13 at 6 23 15 PM

**Safari (macOS):**
Screenshot 2023-02-13 at 6 16 38 PM

## Proposed Change

What if we added a "`note`" property somewhere in [PublicKeyCredentialCreationOptions](https://www.w3.org/TR/webauthn-2/#dictdef-publickeycredentialcreationoptions) to allow RP's to annotate a credential for the management interfaces to then expose to users for the user's ease of management?

Here are examples of places that "`note`" could go:

**1Password ("notes"):**
Screenshot 2023-02-13 at 6 15 44 PM

**Safari ("Notes"):**
Screenshot 2023-02-13 at 6 29 01 PM

Screenshot 2023-02-13 at 6 29 22 PM

This feels beneficial to both RP's and end users.

MasterKale:
This PR attempts to help RP's understand the differences between `user.name` and `user.displayName`. It most importantly establishes `user.name` as the primary value displayed by browsers to users. The definition of `user.displayName` has also been tweaked to emphasize that this value has utility to differentiate "sub-accounts" for a given `user.name`, but without encouraging its use as based on [observations I made back in February](https://github.com/w3c/webauthn/issues/1852#issuecomment-1429062522) only Android currently exposes this value.

Fixes #1852.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/pull/1932.html" title="Last updated on Sep 6, 2023, 5:42 PM UTC (aefe8f2)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/1932/5bd3dd1...aefe8f2.html" title="Last updated on Sep 6, 2023, 5:42 PM UTC (aefe8f2)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2784,7 +2784,7 @@ Note: Invoking this method from a [=browsing context=] where the [=Web Authentic
 <div link-for-hint="WebAuthentication/isPasskeyPlatformAuthenticatorAvailable">
 
 [=[WRPS]=] use this method to determine whether they can create a new [=passkey=] using a [=user-verifying platform authenticator=] or a {{AuthenticatorTransport/hybrid}} authenticator.
-Upon invocation, the [=client=] employs a [=client platform=]-specific procedure to discover available [=user-verifying platform authenticators=] and the 
+Upon invocation, the [=client=] employs a [=client platform=]-specific procedure to discover available [=user-verifying platform authenticators=] and the
 availability of {{AuthenticatorTransport/hybrid}} transport.
 If one or both are discovered, the promise is resolved with the value of [TRUE].
 If neither is discovered, the promise is resolved with the value of [FALSE].
@@ -3186,9 +3186,12 @@ associated with or [=scoped=] to, respectively.
           - When inherited by {{PublicKeyCredentialUserEntity}}, it is a
             [=human palatability|human-palatable=] identifier for a [=user account=]. This
             identifier is the primary value displayed to users by [=Clients=] to help users
-            understand with which [=user account=] a credential is associated. Examples of suitable
-            values for this identifier include, "alexm", "alex.mueller@example.com", or
-            "+14255551234".
+            understand with which [=user account=] a credential is associated. The [=[RP]=] MAY
+            extend the identifier to help a user differentiate between multiple credentials for
+            a single [=user account=] when the same value would otherwise be used for all
+            credentials.
+
+            Examples of suitable values for this identifier include, "alex.mueller@example.com (Production)", "alexm", or "+14255551234".
 
               - The [=[RP]=] MAY let the user choose this value. The [=[RP]=] SHOULD perform enforcement,
                 as prescribed in Section 3.4.3 of [[!RFC8265]] for the UsernameCasePreserved Profile of the PRECIS
@@ -3257,13 +3260,10 @@ credential.
 
     :   <dfn>displayName</dfn>
     ::  A [=human palatability|human-palatable=] name for the [=user account=], intended only for
-        display. [=Clients=] may display this value to help users differentiate between multiple
-        [=credentials=] with the same value of {{PublicKeyCredentialEntity/name}}. Examples of
-        suitable values for this identifier include, "Alex Müller" or "田中倫".
+        display. The [=[RP]=] SHOULD let the user choose this, and SHOULD NOT restrict the choice
+        more than necessary. The [=[RP]=] MAY set this to the same value as {{PublicKeyCredentialEntity/name}}.
 
-        The [=[RP]=] SHOULD let the user choose this, and SHOULD NOT restrict the choice more than necessary.
-
-        The [=[RP]=] MAY set this to the same value as {{PublicKeyCredentialEntity/name}}.
+        Examples of suitable values for this identifier include, "Alex Müller (ACME Co.)" or "田中倫".
 
         - [=[RPS]=] SHOULD perform enforcement, as prescribed in Section 2.3 of
             [[!RFC8266]] for the Nickname Profile of the PRECIS FreeformClass [[!RFC8264]],
--- a/index.bs
+++ b/index.bs
@@ -3258,8 +3258,8 @@ credential.
         [=Clients=] may display this value to help users differentiate between multiple sub-accounts for a
         given [=user account=] for situations in which every credential uses the same value for
         {{PublicKeyCredentialEntity/name}}. When such differentiation is not required, this value
-        MAY be set to the same value as {{PublicKeyCredentialEntity/name}}. Examples of suitable values for this identifier include,
-        "Alex Müller" or "田中倫".
+        MAY be set to the same value as {{PublicKeyCredentialEntity/name}}. Examples of suitable
+        values for this identifier include, "Alex Müller" or "田中倫".
 
         A [=[RP]=] MAY let the user specify this value, and SHOULD NOT restrict the choice
         more than necessary when doing so.
--- a/index.bs
+++ b/index.bs
@@ -3254,8 +3254,12 @@ credential.
         with more than one [=user account=] at the [=[RP]=].
 
     :   <dfn>displayName</dfn>
-    ::  A [=human palatability|human-palatable=] name for the [=user account=], intended only for display. For example, "Alex Müller" or "田中倫". The
-        [=[RP]=] SHOULD let the user choose this, and SHOULD NOT restrict the choice more than necessary.
+    ::  A [=human palatability|human-palatable=] name for the [=user account=], intended only for display.
+        [=Clients=] may display this value to help users differentiate between multiple sub-accounts for a
+        given [=user account=] for situations in which every credential uses the same value for
+        {{PublicKeyCredentialEntity/name}}. Examples of suitable values for this identifier include,
+        "Alex Müller" or "田中倫". The [=[RP]=] SHOULD let the user choose this, and SHOULD NOT restrict
+        the choice more than necessary.
 
         - [=[RPS]=] SHOULD perform enforcement, as prescribed in Section 2.3 of
             [[!RFC8266]] for the Nickname Profile of the PRECIS FreeformClass [[!RFC8264]],
--- a/index.bs
+++ b/index.bs
@@ -3183,10 +3183,12 @@ associated with or [=scoped=] to, respectively.
                 on {{PublicKeyCredentialEntity/name}}'s value prior to displaying the value to the user or
                 including the value as a parameter of the [=authenticatorMakeCredential=] operation.
 
-          - When inherited by {{PublicKeyCredentialUserEntity}}, it is a [=human palatability|human-palatable=] identifier for a
-            [=user account=]. This identifier is the primary value displayed to users by [=Clients=] to help users understand with which
-            [=user account=] a credential is associated. Examples of suitable values for this identifier include, "alexm", "alex.mueller@example.com"
-            or "+14255551234".
+          - When inherited by {{PublicKeyCredentialUserEntity}}, it is a
+            [=human palatability|human-palatable=] identifier for a [=user account=]. This
+            identifier is the primary value displayed to users by [=Clients=] to help users
+            understand with which [=user account=] a credential is associated. Examples of suitable
+            values for this identifier include, "alexm", "alex.mueller@example.com", or
+            "+14255551234".
 
               - The [=[RP]=] MAY let the user choose this value. The [=[RP]=] SHOULD perform enforcement,
                 as prescribed in Section 3.4.3 of [[!RFC8265]] for the UsernameCasePreserved Profile of the PRECIS
--- a/index.bs
+++ b/index.bs
@@ -3186,12 +3186,11 @@ associated with or [=scoped=] to, respectively.
           - When inherited by {{PublicKeyCredentialUserEntity}}, it is a
             [=human palatability|human-palatable=] identifier for a [=user account=]. This
             identifier is the primary value displayed to users by [=Clients=] to help users
-            understand with which [=user account=] a credential is associated. The [=[RP]=] MAY
-            extend the identifier to help a user differentiate between multiple credentials for
-            a single [=user account=] when the same value would otherwise be used for all
-            credentials.
+            understand with which [=user account=] a credential is associated.
 
-            Examples of suitable values for this identifier include, "alex.mueller@example.com (Production)", "alexm", or "+14255551234".
+            Examples of suitable values for this identifier include, "alexm", "+14255551234",
+            "alex.mueller@example.com", "alex.mueller@example.com (prod-env)",
+            "alex.mueller@example.com (ACME Corp.)", or "alex.mueller@example.com (ОАО Примертех)".
 
               - The [=[RP]=] MAY let the user choose this value. The [=[RP]=] SHOULD perform enforcement,
                 as prescribed in Section 3.4.3 of [[!RFC8265]] for the UsernameCasePreserved Profile of the PRECIS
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -3254,8 +3254,12 @@ credential.
...
 
     :   <dfn>displayName</dfn>
-    ::  A [=human palatability|human-palatable=] name for the [=user account=], intended only for display. For example, "Alex Müller" or "田中倫". The
-        [=[RP]=] SHOULD let the user choose this, and SHOULD NOT restrict the choice more than necessary.
+    ::  A [=human palatability|human-palatable=] name for the [=user account=], intended only for display.
+        [=Clients=] may display this value to help users differentiate between multiple sub-accounts for a
+        given [=user account=] for situations in which every credential uses the same value for
+        {{PublicKeyCredentialEntity/name}}. Examples of suitable values for this identifier include,
+        "Alex Müller" or "田中倫". The [=[RP]=] SHOULD let the user choose this, and SHOULD NOT restrict
+        the choice more than necessary.
```
sbweeden, line 3262: In prettty much every case I've ever seen, this gets set to the same value as user.name. Is it worth mentioning that is a perfectly acceptable thing to do?

emlun (approved):
LGTM, thanks!

index.bs
MasterKale: @sbweeden I tried updating the text accordingly, let me know what you think. @emlun I'm curious to hear if the change I made today requires refinement, or if your current approval still stands.

nicksteele (approved):

index.bs
```diff
@@ -3254,8 +3254,15 @@ credential.
...
-        [=[RP]=] SHOULD let the user choose this, and SHOULD NOT restrict the choice more than necessary.
+    ::  A [=human palatability|human-palatable=] name for the [=user account=], intended only for display.
+        [=Clients=] may display this value to help users differentiate between multiple sub-accounts for a
+        given [=user account=] for situations in which every credential uses the same value for
+        {{PublicKeyCredentialEntity/name}}. When such differentiation is not required, this value
+        MAY be set to the same value as {{PublicKeyCredentialEntity/name}}. Examples of suitable
+        values for this identifier include, "Alex Müller" or "田中倫".
+
+        A [=[RP]=] MAY let the user specify this value, and SHOULD NOT restrict the choice
+        more than necessary when doing so.
```
emlun, line 3265: Continuing from https://github.com/w3c/webauthn/pull/1932#discussion_r1281092796 and [commit `f09e3d8c` comments](https://github.com/w3c/webauthn/commit/f09e3d8cea148765c5d634b283e9cc81f583ee6e#commitcomment-123354509):

Hm. Relaxing the last sentence from SHOULD to MAY is strictly speaking a normative change, though a backwards compatible one. It probably doesn't matter either way, but I don't think it needs to change either.

I think it reads marginally better with "The [=[RP]=]", if nothing else for consistency with the description of `name` in the previous section. I think "...when doing so" is a fine addition if the requirement is relaxed from SHOULD to MAY.

I'm not sure the idea of "sub-accounts" is necessary, but I wouldn't say it hurts either.

All in all I think this is still fine to merge as is, but since you asked @MasterKale :slightly_smiling_face: here's how I would phrase these changes as a whole:

```suggestion
    ::  A [=human palatability|human-palatable=] name for the [=user account=], intended only for display. For example, "Alex Müller" or "田中倫".
        [=Clients=] may display this value to help users differentiate between multiple [=credentials=] with the same value of {{PublicKeyCredentialEntity/name}}.

        The [=[RP]=] SHOULD let the user choose this, and SHOULD NOT restrict the choice more than necessary.
        The [=[RP]=] MAY set this to the same value as {{PublicKeyCredentialEntity/name}}.
```

Granted it might look a bit weird to have a SHOULD immediately followed by a conflicting MAY, but I think this should communicate that the latter is fine but the former is preferable when practical.

index.bs
emlun: See reply in https://github.com/w3c/webauthn/pull/1932#discussion_r1281744220

index.bs
emlun: Again, to be clear: I think the PR is fine without the above suggestion, but feel free to use some, all or none of it as you like @MasterKale. :slightly_smiling_face:

MasterKale: > ...but I think this should communicate that the latter is fine but the former is preferable when practical.

Ah, I see the subtle difference. Your suggested wording sounds more spec-like, and you captured the ultimate meaning I want to achieve with this effort. I made some changes accordingly.

emlun: Ok, looks good!

emlun (approved):

index.bs
MasterKale: @sbweeden I'd like to mark this resolve. Can you please review, and if everything looks good to you approve?

sbweeden: lgtm - resolving

MasterKale: Thanks for taking another look. Can you "Approve" the PR too when you get a chance?

akshayku (approved):

index.bs
```diff
@@ -3183,10 +3183,12 @@ associated with or [=scoped=] to, respectively.
...
-          - When inherited by {{PublicKeyCredentialUserEntity}}, it is a [=human palatability|human-palatable=] identifier for a
-            [=user account=]. It is intended only for display, i.e., aiding the user in determining the difference between user
-            accounts with similar {{PublicKeyCredentialUserEntity/displayName}}s. For example, "alexm", "alex.mueller@example.com"
-            or "+14255551234".
+          - When inherited by {{PublicKeyCredentialUserEntity}}, it is a
+            [=human palatability|human-palatable=] identifier for a [=user account=]. This
+            identifier is the primary value displayed to users by [=Clients=] to help users
+            understand with which [=user account=] a credential is associated. Examples of suitable
+            values for this identifier include, "alexm", "alex.mueller@example.com", or
+            "+14255551234".
```
MasterKale, line 3191: This section needs to be tweaked to more strongly emphasize that this is **the** value for users to distinguish multiple credentials that may have the same email address (thinking about multi-tenant environments.)

I also want to add an example of something like "alex.mueller@example.com (production)" to open up perception to possible ways of differentiating multi-tenant credentials.

> [=Clients=] may display this value to help users differentiate between multiple
        [=credentials=] with the same value of {{PublicKeyCredentialEntity/name}}.

I should also remove this from the section **below**.

MasterKale: Example from @timcappalli:

```
dn: Tim Cappalli (Microsoft)
n: tim.cappalli@microsoft.com

is a common pattern I see
```

index.bs
```diff
@@ -3254,8 +3256,14 @@ credential.
         with more than one [=user account=] at the [=[RP]=].
 
     :   <dfn>displayName</dfn>
-    ::  A [=human palatability|human-palatable=] name for the [=user account=], intended only for display. For example, "Alex Müller" or "田中倫". The
-        [=[RP]=] SHOULD let the user choose this, and SHOULD NOT restrict the choice more than necessary.
+    ::  A [=human palatability|human-palatable=] name for the [=user account=], intended only for
+        display. [=Clients=] may display this value to help users differentiate between multiple
+        [=credentials=] with the same value of {{PublicKeyCredentialEntity/name}}. Examples of
```
akshayku, line 3261: Remove this line because name is the unique identifier typically.

index.bs
MasterKale: I took another stab at the wording here to try and emphasize that RP's should add to `name` to differentiate multiple credentials when otherwise all of them would end up with the same value.

index.bs
MasterKale: I believe I've addressed this, if you can take another look @akshayku I'd appreciate it.

index.bs
```diff
@@ -3183,10 +3183,15 @@ associated with or [=scoped=] to, respectively.
...
-            [=user account=]. It is intended only for display, i.e., aiding the user in determining the difference between user
-            accounts with similar {{PublicKeyCredentialUserEntity/displayName}}s. For example, "alexm", "alex.mueller@example.com"
-            or "+14255551234".
+          - When inherited by {{PublicKeyCredentialUserEntity}}, it is a
+            [=human palatability|human-palatable=] identifier for a [=user account=]. This
+            identifier is the primary value displayed to users by [=Clients=] to help users
+            understand with which [=user account=] a credential is associated. The [=[RP]=] MAY
+            extend the identifier to help a user differentiate between multiple credentials for
+            a single [=user account=] when the same value would otherwise be used for all
+            credentials.
```
emlun, line 3192: Sorry, I think this still misses the mark a bit. One authenticator should almost never have two credentials for the same account, but may have credentials for different accounts.

Perhaps we're getting hung up on what an "account" is. You've mentioned that what you're trying to reflect here is a "multi-tenant" use case, where you have one user login account that contains multiple distinct domains of privileges, and uses different WebAuthn credentials to access each domain, but all under the same "umbrella" account. For example, a management console where one user manages several distinct application domains. I think the idea was that all of the user's domain-specific WebAuthn credentials would have the same `user.name`, for example the e-mail address used as a username to log in to the single "umbrella" account that manages all of those domains. Correct?

But within the context of WebAuthn, the term "[user account](https://w3c.github.io/webauthn/#user-account)" is defined as:

>a [user account](https://w3c.github.io/webauthn/#user-account) denotes the mapping of a set of [credentials](https://w3c.github.io/webappsec-credential-management/#concept-credential) [[CREDENTIAL-MANAGEMENT-1]](https://w3c.github.io/webauthn/#biblio-credential-management-1) to a (sub)set of a [Relying Party](https://w3c.github.io/webauthn/#relying-party)'s resources, as maintained and authorized by the [Relying Party](https://w3c.github.io/webauthn/#relying-party).

A WebAuthn credential doesn't need to be tied to an e-mail address, for example. It's perfectly fine for the "multi-tenant" application to associate a WebAuthn credential with a particular privilege domain instead of the "umbrella" account. The distinction here is that each of those privilege domains is a separate "user account" by the WebAuthn definition - they are not "sub-accounts" of the "umbrella" account, because the "umbrella" account is not a WebAuthn "user account" if no WebAuthn credentials are associated with it.

Does that make sense?

emlun (changes requested):

index.bs
emlun: If so, perhaps it could suffice to just extend the examples?

```
          - When inherited by {{PublicKeyCredentialUserEntity}}, it is a
            [=human palatability|human-palatable=] identifier for a [=user account=]. This
            identifier is the primary value displayed to users by [=Clients=] to help users
            understand with which [=user account=] a credential is associated.

            Examples of suitable values for this identifier include, "alexm", "+14255551234",
            "alex.mueller@example.com", "alex.mueller@example.com (ACME Corporation)",
            or "alex.mueller@example.com (ОАО Примертех)".

```

index.bs
```diff
@@ -3254,8 +3259,11 @@ credential.
         with more than one [=user account=] at the [=[RP]=].
 
     :   <dfn>displayName</dfn>
-    ::  A [=human palatability|human-palatable=] name for the [=user account=], intended only for display. For example, "Alex Müller" or "田中倫". The
-        [=[RP]=] SHOULD let the user choose this, and SHOULD NOT restrict the choice more than necessary.
+    ::  A [=human palatability|human-palatable=] name for the [=user account=], intended only for
+        display. The [=[RP]=] SHOULD let the user choose this, and SHOULD NOT restrict the choice
+        more than necessary. The [=[RP]=] MAY set this to the same value as {{PublicKeyCredentialEntity/name}}.
```
agl, line 3264: "The [=[RP]=] MAY set this to the same value as {{PublicKeyCredentialEntity/name}}."

I think we should suggest setting the empty string when no suitable value is available, rather than duplicating.

MasterKale: I've updated the wording here to suggest it's okay to set this to an empty string if needed.

index.bs
MasterKale: > You've mentioned that what you're trying to reflect here is a "multi-tenant" use case, where you have one user login account that contains multiple distinct domains of privileges, and uses different WebAuthn credentials to access each domain, but all under the same "umbrella" account. For example, a management console where one user manages several distinct application domains.

This exactly. For most RP's `user.name` will be as simple as an email address or username. But for a cloud hosting PaaS, Alex could be the admin of multiple customers' VPS' so the PaaS RP should append some unique value per VPS to his PaaS username when they allow Alex to register authenticators for logging into those admin panels using WebAuthn. 

I've incorporated your suggestions and tweaked the examples accordingly. Let me know what you think.

index.bs
MasterKale: @agl Can you please take a look at this latest change and see if it satisfies your concern?

index.bs
MasterKale: @emlun Can you please take a look at this latest change and see if it satisfies your concern?

index.bs
```diff
@@ -3183,10 +3183,14 @@ associated with or [=scoped=] to, respectively.
...
-            accounts with similar {{PublicKeyCredentialUserEntity/displayName}}s. For example, "alexm", "alex.mueller@example.com"
-            or "+14255551234".
+          - When inherited by {{PublicKeyCredentialUserEntity}}, it is a
+            [=human palatability|human-palatable=] identifier for a [=user account=]. This
+            identifier is the primary value displayed to users by [=Clients=] to help users
+            understand with which [=user account=] a credential is associated.
+
+            Examples of suitable values for this identifier include, "alexm", "+14255551234",
+            "alex.mueller@example.com", "alex.mueller@example.com (prod-env)",
+            "alex.mueller@example.com (ACME Corp.)", or "alex.mueller@example.com (ОАО Примертех)".
```
emlun, line 3193: I think with the addition of the (prod-env) example, we can drop the (ACME Corp.) one.

```suggestion
            "alex.mueller@example.com", "alex.mueller@example.com (prod-env)",
            or "alex.mueller@example.com (ОАО Примертех)".
```

index.bs
```diff
@@ -3254,8 +3258,12 @@ credential.
...
 
     :   <dfn>displayName</dfn>
-    ::  A [=human palatability|human-palatable=] name for the [=user account=], intended only for display. For example, "Alex Müller" or "田中倫". The
-        [=[RP]=] SHOULD let the user choose this, and SHOULD NOT restrict the choice more than necessary.
+    ::  A [=human palatability|human-palatable=] name for the [=user account=], intended only for
+        display. The [=[RP]=] SHOULD let the user choose this, and SHOULD NOT restrict the choice
+        more than necessary. If no suitable or [=human palatability|human-palatable=] name is
+        available, the [=[RP]=] SHOULD set this value to an empty string.
+
+        Examples of suitable values for this identifier include, "Alex Müller (ACME Co.)" or "田中倫".
```
emlun, line 3266: I think these examples should still include the one without (ACME) for contrast.

```suggestion
        Examples of suitable values for this identifier include, "Alex Müller", "Alex Müller (ACME Co.)" or "田中倫".
```

emlun (approved):
Looks good to me. :slightly_smiling_face: The attached suggestions are optional.

index.bs
MasterKale: Changes made ✌️

index.bs
MasterKale: Changes made ✌️

PR #1945, opened 2023-08-23
Add importCryptoKey input to PRF extension

### LINKED ISSUE #1895: Proposal/discussion: non-extractable CryptoKey output from the prf extension
# Background
The `prf` extension creates a path for web applications to support end-to-end encryption via hardware authenticators in the browser, in conjunction with [WebCrypto](https://github.com/w3c/webcrypto). [Encrypting Data in the Browser Using WebAuthn](https://blog.millerti.me/2023/01/22/encrypting-data-in-the-browser-using-webauthn/) presents some sample code for how this can be achieved in Chrome Canary today.

From reading discussion around prf, E2EE appears to be one of the core use cases, e.g. for password managers or any applications written with a focus on user privacy. (In my case, I'm developing a PWA that does all user data handling in the browser with encrypted backups.)

## Problem
When crossing the webauthn <> webcrypto api boundary, we have to expose the key material to the js runtime. From [step 2.1](https://blog.millerti.me/2023/01/22/encrypting-data-in-the-browser-using-webauthn/#step-21-import-the-input-key-material) of the article:
```ts
const inputKeyMaterial = new Uint8Array(auth1ExtensionResults.prf.results.first);

const keyDerivationKey = await crypto.subtle.importKey(
  "raw",
  inputKeyMaterial,
  "HKDF",
  false, // n.b. Extractable flag from here on is false, but the raw key material is in memory already...
  ["deriveKey"],
);
```

For this use case, the prf output is ideally never exposed to the js context.

## Proposal
I propose adding an option to the prf extension that supports returning an non-extractable [`CryptoKey`](https://developer.mozilla.org/en-US/docs/Web/API/CryptoKey):
```ts
extensions: {
  prf: {
    eval: {
      first: firstSalt,
      asCryptoKey: true, // New optional field on `AuthenticationExtensionsPRFValues`
    },
  },
},
```

If `asCryptoKey` is set true, then the result object returns a non-extractable `CryptoKey` instead of a `BufferSource`. The resultant key should match the output of the `importKey` call above (i.e. using HKDF).

### Why
To get concrete about why this change can help, we can think about some attack vectors. Note that "key material" below refers to the prf extension results:
- Attacker with js code execution (i.e. from XSS) on the page
  - If an attacker can do code execution in your js context, it's game over already (i.e. attacker could also just exfiltrate data later on, etc). Protecting the key material can limit the blast radius of exposure (i.e. decrypting future/past content offline)
- Attacker with malicious browser extension code execution
  - Again, attacker can't get key material from js, but that may be moot if the user's machine is likely compromised.
- Buggy code
  - Marking key material unexportable can protect from hidden or obfuscated exposure, e.g. if the developer accidentally logs keys somewhere

### Generalized crypto operations
I've seen the [discussion for general crypto operations](https://github.com/w3c/webauthn/issues/1608) which seems relevant but much broader and nebulous. I'm proposing a smaller, concrete change to the prf extension spec.

### The scope of webauthn
I've seen the previous discussion (#1481) on the scope of the webauthn spec and understand the desire to stay out of webcrypto's territory. At the same time, I do also see the [desire to make correct, secure implementations](https://github.com/w3c/webauthn/issues/1595#issuecomment-817217895) easy to do within the spec. It seems like the only way to keep key material out of the js context is to have the webauthn/the prf extension do the webcrypto call behind the scenes.

### Allow arbitrary CryptoKey derivations.
Instead of only allowing HKDF, we could allow passing arbitrary options for use as e.g. RSA or ECDH. This feels out of scope and ties the webauthn api too strongly to the webcrypto one.

Notably, I believe that only outputting HKDF doesn't limit what kinds of constructions folks could implement - they could call `deriveKey` again from this key, etc.

### Don't use prf for this use case
Perhaps this is simply a mismatch in use case, but then what is prf for? I haven't seen much discussion around other use cases. A better fit may be to expose hardware keys in webcrypto and actually decrypt/encrypt bytes on the hardware itself so that keys never leave the HSM, though it's unclear to me if those operations are reasonably performant...

### Use the api as-is and do your best to secure the js context
Seems okay? But not as satisfying.

emlun:
Fixes #1895.

This draft allows as much flexibility in the `importKey` invocation as possible. It might be a better idea to hard-code this key import to use for example HKDF instead; I'll open a separate meta-PR for that.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/pull/1945.html" title="Last updated on Aug 23, 2023, 2:10 PM UTC (65b635b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/1945/bd68fbf...65b635b.html" title="Last updated on Aug 23, 2023, 2:10 PM UTC (65b635b)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -6754,6 +6754,14 @@ Note: this extension may be implemented for [=authenticators=] that do not use [
     dictionary AuthenticationExtensionsPRFInputs {
         AuthenticationExtensionsPRFValues eval;
         record<USVString, AuthenticationExtensionsPRFValues> evalByCredential;
+        AuthenticationExtensionsPRFImportCryptoKeyParams importCryptoKey;
+    };
+
+    dictionary AuthenticationExtensionsPRFImportCryptoKeyParams {
+        required KeyFormat format;
+        required AlgorithmIdentifier algorithm;
+        required boolean extractable;
+        required sequence<KeyUsage> keyUsages;
     };
 
     partial dictionary AuthenticationExtensionsClientInputs {
@@ -6767,6 +6775,19 @@ Note: this extension may be implemented for [=authenticators=] that do not use [
 
         :   <dfn>evalByCredential</dfn>
         ::  A record mapping [=base64url encoding|base64url encoded=] [=credential IDs=] to PRF inputs to evaluate for that credential. Only applicable during [=assertions=] when {{PublicKeyCredentialRequestOptions/allowCredentials}} is not empty.
+
+        :   <dfn>importCryptoKey</dfn>
+        ::  Arguments for the [=client=] to invoke {{SubtleCrypto/importKey}}, along with the PRF result as the `keyData` argument.
+            If present, the extension outputs will be {{Promise}}s resolving to {{CryptoKey}} values;
+            if not present, the extension outputs will be {{BufferSource}} values.
+
+            The client ensures domain separation between {{BufferSource}} and {{CryptoKey}} results,
+            so that an extension input requesting unextractable {{CryptoKey}} values cannot be "downgraded"
+            to requesting the same results as extractable {{BufferSource}} values.
+            For the same reason, the client also ensures domain separation between the [TRUE] and [FALSE] values
+            of <code>{{importCryptoKey}}.{{AuthenticationExtensionsPRFImportCryptoKeyParams/extractable}}</code>.
+
+            Note: The {{KeyFormat}}, {{AlgorithmIdentifier}} and {{KeyUsage}} types are defined in [[!WebCryptoApi]].
     </div>
 
 : Client extension processing ([=registration extension|registration=])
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -6767,17 +6775,56 @@ Note: this extension may be implemented for [=authenticators=] that do not use [
...
+
+        :   <dfn>importCryptoKey</dfn>
+        ::  Arguments for the [=client=] to invoke {{SubtleCrypto/importKey}}, along with the PRF result as the `keyData` argument.
+            If present, the extension outputs will be {{Promise}}s resolving to {{CryptoKey}} values;
+            if not present, the extension outputs will be {{BufferSource}} values.
+
+            The client ensures domain separation between {{BufferSource}} and {{CryptoKey}} results,
+            so that an extension input requesting unextractable {{CryptoKey}} values cannot be "downgraded"
+            to requesting the same results as extractable {{BufferSource}} values.
+            For the same reason, the client also ensures domain separation between the [TRUE] and [FALSE] values
```
agl, line 6787: I suspect the domain separation should cover all the parameters? Here's a silly example, but perhaps there are better ones:

A service uses this facility to get an opaque AES-GCM key and claims that, even if the Javascript is compromised, the compromise is limited to existing ciphertexts and ciphertexts that are encrypted in the future, after the compromise is eliminated, are safe again because the key couldn't be extracted.

But if they use a counter for a nonce, which is safe in general, an attacker could tweak the import parameters to get an opaque AES-ECB key and precompute E<sub>k</sub>(x) for the _x_ that will be used by future AES-GCM ciphertexts.

index.bs
```diff
@@ -6754,6 +6754,14 @@ Note: this extension may be implemented for [=authenticators=] that do not use [
     dictionary AuthenticationExtensionsPRFInputs {
         AuthenticationExtensionsPRFValues eval;
         record<USVString, AuthenticationExtensionsPRFValues> evalByCredential;
+        AuthenticationExtensionsPRFImportCryptoKeyParams importCryptoKey;
+    };
+
+    dictionary AuthenticationExtensionsPRFImportCryptoKeyParams {
+        required KeyFormat format;
```
agl, line 6761: Nothing but [Raw](https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/importKey#raw) can work as a format, right?

index.bs
```diff
@@ -6754,6 +6754,14 @@ Note: this extension may be implemented for [=authenticators=] that do not use [
...
+        required AlgorithmIdentifier algorithm;
```
agl, line 6762: Do we want to subset algorithms?

HMAC is useless, right? Because nobody else could have the HMAC key. (And there's HKDF if someone wants a PRF.)

RSA is out.

PBKDF2 doesn't make sense?

AES-* work.

ECDSA and ECDH? We could define things to work with P-256 if we wanted. Are asymmetric algorithms interesting? The ability to sign things as an identity without throwing UI? The ability to public-key decrypt? I guess this would imply bringing in [CryptoKeyPair](https://developer.mozilla.org/en-US/docs/Web/API/CryptoKeyPair).

index.bs
emlun: Sounds fair. Note that WebCrypto doesn't support AES-ECB, but I suppose you could probably emulate it by feeding single blocks at a time into AES-CBC.

I suppose for that we would need something like a hash of a canonical representation of the arguments - one invariant in order of `EcKeyImportParams` members, for example, or maybe even invariant in `"AES-GCM"` vs. `{name: "AES-GCM"}`.

index.bs
emlun: Right, at least not without preprocessing by the client.

For example, I did one experiment where my application code wrapped the 32 raw bytes in a PKCS#8 header and fed it into `importKey` as a P-256 private key. It "works" (except when it doesn't because of the key domain mismatch, see other thread below), but it's quite an ugly hack.

index.bs
emlun: > Do we want to subset algorithms?

Maybe not? See: #1946

> HMAC is useless, right? Because nobody else could have the HMAC key. (And there's HKDF if someone wants a PRF.)

Mostly yes, unless you import it as extractable and use `wrapKey` to move it somewhere else (or `exportKey` of course, but that defeats the point of all this).

> PBKDF2 doesn't make sense?

Agreed.

> Are asymmetric algorithms interesting? The ability to sign things as an identity without throwing UI? The ability to public-key decrypt?

Signing is a bit awkward due to the lack of UI as you note, but asymmetric encryption certainly is interesting. For example, it would allow for encrypting to multiple PRF-backed keys for redundancy. This can be done today via key wrapping, but [not with never-extractable keys](https://github.com/w3c/webauthn/issues/1895#issuecomment-1691798283).

> ECDSA and ECDH? We could define things to work with P-256 if we wanted.

Right, we could. We can't naively use the PRF directly as a private key, because the P-256 key space is slightly smaller than the full 2<sup>32</sup>-1 range, so we'd need to specify some way to either fail out or recalculate some input salt in case the key falls outside the valid range. I would prefer if we could do something like that generically in WebCrypto, for example extend HKDF with asymmetric keys as output keys, but I presume it would take far more effort to get something like that accepted and rolled out in implementations.

The more I've thought about this, though, the more I've come to think that non-extractability in WebCrypto perhaps isn't that impressive after all. A malicious actor that has the ability to access `CryptoKey` objects probably also has the ability to overwrite properties on `window.crypto.subtle` - and therefore also has the ability to intercept any WebCrypto invocation, set `extractable` to `true` and exfiltrate the key. Even without that, just access to a `CryptoKey` object is enough to use that key for signing and decryption operations - you don't really need to steal the keys if you can just steal what they were meant to protect.

I still think there is value in the ability to import PRF outputs directly to `CryptoKey` objects, as that still reduces an honest application's exposure to raw key material. But I'm not sure that non-extractability is really that important - and if it's not, we might as well hard-code this to output HKDF keys and let application developers go from there. (WebCrypto HKDF cannot derive asymmetric keys, but you can generate asymmetric keys and wrap them with HKDF-derived keys.)

On the other hand, I guess this line of reasoning also comes back around in favour of importing PRF outputs directly to asymmetric private keys - since the domain separation here actually can make truly non-extractable keys. It's still true that malicious code could abuse the resulting `CryptoKey` object, but a truly non-extractable ECDH key is a powerful thing to have access to. Again in the multi-recipient encryption example, you could use ECDH to derive symmetric encryption keys - those `CryptoKey`s would be extractable, but they would exist only alongside the cleartext anyway. And with ECDH you could rotate the encryption keys after every decryption, so that any leaked key won't be able to access any future versions of the contents.

So okay, do you think it would make sense to specify a procedure for the client to transform the PRF output into an asymmetric key? I'm thinking in that case we should probably base it on HKDF or the like, rather than restricting ourselves to key sizes around 32 bits only.

agl:
(Deriving EC private keys from a seed: https://boringssl.googlesource.com/boringssl/+/refs/heads/master/crypto/ec_extra/ec_derive.c#28)

emlun:
Summary of discussion on 2023-09-06 WG call:

There needs to be a point to doing this, and the unique value proposition we can offer here is that the platform can guarantee that keys imported this way are truly unextractable. The most promising use case we imagine is asymmetric encryption: using PRF to derive a truly unextractable private key which can be used for encryption (i.e., RSA) or to derive symmetric encryption keys (i.e., ECDH). Even though any derived or wrapped keys will be extractable, the encryption use case mitigates that issue since the keys will be present only when the cleartext is also, and the unextractable asymmetric key allows for rotating the symmetric keys at frequent intervals to further mitigate the impact of a leaked symmetric key.

As such, the #1946 alternative of hard-coding this to import an HKDF key would be rather pointless, since any keys derived from that would be extractable without enabling the use cases enabled by an asymmetric key. We will instead move this design further toward having the client do more in its controlled environment to enable truly unexportable asymmetric keys.

MasterKale:
**WG meeting @ TPAC 9/12:** We've spent a lot of time trying to figure out what direction to go with this. However we've not yet identified sufficient interest in a discrete use case (e.g. provider-centric need for prf output) to say that this "swiss army knife" PR (vs the more limited #1946 that got closed) should be pursued any further.

Leaving this open for now, but unless a specific use case for `prf` can be identified that cannot be served by this extension in its current form then we should probably close it at the next meeting.

emlun:
2023-09-20 WG call: @emlun to write an explainer of a concrete use case to inform the design and the go/no-go decision.

emlun:
To summarize some of the discussion above, here's the rationale for which types of keys would be relevant to support:

- HMAC: no, because noone else would have the key.
- HKDF: no, because derived keys would still be extractable.
- PBKDF2: not applicable.
- AES: no, because encryption keys do not need to be protected from parties handling the cleartext (assuming each key is used for only one message).
- RSA (encrypt): yes, ignoring key size limitations.
- ECDH: yes, see below.
- RSA (sign): maybe - less useful than RSA (encrypt).
- ECDSA: maybe - less useful than ECDH.

As a concrete use case, I've recently been involved in a pilot project prototyping a web-based wallet app that uses WebAuthn PRF to derive encryption keys used to encrypt wallet contents. The current encryption architecture used in the pilot is described [here](https://open.gunet.gr/wallet-docs/docs/wallet/encryption-architecture). Aside from the self-critique already covered there, this app could benefit from having access to unextractable ECDH private keys.

It is possible with current PRF and WebCrypto capabilities to use RSA or ECDH for asymmetric encryption, so that one can have multiple PRF keys without having all of them present whenever encrypted content changes. But the drawback is that the asymmetric private keys are not hardware bound, and thus could be leaked if there's malicious script on the page. Leaking the asymmetric private keys is worse than leaking the encryption keys derived from them - whenever the encryption keys are in memory, the decrypted cleartext is too, but the encryption keys can be rotated frequently to protect future versions of the encrypted contents. Thus a leaked encryption key would give access only to one version of the encrypted data, which was already decrypted in memory at the time - but a leaked asymmetric private key would give access to all future versions as well.

A similar case could be made for RSA and ECDSA signing keys to sign arbitrary data, but this would not be useful for applications that need single-use signing keys for privacy. I plan to bring that use case up separately.

So, with this in mind, I think the most useful key types to support are ECDH and possibly RSA-OAEP. For that, I imagine the API could look something like this:

```
extensions: {
  prf: {
    eval: {
      first: new Uint8Array([1, 2, 3, 4]),
    },
    deriveCryptoKey:
      // EcKeyGenParams
      { name: "ECDH", namedCurve: "P-256" },

      // OR RsaHashedKeyGenParams
      { name: "RSA-OEAP", modulusLength: 2048, publicExponent: new Uint8Array([1, 0, 1]), hash: "SHA-256" },
  }
}
```

where the client would then use the PRF output as a seed for deriving the private key by some deterministic procedure, for example that described in [NIST SP-800-56Ar3](https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-56Ar3.pdf) section 5.6.1.2.1 ([ECC] Key Pair Generation Using Extra Random Bits) (which seems similar to the [Montgomery reduction in BoringSSL linked above](https://github.com/w3c/webauthn/pull/1945#issuecomment-1709009190)?).

It may be that it's not appropriate to derive RSA keys from only 32 bytes of entropy, in which case ECDH would be the only valid key type.

emlun:
On the other hand one could argue that this still misses the mark - it's still the client handling the private keys, so they are still not hardware-bound. Perhaps we should instead do this with a new extension that enables actually hardware-bound private keys.

But again in favour of doing this in PRF: this would be compatible with existing authenticators that support the CTAP2 `hmac-secret` extension; a new extension would not be compatible with existing authenticators.

Firstyear:
> On the other hand one could argue that this still misses the mark - it's still the client handling the private keys, so they are still not hardware-bound. Perhaps we should instead do this with a new extension that enables actually hardware-bound private keys.

Yes, but we don't have a TPM here. I think if someone wants hardware bound keys, they need to use a proper HSM. We aren't false advertising at all about what this will do. 

Where this is useful is *short lived* keys. Where the key is used and destroyed quickly. For example, a password manager which quickly gets the secret and does the decrypt of the secret and exposes it in memory, then the secret is used and destroyed. Or the ability to sign an email, or something else. 

I think this still has a use and I think it has a place. It's not missing a mark at all. We are allowing short-lived, derived keys that exist in memory, and there is a "chain" that to hold the key, you *very likely* have the hardware authenticator with you, but it's not a strict promise the key is *bound* to the authenticator.

index.bs
fwadnjar, line 1: prf-cryptokey

fwadnjar: prf-cryptokey

fwadnjar (commented):
prf-cryptokey

emlun:
We're pursuing these use cases in other ways which, unlike this, would enable actually hardware-bound keys that do not leave the authenticator.

emlun:
PR #2078 is the "other ways" referred to in my previous comment.

cr6750903177Aguinaldok4:
#2308 diff,patch-emlun.contribute/comment.patch.jul.13 2025;Ext.follows.dr.js

Issue #1862, opened 2023-03-15
WebAuthn Autofill (Conditional UI) for credential registration


nov:
## Description

We're starting WebAuthn support, heavily relying on WebAuthn Autofill UX.
However, WebAuthn Autofill is only available for authentication, not registration.

We'd be happy if we can let password managers register WebAuthn credential in a way similar to what password manager registers strong password via Autofill UX.

## Related Links

emlun:
What would be shown in the autofill UI in this case?

nov:
"iPhone created a strong passkey for this site" or something?

![image](https://user-images.githubusercontent.com/27299/225332524-5a1b9f60-036b-4d50-9610-d75253fa01bf.jpeg)

serianox:
"strong passkey" implies "weak passkey".

Though I'm pretty sure some will have an ECDSA keygen failure sooner or later. 😅

My opinion would be to have a "Use Passkey" button, and have a dimmed "Use Strong Password" and "Choose My Own Password" links below.

timcappalli:
Today, you have to land on a page to create a password. That page could either be the same page where other account details are requested or it could be a dedicated password page as shown in your example. 

In the first one (a registration page), you can call WebAuthn when the user clicks "Create account".

For the second one, you ultimately need to get the user to the "create password" page. So instead of a top level nav, just directing invoke WebAuthn on that link / button instead.

I don't see a scenario where conditional UI is needed for registration.

nov:
"iPhone created a passkey for this site" should be fine.

When user authenticates via WebAuthn Autofill, they won't distinguish whether they are authenticating by password or passkey.
I'd want the same level of seamless UX on registration.

I don't want to put both "create passkey" button and password input field in the flow.
I'd like to password manager create passkey when user focused on the password input field on the registration page.

timcappalli:
Why would you ask for both a password and passkey at the same time? If you want the user to create a passkey, and the platform supports passkeys, there shouldn't be a password field shown to the user.

nov:
We can use the same layout on login page because of WebAuthn Autofill feature.
Why do we have to have the different layout on registration?

timcappalli:
> Why do we have to have the different layout on registration?

In most cases, you're asking for different information on a registration flow vs a sign in flow. But even in that case, if the platform supports passkey, your "create account" button (or whatever your button says) can just result in calling WebAuthn. The user doesn't have to think about it.

nov:
[...]

PR #1951, opened 2023-08-29
Initial text for conditional create

### LINKED ISSUE #1929: Non-modal registration during conditional assertion
## Description

WebAuthn UIs for registration in browsers are typically intrusive. This presents a challenge for sites that are transitioning from usernames and passwords because RPs have to find the right moment to prompt the user to upgrade to a WebAuthn credential.

WebAuthn could provide a way for RPs to signal at time of login that they want to create WebAuthn credential for any users logging in with a username/password. These log in pages can already use conditional assertions to show credentials in the user agent's AutoFill drop down. This existing call could be extended with an extension to optionally register a credential whenever the user opts to AutoFill a username and password.

I've written an explainer linked below about how this might work.

## Related Links
[Explainer: WebAuthn Conditional Registration Extension](https://github.com/w3c/webauthn/wiki/Explainer:-Conditional-Registration-Extension)

pascoej:
Fixes #1929

***
<a href="https://pr-preview.s3.amazonaws.com/pascoej/webauthn/pull/1951.html" title="Last updated on May 15, 2024, 3:25 PM UTC (dde3ab3)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/1951/89f204f...pascoej:dde3ab3.html" title="Last updated on May 15, 2024, 3:25 PM UTC (dde3ab3)">Diff</a>


index.bs
```diff
@@ -6714,6 +6730,52 @@ During a transition from the FIDO U2F JavaScript API, a [=[RP]=] may have a popu
...
+    <xmp class="idl">
+    partial dictionary AuthenticationExtensionsClientInputs {
+        boolean conditionalCreate;
+    };
+    </xmp>
+
+: Client extension processing
+
+:: When [[[#sctn-getAssertion|assertion]]:
+        1. Just after [establishing the RP ID](#GetAssertion-DetermineConditional) perform these steps:
```
agl, line 6755: `#GetAssertion-DetermineConditional` isn't currently an anchor and doesn't seem to be added here?

index.bs
```diff
@@ -6714,6 +6730,52 @@ During a transition from the FIDO U2F JavaScript API, a [=[RP]=] may have a popu
...
+
+            1. If <code>|options|.{{CredentialRequestOptions/mediation}}</code> is not present or does not have the value
+     {{CredentialMediationRequirement/conditional}}:
+
+                1. Throw a "{{NotAllowedError}}" {{DOMException}}
+
+        1. When [showing conditional assertion form control](GetAssn-ConditionalMediation-Interact-FormControl):
```
agl, line 6762: '#' missing before `GetAssn-Cond`&hellip; (and the latter doesn't appear to exist?)

index.bs
```diff
@@ -6714,6 +6730,52 @@ During a transition from the FIDO U2F JavaScript API, a [=[RP]=] may have a popu
...
+
+            1. Inform the user that the RP will create a credential if the user agent mediates the authentication ceremony.
+
+            1. Set conditionalCreateLifetimeTimer to a client-specific default.
```
agl, line 6766:
```suggestion
            1. Set |conditionalCreateLifetimeTimer| to a client-specific default.
```

(Your call. I think we've put bars around variable names elsewhere in the spec but I'm not sure that it's done perfectly consistently.)

index.bs
```diff
@@ -6714,6 +6730,52 @@ During a transition from the FIDO U2F JavaScript API, a [=[RP]=] may have a popu
 : Authenticator extension output
 :: None.
 
+### Conditional Create Extension (<dfn>conditionalCreate</dfn>) ### {#sctn-authenticator-conditional-create-extension}
+
+This [=client extension|client=] [=authentication extension=] indicates that the RP would like to create a credential after an authenticaton ceremony is successfully mediated by the user agent.
+
```
agl, line 6736: Further up this change says:

> prominent modal UI should <i>not</i> be shown <i>unless</i> the user has already consented to create a credential via the [=conditionalCreate|conditionalCreate extension=])

That suggests that a conditional create() shows a modal UI unless the user refused consent to create a credential during the previous conditional get(). It sounds like it's a way to avoid bothering the user with a modal creation UI if they didn't "opt-in" to being bothered during autofill. (Which is different from "the create() can resolve silently if the user consented previously".)

While user agents can ultimately implement any UI, how do you want sites thinking about how to use this extension?

index.bs
emlun: For the record, this has some benefits beyond visual style: this [annotates it semantically as a variable](https://speced.github.io/bikeshed/#var-and-algorithms). Bikeshed does some minimal validation, like warning if a variable is defined but never referenced.

index.bs
```diff
@@ -1726,8 +1726,14 @@ To support obtaining assertions via {{CredentialsContainer/get()|navigator.crede
...
 [=[WRP]=] scripts to call {{CredentialsContainer/create()|navigator.credentials.create()}} to request the creation of a new
-[=public key credential source=], [=bound credential|bound=] to an [=authenticator=]. This
-{{CredentialsContainer/create()|navigator.credentials.create()}} operation can be aborted by leveraging the {{AbortController}};
+[=public key credential source=], [=bound credential|bound=] to an [=authenticator=].
+
+By setting <code>|options|.{{CredentialRequestOptions/mediation}}</code> to {{CredentialMediationRequirement/conditional}},
+[=[RPS]=] can indicate that a prominent modal UI should <i>not</i> be shown <i>unless</i> the user has already consented to create a credential via the [=conditionalCreate|conditionalCreate extension=]). [=[RP]=] script SHOULD first check that [conditionalCreate] is present
+in {=ClientCapabilities=} in order to avoid the possibility of causing a user-visible error to be returned if the user agent does
+not support {{CredentialMediationRequirement/conditional}} [=user mediation=] for {{CredentialsContainer/create()|navigator.credentials.create()}}.
```
emlun, line 1734: The `conditionalCreate` and `ClientCapabilities` that would be referred here are introduced in PR #1923, so we may need to hold this PR until that is merged.

index.bs
```diff
@@ -6714,6 +6730,52 @@ During a transition from the FIDO U2F JavaScript API, a [=[RP]=] may have a popu
 : Authenticator extension output
 :: None.
 
+### Conditional Create Extension (<dfn>conditionalCreate</dfn>) ### {#sctn-authenticator-conditional-create-extension}
+
+This [=client extension|client=] [=authentication extension=] indicates that the RP would like to create a credential after an authenticaton ceremony is successfully mediated by the user agent.
```
emlun, line 6735:
```suggestion
This [=client extension|client=] [=authentication extension=] indicates that the [=[RP]=] would like to create a credential after an authentication ceremony is successfully mediated by the user agent.
```

index.bs
```diff
@@ -6714,6 +6730,52 @@ During a transition from the FIDO U2F JavaScript API, a [=[RP]=] may have a popu
...
+:: The Boolean value [TRUE] to indicate that this extension is requested by the [=[RP]=].
+    <xmp class="idl">
+    partial dictionary AuthenticationExtensionsClientInputs {
+        boolean conditionalCreate;
+    };
+    </xmp>
+
+: Client extension processing
+
+:: When [[#sctn-getAssertion|assertion]]:
```
emlun, line 6754: What does this mean?

This is probably not needed, since the extension is explicitly for assertion only. See for example the `appid` extension, which doesn't specify an initial condition like this.

index.bs
emlun: I also find the phrasing "after an authentication ceremony is successfully mediated by the user agent" a bit weird. I would prefer something more like "if the authentication ceremony is completed successfully".

index.bs
```diff
@@ -6714,6 +6730,52 @@ During a transition from the FIDO U2F JavaScript API, a [=[RP]=] may have a popu
...
+        1. Just after [establishing the RP ID](#GetAssertion-DetermineConditional) perform these steps:
+
+            1. If <code>|options|.{{CredentialRequestOptions/mediation}}</code> is not present or does not have the value
+                {{CredentialMediationRequirement/conditional}}:
+
+                1. Throw a "{{NotAllowedError}}" {{DOMException}}
+
+        1. When [showing conditional assertion form control](GetAssn-ConditionalMediation-Interact-FormControl):
+
+            1. Inform the user that the RP will create a credential if the user agent mediates the authentication ceremony.
```
emlun, line 6764:
```suggestion
            1. Inform the user that the [=[RP]=] will create a credential if the user agent mediates the authentication ceremony.
```

emlun: Also same comment on phrasing as above.

emlun (changes requested):

index.bs
emlun: Right, and it does :slightly_smiling_face:

```
LINE ~1920: The var 'conditionalCreateLifetimeTimer' (in global scope) is only used once.
If this is not a typo, please add an ignore='' attribute to the <var>.
LINE ~1920: The var 'conditionalCreateOrigin' (in global scope) is only used once.
If this is not a typo, please add an ignore='' attribute to the <var>.
```

MasterKale:
Would anything in this proposal potentially blow up if we ended up in a future state where passkey providers offered users an option, entirely outside of the browser, to **not** silently register passkeys? That is, we'd have something like this:

- User turns off "allow sites to silently create passkeys" in their passkey provider

_[A few hours later]_

- RP sets up conditional creation with `conditionalCreate` extension
- User logs in with username + password
- Browser attempts to silently register a passkey
- Passkey provider says "nope"
- ???

Just trying to think ahead a bit, about potential controls that passkey providers might offer users (outside the influence of ostensibly user-centric WebAuthn) that would potentially clash with RP-friendly features like this one.

index.bs
```diff
@@ -1726,8 +1726,14 @@ To support obtaining assertions via {{CredentialsContainer/get()|navigator.crede
 {{PublicKeyCredential}}'s [=interface object=]'s implementation of the <dfn for="PublicKeyCredential" method>\[[Create]](origin,
 options, sameOriginWithAncestors)</dfn> [=internal method=] [[!CREDENTIAL-MANAGEMENT-1]] allows
 [=[WRP]=] scripts to call {{CredentialsContainer/create()|navigator.credentials.create()}} to request the creation of a new
-[=public key credential source=], [=bound credential|bound=] to an [=authenticator=]. This
-{{CredentialsContainer/create()|navigator.credentials.create()}} operation can be aborted by leveraging the {{AbortController}};
+[=public key credential source=], [=bound credential|bound=] to an [=authenticator=].
+
+By setting <code>|options|.{{CredentialRequestOptions/mediation}}</code> to {{CredentialMediationRequirement/conditional}},
```
nsatragno, line 1731: You mean `CredentialCreationOptions/mediation`? This would need to be added to that interface.

nsatragno (changes requested):
Really excited for this! I'll summarize my thoughts I shared in the w3c call:

* As it's written, making a credential through conditional registration requires opting in to a conditional get assertion. We should relax this requirement: it should be acceptable for an RP to offer conditional registration after the user opts in (e.g., after they chose a previously saved password in a sign in form) *without* also requiring them to offer previously registered passkeys on the same form. An example use case would be to create a passkey from a password field during registration (see #1862). I understand you are using the get assertion extension to collect consent, but I don't see why consent needs to be tied in to a get assertion call.
  If we go this route, perhaps we can do away with the assertion extension, and instead have a `create` request with `mediation: conditional` that handles obtaining consent. When it resolves, it could either run a callback or return a promise with a method to fill in the username, challenge, etc.

* `CredentialCreationOptions` needs to be updated to have a `mediation` parameter.

pascoej:
I still need to address the non-requirement of the extension and sort out a procedural issue in order to submit a PR against [credential-management](https://github.com/w3c/webappsec-credential-management).

index.bs
```diff
@@ -1726,8 +1726,14 @@ To support obtaining assertions via {{CredentialsContainer/get()|navigator.crede
 {{PublicKeyCredential}}'s [=interface object=]'s implementation of the <dfn for="PublicKeyCredential" method>\[[Create]](origin,
 options, sameOriginWithAncestors)</dfn> [=internal method=] [[!CREDENTIAL-MANAGEMENT-1]] allows
 [=[WRP]=] scripts to call {{CredentialsContainer/create()|navigator.credentials.create()}} to request the creation of a new
-[=public key credential source=], [=bound credential|bound=] to an [=authenticator=]. This
-{{CredentialsContainer/create()|navigator.credentials.create()}} operation can be aborted by leveraging the {{AbortController}};
+[=public key credential source=], [=bound credential|bound=] to an [=authenticator=].
+
+By setting <code>|options|.{{CredentialCreationOptions/mediation}}</code> to {{CredentialMediationRequirement/conditional}},
+[=[RPS]=] can indicate xc that a prominent modal UI should <i>not</i> be shown <i>unless</i> the user has already consented to create a credential via the [=conditionalCreate|conditionalCreate extension=]). [=[RP]=] script SHOULD first check that [conditionalCreate] is present
```
nicksteele, line 1732: sp

index.bs
```diff
@@ -6714,6 +6735,52 @@ During a transition from the FIDO U2F JavaScript API, a [=[RP]=] may have a popu
...
+
+            1. Set |conditionalCreateLifetimeTimer| to a client-specific default.
+
+            1. Set |conditionalCreateOrigin| to the current origin.
+
+        1. If the user agent mediates the authentication ceremony while the conditional assertion is running:
+
+            1. Start |conditionalCreateLifetimeTimer|.
+
+            1. Optionally collect an [=authorization gesture=] confirming [=user consent=] to create the credential which may later be used to skip (this step)[#op-makecred-step-user-consent].
```
MasterKale, line 6779: In this PR's current incarnation, would an RP that **cannot** do conditional UI (e.g. the RP requires username submission to check that the user can use WebAuthn according to access policy) still be able to conditionally register a passkey? I suppose I wouldn't want to see this functionality require the user to have to click something from the browser's autofill UI.

In Cisco's case, we would want to be able to do the following...

```js
/**
 * No form is actually shown at this point, we're just making this call to allow for
 * the subsequent .create() to silently register
 */
navigator.credentials.get({
  mediation: 'conditional',
  publicKey: { ..., extensions: { conditionalCreate: true } },
});
/**
 * Wait a short period of time since the .get() above is asynchronous and we want to
 * give the browser time to start the `conditionalCreateLifetimeTimer`
 */
setTimeout(() => {
    mediation: 'conditional',
    publicKey: { ... },
  }).then(sendRegRespToServer);
}, 100);
```

...so that we too would be able to use this capability to reduce friction when onboarding users to passkeys. But the way I'm reading this we would **not** be able to leverage conditional creation because the user didn't click the browser's autofill.

Can we consider ways to accommodate a use case like ours too? I'm not 100% sure why autofill interaction is required for this to be pulled off.

MasterKale (changes requested):

index.bs
pascoej: Yes. The extension is not required.

MasterKale: Ah, I missed this recent change to the PR:

Screenshot 2023-09-20 at 12 19 54 PM

So then an RP that can't do conditional UI can simply do the conditional `.create()` call after a successful auth (assuming they check that the conditional create capability is present)...

```
    mediation: 'conditional',
    publicKey: { ... },
  }).then(sendRegRespToServer);
```

...but would be at the mercy of the browser to have **implicitly** started `conditionalCreateLifetimeTimer` after the previous modal `.get()`? That would mean "conditional UI + the `conditionalCreate` extension" is an **explicit** way to ensure `conditionalCreateLifetimeTimer` gets started?

Having thought this through a bit, perhaps we need a way for the RP to query if `conditionalCreateLifetimeTimer` is not expired. Or perhaps this could get factored into whether a "client-create" capability from `getClientCapabilities()` returns true.

pascoej: Taking it as an action item to add a recommended minimum time here.

MasterKale: From the WG meeting on 9/20: I think now what's _actually_ needed is a suggested duration for `conditionalCreateLifetimeTimer` that encourages user agents to give RP's a similar amount of time to make the conditional create call.

pascoej:
Associated credential management PR: https://github.com/w3c/webappsec-credential-management/pull/224

MasterKale (approved):

nsatragno:
I asked for a few changes on the CredMan PR, but I don't think they will affect the integration on webauthn -- thanks!

My question about the extension requirement still stands. We can gain a lot from decoupling conditional creation from conditional assertion at the spec level.

pascoej:
I've addressed comments and added more clarification that the extension is not required.

index.bs
```diff
@@ -1903,6 +1914,16 @@ a numbered step. If outdented, it (today) is rendered as a bullet in the midst o
     [=authenticators=] can be <a href="https://en.wikipedia.org/w/index.php?title=Hot_plug">hot-plugged</a> into (e.g., via USB)
     or discovered (e.g., via NFC or Bluetooth) by the [=client=] by various mechanisms, or permanently built into the [=client=].
 
+1. If <code>|options|.{{CredentialRequestOptions/mediation}}</code> is present with the value
+     {{CredentialMediationRequirement/conditional}}:
+
+    1. If |conditionalCreateLifetimeTimer| is expired or |conditionalCreateOrigin| is not |callerOrigin|, throw a "{{NotAllowedError}}" {{DOMException}}.
+
+    Note: |conditionalCreateLifetimeTimer| and |conditionalCreateOrigin| will be set by the user agent after it believes an authentication ceremony has
+    been completed and the user consents to this type of credential creation. This consent may be collected via means other than the [=conditionalCreate|Conditional Create Extension=].
```
nsatragno, line 1923:
```suggestion
    been completed and the user consents to this type of credential creation. This consent may be collected via means determined by the user agent.
```

index.bs
```diff
@@ -1726,8 +1726,14 @@ To support obtaining assertions via {{CredentialsContainer/get()|navigator.crede
 {{PublicKeyCredential}}'s [=interface object=]'s implementation of the <dfn for="PublicKeyCredential" method>\[[Create]](origin,
```
nsatragno, line 1726: Could you update the language for [isConditionalMediationAvailable()](https://pr-preview.s3.amazonaws.com/w3c/webauthn/1951/410d0f7...pascoej:2b8521b.html#dom-publickeycredential-isconditionalmediationavailable) to reflect that it only applies to `get` requests?

nsatragno (commented):
This is good to land minus two comments. I think the PR also needs a rebase, as the diff is showing a couple unrelated changes.

nsatragno (approved):
Approved, thank you! ^_^

pascoej:
reminder to myself to rebase this to remove the unrelated changes and blank line removals.

nicksteele (approved):

timcappalli (dismissed):
LGTM

index.bs
```diff
@@ -36,6 +36,7 @@ Former Editor: Rolf Lindemann, w3cid 84447, Nok Nok Labs, rolf@noknok.com
 !Contributors: <a href="mailto:agl@google.com">Adam Langley</a> (Google)
 !Contributors: <a href="mailto:mandyam@qti.qualcomm.com">Giridhar Mandyam</a> (Qualcomm)
 !Contributors: <a href="mailto:mattmil3@cisco.com">Matthew Miller</a> (Cisco)
+!Contributors: <a href="mailto:pascoe@apple.com">Pascoe</a> (Apple)
```
emlun, line 39: No first name? (I don't know whether W3C has anything against omitting it, I just noticed it stuck out a bit)

index.bs
```diff
@@ -1903,6 +1916,16 @@ a numbered step. If outdented, it (today) is rendered as a bullet in the midst o
     [=authenticators=] can be <a href="https://en.wikipedia.org/w/index.php?title=Hot_plug">hot-plugged</a> into (e.g., via USB)
     or discovered (e.g., via NFC or Bluetooth) by the [=client=] by various mechanisms, or permanently built into the [=client=].
 
+1. If <code>|options|.{{CredentialRequestOptions/mediation}}</code> is present with the value
+     {{CredentialMediationRequirement/conditional}}:
+
+    1. If |conditionalCreateLifetimeTimer| is expired or |conditionalCreateOrigin| is not |callerOrigin|, throw a "{{NotAllowedError}}" {{DOMException}}.
+
+    Note: |conditionalCreateLifetimeTimer| and |conditionalCreateOrigin| will be set by the user agent after it believes an authentication ceremony has
+    been completed and the user consents to this type of credential creation.
```
emlun, line 1925: Is this specified anywhere? I can't see any other references to `conditionalCreateLifetimeTimer` or `conditionalCreateOrigin` either here or in https://github.com/w3c/webappsec-credential-management/pull/224 .

index.bs
```diff
@@ -1903,6 +1916,16 @@ a numbered step. If outdented, it (today) is rendered as a bullet in the midst o
...
+
+    1. Set |lifetimeTimer| to a client-specific default.
```
emlun, line 1927: This doesn't seem to take `pkOptions.timeout` into account at all, is that intended?

index.bs
```diff
@@ -2265,19 +2288,22 @@ When this method is invoked, the user agent MUST execute the following algorithm
 
 1. Let |pkOptions| be the value of <code>|options|.{{CredentialRequestOptions/publicKey}}</code>.
 
-1. If <code>|options|.{{CredentialRequestOptions/mediation}}</code> is present with the value
-    {{CredentialMediationRequirement/conditional}}:
 
-    1. Let |credentialIdFilter| be the value of <code>|pkOptions|.{{PublicKeyCredentialRequestOptions/allowCredentials}}</code>.
+        <li id='GetAssn-DetermineConditional'>
```
emlun, line 2292: This anchor seems no longer used, so I think this change can be reverted?

emlun: Reverted this in aca236f39be577276a7ff962badc33ad6f0d41fb.

index.bs
pascoej: In most cases, this will resolve quickly. Though I suppose clients might want to have optional authenticator selection upon conditional create one day. I modified this section to take in account `pkOptions.timeout`.

index.bs
pascoej: It was defined in the extension, which got removed. I reworded this section to not use these variables and instead specify that the user agent must believe an authentication ceremony recently took place and that the user consents to the credential creation.

index.bs
pascoej: That is my full government name.

emlun: Oh, my apologies!

index.bs
```diff
@@ -1925,15 +1925,19 @@ a numbered step. If outdented, it (today) is rendered as a bullet in the midst o
     [=authenticators=] can be <a href="https://en.wikipedia.org/w/index.php?title=Hot_plug">hot-plugged</a> into (e.g., via USB)
     or discovered (e.g., via NFC or Bluetooth) by the [=client=] by various mechanisms, or permanently built into the [=client=].
 
-1. If <code>|options|.{{CredentialRequestOptions/mediation}}</code> is present with the value
+1. If <code>|options|.{{CredentialCreationOptions/mediation}}</code> is present with the value
      {{CredentialMediationRequirement/conditional}}:
 
-    1. If |conditionalCreateLifetimeTimer| is expired or |conditionalCreateOrigin| is not |callerOrigin|, throw a "{{NotAllowedError}}" {{DOMException}}.
+    1. If the user agent has not recently mediated an authentication, the origin of said authentication is not |callerOrigin|, or the user
+    does not consent to this type of credential creation, throw a "{{NotAllowedError}}" {{DOMException}}.
```
emlun, line 1932: This should be indented one step further:
```suggestion
        does not consent to this type of credential creation, throw a "{{NotAllowedError}}" {{DOMException}}.
```

index.bs
```diff
@@ -1925,15 +1925,19 @@ a numbered step. If outdented, it (today) is rendered as a bullet in the midst o
...
 
-    Note: |conditionalCreateLifetimeTimer| and |conditionalCreateOrigin| will be set by the user agent after it believes an authentication ceremony has
-    been completed and the user consents to this type of credential creation.
+    Note: The user agent will note when it believes an authentication ceremony has
+    been completed.
```
emlun, line 1935: This needs to be indented one step further in order to not break up the numbered list.

I would also not make this a "Note" since those are [supposed to be non-normative](https://github.com/w3c/webauthn/issues/1979), and this is a fairly important piece.

```suggestion
        It is up to the user agent to decide when it believes an authentication ceremony has been completed.
        That authentication ceremony MAY be performed using other means than the [=Web Authentication API=].
```

index.bs
```diff
@@ -1925,15 +1925,19 @@ a numbered step. If outdented, it (today) is rendered as a bullet in the midst o
...
 
-    1. Set |lifetimeTimer| to a client-specific default.
+    1. If <code>|pkOptions|.{{CredentialCreationOptions/timeout}}</code> is present, check if its value lies
+        within a reasonable range as defined by the [=client=] and if not, correct it to the closest value lying within that range.
+        Set a timer |lifetimeTimer| to this adjusted value. If <code>|pkOptions|.{{CredentialCreationOptions/timeout}}</code>
+        is not present, then set |lifetimeTimer| to a [=client=]-specific default.
```
emlun, line 1941: Now that this is the same as step 4, I think we can just delete this step?
```suggestion
```

emlun (changes requested):

emlun (approved):
LGTM! Sorry for the delayed review.

index.bs
```diff
@@ -2199,7 +2222,7 @@ a numbered step. If outdented, it (today) is rendered as a bullet in the midst o
     [[#sctn-make-credential-privacy]] for details.
 
 During the above process, the user agent SHOULD show some UI to the user to guide them in the process of selecting and
-authorizing an authenticator.
+authorizing an authenticator. When <code>|options|.{{CredentialCreationOptions/mediation}}</code> is set to {{CredentialMediationRequirement/conditional}}, prominent modal UI should <i>not</i> be shown <i>unless</i> credential creation was previously consented to via means determined by the user agent.
```
akshayku, line 2225: What should be the UP and UV flags for the response returned in this silent operation? If UP is zero, then current spec does not allow that and we may need to see that section. Or are we saying that UP is 1, in this conditional create?

index.bs
```diff
@@ -1736,8 +1738,14 @@ To support obtaining assertions via {{CredentialsContainer/get()|navigator.crede
 {{PublicKeyCredential}}'s [=interface object=]'s implementation of the <dfn for="PublicKeyCredential" method>\[[Create]](origin,
 options, sameOriginWithAncestors)</dfn> [=internal method=] [[!CREDENTIAL-MANAGEMENT-1]] allows
 [=[WRP]=] scripts to call {{CredentialsContainer/create()|navigator.credentials.create()}} to request the creation of a new
```
ve7jtb, line 1740: In the CredentialsContainer/create() the RP will pass in user. id and names.   
Are these names the one the authenticator must use or can the passkey provider override them?
Is the assumption that this goes back to the password manager used to fill in the password, or would the user be given a choice of where to create it?

index.bs
pascoej: We would report UP=1 for conditional create. UV depends on the specifics of the authenticator. Some might be able to collect UV, others might not and would be unable to satisfy requests with UV=required.

index.bs
pascoej: The user.id and user.name will be respected as normal. The client may choose pass along which password account was used for authentication for the purposes of linking. The default passkey provider used is up to the client. I imagine it'd be a user configurable setting of all providers that could possibly satisfy such a request.

MasterKale:
There doesn't seem to be anything in here addressing users with multiple available providers. When a user has a security key inserted, has iCloud Keychain enabled, and a third-party provider integrating OS passkey provider APIs, which of these receives the silent registration response?

If this might be handled via client configuration, can we add verbiage in here requiring clients to offer users with such a configuration capability? I think the behavior of conditional creation could get very confusing very quickly if it's left up to the clients to figure out what to do here.

MasterKale (changes requested):
I'm realizing at the WG meeting on 3/27 that this proposal has gone in a different direction since I initially approved. I can't think of another way to revert my approval so I have to request changes.

I'm committing to reviewing and approving later when some of the questions that came up above and at the meeting today get addressed.

timcappalli:
3 concerns I have from the 2024-03-27 meeting:

1. UP and UV are not occurring during the WebAuthn create ceremony
2. There is no binding between the autofill provider and authenticator
3. This doesn't seem realistically interoperable until there is a better client solution for extension-based passkey providers

timcappalli:
Here's a proposal based on the discussion and concerns expressed yesterday:

When conditional create is used, UV and UP **MUST** be `0`/`false` unless UP or UV is performed as part of the WebAuthn ceremony. Conditional creation requests SHOULD use UV=preferred or UV=discouraged.

As was mentioned during the discussion, since conditional creation is an opt in feature for RPs, so changes are required by the RP, and the RP's logic will need to match the use case and flow, which means accepting a response with UP=0 and UV=0.

This balances the concerns about redefining UP and UV with the usability enhancements that are brought by conditional create.

NOTE: there are CTAP changes that will need to be made, likely as part of CTAP 2.3, for this to be supported with security keys for feature/experience parity.

index.bs
akshayku: Typical authenticator have no insight to what has happened in the browser layer. As the operation is silent from authenticator's boundary perspective, UP=0 and UV=0 makes more sense to me. 

This is an opt-in behavior so RP can make changes to expect such a response.

nov: As an RP developer, I'd like to request _"I want UV on authentication, but no need at registration timing"_, and want authenticator answer _"OK, I didn't do UV on registration, but I ensure that you can request UV on following authentications"_.
Does webauthn has such mechanism now?

timcappalli: @nov, creation and authentication are different ceremonies. You as the RP are in control of whether UV is required or not for each ceremony.

nov: I don't want users to register passkeys which cannot fulfill UV requirement on future authentication.

timcappalli: > I don't want users to register passkeys which cannot fulfill UV requirement on future authentication.

Then you need to use UV=required, and this feature won't be applicable to your environment.

index.bs
```diff
@@ -1582,15 +1583,16 @@ that are returned to the caller when a new credential is created, or a new asser
 
     :   {{PublicKeyCredential/isConditionalMediationAvailable()}}
     ::  {{PublicKeyCredential}} overrides this method to indicate availability for {{CredentialMediationRequirement/conditional}}
-        mediation. [=[WRPS]=] SHOULD verify availability before attempting to set
-        <code>|options|.{{CredentialRequestOptions/mediation}}</code> to {{CredentialMediationRequirement/conditional}}.
+        mediation during {{CredentialsContainer/get()|navigator.credentials.get()}}. [=[WRPS]=] SHOULD verify availability before
```
timcappalli, line 1586: s/SHOULD/MUST as discussed on 2024-04-03 call.

timcappalli (changes requested):
One small change from 2024-04-03 call

timcappalli:
The verification step needs to be updated as well to add an if mediation=conditional check.

`1. Verify that the [=UP=] bit of the <code>[=flags=]</code> in |authData| is set.`

https://github.com/w3c/webauthn/blob/89f204f4b1b02a147dfcdee4fbd88319cab5e104/index.bs#L5372

index.bs
```diff
@@ -2000,7 +2025,14 @@ a numbered step. If outdented, it (today) is rendered as a bullet in the midst o
...
                             :   is set to {{UserVerificationRequirement/required}}
-                            ::  Let |userVerification| be [TRUE].
+                            ::  If <code>|options|.{{CredentialCreationOptions/mediation}}</code> is set to
+                                <dl class="switch">
+                                    :   {{CredentialMediationRequirement/conditional}}
+                                    ::  throw a {{NotAllowedError}} {{DOMException}} unless it can be explicitly collected during the ceremony.
+
+                                    :   empty or another value
+                                    ::  Let |userVerification| be [TRUE].
+                                </dl>
```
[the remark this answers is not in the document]
emlun: Also, is `NotAllowedError` the right behaviour here? The closest analogue currently in the spec is the "error code equivalent to `ConstraintError`" returned by the authenticator when UV is required but cannot be satisfied by that authenticator. Returning an error at all also differs from `conditional` mediation in `get()`, which simply hangs forever on most errors. Should `conditional` create do that too?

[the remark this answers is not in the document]
pascoej: The idea with conditional registration is that you invoke it right after logging in and get a result quickly. If the user didn't consent to this type of create, then they will quickly get an error, otherwise a credential. 

I think `ConstraintError` does sound okay in this case. I've made the change.

index.bs
```diff
@@ -1736,8 +1738,16 @@ To support obtaining assertions via {{CredentialsContainer/get()|navigator.crede
...
-{{CredentialsContainer/create()|navigator.credentials.create()}} operation can be aborted by leveraging the {{AbortController}};
+[=public key credential source=], [=bound credential|bound=] to an [=authenticator=].
+
+By setting <code>|options|.{{CredentialCreationOptions/mediation}}</code> to {{CredentialMediationRequirement/conditional}},
+[=[RPS]=] can indicate that they would like to register a credential without prominent modal UI if user has already consented to create a credential. The [=[RP]=] SHOULD first check that {{ClientCapability/conditionalCreate}} is present
+in the result of {{PublicKeyCredential/getClientCapabilities()}} in order to avoid the possibility of causing a user-visible error to be returned if the user agent does
+not support {{CredentialMediationRequirement/conditional}} [=user mediation=] for {{CredentialsContainer/create()|navigator.credentials.create()}}.
+The client MUST set BOTH |requireUserPresence| and |requireUserVerification| to |FALSE| when <code>|options|.{{CredentialCreationOptions/mediation}}</code> is set to {{CredentialMediationRequirement/conditional}}
+unless they may explicitly collected during the ceremony.
```
timcappalli, line 1748:
```suggestion
unless they are explicitly performed during the ceremony.
```

timcappalli (approved):
LGTM

index.bs
```diff
@@ -2000,7 +2025,10 @@ a numbered step. If outdented, it (today) is rendered as a bullet in the midst o
                         <dl class="switch">
 
                             :   is set to {{UserVerificationRequirement/required}}
-                            ::  Let |userVerification| be [TRUE].
+                            ::  1. If <code>|options|.{{CredentialCreationOptions/mediation}}</code> is set to {{CredentialMediationRequirement/conditional}}
+                                    and [=user verification=] cannot be collected during the ceremony,
+                                    throw a {{ConstraintError}} {{DOMException}}.
```
MasterKale, line 2030: Note to self: this new case for `ConstraintError` will need to get documented in the same Registration API Exceptions being proposed in #2047.

index.bs
```diff
@@ -4612,9 +4640,9 @@ It takes the following input parameters:
 : |requireResidentKey|
 :: The [=effective resident key requirement for credential creation=], a Boolean value determined by the [=client=].
 : |requireUserPresence|
-:: The constant Boolean value [TRUE].
+:: The constant Boolean value [TRUE] unless <code>|options|.{{CredentialCreationOptions/mediation}}</code> is set to {{CredentialMediationRequirement/conditional}}.
     It is included here as a pseudo-parameter to simplify applying this abstract authenticator model to implementations that may
-    wish to make a [=test of user presence=] optional although WebAuthn does not.
+    wish to make a [=test of user presence=] optional.
```
MasterKale, line 4645: Trimming "although WebAuthn does not" entirely could open the door to authenticator models that do weird things for sake of making UP optional in more scenarios.

I know the addition of "unless" in the line above helps narrow down cases when `requireUserPresence` might be set to `FASLE`. But perhaps we need to get hyper-specific on this line here too because otherwise I think it introduces some ambiguity.

index.bs
```diff
@@ -4612,9 +4640,9 @@ It takes the following input parameters:
 : |requireResidentKey|
 :: The [=effective resident key requirement for credential creation=], a Boolean value determined by the [=client=].
 : |requireUserPresence|
-:: The constant Boolean value [TRUE].
+:: The constant Boolean value [TRUE] unless <code>|options|.{{CredentialCreationOptions/mediation}}</code> is set to {{CredentialMediationRequirement/conditional}}.
```
MasterKale, line 4643: Would there be a case of `requireUserPresence` still getting set to `True` even if conditional mediation is present in the `.create()` options? If not perhaps we more narrowly define when this line to require `False` here:
```suggestion
:: The constant Boolean value [TRUE], or [FALSE] when <code>|options|.{{CredentialCreationOptions/mediation}}</code> is set to {{CredentialMediationRequirement/conditional}} and the user agent previously collected consent from the user.
```

index.bs
pascoej: I think removing the line entirely makes this most clear with the revised line above.

akshayku (approved):

PR #1954, opened 2023-08-30
Enterprise packed attestation guidance

### LINKED ISSUE #1916: Describe packed enterprise attestation
## Proposed Change

To complement the new [Packed Attestation Statement Certificate Requirements](https://w3c.github.io/webauthn/#sctn-packed-attestation-cert-requirements) section, we should also describe requirements for packed attestations returned from Enterprise Attestation. 

This should describe the use of [`id-fido-gen-ce-sernum`](http://oid-info.com/get/1.3.6.1.4.1.45724.1.1.2) as well as [`id-fido-gen-ce-fw-version`](http://oid-info.com/get/1.3.6.1.4.1.45724.1.1.5)

dwaite:
Fixes #1916

Creates a new section describing verification when packed format is used for enterprise attestation.

***
<a href="https://pr-preview.s3.amazonaws.com/dwaite/webauthn/pull/1954.html" title="Last updated on Sep 18, 2024, 6:51 PM UTC (39733f0)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/1954/403c2b3...dwaite:39733f0.html" title="Last updated on Sep 18, 2024, 6:51 PM UTC (39733f0)">Diff</a>


index.bs
```diff
@@ -5787,6 +5787,18 @@ The attestation certificate MUST have the following fields/extensions:
...
+
+There are two potential sources for Enterprise Attestestations in Packed format:
+
+1. From a hardware device which is manufactured with a device-specific attestation statement and key
+2. Via a provisioned attestation statement, such as with a platform authenticator configured via managed policy
+
+For attestation statements provisioned at manufacturing, the Extension OID `1.3.6.1.4.1.45724.1.1.2` (`id-fido-gen-ce-sernum`) 
+MUST be present, containing a unique serial number for the device as an OCTET STRING. The extension MUST NOT be marked as critical.
+
+As enterprise attestations are normally consumed by relying parties which are looking for particular authenticators,
```
emlun, line 5800:
```suggestion
As enterprise attestations are normally consumed by [=[RPS]=] which are looking for particular authenticators, 
```

index.bs
```diff
@@ -5787,6 +5787,18 @@ The attestation certificate MUST have the following fields/extensions:
...
 
+### Certificate Requirements for Enterprise Packed Attestation Statements ### {#sctn-enterprise-packed-attestation-cert-requirements}
+
+There are two potential sources for Enterprise Attestestations in Packed format:
+
+1. From a hardware device which is manufactured with a device-specific attestation statement and key
+2. Via a provisioned attestation statement, such as with a platform authenticator configured via managed policy
+
+For attestation statements provisioned at manufacturing, the Extension OID `1.3.6.1.4.1.45724.1.1.2` (`id-fido-gen-ce-sernum`) 
+MUST be present, containing a unique serial number for the device as an OCTET STRING. The extension MUST NOT be marked as critical.
```
emlun, line 5798: Does it matter how the serial number (if it is in fact an integer) is encoded as an OCTET STRING? If so, we should define that encoding (probably two's complement, but what byte order?). If not, maybe we should describe it as "containing a unique device identifier" instead of as a number, to make it clear that it's an opaque byte array whose internal structure is undefined?

Also, an example like the one for `id-fido-gen-ce-aaguid` would be helpful to point out that the value needs to be two nested OCTET STRING values.

index.bs
```diff
@@ -5787,6 +5787,18 @@ The attestation certificate MUST have the following fields/extensions:
     are both OPTIONAL as the status of many attestation certificates is available through authenticator metadata services.
     See, for example, the FIDO Metadata Service [[FIDOMetadataService]].
 
+### Certificate Requirements for Enterprise Packed Attestation Statements ### {#sctn-enterprise-packed-attestation-cert-requirements}
+
+There are two potential sources for Enterprise Attestestations in Packed format:
+
+1. From a hardware device which is manufactured with a device-specific attestation statement and key
```
agl, line 5794: (nit: periods at the ends of these sentences.)

index.bs
```diff
@@ -5787,6 +5787,18 @@ The attestation certificate MUST have the following fields/extensions:
...
+2. Via a provisioned attestation statement, such as with a platform authenticator configured via managed policy
+
+For attestation statements provisioned at manufacturing, the Extension OID `1.3.6.1.4.1.45724.1.1.2` (`id-fido-gen-ce-sernum`)
```
agl, line 5797: You enumerate two cases just above, but then only mention one here. Are provisioned enterprise attestations forbidden from using this extension? Or can either use them?

dwaite: I'd raise to someone more familiar such as @ve7jtb

pascoej (approved):

index.bs
Firstyear: The double octet string is just because there is inner ASN in the field that way you could embed a struct or enum into the oid field. It's similar in LDAP in a few places. The outer octet string is parsed as a new blob of asn and the inner can be any ASN value as a result. That way the oid tells a client how to parse/what to expect of the inner ASN blob.

index.bs
agl: Sounds good!

index.bs
```diff
@@ -5787,6 +5787,18 @@ The attestation certificate MUST have the following fields/extensions:
     are both OPTIONAL as the status of many attestation certificates is available through authenticator metadata services.
     See, for example, the FIDO Metadata Service [[FIDOMetadataService]].
 
+### Certificate Requirements for Enterprise Packed Attestation Statements ### {#sctn-enterprise-packed-attestation-cert-requirements}
```
sbweeden, line 5790: What in the spec references this section - i.e. when does an RP detect it has received an enterprise attestation, and should apply these processing rules?
This seems particularly important because there was talk about when an RP requests enterprise attestation, and none is available, it could fallback to direct.

sbweeden: Also, do any of the packed attestation certificate requirements above also apply to enterprise attestation certificates, or are these checks mutually exclusive?

sbweeden (changes requested):

agl (approved):

index.bs
dwaite: We are still pending guidance here on usage for MDM-provisioned EAs. Since these can be dynamic (reprovisioned by the MDM or by switching which software provides the credentials), I would lean toward either saying this should not be used, or that it must indicate a persistent device identifier (and not some software value).

dwaite: will update to recommend against this for software-provisioned EAs

index.bs
```diff
@@ -5872,6 +5872,19 @@ The attributes above are structured within this certificate as such:
...
+
+There are two potential sources for enterprise attestations in Packed format.
+
+1. Attestation statements from a hardware device, which was manufactured with the statement and a corresponding private key.
+2. Attestation statements which have been provisioned, such as a platform authenticator configured via managed policy.
+
+For attestation statements provided at manufacturing, the Extension OID `1.3.6.1.4.1.45724.1.1.2` (`id-fido-gen-ce-sernum`)
+MUST be present, containing a unique serial number for the device as an OCTET STRING. The extension MUST NOT be marked as critical.
```
emlun, line 5883: Is there any way the RP can detect whether this is the case, and thus enforce this requirement? If not, the only actually actionable requirement in this section is that when the extension is present, the RP can verify that it is not marked critical. It's already a bit awkward that many of the requirements set in §8.2.1 (in particular the requirements for the subject field, other than Subject-OU) can only really be verified by a certification auditor, and not by a WebAuthn RP.

If this requirement is not really actionable for RPs, perhaps we should rephrase this section from a set of requirements to verify, to just a description of how these values can be retrieved and parsed if present?

dwaite:
@emlun I took a pass at some new proposed text, based on your above feedback and the corresponding conversation on the call about the variance of enterprise attestations:

> 8.2.2. Certificate Requirements for Enterprise Packed Attestation Statements 
> 
> Enterprise attestations are often used to restrict access to enterprise resources to a small set of specific, vetted authenticators. Examples might include issuing hardware-based roaming authenticators to employees with a specific configuration, or deploying software authenticators via a system for managing corporate-owned devices.
> 
> This close relationship between relying party and authenticator may lead to a higher degree of variance in the attributes available in the enterprise attestation. Similar to non-enterprise attestations, there may be certification bodies that set additional requirements on enterprise attestations in order to achieve certification.
> 
> The Extension OID  1.3.6.1.4.1.45724.1.1.2  (  id-fido-gen-ce-sernum  ) MAY be present, and if so MUST indicate a unique value per device against a particular AAGUID. This value MUST remain constant through factory resets, but MAY be distinct from any other serial number or other hardware identifier associated with the device. This extension MUST NOT be marked as critical, and the corresponding value is encoded as an OCTET STRING. This extension MUST NOT be present in non-enterprise attestations.

emlun:
The text looks mostly fine to me, but I wonder if we even need to comment at all on the "higher degree of variance ..." etc.  To me it seems rather obvious, and rather irrelevant to the WebAuthn spec, that some certification bodies might impose additional requirements. I would also hesitate to write "are often used" about a feature not yet deployed. :slightly_smiling_face:

So yeah, I would omit the second paragraph. Probably the first one too, leaving just the actual requirements. I reckon the use cases of a unique serial number are probably obvious enough.

> 8.2.2. Certificate Requirements for Enterprise Packed Attestation Statements
> The Extension OID  1.3.6.1.4.1.45724.1.1.2  (  id-fido-gen-ce-sernum  ) MAY be present, and if so MUST indicate a unique value per device against a particular AAGUID. This value MUST remain constant through factory resets, but MAY be distinct from any other serial number or other hardware identifier associated with the device. This extension MUST NOT be marked as critical, and the corresponding value is encoded as an OCTET STRING. This extension MUST NOT be present in non-enterprise attestations.

I also haven't yet seen a clear answer to [my question above](https://github.com/w3c/webauthn/pull/1954#discussion_r1311514087): is this OCTET STRING value in general an encoded integer, or just an opaque octet string with undefined internal structure?

dwaite:
> So yeah, I would omit the second paragraph. Probably the first one too, leaving just the actual requirements. I reckon the use cases of a unique serial number are probably obvious enough.

This text exists because there not text describing what enterprise attestations actually are (delegated to CTAP), and it is meant to be a section on requirements on what to expect/how to process these attestations. Those sorts of processing instructions aren't really provided by CTAP (which is not really describing RP behavior). 

We can eliminate the first paragraph by pointing to https://fidoalliance.org/specs/fido-v2.1-ps-20210615/fido-client-to-authenticator-protocol-v2.1-ps-errata-20220621.html#sctn-feature-descriptions-enterp-attstn

I agree the second paragraph doesn't belong in normative text and thus should be removed. Also, I would happily entertain another document where implementation guidance would be more appropriate.

> I also haven't yet seen a clear answer to [my question above](https://github.com/w3c/webauthn/pull/1954#discussion_r1311514087): is this OCTET STRING value in general an encoded integer, or just an opaque octet string with undefined internal structure?

@ve7jtb @dturnerx do you have any guidance here? The only example I have access to is an OCTET_STRING of an OCTET_STRING of a binary serial number.

dwaite:
Will proceed to redo this shortly based on the immediately previous proposal and the example I have access to (OCTET STRING of OCTET STRING of binary serial value)

ve7jtb:
As currently defined by Fido it is an opaque blob.   That allows vendors to encode ehat they like in it.   Some may encode a intiger and others my have alpha numeric serials.   We registerd the OID so that Fido servers could have a standard place to find the info.   The RP should be using it to look up what the public key for the attestation should be in the information they receved from the vendor when purchasing the key.    So yes there should probably be some dicument outside of webAuthn to explain what the RP could/should do with an enterprise attestation.    Diffrent RP will pronbably do diffrent things,   Example:
1) check the attestation sig and AAGUID to see if it is one that they purchased.
2) check the opaque serial string to see if it matches a key that was assigned to a specific user, or just record the serial of the key durring registration.   For example, Google, if a serial is seen coming in from a new user when they create a credential, any credentials created by another user with that same serial are deprovisiond from those users (This is a sort of cleanup). 
3) lookup the attestation public key from the serial blob from outof band provisioning information from the vendor and reject if the key dosen't match the serial.

Some combinatio of all of those are possible.

dwaite:
@emlun, could you review the simplified text (both in this section and the modification on the "enterprise" attestation conveyance if it solves your text issues?
 
@sbweeden and @emlun, since the enterprise attestation requirements really amount to one optional attribute, do you feel this section needs processing requirements or that perhaps we should instead have separate, generalized RP guidance outside "packed" attestations?

@emlun and @ve7jtb, do you want to see additional text around comparing serial numbers beyond the tweaks I made to the text?

dwaite:
I additionally still plan to add a mock enterprise attestation to this section.

sbweeden:
> @sbweeden and @emlun, since the enterprise attestation requirements really amount to one optional attribute, do you feel this section needs processing requirements or that perhaps we should instead have separate, generalized RP guidance outside "packed" attestations?

I find it disappointing that there is no defined structure for the data contained within the enterprise OID extension, similar to what this is for the packed attestation OID 1.3.6.1.4.1.45724.1.1.4. Makes it impossible for an RP to build a product that has consistent processing for different enterprise attestations.

I realise this is not within WebAuthn WG's domain, but still... As for WebAuthn, perhaps text that says it must be validated _using a vendor specified process_, or similar?

dwaite:
> I find it disappointing that there is no defined structure for the data contained within the enterprise OID extension, similar to what there is for the packed attestation OID 1.3.6.1.4.1.45724.1.1.4. Makes it impossible for an RP to build a product that has consistent processing for different enterprise attestations.

The value is meant to be unique per AAGUID as an octet string, so comparison itself is defined. 

The non-defined, vendor specific part is how those binary values are supplied. If the authenticator is listed in a CSV with a serial number of 12345678, there isn't a single industry process to convert that into the OCTET_STRING for comparison.

Likewise, if there is a serial number is printed onto the authenticator, there isn't a process for how the field can be displayed in a way that can be used to correlate the two (or that the two actually be the same).

> I realise this is not within WebAuthn WG's domain, but still... As for WebAuthn, perhaps text that says it must be validated _using a vendor specified process_, or similar?

I think this is a good item to add to the hypothetical implementor's guidance for enterprise attestation.

sbweeden:
> > I find it disappointing that there is no defined structure for the data contained within the enterprise OID extension, similar to what there is for the packed attestation OID 1.3.6.1.4.1.45724.1.1.4. Makes it impossible for an RP to build a product that has consistent processing for different enterprise attestations.
> 
> The value is meant to be unique per AAGUID as an octet string, so comparison itself is defined.
> 
> The non-defined, vendor specific part is how those binary values are supplied. If the authenticator is listed in a CSV with a serial number of 12345678, there isn't a single industry process to convert that into the OCTET_STRING for comparison.
> 
> Likewise, if there is a serial number is printed onto the authenticator, there isn't a process for how the field can be displayed in a way that can be used to correlate the two (or that the two actually be the same).
> 
> > I realise this is not within WebAuthn WG's domain, but still... As for WebAuthn, perhaps text that says it must be validated _using a vendor specified process_, or similar?
> 
> I think this is a good item to add to the hypothetical implementor's guidance for enterprise attestation.

If the value is bytes, unique to the authenticator (per AAGUID), then the processing rules could just say that these bytes are to be validated or interpreted "per an individual authenticator vendor's process" then that's good enough. 

We could suggest an example, such as: "The bytes encapsulate a unique serial number for the authenticator, within the scope of its AAGUID. A relying party would typically ensure that these bytes conform to the expected format defined by the authenticator vendor, and bind the value represented by those bytes to the user account registering the credential".

Does this, or something like it, seem reasonable?

agl (approved):

index.bs
dwaite: resolved for TPAC 2024 meeting

PR #1966, opened 2023-09-18
Specify what an unknown type credential descriptor being ignored means

### LINKED ISSUE #1748: Better specify what an unknown type credential descriptor being ignored means
We use credential descriptors "`PublicKeyCredentialDescriptor`" for allow lists & exclude lists. The spec says

> [...][client platforms](https://w3c.github.io/webauthn/#client-platform) MUST ignore any [PublicKeyCredentialDescriptor](https://w3c.github.io/webauthn/#dictdef-publickeycredentialdescriptor) with an unknown [type](https://w3c.github.io/webauthn/#dom-publickeycredentialdescriptor-type).

There's no further specification for unknown types. If we follow the spec to the letter, a possible interpretation ends up with:

```javascript
navigator.credentials.get({
  publicKey: {
    allowCredentials: [{type: "not-yet-supported-by-browser", id: <id>}],
  }
});
```

Resulting in an empty `allowCredentials` request, which is definitely not what the RP intended. I think we should immediately return `NotAllowedError` in this case, based on prior text:

> If issuedRequests is empty, options.[allowCredentials](https://w3c.github.io/webauthn/#dom-publickeycredentialrequestoptions-allowcredentials) is not empty, and no authenticator will become available for any [public key credentials](https://w3c.github.io/webauthn/#public-key-credential) therein,
Indicate to the user that no eligible credential could be found. When the user acknowledges the dialog, throw a "[NotAllowedError](https://webidl.spec.whatwg.org/#notallowederror)" [DOMException](https://webidl.spec.whatwg.org/#idl-DOMException).

This is not a problem for `excludeCredentials`, since in that case the browser will simply discard the unknown credential type.

amoseui:
The spec describes that client platforms MUST ignore any PublicKeyCredentialDescriptor with an unknown type. However, there is no further specification about the case when this results in an empty allowCredentials. It must not be treated as an empty list.

The client MUST return an error if none of the listed credentials can be used in allowCredentials. For instance, if all of the listed credentials have PublicKeyCredentialDescriptor with an unknown type, the client MUST throw NotAllowedError.

Fixes #1748

***
<a href="https://pr-preview.s3.amazonaws.com/amoseui/webauthn/pull/1966.html" title="Last updated on Sep 18, 2023, 4:09 PM UTC (5f47918)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/1966/58d60d0...amoseui:5f47918.html" title="Last updated on Sep 18, 2023, 4:09 PM UTC (5f47918)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -3529,6 +3529,8 @@ an assertion. Its {{PublicKeyCredentialRequestOptions/challenge}} member MUST be
             of [[#sctn-op-get-assertion]]).
 
         If not [=list/empty=], the client MUST return an error if none of the listed credentials can be used.
+        For instance, if all of the listed credentials have {{PublicKeyCredentialDescriptor}} with an unknown {{PublicKeyCredentialDescriptor/type}},
+        the client MUST throw "{{NotAllowedError}}".
 
         The list is ordered in descending order of preference: the first item in the list is the most
         preferred credential, and the last is the least preferred.


agl:
The chair of the WG wasn't happy with treating this as non-substantive w.r.t IPR, so we'll treat this as an issue rather than a pull request and, if we make a change, it'll be in a different issue.

> However, there is no further specification about the case when this results in an empty allowCredentials.

The [algorithm for making a credential](https://w3c.github.io/webauthn/#ref-for-dom-publickeycredentialdescriptor-type%E2%91%A0) does say that if you filter a list to the point that it becomes empty, then that's an error.

But it's perhaps worth noting here too. I've drawn up #1971 to do that.

agl:
(Moving this to #1971.)

PR #1972, opened 2023-09-22
Make the default value for the attestation member in assertion options be null

### LINKED ISSUE #1941: The default value of `attestation` member in `PublicKeyCredentialRequestOptions` should be null or must not have default value
## Proposed Change

[In section 5.5](https://w3c.github.io/webauthn/#dictionary-assertion-options), `PublicKeyCredentialRequestOptions` dictionary is defined.
Although asking attestation in the assertions is optional, the default value of `attestation` member is `none` attestation conveyance preference.
It means that if the relying party does not set any attestation conveyance preference (do not ask attestation in assertion), the browser (webauthn client) will  set the value of `attestation` member as default value `none`, which will eventually might return `attestationObject` with none attestation statement format.

I'm thinking that attestation in assertion is optional and if there is no explicit request, the attestation conveyance preference must be interpreted as I don't need any attestation for assertion which is not identical to the the `none` attestation.

So, the value of `attestation` member should not be defaulting to `none`, and the absence should be treated as I don't need any attestation for assertion and give me conventional assertion for credential get operation.

Kieun:
The default value of attestation options in **assertion** was `none` and the intension was that the relying party does not want any attestation at assertion time.  In this case, the user agent and authenticator does not return any `attestationObject` which is backward compatible.
But, for `none` attestation options in **attestation** **does** always return attestation statement including `none` attestation statement or any other attestation statement with `self` attestation.

Now, with this PR, the default of attestation options in **assertion** is `null`.
So, if the RP does not set attestation option member value, the user agent now handles that value as `null`. In this case, the user agent does not request any attestation to the authenticator by creating the single entry list with `none` for attestationFormats, and it does not return attestationObject as a assertion response.

In other cases, if the RP explicitly sets attestation options including `none`, the user agent now sends such request like the way of attestation processing, and if any authenticator supports such requests, the authenticator may return attestation in assertions. Then, the user agent handles such response from the authenticator with additional steps such as replacing some potentially identifiable information.

Fixes #1941

***
<a href="https://pr-preview.s3.amazonaws.com/Kieun/webauthn/pull/1972.html" title="Last updated on Sep 22, 2023, 10:01 AM UTC (021293c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/1972/baf774a...Kieun:021293c.html" title="Last updated on Sep 22, 2023, 10:01 AM UTC (021293c)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2677,7 +2677,7 @@ The steps for [=issuing a credential request to an authenticator=] are as follow
 
             <dl class="switch">
 
-                :   is set to {{AttestationConveyancePreference/none}}
+                :   is set to null
                 ::  Set |attestationFormats| be the single-element list containing the string &ldquo;none&rdquo;
 
             </dl>
--- a/index.bs
+++ b/index.bs
@@ -2503,6 +2503,9 @@ When this method is invoked, the user agent MUST execute the following algorithm
                 ::  If the [=authenticator=] returned an [=attestation=], set the value of [=assertionAttestation=] to be the bytes of
                     the [=attestation statement=]. Otherwise set it to null.
 
+                :   <code><dfn for="assertionCreationData">attestationConveyancePreferenceOption</dfn></code>
+                ::  whose value is the value of |pkOptions|.{{PublicKeyCredentialRequestOptions/attestation}}.
+
                 :   <code><dfn for="assertionCreationData">clientExtensionResults</dfn></code>
                 ::  whose value is an {{AuthenticationExtensionsClientOutputs}} object containing [=extension identifier=] →
                     [=client extension output=] entries. The entries are created by running each extension's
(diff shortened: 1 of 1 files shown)


agl (changes requested):
For assertion-time attestation, specifying the formats as `["none"]` to the authenticator tells it [not to return](https://fidoalliance.org/specs/fido-v2.2-rd-20230321/fido-client-to-authenticator-protocol-v2.2-rd-20230321.html#authenticatorGetAssertion) an attestation.

Kieun:
> For assertion-time attestation, specifying the formats as `["none"]` to the authenticator tells it [not to return](https://fidoalliance.org/specs/fido-v2.2-rd-20230321/fido-client-to-authenticator-protocol-v2.2-rd-20230321.html#authenticatorGetAssertion) an attestation.

Indeed. While digging CTAP 2.2, I found that make credential operation also handles attestation format as well. 
@agl as far as I understand, there was no such option so that authenticator always returns the attestation for create operation. Such modification is intentional? @dwaite , @ve7jtb 

Due to such modification, the WebAuthn L3 also sends attestationFormats as ["none"] to the authenticator if the attestation options is `none`.  I think this also needs to reconsider since only CTAP 2.2+ can handle attestationFormats for create.

Kieun:
> For assertion-time attestation, specifying the formats as `["none"]` to the authenticator tells it [not to return](https://fidoalliance.org/specs/fido-v2.2-rd-20230321/fido-client-to-authenticator-protocol-v2.2-rd-20230321.html#authenticatorGetAssertion) an attestation.

@agl I've addressed your suggestion in the following commit: a763f96fa3bf0906846e7f384eff9c2cda9cc428

emlun (changes requested):
Why does `attestation: null` need to be distinct from `attestation: "none"`? It is already the case that the authenticator will not return an attestation statement if `attestation: "none"` ([step 13 of `authenticatorGetAssertion`](https://w3c.github.io/webauthn/#ref-for-attestation-object%E2%91%A0%E2%91%A1)), and if the authenticator does not return an attestation statement, the client will in turn set `AuthenticatorAssertionResponse.attestationObject` to null ([success branch of step 20 in `navigator.credentials.get()`](https://w3c.github.io/webauthn/#assertioncreationdata-assertionattestation)).

In #1941 you note:
>2. Set `attestation` as `none`: meaning is that I don't have any interests to verify the attestation, but if the authenticator could return `self` attestation (self-signed), it would be better than nothing.

But I don't think this is accurate. Self attestation should not be returned for `attestation: "none"` per the references above.

MasterKale:
From WG meeting @ 12/13: This is made moot by https://github.com/w3c/webauthn/pull/1997

MasterKale:
From WG meeting @ Jan 10: We've merged #1997 so this is no longer needed.

PR #1988, opened 2023-10-12
Allow use of credProps extension during auth

MasterKale:
This PR updates the `credProps` extension to be something that can be used during authentication as well.

During discussion of #1880 the idea came up that if credProps were available during auth too then, in the future when passkeys can be moved between providers, RP's might benefit from receiving a new credential nickname after the user migrates providers but a new entry in `supplementalPubKeys` (#1957) indicates to the RP that the credential's provider changed.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/pull/1988.html" title="Last updated on Nov 15, 2023, 7:03 PM UTC (22aa6ff)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/1988/28d90b2...22aa6ff.html" title="Last updated on Nov 15, 2023, 7:03 PM UTC (22aa6ff)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -952,7 +952,7 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
     and [=assertion=].
 
     A [=[WAA]=] could be a [=roaming authenticator=], a dedicated hardware subsystem integrated into the [=client device=],
-    or a software component of the [=client=] or [=client device=]. A [=[WAA]=] is not necessarily confined to operating in 
+    or a software component of the [=client=] or [=client device=]. A [=[WAA]=] is not necessarily confined to operating in
     a local context, and can generate or store a [=credential key pair=] in a server outside of [=client-side=] hardware.
 
     In general, an [=authenticator=] is assumed to have only one user.
@@ -4109,11 +4109,11 @@ considered more trustworthy than the rest of the authenticator.
 Each authenticator stores a <dfn for=authenticator>credentials map</dfn>, a [=map=] from ([=rpId=], [=public key credential source/userHandle=]) to
 [=public key credential source=].
 
-Additionally, each authenticator has an Authenticator Attestation Globally Unique Identifier or <dfn>AAGUID</dfn>, which is a 128-bit identifier 
-indicating the type (e.g. make and model) of the authenticator. The AAGUID MUST be chosen by its maker to be identical across all substantially identical 
-authenticators made by that maker, and different (with high probability) from the AAGUIDs of all other types of authenticators. The AAGUID for a given type 
-of authenticator SHOULD be randomly generated to ensure this. The [=[RP]=] MAY use the AAGUID to infer certain properties of the authenticator, such as 
-certification level and strength of key protection, using information from other sources. The [=[RP]=] MAY use the AAGUID to attempt to identify the maker of 
+Additionally, each authenticator has an Authenticator Attestation Globally Unique Identifier or <dfn>AAGUID</dfn>, which is a 128-bit identifier
+indicating the type (e.g. make and model) of the authenticator. The AAGUID MUST be chosen by its maker to be identical across all substantially identical
+authenticators made by that maker, and different (with high probability) from the AAGUIDs of all other types of authenticators. The AAGUID for a given type
+of authenticator SHOULD be randomly generated to ensure this. The [=[RP]=] MAY use the AAGUID to infer certain properties of the authenticator, such as
+certification level and strength of key protection, using information from other sources. The [=[RP]=] MAY use the AAGUID to attempt to identify the maker of
 the authenticator without requesting and verifying [=attestation=], but the AAGUID is not provably authentic without [=attestation=].
 
 The primary function of the authenticator is to provide [=WebAuthn signatures=], which are bound to various contextual data. These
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -6797,9 +6797,7 @@ This [=client extension|client=] [=registration extension=] facilitates reportin
 ::  None, other than to report on credential properties in the output.
 
 : Client extension output
-:: [=map/Set=] <code>[=credentialCreationData/clientExtensionResults=]["{{AuthenticationExtensionsClientOutputs/credProps}}"]["rk"]</code> to the value of the |requireResidentKey| parameter that was used in the <a href='#CreateCred-InvokeAuthnrMakeCred'>invocation</a> of the [=authenticatorMakeCredential=] operation.
```
emlun, line 6800: Perhaps it's worth keeping this as a recommendation for what to do during a registration ceremony? It should probably move into the **rk** section below in that case.

index.bs
```diff
@@ -6826,8 +6824,8 @@ This [=client extension|client=] [=registration extension=] facilitates reportin
         ::  This OPTIONAL property is a [=human palatability|human-palatable=] description of the credential's [=managing authenticator=],
             chosen by the user.
 
-            The [=client=] MUST allow the user to choose this value,
-            MAY or MAY not present that choice during [=registration ceremonies=],
+            During [=registration ceremonies=] the [=client=] MUST allow the user to choose this value,
+            MAY or MAY not present that choice,
             and MAY reuse the same value for multiple credentials with the same [=managing authenticator=] across multiple [=[RPS]=].
```
emlun, line 6829: "MUST allow the user to choose this value, MAY or MAY not present that choice [at all]" sounds internally inconsistent to me - how could the user choose if never presented with a choice? Perhaps we should rephrase this a bit to make the intended meaning a little clearer, how about this?

```suggestion
            The [=client=] MUST allow the user to choose this value.
            That choice MAY be presented during the [=registration ceremony|registration=] or [=authentication ceremony=]
            or MAY be made available outside the ceremony, for example in client settings.
            The [=client=] MAY reuse the same value for multiple credentials with the same [=managing authenticator=] across multiple [=[RPS]=].
```

index.bs
```diff
@@ -6826,8 +6824,8 @@ This [=client extension|client=] [=registration extension=] facilitates reportin
...
```
emlun, line 6830: Below this, the final paragraph about `[$credential record/authenticatorDisplayName$]` only applies to registration ceremonies. We should add something to cover authentication ceremonies too. How about this?

```diff
             If the [=[RP]=] includes an <code>[$credential record/authenticatorDisplayName$]</code> [=struct/item=] in [=credential records=],
             the [=[RP]=] MAY offer this value, if present,
             as a default value for the <code>[$credential record/authenticatorDisplayName$]</code> of the new [=credential record=].
+            If the {{authenticatorDisplayName}} extension output from an [=authentication ceremony=]
+            is different from the <code>[$credential record/authenticatorDisplayName$]</code> of the [=credential record=],
+            the [=[RP]=] MAY offer the user to update the <code>[$credential record/authenticatorDisplayName$]</code> of the [=credential record=].
```

Unless `authenticatorDisplayName` should be forbidden for assertions? But I can imagine it being useful for assertions too.

index.bs
MasterKale: After considering your feedback and re-reading the line I removed, I kinda want to move this into the "Client extension processing" section above as it reads like instructions for the client.

For developers I suggest that the existing explanation of `credProps.rk` is sufficient...

> If [rk](http://localhost:8000/#dom-credentialpropertiesoutput-rk) is true, the credential is a [discoverable credential](http://localhost:8000/#discoverable-credential). If [rk](http://localhost:8000/#dom-credentialpropertiesoutput-rk) is false, the credential is a [server-side credential](http://localhost:8000/#server-side-credential). If [rk](http://localhost:8000/#dom-credentialpropertiesoutput-rk) is not present, it is not known whether the credential is a [discoverable credential](http://localhost:8000/#discoverable-credential) or a [server-side credential](http://localhost:8000/#server-side-credential).

...And removing this line allows for the overall page structure of "Client extension output" for `credProps` to match the structure of this same section in the entries for `prf` and `largeBlob`.

What are your thoughts on this? 🤔

index.bs
MasterKale: I like that wording a lot better, I've updated this part.

index.bs
emlun: Sounds fair, I think? Please go ahead, and let's see how that looks as a whole.

index.bs
MasterKale: That's a good suggestion, I've added something specific for post-auth next steps for an RP to react to a different value during auth.

> Unless `authenticatorDisplayName` should be forbidden for assertions? But I can imagine it being useful for assertions too.

That was the intention of this PR, extending `authenticatorDisplayName`'s use for `.get()` calls too. Maybe I'm missing something about your comment here?

index.bs
MasterKale: I've gone ahead and restored this line to the "Client extension processing" section for our consideration.

index.bs
emlun: > Maybe I'm missing something about your comment here?

Sorry, you're not. I too don't think it would make sense to forbid `authenticatorDisplayName` for assertions. :laughing:

agl (approved):

timcappalli (approved):

index.bs
```diff
@@ -6794,12 +6794,13 @@ This [=client extension|client=] [=registration extension=] facilitates reportin
     </xmp>
 
 : Client extension processing
-::  None, other than to report on credential properties in the output.
+::  [=map/Set=] <code>[=credentialCreationData/clientExtensionResults=]["{{AuthenticationExtensionsClientOutputs/credProps}}"]["rk"]</code>
+    to the value of the |requireResidentKey| parameter that was used in the
+    <a href='#CreateCred-InvokeAuthnrMakeCred'>invocation</a> of the [=authenticatorMakeCredential=]
+    operation.
```
emlun, line 6800: I think these steps should also include setting `authenticatorDisplayName`? See meta-PR #1995 for that and another bit of polish I noticed.

emlun (changes requested):

index.bs
MasterKale: Thank you for helping flesh out the client processing rules, I've merged your PR into this one! 🙏

MasterKale:
@emlun This is ready for your review again. There doesn't seem to be anything actionable anymore for me atm so maybe we can merge this today...

pascoej (approved):

akshayku (approved):

emlun (approved):

PR #1992, opened 2023-11-01
Add note about typical autocomplete combos for conditional UI

### LINKED ISSUE #1982: Spec is not specific enough about order of conditional UI autofill tokens
## Proposed Change

I've learned recently that Chrome expects the `webauthn` autofill token to either appear by itself, or **after** the typical `username` or `password` tokens. However based on current browser behavior out in the wild this isn't obvious (see screenshots below.)

If you just read the WebAuthn spec itself, there's no obvious order suggested - the presence of the `webauthn` token _somewhere_ sounds like that's all that matters:

> If options.[mediation](https://w3c.github.io/webappsec-credential-management/#dom-credentialrequestoptions-mediation) is [conditional](https://w3c.github.io/webappsec-credential-management/#dom-credentialmediationrequirement-conditional) and the user interacts with an [input](https://html.spec.whatwg.org/multipage/input.html#the-input-element) or [textarea](https://html.spec.whatwg.org/multipage/form-elements.html#the-textarea-element) form control with an [autocomplete](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#attr-fe-autocomplete) attribute whose value contains a "webauthn" [autofill detail token](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#autofill-detail-tokens),

https://w3c.github.io/webauthn/#GetAssn-ConditionalMediation-Interact-FormControl

But I've been informed that in the WHATWG there's text that specifically says that the `webauthn` token must come after either `username` or ~`password`~ `current-password`:

> 3. Either of the following two options...:
>   - ...
>   - `username`
>   - `current-password`
>   - ...
> 4. Optionally, a token that is an [ASCII case-insensitive](https://infra.spec.whatwg.org/#ascii-case-insensitive) match for the string "webauthn", meaning the user agent should show [public key credentials](https://w3c.github.io/webauthn/#public-key-credential) available via [conditional](https://w3c.github.io/webappsec-credential-management/#dom-credentialmediationrequirement-conditional) mediation when the user interacts with the form control. [webauthn](https://html.spec.whatwg.org/#attr-fe-autocomplete-webauthn) is only valid for [input](https://html.spec.whatwg.org/#the-input-element) and [textarea](https://html.spec.whatwg.org/#the-textarea-element) elements.

https://html.spec.whatwg.org/#autofill-detail-tokens

I think we need to clarify in the spec that there _is_ an order for the tokens that must be followed, and either spell out the order in WebAuthn, or emphasize more that the exact order can be confirmed by looking at that autofill detail tokens section of WHATWG.

Personally I'm proposing we address this by **specifically stating the order in the WebAuthn spec** (with a link to WHATWG for those who wish to dive deeper) to make the WebAuthn spec the one place most developers need to go to understand how to properly set up conditional UI.

## autocomplete="username webauthn"

**Chrome**

Screenshot 2023-10-02 at 8 28 01 AM

**Safari**

Screenshot 2023-10-02 at 8 28 17 AM

## autocomplete="webauthn username"

**Chrome (no autofill appears)**

Screenshot 2023-10-02 at 8 24 49 AM

**Safari**

Screenshot 2023-10-02 at 8 24 30 AM

MasterKale:
This PR includes a new note detailing two simple examples of "correct" `autocomplete` attributes that should reliably trigger conditional UI across most browsers.

Fixes #1982.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/pull/1992.html" title="Last updated on Nov 15, 2023, 8:19 PM UTC (139405f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/1992/28d90b2...139405f.html" title="Last updated on Nov 15, 2023, 8:19 PM UTC (139405f)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2413,7 +2413,7 @@ When this method is invoked, the user agent MUST execute the following algorithm
                 </dt>
         ::  <div class="note">
                 Note: The `"webauthn"` [=autofill detail token=] must appear immediately after the last [=autofill detail token=]
-                of type "Normal" or "Contact":
+                of type "Normal" or "Contact". For example:
 
                 - `"username webauthn"`
                 - `"current-password webauthn"`
--- a/index.bs
+++ b/index.bs
@@ -172,6 +172,7 @@ spec: html; urlPrefix: https://html.spec.whatwg.org/multipage/
             text: document.domain; url:dom-document-domain
         urlPrefix: form-control-infrastructure.html
             text: autofill detail token; url: autofill-detail-tokens
+            text: non-autofill credential type; url: non-autofill-credential-type
 
 spec: url; urlPrefix: https://url.spec.whatwg.org
     type: dfn
--- a/index.bs
+++ b/index.bs
@@ -2410,8 +2410,8 @@ When this method is invoked, the user agent MUST execute the following algorithm
                   and the user interacts with an <{input}> or <{textarea}> form control with an <{input/autocomplete}> attribute whose value
                   contains `"webauthn"` as the last [=autofill detail token=],
                 </dt>
-            <div class="note">
-                Note: The following are examples of typical values for the <{input/autocomplete}> attribute with the `"webauthn"`
+        ::  <div class="note">
+                Note: The following are example values for the <{input/autocomplete}> attribute with the `"webauthn"`
                 [=autofill detail token=] in the correct position:
 
                 - `"username webauthn"`
@@ -2421,7 +2421,7 @@ When this method is invoked, the user agent MUST execute the following algorithm
                 of these tokens can be confirmed in the [=autofill detail token=] section of the the WHATWG HTML standard.
             </div>
 
-        ::  1. If |silentlyDiscoveredCredentials| is not [=list/empty=]:
+            1. If |silentlyDiscoveredCredentials| is not [=list/empty=]:
 
                 1. Prompt the user to optionally select a [=DiscoverableCredentialMetadata=] (|credentialMetadata|) from |silentlyDiscoveredCredentials|.
 
--- a/index.bs
+++ b/index.bs
@@ -2408,8 +2408,18 @@ When this method is invoked, the user agent MUST execute the following algorithm
                 <dt id="GetAssn-ConditionalMediation-Interact-FormControl">
                   If <code>|options|.{{CredentialRequestOptions/mediation}}</code> is {{CredentialMediationRequirement/conditional}}
                   and the user interacts with an <{input}> or <{textarea}> form control with an <{input/autocomplete}> attribute whose value
-                  contains a `"webauthn"` [=autofill detail token=],
+                  contains `"webauthn"` as the last [=autofill detail token=],
                 </dt>
+            <div class="note">
+                Note: The following are examples of typical values for the <{input/autocomplete}> attribute with the `"webauthn"`
+                [=autofill detail token=] in the correct position:
+
+                - `"username webauthn"`
+                - `"current-password webauthn"`
+
+                These examples are likely to trigger conditional UI across the greatest number of user agents. The expected ordering
+                of these tokens can be confirmed in the [=autofill detail token=] section of the the WHATWG HTML standard.
+            </div>
 
         ::  1. If |silentlyDiscoveredCredentials| is not [=list/empty=]:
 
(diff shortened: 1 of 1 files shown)


MasterKale:
I'm not a huge fan of how the new Note renders, but I couldn't get it to indent one level closer to the left without tripping CI's HTML validator which complained that I tried to put a `<div>` inside a `<dt>`. The way it renders now CI is happy since the `<div>` is inside the accompanying `<dd>` that contains the subsequent logic block, "If silentlyDiscoveredCredentials is not empty":

Screenshot 2023-10-31 at 10 17 31 PM

timcappalli (approved):
LGTM

arianvp:
drive-by question: Why does the `webauthn` token need to be at the end? This sounds like something many people will get wrong and I think most people don't read w3c specs but e.g. MDN.

index.bs
```diff
@@ -2408,10 +2408,20 @@ When this method is invoked, the user agent MUST execute the following algorithm
                 <dt id="GetAssn-ConditionalMediation-Interact-FormControl">
                   If <code>|options|.{{CredentialRequestOptions/mediation}}</code> is {{CredentialMediationRequirement/conditional}}
                   and the user interacts with an <{input}> or <{textarea}> form control with an <{input/autocomplete}> attribute whose value
-                  contains a `"webauthn"` [=autofill detail token=],
+                  contains `"webauthn"` as the last [=autofill detail token=],
```
nsatragno, line 2411: Generally happy with this and especially the example, but I'm a little worried that the autofill section of html might update and then this section will become stale. Also, `webauthn` is not always a valid token, even if at the end (e.g. you cannot do `autocomplete="off webauthn"`). How about

```
...<{input/autocomplete}> whose [=non-autofill credential type=] is `"webauthn"`.
```

https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#non-autofill-credential-type

Which essentially delegates all parsing to html.

And then, for the note to make RPs lives easier:

```
Note: The "webauthn" autofill detail token must appear immediately after the last autofill detail token
of type "Normal" or "Contact".

[examples]

These examples are likely to trigger conditional UI across the greatest number of user agents. The expected ordering
of these tokens can be confirmed in the [=autofill detail token=] section of the the WHATWG HTML standard.
```

nsatragno (changes requested):

index.bs
emlun: I agree with @nsatragno's suggestions above.

I also don't think the "These examples are likely to trigger..." paragraph is needed, it should be obvious enough that that is the purpose of the examples. I would cut this paragraph to make the size of the note less intrusive.

MasterKale: @nsatragno Is `[=non-autofill credential type=]` supposed to link anywhere? Bikeshed doesn't seem to recognize it:

Screenshot 2023-11-15 at 11 16 26 AM

nsatragno: uhm I did not try the markup I proposed :S

It should ideally link here: https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#non-autofill-credential-type

MasterKale: Thanks for confirming, I figured out how to get it to link 🎉 

Screenshot 2023-11-15 at 12 09 01 PM

MasterKale: I've addressed these changes, marking resolved.

nsatragno (approved):
Thank you! Approved.

index.bs
```diff
@@ -2407,11 +2408,18 @@ When this method is invoked, the user agent MUST execute the following algorithm
...
+                  and the user interacts with an <{input}> or <{textarea}> form control with an <{input/autocomplete}> attribute whose
+                  [=non-autofill credential type=] is `"webauthn"`,
                 </dt>
+        ::  <div class="note">
+                Note: The `"webauthn"` [=autofill detail token=] must appear immediately after the last [=autofill detail token=]
+                of type "Normal" or "Contact". For example:
 
-        ::  1. If |silentlyDiscoveredCredentials| is not [=list/empty=]:
+                - `"username webauthn"`
+                - `"current-password webauthn"`
```
emlun, line 2419: I would add `autofill=` to make the examples that tiny bit clearer, but this is fine as is too.

```suggestion
                - `autofill="username webauthn"`
                - `autofill="current-password webauthn"`
```

emlun (approved):

emlun:
If I understood the HTML spec right it seems like `autofill="webauthn"` alone should also work (this will set the _non-autofill credential type_ since it jumps to the _Done_ step), is that right @nsatragno? If so, should we point that out too?

nsatragno:
Yes, that should work too. Either pointing it out or not sounds good to me.

MasterKale:
> If I understood the HTML spec right it seems like `autofill="webauthn"` alone should also work (this will set the _non-autofill credential type_ since it jumps to the _Done_ step), is that right @nsatragno? If so, should we point that out too?

I think it'd be better to encourage here more use of proper autofill tokens so I'm inclined to not call this otherwise valid combination out...

emlun:
Sounds fair. :slightly_smiling_face:

PR #1920, opened 2023-07-08
Add missing fields in dictionaries for JSON representation

Kieun:
Closes #1918

This pr is re-opened.

Also found that attestationFormats field is missing in PublicKeyCredentialRequestOptionsJSON as well, this issue is also addressed with this pr.

Issue #1818, opened 2022-10-22
Is there a way to store user secret key in the authenticator with/without an extension?


KuznetsovNikita:
The webAuthn will be very useful if developers could have an option to store and get user secret key, from my perspective.

As I understand the developers have a few options:

1) [`user.id` aka `userHandle`](https://w3c.github.io/webauthn/#user-handle) with is supported for most of the authenticator.

2) The ["small blob" extension](https://fidoalliance.org/specs/fido-v2.1-ps-20210615/fido-client-to-authenticator-protocol-v2.1-ps-20210615.html#sctn-credBlob-extension) with is not supported well in nowadays.

3) The ["large blob" extension](https://www.w3.org/TR/webauthn-2/#sctn-large-blob-extension) with is an experimental feature in nowadays. 

As I understand the value for `user.id` aka `userHandle` is stored in the authenticator, but means non sensitive data and used for technical reason.  

Why I should not store user secret in the `user.id` aka `userHandle`? 
Is it possible to get a list of `user.ids` from authenticator programmatically?
Is it possible some how hack and get this value from authenticator? 
Is the hack required physical contact with authenticator or could it be done programmatically? 

Thank you in advance!

timcappalli:
Hi! Please use [FIDO-DEV](https://groups.google.com/a/fidoalliance.org/g/fido-dev?pli=1) for implementation questions/discussions.

![Screenshot_20221022-072220.png](https://user-images.githubusercontent.com/6476604/197344345-cce70c1e-0209-41ff-8dcd-61fb9157453b.png)

emlun:
On the WebAuthn layer, the small and large blob extensions are indeed the only way to store a chosen value. As you note you could technically abuse `user.id` for this as well, but that is strongly discouraged because authenticators and client platforms make assumptions about what it's used for and do not treat it as sensitive.

If a random but deterministically derived value is sufficient - for example, to use as key derivation material rather than a finished key - there is the [`prf` extension](https://w3c.github.io/webauthn/#prf-extension), though I don't think there are any client implementations available yet.

emlun:
Closing this, please let us know if there's discussion left to have here.

PR #1812, opened 2022-10-05
Use credential record abstraction in devicePubKey extension

emlun:
This applies the new abstraction from #1773 and #1807 to the `devicePubKey` extension, which can substantially streamline the RP processing steps for the new extension.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/pull/1812.html" title="Last updated on Jan 13, 2023, 4:43 PM UTC (67dc54b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/1812/33e7507...67dc54b.html" title="Last updated on Jan 13, 2023, 4:43 PM UTC (67dc54b)">Diff</a>

PR #1997, opened 2023-11-15
Drop assertion-time attestation.

agl:
We don't believe that there's a use for it now. The cases that wish to plumb this sort of data back can do so via other means.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/pull/1997.html" title="Last updated on Jan 10, 2024, 12:44 PM UTC (a0cfb6c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/1997/a83c764...a0cfb6c.html" title="Last updated on Jan 10, 2024, 12:44 PM UTC (a0cfb6c)">Diff</a>


rlin1:
Does that imply that device-bound keys in the "supplemental keys" extension cannot be attested?

agl:
> Does that imply that device-bound keys in the "supplemental keys" extension cannot be attested?

No, they certainly can be. This change just reduces the number of places that can have attestations at assertion time from three to two.

arianvp:
Big 👍 for this change.   Attestation should be stored at creation time and this simplifies the spec.  for DBKs creation time can overlap with assertion time of the main credential so it's still during "creation".

index.bs
```diff
@@ -5057,24 +4996,13 @@ The privacy, security and operational characteristics of [=attestation=] depend
     operating environment, and so on.
 
 The [=attestation type=] and [=attestation statement format=] is chosen by the [=authenticator=];
-[=[RPS]=] can only signal their preferences by setting the {{PublicKeyCredentialCreationOptions/attestation}} and {{PublicKeyCredentialCreationOptions/attestationFormats}} parameters, or those with the same names in {{PublicKeyCredentialRequestOptions}}.
+[=[RPS]=] can only signal their preferences by setting the {{PublicKeyCredentialCreationOptions/attestation}} and {{PublicKeyCredentialCreationOptions/attestationFormats}} parameters.
```
emlun, line 4999: Above this, [near the beginning of the "Attestation" section](https://github.com/w3c/webauthn/pull/1997/files#diff-5e793325cd2bfc452e268a4aa2f02b4024dd9584bd1db3c2595f61f1ecf7b985L5013), remains some language about "attestation is returned when a credential is created, and optionally exercised". See the diff that introduced it: https://github.com/w3c/webauthn/pull/1663/commits/20dd35c00a4caf07b1514c5495afaaf1482e04ae#diff-5e793325cd2bfc452e268a4aa2f02b4024dd9584bd1db3c2595f61f1ecf7b985L4308-R4358

index.bs
```diff
@@ -2710,23 +2676,19 @@ The steps for [=issuing a credential request to an authenticator=] are as follow
...
-                            |enterpriseAttestationPossible|, |attestationFormats|, and |authenticatorExtensions| as parameters.
+                            and |authenticatorExtensions| as parameters.
                     </dl>
 
             :   [=list/is empty=]
             ::  Using local configuration knowledge of the appropriate transport to use with |authenticator|, invoke the
                 [=authenticatorGetAssertion=] operation on |authenticator| with |rpId|, |clientDataHash|,
                 |userVerification|,
-                |enterpriseAttestationPossible|,
-                |attestationFormats|,
```
emlun, line 2729: These need to be deleted from the [authenticatorGetAssertion parameters](https://github.com/w3c/webauthn/pull/1997/files#diff-5e793325cd2bfc452e268a4aa2f02b4024dd9584bd1db3c2595f61f1ecf7b985R4736-R4739) too.

index.bs
emlun: Similarly, I think we can revert the change to the `<figcaption>` in that section too: https://github.com/w3c/webauthn/pull/1663/commits/20dd35c00a4caf07b1514c5495afaaf1482e04ae#diff-5e793325cd2bfc452e268a4aa2f02b4024dd9584bd1db3c2595f61f1ecf7b985L4319-R4368

emlun (changes requested):

MasterKale (approved):

index.bs
```diff
@@ -3026,7 +2986,6 @@ optionally evidence of [=user consent=] to a specific transaction.
         [SameObject] readonly attribute ArrayBuffer      authenticatorData;
         [SameObject] readonly attribute ArrayBuffer      signature;
         [SameObject] readonly attribute ArrayBuffer?     userHandle;
-        [SameObject] readonly attribute ArrayBuffer?     attestationObject;
```
MasterKale, line 3029: This will need to be dropped from `AuthenticatorAssertionResponseJSON` as well (it was added there too back in #1920)

MasterKale, line 3029: This will need to be dropped from `AuthenticatorAssertionResponseJSON` as well (it was added there too back in #1920)

MasterKale (changes requested):

index.bs
agl: Thanks. Done.

index.bs
agl: Yep, and more needs to be tidied up from there too. Done.

index.bs
agl: Done.

MasterKale (approved):
Looks good to me.

emlun (approved):
Heh, I spotted this looking strange in the HTML diff:

screenshot-2024-01-10T13:37:43+01:00

This was added in [meta-PR #1818](https://github.com/w3c/webauthn/pull/1812/files#diff-5e793325cd2bfc452e268a4aa2f02b4024dd9584bd1db3c2595f61f1ecf7b985R5508-R5512), and apparently Bikeshed now started pulling in `{{AuthenticatorAssertionResponse/attestationObject}}` as an external ref from the spec database since the member definition no longer exists within the document.

I've pushed commit a0cfb6c897df6fc85c939062c9d46ecb7ac8c302 which fixes this.

Issue #1987, opened 2023-10-12
Virtual Authenticator API does not expose a way to set backup-eligible or backup-status flags


MasterKale:
## Proposed Change

The current state of the Virtual Authenticators API does not allow users of the API any way to influence the `be` and `bs` flags in authenticator data. This makes it difficult to use this API for E2E tests that want to test "synced passkey" scenarios.

We should update this API to add in new parameters to the [Add Virtual Authenticator](https://w3c.github.io/webauthn/#sctn-automation-add-virtual-authenticator) API call payload that can make it possible to set `be` and `bs`, a la the current `isUserConsenting` and `isUserVerified` parameters.

emlun:
I think the [Add Credential](https://w3c.github.io/webauthn/#sctn-automation-add-credential) API call would be the be more appropriate place for these, but I agree otherwise.

I also don't think we need to define any new operations to enable changing `bs` between assertions - since _Add Credential_ includes all credential state (including the private key and signature counter), the caller can "update" `bs` by first calling [Remove Credential](https://w3c.github.io/webauthn/#sctn-automation-remove-credential) and then calling _Add Credential_ with the new `bs` value.

MasterKale:
> I think the [Add Credential](https://w3c.github.io/webauthn/#sctn-automation-add-credential) API call would be the be more appropriate place for these, but I agree otherwise.

I honestly agree with you, as `be` and `bs` are definitely credential properties vs authenticator properties. However the current means of manipulating flags in `authData` are centralized in the Add Virtual Authenticator API call as `isUserConsenting` and `isUserVerified` so that's why I suggested the authenticator API instead...

Actually no, now I agree 100%, the means of specifying `be` and `bs` for a credential **should** be on Add Credential.

> ...the caller can "update" bs by first calling [Remove Credential](https://w3c.github.io/webauthn/#sctn-automation-remove-credential) and then calling Add Credential with the new bs value.

I think I can get behind this, but we _do_ have precedent for an "update" API [Set User Verified](https://w3c.github.io/webauthn/#sctn-automation-set-user-verified). Maybe we can add a similar "Set Backup Status" API for individual credentials.

agl:
@nsatragno

nsatragno:
If we put only the option on `AddCredential`, then how would you test your website's behaviour on credential creation?

Here's a proposal:
[...]

PR #1999, opened 2023-11-20
Add backup flags to virtual authenticator

nsatragno:
Allow setting and changing the backup eligibility (BE) and backup state (BS) flags through the virtual authenticator API.

Fixed: #1987

***
<a href="https://pr-preview.s3.amazonaws.com/nsatragno/webauthn/pull/1999.html" title="Last updated on Nov 22, 2023, 6:16 PM UTC (b0c79de)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/1999/f8163ea...nsatragno:b0c79de.html" title="Last updated on Nov 22, 2023, 6:16 PM UTC (b0c79de)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -7765,9 +7765,13 @@ Each stored [=virtual authenticator=] has the following properties:
 
     Note: This property has no effect if the [=Virtual Authenticator=] does not support the [=User Verification Method=] extension.
 : |defaultBackupEligibility|
-:: Determines the default state of the [=backup eligibility=] [=flag=] ([=BE=]) for any newly created [=Public Key Credential Source=].
+:: Determines the default state of the [=backup eligibility=] [=credential property=] for any newly created [=Public Key Credential Source=].
+    This value MUST be reflected by the [=BE=] [=authenticator data=] [=flag=] when performing an [=authenticatorMakeCredential=]
+    operation with this [=virtual authenticator=].
 : |defaultBackupState|
-:: Determines the default state of the [=backup state=] [=flag=] ([=BS=]) for any newly created [=Public Key Credential Source=].
+:: Determines the default state of the [=backup state=] [=credential property=] for any newly created [=Public Key Credential Source=].
+    This value MUST be reflected by the [=BS=] [=authenticator data=] [=flag=] when performing an [=authenticatorMakeCredential=]
+    operation with this [=virtual authenticator=].
 
 ## <dfn>Add Virtual Authenticator</dfn> ## {#sctn-automation-add-virtual-authenticator}
 
@@ -8011,6 +8015,8 @@ The <dfn>Credential Parameters</dfn> is a JSON [=Object=] passed to the [=remote
                 <td>
                     The simulated [=backup eligibility=] for the [=public key credential source=]. If unset, the value will default to the
                     [=virtual authenticator=]'s |defaultBackupEligibility| property.
+                    This value MUST be reflected by the [=BE=] [=authenticator data=] [=flag=] when performing an [=authenticatorGetAssertion=]
+                    operation with this [=public key credential source=].
                 </td>
                 <td>boolean</td>
             </tr>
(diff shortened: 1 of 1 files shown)


MasterKale (approved):
Thank you for taking this on, @nsatragno, this looks great!

The only thing I could think to suggest is that we call it `defaultBackupEligible` and `backupEligible` over `...Eligibility`, but looking at the current state of the spec right now we seem to use "eligibility" more often than "eligible" so I don't think I have a strong case here:

Screenshot 2023-11-21 at 2 15 14 PM

Screenshot 2023-11-21 at 2 15 18 PM

index.bs
```diff
@@ -7764,6 +7764,14 @@ Each stored [=virtual authenticator=] has the following properties:
 :: A {{UvmEntries}} array to be set as the [=authenticator extension output=] when processing the [=User Verification Method=] extension.
 
     Note: This property has no effect if the [=Virtual Authenticator=] does not support the [=User Verification Method=] extension.
+: |defaultBackupEligibility|
+:: Determines the default state of the [=backup eligibility=] [=credential property=] for any newly created [=Public Key Credential Source=].
+    This value SHOULD be reflected by the [=BE=] [=authenticator data=] [=flag=] when performing an [=authenticatorMakeCredential=]
```
emlun, line 7769: Why should these SHOULDs not be MUSTs? (also applies to other occurrences below)

index.bs
```diff
@@ -8171,6 +8217,75 @@ The [=remote end steps=] are:
  1. Set the |authenticator|'s |isUserVerified| property to the |parameters|' |isUserVerified| property.
  1. Return [=success=].
 
+## <dfn>Set Credential Properties</dfn> ## {#sctn-automation-set-credential-properties}
```
emlun, line 8220: Should this be able to set `extensions` and `uvm` too?

index.bs
```diff
@@ -8171,6 +8217,75 @@ The [=remote end steps=] are:
...
+ 1. Let |credential| be the [=public key credential source=] managed by |authenticator| matched by |credentialId|.
+ 1. If |credential| is empty, return a [=WebDriver error=] with [=WebDriver error code=] [=invalid argument=].
+ 1. Let |backupEligibility| be the |parameters|' |backupEligibility| property.
+ 1. If |backupEligibility| is defined, set the [=backup eligibility=] [=credential property=] of |credential| to its value.
+
+    Note: Normally, the |backupEligibility| property is permanent to a [=public key credential source=].
+    [=Set Credential Properties=] allows changing it for testing and debugging purposes.
+
+ 1. Let |backupState| be the |parameters|' |backupState| property.
+ 1. If |backupState| is defined, set the [=backup state=] [=credential property=] of |credential| to its value.
```
emlun, line 8286: I think it's probably worth being a bit more explicit here, to eliminate any doubt of whether "it" refers to `|backupEligibility|` or to `the [=backup eligibility=] [=credential property=] of |credential|`.

```suggestion
 1. If |backupEligibility| is defined, set the [=backup eligibility=] [=credential property=] of |credential| to the value of |backupEligibility|.

    Note: Normally, the |backupEligibility| property is permanent to a [=public key credential source=].
    [=Set Credential Properties=] allows changing it for testing and debugging purposes.

 1. Let |backupState| be the |parameters|' |backupState| property.
 1. If |backupState| is defined, set the [=backup state=] [=credential property=] of |credential| to the value of |backupState|.
```

emlun (commented):
Should the new Set Credential Properties be able to set `extensions` and `uvm` too?

index.bs
emlun: Ah right, what set of `extensions` the authenticator supports shouldn't change between assertions. But the `uvm` response can vary per request, not just per authenticator - for example if an authenticator supports both PIN and fingerprint UV, each assertion (even with the same credential) may report either method in the `uvm` response. So perhaps that should be variable per assertion, so you can test against that behaviour? But yeah, that should be a separate issue/PR in that case.

emlun (approved):
LGTM, thanks!

nsatragno:
+@pascoej for review from webkit.

timcappalli (approved):

pascoej (approved):

PR #2012, opened 2024-01-09
Expand upon the definition of "unsigned extension outputs"

### LINKED ISSUE #1964: Further clarification of Unsigned Extension Outputs
## Proposed Change
The current level 3 drafts introduce a concept of Unsigned Extension Outputs.
It would be helpful if the specification could add a little more detail about what  "Unsigned" means in the context of Extension Outputs? Maybe add a use case for  this type of extension?  

Are there new "unsigned" extensions being added to the iana registry?

agl:
Fixes #1964

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/pull/2012.html" title="Last updated on Jan 10, 2024, 8:30 PM UTC (8f8c7ea)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/2012/a83c764...8f8c7ea.html" title="Last updated on Jan 10, 2024, 8:30 PM UTC (8f8c7ea)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -6573,7 +6573,7 @@ and the corresponding value is the <dfn>authenticator extension output</dfn> for
 as a separate map, keyed with the same [=extension identifier=]. This map only contains entries for authenticator
 extensions that make use of unsigned outputs. Unsigned outputs are useful when extensions output a signature over
 the [=authenticator data=] (because otherwise a signature would have to sign over itself, which isn't possible) or when
-the extension outputs should not be sent to the RP.
+some extension outputs should not be sent to the [=[RP]=].
 
 Note: In [[!FIDO-CTAP]] [=unsigned extension outputs=] are returned as a CBOR map in a top-level field named
 `unsignedExtensionOutputs` from both [=authenticatorMakeCredential=] and [=authenticatorGetAssertion=].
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -6571,7 +6571,9 @@ and the corresponding value is the <dfn>authenticator extension output</dfn> for
 
 <dfn>Unsigned extension outputs</dfn> are represented independently from [=authenticator data=] and returned by authenticators
 as a separate map, keyed with the same [=extension identifier=]. This map only contains entries for authenticator
-extensions that make use of unsigned outputs.
+extensions that make use of unsigned outputs. Unsigned outputs are useful when extensions output a signature over
+the [=authenticator data=] (because otherwise a signature would have to sign over itself, which isn't possible) or when
+the extension outputs should not be sent to the RP.
```
MasterKale, line 6576: Are there any examples currently in the spec of extension output that **should not** be sent to the RP? I've never really considered this as a possibility before.

emlun, line 6576:
```suggestion
the extension outputs should not be sent to the [=[RP]=].
```

emlun: Maybe also rewrite "the outputs" to "some outputs"? (Since one extension may emit both signed and unsigned outputs)

```suggestion
some extension outputs should not be sent to the [=[RP]=].
```

emlun (approved):

index.bs
agl: The PRF extension.

emlun (approved):

index.bs
MasterKale: I didn't appreciate that this is was about an interaction between Client and Authenticator and isn't quite related to e.g. client extension outputs that RPs ultimately see.

PR #2017, opened 2024-01-19
Explicitly specify binary encoding for string truncation

### LINKED ISSUE #1994: rp.name, user.name and user.displayName length limit does not state binary encoding
The definitions of [`PublicKeyCredentialEntity.name`](https://w3c.github.io/webauthn/#dom-publickeycredentialentity-name) and [`PublicKeyCredentialUserEntity.displayName`](https://w3c.github.io/webauthn/#dom-publickeycredentialuserentity-displayname) state that

>Authenticators MAY truncate a [name](https://w3c.github.io/webauthn/#dom-publickeycredentialentity-name) member’s value so that it fits within 64 bytes [...]

and

>[Authenticators](https://w3c.github.io/webauthn/#authenticator) MUST accept and store a 64-byte minimum length for a [displayName](https://w3c.github.io/webauthn/#dom-publickeycredentialuserentity-displayname) member’s value. Authenticators MAY truncate a [displayName](https://w3c.github.io/webauthn/#dom-publickeycredentialuserentity-displayname) member’s value so that it fits within 64 bytes. [...]

but do not state what binary encoding the 64 byte limit applies to. Both reference [§ 6.4.1 String Truncation](https://w3c.github.io/webauthn/#sctn-strings-truncation) which states that

>[...] truncation SHOULD also respect UTF-8 sequence boundaries or [grapheme cluster](https://w3c.github.io/i18n-glossary/#dfn-grapheme-cluster) boundaries [...]

so presumably UTF-8 is intended, but again this is not explicitly stated - just that the truncation should respect UTF-8 considerations. CTAP2 explicitly states that UTF-8 is used, but not all authenticators use CTAP.

In fact IDL [DOMString](https://webidl.spec.whatwg.org/#idl-DOMString)s are explicitly [sequences of 16-bit code units](https://infra.spec.whatwg.org/#string), which are more naturally represented by UTF-16 or UCS-2. So it could be argued that the length limit of `name` and `displayName` varies depending on what encoding the authenticator happens to use. This makes it practically impossible (in theory, even if in practice most probably (?) use UTF-8) for an RP to determine if a user input is likely to be truncated or not.

## Proposed Change

The length limit for [`PublicKeyCredentialEntity.name`](https://w3c.github.io/webauthn/#dom-publickeycredentialentity-name) and [`PublicKeyCredentialUserEntity.displayName`](https://w3c.github.io/webauthn/#dom-publickeycredentialuserentity-displayname) should explicitly state the binary encoding the limit applies to. For example:

>Authenticators MAY truncate a [name](https://w3c.github.io/webauthn/#dom-publickeycredentialentity-name) member’s value so that its UTF-8 encoding fits within 64 bytes [...]

and

>[Authenticators](https://w3c.github.io/webauthn/#authenticator) MUST accept and store a [displayName](https://w3c.github.io/webauthn/#dom-publickeycredentialuserentity-displayname) member’s value whose UTF-8 encoding is 64 bytes or shorter. Authenticators MAY truncate a [displayName](https://w3c.github.io/webauthn/#dom-publickeycredentialuserentity-displayname) member’s value so that its UTF-8 encoding fits within 64 bytes. [...]

emlun:
Fixes #1994.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/pull/2017.html" title="Last updated on Mar 20, 2024, 3:26 PM UTC (ccf4e32)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/2017/82db709...ccf4e32.html" title="Last updated on Mar 20, 2024, 3:26 PM UTC (ccf4e32)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -4923,28 +4923,62 @@ Authenticators may be required to store arbitrary strings chosen by a [=[RP]=], 
 ### String Truncation ### {#sctn-strings-truncation}
 
 Each arbitrary string in the API will have some accommodation for the potentially limited resources available to an [=authenticator=].
-If string truncation is the chosen accommodation,
-then the authenticator MAY choose a size limit equal to or greater than the specified minimum supported length
-and truncate the string so that its length in the UTF-8 character encoding satisfies that limit.
-Such truncation SHOULD also respect UTF-8 code point boundaries or [=grapheme cluster=] boundaries [[UAX29]].
-The resulting truncated value MAY be shorter than the chosen size limit
-but MUST NOT be shorter than the longest prefix substring that satisfies the size limit and ends on a [=grapheme cluster=] boundary.
+When the chosen accommodation is string truncation, care needs to be taken to not corrupt the string value.
 
-For example, in <a href="#fig-stringTruncation">figure <span class="figure-num-following"></span></a> the string is 65 bytes long. If truncating to 64 bytes then the final 0x88 byte must be removed purely because of space reasons. Since that leaves a partial UTF-8 code point the remainder of that code point may also be removed. Since that leaves a partial [=grapheme cluster=] an authenticator may remove the remainder of that cluster.
+For example, truncation based on Unicode code points alone may cause a [=grapheme cluster=] to be truncated.
+This could make the grapheme cluster render as a different glyph,
+potentially changing the meaning of the string, instead of removing the glyph entirely.
+For example, <a href="#fig-stringTruncation">figure <span class="figure-num-following"></span></a>
+shows the end of a UTF-8 encoded string whose encoding is 65 bytes long.
+If truncating to 64 bytes then the final 0x88 byte is removed first to satisfy the size limit.
+Since that leaves a partial UTF-8 code point, the remainder of that code point must also be removed.
+Since that leaves a partial [=grapheme cluster=], the remainder of that cluster should also be removed.
 
 <figure id="fig-stringTruncation">
     <img src="images/string-truncation.svg"></img>
     <figcaption>The end of a UTF-8 encoded string showing the positions of different truncation boundaries.</figcaption>
 </figure>
 
-[=Conforming User Agents=] are responsible for ensuring that the authenticator behavior observed by [=[RPS]=] conforms to this specification with respect to string handling. For example, if an authenticator is known to behave incorrectly when asked to store large strings, the user agent SHOULD perform the truncation for it in order to maintain the model from the point of view of the [=[RP]=]. User-agents that do this SHOULD truncate at [=grapheme cluster=] boundaries.
+The responsibility for handling these concerns falls primarily on the [=client=],
+to avoid burdening [=authenticators=] with understanding character encodings and Unicode character properties.
+The following subsections define requirements for how clients and authenticators,
+respectively, may perform string truncation.
 
-Truncation based on UTF-8 code points alone may cause a [=grapheme cluster=] to be truncated. This could make the grapheme cluster render as a different glyph, potentially changing the meaning of the string, instead of removing the glyph entirely.
 
-In addition to that, truncating on byte boundaries alone causes a known issue that user agents should be aware of: if the authenticator is using [[!FIDO-CTAP]] then future messages from the authenticator may contain invalid CBOR since the value is typed as a CBOR string and thus is required to be valid UTF-8. User agents are tasked with handling this to avoid burdening authenticators with understanding character encodings and Unicode character properties. Thus, when dealing with [=authenticators=], user agents SHOULD:
+#### String Truncation by Clients #### {#sctn-strings-truncation-client}
+
+When a [=[WAC]=] truncates a string,
+the truncation behaviour observable by the [=[RP]=] MUST satisfy the following requirements:
+
+Choose a size limit equal to or greater than the specified minimum supported length.
+The string MAY be truncated so that its length in bytes in the UTF-8 character encoding satisfies that limit.
+This truncation MUST respect UTF-8 code point boundaries, and SHOULD respect [=grapheme cluster=] boundaries [[UAX29]].
+The resulting truncated value MAY be shorter than the chosen size limit
+but MUST NOT be shorter than the longest prefix substring that satisfies the size limit and ends on a [=grapheme cluster=] boundary.
+
+The client MAY let the [=authenticator=] perform the truncation if it satisfies these requirements;
+otherwise the client MUST perform the truncation before relaying the string value to the authenticator.
+
+In addition to the above, truncating on byte boundaries alone causes a known issue that user agents should be aware of: if the authenticator is using [[!FIDO-CTAP]] then future messages from the authenticator may contain invalid CBOR since the value is typed as a CBOR string and thus is required to be valid UTF-8. Thus, when dealing with [=authenticators=], user agents SHOULD:
 
 1. Ensure that any strings sent to authenticators are validly encoded.
 1. Handle the case where strings have been truncated resulting in an invalid encoding. For example, any partial code point at the end may be dropped or replaced with [U+FFFD](http://unicode.org/cldr/utility/character.jsp?a=FFFD).
+
+
+#### String Truncation by Authenticators #### {#sctn-strings-truncation-authenticator}
+
+Because a [=[WAA]=] may be implemented in a constrained environment,
+the requirements on authenticators are relaxed compared to those for [=clients=].
+
+When a [=[WAA]=] truncates a string,
+the truncation behaviour MUST satisfy the following requirements:
+
+Choose a size limit equal to or greater than the specified minimum supported length.
+The string MAY be truncated so that its length in bytes in the UTF-8 character encoding satisfies that limit.
+This truncation SHOULD respect UTF-8 code point boundaries, and MAY respect [=grapheme cluster=] boundaries [[UAX29]].
+The resulting truncated value MAY be shorter than the chosen size limit
+but MUST NOT be shorter than the longest prefix substring that satisfies the size limit and ends on a [=grapheme cluster=] boundary.
+
 
 ### Language and Direction Encoding ### {#sctn-strings-langdir}
 
--- a/index.bs
+++ b/index.bs
@@ -3194,7 +3194,9 @@ associated with or [=scoped=] to, respectively.
 
         When [=clients=], [=client platforms=], or [=authenticators=] display a {{PublicKeyCredentialEntity/name}}'s value, they should always use UI elements to provide a clear boundary around the displayed value, and not allow overflow into other elements [[css-overflow-3]].
 
-        Authenticators MAY truncate characters from the end of a {{PublicKeyCredentialEntity/name}} member's value so that when encoded in UTF-8, its length fits within 64 bytes. See [[#sctn-strings-truncation]] about truncation and other considerations.
+        When storing a {{PublicKeyCredentialEntity/name}} member's value,
+        [=authenticators=] MUST support any value whose length in the UTF-8 character encoding is 64 bytes or less.
+        Authenticators MAY truncate the value as described in [[#sctn-strings-truncation]].
 </div>
 
 
@@ -3265,9 +3267,9 @@ credential.
 
         When [=clients=], [=client platforms=], or [=authenticators=] display a {{PublicKeyCredentialUserEntity/displayName}}'s value, they should always use UI elements to provide a clear boundary around the displayed value, and not allow overflow into other elements [[css-overflow-3]].
 
-        [=Authenticators=] MUST accept and store a {{PublicKeyCredentialUserEntity/displayName}} member's value
-        whose length in the UTF-8 character encoding is 64 bytes or shorter.
-        Authenticators MAY truncate characters from the end of a {{PublicKeyCredentialEntity/displayName}} member's value so that when encoded in UTF-8, its length fits within 64 bytes. See [[#sctn-strings-truncation]] about truncation and other considerations.
+        When storing a {{PublicKeyCredentialUserEntity/displayName}} member's value,
+        [=authenticators=] MUST support any value whose length in the UTF-8 character encoding is 64 bytes or less.
+        Authenticators MAY truncate the value as described in [[#sctn-strings-truncation]].
 </div>
 
 
--- a/index.bs
+++ b/index.bs
@@ -3194,7 +3194,7 @@ associated with or [=scoped=] to, respectively.
 
         When [=clients=], [=client platforms=], or [=authenticators=] display a {{PublicKeyCredentialEntity/name}}'s value, they should always use UI elements to provide a clear boundary around the displayed value, and not allow overflow into other elements [[css-overflow-3]].
 
-        Authenticators MAY truncate a {{PublicKeyCredentialEntity/name}} member's value so that it fits within 64 bytes, if the authenticator stores the value. See [[#sctn-strings-truncation]] about truncation and other considerations.
+        Authenticators MAY truncate a {{PublicKeyCredentialEntity/name}} member's value so that its UTF-8 encoding fits within 64 bytes, if the authenticator stores the value. See [[#sctn-strings-truncation]] about truncation and other considerations.
 </div>
 
 
@@ -3265,8 +3265,9 @@ credential.
 
         When [=clients=], [=client platforms=], or [=authenticators=] display a {{PublicKeyCredentialUserEntity/displayName}}'s value, they should always use UI elements to provide a clear boundary around the displayed value, and not allow overflow into other elements [[css-overflow-3]].
 
-        [=Authenticators=] MUST accept and store a 64-byte minimum length for a {{PublicKeyCredentialUserEntity/displayName}}
-        member's value. Authenticators MAY truncate a {{PublicKeyCredentialUserEntity/displayName}} member's value so that it fits within 64 bytes. See [[#sctn-strings-truncation]] about truncation and other considerations.
+        [=Authenticators=] MUST accept and store a {{PublicKeyCredentialUserEntity/displayName}} member's value
+        whose UTF-8 encoding is 64 bytes or shorter.
+        Authenticators MAY truncate a {{PublicKeyCredentialUserEntity/displayName}} member's value so that its UTF-8 encoding fits within 64 bytes. See [[#sctn-strings-truncation]] about truncation and other considerations.
 </div>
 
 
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -3194,7 +3194,7 @@ associated with or [=scoped=] to, respectively.
 
         When [=clients=], [=client platforms=], or [=authenticators=] display a {{PublicKeyCredentialEntity/name}}'s value, they should always use UI elements to provide a clear boundary around the displayed value, and not allow overflow into other elements [[css-overflow-3]].
 
-        Authenticators MAY truncate a {{PublicKeyCredentialEntity/name}} member's value so that it fits within 64 bytes, if the authenticator stores the value. See [[#sctn-strings-truncation]] about truncation and other considerations.
+        Authenticators MAY truncate a {{PublicKeyCredentialEntity/name}} member's value so that its UTF-8 encoding fits within 64 bytes, if the authenticator stores the value. See [[#sctn-strings-truncation]] about truncation and other considerations.
```
aphillips, line 3197: The "UTF-8 encoding" reads oddly to me. Also, we want to be clear that we mean characters/code points here.

```suggestion
Authenticators MAY truncate characters from the end of a {{PublicKeyCredentialEntity/name}} member's value so that when the name is encoded in UTF-8, its length fits within 64 bytes, if the authenticator stores the value. See [[#sctn-strings-truncation]] about truncation and other considerations.
```

Also, is "if the authenticator stores the value" necessary?

index.bs
```diff
@@ -3265,8 +3265,9 @@ credential.
 
         When [=clients=], [=client platforms=], or [=authenticators=] display a {{PublicKeyCredentialUserEntity/displayName}}'s value, they should always use UI elements to provide a clear boundary around the displayed value, and not allow overflow into other elements [[css-overflow-3]].
 
-        [=Authenticators=] MUST accept and store a 64-byte minimum length for a {{PublicKeyCredentialUserEntity/displayName}}
-        member's value. Authenticators MAY truncate a {{PublicKeyCredentialUserEntity/displayName}} member's value so that it fits within 64 bytes. See [[#sctn-strings-truncation]] about truncation and other considerations.
+        [=Authenticators=] MUST accept and store a {{PublicKeyCredentialUserEntity/displayName}} member's value
+        whose UTF-8 encoding is 64 bytes or shorter.
+        Authenticators MAY truncate a {{PublicKeyCredentialUserEntity/displayName}} member's value so that its UTF-8 encoding fits within 64 bytes. See [[#sctn-strings-truncation]] about truncation and other considerations.
```
aphillips, line 3270:
```suggestion
Authenticators MAY truncate characters from the end of a {{PublicKeyCredentialEntity/displayName}} member's value so that when the display name is encoded in UTF-8, its length fits within 64 bytes. See [[#sctn-strings-truncation]] about truncation and other considerations.
```

index.bs
```diff
@@ -3265,8 +3265,9 @@ credential.
 
         When [=clients=], [=client platforms=], or [=authenticators=] display a {{PublicKeyCredentialUserEntity/displayName}}'s value, they should always use UI elements to provide a clear boundary around the displayed value, and not allow overflow into other elements [[css-overflow-3]].
 
-        [=Authenticators=] MUST accept and store a 64-byte minimum length for a {{PublicKeyCredentialUserEntity/displayName}}
-        member's value. Authenticators MAY truncate a {{PublicKeyCredentialUserEntity/displayName}} member's value so that it fits within 64 bytes. See [[#sctn-strings-truncation]] about truncation and other considerations.
+        [=Authenticators=] MUST accept and store a {{PublicKeyCredentialUserEntity/displayName}} member's value
+        whose UTF-8 encoding is 64 bytes or shorter.
```
aphillips, line 3269:
```suggestion
        [=Authenticators=] MUST accept and store a {{PublicKeyCredentialUserEntity/displayName}} member's value
        whose length in the UTF-8 character encoding is 64 bytes or shorter.
```

aphillips (changes requested):
Thanks for copying me on this PR. Some suggestions...

index.bs
emlun: Thanks! I agree that "if the authenticator stores the value" wasn't necessary. I also saw some opportunities to make this language both shorter and more strict, please let me know what you think.

index.bs
emlun: Thanks! I saw some opportunities to make this language both shorter and more strict, please let me know what you think.

index.bs
emlun: Thanks! I rewrote this a bit, and changed "shorter" to "less" since the subject is now the length rather than the byte string. Also applied the same language to `name` as well.

index.bs
```diff
@@ -4919,7 +4922,7 @@ Authenticators may be required to store arbitrary strings chosen by a [=[RP]=],
 
 ### String Truncation ### {#sctn-strings-truncation}
 
-Each arbitrary string in the API will have some accommodation for the potentially limited resources available to an [=authenticator=]. If string value truncation is the chosen accommodation then authenticators MAY truncate in order to make the string fit within a length equal or greater than the specified minimum supported length. Such truncation SHOULD also respect UTF-8 sequence boundaries or [=grapheme cluster=] boundaries [[UAX29]]. This defines the maximum truncation permitted and authenticators MUST NOT truncate further.
+Each arbitrary string in the API will have some accommodation for the potentially limited resources available to an [=authenticator=]. If string value truncation is the chosen accommodation then authenticators MAY truncate in order to make the string fit within a length equal to or greater than the specified minimum supported length. Such truncation SHOULD also respect UTF-8 sequence boundaries or [=grapheme cluster=] boundaries [[UAX29]]. This defines the maximum truncation permitted and authenticators MUST NOT truncate further.
```
aphillips, line 4925: I think "UTF-8 sequence boundaries" is unclear. I would suggest saying "UTF-8 character boundaries" or "UTF-8 code point boundaries". I would also make SHOULD->MUST. Arbitrary truncation of UTF-8 mid-character should not be permitted. I'm okay with SHOULD for grapheme cluster boundaries. Ultimately, I'd make that sentence say:

> Such truncation MUST only be done on UTF-8 character boundaries 
> and SHOULD respect [=grapheme cluster]= boundaries [[UAX29]].

Note that the "equal or greater to" discussion is super tricky. Because UTF-8 is a multibyte encoding, the truncation to (say) 64-bytes might end up with a string that is 62 or 63 bytes long (so as to avoid mid-character truncation).

emlun: > I think "UTF-8 sequence boundaries" is unclear. I would suggest saying "UTF-8 character boundaries" or "UTF-8 code point boundaries".

@agl thoughts on this?

> I would also make SHOULD->MUST.

I believe we wrote this as a SHOULD instead of a MUST because there already exist authenticator implementations that do not respect these guidelines. This is also why we recommend that the client performs this truncation instead, because clients are better equipped to handle all the Unicode intricacies like identifying grapheme clusters that span multiple code points.

> Note that the "equal or greater to" discussion is super tricky. Because UTF-8 is a multibyte encoding, the truncation to (say) 64-bytes might end up with a string that is 62 or 63 bytes long (so as to avoid mid-character truncation).

Note that what's specified here with "equal to or greater than" is the _length the string may be truncated to fit within_: authenticators may truncate strings to _fit within_ X bytes, where X is "equal to or greater than the specified minimum supported length", so in this case X >= 64. So what this says is that authenticators may truncate the string _to fit within 64 bytes_, not just to exactly 64 bytes, for the reasons you point out. But it also gives authenticators the option to choose X _higher_ than 64 if they like. Then there's also the next requirement: "This defines the maximum truncation permitted and authenticators MUST NOT truncate further.", so an (X+1)-byte string may be truncated to X-3 bytes but not to, say, X-32 bytes.

aphillips: > Note that what's specified here with "equal to or greater than" is the length the string may be truncated to fit within: authenticators may truncate strings to fit within X bytes, where X is "equal to or greater than the specified minimum supported length", so in this case X >= 64. So what this says is that authenticators may truncate the string to fit within 64 bytes, not just to exactly 64 bytes, for the reasons you point out. But it also gives authenticators the option to choose X higher than 64 if they like. Then there's also the next requirement: "This defines the maximum truncation permitted and authenticators MUST NOT truncate further.", so an (X+1)-byte string may be truncated to X-3 bytes but not to, say, X-32 bytes.

I have to admit that the "equal to or greater than" phrasing is hard to parse. I read this (as intended) as allowing an implementation to choose a longer limit but not one that is shorter than the given limit (64 in most of our examples).

It's possible that people will read the "MUST NOT truncate further" requirement as meaning "exactly X bytes", when UTF-8 will produce some truncated strings of less than X bytes (but never more than 3 fewer).

> I believe we wrote this as a SHOULD instead of a MUST because there already exist authenticator implementations that do not respect these guidelines.

Perhaps there are such authenticators: they should feel icky and awful and fix their code 😉. I don't see why your spec should not require it. Note that I do not think we should hold up the same bar for grapheme cluster boundaries.

aphillips: > Would this phrasing be better?

>> Authenticators MUST NOT truncate further than the minimum amount required to align with a [=grapheme cluster=] boundary.

That's pretty good.

If we took this I would still make the change I requested to say "UTF-8 character boundaries" or "UTF-8 code point boundaries" in the preceding sentence.

---

Regarding UTF-8 truncation, I understand what you're saying and my comment was (perhaps inappropriately) dismissive of legacy implementations. 

"UTF-8 engine" makes it sound complicated. The design of UTF-8 is such that truncation on character boundaries can rely on a bitmask to check the byte at the truncation location. The code doesn't have to have a deep (or even shallow) understanding of Unicode to find code point boundaries. 

My concern about using SHOULD here is that it leaves the wrong impression. It might allow _more_ of these bad actors to come into the world. If we must preserve compatibility with these stale implementations (and I agree that we probably do), we should look for a formulation that does what I suggested before: makes people feel icky about not fixing their code.

Perhaps we can sidestep the problem:

```suggestion
Each arbitrary string in the API will have some accommodation for the potentially limited resources available to an
[=authenticator=]. If string value truncation is the chosen accommodation then authenticators MAY truncate the 
string value using a size limit equal to or greater than the specified minimum supported length. The resulting 
truncated value MAY be shorter than the chosen limit only by the number of bytes necessary to ensure that the
truncation falls on [=code point=] or [=grapheme cluster=][[UAX29]] boundaries in the UTF-8 encoded text. 
```

index.bs
```diff
@@ -3194,7 +3194,9 @@ associated with or [=scoped=] to, respectively.
 
         When [=clients=], [=client platforms=], or [=authenticators=] display a {{PublicKeyCredentialEntity/name}}'s value, they should always use UI elements to provide a clear boundary around the displayed value, and not allow overflow into other elements [[css-overflow-3]].
 
-        Authenticators MAY truncate a {{PublicKeyCredentialEntity/name}} member's value so that it fits within 64 bytes, if the authenticator stores the value. See [[#sctn-strings-truncation]] about truncation and other considerations.
+        When storing a {{PublicKeyCredentialEntity/name}} member's value,
+        [=authenticators=] MUST support any value whose length in the UTF-8 character encoding is 64 bytes or less.
```
agl, line 3198: "support" seems like an imprecise verb here. ("Sure, I support it&mdash;by truncating to 16 bytes.)

index.bs
```diff
@@ -3265,8 +3267,9 @@ credential.
 
         When [=clients=], [=client platforms=], or [=authenticators=] display a {{PublicKeyCredentialUserEntity/displayName}}'s value, they should always use UI elements to provide a clear boundary around the displayed value, and not allow overflow into other elements [[css-overflow-3]].
 
-        [=Authenticators=] MUST accept and store a 64-byte minimum length for a {{PublicKeyCredentialUserEntity/displayName}}
-        member's value. Authenticators MAY truncate a {{PublicKeyCredentialUserEntity/displayName}} member's value so that it fits within 64 bytes. See [[#sctn-strings-truncation]] about truncation and other considerations.
+        When storing a {{PublicKeyCredentialUserEntity/displayName}} member's value,
+        [=authenticators=] MUST support any value whose length in the UTF-8 character encoding is 64 bytes or less.
```
agl, line 3271: (Ditto.)

index.bs
emlun: > "UTF-8 engine" makes it sound complicated. The design of UTF-8 is such that truncation on character boundaries can rely on a bitmask to check the byte at the truncation location. The code doesn't have to have a deep (or even shallow) understanding of Unicode to find code point boundaries.

This is true of code points, but the greater point of this section is that Unicode code point boundaries don't necessarily align with what a user considers a "character". See section [3 Grapheme Cluster Boundaries](https://www.unicode.org/reports/tr29/#Grapheme_Cluster_Boundaries) in our referenced [UAX#29]:

>[...] For example, “G” + _grave-accent_ is a _user-perceived character_: users think of it as a single character, yet is actually represented by two Unicode code points. [...]

See also the reference to [grapheme cluster](https://w3c.github.io/i18n-glossary/#dfn-grapheme-cluster) in the Internationalization Glossary:
>[...] For example, a Slovak user might wish to treat the default pair of grapheme clusters "ch" as a single grapheme cluster. Note that the interaction between the language of string content and the end-user's preferences might be complex.

This is what I meant vaguely by "UTF-8 engine" earlier: the default rules for grapheme cluster segmentation are already complex, but a truly faithful implementation should also account for the user's preferences, locale settings and possibly even other context information. An authenticator separate from the client (user agent) cannot feasibly take all of this into account. The client is much more likely to both have enough storage and update capabilities to keep up with these language processing rules and to have access to the necessary context information in the first place.

All this in mind, I came up with this formulation. I feel like this may be a bit more stuffy than necessary, but at least it should be unambiguous. What do you think?

>If string truncation is the chosen accommodation, then the authenticator MAY choose a size limit equal to or greater than the specified minimum supported length and truncate the string to fit within that limit. Such truncation SHOULD also respect UTF-8 code point boundaries or [grapheme cluster](https://w3c.github.io/i18n-glossary/#dfn-grapheme-cluster) boundaries [[UAX29]](file:///home/emlun/dev/webauthn/index.html#biblio-uax29). The resulting truncated value MAY be shorter than the chosen size limit but MUST NOT be shorter than the longest prefix substring that satisfies the size limit and ends on a [grapheme cluster](https://w3c.github.io/i18n-glossary/#dfn-grapheme-cluster) boundary.

index.bs
emlun: Agreed. Rephrased as below, is that better?

>When storing a [name](https://pr-preview.s3.amazonaws.com/w3c/webauthn/pull/2017.html#dom-publickeycredentialentity-name) member’s value, [authenticators](https://pr-preview.s3.amazonaws.com/w3c/webauthn/pull/2017.html#authenticator) MAY truncate the value as described in [§ 6.4.1 String Truncation](https://pr-preview.s3.amazonaws.com/w3c/webauthn/pull/2017.html#sctn-strings-truncation) using a size limit greater than or equal to 64 bytes.

index.bs
```diff
@@ -4919,9 +4922,15 @@ Authenticators may be required to store arbitrary strings chosen by a [=[RP]=],
...
+Such truncation SHOULD also respect UTF-8 code point boundaries or [=grapheme cluster=] boundaries [[UAX29]].
+The resulting truncated value MAY be shorter than the chosen size limit
+but MUST NOT be shorter than the longest prefix substring that satisfies the size limit and ends on a [=grapheme cluster=] boundary.
 
-For example, in <a href="#fig-stringTruncation">figure <span class="figure-num-following"></span></a> the string is 65 bytes long. If truncating to 64 bytes then the final 0x88 byte must be removed purely because of space reasons. Since that leaves a partial UTF-8 sequence the remainder of that sequence may also be removed. Since that leaves a partial [=grapheme cluster=] an authenticator may remove the remainder of that cluster.
+For example, in <a href="#fig-stringTruncation">figure <span class="figure-num-following"></span></a> the string is 65 bytes long. If truncating to 64 bytes then the final 0x88 byte must be removed purely because of space reasons. Since that leaves a partial UTF-8 code point the remainder of that code point may also be removed. Since that leaves a partial [=grapheme cluster=] an authenticator may remove the remainder of that cluster.
```
aphillips, line 4933: "code point may also be removed" => "code point must also be removed"

index.bs
```diff
@@ -4919,9 +4922,15 @@ Authenticators may be required to store arbitrary strings chosen by a [=[RP]=],
 
 ### String Truncation ### {#sctn-strings-truncation}
 
-Each arbitrary string in the API will have some accommodation for the potentially limited resources available to an [=authenticator=]. If string value truncation is the chosen accommodation then authenticators MAY truncate in order to make the string fit within a length equal or greater than the specified minimum supported length. Such truncation SHOULD also respect UTF-8 sequence boundaries or [=grapheme cluster=] boundaries [[UAX29]]. This defines the maximum truncation permitted and authenticators MUST NOT truncate further.
+Each arbitrary string in the API will have some accommodation for the potentially limited resources available to an [=authenticator=].
+If string truncation is the chosen accommodation,
+then the authenticator MAY choose a size limit equal to or greater than the specified minimum supported length
+and truncate the string so that its length in the UTF-8 character encoding satisfies that limit.
+Such truncation SHOULD also respect UTF-8 code point boundaries or [=grapheme cluster=] boundaries [[UAX29]].
```
aphillips, line 4929:
```suggestion
Such truncation MUST respect Unicode [=code point=] boundaries and SHOULD respect [=grapheme cluster=] boundaries [[UAX29]].
```

index.bs
emlun: Thanks for your comments! I've rewritten the section to more clearly distinguish between the responsibilities of clients vs. authenticators, where the requirements are now MUST/SHOULD (code points/clusters) for clients and SHOULD/MAY for authenticators. Please have another look! I'll resolve this thread since the whole text is rearranged.

index.bs

emlun: (But note that with the RFC 2119 blurb updated in #2039, this "must" is not normative. There is a normative restatement later in the text.)

index.bs
emlun: Done (for clients). As mentioned above I've rewritten the section, please have another look!

agl:
From the call of 2024-05-15: ok to land after agl reviews.

agl (approved):

index.bs
```diff
@@ -3194,7 +3194,9 @@ associated with or [=scoped=] to, respectively.
 
         When [=clients=], [=client platforms=], or [=authenticators=] display a {{PublicKeyCredentialEntity/name}}'s value, they should always use UI elements to provide a clear boundary around the displayed value, and not allow overflow into other elements [[css-overflow-3]].
 
-        Authenticators MAY truncate a {{PublicKeyCredentialEntity/name}} member's value so that it fits within 64 bytes, if the authenticator stores the value. See [[#sctn-strings-truncation]] about truncation and other considerations.
+        When storing a {{PublicKeyCredentialEntity/name}} member's value,
```
agl, line 3197: Very minor nit: this comma (and in the next diff hunk) read oddly to me, but I'm happy either way.

emlun: Hm, I see. If you don't mind I'll keep it as is, if only for consistency with the first sentence in the paragraph immediately preceding this (and also the next hunk).

PR #2020, opened 2024-02-08
Improved version of extension for Transaction Confirmation

rlin1:
supporting security keys or platform authenticators to show the transaction.  Also supports the platform to show the transaction text.
If the client displayed the transaction text but not the authenticator (for example when using a security without display), the transaction text will be only included in the collectedClientData - similar to what SPC is doing.
If the authenticator showed the transaction text, it will also be included in the extension that was signed by the authenticator (known and potentially attested entity showed the transaction text as opposed to a unknown client).

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/pull/2020.html" title="Last updated on Oct 27, 2024, 2:48 PM UTC (704decb)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/2020/061a649...704decb.html" title="Last updated on Oct 27, 2024, 2:48 PM UTC (704decb)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -7669,23 +7669,25 @@ This extension allows for a simple form of transaction authorization. A
 :: A single USVString prompt.
     <xmp class="idl">
     partial dictionary AuthenticationExtensionsClientInputs {
-      USVString txAuthSimple2;
+      USVString txAuthSimple;
     };
     </xmp>
 
 : Client extension processing
-:: If (and only if) the authenticator doesn't support this extension (to be determined by the platform via the response to CTAP getClientInfo), then the Client is expected to display
-    the transaction text to the user and to include the transaction text in the collectedClientData structure as described below.
-    If the authenticator supports this extension, then the client SHALL pass-through the extension to the
-    authenticator and not perform any other client processing.
+:: If this extension is present, the client SHALL
+       1. use a dialog to the user that makes sense in the context of approving a transaction (as opposed to sign in).
+       1. display the transaction text to the user. The client SHOULD
+             indicate that the transaction text originates from a specific relying party
+	     (as opposed to the platform itself).
+	     
+       1. use the {{CollectedClientTxAuthSimple2Data}} structure containing the transaction text instead of using
+             the {{CollectedClientData}} structure.
+	     
+       1. pass-through the extension to the authenticator (see "client extension output" below)
+       1. pass-through the "authenticator extension output" to the caller as part of the assertion
 
 
     <div class="issue">
-      Mention that the client shouldn't prompt the user for "sign-in" but instead for approving a transaction.
-
-      Should we always let the client use CollectedClientTxAuthSimple2Data instead of CollectedClientData to have more consistency (or only if the client displayed it)?
-      If yes: Should the client verify it matches the transactionText returned by the authenticator (if that exists)?
-
       Should we keep the "type" on "authenticate" or change to something closer to transaction (SPC uses 'payment.get')?
 
       Should we change the name back to txAuthSimple - cannot really hurt - but makes it easier for authenticators already implementing it.
@@ -7728,7 +7730,7 @@ This extension allows for a simple form of transaction authorization. A
 :: Returns the authenticator extension output string UTF-8 decoded into a USVString.
     <xmp class="idl">
     partial dictionary AuthenticationExtensionsClientOutputs {
-      USVString txAuthSimple2;
+      USVString txAuthSimple;
     };
     </xmp>
 
@@ -7741,11 +7743,12 @@ txAuthSimpleInput = (tstr)
 </pre>
 
 : Authenticator extension processing
-:: The authenticator MUST display the prompt to the user before performing either [=user verification=] or [=test of user
+:: The authenticator MUST display the transaction text to the user
+    before performing either [=user verification=] or [=test of user
     presence=]. The authenticator MAY insert line breaks if needed.
 
 : Authenticator extension output
-:: A single CBOR string, representing the prompt as displayed (including any eventual line breaks).
+:: A single CBOR string, representing the transaction text (excluding any eventual line breaks that were inserted).
 
 <pre>
 CDDL:
--- a/index.bs
+++ b/index.bs
@@ -7687,12 +7687,6 @@ This extension allows for a simple form of transaction authorization. A
        1. pass-through the "authenticator extension output" to the caller as part of the assertion
 
 
-    <div class="issue">
-      Should we keep the "type" on "authenticate" or change to something closer to transaction (SPC uses 'payment.get')?
-
-      Should we change the name back to txAuthSimple - cannot really hurt as no platform supports it today and authenticator implementations don't have to change.
-    </div>
-
     <xmp class="idl">
         dictionary CollectedClientTxAuthSimpleData : CollectedClientData {
             required CollectedClientAdditionalTxAuthSimpleData transactionText;
@@ -9360,6 +9354,22 @@ for their contributions as our W3C Team Contacts.
     "href": "https://tools.ietf.org/html/rfc8610",
     "status": "IETF Proposed Standard",
     "date": "June 2019"
+  },
+
+  "RFC9052": {
+    "authors": ["Jim Schaad"],
+    "title": "CBOR Object Signing and Encryption (COSE): Structures and Process",
+    "href": "https://datatracker.ietf.org/doc/rfc9052/",
+    "status": "IETF Internet Standard",
+    "date": "August 2022"
+  },
+
+  "RFC9053": {
+    "authors": ["Jim Schaad"],
+    "title": "CBOR Object Signing and Encryption (COSE): Initial Algorithms",
+    "href": "https://datatracker.ietf.org/doc/rfc9053/",
+    "status": "RFC Informational",
+    "date": "August 2022"
   },
 
   "ISOBiometricVocabulary": {
--- a/index.bs
+++ b/index.bs
@@ -7652,6 +7652,96 @@ To <dfn abstract-op>Create a new supplemental public key record</dfn>, perform t
         [=set/append=] this [=supplemental public key record=] to |credentialRecord|.[$credential record/supplementalPubKeys$].
 
 
+### Simple Transaction Authorization Extension (txAuthSimple2) ### {#sctn-simple-txauth-extension2}
+
+This extension allows for a simple form of transaction authorization. A
+   [=[RP]=] can specify a prompt string, intended for display by the platform and by the authenticator (if supported).
+   With this approach, [=Relying Parties=] can use the feature independently of the capabilities of the authenticator used by the user,
+   while still benefitting from the increased security level if the authenticator itself supports showing the transaction text.
+
+: Extension identifier
+:: `txAuthSimple2`
+
+: Operation applicability
+:: [=authentication extension|Authentication=]
+
+: Client extension input
+:: A single USVString prompt.
+    <xmp class="idl">
+    partial dictionary AuthenticationExtensionsClientInputs {
+      USVString txAuthSimple2;
+    };
+    </xmp>
+
+: Client extension processing
+:: If (and only if) the authenticator doesn't support this extension (to be determined by the platform via the response to CTAP getClientInfo), then the Client is expected to display
+    the transaction text to the user and to include the transaction text in the collectedClientData structure as described below.
+    If the authenticator supports this extension, then the client SHALL pass-through the extension to the
+    authenticator and not perform any other client processing.
+
+    <xmp class="idl">
+        dictionary CollectedClientTxAuthSimple2Data : CollectedClientData {
+            required CollectedClientAdditionalTxAuthSimple2Data transactionText;
+        };
+    </xmp>
+
+    The {{CollectedClientTxAuthSimple2Data}} dictionary inherits from
+    {{CollectedClientData}}. It contains the following additional field:
+
+    <dl dfn-type="dict-member" dfn-for="CollectedClientTxAuthSimple2Data">
+        :  <dfn>transactionText</dfn> member
+        :: The full context of the transaction to sign.
+    </dl>
+
+
+    <xmp class="idl">
+        dictionary CollectedClientAdditionalTxAuthSimple2Data {
+            required USVString rpId;
+            required USVString transactionText;
+        };
+    </xmp>
+
+    The {{CollectedClientAdditionalTxAuthSimple2Data}} dictionary contains the following
+    fields:
+
+    <dl dfn-type="dict-member" dfn-for="CollectedClientAdditionalTxAuthSimple2Data">
+        :  <dfn>rpId</dfn> member
+        :: The id of the [=Relying Party=] that created the credential.
+
+        :  <dfn>transactionText</dfn> member
+        :: The full context of the transaction that the user is intended to approve.
+    </dl>
+
+: Client extension output
+:: Returns the authenticator extension output string UTF-8 decoded into a USVString.
+    <xmp class="idl">
+    partial dictionary AuthenticationExtensionsClientOutputs {
+      USVString txAuthSimple2;
+    };
+    </xmp>
+
+: Authenticator extension input
+:: If the authenticator supports this extension, then the client provides the extension input encoded as a CBOR text string (major type 3).
+
+<pre>
+CDDL:
+txAuthSimpleInput = (tstr)
+</pre>
+
+: Authenticator extension processing
+:: The authenticator MUST display the prompt to the user before performing either [=user verification=] or [=test of user
+    presence=]. The authenticator MAY insert line breaks if needed.
+
+: Authenticator extension output
+:: A single CBOR string, representing the prompt as displayed (including any eventual line breaks).
+
+<pre>
+CDDL:
+txAuthSimpleOutput = (tstr)
+</pre>
+
+
+
 # User Agent Automation # {#sctn-automation}
 
 For the purposes of user agent automation and [=web application=] testing, this document defines a number of [[WebDriver]] [=extension commands=].
--- a/index.bs
+++ b/index.bs
@@ -7674,22 +7674,20 @@ Example uses cases could be "I want to move $1234 from account A to account B" o
 
     <xmp class="idl">
     partial dictionary AuthenticationExtensionsClientInputs {
-      USVString confirmationPrompt;
+      USVString confirmation;
     };
     </xmp>
 
 : Client extension processing
-:: If this extension is present, the client SHALL
-       1. use a dialog to the user that makes the user aware of the confirmation to be provided (as opposed to doing a simple sign in).
-       1. display the confirmation prompt to the user. The client SHOULD
-             indicate that the confirmation prompt originates from a specific relying party
-             (as opposed to the platform itself).
+::   1. use a dialog to the user that makes the user aware of the confirmation to be provided (as opposed to doing a simple sign in).
+     1. display the confirmation prompt to the user. The client SHOULD
+           indicate that the confirmation prompt originates from a specific relying party
+           (as opposed to the platform itself).
      
-       1. use the {{CollectedClientConfirmationData}} structure containing the confirmation prompt instead of using
-             the {{CollectedClientData}} structure.
+     1. use the {{CollectedClientConfirmationData}} structure containing the confirmation prompt instead of using the {{CollectedClientData}} structure.
      
-       1. pass-through the extension to the authenticator (see "client extension output" below)
-       1. pass-through the "authenticator extension output" to the caller as part of the assertion
+     1. pass-through the extension to the authenticator (see "client extension output" below)
+     1. pass-through the "authenticator extension output" to the caller as part of the assertion
 
 
     <xmp class="idl">
@@ -7722,20 +7720,21 @@ Example uses cases could be "I want to move $1234 from account A to account B" o
     </dl>
 
 : Client extension output
-:: Returns the authenticator extension output string UTF-8 decoded into a USVString.
+:: Returns the value [TRUE] to indicate to the [[RP]=] that the extension was acted upon.
     <xmp class="idl">
     partial dictionary AuthenticationExtensionsClientOutputs {
-      USVString confirmationPrompt;
+      boolean confirmation;
     };
     </xmp>
 
 : Authenticator extension input
 :: The client provides the extension input encoded as a CBOR text string (major type 3).
 
-<pre>
-CDDL:
-confirmationPromptInput = (tstr)
-</pre>
+    ```
+    $$extensionInput //= (
+      confirmation = tstr,
+    )
+    ```
 
 : Authenticator extension processing
 :: The authenticator supporting this extension MUST display the confirmation prompt to the user
@@ -7743,29 +7742,23 @@ confirmationPromptInput = (tstr)
     presence=]. The authenticator MAY wrap lines that are too wide to be shown if needed.
 
 : Authenticator extension output
-:: If the authenticator supports this extension (and only then): A single CBOR string, representing the
-    confirmation prompt.
+:: A single CBOR text string, representing the confirmation prompt.
 
-<pre>
-CDDL:
-confirmationPromptOutput = (tstr)
-</pre>
+    ```
+    $$extensionOutput //= (
+      confirmation = tstr,
+    )
+    ```
 
 #### `confirmation` Extension Output Verification Procedures #### {#sctn-confirmation-extension-verification}
 
-Verifying the <code>[=confirmation=]</code> extension output is performed by the [=[RP]=] whenever a the use of <code>[=confirmatione=]</code> was requested.
-
-##### Authentication (`get()`) ##### {#sctn-confirmation-extension-verification-get}
-
-This extension is only specified to be used in conjunction with a {{CredentialsContainer/get()|navigator.credentials.get()}} call.
-
-If the [=[RP]=] requested the `confirmation` extension in a {{CredentialsContainer/get()|navigator.credentials.get()}} call, then the below verification step are performed in the context of [step 19](#authn-ceremony-verify-extension-outputs)
+The following verification steps are performed in the context of [step 19](#authn-ceremony-verify-extension-outputs)
 of [[#sctn-verifying-assertion]] using these variables established therein: |credential|, |clientExtensionResults|, |authData|, |hash|, and |credentialRecord|.
 [=[RP]=] policy may specify whether a response without a `confirmation` extension output is acceptable.
 
-1. Verify that the {{AuthenticationExtensionsClientOutputs/confirmation}} member of |clientExtensionResults| exists.
-1. Verify that the {{AuthenticationExtensionsClientOutputs/confirmation}} member contains the confirmation prompt that is expected (i.e., that was used when requesting the `confirmation` extension.
-
+1. Verify that the `confirmation` key exists in the [=authData/extensions=] in |authData|.
+1. Verify that the `confirmation` value in the [=authData/extensions=] in |authData| equals the confirmation prompt that is expected (i.e., that was used when requesting the `confirmation` extension).
+1. If processing a response without the `confirmation` extension is acceptable by policy: Fall back to <code>|credential|.{{PublicKeyCredential/response}}.{{AuthenticatorResponse/clientDataJSON}}</code> entry if authenticator extension output is absent.
 
 
 # User Agent Automation # {#sctn-automation}
@@ -9370,22 +9363,6 @@ for their contributions as our W3C Team Contacts.
     "href": "https://tools.ietf.org/html/rfc8610",
     "status": "IETF Proposed Standard",
     "date": "June 2019"
-  },
-
-  "RFC9052": {
-    "authors": ["Jim Schaad"],
-    "title": "CBOR Object Signing and Encryption (COSE): Structures and Process",
-    "href": "https://datatracker.ietf.org/doc/rfc9052/",
-    "status": "IETF Internet Standard",
-    "date": "August 2022"
-  },
-
-  "RFC9053": {
-    "authors": ["Jim Schaad"],
-    "title": "CBOR Object Signing and Encryption (COSE): Initial Algorithms",
-    "href": "https://datatracker.ietf.org/doc/rfc9053/",
-    "status": "RFC Informational",
-    "date": "August 2022"
   },
 
   "ISOBiometricVocabulary": {
--- a/index.bs
+++ b/index.bs
@@ -7735,7 +7735,7 @@ This extension allows for a simple form of transaction authorization. A
     </xmp>
 
 : Authenticator extension input
-:: If the authenticator supports this extension, then the client provides the extension input encoded as a CBOR text string (major type 3).
+:: The client provides the extension input encoded as a CBOR text string (major type 3).
 
 <pre>
 CDDL:
@@ -7743,12 +7743,13 @@ txAuthSimpleInput = (tstr)
 </pre>
 
 : Authenticator extension processing
-:: The authenticator MUST display the transaction text to the user
+:: The authenticator supporting this extension MUST display the transaction text to the user
     before performing either [=user verification=] or [=test of user
-    presence=]. The authenticator MAY insert line breaks if needed.
+    presence=]. The authenticator MAY insert line breaks when displaying the text if needed.
 
 : Authenticator extension output
-:: A single CBOR string, representing the transaction text (excluding any eventual line breaks that were inserted).
+:: If the authenticator supports this extension (and only then): A single CBOR string, representing the
+    transaction text (excluding any eventual line breaks that were inserted).
 
 <pre>
 CDDL:
--- a/index.bs
+++ b/index.bs
@@ -7652,7 +7652,7 @@ To <dfn abstract-op>Create a new supplemental public key record</dfn>, perform t
         [=set/append=] this [=supplemental public key record=] to |credentialRecord|.[$credential record/supplementalPubKeys$].
 
 
-### Simple Transaction Authorization Extension (txAuthSimple2) ### {#sctn-simple-txauth-extension2}
+### Simple Transaction Authorization Extension (txAuthSimple) ### {#sctn-simple-txauth-extension}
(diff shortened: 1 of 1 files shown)


rlin1:
To be decided: should we change the extension name back to txAuthSimple?
Pro: Authenticators that implemented it would still work and don't need a change.  
Can it harm: Don't think so.  Clients never practically supported the extension and the functionality change relates to the clients.

rlin1:
Close #2022

timcappalli:
@christiaanbrand or @rlin1 can you join a WebAuthn WG call in March to discuss this?

FlxMgdnz:
We would love to see this extension added to L3 and get browser/OS support. It is fundamentally relevant - even crucial - for any PSD2 regulated bank and other financial institutions.

rlin1:
> @christiaanbrand or @rlin1 can you join a WebAuthn WG call in March to discuss this?

I plan to join the call on March 6th.

petrdvorak:
I want to vouch for this for the reason already mentioned - without the ability to send transaction data to an authenticator in a way it can display them to the user (what you see is what you sign), achieving compliance in banking will be difficult for any FIDO2-related efforts. Standards such as Secure Payment Confirmation help a bit, but having the extension available would be much appreciated.

mage28:
Most important stuff has already been mentioned by @rlin1, @FlxMgdnz and @petrdvorak. Thank you!
Having the option to send/display some data (ideally) on the authenticator would be greatly extend the usage for FIDO2/Passkey beyond authentication/sign-in usecases and open it up, e.g. for financial usecases (regulated by PSD2) that require the secure display of transaction details. But it is actually helpful for all usecases where it makes sense to show a user a compact information WHAT he is actually doing, in a way that is almost impossible to modify by malware and such.

petrdvorak:
Maybe one more "storytelling comment" on this...

When we started working on our FIDO2 HW token about two years ago, I read through the FIDO2 documents and thought: "This is great. We can send a challenge to an authenticator, and it will sign it."

Then we started implementing things and found out that `clientDataJSON`, which actually contains the challenge, is essentially useless as the bearer of the data, as it is sent to the authenticator ironed out via `SHA256` hash. While I understand the need for efficiency in interfaces like NFC/Bluetooth/USB (after all, some enterprises probably still run their COBOL apps on Windows 95), this is a neat candidate for abstraction in the standard so that we can decide to send `clientDataJSON` as a plain-text form to the authenticator, rather than hash and use the challenge itself for the WYSIWYS purpose.

From the specification, I understand the `clientDataJSON` structure is the following:

```
dictionary CollectedClientData {
    required DOMString           type;
    DOMString                    topOrigin;
    boolean                      crossOrigin;
};
```

So, practically, it is something like this (non-minified, intentionally randomly found a bank with a long name, and used long challenge):

```json
{
   "type": "webauthn.get",
   "challenge": "cb48493c-a0c7-4a9d-9aed-a172724d4357&A1*A100CZK*ICZ2730300000001165254011*D20180425",
   "origin": "moldindcondbank.md"
   "topOrigin": "moldindcondbank.md",
   "boolean": false
}
```

Is this so horrible in terms of size? I mean... the SHA256 hash almost makes the payload longer. ;-)

prusnak:
Any thoughts on this @andrewkozlik?

petrdvorak:
Maybe one more comment from my side after extensive reading on this subject (apologies, but I am fascinated that a challenge-response protocol does not allow sending the challenge to the authenticator in plain form, or at least augment it with custom data).

From what I understood, one of the reasons why web browser vendors did not implement `txAuthSimple` extension was that they did not want to display data provided by the relying party (web page) in the browser UI. We can live without this - no need to display anything. This is why we added a display to our authenticator. The user can only see the challenge data / associated data in the authenticator - we are OK with this. We just need a way to sneak structured data to the authenticator.

I still think the best way to do this is not to destroy the challenge object by hashing it, but having an extension that works the best effort is enough.

FlxMgdnz:
> Maybe one more comment from my side after extensive reading on this subject (apologies, but I am fascinated that a challenge-response protocol does not allow sending the challenge to the authenticator in plain form, or at least augment it with custom data).
> 
> From what I understood, one of the reasons why web browser vendors did not implement `txAuthSimple` extension was that they did not want to display data provided by the relying party (web page) in the browser UI. We can live without this - no need to display anything. This is why we added a display to our authenticator. The user can only see the challenge data / associated data in the authenticator - we are OK with this. We just need a way to sneak structured data to the authenticator.
> 
> I still think the best way to do this is not to destroy the challenge object by hashing it, but having an extension that works the best effort is enough.

I agree that it would be a good first step to simply hand over tx data to authenticators and enable scenarios where dedicated hardware authenticators with display capabilities can be used. Ultimately, we should aim for platform support, as this will have a much bigger impact.

From what I remember, there were concerns about tx data formatting and encoding, which need to be handled in secure context where you'd want as little of such complexity as possible. So the challenge is to define a format that is comprehensive enough for a user to understand what they are signing off, but simple enough for the platform providers' implementation teams to follow.

Windows hello txAuth

petrdvorak:
OK, it makes sense.

One solution could be just to ask the RP to provide a meaningful message, as you suggest.

If a structured data is required, I can contribute my experience in this by suggesting formats to look into, then. If this is primarily about payments and PSD2 (as many people above suggest), you can look into:

- [SPAYD](https://en.wikipedia.org/wiki/Short_Payment_Descriptor) (short payment description) which I "invented" and is used on every invoice in the Czech Republic.
- [EPC Code](https://en.wikipedia.org/wiki/EPC_QR_code) the same thing, proposed on European level

In our systems, we wanted to have more generic data encoded, so we took the experience from above to design a generic [data payload format for operation data](https://developers.wultra.com/components/powerauth-webflow/develop/documentation/Operation-Data.html#section-documentation).

As a result, it allows you to encode data such as payments, or configuration changes into the short fixed format payload, i.e.:

### Example 1

```
A1*A100CZK*ICZ2730300000001165254011*D20180425
```

> Payment of 100 CZK to IBAN CZ2730300000001165254011 with associated date 2018-04-25.

Displayed as:

```
AMOUNT: 100 CZK
IBAN: CZ2730300000001165254011
DATE: 2018-04-25
```

### Example 2

```
A0*TClosing Ticket*R32784237923
```

> Confirmation of closing ticket with reference ID 32784237923.

Displayed as:

```
MESSAGE: Closing Ticket
REFERENCE: 32784237923
```

mage28:
> I agree that it would be a good first step to simply hand over tx data to authenticators and enable scenarios where dedicated hardware authenticators with display capabilities can be used. Ultimately, we should aim for platform support, as this will have a much bigger impact.

I need to agree with Felix here. We need a solution that has browser/platform authenticator support, otherwise we won't have enough reach. While not the kind of security as with authenticators, it is still a major step forward from signing "something shown on the website".

> From what I remember, there were concerns about tx data formatting and encoding, which need to be handled in secure context where you'd want as little of such complexity as possible. 

Exactly. There won't be many - if any - formatting options. Thanks for the screenshot, I hope I can use it in our internal marketing slides for FIDP/passkey :-). It helps to understand what MAY be possible in future.

>If a structured data is required..

I worked a while ago on some similiar templates/patterns like SPAYD and EPC Code for other QR-code like graphics. It was horrible to design and you quickly get into issues with languages and still missing use cases, so after a couple of years we have many usecases that are ending up in the "plain text" variant, while there was not fitting variant. So even in the case we support a format for structured data, we always need an option was well as a fallback to plaintext. Currently I would argue that anything else than plaintext is nothing that could be achieve easily.

petrdvorak:
@mage28 Structured data can accommodate any use case where the test is sufficient, i.e.:

```
A0*TAny text with in any language to be displayed...
```

But I get the point.

Currently, we designed a way to have some workaround for this:

- We will support `txAuthSimple` as proposed in this PR.
- We will hack `hmac-secret` extension as a carrier of data to authenticator as workaround (we are OK with 64 bytes it gives us)

andrewkozlik:
I am very much in favor of reintroducing the `txAuthSimple` extension for all the reasons already mentioned above.

I agree with @mage28's arguments for using plain text in the `txAuthSimple` extension. After all, I think it was named "simple" for a reason. It should be as simple as possible to support by all parties involved and you can't get any more basic than plain text. Dealing with structured data is a job for something like `txAuthGeneric` which had a `contentType` field to specify how the data should be interpreted. I am not disputing @petrdvorak's statement that "structured data can accommodate any use case where the plain text is sufficient", but I am a bit worried that adding any complexity could become a blocker for wider adoption of the `txAuthSimple` extension.

One thing that deserves discussion IMHO is how to handle the fact that the authenticator may insert line breaks if needed. The question is whether the authenticator extension output should include the line breaks, like in the earlier L1 version of the spec, or exclude them, as is the case of the present PR. I can see arguments on both sides. If the way the message is displayed on the authenticator is broken up such that the user misinterprets its meaning, then there can be legal consequences. The signature issued by the authenticator should therefore reflect the exact string that the user authorized so that it can serve as evidence in case of a dispute. On the other hand, this of course puts extra work on the relying party, because it needs to evaluate whether the returned `txAuthSimpleOutput` is acceptable. In the simplest case this could just mean replacing all line breaks in `txAuthSimpleOutput` with spaces and comparing to the original `transactionText`. But things can quickly become more complicated, for example, with narrow screens which may require breaking up long words or other expressions without spaces, e.g. IBAN.

petrdvorak:
I am OK with keeping the content plain text. Anything is better for us than nothing. 😊

@andrewkozlik I have spent quite some time in mobile banking, dealing with various screen sizes and varying line-break outputs. This does not present a problem in practice. I would even say that the WYSIWYS principle is even more open. You always sign some bits, but you never see those - WYSIWIS does not have to take into account:

- font family, color and size (which the user can modify by enabling color filters or assistive features)
- text breaking on a display (which may depend on screen resolution)
- symbol replacement (i.e., changing EUR for € sign)
- even the localization (meaning it is OK to see "Please approve payment of 100.00 EUR to account 123123/0100" message while signing byte representation of `A2*A100.00EUR*Q123123/0100`, as long as you can justify the texts have equal meaning and the mapping is correct, plus this allows you to change language dynamically)

After all, at various points, you need to trust the system for the correct WYSIWYS, i.e., the display has to display what you tell it to (which is kind of expected), the fonts installed in the system must result in displaying the right message, etc.

We have confirmed this principle when discussing signing in our mobile authenticator.

As a result, if we put some text in the `txAuthSimple`, I would argue that we should work with this text "as is" in anything related to cryptography. Authenticators can put any formatting to the text for presentation purposes, but this should live independently from signing...

index.bs
```diff
@@ -7652,6 +7652,121 @@ To <dfn abstract-op>Create a new supplemental public key record</dfn>, perform t
         [=set/append=] this [=supplemental public key record=] to |credentialRecord|.[$credential record/supplementalPubKeys$].
 
 
+### Simple Confirmation Extension (confirmation) ### {#sctn-simple-txauth-extension}
```
MasterKale, line 7655: Do we still want the link to this section to reference txAuthExtension?

index.bs
```diff
@@ -7652,6 +7652,121 @@ To <dfn abstract-op>Create a new supplemental public key record</dfn>, perform t
...
+      USVString confirmationPrompt;
+    };
+    </xmp>
+
+: Authenticator extension input
+:: The client provides the extension input encoded as a CBOR text string (major type 3).
+
+<pre>
+CDDL:
+confirmatonPromptInput = (tstr)
```
MasterKale, line 7736: There's a tiny typo here, "`confirmaton`"

petrdvorak:
Hello @rlin1, could you please provide some context behind renaming the extension? We are OK with a different name, just curious...

rlin1:
Consider using a short name for the extension "conf" - instead of "confirmation".

petrdvorak:
Is introducing the confirmation length limit (1024 bytes?) helpful or harmful? What do you think?

index.bs
```diff
@@ -7652,6 +7652,121 @@ To <dfn abstract-op>Create a new supplemental public key record</dfn>, perform t
...
+    confirmation prompt.
+
+<pre>
+CDDL:
+confirmationPromptOutput = (tstr)
+</pre>
+
+#### `confirmation` Extension Output Verification Procedures #### {#sctn-txauthsimple-extension-verification}
+
+Verifying the <code>[=txAuthSimple=]</code> extension output is performed by the [=[RP]=] whenever a the use of <code>[=confirmatione=]</code> was requested.
```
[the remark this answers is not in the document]
rlin1: addressed

index.bs
rlin1: addressed

rlin1:
> Is introducing the confirmation length limit (1024 bytes?) helpful or harmful? What do you think?

likely helpful.  Potentially even shorter...

mage28:
> Is introducing the confirmation length limit (1024 bytes?) helpful or harmful? What do you think?

Characters or bytes? Practically 1024 is more than sufficient, probably you will need to stick with very short and clear text informations, probably of less than 100 or 200 characters. So setting a max length limit of 1024 is surely not harmful for all usecases I can imagine and sets some technical restrictions. It can potentially be even short, but then we start to discuss when it starts to hurt the one or other.

petrdvorak:
@mage28 We are looking at this from the connected authenticator perspective, so bytes are a bit better measure of the data transfer limit. For our use case, the typical message will be much shorter. Even rich transaction details are shorter, i.e.:

```
Please approve a recurring payment of 43,289.99 EUR from an account
FI211234569876543210 to account IE64BOFI90583812345678, payment
reference: 9458711390468640.
```

_(160 characters)_

_Note: PSD3 might change the scope of dynamic linking, so that the authentication code is linked to an amount, currency, destination account, and source account, at least._

prusnak:
256 or 512 or 1024 bytes (not chars) including the zero terminator char, please.

That is 255 or 511 or 1023 printable characters.

petrdvorak:
@prusnak Reading through the conversation, this will primarily be about platform authenticators. Also, regarding the "confirmation" naming semantics, I read it as "it is the text the user can see in Windows UI alongside the standard confirmation window"). As a result, the JavaScript code will likely work by just providing a text... I would suggest keeping this USVString ("characters") and trimming the limit to 256 characters (which is plenty, as my example above suggests). UTF-8 encoding will result in at most 1024 bytes and will fit the zero terminating byte, or will fill the limit (CTAP protocol should hopefully take care of working with the data as bytes).

index.bs
```diff
@@ -7652,6 +7652,122 @@ To <dfn abstract-op>Create a new supplemental public key record</dfn>, perform t
...
+       1. use a dialog to the user that makes the user aware of the confirmation to be provided (as opposed to doing a simple sign in).
+       1. display the confirmation prompt to the user. The client SHOULD
+             indicate that the confirmation prompt originates from a specific relying party
+             (as opposed to the platform itself).
+     
+       1. use the {{CollectedClientConfirmationData}} structure containing the confirmation prompt instead of using
+             the {{CollectedClientData}} structure.
+     
+       1. pass-through the extension to the authenticator (see "client extension output" below)
+       1. pass-through the "authenticator extension output" to the caller as part of the assertion
```
emlun, line 7692: "If this extension is present, the client SHALL" is implicit and can be dropped. The last step is also implicit in the "Client extension output" section. Please begin sentences with an uppercase letter and use the `[=[RP]=]` term.

I would also write the fourth step as "set the authenticator extension input to...":

```suggestion
::  1. Use a dialog to the user that makes the user aware of the confirmation to be provided (as opposed to doing a simple sign in).
    1. Display the confirmation prompt to the user. The client SHOULD
        indicate that the confirmation prompt originates from a specific [=[RP]=]
        (as opposed to the platform itself).
     
    1. Use the {{CollectedClientConfirmationData}} structure containing the confirmation prompt instead of using
        the {{CollectedClientData}} structure.
    1. Set the `confirmation` [=authenticator extension input=] to the value of {{AuthenticationExtensionsClientInputs/confirmation}}, encoded as a CBOR text string.
```

index.bs
```diff
@@ -7652,6 +7652,122 @@ To <dfn abstract-op>Create a new supplemental public key record</dfn>, perform t
...
+
+: Operation applicability
+:: [=authentication extension|Authentication=]
+
+: Client extension input
+:: A single USVString confirmationPrompt.  This string might contain the following formatting tags: <code>&lt;b&gt;&lt;/b&gt;</code> to mark the text inbetween as bold and <code>&lt;br /&gt;</code> to force a line break.  Support of these formatting tags is "best effort".  Implementations not supporting it SHALL NOT display the tags in "verbatim".
+
+    <xmp class="idl">
+    partial dictionary AuthenticationExtensionsClientInputs {
+      USVString confirmationPrompt;
```
emlun, line 7677: This property name should be the extension identifier defined above.
```suggestion
      USVString confirmation;
```

index.bs
```diff
@@ -7652,6 +7652,122 @@ To <dfn abstract-op>Create a new supplemental public key record</dfn>, perform t
...
+    <dl dfn-type="dict-member" dfn-for="CollectedClientAdditionalConfirmationData">
+        :  <dfn>confirmationPrompt</dfn> member
+        :: The full context of the confirmation to give.
+    </dl>
+
+: Client extension output
+:: Returns the authenticator extension output string UTF-8 decoded into a USVString.
+    <xmp class="idl">
+    partial dictionary AuthenticationExtensionsClientOutputs {
+      USVString confirmationPrompt;
```
emlun, line 7728: This property name should equal the extension identifier defined above.
```suggestion
      USVString confirmation;
```

index.bs
```diff
@@ -7652,6 +7652,122 @@ To <dfn abstract-op>Create a new supplemental public key record</dfn>, perform t
...
+    };
+    </xmp>
+
+: Authenticator extension input
+:: The client provides the extension input encoded as a CBOR text string (major type 3).
+
+<pre>
+CDDL:
+confirmationPromptInput = (tstr)
+</pre>
```
emlun, line 7738: The key should be the extension identifier defined above, and this should define an extension to the `$$extensionInput` socket:

```suggestion
:: The [=client extension input=] encoded as a CBOR text string (major type 3).

    ```
    $$extensionInput //= (
      confirmation: tstr,
    )
    ```
```

index.bs
```diff
@@ -7652,6 +7652,122 @@ To <dfn abstract-op>Create a new supplemental public key record</dfn>, perform t
...
+    presence=]. The authenticator MAY wrap lines that are too wide to be shown if needed.
+
+: Authenticator extension output
+:: If the authenticator supports this extension (and only then): A single CBOR string, representing the
+    confirmation prompt.
+
+<pre>
+CDDL:
+confirmationPromptOutput = (tstr)
+</pre>
```
emlun, line 7752: The first sentence is implicit and can be dropped. The key should be the extension identifier defined above, and this should define an extension to the `$$extensionInput` socket:


```suggestion
:: A single CBOR text string, representing the confirmation prompt.

    ```
    $$extensionOutput //= (
        confirmation: tstr,
    )
    ```
```

index.bs
```diff
@@ -7652,6 +7652,122 @@ To <dfn abstract-op>Create a new supplemental public key record</dfn>, perform t
...
+
+#### `confirmation` Extension Output Verification Procedures #### {#sctn-confirmation-extension-verification}
+
+Verifying the <code>[=confirmation=]</code> extension output is performed by the [=[RP]=] whenever a the use of <code>[=confirmatione=]</code> was requested.
+
+##### Authentication (`get()`) ##### {#sctn-confirmation-extension-verification-get}
+
+This extension is only specified to be used in conjunction with a {{CredentialsContainer/get()|navigator.credentials.get()}} call.
+
+If the [=[RP]=] requested the `confirmation` extension in a {{CredentialsContainer/get()|navigator.credentials.get()}} call, then the below verification step are performed in the context of [step 19](#authn-ceremony-verify-extension-outputs)
```
emlun, line 7762: While used in the `supplementalPubKeys` extension, this section structure is not (yet?) a general pattern, so I think we can just collapse these to one. We also shouldn't need to restate here that it's only relevant for assertions.

```suggestion
#### `confirmation` Extension Output Verification Procedure #### {#sctn-confirmation-extension-verification}

The below verification step are performed in the context of [step 19](#authn-ceremony-verify-extension-outputs)
```

index.bs
emlun: On second thought: why is a client extension output needed? It's not signed, unlike the `collectedClientData` entry and the authenticator extension output, so the RP should rather be using those for verification, no?

If so, the extension output should just be `true` to indicate that the extension was respected.

```
: Client extension output
:: Returns the value [TRUE] to indicate to the [=[RP]=] that the extension was acted upon.
    <xmp class="idl">
      boolean confirmation;
    };
    </xmp>
```

index.bs
```diff
@@ -7652,6 +7652,122 @@ To <dfn abstract-op>Create a new supplemental public key record</dfn>, perform t
...
+of [[#sctn-verifying-assertion]] using these variables established therein: |credential|, |clientExtensionResults|, |authData|, |hash|, and |credentialRecord|.
+[=[RP]=] policy may specify whether a response without a `confirmation` extension output is acceptable.
+
+1. Verify that the {{AuthenticationExtensionsClientOutputs/confirmation}} member of |clientExtensionResults| exists.
+1. Verify that the {{AuthenticationExtensionsClientOutputs/confirmation}} member contains the confirmation prompt that is expected (i.e., that was used when requesting the `confirmation` extension.
```
emlun, line 7767: As noted earlier, this should rather be using the authenticator extension output or the `clientDataJSON` entry, right? Maybe look for the authenticator extension output first, and fall back to the `clientDataJSON` entry if the authenticator extension output is absent?

```suggestion
[=[RP]=] policy may specify whether a response without a `confirmation` extension output is acceptable.

1. Verify that the `confirmation` key exists in the [=authData/extensions=] in |authData|.
1. Verify that the `confirmation` value in the [=authData/extensions=] in |authData| equals the confirmation prompt that is expected (i.e., that was used when requesting the `confirmation` extension).
1. (TODO: Fall back to <code>|credential|.{{PublicKeyCredential/response}}.{{AuthenticatorResponse/clientDataJSON}}</code> entry if authenticator extension output is absent?)
```

index.bs
```diff
@@ -9256,6 +9372,22 @@ for their contributions as our W3C Team Contacts.
...
+  },
+
+  "RFC9053": {
+    "authors": ["Jim Schaad"],
+    "title": "CBOR Object Signing and Encryption (COSE): Initial Algorithms",
+    "href": "https://datatracker.ietf.org/doc/rfc9053/",
+    "status": "RFC Informational",
+    "date": "August 2022"
+  },
+
```
emlun, line 9390: These are already in the Bikeshed index, and aren't referenced in this extension anyway. Please revert this change.

emlun (changes requested):

petrdvorak:
To continue our efforts to implement WYSIWYS in the connected authenticator: We found that besides not having `txAuthSimple`, even `hmac-secret` is not implemented as expected.

We were a bit desperate, as we really needed to get the data to the connected authenticator device so that the user could confirm it... and we managed to hack it.

By only supporting non-discoverable credentials, we get the list of `allowCredential` in WebAuthn, and - voila - this is the only data bundle we can pass in the authenticator and obtain it without modifications. So, we now append our challenge to the credential ID. Our authenticator can interpret this. Other authenticators don't need to care. Credential ID is not them. We then also compute a proprietary signature response (still conforming to ES256), binding the "plain challenge" we get in `allowCredential` list to the "hashed challenge" constructed from ClientDataJSON by the WebAuthn protocol... Cool, it works. It is ugly. But it works. We can do better...

Introducing "WebAuthn2: This time, we will get it right.", starring Steven Segal.

The signable challenge is in Verifiable Credentials format so that it is signed by the relying party, and the browser (or connected authenticator) can display it, as it is a standard format. Then, signing the challenge in the authenticator (platform or cross-platform) results in a Verifiable Presentation.

Is there any effort alongside this outline?

index.bs
rlin1: done

index.bs
rlin1: Done.

index.bs
rlin1: Done.

index.bs
rlin1: Done.

index.bs
rlin1: Done.

index.bs
```diff
@@ -7652,6 +7652,122 @@ To <dfn abstract-op>Create a new supplemental public key record</dfn>, perform t
...
+: Client extension processing
+:: If this extension is present, the client SHALL
+       1. use a dialog to the user that makes the user aware of the confirmation to be provided (as opposed to doing a simple sign in).
+       1. display the confirmation prompt to the user. The client SHOULD
+             indicate that the confirmation prompt originates from a specific relying party
+             (as opposed to the platform itself).
+     
+       1. use the {{CollectedClientConfirmationData}} structure containing the confirmation prompt instead of using
+             the {{CollectedClientData}} structure.
```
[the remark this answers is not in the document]
rlin1: Would you define the CollectedClientDataExtensions in this section (regarding confirmation prompt) or add that to the original location where CollectedClientData is defined?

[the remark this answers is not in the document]
rlin1: See latest version (a175ab9): Is this what you proposed?

index.bs
```diff
@@ -7583,6 +7587,94 @@ To <dfn abstract-op>Create a new supplemental public key record</dfn>, perform t
...
+: Client extension processing
+::   1. use a dialog to the user that makes the user aware of the confirmation to be provided (as opposed to doing a simple sign in).
+     1. display the confirmation prompt to the user. The client SHOULD
+           indicate that the confirmation prompt originates from a specific relying party
+           (as opposed to the platform itself).
+     
+     1. use the {{CollectedClientConfirmationData}} structure containing the confirmation prompt instead of using the {{CollectedClientData}} structure.
+     
+     1. pass-through the extension to the authenticator (see "client extension output" below)
+     1. pass-through the "authenticator extension output" to the caller as part of the assertion
```
MasterKale, line 7625: Nitpick: sentences should start with capital letters.

index.bs
```diff
@@ -7583,6 +7587,94 @@ To <dfn abstract-op>Create a new supplemental public key record</dfn>, perform t
...
+Example uses cases could be "I want to move $1234 from account A to account B" or "I want to share my health data with hospital X".
+
+: Extension identifier
+:: `confirmation`
+
+: Operation applicability
+:: [=authentication extension|Authentication=]
+
+: Client extension input
+:: A single USVString confirmationPrompt.  This string might contain the following formatting tags: <code>&lt;b&gt;&lt;/b&gt;</code> to mark the text inbetween as bold and <code>&lt;br /&gt;</code> to force a line break.  Support of these formatting tags is "best effort".  Implementations not supporting it SHALL NOT display the tags in "verbatim".
```
MasterKale, line 7608: I think this needs to get trimmed down to just `confirmation`:
```suggestion
:: A single USVString confirmation.  This string might contain the following formatting tags: <code>&lt;b&gt;&lt;/b&gt;</code> to mark the text inbetween as bold and <code>&lt;br /&gt;</code> to force a line break.  Support of these formatting tags is "best effort".  Implementations not supporting it SHALL NOT display the tags in "verbatim".
```

index.bs
```diff
@@ -7685,6 +7689,101 @@ To <dfn abstract-op>Create a new supplemental public key record</dfn>, perform t
...
+
+#### `confirmation` Extension Output Verification Procedures #### {#sctn-confirmation-extension-verification}
+
+The following verification steps are performed in the context of [step 19](#authn-ceremony-verify-extension-outputs)
+of [[#sctn-verifying-assertion]] using these variables established therein: |credential|, |clientExtensionResults|, |authData|, |hash|, and |credentialRecord|.
+[=[RP]=] policy may specify whether a response without a `confirmation` extension output is acceptable.
+
+1. Verify that the `confirmation` key exists in the [=authData/extensions=] in |authData|.
+1. Verify that the `confirmation` value in the [=authData/extensions=] in |authData| equals the confirmation prompt that is expected (i.e., that was used when requesting the `confirmation` extension).
+1. If processing a response without the `confirmation` extension is acceptable by policy: Fall back to <code>|credential|.{{PublicKeyCredential/response}}.{{AuthenticatorResponse/clientDataJSON}}</code> entry if authenticator extension output is absent.
```
jaiminbhatt7, line 7784: Copy-pasting offline chat with @rlin1 

> How does the extension processing work when `uv` is `preferred` or `required`. Does the authenticator/client need to show two prompts?

It is up to the relying party to handle the different user verification options appropriately. This includes appropriate settings for userVerification and appropriate processing of UV and UP. This confirmation extension doesn’t change the way user verification is handled by the client platform / authenticator. Have added a note regarding this.

> If the authenticator does not support displaying the confirmation prompt but supports signing the confirmation, does the browser/OS take over this responsibility? Do we need corresponding changes in CTAP to let providers indicate to the browser/OS whether this extension is fully/partially supported.

In that case the authenticator doesn’t support the confirmation extension. The client platform will add the confirmation details to the collectedClientData structure so that *any* authenticator would just sign over it- The relying arty can detect that by not finding the confirmation extension in the assertion, but only seeing the confirmation details in the collectedClientData.

> “If processing a response without the `confirmation` extension is acceptable by policy” – Not sure what this means. How does an RP specify that confirmation prompt is optional. And if the confirmation is optional then it does not provide auditable evidence that consent was visible to the user as you stated below.

It is up to the relying party to decide whether or not to accept an assertion without the confirmation extension (but with confirmation details in the collected clientData).  This situation indicates that a privileged component (i.e. client platform) has shown the confirmation details, but it wasn’t done by the authenticator itself (which could be an attested component).

jaiminbhatt7: I think the approach of letting RPs decide whether to accept an assertion without confirmation makes sense but its not clearly called out in the PR. We should further clarify here that if authenticator does not support the confirmation extension, the confirmation string is still signed over.

rlin1: Should be addressed by commit 9f2723f3cbba6f0d785200a97c132518ee92fbf4

mage28:
> 2025-03-26 WG call: this hasn't been discussed in quite some time, and the last time it was discussed there were concerns around untrusted text in a trusted UI. Planning to close this at the next meeting unless there is activity.

That does not sound good. Does it mean trying putting it to the backlog again? Concerns are still from Browsers vendors?  Concerns on security issues even with plain text? From the users perspective, the idea behind is to provide the best possible option to provide information what actually is authenticated/signed. Can we perhaps more details on concerns or impact? It really hinders the introduction of FIDO2/Passkeys as an alternative method of transaction signing.

petrdvorak:
I fully agree with @mage28 — building a challenge-response protocol support that does not allow showing the challenge in the authenticator and actively defending the decision requires skills.

- Why is this not a problem for Secure Payment Confirmation? Or will this be canceled, too? 
- Would providing signed content (like verifiable credentials) help?

mage28:
There is also a recent public statement from the german DK / GBIC proposing their interest in this feature:
https://die-dk.de/en/topics/article/gbic-statement-secure-display-transaction-data-fido2-standard/

petrdvorak:
@mage28 Let's drop it from this and standardize [our approach](https://developers.wultra.com/components/talisman/develop/documentation/Visual-Transaction-Signing), based on mapping the data payload on credential ID. It only works for cross-platform authenticators, but for banking, this is what we are currently working for. It is stable, and can be implemented by your authenticators without impacting any other authenticators.

ve7jtb:
This has no impact on Secure Payment confirmation.   That W3C specification is built on top of the existing WebAuthn spec not using these extensions.   The browsers are OK with displaying that as the transaction information is tightly scoped and presented to the user as financial transaction information.    That is different from having freeform text presented to the user during an authentication interaction.   Part of the problem is getting the user to understand that WebAuthn now is doing two things 1) authentication and now 2) Digital signature.  Could it be done, yes, but it requires significant UX work to make it understandable.

Work is being done in the EWC Large Scale Pilot and other pilots to to do transaction signing based on Verifiable Credentials.   That is using transaction signing as part of openID4VP over the W3C Digital Crtedentials API.   That API is also part of the Credential Management API that WebAuthn is under.    That work also is also not use the Tx Signing extension.

These extensions don't exist in practice and never have been deployed in WebAuthn (yes there was a short lived Android experiment that is the exception).  They were part of an older Fido UAF specification that were documented but never implemented for users.

The GBIC probably wants something more specific than general transaction signing, like Secure Transaction Confirmation. I think that could be looked at. I would like to understand why Secure Transaction Confirmation is not good enough for them before starting on something else. I also happen to be working with SPRIN-D on using Digital credentials for financial transactions. So, Germany is looking at multiple things in parallel.  

@petrdvorak I would be cautious with this approach.  It won't work with discoverable credentials, and large credential ID currently trigger bugs in iOS and MacOS processing of allow and exclude lists(I hope that gets fixed soon).  It likely also violates CTAP requirements for certification ( This is a Fido discussion).    Authenticators have tried similar things in the past and have been blocked by Chrome and other browsers.    Just giving you a heads up.

Given that the Digital Credentials API is working on Transaction signing and other qualified digital signature applications, trying to retrofit something into WebAuthn may just add to the confusion.

petrdvorak:
@ve7jtb One more comment to:

> I would like to understand why Secure Transaction Confirmation is not good enough for them before starting on something else.

SPC is great for e-commerce purchases online but these are barely all transactions that are out there. Banks must apply SCA to all payment types, including:

- Domestic account to account payments
- Foreign payments
- Bulk payments
- Collections
- Standing orders (establishment)
- Payments to contact
- ... any other payment type

Besides that, banks must apply SCA for new SCA element binding, meaning that "adding a new authenticator with the following parameters" is a mandatory transaction to cover.

Banks also typically use SCA for other use-cases, such as:

- Change of card limits
- Enabling e-commerce payments on the card
- Setting up or modifying an insurance
- Change of other credentials
- Adding another person to account
- Changing registered contact info (address, email, phone number)

... our typical customer has around 200 operation types with different attributes.

So, we would need a slight extension for SPC...

timcappalli:
2025-04-02: there is no current "sponsor" for this work actively engaged in the working group. These use cases are being actively discussed in the context of Digital Payment Credentials (DPCs) with the Digital Credentials API and OpenID4VP.

PR #2021, opened 2024-02-12
Acceptable UV caching time affordance extension added

rlin1:
The current behavior of some Passkey Providers to “cache” the user verification status cannot be modeled in FIDO/WebAuthn today and hence might surprise relying parties.

This PR adds the user verification extension that allows the RP to indicate its willingness to accept cached user verification for a given time.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/pull/2021.html" title="Last updated on Mar 27, 2024, 9:19 PM UTC (f342836)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/2021/73b3562...f342836.html" title="Last updated on Mar 27, 2024, 9:19 PM UTC (f342836)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -7600,7 +7600,7 @@ For example: Do not ask the user for a fresh [=user verification=] for sign-in i
 :: The maxTimeSinceLastUV denotes the maximum acceptable number of milliseconds elapsed since the last time the user was successfully verified.
     <xmp class="idl">
     partial dictionary AuthenticationExtensionsClientInputs {
-      uint maxTimeSinceLastUV;
+      unsigned long long maxTimeSinceLastUV;
     };
     </xmp>
 
@@ -7611,7 +7611,7 @@ For example: Do not ask the user for a fresh [=user verification=] for sign-in i
 :: Returns the number of milliseconds elapsed since the last time the user was successfully verified as returned by the [=authenticator=].
     <xmp class="idl">
     partial dictionary AuthenticationExtensionsClientOutputs {
-      uint timeSinceLastUV;
+      unsigned long long timeSinceLastUV;
     };
     </xmp>
 
--- a/index.bs
+++ b/index.bs
@@ -7580,7 +7580,7 @@ To <dfn abstract-op>Create a new supplemental public key record</dfn>, perform t
         [=set/append=] this [=supplemental public key record=] to |credentialRecord|.[$credential record/supplementalPubKeys$].
 
 
-### User verification caching extension (<dfn>userVerificationCaching</dfn>) ### {#sctn-user-verification-caching-extension}
+### User verification caching extension (userVerificationCaching) ### {#sctn-user-verification-caching-extension}
 
 In some cases it is good enough for the [=[RP]=] to know whether the user was verified by the authenticator "recently".
 
--- a/index.bs
+++ b/index.bs
@@ -7588,8 +7588,6 @@ This extension allows the [=[RP]=] to specify such [=user verification=] caching
 
 For example: Do not ask the user for a fresh [=user verification=] for sign-in if the user was verified by this authenticator within the past 300 seconds.
 
-#### Extension Definition #### {#sctn-user-verification-cacing-extension-definition}
-
 : Extension identifier
 :: `userVerificationCaching`
 
--- a/index.bs
+++ b/index.bs
@@ -9244,22 +9244,6 @@ for their contributions as our W3C Team Contacts.
     "date": "June 2019"
   },
 
-  "RFC9052": {
-    "authors": ["Jim Schaad"],
-    "title": "CBOR Object Signing and Encryption (COSE): Structures and Process",
-    "href": "https://datatracker.ietf.org/doc/rfc9052/",
-    "status": "IETF Internet Standard",
-    "date": "August 2022"
-  },
-
-  "RFC9053": {
-    "authors": ["Jim Schaad"],
-    "title": "CBOR Object Signing and Encryption (COSE): Initial Algorithms",
-    "href": "https://datatracker.ietf.org/doc/rfc9053/",
-    "status": "RFC Informational",
-    "date": "August 2022"
-  },
-
   "ISOBiometricVocabulary": {
     "authors": ["ISO/IEC JTC1/SC37"],
     "title": "Information technology — Vocabulary — Biometrics",
--- a/index.bs
+++ b/index.bs
@@ -7582,14 +7582,11 @@ To <dfn abstract-op>Create a new supplemental public key record</dfn>, perform t
 
 ### User verification caching extension (<dfn>userVerificationCaching</dfn>) ### {#sctn-user-verification-caching-extension}
 
-In some cases it is good enough for the [=RP=] to know whether
-    the user was verified by the authenticator "recently".
-    This extension allows the [=RP=] to specify such [=user verification=]
-    caching time, i.e. the time for which the
-    [=user verification=] status can be "cached" by the [=authenticator=].
+In some cases it is good enough for the [=RP=] to know whether the user was verified by the authenticator "recently".
 
-For example: Do not ask the user for a fresh [=user verification=] for sign-in
-    if the user was verified by this authenticator within the past 300 seconds.
+This extension allows the [=RP=] to specify such [=user verification=] caching time, i.e. the time for which the [=user verification=] status can be "cached" by the [=authenticator=].
+
+For example: Do not ask the user for a fresh [=user verification=] for sign-in if the user was verified by this authenticator within the past 300 seconds.
 
 #### Extension Definition #### {#sctn-user-verification-cacing-extension-definition}
 
@@ -7600,10 +7597,10 @@ For example: Do not ask the user for a fresh [=user verification=] for sign-in
 :: [=authentication extension|authentication=]
 
 : Client extension input
-:: The maximum [=user verification=] caching time denotes the maximum acceptable number of seconds elapsed since the last time the user was successfully verified.
+:: The maxTimeSinceLastUV denotes the maximum acceptable number of milliseconds elapsed since the last time the user was successfully verified.
     <xmp class="idl">
     partial dictionary AuthenticationExtensionsClientInputs {
-      unsigned short maxUVC;
+      uint maxTimeSinceLastUV;
     };
     </xmp>
 
@@ -7611,33 +7608,31 @@ For example: Do not ask the user for a fresh [=user verification=] for sign-in
 :: None, except creating the [=authenticator extension input=] from the client extension input.
 
 : Client extension output
-:: Returns a JSON object denoting the maximum acceptable number of seconds elapsed since the last time the user was successfully verified as returned by the [=authenticator=].
+:: Returns the number of milliseconds elapsed since the last time the user was successfully verified as returned by the [=authenticator=].
     <xmp class="idl">
     partial dictionary AuthenticationExtensionsClientOutputs {
-      unsigned short maxUVC;
+      uint timeSinceLastUV;
     };
     </xmp>
 
 : Authenticator extension input
-:: The maxUVC time in seconds, encoded in CBOR.
+:: The maximum acceptable time in milliseconds elapsed since last user verification, encoded in CBOR.
 
     ```
     $$extensionInput //= (
-      maxUVC: uint .size 2
+      mtslUV: uint .size 4
     )
     ```
 
 : Authenticator extension processing
-:: When user verification is requested, the [=authenticator=] triggers user verification only if more seconds have elapsed since the last time the user was verified
-    than indicated by the maxUVC extension.
+:: When user verification is requested, the [=authenticator=] triggers user verification only if more milliseconds have elapsed since the last time the user was verified than indicated by the maxTimeSinceLastUV value in the extension.
 
 : Authenticator extension output
-:: If no fresh user verification needed to be triggered triggered, the authenticator reports the maxUVC time back to the [=RP=] to ensure the [=RP=] is aware
-    that no fresh user verification was triggered.
+:: If no fresh user verification needed to be triggered triggered, the authenticator reports the time last last user verification time back to the [=RP=] to ensure the [=RP=] is aware that no fresh user verification was triggered.  It is up to the authenticator to decide whether to return the real elapsed time, or a "rounded" value. If user verification was requested, this value SHALL not exceed the value originally provided in the extension input.
 
     ```
     $$extensionOutput //= (
-      maxUVC: uint .size 2
+      tslUV: uint .size 4
     )
     ```
 
--- a/index.bs
+++ b/index.bs
@@ -7580,6 +7580,68 @@ To <dfn abstract-op>Create a new supplemental public key record</dfn>, perform t
         [=set/append=] this [=supplemental public key record=] to |credentialRecord|.[$credential record/supplementalPubKeys$].
 
 
+### User verification caching extension (<dfn>userVerificationCaching</dfn>) ### {#sctn-user-verification-caching-extension}
+
+In some cases it is good enough for the [=RP=] to know whether
+    the user was verified by the authenticator "recently".
+    This extension allows the [=RP=] to specify such [=user verification=]
+    caching time, i.e. the time for which the
+    [=user verification=] status can be "cached" by the [=authenticator=].
+
+For example: Do not ask the user for a fresh [=user verification=] for sign-in
+    if the user was verified by this authenticator within the past 300 seconds.
+
+#### Extension Definition #### {#sctn-user-verification-cacing-extension-definition}
+
+: Extension identifier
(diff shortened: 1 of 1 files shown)


timcappalli:
@rlin1 we talked about this extensively at the WebAuthn F2F last year. The direction we were discussing was adding an ability for an authenticator to convey when it last performed UV and the authenticator would be truthful in its UV response in authData. This would provide the balance of an RP knowing the UV response is truthful along with some additional context, and the client and/or provider being able to make the best decision based on context they know.

This extension could potentially be part of that solution, but it is unclear whether the authenticator would reply truthfully about UV for the assertion or would lie based on the caching being acceptable to the RP.

rlin1:
With this approach, the maxUVC included in the authenticator output could be defined differently to be less or equal to the maxUVC as provided by the RP - so that the authenticator could set it to the elapsed time.
The Authenticator today already indicates whether UV was performed.  The current expectation is that this is fresh user verification.  I think it would be good to let the RP indicate whether UV caching is acceptable for the specific use case.

rlin1:
Close #2023

rlin1:
Should this extension cover userPresence as well? If yes: we might want to rename it to userGestureCaching...

index.bs
```diff
@@ -7580,6 +7580,68 @@ To <dfn abstract-op>Create a new supplemental public key record</dfn>, perform t
...
+### User verification caching extension (<dfn>userVerificationCaching</dfn>) ### {#sctn-user-verification-caching-extension}
+
+In some cases it is good enough for the [=RP=] to know whether 
+    the user was verified by the authenticator "recently".
+    This extension allows the [=RP=] to specify such [=user verification=]
+    caching time, i.e. the time for which the
+    [=user verification=] status can be "cached" by the [=authenticator=].
+
+For example: Do not ask the user for a fresh [=user verification=] for sign-in 
+    if the user was verified by this authenticator within the past 300 seconds.
```
emlun, line 7592: This should use the `[RP]` macro and remove the leading indentation.

```suggestion
In some cases it is good enough for the [=[RP]=] to know whether 
the user was verified by the authenticator "recently".
This extension allows the [=[RP]=] to specify such [=user verification=]
caching time, i.e. the time for which the
[=user verification=] status can be "cached" by the [=authenticator=].

For example: Do not ask the user for a fresh [=user verification=] for sign-in 
if the user was verified by this authenticator within the past 300 seconds.
```

index.bs
```diff
@@ -7580,6 +7580,68 @@ To <dfn abstract-op>Create a new supplemental public key record</dfn>, perform t
...
+#### Extension Definition #### {#sctn-user-verification-cacing-extension-definition}
+
+: Extension identifier
+:: `userVerificationCaching`
+
+: Operation applicability
+:: [=authentication extension|authentication=]
+
+: Client extension input
+:: The maximum [=user verification=] caching time denotes the maximum acceptable number of seconds elapsed since the last time the user was successfully verified.
```
emlun, line 7603: Using seconds here would differ from the `timeout` parameter which is defined in milliseconds. I think we should use stay with a single unit throughout the whole API to avoid confusion (absent other signals like time unit-aware types or a unit suffix naming convention, neither of which we have in place).

That probably also means the integer needs to be 32 bits wide rather than 16.

index.bs
```diff
@@ -7580,6 +7580,68 @@ To <dfn abstract-op>Create a new supplemental public key record</dfn>, perform t
...
+    <xmp class="idl">
+    partial dictionary AuthenticationExtensionsClientInputs {
+      unsigned short maxUVC;
```
emlun, line 7606: The extension input property should have the extension identifier as its name.

```suggestion
      unsigned short userVerificationCaching;
```

index.bs
```diff
@@ -7580,6 +7580,68 @@ To <dfn abstract-op>Create a new supplemental public key record</dfn>, perform t
...
+    </xmp>
+
+: Client extension processing
+:: None, except creating the [=authenticator extension input=] from the client extension input.
+
+: Client extension output
+:: Returns a JSON object denoting the maximum acceptable number of seconds elapsed since the last time the user was successfully verified as returned by the [=authenticator=].
+    <xmp class="idl">
+    partial dictionary AuthenticationExtensionsClientOutputs {
+      unsigned short maxUVC;
```
emlun, line 7617: Ditto:
```suggestion
:: Returns a JSON object denoting the maximum acceptable number of milliseconds elapsed since the last time the user was successfully verified as returned by the [=authenticator=].
    <xmp class="idl">
    partial dictionary AuthenticationExtensionsClientOutputs {
      unsigned short userVerificationCaching;
```

emlun: Also, this output is not a JSON object, just an integer. I think we can just reference the authenticator output here.

```suggestion
:: The authenticator extension output parsed as an integer value.
    <xmp class="idl">
    partial dictionary AuthenticationExtensionsClientOutputs {
      unsigned long userVerificationCaching;
```

index.bs
```diff
@@ -7580,6 +7580,68 @@ To <dfn abstract-op>Create a new supplemental public key record</dfn>, perform t
...
+    </xmp>
+
+: Authenticator extension input
+:: The maxUVC time in seconds, encoded in CBOR.
+
+    ```
+    $$extensionInput //= (
+      maxUVC: uint .size 2
+    )
+    ```
```
emlun, line 7628: Ditto:
```suggestion
:: The client extension input encoded as a CBOR unsigned integer.

    ```
    $$extensionInput //= (
      userVerificationCaching: uint .size 4
    )
    ```
```

emlun: I see here that this highlights that perhaps we should instead change the extension identifier to a shorter value such as `maxUVC` or `uvCache`, to save space in authenticator protocols.

index.bs
```diff
@@ -7580,6 +7580,68 @@ To <dfn abstract-op>Create a new supplemental public key record</dfn>, perform t
...
+
+: Authenticator extension processing
+:: When user verification is requested, the [=authenticator=] triggers user verification only if more seconds have elapsed since the last time the user was verified 
+    than indicated by the maxUVC extension.
+
+: Authenticator extension output
+:: If no fresh user verification needed to be triggered triggered, the authenticator reports the maxUVC time back to the [=RP=] to ensure the [=RP=] is aware
+    that no fresh user verification was triggered.
```
emlun, line 7636: What happens if the time elapsed is less than the requested limit?

index.bs
```diff
@@ -9189,6 +9252,22 @@ for their contributions as our W3C Team Contacts.
...
+  },
+
+  "RFC9053": {
+    "authors": ["Jim Schaad"],
+    "title": "CBOR Object Signing and Encryption (COSE): Initial Algorithms",
+    "href": "https://datatracker.ietf.org/doc/rfc9053/",
+    "status": "RFC Informational",
+    "date": "August 2022"
+  },
+
```
emlun, line 9270: These don't seem to be referenced anywhere in the added text; was this added by mistake?

index.bs
```diff
@@ -8786,6 +8848,7 @@ These recommendations serve to prevent an adversary with physical access to an [
 
 ## Privacy considerations for [=clients=] ## {#sctn-privacy-considerations-client}
 
+
```
emlun, line 8851: Please amend the commit to remove the spurious whitespace changes here and on lines 325, 368, 8378 and 8477.

emlun (changes requested):

index.bs
rlin1: no additional UV would be triggered.  And the authenticator could decide whether to return the measured time since last UV or (for privacy reasons) a "rounded" value, but not exceeding the maxUV value provided by the relying party.

index.bs
rlin1: was required to make bikeshed run successfully

index.bs
rlin1: done.

index.bs
rlin1: done

index.bs
rlin1: done

index.bs
emlun: How are you running Bikeshed? You may need to run `bikeshed update` to update your snapshot of the bibliography index.

emlun:
I force-pushed the branch to remove some unrelated whitespace changes from the commit history; you may need to hard-reset any local clones to the new HEAD commit.

index.bs
```diff
@@ -7580,6 +7580,61 @@ To <dfn abstract-op>Create a new supplemental public key record</dfn>, perform t
...
+:: `userVerificationCaching`
+
+: Operation applicability
+:: [=authentication extension|authentication=]
+
+: Client extension input
+:: The maxTimeSinceLastUV denotes the maximum acceptable number of milliseconds elapsed since the last time the user was successfully verified.
+    <xmp class="idl">
+    partial dictionary AuthenticationExtensionsClientInputs {
+      unsigned long long maxTimeSinceLastUV;
```
emlun, line 7601: This member name needs to be the same as the extension identifier.
```suggestion
      unsigned long long userVerificationCaching;
```

index.bs
```diff
@@ -7580,6 +7580,61 @@ To <dfn abstract-op>Create a new supplemental public key record</dfn>, perform t
...
+    </xmp>
+
+: Client extension processing
+:: None, except creating the [=authenticator extension input=] from the client extension input.
+
+: Client extension output
+:: Returns the number of milliseconds elapsed since the last time the user was successfully verified as returned by the [=authenticator=].
+    <xmp class="idl">
+    partial dictionary AuthenticationExtensionsClientOutputs {
+      unsigned long long timeSinceLastUV;
```
emlun, line 7612: This member name needs to be the same as the extension identifier.

```suggestion
      unsigned long long userVerificationCaching;
```

index.bs
```diff
@@ -7580,6 +7580,61 @@ To <dfn abstract-op>Create a new supplemental public key record</dfn>, perform t
...
+    };
+    </xmp>
+
+: Authenticator extension input
+:: The maximum acceptable time in milliseconds elapsed since last user verification, encoded in CBOR.
+
+    ```
+    $$extensionInput //= (
+      mtslUV: uint .size 4
```
emlun, line 7621: This member name needs to be the same as the extension identifier.
```suggestion
      userVerificationCaching: uint .size 4
```

index.bs
```diff
@@ -7580,6 +7580,61 @@ To <dfn abstract-op>Create a new supplemental public key record</dfn>, perform t
...
+
+: Authenticator extension processing
+:: When user verification is requested, the [=authenticator=] triggers user verification only if more milliseconds have elapsed since the last time the user was verified than indicated by the maxTimeSinceLastUV value in the extension.
+
+: Authenticator extension output
+:: If no fresh user verification needed to be triggered triggered, the authenticator reports the time last last user verification time back to the [=[RP]=] to ensure the [=[RP]=] is aware that no fresh user verification was triggered.  It is up to the authenticator to decide whether to return the real elapsed time, or a "rounded" value. If user verification was requested, this value SHALL not exceed the value originally provided in the extension input.
+
+    ```
+    $$extensionOutput //= (
+      tslUV: uint .size 4
```
emlun, line 7633: This member name needs to be the same as the extension identifier.

```suggestion
      userVerificationCaching: uint .size 4
```

emlun (changes requested):

index.bs
```diff
@@ -7580,6 +7580,61 @@ To <dfn abstract-op>Create a new supplemental public key record</dfn>, perform t
...
+    $$extensionInput //= (
+      mtslUV: uint .size 4
+    )
+    ```
+
+: Authenticator extension processing
+:: When user verification is requested, the [=authenticator=] triggers user verification only if more milliseconds have elapsed since the last time the user was verified than indicated by the maxTimeSinceLastUV value in the extension.
+
+: Authenticator extension output
+:: If no fresh user verification needed to be triggered triggered, the authenticator reports the time last last user verification time back to the [=[RP]=] to ensure the [=[RP]=] is aware that no fresh user verification was triggered.  It is up to the authenticator to decide whether to return the real elapsed time, or a "rounded" value. If user verification was requested, this value SHALL not exceed the value originally provided in the extension input.
```
sbweeden, line 7629: What is the extension output if UV is done as part of the current ceremony? I know that the UV bit should be set in authenticatorData, but should the extension output also be set with a value of 0 perhaps?

timcappalli:
I'm largely good with this one, as long as it is restricted for use with UV=preferred, similar to #2052.

index.bs
MasterKale: To be clear, the suggestion here is that extension output _should_ be set to `0` if UV takes place during the ceremony? Which naturally means the extension output should never be greater than the duration specified as `maxTimeSinceLastUV`.

MasterKale:
I've been convinced that this proposal enables RP's to gracefully handle more use cases than the simpler reporting mechanism proposed in #2052. I'd say let's try and get this one into the spec instead.

timcappalli:
I'd also add that one of the two open questions for #2052 also applies here: 

> Technically an out of band vault unlock for passkey provider doesn't satisfy user verification as it was not part of a WebAuthn ceremony. How do we want to handle that?

index.bs
```diff
@@ -7580,6 +7580,61 @@ To <dfn abstract-op>Create a new supplemental public key record</dfn>, perform t
...
+For example: Do not ask the user for a fresh [=user verification=] for sign-in if the user was verified by this authenticator within the past 300 seconds.
+
+: Extension identifier
+:: `userVerificationCaching`
+
+: Operation applicability
+:: [=authentication extension|authentication=]
+
+: Client extension input
+:: The maxTimeSinceLastUV denotes the maximum acceptable number of milliseconds elapsed since the last time the user was successfully verified.
```
timcappalli, line 7598: Should we bound this with a min and max? (https://github.com/w3c/webauthn/issues/2034#issuecomment-1969890072)

index.bs
```diff
@@ -7580,6 +7580,61 @@ To <dfn abstract-op>Create a new supplemental public key record</dfn>, perform t
...
+    partial dictionary AuthenticationExtensionsClientInputs {
+      unsigned long long maxTimeSinceLastUV;
+    };
+    </xmp>
+
+: Client extension processing
+:: None, except creating the [=authenticator extension input=] from the client extension input.
+
+: Client extension output
+:: Returns the number of milliseconds elapsed since the last time the user was successfully verified as returned by the [=authenticator=].
```
timcappalli, line 7609: s/verified as/verified, as

index.bs
```diff
@@ -7580,6 +7580,61 @@ To <dfn abstract-op>Create a new supplemental public key record</dfn>, perform t
...
+:: The maximum acceptable time in milliseconds elapsed since last user verification, encoded in CBOR.
+
+    ```
+    $$extensionInput //= (
+      mtslUV: uint .size 4
+    )
+    ```
+
+: Authenticator extension processing
+:: When user verification is requested, the [=authenticator=] triggers user verification only if more milliseconds have elapsed since the last time the user was verified than indicated by the maxTimeSinceLastUV value in the extension.
```
timcappalli, line 7626: "If more milliseconds have elapsed since the last time the user was verified than indicated by the maxTimeSinceLastUV value in the extension, request [=user verification=]."

timcappalli (changes requested):

index.bs
```diff
@@ -7580,6 +7580,61 @@ To <dfn abstract-op>Create a new supplemental public key record</dfn>, perform t
...
+: Client extension output
+:: Returns the number of milliseconds elapsed since the last time the user was successfully verified as returned by the [=authenticator=].
+    <xmp class="idl">
+    partial dictionary AuthenticationExtensionsClientOutputs {
+      unsigned long long timeSinceLastUV;
+    };
+    </xmp>
+
+: Authenticator extension input
+:: The maximum acceptable time in milliseconds elapsed since last user verification, encoded in CBOR.
```
Firstyear, line 7617: What's missing here is the negative case - if this extension is *not* present, do *not* allow caching at all. This needs to be clearly stated in the spec.

index.bs
emlun: Yes, `unsigned long long` is the [WebIDL type for a 64-bit integer](https://webidl.spec.whatwg.org/#idl-unsigned-long-long).

...oh, I was about to say we use the same type for [`timeout`](https://w3c.github.io/webauthn/#dom-publickeycredentialcreationoptions-timeout), but it turns out it's just `unsigned long` (32-bit). So yeah, let's make it just `unsigned long` here too for consistency. 2<sup>64</sup> milliseconds is almost 50 days, I can't imagine anyone needing more range than that.

timcappalli:
2024-04-19 WebAuthn F2F meeting: consensus to put this PR (and its alternative #2052) on hold until some details around attestation and conformance in FIDO are ironed out.

nicksteele:
Closed due to reasons mentioned in #2034 but could revisit in the future

PR #2040, opened 2024-03-12
adds Related Origin Requests

### LINKED ISSUE #1372: Consider allowing cross-domain credential use
This issue is to capture discussion / decision making related to a discussion opened on the mailing list:

https://lists.w3.org/Archives/Public/public-webauthn/2020Feb/0001.html

The summary of the request is to allow an RP to advertise an allowed list of other web origins at which a credential issued to the RPID controlled by the RP may be used.

For example consider a credential created via WebAuthn at a site login.a.com, using RPID a.com.

The authority in control of a.com would like to permit use of this credential at other web origins such as b.com and c.com, without having to have separate credentials registered at those web domains.

This is similar in principal to the facet list feature of FIDO U2F. It has application within a company that controls different web domains (perhaps brands, or mergers and acquisitions).

Alternatives in this field require browser redirects or iframes that access both a.com (for authentication) and then perform federated SSO to b.com. This is an undersirable user experience in many cases.

A proposed solution could allow a.com (via some form of web discovery including via DNS, or a hosted page) to control the list of RPIDs or web origins that are permitted to use a.com's RPID in calls to navigator.credentials.get(). The solution requires the client (in this case the browser) to understand and discover this trust relationship, something that is already done for other web access control protocols today. I don't plan to prescribe a particular method here, just ask that one be considered.

This is already done for Android native apps today to permit sharing of a credential with a WebAuthn RPID via the assetlinks.json hosted discovery document. Clearly non-browser and browser clients want to be able to do this - the same *should* be able to apply for cross-domain browser-based use of credentials. Preferably the trust model is sharable across different client types so that we don't end up with every client type building their own bespoke model.

timcappalli:
Adds the Related Origin Requests feature ([explainer](https://github.com/w3c/webauthn/wiki/Explainer:-Related-origin-requests))

- Updates "Create a New Credential"
- Updates "Use an Existing Credential"
- Adds a new subsection: "Using Web Authentication across related origins"
- Adds a new definition for "Origin Label"
- Updates "Validating the origin of a credential"
- Updates "enum ClientCapability"

Resolves #1372

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/pull/2040.html" title="Last updated on Jul 17, 2024, 6:35 PM UTC (017a5e3)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/2040/c75b69e...017a5e3.html" title="Last updated on Jul 17, 2024, 6:35 PM UTC (017a5e3)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -4119,8 +4119,8 @@ To validate the calling origin is an authorized related origin for a given cerem
     1. Let |domain| be the [=effective domain=] of |url|. If that is null, [=continue=].
     1. Remove any [=public suffix=] from the end of |domain|, including private registries and unknown registries. If |domain| is now empty, [=continue=].
     1. Split |domain| into |labels| and let |label| be the right-most one.
-    1. If |labelsSeen| [=set/contains=] |label|, [=continue=].
     1. If |callerOrigin| and |url| are [=same origin=], return [TRUE].
+    1. If |labelsSeen| [=set/contains=] |label|, [=continue=].
     1. [=set/Append=] |label| to |labelsSeen|.
     1. If the [=set/size=] of |labelsSeen| is greater than or equal to |maxLabels|, [=break=].
 1. Return [FALSE].
--- a/index.bs
+++ b/index.bs
@@ -1836,7 +1836,8 @@ a numbered step. If outdented, it (today) is rendered as a bullet in the midst o
                         :   supports [[#sctn-related-origins|related origin requests]]
                         ::  1. Let |rpIdRequested| be the value of <code>|pkOptions|.{{PublicKeyCredentialCreationOptions/rp}}.{{PublicKeyCredentialRpEntity/id}}</code>.
 
-                            1. If the result of running the [[#sctn-validating-relation-origin|related origins validation procedure]] is [FALSE], throw a "{{SecurityError}}" {{DOMException}}.
+                            1. Run the [$related origins validation procedure$] with arguments |callerOrigin| and |rpIdRequested|.
+                                If the result is [FALSE], throw a "{{SecurityError}}" {{DOMException}}.
 
                         :   does not support [[#sctn-related-origins|related origin requests]]
                         ::  throw a "{{SecurityError}}" {{DOMException}}.
@@ -2340,7 +2341,8 @@ When this method is invoked, the user agent MUST execute the following algorithm
                             :   supports [[#sctn-related-origins|related origin requests]]
                             ::  1. Let |rpIdRequested| be the value of <code>|pkOptions|.{{PublicKeyCredentialRequestOptions/rpId}}</code>
 
-                                1. If the result of running the [[#sctn-validating-relation-origin|related origins validation procedure]] is [FALSE], throw a "{{SecurityError}}" {{DOMException}}.
+                                1. Run the [$related origins validation procedure$] with arguments |callerOrigin| and |rpIdRequested|.
+                                    If the result is [FALSE], throw a "{{SecurityError}}" {{DOMException}}.
 
                             :   does not support [[#sctn-related-origins|related origin requests]]
                             ::  throw a "{{SecurityError}}" {{DOMException}}.
@@ -4115,7 +4117,7 @@ For example, for the RP ID `example.com`:
 
 ### Validating Related Origins ### {#sctn-validating-relation-origin}
 
-To validate the calling origin is an authorized related origin for a given ceremony:
+The <dfn abstract-op>related origins validation procedure</dfn>, given arguments |callerOrigin| and |rpIdRequested|, is as follows:
 
 1. Let |maxLabels| be the maximum number of [=registrable origin labels=] allowed by client policy.
 1. Fetch the well-known URL for the RP ID (<code>https://|rpIdRequested|/.well-known/webauthn</code>).
--- a/index.bs
+++ b/index.bs
@@ -1839,7 +1839,7 @@ a numbered step. If outdented, it (today) is rendered as a bullet in the midst o
                         
                                 1. If the fetch fails, the response does not have a content type of `application/json`, or does not have a status code (after following redirects) of 200, then throw a "{{SecurityError}}" {{DOMException}}.
                                 1. If the body of the resource is not a valid JSON object, then throw a "{{SecurityError}}" {{DOMException}}.
-                                1. If the value of the |origins| member of the JSON object is missing, or is not a list of strings, then throw a "{{SecurityError}}" {{DOMException}}.
+                                1. If the value of the |origins| property of the JSON object is missing, or is not an array of strings, then throw a "{{SecurityError}}" {{DOMException}}.
     
                             1. Let |originLabelsSeen| be an empty set.
                             1. [=set/For each=] string in |origins|:
@@ -2360,7 +2360,7 @@ When this method is invoked, the user agent MUST execute the following algorithm
 
                                     1. If the fetch fails, the response does not have a content type of `application/json`, or does not have a status code (after following redirects) of 200, then throw a "{{SecurityError}}" {{DOMException}}.
                                     1. If the body of the resource is not a valid JSON object, then throw a "{{SecurityError}}" {{DOMException}}.
-                                    1. If the value of the |origins| member of the JSON object is missing, or is not a list of strings, then throw a "{{SecurityError}}" {{DOMException}}.
+                                    1. If the value of the |origins| property of the JSON object is missing, or is not an array of strings, then throw a "{{SecurityError}}" {{DOMException}}.
 
                                 1. Let |originLabelsSeen| be an empty set.
                                 1. [=set/For each=] string in |origins|:
--- a/index.bs
+++ b/index.bs
@@ -4088,8 +4088,9 @@ A JSON document MUST be hosted at the `webauthn` well-known URL ([[!RFC8615]]) f
     - The top-level JSON object MUST contain a key named `origins` whose value MUST be an array of one or more strings containing web origins.
     - Origins with the same [=origin label=] SHOULD be listed together as the first elements in the array, with unique origins as the last elements.
 
-:: Note: When adding additional origins matching an existing [=origin label=], add them after the existing elements with the same [=origin label=] instead of at the end of the array.
-   This will ensure their processing will not impact the [=origin label=] count.
+    :: Note: When adding additional origins matching an existing [=origin label=], add them after the existing elements with the same [=origin label=] instead of at the end of the array.
+        This will ensure their processing will not impact the [=origin label=] count.
+
 
 For example, for the RP ID `example.com`:
 
--- a/index.bs
+++ b/index.bs
@@ -4117,9 +4117,9 @@ Relying Parties MUST choose a common [=RP ID=] to use across all ceremonies from
 
 A JSON document MUST be hosted at the `webauthn` well-known URL ([[!RFC8615]]) for the [=RP ID=]. The JSON document MUST be returned as follows:
 
-    - the content type must be `application/json`
-    - JSON document MUST contain a single object named `origins`
-    - the `origins` object MUST contain an array of one or more web origins
+    - The content type must be `application/json`.
+    - JSON document MUST contain a single object named `origins`.
+    - The `origins` object MUST contain an array of one or more web origins.
 
 For example, for the RP ID `example.com`:
 
--- a/index.bs
+++ b/index.bs
@@ -4086,6 +4086,10 @@ A JSON document MUST be hosted at the `webauthn` well-known URL ([[!RFC8615]]) f
 
     - The content type MUST be `application/json`.
     - The top-level JSON object MUST contain a key named `origins` whose value MUST be an array of one or more strings containing web origins.
+    - Origins with the same [=origin label=] SHOULD be listed together as the first elements in the array, with unique origins as the last elements.
+
+:: Note: When adding additional origins matching an existing [=origin label=], add them after the existing elements with the same [=origin label=] instead of at the end of the array.
+   This will ensure their processing will not impact the [=origin label=] count.
 
 For example, for the RP ID `example.com`:
 
@@ -4094,8 +4098,14 @@ For example, for the RP ID `example.com`:
     "origins": [
         "https://example.co.uk",
         "https://example.de",
-        "https://example.sg"
-        "https://myexamplerewards.com"
+        "https://example.sg",
+        "https://example.net",
+        "https://exampledelivery.com",
+        "https://exampledelivery.co.uk",
+        "https://exampledelivery.de",
+        "https://exampledelivery.sg",
+        "https://myexamplerewards.com",
+        "https://examplecars.com"
     ]
 }
 </xmp>
--- a/index.bs
+++ b/index.bs
@@ -4080,8 +4080,7 @@ By default, Web Authentication requires that the [=RP ID=] be equal to the [=det
 This can make deployment challenging for large environments where multiple country-specific domains are in use (e.g. example.com vs example.co.uk vs example.sg), where alternative or brand domains are required (e.g. myexampletravel.com vs examplecruises.com), and/or where platform as a service providers are used to support mobile apps. 
 
 [=[WRPS]=] can opt in to allowing [=WebAuthn Clients=] to enable a credential to be created and used across a limited set of related [=origin|origins=].
-
-Relying Parties MUST choose a common [=RP ID=] to use across all ceremonies from related origins.
+Such [=[RPS]=] MUST choose a common [=RP ID=] to use across all ceremonies from related origins.
 
 A JSON document MUST be hosted at the `webauthn` well-known URL ([[!RFC8615]]) for the [=RP ID=]. The JSON document MUST be returned as follows:
 
--- a/index.bs
+++ b/index.bs
@@ -1834,7 +1834,7 @@ a numbered step. If outdented, it (today) is rendered as a bullet in the midst o
                         ::  1. Let |rpIdRequested| be the value of <code>|pkOptions|.{{PublicKeyCredentialCreationOptions/rp}}.{{PublicKeyCredentialRpEntity/id}}</code>.
 
                             1. If the result of running the [[#sctn-validating-relation-origin|related origins validation procedure]] is [FALSE], throw a "{{SecurityError}}" {{DOMException}}.
-                            
+
                         :   does not support [[#sctn-related-origins|related origin requests]]
                         ::  throw a "{{SecurityError}}" {{DOMException}}.
                     </dl>
@@ -2337,13 +2337,7 @@ When this method is invoked, the user agent MUST execute the following algorithm
                             :   supports [[#sctn-related-origins|related origin requests]]
                             ::  1. Let |rpIdRequested| be the value of <code>|pkOptions|.{{PublicKeyCredentialRequestOptions/rpId}}</code>
 
-                                1. If the result of running the [[#sctn-validating-relation-origin|related origins validation procedure]] is
-                                    <dl class="switch">
-                                        :   `false`
-                                        ::  throw a "{{SecurityError}}" {{DOMException}}.
-                                        :   `true`
-                                        ::  [=break=].
-                                    </dl>
+                                1. If the result of running the [[#sctn-validating-relation-origin|related origins validation procedure]] is [FALSE], throw a "{{SecurityError}}" {{DOMException}}.
 
                             :   does not support [[#sctn-related-origins|related origin requests]]
                             ::  throw a "{{SecurityError}}" {{DOMException}}.
--- a/index.bs
+++ b/index.bs
@@ -4115,7 +4115,7 @@ To validate the calling origin is an authorized related origin for a given cerem
     1. If the fetch fails, the response does not have a content type of `application/json`, or does not have a status code (after following redirects) of 200, then throw a "{{SecurityError}}" {{DOMException}}.
     1. If the body of the resource is not a valid JSON object, then throw a "{{SecurityError}}" {{DOMException}}.
     1. If the value of the |origins| property of the JSON object is missing, or is not an array of strings, then throw a "{{SecurityError}}" {{DOMException}}.
-1. Let |originLabelsSeen| be an empty set.
+1. Let |labelsSeen| be a new empty [=set=].
 1. [=set/For each=] string in |origins|:
     1. Let |url| be the result of parsing the string as a URL. If that fails, [=continue=].
     1. Let |domain| be the [=effective domain=] of |url|. If that is null, [=continue=].
--- a/index.bs
+++ b/index.bs
@@ -4096,7 +4096,7 @@ Relying Parties MUST choose a common [=RP ID=] to use across all ceremonies from
 
 A JSON document MUST be hosted at the `webauthn` well-known URL ([[!RFC8615]]) for the [=RP ID=]. The JSON document MUST be returned as follows:
 
-    - The content type must be `application/json`.
+    - The content type MUST be `application/json`.
     - JSON document MUST contain a single object named `origins`.
     - The `origins` object MUST contain an array of one or more web origins.
 
--- a/index.bs
+++ b/index.bs
@@ -4109,7 +4109,7 @@ To override this default policy and indicate that a cross-origin <{iframe}> is a
 
 By default, Web Authentication requires that the [=RP ID=] be equal to the [=determines the set of origins on which the public key credential may be exercised|origin=]'s [=effective domain=], or a [=is a registrable domain suffix of or is equal to|registrable domain suffix=] of the [=determines the set of origins on which the public key credential may be exercised|origin=]'s [=effective domain=].
 
-This can make deployment challenging for large environments where multiple country-specific domains are in use (e.g. example.com vs example.co.uk vs example.sg), where vanity or brand domains are required (e.g. myexampletravel.com vs examplecruises.com), and/or where platform as a service providers are used to support mobile apps. 
+This can make deployment challenging for large environments where multiple country-specific domains are in use (e.g. example.com vs example.co.uk vs example.sg), where alternative or brand domains are required (e.g. myexampletravel.com vs examplecruises.com), and/or where platform as a service providers are used to support mobile apps. 
 
 [=[WRPS]=] can opt in to allowing [=WebAuthn Clients=] to enable a credential to be created and used across a limited set of related [=origin|origins=]
 
--- a/index.bs
+++ b/index.bs
@@ -4130,9 +4130,10 @@ To validate the calling origin is an authorized related origin for a given cerem
     1. Let |domain| be the [=effective domain=] of |url|. If that is null, [=continue=].
     1. Remove any [=public suffix=] from the end of |domain|, including private registries and unknown registries. If |domain| is now empty, [=continue=].
     1. Split |domain| into |labels| and let |label| be the right-most one.
+    1. If the [=set/size=] of |labelsSeen| is greater than or equal to |maxLabels| and |labelsSeen|does not [=contains|contain=] |label|, [=continue=].
     1. If |callerOrigin| and |url| are [=same origin=], return [TRUE].
-    1. [=set/Append=] |label| to |labelsSeen|.
-    1. If the [=set/size=] of |labelsSeen| is greater than or equal to |maxLabels|, [=break=].
+    1. If |labelsSeen| [=contains=] |label|, [=continue=].
+    1. If the [=set/size=] of |labelsSeen| is less than |maxLabels|, [=set/Append=] |label| to |labelsSeen|.
 1. Return [FALSE].
 
 # WebAuthn <dfn>Authenticator Model</dfn> # {#sctn-authenticator-model}
--- a/index.bs
+++ b/index.bs
@@ -4121,8 +4121,10 @@ To validate the calling origin is an authorized related origin for a given cerem
     1. Let |domain| be the [=effective domain=] of |url|. If that is null, [=continue=].
     1. Remove any [=public suffix=] from the end of |domain|, including private registries and unknown registries. If |domain| is now empty, [=continue=].
     1. Split |domain| into [=origin labels|labels=] and let |label| be the right-most one.
-    1. If |label| is not in |labelsSeen| and the number of elements in |labelsSeen| is less than |maxLabels|, then insert |label| into |labelsSeen|. Otherwise, [=continue=]. 
+    1. If |labelsSeen| [=set/contains=] |label|, [=continue=].
     1. If |rpIdRequested| and |url| are [=same origin=], return `true`.
+    1. [=set/Append=] |label| to |labelsSeen|.
+    1. If the [=set/size=] of |labelsSeen| is greater than or equal to |maxLabels|, [=break=].
 1. Return `false`.
 
 # WebAuthn <dfn>Authenticator Model</dfn> # {#sctn-authenticator-model}
--- a/index.bs
+++ b/index.bs
@@ -1228,8 +1228,11 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
 : <dfn>Non-Discoverable Credential</dfn>
 :: This is a [=credential=] whose [=credential ID=] must be provided in {{PublicKeyCredentialRequestOptions/allowCredentials}} when calling {{CredentialsContainer/get()|navigator.credentials.get()}} because it is not [=client-side discoverable credential|client-side discoverable=]. See also [=server-side credentials=].
 
-: <dfn>Origin Label</dfn>
-:: The portion of a domain name preceding the effective top-level domain, as defined in the Public Suffix List [[PSL]]. This is often referred to as the eTLD+1. For example, the origin label for example.co.uk and example.de is `example`.
+: <dfn>Registrable Origin Label</dfn>
+:: The first [=domain label=] of the [=registrable domain=] of a [=domain=],
+    or null if the [=registrable domain=] is null.
+    For example, the [=registrable origin label=] of both `example.co.uk` and `www.example.de` is `example`
+    if both `co.uk` and `de` are [=public suffixes=].
 
 : <dfn>Public Key Credential</dfn>
 :: Generically, a *credential* is data one entity presents to another in order to *authenticate* the former to the latter
@@ -4106,7 +4109,7 @@ For example, for the RP ID `example.com`:
 }
 </xmp>
 
-[=WebAuthn Clients=] supporting this feature MUST support at least five [=origin labels=]. Client policy SHOULD define an upper limit to prevent abuse.
+[=WebAuthn Clients=] supporting this feature MUST support at least five [=registrable origin labels=]. Client policy SHOULD define an upper limit to prevent abuse.
 
 [=WebAuthn Clients=] supporting this feature SHOULD include {{ClientCapability/relatedOrigins}} in their response to [[#sctn-getClientCapabilities|getClientCapabilities()]].
 
@@ -4114,21 +4117,20 @@ For example, for the RP ID `example.com`:
 
 To validate the calling origin is an authorized related origin for a given ceremony:
 
-1. Let |maxLabels| be the maximum number of [=origin labels=] allowed by client policy.
+1. Let |maxLabels| be the maximum number of [=registrable origin labels=] allowed by client policy.
 1. Fetch the well-known URL for the RP ID (<code>https://|rpIdRequested|/.well-known/webauthn</code>).
     1. If the fetch fails, the response does not have a content type of `application/json`, or does not have a status code (after following redirects) of 200, then throw a "{{SecurityError}}" {{DOMException}}.
     1. If the body of the resource is not a valid JSON object, then throw a "{{SecurityError}}" {{DOMException}}.
     1. If the value of the |origins| property of the JSON object is missing, or is not an array of strings, then throw a "{{SecurityError}}" {{DOMException}}.
 1. Let |labelsSeen| be a new empty [=set=].
-1. [=set/For each=] string in |origins|:
-    1. Let |url| be the result of parsing the string as a URL. If that fails, [=continue=].
+1. [=set/For each=] |originItem| of |origins|:
+    1. Let |url| be the result of running the [=URL parser=] with |originItem| as the input. If that fails, [=continue=].
     1. Let |domain| be the [=effective domain=] of |url|. If that is null, [=continue=].
-    1. Remove any [=public suffix=] from the end of |domain|, including private registries and unknown registries. If |domain| is now empty, [=continue=].
-    1. Split |domain| into |labels| and let |label| be the right-most one.
-    1. If the [=set/size=] of |labelsSeen| is greater than or equal to |maxLabels| and |labelsSeen|does not [=contains|contain=] |label|, [=continue=].
+    1. Let |label| be [=registrable origin label=] of |domain|.
+    1. If |label| is empty or null, [=continue=].
+    1. If the [=set/size=] of |labelsSeen| is greater than or equal to |maxLabels| and |labelsSeen| does not [=set/contain=] |label|, [=continue=].
     1. If |callerOrigin| and |url| are [=same origin=], return [TRUE].
-    1. If |labelsSeen| [=contains=] |label|, [=continue=].
-    1. If the [=set/size=] of |labelsSeen| is less than |maxLabels|, [=set/Append=] |label| to |labelsSeen|.
+    1. If the [=set/size=] of |labelsSeen| is less than |maxLabels|, [=set/append=] |label| to |labelsSeen|.
 1. Return [FALSE].
 
 # WebAuthn <dfn>Authenticator Model</dfn> # {#sctn-authenticator-model}
--- a/index.bs
+++ b/index.bs
@@ -4110,7 +4110,7 @@ For example, for the RP ID `example.com`:
 
 To validate the calling origin is an authorized related origin for a given ceremony:
 
-1. Let |maxLabels| be the number of maximum [=origin labels=] allowed by client policy.
+1. Let |maxLabels| be the maximum number of [=origin labels=] allowed by client policy.
 1. Fetch the well-known URL for the RP ID (<code>https://|rpIdRequested|/.well-known/webauthn</code>).
     1. If the fetch fails, the response does not have a content type of `application/json`, or does not have a status code (after following redirects) of 200, then throw a "{{SecurityError}}" {{DOMException}}.
     1. If the body of the resource is not a valid JSON object, then throw a "{{SecurityError}}" {{DOMException}}.
--- a/index.bs
+++ b/index.bs
@@ -1228,6 +1228,9 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
 : <dfn>Non-Discoverable Credential</dfn>
 :: This is a [=credential=] whose [=credential ID=] must be provided in {{PublicKeyCredentialRequestOptions/allowCredentials}} when calling {{CredentialsContainer/get()|navigator.credentials.get()}} because it is not [=client-side discoverable credential|client-side discoverable=]. See also [=server-side credentials=].
 
+: <dfn>Origin Label</dfn>
+:: The portion of a domain name preceding the effective top-level domain, as defined in the Public Suffix List [[PSL]]. This is often referred to as the eTLD+1. For example, the origin label for example.co.uk and example.de is `example`.
+
 : <dfn>Public Key Credential</dfn>
 :: Generically, a *credential* is data one entity presents to another in order to *authenticate* the former to the latter
     [[RFC4949]]. The term [=public key credential=] refers to one of: a [=public key credential source=], the
@@ -1824,11 +1827,37 @@ a numbered step. If outdented, it (today) is rendered as a bullet in the midst o
 
                 :   is present
                 ::  If <code>|pkOptions|.{{PublicKeyCredentialCreationOptions/rp}}.{{PublicKeyCredentialRpEntity/id}}</code> [=is not a
-                    registrable domain suffix of and is not equal to=] |effectiveDomain|, throw a "{{SecurityError}}" {{DOMException}}.
+                    registrable domain suffix of and is not equal to=] |effectiveDomain|, and if the client
+                
+                    <dl class="switch">
+                        :   supports [[#sctn-related-origins|related origin requests]]
+                        ::  1. let |rpIdRequested| be the value of |pkOptions|.{{PublicKeyCredentialCreationOptions/rp}}.{{PublicKeyCredentialRpEntity/id}}
 
-                :   Is not present
-                ::  Set <code>|pkOptions|.{{PublicKeyCredentialCreationOptions/rp}}.{{PublicKeyCredentialRpEntity/id}}</code> to
-                    |effectiveDomain|.
+                            1. let |maxLabels| be 
+                        
+                            1. fetch the well-known URL for the RP ID (<code>https://|rpIdRequested|/.well-known/webauthn-origins</code>).
+                        
+                                1. if the fetch fails, the response does not have a content type of `application/json`, or does not have a status code (after following redirects) of 200, then throw a "{{SecurityError}}" {{DOMException}}.
+                                1. if the body of the resource is not a valid JSON object, then throw a "{{SecurityError}}" {{DOMException}}.
+                                1. if the value of the |origins| member of the JSON object is missing, or is not a list of strings, then throw a "{{SecurityError}}" {{DOMException}}.
+    
+                            1. Let |originLabelsSeen| be an empty set
+                            1. [=set/For each=] string in |origins|:
+                                1. Let |url| be the result of parsing the string as a URL. If that fails, continue with the next element in the list.
+                                1. Let |domain| be the !!effective domain of |url|. If that is null, continue with the next element of the list.
+                                1. Remove any !!public suffix from the end of |domain|, including private registries and unknown registries. If |domain| is now empty, continue with the next element of the list.
+                                1. Split |domain| into [=origin labels|labels=] and let |label| be the right-most one.
+                                1. If |label| is not in |labelsSeen| then:
+                                    1. If the number of elements in |labelsSeen| is less than |maxLabels|, then insert |label| into |labelsSeen|.
+                                    1. Otherwise, continue with the next element of the list.
+                                1. If |rpIdRequested| and |url| are !!same origin!!, then [=continue=].
+
+                        :   does not support [[#sctn-related-origins|related origin requests]]
+                        ::  throw a "{{SecurityError}}" {{DOMException}}.
+                    </dl>
+
+                :   is not present
+                ::  Set <code>|pkOptions|.{{PublicKeyCredentialCreationOptions/rp}}.{{PublicKeyCredentialRpEntity/id}}</code> to |effectiveDomain|.
             </dl>
 
         Note: <code>|pkOptions|.{{PublicKeyCredentialCreationOptions/rp}}.{{PublicKeyCredentialRpEntity/id}}</code> represents the
@@ -4046,7 +4075,39 @@ To override this default policy and indicate that a cross-origin <{iframe}> is a
 
 [=[RPS]=] utilizing the WebAuthn API in an embedded context should review [[#sctn-seccons-visibility]] regarding [=UI redressing=] and its possible mitigations.
 
+## Using Web Authentication across related origins ## {#sctn-related-origins}
 
+By default, 
+
+[=[WRP]=] can opt in to allowing a [=WebAuthn Clients=] to select credentials created with a different, but related [=RP ID=] than the caller's [=origin=].
+
+WebAuthn credential to be used across a limited set of [=origin|origins=] to support use cases such as country code TLDs(!add reference!), vanity or brand domains, and platform as a service providers supporting mobile apps.
+
+Relying Parties MUST choose a common Relying Party ID to use across all ceremonies from related origins.
+
+A JSON document MUST be hosted at the `webauthn-origins` well-known URL ([[!RFC8615]]) for the [=RP ID=].
+
+The JSON document MUST be returned as follows:
+
+    - the content type must be `application/json`
+    - JSON document MUST contain a single object named `origins`
+    - the `origins` object MUST contain an array of one or more web origins
+
+For example, for the RP ID `example.com`:
+
+<xmp class="json" highlight="json">
+{
+    "origins": [
+        "https://example.co.uk",
+        "https://example.de",
+        "https://myexamplerewards.com"
+    ]
+}
+</xmp>
+
+[=WebAuthn Clients=] SHOULD limit the number of related origins, with a recommended maximum of 5 [=origin labels=].
+
+[=WebAuthn Clients=] supporting this feature SHOULD include `related-origin-requests` in their response to getClientCapabilities() as defined in !XYZ!.
 
 # WebAuthn <dfn>Authenticator Model</dfn> # {#sctn-authenticator-model}
 
--- a/index.bs
+++ b/index.bs
@@ -4122,10 +4122,10 @@ To validate the calling origin is an authorized related origin for a given cerem
     1. Remove any [=public suffix=] from the end of |domain|, including private registries and unknown registries. If |domain| is now empty, [=continue=].
     1. Split |domain| into [=origin labels|labels=] and let |label| be the right-most one.
     1. If |labelsSeen| [=set/contains=] |label|, [=continue=].
-    1. If |rpIdRequested| and |url| are [=same origin=], return `true`.
+    1. If |rpIdRequested| and |url| are [=same origin=], return [TRUE].
     1. [=set/Append=] |label| to |labelsSeen|.
     1. If the [=set/size=] of |labelsSeen| is greater than or equal to |maxLabels|, [=break=].
-1. Return `false`.
+1. Return [FALSE].
 
 # WebAuthn <dfn>Authenticator Model</dfn> # {#sctn-authenticator-model}
 
--- a/index.bs
+++ b/index.bs
@@ -4086,11 +4086,6 @@ A JSON document MUST be hosted at the `webauthn` well-known URL ([[!RFC8615]]) f
 
     - The content type MUST be `application/json`.
     - The top-level JSON object MUST contain a key named `origins` whose value MUST be an array of one or more strings containing web origins.
-    - Origins with the same [=origin label=] SHOULD be listed together as the first elements in the array, with unique origins as the last elements.
-
-    :: Note: When adding additional origins matching an existing [=origin label=], add them after the existing elements with the same [=origin label=] instead of at the end of the array.
-        This will ensure their processing will not impact the [=origin label=] count.
-
 
 For example, for the RP ID `example.com`:
 
--- a/index.bs
+++ b/index.bs
@@ -4079,7 +4079,7 @@ By default, Web Authentication requires that the [=RP ID=] be equal to the [=det
 
 This can make deployment challenging for large environments where multiple country-specific domains are in use (e.g. example.com vs example.co.uk vs example.sg), where alternative or brand domains are required (e.g. myexampletravel.com vs examplecruises.com), and/or where platform as a service providers are used to support mobile apps. 
 
-[=[WRPS]=] can opt in to allowing [=WebAuthn Clients=] to enable a credential to be created and used across a limited set of related [=origin|origins=]
+[=[WRPS]=] can opt in to allowing [=WebAuthn Clients=] to enable a credential to be created and used across a limited set of related [=origin|origins=].
 
 Relying Parties MUST choose a common [=RP ID=] to use across all ceremonies from related origins.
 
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -4046,7 +4107,38 @@ To override this default policy and indicate that a cross-origin <{iframe}> is a
...
+    "origins": [
+        "https://example.co.uk",
+        "https://example.de",
+        "https://example.sg"
+        "https://myexamplerewards.com"
+    ]
+}
+</xmp>
+
+To mitigate abuse, [=WebAuthn Clients=] SHOULD limit the number of related origins, with a recommended maximum of 5 [=origin labels=].
```
sbweeden, line 4139: 5 seems like an arbitrary and small number. It also leaves things wide open to client inconsistency. I would suggest this be discussed and agreed upon amongst major browser vendors before the ink dries. RPs and users will suffer if support is inconsistent.

timcappalli: It was agreed upon as part of the explainer. 5 labels are more than adequate for the use case it is aiming to solve.

sbweeden: Then can we say MUST support at least 5 .... and SHOULD limit ...

timcappalli: Not sure I agree. There may be clients who only want to support ccTLD usage. I don't think we should prevent them from implementing because of that.

index.bs
```diff
@@ -1824,11 +1827,37 @@ a numbered step. If outdented, it (today) is rendered as a bullet in the midst o
...
+                        :   supports [[#sctn-related-origins|related origin requests]]
+                        ::  1. Let |rpIdRequested| be the value of <code>|pkOptions|.{{PublicKeyCredentialCreationOptions/rp}}.{{PublicKeyCredentialRpEntity/id}}</code>.
+
+                            1. Let |maxLabels| be the number of maximum [=origin labels=] allowed by client policy.
+                        
+                            1. Fetch the well-known URL for the RP ID (<code>https://|rpIdRequested|/.well-known/webauthn-origins</code>).
+                        
+                                1. If the fetch fails, the response does not have a content type of `application/json`, or does not have a status code (after following redirects) of 200, then throw a "{{SecurityError}}" {{DOMException}}.
+                                1. If the body of the resource is not a valid JSON object, then throw a "{{SecurityError}}" {{DOMException}}.
+                                1. If the value of the |origins| member of the JSON object is missing, or is not a list of strings, then throw a "{{SecurityError}}" {{DOMException}}.
```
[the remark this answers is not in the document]
timcappalli: I think because the processing rules are pseudocode, it is `list`. @emlun do you know?

index.bs
```diff
@@ -1824,11 +1827,37 @@ a numbered step. If outdented, it (today) is rendered as a bullet in the midst o
...
+                
+                    <dl class="switch">
+                        :   supports [[#sctn-related-origins|related origin requests]]
+                        ::  1. Let |rpIdRequested| be the value of <code>|pkOptions|.{{PublicKeyCredentialCreationOptions/rp}}.{{PublicKeyCredentialRpEntity/id}}</code>.
+
+                            1. Let |maxLabels| be the number of maximum [=origin labels=] allowed by client policy.
+                        
+                            1. Fetch the well-known URL for the RP ID (<code>https://|rpIdRequested|/.well-known/webauthn-origins</code>).
+                        
+                                1. If the fetch fails, the response does not have a content type of `application/json`, or does not have a status code (after following redirects) of 200, then throw a "{{SecurityError}}" {{DOMException}}.
```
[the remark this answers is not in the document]
timcappalli: Apex domains can be challenging (and in some cases impossible) to directly add resources to, but adding a redirect  is often allowed.

index.bs
```diff
@@ -1824,11 +1827,37 @@ a numbered step. If outdented, it (today) is rendered as a bullet in the midst o
...
-                    registrable domain suffix of and is not equal to=] |effectiveDomain|, throw a "{{SecurityError}}" {{DOMException}}.
+                    registrable domain suffix of and is not equal to=] |effectiveDomain|, and if the client
+                
+                    <dl class="switch">
+                        :   supports [[#sctn-related-origins|related origin requests]]
+                        ::  1. Let |rpIdRequested| be the value of <code>|pkOptions|.{{PublicKeyCredentialCreationOptions/rp}}.{{PublicKeyCredentialRpEntity/id}}</code>.
+
+                            1. Let |maxLabels| be the number of maximum [=origin labels=] allowed by client policy.
+                        
+                            1. Fetch the well-known URL for the RP ID (<code>https://|rpIdRequested|/.well-known/webauthn-origins</code>).
```
MasterKale, line 1838: If the JSON served here already needs an `"origins"` key with the array of origins, what if this became the basis for a general WebAuthn configuration by changing the URL to `/.well-known/webauthn`?

MasterKale: During WG @ 3/13: Could make collaboration with e.g. passkey management URLs trickier if WebAppSec WG wants to work on a similar well-known document.

abergs:
Is there any considerations to non https origins, like browser extensions?

Our use case may be niche, but I'll share it:

As a Bitwarden user, you can self-host your server. For example, i can host my vault and data on "vault.andersaberg.com", but still use the bitwarden clients (browser, ios, android, native desktop) to access my data. You simply configure your client to go to your self-hosted domain for the auth and api calls.

Would the bitwarden extension be able to use a passkey, bound to "vault.andersaberg.com", if there is a "vault.andersaberg.com/.well-known/webauthn"-file proclaiming "bitwarden.com" to be an associated domain?

I have the same question for mobile apps, but I don't think anyone here can respond to that.

index.bs
timcappalli: Made this change in [689647c](https://github.com/w3c/webauthn/pull/2040/commits/689647ca72d2438f8dd849b398f891779caf6a91) but can revert it back or change it.

fyi @pascoej

index.bs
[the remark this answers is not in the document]
timcappalli: Ah I see what you mean now @sbweeden. I was thinking about a different line. This is addressed in 3aad0542be39fdf0aabadbfb21fac9a729e75bfe

index.bs
```diff
@@ -1824,11 +1827,37 @@ a numbered step. If outdented, it (today) is rendered as a bullet in the midst o
 
                 :   is present
                 ::  If <code>|pkOptions|.{{PublicKeyCredentialCreationOptions/rp}}.{{PublicKeyCredentialRpEntity/id}}</code> [=is not a
-                    registrable domain suffix of and is not equal to=] |effectiveDomain|, throw a "{{SecurityError}}" {{DOMException}}.
+                    registrable domain suffix of and is not equal to=] |effectiveDomain|, and if the client
+                
+                    <dl class="switch">
+                        :   supports [[#sctn-related-origins|related origin requests]]
+                        ::  1. Let |rpIdRequested| be the value of <code>|pkOptions|.{{PublicKeyCredentialCreationOptions/rp}}.{{PublicKeyCredentialRpEntity/id}}</code>.
```
emlun, line 1834: This procedure looks like it's the same between both `create()` and `get()`, is it? If so I think we should extract this to a reusable procedure we can call from both places. Probably make it a subsection of §5.11?

index.bs
```diff
@@ -1824,11 +1827,37 @@ a numbered step. If outdented, it (today) is rendered as a bullet in the midst o
...
+    
+                            1. Let |originLabelsSeen| be an empty set.
+                            1. [=set/For each=] string in |origins|:
+                                1. Let |url| be the result of parsing the string as a URL. If that fails, continue with the next element in the list.
+                                1. Let |domain| be the [=effective domain=] of |url|. If that is null, continue with the next element of the list.
+                                1. Remove any [=public suffix=] from the end of |domain|, including private registries and unknown registries. If |domain| is now empty, continue with the next element of the list.
+                                1. Split |domain| into [=origin labels|labels=] and let |label| be the right-most one.
+                                1. If |label| is not in |labelsSeen| and:
+                                    1. If the number of elements in |labelsSeen| is less than |maxLabels|, then insert |label| into |labelsSeen|.
+                                    1. Otherwise, continue with the next element of the list.
```
emlun, line 1852: I think we can use `[=continue=]` for all of these?
```suggestion
                                1. Let |url| be the result of parsing the string as a URL. If that fails, [=continue=].
                                1. Let |domain| be the [=effective domain=] of |url|. If that is null, [=continue=].
                                1. Remove any [=public suffix=] from the end of |domain|, including private registries and unknown registries. If |domain| is now empty, [=continue=].
                                1. Split |domain| into [=origin labels|labels=] and let |label| be the right-most one.
                                1. If |label| is not in |labelsSeen| and:
                                    1. If the number of elements in |labelsSeen| is less than |maxLabels|, then insert |label| into |labelsSeen|.
                                    1. Otherwise, [=continue=].
```

Also, the "...and:" formatting in step 5 looks a bit off from the document style so far. I would just write it as a single inline step:
```suggestion
                                1. Let |url| be the result of parsing the string as a URL. If that fails, continue with the next element in the list.
                                1. Let |domain| be the [=effective domain=] of |url|. If that is null, continue with the next element of the list.
                                1. Remove any [=public suffix=] from the end of |domain|, including private registries and unknown registries. If |domain| is now empty, continue with the next element of the list.
                                1. Split |domain| into [=origin labels|labels=] and let |label| be the right-most one.
                                1. If |label| is not in |labelsSeen| and the number of elements in |labelsSeen| is less than |maxLabels|, then insert |label| into |labelsSeen|. Otherwise, [=continue=].
```

index.bs
```diff
@@ -1824,11 +1827,37 @@ a numbered step. If outdented, it (today) is rendered as a bullet in the midst o
...
+                                1. If |rpIdRequested| and |url| are [=same origin=], then [=continue=].
```
emlun, line 1853: ...so I think you may have meant `[=break=]` instead of `[=continue=]` here (i.e., "continue" with the registration flow if a match is found)?

If I got your intention right I think we also need to throw the error if the loop exits without finding a match, but skip the error if we don't find a match. That could perhaps be handled easiest as a return value from the extracted procedure I mentioned above.

index.bs
```diff
@@ -1824,11 +1827,37 @@ a numbered step. If outdented, it (today) is rendered as a bullet in the midst o
...
+                                1. If the body of the resource is not a valid JSON object, then throw a "{{SecurityError}}" {{DOMException}}.
+                                1. If the value of the |origins| property of the JSON object is missing, or is not an array of strings, then throw a "{{SecurityError}}" {{DOMException}}.
+    
+                            1. Let |originLabelsSeen| be an empty set.
+                            1. [=set/For each=] string in |origins|:
+                                1. Let |url| be the result of parsing the string as a URL. If that fails, continue with the next element in the list.
+                                1. Let |domain| be the [=effective domain=] of |url|. If that is null, continue with the next element of the list.
+                                1. Remove any [=public suffix=] from the end of |domain|, including private registries and unknown registries. If |domain| is now empty, continue with the next element of the list.
+                                1. Split |domain| into [=origin labels|labels=] and let |label| be the right-most one.
```
emlun, line 1849: I thought all of what precedes the eTLD is the label (i.e., the origin label of `login.example.co.uk` is `"login.example"`, not `["login", "example"]`)? Perhaps we could just define the "origin label" to be just the right-most domain segment apart from the eTLD, then we don't need to repeat the splitting logic here.

index.bs
```diff
@@ -1824,11 +1827,37 @@ a numbered step. If outdented, it (today) is rendered as a bullet in the midst o
...
+
+                            1. Let |maxLabels| be the number of maximum [=origin labels=] allowed by client policy.
+                        
+                            1. Fetch the well-known URL for the RP ID (<code>https://|rpIdRequested|/.well-known/webauthn</code>).
+                        
+                                1. If the fetch fails, the response does not have a content type of `application/json`, or does not have a status code (after following redirects) of 200, then throw a "{{SecurityError}}" {{DOMException}}.
+                                1. If the body of the resource is not a valid JSON object, then throw a "{{SecurityError}}" {{DOMException}}.
+                                1. If the value of the |origins| property of the JSON object is missing, or is not an array of strings, then throw a "{{SecurityError}}" {{DOMException}}.
+    
+                            1. Let |originLabelsSeen| be an empty set.
```
emlun, line 1844: Should this be `|labelsSeen|`? Even then, `labelsSeen` appears to not be used for anything after it's constructed. I had expected the loop would abort after this reaches the max size?

index.bs
```diff
@@ -3963,6 +4020,7 @@ Note: The {{UserVerificationRequirement}} enumeration is deliberately not refere
         "hybridTransport",
         "passkeyPlatformAuthenticator",
         "userVerifyingPlatformAuthenticator",
+        "relatedOriginRequests"
```
emlun, line 4023: Maybe `relatedOrigins` is enough?

index.bs
```diff
@@ -4046,7 +4107,38 @@ To override this default policy and indicate that a cross-origin <{iframe}> is a
 
 [=[RPS]=] utilizing the WebAuthn API in an embedded context should review [[#sctn-seccons-visibility]] regarding [=UI redressing=] and its possible mitigations.
 
+## Using Web Authentication across related origins ## {#sctn-related-origins}
+
+By default, Web Authentication requires that the [=RP ID=] be equal to the [=determines the set of origins on which the public key credential may be exercised|origin=]'s [=effective domain=], or a [=is a registrable domain suffix of or is equal to|registrable domain suffix=] of the [=determines the set of origins on which the public key credential may be exercised|origin=]'s [=effective domain=].
+
+This can make deployment challenging for large environments where multiple country-specific domains are in use (e.g. example.com vs example.co.uk vs example.sg), where vanity or brand domains are required (e.g. myexampletravel.com vs examplecruises.com), and/or where platform as a service providers are used to support mobile apps.
```
emlun, line 4114: Could "vanity" be seen as a judgmental term? It always sounds derisive to me (English is my second language).

index.bs
```diff
@@ -4046,7 +4107,38 @@ To override this default policy and indicate that a cross-origin <{iframe}> is a
...
+
+By default, Web Authentication requires that the [=RP ID=] be equal to the [=determines the set of origins on which the public key credential may be exercised|origin=]'s [=effective domain=], or a [=is a registrable domain suffix of or is equal to|registrable domain suffix=] of the [=determines the set of origins on which the public key credential may be exercised|origin=]'s [=effective domain=].
+
+This can make deployment challenging for large environments where multiple country-specific domains are in use (e.g. example.com vs example.co.uk vs example.sg), where vanity or brand domains are required (e.g. myexampletravel.com vs examplecruises.com), and/or where platform as a service providers are used to support mobile apps. 
+
+[=[WRPS]=] can opt in to allowing [=WebAuthn Clients=] to enable a credential to be created and used across a limited set of related [=origin|origins=]
+
+Relying Parties MUST choose a common [=RP ID=] to use across all ceremonies from related origins.
 
+A JSON document MUST be hosted at the `webauthn` well-known URL ([[!RFC8615]]) for the [=RP ID=]. The JSON document MUST be returned as follows:
```
emlun, line 4120: Looks like we'll need to request a [registration in the registry](https://www.rfc-editor.org/rfc/rfc8615#section-3.1) for this (once we've worked out what its substance would be)?

index.bs
```diff
@@ -4046,7 +4107,38 @@ To override this default policy and indicate that a cross-origin <{iframe}> is a
...
+
+    - the content type must be `application/json`
+    - JSON document MUST contain a single object named `origins`
+    - the `origins` object MUST contain an array of one or more web origins
```
emlun, line 4124:
```suggestion
    - The content type must be `application/json`.
    - JSON document MUST contain a single object named `origins`.
    - The `origins` object MUST contain an array of one or more web origins.
```

index.bs
timcappalli: @emlun what is the name of the section? I don't see a 5.11

index.bs
timcappalli: changed to "alternative"

index.bs
```diff
@@ -4046,7 +4084,57 @@ To override this default policy and indicate that a cross-origin <{iframe}> is a
...
+1. Fetch the well-known URL for the RP ID (<code>https://|rpIdRequested|/.well-known/webauthn</code>).
+    1. If the fetch fails, the response does not have a content type of `application/json`, or does not have a status code (after following redirects) of 200, then throw a "{{SecurityError}}" {{DOMException}}.
+    1. If the body of the resource is not a valid JSON object, then throw a "{{SecurityError}}" {{DOMException}}.
+    1. If the value of the |origins| property of the JSON object is missing, or is not an array of strings, then throw a "{{SecurityError}}" {{DOMException}}.
+1. Let |originLabelsSeen| be an empty set.
+1. [=set/For each=] string in |origins|:
+    1. Let |url| be the result of parsing the string as a URL. If that fails, [=continue=].
+    1. Let |domain| be the [=effective domain=] of |url|. If that is null, [=continue=].
+    1. Remove any [=public suffix=] from the end of |domain|, including private registries and unknown registries. If |domain| is now empty, [=continue=].
+    1. Split |domain| into [=origin labels|labels=] and let |label| be the right-most one.
```
emlun, line 4134: https://github.com/w3c/webauthn/pull/2040#discussion_r1531243743 is still relevant here:

>I thought all of what precedes the eTLD is the label (i.e., the origin label of `login.example.co.uk` is `"login.example"`, not `["login", "example"]`)? Perhaps we could just define the "origin label" to be just the right-most domain segment apart from the eTLD, then we don't need to repeat the splitting logic here.

index.bs
```diff
@@ -4046,7 +4084,57 @@ To override this default policy and indicate that a cross-origin <{iframe}> is a
...
+    1. If |label| is not in |labelsSeen| and the number of elements in |labelsSeen| is less than |maxLabels|, then insert |label| into |labelsSeen|. Otherwise, [=continue=]. 
+    1. If |rpIdRequested| and |url| are [=same origin=], return `true`.
```
emlun, line 4136: I think splitting this up could make it a bit easier to follow. Did I understand this correctly?
```suggestion
    1. If |labelsSeen| [=set/contains=] |label|, [=continue=].
    1. [=set/Append=] |label| to |labelsSeen|.
    1. If the [=set/size=] of |labelsSeen| is greater than |maxLabels|, [=break=].
    1. If |rpIdRequested| and |url| are [=same origin=], return `true`.
```

emlun: I think you could even rearrange the steps a little bit to possibly exit earlier:

```suggestion
    1. If |labelsSeen| [=set/contains=] |label|, [=continue=].
    1. If |rpIdRequested| and |url| are [=same origin=], return `true`.
    1. [=set/Append=] |label| to |labelsSeen|.
    1. If the [=set/size=] of |labelsSeen| is greater than or equal to |maxLabels|, [=break=].
```

emlun, line 4136: Oh, and we can use the `[TRUE]` macro for formatting consistency.

```suggestion
    1. If |label| is not in |labelsSeen| and the number of elements in |labelsSeen| is less than |maxLabels|, then insert |label| into |labelsSeen|. Otherwise, [=continue=]. 
    1. If |rpIdRequested| and |url| are [=same origin=], return [TRUE].
```

index.bs
```diff
@@ -1824,11 +1827,25 @@ a numbered step. If outdented, it (today) is rendered as a bullet in the midst o
...
-                :   Is not present
-                ::  Set <code>|pkOptions|.{{PublicKeyCredentialCreationOptions/rp}}.{{PublicKeyCredentialRpEntity/id}}</code> to
-                    |effectiveDomain|.
+                            1. If the result of running the [[#sctn-validating-relation-origin|related origins validation procedure]] is
+                                <dl class="switch">
+                                    :   `false`
+                                    ::  throw a "{{SecurityError}}" {{DOMException}}.
+                                    :   `true`
+                                    ::  [=break=].
+                                </dl>
```
emlun, line 1842: I don't think there's any loop to `[=break=]` here anymore. Looks like this now reduces to just "If the result of running the related origins validation procedure is [FALSE], throw a "SecurityError" DOMException."?

index.bs
```diff
@@ -2314,19 +2331,36 @@ When this method is invoked, the user agent MUST execute the following algorithm
...
+                            :   supports [[#sctn-related-origins|related origin requests]]
+                            ::  1. Let |rpIdRequested| be the value of <code>|pkOptions|.{{PublicKeyCredentialRequestOptions/rpId}}</code>
+
+                                1. If the result of running the [[#sctn-validating-relation-origin|related origins validation procedure]] is
+                                    <dl class="switch">
+                                        :   `false`
+                                        ::  throw a "{{SecurityError}}" {{DOMException}}.
+                                        :   `true`
+                                        ::  [=break=].
+                                    </dl>
```
emlun, line 2351: (Ditto)

index.bs
```diff
@@ -4046,7 +4084,57 @@ To override this default policy and indicate that a cross-origin <{iframe}> is a
...
+### Validating Related Origins ### {#sctn-validating-relation-origin}
+
+To validate the calling origin is an authorized related origin for a given ceremony:
+
+1. Let |maxLabels| be the number of maximum [=origin labels=] allowed by client policy.
+1. Fetch the well-known URL for the RP ID (<code>https://|rpIdRequested|/.well-known/webauthn</code>).
+    1. If the fetch fails, the response does not have a content type of `application/json`, or does not have a status code (after following redirects) of 200, then throw a "{{SecurityError}}" {{DOMException}}.
+    1. If the body of the resource is not a valid JSON object, then throw a "{{SecurityError}}" {{DOMException}}.
+    1. If the value of the |origins| property of the JSON object is missing, or is not an array of strings, then throw a "{{SecurityError}}" {{DOMException}}.
+1. Let |originLabelsSeen| be an empty set.
```
emlun, line 4129: I think this should be `labelsSeen`?

Also, elsewhere in the spec we currently use the phrases "new empty set" or "new ordered set" ([=set=] and [=ordered set=] are equivalent Infra terms), so let's say "new" here too for consistency.

```suggestion
1. Let |labelsSeen| be a new empty [=set=].
```

emlun (changes requested):
The extracted procedure is definitely an improvement! :smile: Some more things to clean up, though...

pascoej (approved):

index.bs
timcappalli: https://github.com/protocol-registries/well-known-uris/issues/45

index.bs
timcappalli: We only want to count based on the eTLD+1 and ignore anything further to the left (e.g. subdomains).

timcappalli:
> Would the bitwarden extension be able to use a passkey, bound to "vault.andersaberg.com", if there is a "vault.andersaberg.com/.well-known/webauthn"-file proclaiming "bitwarden.com", or a browser extension ID to be an associated domain? I'm guessing _yes_, if the extension has host permission on that domain (vault.andersaberg.com)?


@abergs (and @nsatragno) can the extension dynamically change its identifier (locally) after install based on the vault configuration? (https://lists.w3.org/Archives/Public/public-webauthn/2023Dec/0078.html)

index.bs
```diff
@@ -4046,7 +4073,59 @@ To override this default policy and indicate that a cross-origin <{iframe}> is a
...
+}
+</xmp>
+
+[=WebAuthn Clients=] supporting this feature MUST support at least five [=origin labels=]. Client policy SHOULD define an upper limit to prevent abuse.
+
+[=WebAuthn Clients=] supporting this feature SHOULD include {{ClientCapability/relatedOrigins}} in their response to [[#sctn-getClientCapabilities|getClientCapabilities()]].
+
+### Validating Related Origins ### {#sctn-validating-relation-origin}
+
+To validate the calling origin is an authorized related origin for a given ceremony:
```
MasterKale, line 4111: I think this needs something to qualify that these steps are only required if `rp.id` is **not** appropriate for the current website.

Right now an RP with `rp.id` of `"example.com"` hosting auth on `https://example.com` (the typical use case) would experience broken auth if the RP didn't set up `https://example.com/.well-known/webauthn`.

Sure, in the section above it's made clear that these steps are for use of `"example.com"` on `https://example.co.uk` but the validation steps should probably explicitly declare this.

index.bs
```diff
@@ -4046,7 +4073,59 @@ To override this default policy and indicate that a cross-origin <{iframe}> is a
...
+
+1. Let |maxLabels| be the maximum number of [=origin labels=] allowed by client policy.
+1. Fetch the well-known URL for the RP ID (<code>https://|rpIdRequested|/.well-known/webauthn</code>).
+    1. If the fetch fails, the response does not have a content type of `application/json`, or does not have a status code (after following redirects) of 200, then throw a "{{SecurityError}}" {{DOMException}}.
+    1. If the body of the resource is not a valid JSON object, then throw a "{{SecurityError}}" {{DOMException}}.
+    1. If the value of the |origins| property of the JSON object is missing, or is not an array of strings, then throw a "{{SecurityError}}" {{DOMException}}.
```
MasterKale, line 4117: Note: these new `SecurityError` instances should get documented in both API exceptions sections being added with #2047

index.bs
MasterKale: Wait, I understand now, this is the verification algorithm and it assumes that `rp.id` is matched against the domain before it gets executed. Nothing to do here then.

index.bs
```diff
@@ -4046,7 +4073,59 @@ To override this default policy and indicate that a cross-origin <{iframe}> is a
 
 [=[RPS]=] utilizing the WebAuthn API in an embedded context should review [[#sctn-seccons-visibility]] regarding [=UI redressing=] and its possible mitigations.
 
+## Using Web Authentication across related origins ## {#sctn-related-origins}
+
+By default, Web Authentication requires that the [=RP ID=] be equal to the [=determines the set of origins on which the public key credential may be exercised|origin=]'s [=effective domain=], or a [=is a registrable domain suffix of or is equal to|registrable domain suffix=] of the [=determines the set of origins on which the public key credential may be exercised|origin=]'s [=effective domain=].
+
+This can make deployment challenging for large environments where multiple country-specific domains are in use (e.g. example.com vs example.co.uk vs example.sg), where alternative or brand domains are required (e.g. myexampletravel.com vs examplecruises.com), and/or where platform as a service providers are used to support mobile apps. 
+
+[=[WRPS]=] can opt in to allowing [=WebAuthn Clients=] to enable a credential to be created and used across a limited set of related [=origin|origins=]
```
emlun, line 4082:
```suggestion
[=[WRPS]=] can opt in to allowing [=WebAuthn Clients=] to enable a credential to be created and used across a limited set of related [=origin|origins=].
```

MasterKale (approved):

index.bs
```diff
@@ -4046,7 +4073,59 @@ To override this default policy and indicate that a cross-origin <{iframe}> is a
...
+
+Relying Parties MUST choose a common [=RP ID=] to use across all ceremonies from related origins.
 
+A JSON document MUST be hosted at the `webauthn` well-known URL ([[!RFC8615]]) for the [=RP ID=]. The JSON document MUST be returned as follows:
+
+    - The content type MUST be `application/json`.
+    - JSON document MUST contain a single object named `origins`.
+    - The `origins` object MUST contain an array of one or more web origins.
```
emlun, line 4090: Is this intended to disallow other keys? If so, how about this formulation?

```suggestion
    - The top-level JSON object MUST contain only the key `"origins"`,
        whose value MUST be an array of one or more strings containing web origins.
```

index.bs
```diff
@@ -4046,7 +4073,59 @@ To override this default policy and indicate that a cross-origin <{iframe}> is a
...
 
+## Using Web Authentication across related origins ## {#sctn-related-origins}
+
+By default, Web Authentication requires that the [=RP ID=] be equal to the [=determines the set of origins on which the public key credential may be exercised|origin=]'s [=effective domain=], or a [=is a registrable domain suffix of or is equal to|registrable domain suffix=] of the [=determines the set of origins on which the public key credential may be exercised|origin=]'s [=effective domain=].
+
+This can make deployment challenging for large environments where multiple country-specific domains are in use (e.g. example.com vs example.co.uk vs example.sg), where alternative or brand domains are required (e.g. myexampletravel.com vs examplecruises.com), and/or where platform as a service providers are used to support mobile apps. 
+
+[=[WRPS]=] can opt in to allowing [=WebAuthn Clients=] to enable a credential to be created and used across a limited set of related [=origin|origins=]
+
+Relying Parties MUST choose a common [=RP ID=] to use across all ceremonies from related origins.
```
emlun, line 4084: I think this should perhaps be part of the previous paragraph? So we don't imply that this is required for all RPs, just the ones that opt in.
```suggestion
Such [=[RPS]=] MUST choose a common [=RP ID=] to use across all ceremonies from related origins.
```

index.bs
```diff
@@ -4046,7 +4073,59 @@ To override this default policy and indicate that a cross-origin <{iframe}> is a
...
 
+A JSON document MUST be hosted at the `webauthn` well-known URL ([[!RFC8615]]) for the [=RP ID=]. The JSON document MUST be returned as follows:
+
+    - The content type MUST be `application/json`.
+    - JSON document MUST contain a single object named `origins`.
```
sbweeden, line 4089: Very minor nit. What is meant by "MUST contain a single object named `origins`" here? The way it reads to me sounds like *nothing else* is allowed in the well-known URL for webauthn, which seems limiting if we want to allow room for future expansion without having to come back and change this text later. Couldn't this say something like:
```
Support for related origins is declared via a single key within the well-known document named `origins`.
```

sbweeden (changes requested):
Please consider the wording for the constraints related origins well-known document. It's a very minor consideration. Otherwise I'm perfectly ok with these changes.

index.bs
```diff
@@ -4046,7 +4073,59 @@ To override this default policy and indicate that a cross-origin <{iframe}> is a
...
+[=WebAuthn Clients=] supporting this feature MUST support at least five [=origin labels=]. Client policy SHOULD define an upper limit to prevent abuse.
+
+[=WebAuthn Clients=] supporting this feature SHOULD include {{ClientCapability/relatedOrigins}} in their response to [[#sctn-getClientCapabilities|getClientCapabilities()]].
+
+### Validating Related Origins ### {#sctn-validating-relation-origin}
+
+To validate the calling origin is an authorized related origin for a given ceremony:
+
+1. Let |maxLabels| be the maximum number of [=origin labels=] allowed by client policy.
+1. Fetch the well-known URL for the RP ID (<code>https://|rpIdRequested|/.well-known/webauthn</code>).
```
sbweeden, line 4114: I have a question about the fetching of the well-known URL. Could this be abused given that the origin of the page that contains the WebAuthn call is necessarily a different origin from the site where the well-known document will be retrieved? This seems to constitute a cross origin request.

timcappalli: It is the WebAuthn client software fetching the document, not anything in the web session. What types of abuse were you thinking?

index.bs
timcappalli: Yeah, that's fair. I will change to something similar to what you suggested.

index.bs
```diff
@@ -4046,7 +4073,59 @@ To override this default policy and indicate that a cross-origin <{iframe}> is a
...
+    1. If the fetch fails, the response does not have a content type of `application/json`, or does not have a status code (after following redirects) of 200, then throw a "{{SecurityError}}" {{DOMException}}.
+    1. If the body of the resource is not a valid JSON object, then throw a "{{SecurityError}}" {{DOMException}}.
+    1. If the value of the |origins| property of the JSON object is missing, or is not an array of strings, then throw a "{{SecurityError}}" {{DOMException}}.
+1. Let |labelsSeen| be a new empty [=set=].
+1. [=set/For each=] string in |origins|:
+    1. Let |url| be the result of parsing the string as a URL. If that fails, [=continue=].
+    1. Let |domain| be the [=effective domain=] of |url|. If that is null, [=continue=].
+    1. Remove any [=public suffix=] from the end of |domain|, including private registries and unknown registries. If |domain| is now empty, [=continue=].
+    1. Split |domain| into [=origin labels|labels=] and let |label| be the right-most one.
```
emlun, line 4123: https://github.com/w3c/webauthn/pull/2040#discussion_r1548479670 is still relevant here:

>I thought all of what precedes the eTLD is the label (i.e., the origin label of `login.example.co.uk` is `"login.example"`, not `["login", "example"]`)? Perhaps we could just define the "origin label" to be just the right-most domain segment apart from the eTLD, then we don't need to repeat the splitting logic here.

index.bs
```diff
@@ -4046,7 +4073,59 @@ To override this default policy and indicate that a cross-origin <{iframe}> is a
...
+    1. If |labelsSeen| [=set/contains=] |label|, [=continue=].
+    1. If |rpIdRequested| and |url| are [=same origin=], return [TRUE].
+    1. [=set/Append=] |label| to |labelsSeen|.
+    1. If the [=set/size=] of |labelsSeen| is greater than or equal to |maxLabels|, [=break=].
```
emlun, line 4127: On second read I think there's a wrinkle in this logic, namely that it returns immediately as soon as the 5th (say) _labelsSeen_ element is added, so the "If _labelsSeen_ contains _label_, continue" condition doesn't get a chance to trigger for _domain_ values past that. I.e., the origin `https://www.foo1.org` would fail with this related origin list:

- `https://foo1.org`
- `https://foo2.org`
- `https://foo3.org`
- `https://foo4.org`
- `https://foo5.org` (_labelsSeen_ is now `["foo1", "foo2", "foo3", "foo4", "foo5"]`, break and return `false`)
- `https://www.foo1.org` (_labelsSeen_ is still only 5 elements, so procedure would return `true` here)

I think this could be fixed by simply changing "greater than or equal to" to "greater than", if that is closer to the intent?

emlun (changes requested):

index.bs
emlun: Sorry, I saw [your reply](https://github.com/w3c/webauthn/pull/2040#discussion_r1583640476) now. In that case this use of the [origin label](https://pr-preview.s3.amazonaws.com/w3c/webauthn/pull/2040.html#origin-label) term is inconsistent with its definition, as this step seems to assume that "a label" is a single period-separated segment of an "origin label".

index.bs
timcappalli: Addressed in https://github.com/w3c/webauthn/pull/2040/commits/a2ac319dcf0d29ff6c38eadb7f6db8086a630a7b

index.bs
timcappalli: @emlun did a quick test with this code and I think `>=` is correct.

```javascript
var psl = require('psl');

const MAX_LABELS = 5;

//const WELL_KNOWN = {
//    origins: [
//        "https://shopping.sg",
//        "https://shopping.ie",
//        "https://shopping.co.uk",
//        "https://otherdomain1.com",
//        "https://otherdomain2.com",
//        "https://otherdomain3.com",
//        "https://otherdomain4.com",
//        "https://otherdomain5.com"
//    ]
//}

const WELL_KNOWN = {
    origins: [
        "https://otherdomain1.com",
        "https://otherdomain2.com",
        "https://otherdomain3.com",
        "https://otherdomain4.com",
        "https://otherdomain5.com",
        "https://otherdomain6.com"
    ]
}

function checkRpId(rpIdRequested) {
    const labelsSeen = new Set();

    for (const origin of WELL_KNOWN.origins) {
        const url = new URL(origin);

        let domain = url.hostname;

        if (!domain) continue;

        // grab eTLD+1 label
        const parsed = psl.parse(domain)
        const label = parsed.sld;

        if (labelsSeen.has(label)) continue;

        if (`https://${rpIdRequested}` === url.origin) return true;

        labelsSeen.add(label);

        if (labelsSeen.size >= MAX_LABELS) break;
    }
    return false;
}
```

Examples:

```javascript
console.log(checkRpId("otherdomain6.com"))
```
`false`

<hr>

```javascript
console.log(checkRpId("otherdomain5.com"))
```
`true`

index.bs
timcappalli: Good catch, addressed in https://github.com/w3c/webauthn/pull/2040/commits/b120220e6c0eed44f3c1eeb08cad18526c701977

index.bs
timcappalli: @MasterKale please sanity check ^

MasterKale: > If the [=set/size=] of |labelsSeen| is greater than or equal to |maxLabels|, [=break=].

If anything for sake of readability I'd put this check at the top of the list of steps. Consider the following amendment to the code above:

From this...
```js
for (const origin of WELL_KNOWN.origins) {
    // ...
    if (labelsSeen.size >= MAX_LABELS) break;
}
```
...to this:
```js
for (const origin of WELL_KNOWN.origins) {
    if (labelsSeen.size === MAX_LABELS) break;
    // ...
}
```

The latter leaves less room for edge cases as it more explicitly communicates the fact that only up to five related labels are supported.

MasterKale: I'm still considering Emil's question, one sec...

MasterKale: There appears to be a bug in this logic. RP ID scoping rules allow for RP ID `otherdomain5.com` to be specified if WebAuthn is handled from `https://www.otherdomain5.com`.

However, if I call `checkRpId("otherdomain5.com")` as written above, with the following list of related origins...

```js
let WELL_KNOWN = {
    origins: [
        "https://otherdomain1.com",
        "https://otherdomain2.com",
        "https://otherdomain3.com",
        "https://otherdomain4.com",
        "https://www.otherdomain5.com",
    ]
}
```

...the method return `false`. This list is still within the client's maximum of five labels, however I think this step is part of the problem:

> If |rpIdRequested| and |url| are [=same origin=], return [TRUE].

This likely needs to get tweaked to say something like, "rpIdRequested is valid for the current URL" or some such 🤔

timcappalli: > There appears to be a bug in this logic. RP ID scoping rules allow for RP ID otherdomain5.com to be specified if WebAuthn is handled from https://www.otherdomain5.com.

Related Origins logic would not be called in this case, as this would pass the existing RP < > origin check in the spec.

MasterKale: I forget about the checks that must occur before this logic is evaluated:

Screenshot 2024-06-14 at 10 23 29 AM

index.bs
```diff
@@ -4046,7 +4073,57 @@ To override this default policy and indicate that a cross-origin <{iframe}> is a
...
+    1. If the fetch fails, the response does not have a content type of `application/json`, or does not have a status code (after following redirects) of 200, then throw a "{{SecurityError}}" {{DOMException}}.
+    1. If the body of the resource is not a valid JSON object, then throw a "{{SecurityError}}" {{DOMException}}.
+    1. If the value of the |origins| property of the JSON object is missing, or is not an array of strings, then throw a "{{SecurityError}}" {{DOMException}}.
+1. Let |labelsSeen| be a new empty [=set=].
+1. [=set/For each=] string in |origins|:
+    1. Let |url| be the result of parsing the string as a URL. If that fails, [=continue=].
+    1. Let |domain| be the [=effective domain=] of |url|. If that is null, [=continue=].
+    1. Remove any [=public suffix=] from the end of |domain|, including private registries and unknown registries. If |domain| is now empty, [=continue=].
+    1. Split |domain| into |labels| and let |label| be the right-most one.
+    1. If |labelsSeen| [=set/contains=] |label|, [=continue=].
```
MasterKale, line 4122: If https://example.com/.well-known/webauthn served a list of related origins containing `["https://example.co.jp", "https://example.us"]`, this logic would not allow https://example.us to use `"example.com"` as an RP ID because the label `"example"` would end up in `labelsSeen` when processing `"https://www.example.co.jp"`. The next step wouldn't get to evaluate to return `TRUE`.

MasterKale (changes requested):

index.bs
MasterKale:
```js
const CLIENT_MAX_LABELS = 5;

// Hosted from https://example.com/.well-known
const WELL_KNOWN_WEBAUTHN = {
  origins: [
    "https://example.co.jp",
    "https://example.us",
    "https://otherexample.com",
    "https://onemoreexample.com",
  ]
};

function checkOriginIsRelated(currentOrigin) {
  const labelsSeen = new Set();

  const originURL = new URL(currentOrigin);
  const originDomain = originURL.hostname;
  const parsedOrigin = psl.parse(originDomain);

  for (let origin of WELL_KNOWN_WEBAUTHN.origins) {
    // 1. Let url be the result of parsing the string as a URL. If that fails, continue.
    const url = new URL(origin);

    // 2. Let domain be the effective domain of url. If that is null, continue.
    const domain = url.hostname;

    // 3. Remove any public suffix from the end of domain, including private registries and unknown
    // registries. If domain is now empty, continue.
    const parsedDomain = psl.parse(domain);

    if (parsedDomain.error) continue;

    // 4. Split domain into labels and let label be the right-most one.
    const label = parsedDomain.sld;

    if (!label) continue;

    // 5. If labelsSeen contains label, continue.
    if (labelsSeen.has(label)) continue;

    // 6. If callerOrigin's effective domain and domain are same origin, return True.
    if (parsedDomain.domain === parsedOrigin.domain) return true;

    // 7. Append label to labelsSeen.
    labelsSeen.add(label);

    // 8. If the size of labelsSeen is greater than or equal to maxLabels, break.
    if (labelsSeen.size >= CLIENT_MAX_LABELS) break;
  }

  return false;
}

// This is the second origin in the list, should be `true`
console.log(checkOriginIsRelated('https://example.us')) // false
```

MasterKale: I'm trying to think of the best way to fix this and right now I think removing this step completely might be it.

```suggestion
```

Keeping the step to add the label to `labelsSeen` after the domain check still seems valid, as does checking the size of `labelsSeen` afterwards to make sure it conforms to the client's maximum number of allowed labels.

But now I'm questioning the value of the limiting the number of allowed labels in this way. A list of origins like this would also fail to understand a related origin, despite the fact that the last item in the list of origins should be valid:

```
[
  "https://example2.com",
  "https://example3.com",
  "https://example4.com",
  "https://example5.com",
  "https://example6.com",
  "https://example2.net",
]
```

`"example6"` would be the fifth label so the loop would be terminated. However `"https://example2.net" should get evaluated because its label, `"example2"` is still within the five-label minimum allowed by this capability.

MasterKale: I just realized that `"example2.com"`, `"example2.net"`, and `"www.example2.us"` all have the same label `"example2"`. In a long enough list of related origins they may not get evaluated if the maximum number of labels gets hit before they're evaluated.

To what degree could we ask clients to lexicographically sort such a list of related origins by their label? It would remove some implementation burden from the RP, and prevent weird edge cases like this from complicating this logic further.

MasterKale: Alternatively if we wanted to maintain the implementation friction we could add in a step for RPs to perform the lexicographical sort manually when maintaining their list of related origins.

index.bs
```diff
@@ -4046,7 +4073,57 @@ To override this default policy and indicate that a cross-origin <{iframe}> is a
...
+    1. If |rpIdRequested| and |url| are [=same origin=], return [TRUE].
```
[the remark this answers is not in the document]
timcappalli: Addressed slightly different in https://github.com/w3c/webauthn/pull/2040/commits/bcec7faef08c3f5128e88d0b009492f544745e6b

index.bs
timcappalli: addressed in https://github.com/w3c/webauthn/pull/2040/commits/a108a2c79e081e78d68646de08cf78dafd2d0f21

timcappalli:
https://github.com/w3c/webauthn/pull/2040/commits/e4f24d9b3d1ec57f2dad79d6c210285699f4bda8 should address the issue where a new origin is added to the well-known that matches an existing label, but the max label count was already hit.

Some code to prove out the updated algorithm:

```javascript
// UPDATED LOGIC TO ADDRESS NEW ORIGINS WITH EXISTING LABELS

var psl = require('psl');

const MAX_LABELS = 5;

const WELL_KNOWN = {
  origins: [
    "https://shopping.sg",
    "https://shopping.co.uk",
    "https://otherdomain1.com",
    "https://otherdomain2.com",
    "https://otherdomain3.com",
    "https://shopping.ie",
    "https://otherdomain4.com",
    "https://otherdomain5.com",
    "https://otherdomain4.sg",
    "https://shopping.ms"
  ]
}

function checkCurrentOrigin(callingOrigin) {
  const labelsSeen = new Set();
  
  console.log(`Calling Origin: ${callingOrigin}`);

  for (const origin of WELL_KNOWN.origins) {
    console.log("--------------")
    console.log(`[0] START WK Origin: ${origin}`);

    const url = new URL(origin);
    console.log(`[1] WK URL: ${url}`);

    let domain = url.hostname;
    console.log(`[2] WK Domain: ${domain}`);

    if (!domain) continue;

    // grab eTLD+1 label
    const parsed = psl.parse(domain)
    console.log(`[3] WK Parsed: ${JSON.stringify(parsed)}`);

    const label = parsed.sld;
    console.log(`[4] WK Label: ${label}`);

    if (labelsSeen.size >= MAX_LABELS && !labelsSeen.has(label)) {
      console.log(`New label found, but limit exceeded, moving to next orign`);
      continue
    } ;

    if (`${callingOrigin}` == `${url.origin}`) return true;

    if (labelsSeen.has(label)) {
      console.log(`[5] label found`);
      console.log("moving to next origin...")
      continue;
    }

    if (labelsSeen.size < MAX_LABELS) {
      console.log(`[6] adding label: ${label}`);
      labelsSeen.add(label);
    }
    console.log(`labels seen: ${Array.from(labelsSeen)}`);
    console.log(`...got to end of loop`);
  }
  return false;
}

const callingOrigin = "https://shopping.ms"
console.log(checkCurrentOrigin(callingOrigin));
``

**Review by pascoej — APPROVED** [review]

**Review by agl — APPROVED** [review]

**Review by emlun — CHANGES_REQUESTED** [review]
I think the fact we've been dancing back and forth between issues caused by the max label count and origin ordering indicates we may need a different approach. I'm thinking a better design might be like this:

- Instead of a flat list of origins, `origins` is an object where keys are [_second level domains_ (SLDs)](https://developer.mozilla.org/en-US/docs/Glossary/Second-level_Domain).
- Each value in `origins` is a list of origins permitted for that SLD.
- If `origins` has too many (> ~5?) keys, or `origins[inputSld]` is too long (> ~100?), then the origin validation procedure returns `false`. These limits are determined by client policy.

So the example from the test code (I added `foo.bar.otherdomain4.sg` so there is a subdomain example):

```
{
  origins: [
    "https://shopping.sg",
    "https://shopping.co.uk",
    "https://otherdomain1.com",
    "https://otherdomain2.com",
    "https://otherdomain3.com",
    "https://shopping.ie",
    "https://otherdomain4.com",
    "https://otherdomain5.com",
    "https://otherdomain4.sg",
    "https://foo.bar.otherdomain4.sg",
    "https://shopping.ms"
  ]
}
```

would instead become (and be invalid since `origins` has more than 5 keys):

```
{
  origins: {
    "shopping": [
      "https://shopping.co.uk",
      "https://shopping.ie",
      "https://shopping.ms",
      "https://shopping.sg"
    ],
    "otherdomain1": ["https://otherdomain1.com"],
    "otherdomain2": ["https://otherdomain2.com"],
    "otherdomain3": ["https://otherdomain3.com"],
    "otherdomain4": [
      "https://otherdomain4.com",
      "https://otherdomain4.sg",
      "https://foo.bar.otherdomain4.sg"
    ],
    "otherdomain5": ["https://otherdomain5.com"]
  }
}
```

Would that work? This would at least make the origin label limit much less sensitive to ordering issues and much easier to understand.

I'm not sure whether there should be the limit on list lengths in that case, but I reckon there probably should be some kind of bound on the workload dumped on the client? Alternatively, we could require the RP to sort the origin lists so that the client can binary search for a match instead of having to linear search an unbounded list (and if the RP doesn't, the validation may result in a false negative).

Thoughts on that?

**Comment by timcappalli** [comment]
@emlun this was discussed on the last call. The preference was to put all the work on the client and keeping the well-known simple for RPs.

File: index.bs [hunk 30]
```diff
@@ -4046,7 +4073,63 @@ To override this default policy and indicate that a cross-origin <{iframe}> is a
...
+}
+</xmp>
+
+[=WebAuthn Clients=] supporting this feature MUST support at least five [=origin labels=]. Client policy SHOULD define an upper limit to prevent abuse.
+
+[=WebAuthn Clients=] supporting this feature SHOULD include {{ClientCapability/relatedOrigins}} in their response to [[#sctn-getClientCapabilities|getClientCapabilities()]].
+
+### Validating Related Origins ### {#sctn-validating-relation-origin}
+
+To validate the calling origin is an authorized related origin for a given ceremony:
```
emlun, line 4115: This should explicitly define _callerOrigin_ and _rpIdRequested_ as parameters of the procedure, and the references in the client algorithms should explicitly specify the arguments.

I think we should also put a `<dfn>` here and reference that rather than simply linking to the section with a different link text. Right now the link text "related origins validation procedure" appears nowhere else in the document, so readers have to rely solely on the hyperlink in order to discover what is actually referenced (this is not the only case of that throughout the spec, but IMO it's preferable to minimize them).

index.bs
```diff
@@ -4046,7 +4073,63 @@ To override this default policy and indicate that a cross-origin <{iframe}> is a
...
+1. Let |labelsSeen| be a new empty [=set=].
+1. [=set/For each=] string in |origins|:
+    1. Let |url| be the result of parsing the string as a URL. If that fails, [=continue=].
+    1. Let |domain| be the [=effective domain=] of |url|. If that is null, [=continue=].
+    1. Remove any [=public suffix=] from the end of |domain|, including private registries and unknown registries. If |domain| is now empty, [=continue=].
+    1. Split |domain| into |labels| and let |label| be the right-most one.
+    1. If the [=set/size=] of |labelsSeen| is greater than or equal to |maxLabels| and |labelsSeen|does not [=contains|contain=] |label|, [=continue=].
+    1. If |callerOrigin| and |url| are [=same origin=], return [TRUE].
+    1. If |labelsSeen| [=contains=] |label|, [=continue=].
+    1. If the [=set/size=] of |labelsSeen| is less than |maxLabels|, [=set/Append=] |label| to |labelsSeen|.
```
emlun, line 4131: Some space/casing typos, the alt-link-texts are not necessary, and the existence check is implicit in `set/append`:

```suggestion
    1. If the [=set/size=] of |labelsSeen| is greater than or equal to |maxLabels| and |labelsSeen| does not [=set/contain=] |label|, [=continue=].
    1. If |callerOrigin| and |url| are [=same origin=], return [TRUE].
    1. If the [=set/size=] of |labelsSeen| is less than |maxLabels|, [=set/append=] |label| to |labelsSeen|.
```

index.bs
```diff
@@ -4046,7 +4073,63 @@ To override this default policy and indicate that a cross-origin <{iframe}> is a
...
+To validate the calling origin is an authorized related origin for a given ceremony:
+
+1. Let |maxLabels| be the maximum number of [=origin labels=] allowed by client policy.
+1. Fetch the well-known URL for the RP ID (<code>https://|rpIdRequested|/.well-known/webauthn</code>).
+    1. If the fetch fails, the response does not have a content type of `application/json`, or does not have a status code (after following redirects) of 200, then throw a "{{SecurityError}}" {{DOMException}}.
+    1. If the body of the resource is not a valid JSON object, then throw a "{{SecurityError}}" {{DOMException}}.
+    1. If the value of the |origins| property of the JSON object is missing, or is not an array of strings, then throw a "{{SecurityError}}" {{DOMException}}.
+1. Let |labelsSeen| be a new empty [=set=].
+1. [=set/For each=] string in |origins|:
+    1. Let |url| be the result of parsing the string as a URL. If that fails, [=continue=].
```
emlun, line 4124: Infra defines the phrasing "For each _item_ of _list_" for ["for each"](https://infra.spec.whatwg.org/#list-iterate):

```suggestion
1. [=set/For each=] |originItem| of |origins|:
    1. Let |url| be the result of parsing |originItem| as a URL. If that fails, [=continue=].
```

emlun: Hm, and maybe we should use a reference to [URL parser](https://url.spec.whatwg.org/#concept-url-parser) here?

```suggestion
1. [=set/For each=] |originItem| of |origins|:
    1. Let |url| be the result of running the [=URL parser=] with |originItem| as the input. If that fails, [=continue=].
```

emlun (changes requested):

LeviSchuck:
Hello, I have some feedback regarding the minimum supported limit of five origins.

Amazon appears to have at least 17 TLDs that serve customers (rather than redirect customers to .com). Authentication occurs without redirecting to the `.com` origin. Should Amazon adopt related origins and somehow overcome the issue of prior passkey registrations on a localized TLD, they would be constrained by the five origin limit. Perhaps a limit of 20 would be more appropriate?

Here's the list I processed:
```
ae,ca,co.jp,co.uk,com,com.au,de,eg,es,fr,in,it,nl,pl,sa,se,sg
```

There may be more `co.{cctld}` and similar `com.{cctld}` that I missed. 

I have no affiliation with Amazon. They just seemed to be an interesting use case for this specification change.

Firehed:
@LeviSchuck The `label` described in the algorithm (if I've followed it correctly through my own local testing) would resolve to `amazon` on all of their owned TLDs, which counts as a singular entry towards the limit. I assume it was designed specifically with this use-case in mind.

I'd love to see this confirmed (or refuted) by one of the authors, but my read was that you'll only run into the limit if you're authenticating across a lot of wildly different domains (such as how signing in to Google often sends you to `youtube.com`)

MasterKale:
The algorithm is indeed focused more on the part after the TLD, a.k.a. the "label". So long as all 17 of those TLDs use `amazon` before them then the current algorithm defined in this PR will handle those just fine.

LeviSchuck:
Thank you for replying @Firehed ! @MasterKale helped me understand what a label means outside of this thread. 

I can think of no other case where 
1. Identity pools are shared across many origins, and
2. A consistent authentication origin is *not* used for each identity pool.

The limit of 5 labels seems fine.

index.bs
```diff
@@ -1228,6 +1228,9 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
 : <dfn>Non-Discoverable Credential</dfn>
 :: This is a [=credential=] whose [=credential ID=] must be provided in {{PublicKeyCredentialRequestOptions/allowCredentials}} when calling {{CredentialsContainer/get()|navigator.credentials.get()}} because it is not [=client-side discoverable credential|client-side discoverable=]. See also [=server-side credentials=].
 
+: <dfn>Origin Label</dfn>
+:: The portion of a domain name preceding the effective top-level domain, as defined in the Public Suffix List [[PSL]]. This is often referred to as the eTLD+1. For example, the origin label for example.co.uk and example.de is `example`.
```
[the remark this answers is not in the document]
emlun: Heh, and upon re-reading this I notice that my suggestion in prose was "registrable **domain** label" while my suggestions in code (here and below) are "registrable **origin** label". Another example of how confusingly synonymous these terms can be, I guess.

Anyway, I think "registrable **domain** label" is slightly better as a "subtype" of "domain label", but I don't have a strong preference for either variant.

index.bs
timcappalli: > This should explicitly define callerOrigin and rpIdRequested as parameters of the procedure, and the references in the client algorithms should explicitly specify the arguments.

Do you have an example of what that should look like in Bikeshed?

> I think we should also put a <dfn> here and reference that rather than simply linking to the section with a different link text. Right now the link text "related origins validation procedure" appears nowhere else in the document, so readers have to rely solely on the hyperlink in order to discover what is actually referenced (this is not the only case of that throughout the spec, but IMO it's preferable to minimize them).

Not sure I understand. What would a definition look like for a procedure? Isn't that just the section that defines the procedue?

emlun: We have a few examples in the spec:

- [**determines the set of origins on which the public key credential may be exercised**](https://w3c.github.io/webauthn/#determines-the-set-of-origins-on-which-the-public-key-credential-may-be-exercised) ([source](https://github.com/w3c/webauthn/blob/bb9f9bb53c17e50f20a5ab31e754f1f644362fb1/index.bs#L1340))
- [**looking up**](https://w3c.github.io/webauthn/#credential-id-looking-up) a credential ID in an authenticator ([source](https://github.com/w3c/webauthn/blob/bb9f9bb53c17e50f20a5ab31e754f1f644362fb1/index.bs#L4644))
- [**Create a new supplemental public key record**](https://w3c.github.io/webauthn/#abstract-opdef-create-a-new-supplemental-public-key-record) ([source](https://github.com/w3c/webauthn/blob/bb9f9bb53c17e50f20a5ab31e754f1f644362fb1/index.bs#L7582))
- (Arguably) [**effective resident key requirement for credential creation**](https://w3c.github.io/webauthn/#effective-resident-key-requirement-for-credential-creation) ([source](https://github.com/w3c/webauthn/blob/bb9f9bb53c17e50f20a5ab31e754f1f644362fb1/index.bs#L1998))

So for example, this section could start with:

```
### Validating Related Origins ### {#sctn-validating-relation-origin}

The <dfn abstract-op>related origins validation procedure</dfn>, given arguments |callerOrigin| and |rpIdRequested|,
is as follows:
```

and then the references in `create()` and `get()` could look like this:
```
                            1. Run the [$related origins validation procedure$] with arguments |callerOrigin| and |rpIdRequested|.
                                If the result is [FALSE], throw a "{{SecurityError}}" {{DOMException}}.
```

This has a few advantages: most importantly it makes the connection clearer and less reliant on hidden metadata, so it's more likely to survive format conversions such as into PDF or plain text (or even actual print, if someone were to actually do that), or quotations. Less importantly it also shows Bikeshed's list of references when you click the definition, which doesn't happen for links to sections.

Alternatively, the link to the section should be a plain section reference without replacing the link text, so that it renders the whole section heading, for the reasons above.

Like I said, we don't always respect these concerns throughout the spec, and I don't think we _always_ should, but I think this is one of the cases where there's little or nothing to gain from not doing so.

emlun: But to be fair, there is also at least this counter-example where we do link to a section with a (slightly) different link text:

- [limited verification algorithm](https://w3c.github.io/webauthn/#clientdatajson-development) ([source](https://github.com/w3c/webauthn/blob/bb9f9bb53c17e50f20a5ab31e754f1f644362fb1/index.bs#L3836))

timcappalli: Makes total sense. Addressed in [9da2c4b](https://github.com/w3c/webauthn/pull/2040/commits/9da2c4b4f55291376bf8da81ac9a4479e39c029b). Thanks for explaining @emlun!

index.bs
timcappalli: @nsatragno can you comment based on the discussion on the call?

index.bs
```diff
@@ -4046,7 +4078,62 @@ To override this default policy and indicate that a cross-origin <{iframe}> is a
...
+
+By default, Web Authentication requires that the [=RP ID=] be equal to the [=determines the set of origins on which the public key credential may be exercised|origin=]'s [=effective domain=], or a [=is a registrable domain suffix of or is equal to|registrable domain suffix=] of the [=determines the set of origins on which the public key credential may be exercised|origin=]'s [=effective domain=].
+
+This can make deployment challenging for large environments where multiple country-specific domains are in use (e.g. example.com vs example.co.uk vs example.sg), where alternative or brand domains are required (e.g. myexampletravel.com vs examplecruises.com), and/or where platform as a service providers are used to support mobile apps. 
+
+[=[WRPS]=] can opt in to allowing [=WebAuthn Clients=] to enable a credential to be created and used across a limited set of related [=origin|origins=].
+Such [=[RPS]=] MUST choose a common [=RP ID=] to use across all ceremonies from related origins.
+
+A JSON document MUST be hosted at the `webauthn` well-known URL ([[!RFC8615]]) for the [=RP ID=]. The JSON document MUST be returned as follows:
```
emlun, line 4090: I don't think the parentheses are needed.
```suggestion
A JSON document MUST be hosted at the `webauthn` well-known URL [[!RFC8615]] for the [=RP ID=]. The JSON document MUST be returned as follows:
```

index.bs
```diff
@@ -4046,7 +4078,62 @@ To override this default policy and indicate that a cross-origin <{iframe}> is a
...
+[=WebAuthn Clients=] supporting this feature MUST support at least five [=registrable origin labels=]. Client policy SHOULD define an upper limit to prevent abuse.
+
+[=WebAuthn Clients=] supporting this feature SHOULD include {{ClientCapability/relatedOrigins}} in their response to [[#sctn-getClientCapabilities|getClientCapabilities()]].
+
+### Validating Related Origins ### {#sctn-validating-relation-origin}
+
+The <dfn abstract-op>related origins validation procedure</dfn>, given arguments |callerOrigin| and |rpIdRequested|, is as follows:
+
+1. Let |maxLabels| be the maximum number of [=registrable origin labels=] allowed by client policy.
+1. Fetch the well-known URL for the RP ID (<code>https://|rpIdRequested|/.well-known/webauthn</code>).
```
emlun, line 4123: I don't think we've formally defined the string interpolation syntax, so let's also reference `rpIdRequested` outside it and leave the parentheses as an informational hint.

```suggestion
1. Fetch the `webauthn` well-known URL [[!RFC8615]] for the RP ID |rpIdRequested| (i.e., <code>https://|rpIdRequested|/.well-known/webauthn</code>).
```

emlun (approved):
LGTM, good to merge with the addition of a couple more polish opportunities:

supersensen:
From the perspective of decentralized network user authentication, the /.well-known/webauthn file is like providing hackers with a complete target list. 

Therefore, it is hoped that from the "PublicKeyCredentialCreationOptions" before creating credentials at registration, it can be defined whether origin verification is required. If it is not required, the browser should be informed that any domain can pass the verification. 

The right to publicly disclose the origins list should be entrusted to the users who make use of it, and hopes that it will be given due attention.

agl:
> Therefore, it is hoped that from the "PublicKeyCredentialCreationOptions" before creating credentials at registration, it can be defined whether origin verification is required. If it is not required, the browser should be informed that any domain can pass the verification.

No domain is required to populate that file. The existing RP ID validation rules still apply and, if satisfied, no .well-known request will be made.

PR #2041, opened 2024-03-13
Include supported extensions in getClientCapabilities.

agl:
It is of somewhat limited value because getClientCapabilities can't know what extensions the user's authenticator will support, but it can be meaningful to know that an extension isn't supported. Absent this, sites are guessing based on browser names and versions, but that is fragile.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/pull/2041.html" title="Last updated on Mar 13, 2024, 11:12 PM UTC (1bf960f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/2041/354a717...1bf960f.html" title="Last updated on Mar 13, 2024, 11:12 PM UTC (1bf960f)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2775,12 +2775,16 @@ This method has no arguments and returns a record of capability keys to Boolean 
 </xmp>
 
 [=map/Keys=] in {{PublicKeyCredentialClientCapabilities}} MUST be sorted in ascending lexicographical order.
-The set of [=map/keys=] SHOULD equal the set of [=enumeration values=] of {{ClientCapability}},
+The set of [=map/keys=] SHOULD contain the set of [=enumeration values=] of {{ClientCapability}},
 but the client MAY omit keys as it deems necessary; see [[#sctn-disclosing-client-capabilities]].
 
 When the value for a given capability is [TRUE], the feature is known to be currently supported by the client.
 When the value for a given capability is [FALSE], the feature is known to be not currently supported by the client.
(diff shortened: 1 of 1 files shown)


MasterKale (approved):
I like this change a lot 🔥

Kieun:
It took five years since I brought this requirement.

timcappalli (approved):
LGTM

emlun:
I'm considering whether clients should also be allowed to set `false` for extensions that are known and not supported, to give RPs a definitive answer in those cases. That would probably include the extensions defined inline in the spec, but probably not many extensions defined elsewhere. But maybe that makes for too much variation? I'm not sure, this is probably fine either way.

g-davidson:
This looks great!

msft-bob:
Looks good. Thanks for putting this together.

nsatragno:
Are we concerned that there will be a gap until we get the browsers updated where extensions will be supported, but this capability will not. Are we risking those RPs assuming the extension is not supported when it is during that gap?

(same comment as in #2051)

emlun:
Both the method and the `extension:` capabilities are added in L3, so formally speaking there cannot be any mature implementation of `getClientCapabilities()` that does not include the `extension:` capabilities. At least in my opinion, any current implementations of `getClientCapabilities()` should be considered experimental from a spec point of view until L3 reaches at least CR maturity. So no, I don't think we should worry about a gap like that, because RPs should not be relying on these features yet.

Issue #1859, opened 2023-02-24
Clarify how to differentiate between exceptions


dolda2000:
## Proposed Change

As far as I have been able to tell, exceptions from the WebAuthn UI interactions fall into three general categories, that are fairly important to distinguish:

1. Totally normal exceptions that the user initiated and is aware of. An example would be the user simply cancelling the interaction. The user is well aware of what happened and doesn't need to (shouldn't, in fact) be further informed of it.
2. Semi-normal abortions of the UI, that happened because of user interaction, but where the user may not know exactly what happened and should be informed of it. An example would be that, on Firefox, if the user uses an authenticator that doesn't contain any of the credentials among `allowCredentials`, Firefox simply closes the UI and returns this fact to the RP as an exception. The user might not be aware that he used the wrong authenticator and should be told about it. (Interestingly this seems to be different on Chrome, where the UI directly informs the user of this fact and offers to retry or cancel.)
3. Actual exceptional conditions where the RP did something wrong. An example would be providing malformed or otherwise invalid options objects. In this case, the RP should be able to tell the user that something unexpected happened and to perhaps contact support if the problem persists.

However, I cannot find anything in the spec about how to make these distinctions, and neither have I been able to infer anything from actually observed exceptions. Exceptions that fall into different categories seem to commonly use the same error name, leaving only the exception message to infer anything from, but they are apparently primarily intended to be human-readable, localizable, and differ between different browsers.

I am currently just relying on Chrome's behavior to directly tell the user of various aborting conditions (which is nice) and the complete and entire bug-freeness of my site, but obviously neither of those are particularly functional assumptions, and in fact on Firefox several interactions are quite suboptimal, so I'd really appreciate some clarity on the issue.

MasterKale:
You have to go digging into the spec to connect the pieces, but the spec definitely outlines discrete error scenarios that should be reliably detectable. For sake of discussion, I've done so with my own library for both registration and authentication:

- https://github.com/MasterKale/SimpleWebAuthn/blob/master/packages/browser/src/helpers/identifyRegistrationError.ts
- https://github.com/MasterKale/SimpleWebAuthn/blob/master/packages/browser/src/helpers/identifyAuthenticationError.ts

[...]

PR #2047, opened 2024-03-20
Help RP's understand actionable exceptions from `create()` and `get()`

MasterKale:
This PR attempts to pull together any exceptions raised by `create()` and `get()` to help RP's understand what exceptions may be encountered when using WebAuthn. The intention here is to help RP's understand which errors might be surfaced to the user, and which should not.

Addresses #1859.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/pull/2047.html" title="Last updated on Jul 31, 2024, 4:56 PM UTC (2b692fa)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/2047/2b8e368...2b692fa.html" title="Last updated on Jul 31, 2024, 4:56 PM UTC (2b692fa)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2253,7 +2253,7 @@ authorizing an authenticator. When <code>|options|.{{CredentialCreationOptions/m
         or <code>{{PublicKeyCredentialCreationOptions/rp}}.{{PublicKeyCredentialRpEntity/id}}</code> was not a registrable domain suffix of nor was equal to the [=effective domain=].
 
     :   {{TypeError}}
-    ::  The <code>|options|</code> argument was not a valid {{CredentialCreationOptions}} value,
+    ::  The <code>|options|</code> argument was not a valid [[#sctn-credentialcreationoptions-extension|CredentialCreationOptions]] value,
         or the value of <code>{{PublicKeyCredentialCreationOptions/user}}.{{PublicKeyCredentialUserEntity/id}}</code> was empty or was longer than 64 bytes.
 
     :   {{UnknownError}}
--- a/index.bs
+++ b/index.bs
@@ -2229,6 +2229,8 @@ authorizing an authenticator. When <code>|options|.{{CredentialCreationOptions/m
 
 #### Create Request Exceptions #### {#sctn-create-request-exceptions}
 
+[INFORMATIVE]
+
 [=[WRPS]=] can encounter the following {{DOMException|DOMExceptions}} from a call to {{CredentialsContainer/create()|navigator.credentials.create()}}. Some errors can have multiple reasons for why they happened, requiring the [=[WRPS]=] to infer the actual reason based on their use of WebAuthn:
 
 <dl>
@@ -2770,6 +2772,8 @@ The steps for [=issuing a credential request to an authenticator=] are as follow
     1. Return [TRUE].
 
 #### Get Request Exceptions #### {#sctn-get-request-exceptions}
+
+[INFORMATIVE]
 
 [=[WRPS]=] can encounter the following {{DOMException|DOMExceptions}} from a call to {{CredentialsContainer/get()|navigator.credentials.get()}}:
 
--- a/index.bs
+++ b/index.bs
@@ -2745,20 +2745,26 @@ The steps for [=issuing a credential request to an authenticator=] are as follow
 
 #### Authentication API Exceptions
 
-[=[RP]=]'s should prepare to encounter the following {{DOMException|DOMExceptions}} from a call to {{CredentialsContainer/get()|navigator.credentials.get()}}:
+[=[WRPS]=] can encounter the following {{DOMException|DOMExceptions}} from a call to {{CredentialsContainer/get()|navigator.credentials.get()}}:
 
 <dl dfn-type="argument" dfn-for="Authentication API Exceptions">
     :   <dfn>AbortError</dfn>
-    ::  Definition here
-
-    :   <dfn>NotAllowedError</dfn>
-    ::  Definition here
+    ::  The ceremony was cancelled via an {{AbortController}}
+        (see [[#sctn-abortoperation]] and [[#sctn-sample-aborting]] for more information.)
 
     :   <dfn>SecurityError</dfn>
-    ::  Definition here
+    ::  The [=effective domain=] was not a [=valid domain=],
+        or {{PublicKeyCredentialCreationOptions/rp}}.{{PublicKeyCredentialRpEntity/id}} was not a registrable domain suffix of nor was equal to the [=effective domain=].
 
     :   <dfn>UnknownError</dfn>
-    ::  Definition here
+    ::  The [=authenticator=] could not process the supplied options,
+        or encountered an error while generating an [=assertion signature=].
+
+    :   <dfn>NotAllowedError</dfn>
+    ::  A catch-all error covering a wide range of possible reasons,
+        including common ones like the user canceling out of the ceremony.
+        Some of these causes are documented throughout this spec,
+        while others are client-specific.
 </dl>
 
 ### Store an Existing Credential - PublicKeyCredential's `[[Store]](credential, sameOriginWithAncestors)` Method ### {#sctn-storeCredential}
@@ -5919,7 +5925,7 @@ The attributes above are structured within this certificate as such:
   04 12                                  -- OCTET STRING
     04  10                               -- OCTET STRING
       CD 8C 39 5C 26 ED EE DE            -- AAGUID cd8c395c-26ed-eede-653b-00797d03ca3c
-      65 3B 00 79 7D 03 CA 3C 
+      65 3B 00 79 7D 03 CA 3C
 
 30 12                                    -- SEQUENCE
   06 0B 2B 06 01 04 01 82 E5 1C 01 01 05 -- OID 1.3.6.1.4.1.45724.1.1.5
--- a/index.bs
+++ b/index.bs
@@ -2285,6 +2285,9 @@ The following {{DOMException}} exceptions can be raised:
     :   {{SecurityError}}
     ::  The [=effective domain=] was not a [=valid domain=],
         or <code>{{PublicKeyCredentialCreationOptions/rp}}.{{PublicKeyCredentialRpEntity/id}}</code> was not equal to or a registrable domain suffix of the [=effective domain=].
+        In the case of [[#sctn-related-origins|related origin requests]],
+        the [=authenticator=] does not support it
+        or there were issues when [[#sctn-validating-relation-origin|validating related origins]].
 
     :   {{TimeoutError}}
     ::  The ceremony was cancelled by the user agent after exceeding the time limit permitted for the ceremony.
@@ -2848,6 +2851,9 @@ The following {{DOMException}} exceptions can be raised:
     :   {{SecurityError}}
     ::  The [=effective domain=] was not a [=valid domain=],
         or <code>{{PublicKeyCredentialCreationOptions/rp}}.{{PublicKeyCredentialRpEntity/id}}</code> was not equal to or a registrable domain suffix of the [=effective domain=].
+        In the case of [[#sctn-related-origins|related origin requests]],
+        the [=authenticator=] does not support it
+        or there were issues when [[#sctn-validating-relation-origin|validating related origins]].
 
     :   {{TimeoutError}}
     ::  The ceremony was cancelled by the user agent after exceeding the time limit permitted for the ceremony.
--- a/index.bs
+++ b/index.bs
@@ -2204,32 +2204,42 @@ authorizing an authenticator.
 
 #### Registration API Exceptions
 
-[=[RP]=]'s should prepare to encounter the following {{DOMException|DOMExceptions}} from a call to {{CredentialsContainer/create()|navigator.credentials.create()}}:
+[=[WRPS]=] can encounter the following {{DOMException|DOMExceptions}} from a call to {{CredentialsContainer/create()|navigator.credentials.create()}}. Some errors can have multiple reasons for why they happened, requiring the [=[WRPS]=] to infer the actual reason based on their use of WebAuthn:
 
 <dl dfn-type="argument" dfn-for="Registration API Exceptions">
     :   <dfn>AbortError</dfn>
-    ::  Definition here
+    ::  The ceremony was cancelled via an {{AbortController}}
+        (see [[#sctn-abortoperation]] and [[#sctn-sample-aborting]] for more information.)
 
     :   <dfn>ConstraintError</dfn>
-    ::  Definition here
+    ::  Either {{residentKey}} was set to "{{ResidentKeyRequirement/required}}" and no available authenticator supported resident keys,
+        or {{userVerification}} was set to "{{UserVerificationRequirement/required}}" and no available authenticator could perform [=user verification=].
 
     :   <dfn>InvalidStateError</dfn>
-    ::  Definition here
-
-    :   <dfn>{{NotAllowedError}}</dfn>
-    ::  Definition here
+    ::  The authenticator used in the ceremony recognized an entry in {{excludeCredentials}}
+        after the user [=user consent|consented=] to registering a credential.
 
     :   <dfn>NotSupportedError</dfn>
-    ::  Definition here
+    ::  No entry in {{PublicKeyCredentialCreationOptions/pubKeyCredParams}} had a `type` property of "{{PublicKeyCredentialType/public-key}}",
+        or the [=authenticator=] did not support any of the specified cryptographic parameters in {{PublicKeyCredentialCreationOptions/pubKeyCredParams}}.
 
     :   <dfn>SecurityError</dfn>
-    ::  Definition here
+    ::  The [=effective domain=] was not a [=valid domain=],
+        or {{PublicKeyCredentialCreationOptions/rp}}.{{PublicKeyCredentialRpEntity/id}} was not a registrable domain suffix of nor was equal to the [=effective domain=].
 
     :   <dfn>TypeError</dfn>
-    ::  Definition here
+    ::  The value of {{PublicKeyCredentialCreationOptions/user}}.{{PublicKeyCredentialUserEntity/id}} was not between 1 and 64 bytes (inclusive.)
 
     :   <dfn>UnknownError</dfn>
-    ::  Definition here
+    ::  The [=authenticator=] could not process the supplied options,
+        or encountered an error while creating the new credential object.
+
+    :   <dfn>{{NotAllowedError}}</dfn>
+    ::  A catch-all error covering a wide range of possible reasons,
+        including common ones like the user canceling out of the ceremony.
+        Some of these causes are documented throughout this spec,
+        while others are client-specific.
+
 </dl>
 
 ### Use an Existing Credential to Make an Assertion - PublicKeyCredential's `[[Get]](options)` Method ### {#sctn-getAssertion}
--- a/index.bs
+++ b/index.bs
@@ -2289,10 +2289,6 @@ The following {{DOMException}} exceptions can be raised:
         the [=authenticator=] does not support it
         or there were issues when [[#sctn-validating-relation-origin|validating related origins]].
 
-    :   {{TimeoutError}}
-    ::  The ceremony was cancelled by the user agent after exceeding the time limit permitted for the ceremony.
-        See [[#sctn-timeout-recommended-range]] for more information.
-
     :   {{UnknownError}}
     ::  The [=authenticator=] could not process the supplied options,
         or encountered an error while creating the new credential.
@@ -2854,10 +2850,6 @@ The following {{DOMException}} exceptions can be raised:
         In the case of [[#sctn-related-origins|related origin requests]],
         the [=authenticator=] does not support it
         or there were issues when [[#sctn-validating-relation-origin|validating related origins]].
-
-    :   {{TimeoutError}}
-    ::  The ceremony was cancelled by the user agent after exceeding the time limit permitted for the ceremony.
-        See [[#sctn-timeout-recommended-range]] for more information.
 
     :   {{UnknownError}}
     ::  The [=authenticator=] could not process the supplied options,
--- a/index.bs
+++ b/index.bs
@@ -2256,7 +2256,7 @@ authorizing an authenticator. When <code>|options|.{{CredentialCreationOptions/m
 [INFORMATIVE]
 
 [=[WRPS]=] can encounter a number of exceptions from a call to {{CredentialsContainer/create()|navigator.credentials.create()}}.
-Some errors can have multiple reasons for why they happened,
+Some exceptions can have multiple reasons for why they happened,
 requiring the [=[WRPS]=] to infer the actual reason based on their use of WebAuthn.
 Some exceptions that can be raised during processing of [=WebAuthn Extensions=]
 defined outside of this specification may not be listed here.
@@ -2829,7 +2829,7 @@ The steps for [=issuing a credential request to an authenticator=] are as follow
 [INFORMATIVE]
 
 [=[WRPS]=] can encounter a number of exceptions from a call to {{CredentialsContainer/get()|navigator.credentials.get()}}.
-Some errors can have multiple reasons for why they happened,
+Some exceptions can have multiple reasons for why they happened,
 requiring the [=[WRPS]=] to infer the actual reason based on their use of WebAuthn.
 Some exceptions that can be raised during processing of [=WebAuthn Extensions=]
 defined outside of this specification may not be listed here.
--- a/index.bs
+++ b/index.bs
@@ -2285,9 +2285,9 @@ The following {{DOMException}} exceptions can be raised:
     :   {{SecurityError}}
     ::  The [=effective domain=] was not a [=valid domain=],
         or <code>{{PublicKeyCredentialCreationOptions/rp}}.{{PublicKeyCredentialRpEntity/id}}</code> was not equal to or a registrable domain suffix of the [=effective domain=].
-        In the case of [[#sctn-related-origins|related origin requests]],
-        the [=authenticator=] does not support it
-        or there were issues when [[#sctn-validating-relation-origin|validating related origins]].
+        In the latter case,
+        the [=client=] does not support [[#sctn-related-origins|related origin requests]]
+        or the [$related origins validation procedure$] failed.
 
     :   {{UnknownError}}
     ::  The [=authenticator=] could not process the supplied options,
@@ -2847,9 +2847,9 @@ The following {{DOMException}} exceptions can be raised:
     :   {{SecurityError}}
     ::  The [=effective domain=] was not a [=valid domain=],
         or <code>{{PublicKeyCredentialCreationOptions/rp}}.{{PublicKeyCredentialRpEntity/id}}</code> was not equal to or a registrable domain suffix of the [=effective domain=].
-        In the case of [[#sctn-related-origins|related origin requests]],
-        the [=authenticator=] does not support it
-        or there were issues when [[#sctn-validating-relation-origin|validating related origins]].
+        In the latter case,
+        the [=client=] does not support [[#sctn-related-origins|related origin requests]]
+        or the [$related origins validation procedure$] failed.
 
     :   {{UnknownError}}
     ::  The [=authenticator=] could not process the supplied options,
--- a/index.bs
+++ b/index.bs
@@ -264,7 +264,6 @@ spec: SP800-800-63r3; urlPrefix: https://pages.nist.gov/800-63-3/sp800-63-3.html
 spec: webidl; urlPrefix: https://webidl.spec.whatwg.org
     type: dfn;
         text: get a copy of the bytes held by the buffer source; url: dfn-get-buffer-source-copy
-        text: Exceptions; url: idl-exceptions
 
 spec: webdriver; urlPrefix: https://w3c.github.io/webdriver/
     type: dfn
@@ -2303,7 +2302,7 @@ The following {{DOMException}} exceptions can be raised:
 
 </dl>
 
-The following <code>[=Exceptions|Exception=]</code> exceptions can be raised:
+The following [=simple exceptions=] can be raised:
 
 <dl>
 
@@ -2865,7 +2864,7 @@ The following {{DOMException}} exceptions can be raised:
         while others are client-specific.
 </dl>
 
-The following <code>[=Exceptions|Exception=]</code> exceptions can be raised:
+The following [=simple exceptions=] can be raised:
 
 <dl>
 
--- a/index.bs
+++ b/index.bs
@@ -2253,7 +2253,8 @@ authorizing an authenticator. When <code>|options|.{{CredentialCreationOptions/m
         or <code>{{PublicKeyCredentialCreationOptions/rp}}.{{PublicKeyCredentialRpEntity/id}}</code> was not a registrable domain suffix of nor was equal to the [=effective domain=].
 
     :   {{TypeError}}
-    ::  The value of <code>{{PublicKeyCredentialCreationOptions/user}}.{{PublicKeyCredentialUserEntity/id}}</code> was empty or was longer than 64 bytes.
+    ::  The <code>|options|</code> argument was not a valid {{CredentialCreationOptions}} value,
+        or the value of <code>{{PublicKeyCredentialCreationOptions/user}}.{{PublicKeyCredentialUserEntity/id}}</code> was empty or was longer than 64 bytes.
 
     :   {{UnknownError}}
     ::  The [=authenticator=] could not process the supplied options,
--- a/index.bs
+++ b/index.bs
@@ -2227,7 +2227,7 @@ During the above process, the user agent SHOULD show some UI to the user to guid
 authorizing an authenticator. When <code>|options|.{{CredentialCreationOptions/mediation}}</code> is set to {{CredentialMediationRequirement/conditional}}, prominent modal UI should <i>not</i> be shown <i>unless</i> credential creation was previously consented to via means determined by the user agent.
 </div>
 
-#### Registration API Exceptions #### {#sctn-registration-api-exceptions}
+#### Create Request Exceptions #### {#sctn-create-request-exceptions}
 
 [=[WRPS]=] can encounter the following {{DOMException|DOMExceptions}} from a call to {{CredentialsContainer/create()|navigator.credentials.create()}}. Some errors can have multiple reasons for why they happened, requiring the [=[WRPS]=] to infer the actual reason based on their use of WebAuthn:
 
@@ -2769,7 +2769,7 @@ The steps for [=issuing a credential request to an authenticator=] are as follow
 
     1. Return [TRUE].
 
-#### Authentication API Exceptions #### {#sctn-authentication-api-exceptions}
+#### Get Request Exceptions #### {#sctn-get-request-exceptions}
 
 [=[WRPS]=] can encounter the following {{DOMException|DOMExceptions}} from a call to {{CredentialsContainer/get()|navigator.credentials.get()}}:
 
--- a/index.bs
+++ b/index.bs
@@ -2787,6 +2787,9 @@ The steps for [=issuing a credential request to an authenticator=] are as follow
     ::  The [=effective domain=] was not a [=valid domain=],
         or <code>{{PublicKeyCredentialCreationOptions/rp}}.{{PublicKeyCredentialRpEntity/id}}</code> was not equal to or a registrable domain suffix of the [=effective domain=].
 
+    :   {{TypeError}}
+    ::  The <code>|options|</code> argument was not a valid [[#sctn-credentialrequestoptions-extension|CredentialRequestOptions]] value.
+
     :   {{UnknownError}}
     ::  The [=authenticator=] could not process the supplied options,
         or encountered an error while generating an [=assertion signature=].
--- a/index.bs
+++ b/index.bs
@@ -2227,7 +2227,7 @@ During the above process, the user agent SHOULD show some UI to the user to guid
 authorizing an authenticator. When <code>|options|.{{CredentialCreationOptions/mediation}}</code> is set to {{CredentialMediationRequirement/conditional}}, prominent modal UI should <i>not</i> be shown <i>unless</i> credential creation was previously consented to via means determined by the user agent.
 </div>
 
-#### Registration API Exceptions ### {#sctn-registration-api-exceptions}
+#### Registration API Exceptions #### {#sctn-registration-api-exceptions}
 
 [=[WRPS]=] can encounter the following {{DOMException|DOMExceptions}} from a call to {{CredentialsContainer/create()|navigator.credentials.create()}}. Some errors can have multiple reasons for why they happened, requiring the [=[WRPS]=] to infer the actual reason based on their use of WebAuthn:
 
@@ -2768,7 +2768,7 @@ The steps for [=issuing a credential request to an authenticator=] are as follow
 
     1. Return [TRUE].
 
-#### Authentication API Exceptions ### {#sctn-authentication-api-exceptions}
+#### Authentication API Exceptions #### {#sctn-authentication-api-exceptions}
 
 [=[WRPS]=] can encounter the following {{DOMException|DOMExceptions}} from a call to {{CredentialsContainer/get()|navigator.credentials.get()}}:
 
--- a/index.bs
+++ b/index.bs
@@ -5999,7 +5999,7 @@ The attributes above are structured within this certificate as such:
   04 12                                  -- OCTET STRING
     04  10                               -- OCTET STRING
       CD 8C 39 5C 26 ED EE DE            -- AAGUID cd8c395c-26ed-eede-653b-00797d03ca3c
-      65 3B 00 79 7D 03 CA 3C
+      65 3B 00 79 7D 03 CA 3C 
 
 30 12                                    -- SEQUENCE
   06 0B 2B 06 01 04 01 82 E5 1C 01 01 05 -- OID 1.3.6.1.4.1.45724.1.1.5
--- a/index.bs
+++ b/index.bs
@@ -2259,8 +2259,10 @@ authorizing an authenticator. When <code>|options|.{{CredentialCreationOptions/m
 [=[WRPS]=] can encounter a number of exceptions from a call to {{CredentialsContainer/create()|navigator.credentials.create()}}.
 Some exceptions can have multiple reasons for why they happened,
 requiring the [=[WRPS]=] to infer the actual reason based on their use of WebAuthn.
-Some exceptions that can be raised during processing of [=WebAuthn Extensions=]
-defined outside of this specification may not be listed here.
+
+Note: Exceptions that can be raised during processing of any [=WebAuthn Extensions=],
+including ones defined outside of this specification,
+are not listed here.
 
 The following {{DOMException}} exceptions can be raised:
 
@@ -2832,8 +2834,10 @@ The steps for [=issuing a credential request to an authenticator=] are as follow
 [=[WRPS]=] can encounter a number of exceptions from a call to {{CredentialsContainer/get()|navigator.credentials.get()}}.
 Some exceptions can have multiple reasons for why they happened,
 requiring the [=[WRPS]=] to infer the actual reason based on their use of WebAuthn.
-Some exceptions that can be raised during processing of [=WebAuthn Extensions=]
-defined outside of this specification may not be listed here.
+
+Note: Exceptions that can be raised during processing of any [=WebAuthn Extensions=],
+including ones defined outside of this specification,
+are not listed here.
 
 The following {{DOMException}} exceptions can be raised:
 
--- a/index.bs
+++ b/index.bs
@@ -2280,7 +2280,7 @@ The following {{DOMException}} exceptions can be raised:
 
     :   {{NotSupportedError}}
     ::  No entry in {{PublicKeyCredentialCreationOptions/pubKeyCredParams}} had a {{PublicKeyCredentialDescriptor/type}} property of {{PublicKeyCredentialType/public-key}},
-        or the [=authenticator=] did not support any of the specified cryptographic parameters in {{PublicKeyCredentialCreationOptions/pubKeyCredParams}}.
+        or the [=authenticator=] did not support any of the signature algorithms specified in {{PublicKeyCredentialCreationOptions/pubKeyCredParams}}.
 
     :   {{SecurityError}}
     ::  The [=effective domain=] was not a [=valid domain=],
--- a/index.bs
+++ b/index.bs
@@ -2231,7 +2231,11 @@ authorizing an authenticator. When <code>|options|.{{CredentialCreationOptions/m
 
 [INFORMATIVE]
 
-[=[WRPS]=] can encounter the following {{DOMException|DOMExceptions}} from a call to {{CredentialsContainer/create()|navigator.credentials.create()}}. Some errors can have multiple reasons for why they happened, requiring the [=[WRPS]=] to infer the actual reason based on their use of WebAuthn:
+[=[WRPS]=] can encounter the following {{DOMException|DOMExceptions}} from a call to {{CredentialsContainer/create()|navigator.credentials.create()}}.
+Some errors can have multiple reasons for why they happened,
+requiring the [=[WRPS]=] to infer the actual reason based on their use of WebAuthn.
+Some exceptions that can be raised during processing of [=WebAuthn Extensions=]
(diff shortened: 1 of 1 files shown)


MasterKale:
I've finally cobbled together reasons for all of the exceptions during both registration and authentication.

...Except I cop out a bit with `NotAllowedError` because it has many more possible reasons it gets raised, and in practice clients have overloaded this exception with causes not documented in the spec. I thought it prudent to present this error as one that RPs should prepare to handle as a general, "the user canceled the ceremony, or something went wrong" exception and handle it as such. This is as opposed to encouraging each RP to try and interpret all possible reasons the issue was raised. I'm open to feedback on this approach.

herrjemand:
100% support this @MasterKale. Recently working on client lib, NotAllowedError is kinda useless. *)

index.bs
```diff
@@ -2202,6 +2202,45 @@ During the above process, the user agent SHOULD show some UI to the user to guid
...
+    :   <dfn>NotSupportedError</dfn>
+    ::  No entry in {{PublicKeyCredentialCreationOptions/pubKeyCredParams}} had a `type` property of "{{PublicKeyCredentialType/public-key}}",
+        or the [=authenticator=] did not support any of the specified cryptographic parameters in {{PublicKeyCredentialCreationOptions/pubKeyCredParams}}.
+
+    :   <dfn>SecurityError</dfn>
+    ::  The [=effective domain=] was not a [=valid domain=],
+        or {{PublicKeyCredentialCreationOptions/rp}}.{{PublicKeyCredentialRpEntity/id}} was not a registrable domain suffix of nor was equal to the [=effective domain=].
+
+    :   <dfn>TypeError</dfn>
+    ::  The value of {{PublicKeyCredentialCreationOptions/user}}.{{PublicKeyCredentialUserEntity/id}} was not between 1 and 64 bytes (inclusive.)
```
emlun, line 2231: `TypeError` is thrown in other type error cases as well, like passing a string in place of a byte array or an array in place of an object (but notably not passing a byte array in place of a string - I assume it gets implicitly `.toString()`ed):

```js
  publicKey: {
    rp: { name: "Test" },
    user: { id: "1234", name: "Test", displayName: "Test" },
    pubKeyCredParams: [{ type: "public-key", alg: -7 }],
    challenge: crypto.getRandomValues(new Uint8Array(32)),
  }
})

Uncaught (in promise) TypeError: Failed to execute 'create' on 'CredentialsContainer': Failed to read the 'publicKey' property from 'CredentialCreationOptions': Failed to read the 'user' property from 'PublicKeyCredentialCreationOptions': Failed to read the 'id' property from 'PublicKeyCredentialUserEntity': The provided value is not of type '(ArrayBuffer or ArrayBufferView)'.

```

index.bs
```diff
@@ -2202,6 +2202,45 @@ During the above process, the user agent SHOULD show some UI to the user to guid
...
 </div>
 
+#### Registration API Exceptions
+
+[=[WRPS]=] can encounter the following {{DOMException|DOMExceptions}} from a call to {{CredentialsContainer/create()|navigator.credentials.create()}}. Some errors can have multiple reasons for why they happened, requiring the [=[WRPS]=] to infer the actual reason based on their use of WebAuthn:
+
+<dl dfn-type="argument" dfn-for="Registration API Exceptions">
+    :   <dfn>AbortError</dfn>
+    ::  The ceremony was cancelled via an {{AbortController}}
+        (see [[#sctn-abortoperation]] and [[#sctn-sample-aborting]] for more information.)
```
emlun, line 2212: I'm not sure what writing style guides say about whether the period should be inside or outside the parentheses (I personally prefer outside), but I think this text is better anyway if we just sidestep that issue altogether:

```suggestion
    ::  The ceremony was cancelled by an {{AbortController}}.
        See [[#sctn-abortoperation]] and [[#sctn-sample-aborting]].
```

index.bs
```diff
@@ -2704,6 +2743,29 @@ The steps for [=issuing a credential request to an authenticator=] are as follow
...
     1. Return [TRUE].
 
+#### Authentication API Exceptions
+
+[=[WRPS]=] can encounter the following {{DOMException|DOMExceptions}} from a call to {{CredentialsContainer/get()|navigator.credentials.get()}}:
+
+<dl dfn-type="argument" dfn-for="Authentication API Exceptions">
+    :   <dfn>AbortError</dfn>
+    ::  The ceremony was cancelled via an {{AbortController}}
+        (see [[#sctn-abortoperation]] and [[#sctn-sample-aborting]] for more information.)
```
emlun, line 2753: Ditto here:
```suggestion
    ::  The ceremony was cancelled by an {{AbortController}}.
        See [[#sctn-abortoperation]] and [[#sctn-sample-aborting]].
```

index.bs
```diff
@@ -2202,6 +2202,45 @@ During the above process, the user agent SHOULD show some UI to the user to guid
...
+[=[WRPS]=] can encounter the following {{DOMException|DOMExceptions}} from a call to {{CredentialsContainer/create()|navigator.credentials.create()}}. Some errors can have multiple reasons for why they happened, requiring the [=[WRPS]=] to infer the actual reason based on their use of WebAuthn:
+
+<dl dfn-type="argument" dfn-for="Registration API Exceptions">
+    :   <dfn>AbortError</dfn>
+    ::  The ceremony was cancelled via an {{AbortController}}
+        (see [[#sctn-abortoperation]] and [[#sctn-sample-aborting]] for more information.)
+
+    :   <dfn>ConstraintError</dfn>
+    ::  Either {{residentKey}} was set to "{{ResidentKeyRequirement/required}}" and no available authenticator supported resident keys,
+        or {{userVerification}} was set to "{{UserVerificationRequirement/required}}" and no available authenticator could perform [=user verification=].
```
emlun, line 2216: We typically don't wrap references to enum values with quotes (my reasoning for this is that the quotes are part of the value definition, but a reference only needs to reference the name definition).

```suggestion
    ::  Either {{residentKey}} was set to {{ResidentKeyRequirement/required}} and no available authenticator supported resident keys,
        or {{userVerification}} was set to {{UserVerificationRequirement/required}} and no available authenticator could perform [=user verification=].
```

index.bs
```diff
@@ -2202,6 +2202,45 @@ During the above process, the user agent SHOULD show some UI to the user to guid
...
+
+    :   <dfn>InvalidStateError</dfn>
+    ::  The authenticator used in the ceremony recognized an entry in {{excludeCredentials}}
+        after the user [=user consent|consented=] to registering a credential.
+
+    :   <dfn>NotSupportedError</dfn>
+    ::  No entry in {{PublicKeyCredentialCreationOptions/pubKeyCredParams}} had a `type` property of "{{PublicKeyCredentialType/public-key}}",
```
emlun, line 2223: Ditto here:
```suggestion
    ::  No entry in {{PublicKeyCredentialCreationOptions/pubKeyCredParams}} had a `type` property of {{PublicKeyCredentialType/public-key}},
```

emlun, line 2223: Reference {{PublicKeyCredentialDescriptor/type}} here?

```suggestion
    ::  No entry in {{PublicKeyCredentialCreationOptions/pubKeyCredParams}} had a {{PublicKeyCredentialDescriptor/type}} property of "{{PublicKeyCredentialType/public-key}}",
```

index.bs
```diff
@@ -2202,6 +2202,45 @@ During the above process, the user agent SHOULD show some UI to the user to guid
...
+        or the [=authenticator=] did not support any of the specified cryptographic parameters in {{PublicKeyCredentialCreationOptions/pubKeyCredParams}}.
+
+    :   <dfn>SecurityError</dfn>
+    ::  The [=effective domain=] was not a [=valid domain=],
+        or {{PublicKeyCredentialCreationOptions/rp}}.{{PublicKeyCredentialRpEntity/id}} was not a registrable domain suffix of nor was equal to the [=effective domain=].
```
emlun, line 2228: We typically wrap multi-piece "code" segments with `<code>` so that the constant-width font also applies to the `.` or other operators.
```suggestion
        or <code>{{PublicKeyCredentialCreationOptions/rp}}.{{PublicKeyCredentialRpEntity/id}}</code> was not a registrable domain suffix of nor was equal to the [=effective domain=].
```

index.bs
```diff
@@ -2704,6 +2743,29 @@ The steps for [=issuing a credential request to an authenticator=] are as follow
...
+[=[WRPS]=] can encounter the following {{DOMException|DOMExceptions}} from a call to {{CredentialsContainer/get()|navigator.credentials.get()}}:
+
+<dl dfn-type="argument" dfn-for="Authentication API Exceptions">
+    :   <dfn>AbortError</dfn>
+    ::  The ceremony was cancelled via an {{AbortController}}
+        (see [[#sctn-abortoperation]] and [[#sctn-sample-aborting]] for more information.)
+
+    :   <dfn>SecurityError</dfn>
+    ::  The [=effective domain=] was not a [=valid domain=],
+        or {{PublicKeyCredentialCreationOptions/rp}}.{{PublicKeyCredentialRpEntity/id}} was not a registrable domain suffix of nor was equal to the [=effective domain=].
```
emlun, line 2757: Ditto here:
```suggestion
        or <code>{{PublicKeyCredentialCreationOptions/rp}}.{{PublicKeyCredentialRpEntity/id}}</code> was not a registrable domain suffix of nor was equal to the [=effective domain=].
```

index.bs
```diff
@@ -2202,6 +2202,45 @@ During the above process, the user agent SHOULD show some UI to the user to guid
 authorizing an authenticator.
 </div>
 
+#### Registration API Exceptions
+
+[=[WRPS]=] can encounter the following {{DOMException|DOMExceptions}} from a call to {{CredentialsContainer/create()|navigator.credentials.create()}}. Some errors can have multiple reasons for why they happened, requiring the [=[WRPS]=] to infer the actual reason based on their use of WebAuthn:
+
+<dl dfn-type="argument" dfn-for="Registration API Exceptions">
+    :   <dfn>AbortError</dfn>
```
emlun, line 2210: These `<dfn>`s are making existing references ambiguous, see Bikeshed output below.

Do we need to refer to these locations in the document, or could we use references instead of definitions here?

(This also applies to the authentication errors section)

```suggestion
<dl>
    :   {{AbortError}}
```

```
LINK ERROR: Ambiguous for-less link for 'NotAllowedError', please see <https://speced.github.io/bikeshed/#ambi-for> for instructions:
Local references:
spec:webauthn-3; type:argument; for:Registration API Exceptions; text:NotAllowedError
spec:webauthn-3; type:argument; for:Authentication API Exceptions; text:NotAllowedError
for-less references:
spec:webidl; type:exception; for:/; text:NotAllowedError
{{NotAllowedError}}
LINK ERROR: Ambiguous for-less link for 'TypeError', please see <https://speced.github.io/bikeshed/#ambi-for> for instructions:
Local references:
spec:webauthn-3; type:argument; for:Registration API Exceptions; text:TypeError
for-less references:
spec:webidl; type:exception; for:/; text:TypeError
{{TypeError}}
LINK ERROR: Ambiguous for-less link for 'SecurityError', please see <https://speced.github.io/bikeshed/#ambi-for> for instructions:
Local references:
spec:webauthn-3; type:argument; for:Registration API Exceptions; text:SecurityError
spec:webauthn-3; type:argument; for:Authentication API Exceptions; text:SecurityError
for-less references:
spec:webidl; type:exception; for:/; text:SecurityError
{{SecurityError}}
LINK ERROR: Ambiguous for-less link for 'NotSupportedError', please see <https://speced.github.io/bikeshed/#ambi-for> for instructions:
Local references:
spec:webauthn-3; type:argument; for:Registration API Exceptions; text:NotSupportedError
for-less references:
spec:webidl; type:exception; for:/; text:NotSupportedError
{{NotSupportedError}}
LINK ERROR: Ambiguous for-less link for 'InvalidStateError', please see <https://speced.github.io/bikeshed/#ambi-for> for instructions:
Local references:
spec:webauthn-3; type:argument; for:Registration API Exceptions; text:InvalidStateError
for-less references:
spec:webidl; type:exception; for:/; text:InvalidStateError
{{InvalidStateError}}
LINK ERROR: Multiple possible 'idl' local refs for 'userVerification'.
Randomly chose one of them; other instances might get a different random choice.
{{userVerification}}
LINK ERROR: Multiple possible 'idl' local refs for 'excludeCredentials'.
Randomly chose one of them; other instances might get a different random choice.
{{excludeCredentials}}
LINK ERROR: Ambiguous for-less link for 'UnknownError', please see <https://speced.github.io/bikeshed/#ambi-for> for instructions:
Local references:
spec:webauthn-3; type:argument; for:Registration API Exceptions; text:UnknownError
spec:webauthn-3; type:argument; for:Authentication API Exceptions; text:UnknownError
for-less references:
spec:webidl; type:exception; for:/; text:UnknownError
{{UnknownError}}
LINK ERROR: Ambiguous for-less link for 'ConstraintError', please see <https://speced.github.io/bikeshed/#ambi-for> for instructions:
Local references:
spec:webauthn-3; type:argument; for:Registration API Exceptions; text:ConstraintError
for-less references:
spec:webidl; type:exception; for:/; text:ConstraintError
{{ConstraintError}}
```

index.bs
```diff
@@ -2202,6 +2202,45 @@ During the above process, the user agent SHOULD show some UI to the user to guid
...
+    ::  The ceremony was cancelled via an {{AbortController}}
+        (see [[#sctn-abortoperation]] and [[#sctn-sample-aborting]] for more information.)
+
+    :   <dfn>ConstraintError</dfn>
+    ::  Either {{residentKey}} was set to "{{ResidentKeyRequirement/required}}" and no available authenticator supported resident keys,
+        or {{userVerification}} was set to "{{UserVerificationRequirement/required}}" and no available authenticator could perform [=user verification=].
+
+    :   <dfn>InvalidStateError</dfn>
+    ::  The authenticator used in the ceremony recognized an entry in {{excludeCredentials}}
```
emlun, line 2219: If the above suggestion of removing `dfn-for="Registration API Exceptions"` is accepted, then these `{{userVerification}}` and `{{excludeCredentials}}` references instead become ambiguous between the primary and JSON variants of the dictionary definitions, so we'd have to disambiguate these:

```suggestion
        or {{AuthenticatorSelectionCriteria/userVerification}} was set to "{{UserVerificationRequirement/required}}" and no available authenticator could perform [=user verification=].

    :   <dfn>InvalidStateError</dfn>
    ::  The authenticator used in the ceremony recognized an entry in {{PublicKeyCredentialCreationOptions/excludeCredentials}}
```

index.bs
```diff
@@ -2202,6 +2202,45 @@ During the above process, the user agent SHOULD show some UI to the user to guid
 authorizing an authenticator.
 </div>
 
+#### Registration API Exceptions
```
emlun, line 2205: Bikeshed lint:

```
LINE 2205: The heading 'Registration API Exceptions' needs a manually-specified ID.
```

index.bs
```diff
@@ -2704,6 +2743,29 @@ The steps for [=issuing a credential request to an authenticator=] are as follow
 
     1. Return [TRUE].
 
+#### Authentication API Exceptions
```
emlun, line 2746: Bikeshed lint:

```
LINE 2746: The heading 'Authentication API Exceptions' needs a manually-specified ID.
```

index.bs
emlun, line 2231: Ditto here about `<code>`:
```suggestion
    ::  The value of <code>{{PublicKeyCredentialCreationOptions/user}}.{{PublicKeyCredentialUserEntity/id}}</code> was not between 1 and 64 bytes (inclusive.)
```

emlun: Also a similar comment about punctuation and parentheses: I think this too can sidestep that issue while also making the text a bit clearer. For example:

```suggestion
    ::  The value of {{PublicKeyCredentialCreationOptions/user}}.{{PublicKeyCredentialUserEntity/id}} was empty or was longer than 64 bytes.
```

index.bs
```diff
@@ -2202,6 +2202,45 @@ During the above process, the user agent SHOULD show some UI to the user to guid
...
+    :   <dfn>SecurityError</dfn>
+    ::  The [=effective domain=] was not a [=valid domain=],
+        or {{PublicKeyCredentialCreationOptions/rp}}.{{PublicKeyCredentialRpEntity/id}} was not a registrable domain suffix of nor was equal to the [=effective domain=].
+
+    :   <dfn>TypeError</dfn>
+    ::  The value of {{PublicKeyCredentialCreationOptions/user}}.{{PublicKeyCredentialUserEntity/id}} was not between 1 and 64 bytes (inclusive.)
+
+    :   <dfn>UnknownError</dfn>
+    ::  The [=authenticator=] could not process the supplied options,
+        or encountered an error while creating the new credential object.
```
emlun, line 2235: "Credential object" seems a bit redundant, especially considering the authenticator isn't the party that constructs a `PublicKeyCredential` instance ("object").

```suggestion
        or encountered an error while creating the new credential.
```

emlun (changes requested):
Substance wise this looks good to me :slightly_smiling_face: but this does highlight that we're not at all consistent about actually forwarding some of these "error code equivalent to..." errors from the authenticator through the client layer. But that should be a different issue.

Some editorial issues to take care of, though:

MasterKale:
@emlun I incorporated all but one of your suggested changes; I left a comment on the one about `TypeError`. Can you please take another look when you have a chance?

And @timcappalli and @sbweeden can you also please take a look when you can?

sbweeden:
Whilst I cannot vouch for the completeness or accuracy of the documented errors, I certainly support the idea that they be listed, along with their reasons when known. I do feel however that this effort is more about document encumbent behaviour than defining a set of useful-to-the-RP error conditions that the clients should implement. As a simple example, it would be good for an RP to be able to differentiate a user canceling an operation vs timeout. At the next WG call I'd like to at least raise whether this effort should consider the approach of trying to define what RPs want, rather than just what browsers currently do.

MasterKale:
@sbweeden This PR will pave a path towards introducing more nuanced errors like the ones I detailed here in response to a similar statement you made a month ago:

https://github.com/w3c/webauthn/issues/2062#issuecomment-2093823953

The hope is that we establish a sensible way of centrally capturing errors, and keep it updated as we consider introducing more nuanced errors for the benefit of RPs.

index.bs
emlun: Yeah, I agree we shouldn't reference implementation-specific behaviours in the spec (unless we're specifying that behaviour as canonical and required), but this is rather an implicit consequence of converting JS values to WebIDL values, which is a specified (if convoluted) process. If I'm reading WebIDL right, I think this particular `TypeError` comes from the JS bindings for [Operations](https://webidl.spec.whatwg.org/#js-operations), which reference [creating an operation function](https://webidl.spec.whatwg.org/#dfn-create-operation-function), where step 2.1.5 invokes the [overload resolution algorithm](https://webidl.spec.whatwg.org/#dfn-overload-resolution-algorithm) whose step 12.19 throws this `TypeError` after failing to convert a JS argument value into a WebIDL value.

Anyway, to be clear, I certainly don't think we should enumerate every particular way a type mismatch could cause a `TypeError`, but I think it's worth mentioning that `TypeError` may be thrown not only specifically because of a bad `user.id` length. Maybe something like this?

```
    :   {{TypeError}}
    ::  The argument was not a valid {{CredentialCreationOptions}} value,
        or the value of <code>{{PublicKeyCredentialCreationOptions/user}}.{{PublicKeyCredentialUserEntity/id}}</code> was empty or longer than 64 bytes.
```

MasterKale: Not a bad idea, I'll go ahead and incorporate this.

Hmm, `{{CredentialCreationOptions}}` links to the credman version of the type, how can I link it to our extended version instead? 🤔

emlun: I don't think you can link directly to the `partial dictionary` WebIDL, but I guess you could link to [§5.1.1. CredentialCreationOptions Dictionary Extension](https://w3c.github.io/webauthn/#sctn-credentialcreationoptions-extension) instead?

MasterKale: Sweet, I figured out how to link to that anchor while still having it say "CredentialCreationOptions". This is exactly what I wanted, thanks for the tip.

Screenshot 2024-06-21 at 10 05 11 AM

MasterKale: @emlun This is ready for your review again ✌️

index.bs
```diff
@@ -2227,6 +2227,46 @@ During the above process, the user agent SHOULD show some UI to the user to guid
 authorizing an authenticator. When <code>|options|.{{CredentialCreationOptions/mediation}}</code> is set to {{CredentialMediationRequirement/conditional}}, prominent modal UI should <i>not</i> be shown <i>unless</i> credential creation was previously consented to via means determined by the user agent.
 </div>
 
+#### Registration API Exceptions #### {#sctn-registration-api-exceptions}
```
timcappalli, line 2230: Nit: since this is a subsection of PKC Create, usage of the Create term would be more consistent. "Registration API" almost sounds like its a different API.
```suggestion
#### Create Request Exceptions #### {#sctn-create-request-exceptions}
```

index.bs
```diff
@@ -2729,6 +2769,29 @@ The steps for [=issuing a credential request to an authenticator=] are as follow
 
     1. Return [TRUE].
 
+#### Authentication API Exceptions #### {#sctn-authentication-api-exceptions}
```
timcappalli, line 2772: Same here.

```suggestion
#### Get Request Exceptions #### {#sctn-get-request-exceptions}
```

timcappalli (approved):
These are just nits (I don't feel too strongly either way).

Looks good otherwise.

index.bs
```diff
@@ -2227,6 +2227,46 @@ During the above process, the user agent SHOULD show some UI to the user to guid
 authorizing an authenticator. When <code>|options|.{{CredentialCreationOptions/mediation}}</code> is set to {{CredentialMediationRequirement/conditional}}, prominent modal UI should <i>not</i> be shown <i>unless</i> credential creation was previously consented to via means determined by the user agent.
 </div>
 
+#### Registration API Exceptions #### {#sctn-registration-api-exceptions}
+
+[=[WRPS]=] can encounter the following {{DOMException|DOMExceptions}} from a call to {{CredentialsContainer/create()|navigator.credentials.create()}}. Some errors can have multiple reasons for why they happened, requiring the [=[WRPS]=] to infer the actual reason based on their use of WebAuthn:
```
nsatragno, line 2232: Let's mark this section non-normative.

index.bs
nsatragno, line 2772: ditto

MasterKale:
WAWG Meeting @ 6/26:

- @nsatragno suggested I mark these sections non-normative so the processing steps remain the authority on what errors clients should raise
- I'm going to make @timcappalli's suggested section title changes, it's a bit confusing as-is given parent header titles

index.bs
```diff
@@ -2729,6 +2769,29 @@ The steps for [=issuing a credential request to an authenticator=] are as follow
...
+    :   {{UnknownError}}
+    ::  The [=authenticator=] could not process the supplied options,
+        or encountered an error while generating an [=assertion signature=].
+
+    :   {{NotAllowedError}}
+    ::  A catch-all error covering a wide range of possible reasons,
+        including common ones like the user canceling out of the ceremony.
+        Some of these causes are documented throughout this spec,
+        while others are client-specific.
+</dl>
```
nsatragno, line 2794: Is there no TypeError for get credential if the options object is not valid?

it's strange we don't seem to specify that anywhere, but e.g. a wrong type or missing mandatory parameter results in TypeError.

index.bs
```diff
@@ -5936,7 +5999,7 @@ The attributes above are structured within this certificate as such:
   04 12                                  -- OCTET STRING
     04  10                               -- OCTET STRING
       CD 8C 39 5C 26 ED EE DE            -- AAGUID cd8c395c-26ed-eede-653b-00797d03ca3c
-      65 3B 00 79 7D 03 CA 3C 
+      65 3B 00 79 7D 03 CA 3C
```
nsatragno, line 6002: unintentional change?

index.bs
```diff
@@ -2227,6 +2227,46 @@ During the above process, the user agent SHOULD show some UI to the user to guid
...
+    ::  The authenticator used in the ceremony recognized an entry in {{PublicKeyCredentialCreationOptions/excludeCredentials}}
+        after the user [=user consent|consented=] to registering a credential.
+
+    :   {{NotSupportedError}}
+    ::  No entry in {{PublicKeyCredentialCreationOptions/pubKeyCredParams}} had a {{PublicKeyCredentialDescriptor/type}} property of {{PublicKeyCredentialType/public-key}},
+        or the [=authenticator=] did not support any of the specified cryptographic parameters in {{PublicKeyCredentialCreationOptions/pubKeyCredParams}}.
+
+    :   {{SecurityError}}
+    ::  The [=effective domain=] was not a [=valid domain=],
+        or <code>{{PublicKeyCredentialCreationOptions/rp}}.{{PublicKeyCredentialRpEntity/id}}</code> was not a registrable domain suffix of nor was equal to the [=effective domain=].
```
emlun, line 2253: Minor improvement (IMO) to text flow:
```suggestion
        or <code>{{PublicKeyCredentialCreationOptions/rp}}.{{PublicKeyCredentialRpEntity/id}}</code> was not equal to or a registrable domain suffix of the [=effective domain=].
```

index.bs
```diff
@@ -2227,6 +2227,46 @@ During the above process, the user agent SHOULD show some UI to the user to guid
...
+
+    :   {{UnknownError}}
+    ::  The [=authenticator=] could not process the supplied options,
+        or encountered an error while creating the new credential.
+
+    :   {{NotAllowedError}}
+    ::  A catch-all error covering a wide range of possible reasons,
+        including common ones like the user canceling out of the ceremony.
+        Some of these causes are documented throughout this spec,
+        while others are client-specific.
```
emlun, line 2267: What are the client-specific cases of this? It would (as I'm sure you agree) be best for cross-platform consistency if we could minimize unspecified behaviour.

index.bs
```diff
@@ -2227,6 +2227,46 @@ During the above process, the user agent SHOULD show some UI to the user to guid
...
+
```
emlun, line 2268: Perhaps we should also point out that extensions may also introduce new error cases which are not listed here (because extensions may be defined outside the WebAuthn spec).

index.bs
```diff
@@ -2729,6 +2769,29 @@ The steps for [=issuing a credential request to an authenticator=] are as follow
...
+[=[WRPS]=] can encounter the following {{DOMException|DOMExceptions}} from a call to {{CredentialsContainer/get()|navigator.credentials.get()}}:
+
+<dl>
+    :   {{AbortError}}
+    ::  The ceremony was cancelled by an {{AbortController}}.
+        See [[#sctn-abortoperation]] and [[#sctn-sample-aborting]].
+
+    :   {{SecurityError}}
+    ::  The [=effective domain=] was not a [=valid domain=],
+        or <code>{{PublicKeyCredentialCreationOptions/rp}}.{{PublicKeyCredentialRpEntity/id}}</code> was not a registrable domain suffix of nor was equal to the [=effective domain=].
```
emlun, line 2783: Same minor improvement (IMO) to text flow:
```suggestion
        or <code>{{PublicKeyCredentialCreationOptions/rp}}.{{PublicKeyCredentialRpEntity/id}}</code> was not equal to or a registrable domain suffix of the [=effective domain=].
```

index.bs
```diff
@@ -2729,6 +2769,29 @@ The steps for [=issuing a credential request to an authenticator=] are as follow
...
+
+    :   {{UnknownError}}
+    ::  The [=authenticator=] could not process the supplied options,
+        or encountered an error while generating an [=assertion signature=].
+
+    :   {{NotAllowedError}}
+    ::  A catch-all error covering a wide range of possible reasons,
+        including common ones like the user canceling out of the ceremony.
+        Some of these causes are documented throughout this spec,
+        while others are client-specific.
```
emlun, line 2793: (Ditto about unspecified behaviours and extensions)

index.bs
MasterKale: Addressed in https://github.com/w3c/webauthn/pull/2047/commits/6760c571dc51670d256ddbc37fd4ddf8dda8b8c0 ✌️

index.bs
MasterKale: Addressed in https://github.com/w3c/webauthn/pull/2047/commits/6760c571dc51670d256ddbc37fd4ddf8dda8b8c0 ✌️

index.bs
MasterKale: I've gone ahead and made this change ✌️

index.bs
MasterKale: I've gone ahead and made this change ✌️

index.bs
MasterKale: I've replaced the space for sake of minimizing PR content.

index.bs
MasterKale: One of the wilder incidents of `NotAllowedError` being overloaded are webviews that error out when WebAuthn is invoked because the native app lacks platform-specific permissions to handle the call. The RP sees a `NotAllowedError` rejection with a platform-specific error message. This can also manifest as the immediate rejection of an API call without any meaningful error message.

I doubt we'd be able to define client-specific behavior like this within the spec, but would be happy to be wrong here.

nsatragno (approved):
Looks good to me. Thank you!

index.bs
MasterKale: I tweaked both of these `NotAllowedError` descriptions with text to the following effect:
Screenshot 2024-07-09 at 3 15 00 PM
Is this what you had in mind?

index.bs
MasterKale: Addressing this lack of `TypeError` during auth definitely seems like a fast-follow to this PR.

MasterKale: I created https://github.com/w3c/webauthn/issues/2094 to track this.

index.bs
MasterKale: I was overthinking things, I added `TypeError` to the list of `.get()` exceptions:

Screenshot 2024-07-10 at 12 51 52 PM

MasterKale:
I was overthinking things, I added `TypeError` to the list of `.get()` exceptions:

Screenshot 2024-07-10 at 12 51 52 PM

I'm going to close out #2094 with this.

index.bs
emlun: Hm, yeah, that seems probably out of scope for the WebAuthn spec. Thanks!

index.bs
emlun: I didn't mean `NotAllowedError` specifically, rather that extensions may introduce additional error cases in genera. For example, the [PRF extension](https://w3c.github.io/webauthn/#prf-extension) returns `NotSupportedError` and `SyntaxError` in some circumstances.

index.bs
emlun: Fixed in 6afa223fd4f661ece8e2eab469fb904a6082599f.

index.bs
```diff
@@ -2729,6 +2772,35 @@ The steps for [=issuing a credential request to an authenticator=] are as follow
...
+    :   {{AbortError}}
+    ::  The ceremony was cancelled by an {{AbortController}}.
+        See [[#sctn-abortoperation]] and [[#sctn-sample-aborting]].
+
+    :   {{SecurityError}}
+    ::  The [=effective domain=] was not a [=valid domain=],
+        or <code>{{PublicKeyCredentialCreationOptions/rp}}.{{PublicKeyCredentialRpEntity/id}}</code> was not equal to or a registrable domain suffix of the [=effective domain=].
+
+    :   {{TypeError}}
+    ::  The <code>|options|</code> argument was not a valid [[#sctn-credentialrequestoptions-extension|CredentialRequestOptions]] value.
```
emlun, line 2791: (Ditto about `TypeError`)

emlun (changes requested):

index.bs
MasterKale: Thanks for clarifying, I took another stab at it, moving the mention of the possiblity of unlisted extension exceptions up into the intro:

Screenshot 2024-07-17 at 9 31 55 AM

emlun: Looks good! But now that implies that errors from extensions defined within the spec _will_ be listed here. Should we do that? :slightly_smiling_face:

index.bs
```diff
@@ -2227,6 +2227,49 @@ During the above process, the user agent SHOULD show some UI to the user to guid
...
+    ::  No entry in {{PublicKeyCredentialCreationOptions/pubKeyCredParams}} had a {{PublicKeyCredentialDescriptor/type}} property of {{PublicKeyCredentialType/public-key}},
+        or the [=authenticator=] did not support any of the specified cryptographic parameters in {{PublicKeyCredentialCreationOptions/pubKeyCredParams}}.
+
+    :   {{SecurityError}}
+    ::  The [=effective domain=] was not a [=valid domain=],
+        or <code>{{PublicKeyCredentialCreationOptions/rp}}.{{PublicKeyCredentialRpEntity/id}}</code> was not equal to or a registrable domain suffix of the [=effective domain=].
+
+    :   {{TypeError}}
+    ::  The <code>|options|</code> argument was not a valid [[#sctn-credentialcreationoptions-extension|CredentialCreationOptions]] value,
+        or the value of <code>{{PublicKeyCredentialCreationOptions/user}}.{{PublicKeyCredentialUserEntity/id}}</code> was empty or was longer than 64 bytes.
```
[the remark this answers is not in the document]
MasterKale: I ended up splitting up the exceptions into groups by type. Thoughts?

## DOMExceptions
Screenshot 2024-07-19 at 2 12 46 PM

## Simple exceptions
Screenshot 2024-07-19 at 2 12 51 PM

index.bs
MasterKale: Alright, point taken. I've decided to punt on capturing extension exceptions here as, after reviewing a few extensions, I think it'd make these new sections too noisy. I'd like to keep this to exceptions that can occur from typical calls to `.create()` and `.get()`, which I'm going to assume don't include any use of extensions.

MasterKale: @emlun Here's a preview of what I have so far:

Screenshot 2024-07-23 at 2 50 02 PM

index.bs
```diff
@@ -261,9 +261,10 @@ spec: SP800-800-63r3; urlPrefix: https://pages.nist.gov/800-63-3/sp800-63-3.html
         text: something you have; url: af
         text: something you are; url: af
 
-spec: webidl; urlPrefix: https://heycam.github.io/webidl
+spec: webidl; urlPrefix: https://webidl.spec.whatwg.org
     type: dfn;
         text: get a copy of the bytes held by the buffer source; url: dfn-get-buffer-source-copy
+        text: Exceptions; url: idl-exceptions
```
marcoscaceres, line 267: this shouldn't be needed, fwiw... it should hopefully autolink?

index.bs
```diff
@@ -2251,6 +2252,66 @@ During the above process, the user agent SHOULD show some UI to the user to guid
...
+
+    :   {{NotAllowedError}}
+    ::  A catch-all error covering a wide range of possible reasons,
+        including common ones like the user canceling out of the ceremony.
+        Some of these causes are documented throughout this spec,
+        while others are client-specific.
+
+</dl>
+
+The following <code>[=Exceptions|Exception=]</code> exceptions can be raised:
```
emlun, line 2306: I would use the **simple exceptions** term here. And I agree with @marcoscaceres, we shouldn't need the new anchor definition in the prelude for this.

```suggestion
The following [=simple exceptions=] can be raised:
```

index.bs
```diff
@@ -2766,6 +2827,52 @@ The steps for [=issuing a credential request to an authenticator=] are as follow
...
+        or encountered an error while generating an [=assertion signature=].
+
+    :   {{NotAllowedError}}
+    ::  A catch-all error covering a wide range of possible reasons,
+        including common ones like the user canceling out of the ceremony.
+        Some of these causes are documented throughout this spec,
+        while others are client-specific.
+</dl>
+
+The following <code>[=Exceptions|Exception=]</code> exceptions can be raised:
```
emlun, line 2868: Ditto:
```suggestion
The following [=simple exceptions=] can be raised:
```

index.bs
```diff
@@ -2251,6 +2252,66 @@ During the above process, the user agent SHOULD show some UI to the user to guid
...
+    ::  The authenticator used in the ceremony recognized an entry in {{PublicKeyCredentialCreationOptions/excludeCredentials}}
+        after the user [=user consent|consented=] to registering a credential.
+
+    :   {{NotSupportedError}}
+    ::  No entry in {{PublicKeyCredentialCreationOptions/pubKeyCredParams}} had a {{PublicKeyCredentialDescriptor/type}} property of {{PublicKeyCredentialType/public-key}},
+        or the [=authenticator=] did not support any of the specified cryptographic parameters in {{PublicKeyCredentialCreationOptions/pubKeyCredParams}}.
+
+    :   {{SecurityError}}
+    ::  The [=effective domain=] was not a [=valid domain=],
+        or <code>{{PublicKeyCredentialCreationOptions/rp}}.{{PublicKeyCredentialRpEntity/id}}</code> was not equal to or a registrable domain suffix of the [=effective domain=].
```
emlun, line 2288: With #2040 merged, I think we need another case here?

index.bs
```diff
@@ -2766,6 +2827,52 @@ The steps for [=issuing a credential request to an authenticator=] are as follow
...
+The following {{DOMException}} exceptions can be raised:
+
+<dl>
+    :   {{AbortError}}
+    ::  The ceremony was cancelled by an {{AbortController}}.
+        See [[#sctn-abortoperation]] and [[#sctn-sample-aborting]].
+
+    :   {{SecurityError}}
+    ::  The [=effective domain=] was not a [=valid domain=],
+        or <code>{{PublicKeyCredentialCreationOptions/rp}}.{{PublicKeyCredentialRpEntity/id}}</code> was not equal to or a registrable domain suffix of the [=effective domain=].
```
emlun, line 2851: With #2040 merged, I think we need another case here?

emlun (changes requested):

index.bs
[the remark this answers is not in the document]
emlun: Looks good!

index.bs
emlun: Looks good!

index.bs
MasterKale: Thanks for this, I took care of it in https://github.com/w3c/webauthn/pull/2047/commits/f15b18b1609d96956016fdefb7457eed944263cb

index.bs
MasterKale: Thanks for this, I took care of it in https://github.com/w3c/webauthn/pull/2047/commits/f15b18b1609d96956016fdefb7457eed944263cb

index.bs
MasterKale: Thanks for pointing this out, I'm clearly still learning the ropes. I've replaced this with a simpler `[=simple exceptions=]` as @emlun called out below.

index.bs
MasterKale: Thank you for the reminder, I've added a bit to both sections about the Related Origins-related `SecurityError` causes:

Screenshot 2024-07-30 at 4 02 56 PM

index.bs
MasterKale: Thank you for the reminder, I've added a bit to both sections about the Related Origins-related `SecurityError` causes:

Screenshot 2024-07-30 at 4 02 56 PM

index.bs
```diff
@@ -2766,6 +2829,55 @@ The steps for [=issuing a credential request to an authenticator=] are as follow
...
+        In the case of [[#sctn-related-origins|related origin requests]],
+        the [=authenticator=] does not support it
+        or there were issues when [[#sctn-validating-relation-origin|validating related origins]].
```
emlun, line 2856: Ditto:
```suggestion
        In the latter case, 
        the [=client=] does not support [[#sctn-related-origins|related origin requests]]
        or the [$related origins validation procedure$] failed.
``````

File: index.bs [hunk 31]
```diff
@@ -2251,6 +2251,69 @@ During the above process, the user agent SHOULD show some UI to the user to guid
...
+
+    :   {{TimeoutError}}
+    ::  The ceremony was cancelled by the user agent after exceeding the time limit permitted for the ceremony.
+        See [[#sctn-timeout-recommended-range]] for more information.
+
+    :   {{UnknownError}}
```
emlun, line 2296: We don't actually throw TimeoutError (yet), that would be introduced in #2095.
```suggestion
    :   {{UnknownError}}
```

index.bs
```diff
@@ -2766,6 +2829,55 @@ The steps for [=issuing a credential request to an authenticator=] are as follow
...
```
emlun, line 2862: We don't actually throw TimeoutError (yet), that would be introduced in https://github.com/w3c/webauthn/pull/2095.
```suggestion
    :   {{UnknownError}}
```

index.bs
```diff
@@ -2251,6 +2251,69 @@ During the above process, the user agent SHOULD show some UI to the user to guid
...
+    ::  The [=authenticator=] could not process the supplied options,
+        or encountered an error while creating the new credential.
```
emlun, line 2298: I'm not sure UnknownError is actually surfaced to the caller - the only occurrences in the spec are that the authenticator should "return an error code equivalent to UnknownError", but these errors are silently ignored by the 'If any authenticator returns an error status not equivalent to "[InvalidStateError](https://webidl.spec.whatwg.org/#invalidstateerror)"' case in `create()` and the "If any authenticator returns an error status" case in `get()`.

index.bs
```diff
@@ -2766,6 +2829,55 @@ The steps for [=issuing a credential request to an authenticator=] are as follow
...
+        the [=authenticator=] does not support it
+        or there were issues when [[#sctn-validating-relation-origin|validating related origins]].
+
+    :   {{TimeoutError}}
+    ::  The ceremony was cancelled by the user agent after exceeding the time limit permitted for the ceremony.
+        See [[#sctn-timeout-recommended-range]] for more information.
+
+    :   {{UnknownError}}
+    ::  The [=authenticator=] could not process the supplied options,
+        or encountered an error while generating an [=assertion signature=].
```
emlun, line 2864: I'm not sure UnknownError is actually surfaced to the caller - the only occurrences in the spec are that the authenticator should "return an error code equivalent to UnknownError", but these errors are silently ignored by the 'If any authenticator returns an error status not equivalent to "[InvalidStateError](https://webidl.spec.whatwg.org/#invalidstateerror)"' case in `create()` and the "If any authenticator returns an error status" case in `get()`.

index.bs
```diff
@@ -2251,6 +2251,69 @@ During the above process, the user agent SHOULD show some UI to the user to guid
...
+        while others are client-specific.
+
+</dl>
+
+The following [=simple exceptions=] can be raised:
+
+<dl>
+
+    :   {{TypeError}}
+    ::  The <code>|options|</code> argument was not a valid [[#sctn-credentialcreationoptions-extension|CredentialCreationOptions]] value,
```
emlun, line 2313: Wrap with `<code>` for visual consistency with `{{}}`-type autolinks?
```suggestion
    ::  The <code>|options|</code> argument was not a valid <code>[[#sctn-credentialcreationoptions-extension|CredentialCreationOptions]]</code> value,
```

index.bs
```diff
@@ -2766,6 +2829,55 @@ The steps for [=issuing a credential request to an authenticator=] are as follow
...
+        Some of these causes are documented throughout this spec,
+        while others are client-specific.
+</dl>
+
+The following [=simple exceptions=] can be raised:
+
+<dl>
+
+    :   {{TypeError}}
+    ::  The <code>|options|</code> argument was not a valid [[#sctn-credentialrequestoptions-extension|CredentialRequestOptions]] value.
```
emlun, line 2878: Wrap with `<code>` for visual consistency with `{{}}`-type autolinks?
```suggestion
    ::  The <code>|options|</code> argument was not a valid <code>[[#sctn-credentialrequestoptions-extension|CredentialRequestOptions]]</code> value.
```

index.bs
MasterKale: Ah, nice insight to pull out of all of this, you're right. I'll remove `UnknownError` then, the RP will only ever get a `NotAllowedError` out of an authenticator raising an `UnknownError` (probably because the user will have to cancel out of the modal which of course becomes a `NotAllowedError`.)

MasterKale: I've removed `UnknownError`

index.bs
MasterKale: As mentioned above, you're right, the RP will get a NotAllowedError from the client after an authenticator raises UnknownError and the user cancels out of the browser modal. I've removed `UnknownError`

index.bs
MasterKale: Whoops, I was juggling branches and accidentally committed this from #2095. I've removed `TimeoutError` from this PR.

index.bs
MasterKale: Whoops, I was juggling branches and accidentally committed this from #2095. I've removed `TimeoutError` from this PR.

index.bs
```diff
@@ -2251,6 +2251,69 @@ During the above process, the user agent SHOULD show some UI to the user to guid
...
+    :   {{NotSupportedError}}
+    ::  No entry in {{PublicKeyCredentialCreationOptions/pubKeyCredParams}} had a {{PublicKeyCredentialDescriptor/type}} property of {{PublicKeyCredentialType/public-key}},
+        or the [=authenticator=] did not support any of the specified cryptographic parameters in {{PublicKeyCredentialCreationOptions/pubKeyCredParams}}.
+
+    :   {{SecurityError}}
+    ::  The [=effective domain=] was not a [=valid domain=],
+        or <code>{{PublicKeyCredentialCreationOptions/rp}}.{{PublicKeyCredentialRpEntity/id}}</code> was not equal to or a registrable domain suffix of the [=effective domain=].
+        In the case of [[#sctn-related-origins|related origin requests]],
+        the [=authenticator=] does not support it
+        or there were issues when [[#sctn-validating-relation-origin|validating related origins]].
```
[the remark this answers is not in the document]
MasterKale: Thanks for the suggestion, I've gone ahead with it.

index.bs
MasterKale: Thanks for the suggestion, I've gone ahead with it.

index.bs
```diff
@@ -2251,6 +2251,69 @@ During the above process, the user agent SHOULD show some UI to the user to guid
...
+    ::  Either {{residentKey}} was set to {{ResidentKeyRequirement/required}} and no available authenticator supported resident keys,
+        or {{AuthenticatorSelectionCriteria/userVerification}} was set to {{UserVerificationRequirement/required}} and no available authenticator could perform [=user verification=].
+
+    :   {{InvalidStateError}}
+    ::  The authenticator used in the ceremony recognized an entry in {{PublicKeyCredentialCreationOptions/excludeCredentials}}
+        after the user [=user consent|consented=] to registering a credential.
+
+    :   {{NotSupportedError}}
+    ::  No entry in {{PublicKeyCredentialCreationOptions/pubKeyCredParams}} had a {{PublicKeyCredentialDescriptor/type}} property of {{PublicKeyCredentialType/public-key}},
+        or the [=authenticator=] did not support any of the specified cryptographic parameters in {{PublicKeyCredentialCreationOptions/pubKeyCredParams}}.
```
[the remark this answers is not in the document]
MasterKale: Done and done ✌️

index.bs
```diff
@@ -261,7 +261,7 @@ spec: SP800-800-63r3; urlPrefix: https://pages.nist.gov/800-63-3/sp800-63-3.html
         text: something you have; url: af
         text: something you are; url: af
 
-spec: webidl; urlPrefix: https://heycam.github.io/webidl
+spec: webidl; urlPrefix: https://webidl.spec.whatwg.org
     type: dfn;
         text: get a copy of the bytes held by the buffer source; url: dfn-get-buffer-source-copy
```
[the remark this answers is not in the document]
MasterKale: Bikeshed seems fine with this locally, so I've removed this spec reference completely.

index.bs
MasterKale: Works for me:

Screenshot 2024-07-31 at 9 55 35 AM

index.bs
MasterKale: Sounds good, done:

Screenshot 2024-07-31 at 9 55 27 AM

index.bs
emlun: Ping @nsatragno to make sure we interpreted this the same way browsers do?

nsatragno: This seems correct: unknown errors should always be translated to `NotAllowedError`. Good catch!

emlun (approved):
Thanks @MasterKale!

sbweeden (approved):

MasterKale:
Heya @pascoej and @marcoscaceres is there a chance you can take a look at this and comment/approve in time for tomorrow's meeting?

pascoej (approved):

selfissued (approved):
This seems useful.

PR #2051, opened 2024-03-27
Include enterpriseAttestation in getClientClientCapabilities

### LINKED ISSUE #1742: Should enterprise attestation support be flagged explicitly?
At the June F2F, the topic of how painful it can be to require enterprise attestation came up. Right now browsers throw a type error if not supported, but that would turn into a request with no attestation after updating browsers to ignore unknown enum values. This is undesirable: RPs might want to know in advance whether enterprise attestation can succeed or not.

timcappalli:
- Adds `enterpriseAttestation` to getClientClientCapabilities enum
- Adds blurb to "enterprise" definition that clients should include it

Resolves #1742

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/pull/2051.html" title="Last updated on Apr 2, 2024, 10:01 PM UTC (51b4b40)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/2051/89f204f...51b4b40.html" title="Last updated on Apr 2, 2024, 10:01 PM UTC (51b4b40)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -3431,6 +3431,8 @@ Note: The {{AttestationConveyancePreference}} enumeration is deliberately not re
     ::  The [=[RP]=] wants to receive an [=attestation statement=] that may include uniquely identifying information. This is intended for controlled deployments within an enterprise where the organization wishes to tie registrations to specific authenticators. User agents MUST NOT provide such an attestation unless the user agent or authenticator configuration permits it for the requested [=RP ID=].
 
         If permitted, the user agent SHOULD signal to the authenticator (at [invocation time](#CreateCred-InvokeAuthnrMakeCred)) that enterprise attestation is requested, and convey the resulting [=/AAGUID=] and [=attestation statement=], unaltered, to the [=[RP]=].
+
+        [=WebAuthn Clients=] supporting enterprise attestation SHOULD include {{ClientCapability/enterpriseAttestation}} in their response to [[#sctn-getClientCapabilities|getClientCapabilities()]].
 </div>
 
 
@@ -3961,6 +3963,7 @@ Note: The {{UserVerificationRequirement}} enumeration is deliberately not refere
     enum ClientCapability {
         "conditionalCreate",
         "conditionalMediation",
+        "enterpriseAttestation",
         "hybridTransport",
         "passkeyPlatformAuthenticator",
         "userVerifyingPlatformAuthenticator",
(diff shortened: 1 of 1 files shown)


timcappalli:
@msft-bob looks like you're not in the repo so I can't assign review to you, but flagging for your feedback via comment

akshayku (approved):

emlun (commented):
This probably needs to take into account not only whether the client _can_ support enterprise attestation, but also whether enterprise attestation is allowed for the calling RP.

There's also the possible complication that it might be allowed by the client, but not by the authenticator if it is configured for [vendor-facilitated enterprise attestation](https://fidoalliance.org/specs/fido-v2.1-ps-20210615/fido-client-to-authenticator-protocol-v2.1-ps-20210615.html#vendor-facilitated-enterprise-attestation) instead of [platform-managed enterprise attestation](https://fidoalliance.org/specs/fido-v2.1-ps-20210615/fido-client-to-authenticator-protocol-v2.1-ps-20210615.html#platform-managed-enterprise-attestation). Do we need to care about that here? @ve7jtb

timcappalli:
> This probably needs to take into account not only whether the client can support enterprise attestation, but also whether enterprise attestation is allowed for the calling RP.

@emlun the intent was to only convey whether the WebAuthn client supported it, not whether it was allowed for a given origin/RP.

emlun:
Ah, ok. In that case it's probably worth mentioning in the `{{ClientCapability/enterpriseAttestation}}` definition that the presence of this capability is no guarantee that the client's/authenticator's policy allows EA for the calling RP.

timcappalli:
> Ah, ok. In that case it's probably worth mentioning in the {{ClientCapability/enterpriseAttestation}} definition that the presence of this capability is no guarantee that the client's/authenticator's policy allows EA for the calling RP.

@emlun addressed in https://github.com/w3c/webauthn/pull/2051/commits/51b4b409f5605f3d1ff6fd77baa4e542ef52e782

nsatragno:
The type error if an attestation type is not supported should not be a reason to merge this. Browsers should not error out on unknown values to begin with, and we should not patch non compliant behaviour with more feature detection.

However, when an RP requires Enterprise Attestation, it probably doesn't make any sense to continue the ceremony when it has no chance to succeed -- so I can see value in this capability. My only concern is there will be a gap until we get the browsers updated where EA will be supported, but this capability will not. Are we risking those RPs assuming EA is not supported when it is during that gap?

MasterKale (approved):

timcappalli:
Closing for the time being.

PR #2058, opened 2024-04-20
Don't zero platform-authenticator AAGUIDs.

agl:
As discussed at the face-to-face, this reflects current practice where the AAGUID of platform authenticators are passed through even when attestation is not requested.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/pull/2058.html" title="Last updated on Apr 20, 2024, 10:56 PM UTC (8f87400)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/2058/3ad0eda...8f87400.html" title="Last updated on Apr 20, 2024, 10:56 PM UTC (8f87400)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2136,9 +2136,9 @@ a numbered step. If outdented, it (today) is rendered as a bullet in the midst o
                             ::  Replace potentially uniquely identifying information with non-identifying versions of the
                                 same:
                                   1. If the [=authData/attestedCredentialData/aaguid=] in the [=attested credential data=] is 16 zero bytes, <code>|credentialCreationData|.[=attestationObjectResult=].fmt</code> is "packed", and "x5c" is absent from <code>|credentialCreationData|.[=attestationObjectResult=]</code>, then [=self attestation=] is being used and no further action is needed.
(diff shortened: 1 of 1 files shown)


pascoej (approved):

MasterKale (approved):

timcappalli:
> I'm sure we've discussed this at some point, but please remind me: what is the issue with the currently specified behaviour of zeroing the AAGUID for _all_ authenticators, including platform authenticators, unless attestation is requested?

The AAGUID is valuable for end user credential names/icons, so many in market deployments are passing an AAGUID even when attestation is not requested. There was consensus in the group that AAGUID should be allowed without attestation. 

At the F2F a few weeks back, there were concerns about only allowing this for platform providers, so the consensus was that there will be 2 PRs: one that just allows the current behavior (this one) and another that allows AAGUIDs from all authenticators.

jschanck:
Is this PR is intended to allow clients to return the AAGUID? Or is it mandating that clients return the AAGUID?

akshayku (approved):

Issue #2024, opened 2024-02-14
Make user.displayName optional?


emlun:
This was brought up in #1942: Since PR #1932, the description of [`PublicKeyCredentialUserEntity.displayName`](https://w3c.github.io/webauthn/#dom-publickeycredentialuserentity-displayname) includes:

>[...] If no suitable or [human-palatable](https://w3c.github.io/webauthn/#human-palatability) name is available, the [Relying Party](https://w3c.github.io/webauthn/#relying-party) SHOULD set this value to an empty string.

The motivation for this being that [setting `displayName` to empty seemed preferable to setting both `name` and `displayName` to the same value](https://github.com/w3c/webauthn/pull/1932#discussion_r1296167484). With this in mind, should [`PublicKeyCredentialUserEntity.displayName`](https://w3c.github.io/webauthn/#dom-publickeycredentialuserentity-displayname) be made optional, defaulting to an empty string?

emlun:
This change may be backwards-incompatible: RPs would still have to set a `displayName` value in case the browser hasn't yet updated to WebAuthn L3 and therefore would return an error instead of assigning `displayName` the new default value.

Issue #1915, opened 2023-07-02
username and display name should not be mandatory (rp, challange either) and OS UX should be simplified if not present


r-jo:
## Description

PublicKeyCredentialUserEntity has mandatory fields by create(): **name** and **displayName** even if there is an anonym user id which is given to the server as "user handle" after a get() and what we can and should use for possibly private-as-possible user identification.

**These fields should not be mandatory** and **OS UX should be much simpler** if these fields are not used. For **simplicity and privacy** it is possible not to use phone numbers, emails or user names. OS should create a ux: Creating passkey for xyz.com web domain (which will be synced to your Apple/Google/Microsoft/Third Party pass manager).

I think there is a username fetish which can be tied to passwords that are going to be replaced with **device based security** (even if private keys are synced and stored, passkey is rather device based if I understand well). With passwords you could log in from everywhere and though you do not really need usernames if the pass is long enough, it was personal and intuitive and now people seem to think it is a must (which is not the case).

We actually **do not need usernames anymore**. The new way is to have a **personal protected OS account** even if the device is common. You protect it with biometrics and PIN. The user is tied to devices (actually OS accounts) that (s)he should own. We can use **access tokens and crypto keys** to create an implicit account and only if the user wants some more robust recovery, we can give a recovery key (a long password without username) or even better as I recon: a passkey which is synced and stored in the pass manager. For normal people it will be Apple and Google pass manager which is ok because those who dont care, save everything via Apple or Gmail anyways. Those who care (gmx, protonmail, 3rd party pass manager etc), can have the option for a third party pass manager as I see (os api).

Again: the user identity is actually the normally 1-3 personal protected os account (and efforts should be made that people create own os account on common devices). The implicit account can be verified via session management and subtle crypto (netflix pushed it and netflix uses it too), actually the same way as by passkeys: private key in browser indexed db non extractable key, public key on server. Of course clearing the browser is equivavalent to device loss, so **at the right point, preferably with passkeys** users can create explicit accounts where the private key is stored in a pass manager.

[...]

PR #2073, opened 2024-05-20
Apply RFC 8264 enforcement only to non-empty strings

### LINKED ISSUE #2068: Empty strings are not valid RFC 8266 Nicknames
[The enforcement rule for the Nickname Profile in RFC 8266](https://www.rfc-editor.org/rfc/rfc8266#section-2.3) expressly forbids empty strings:

> After all of the foregoing rules have been enforced, the entity MUST ensure that the nickname is not zero bytes in length (this is done after enforcing the rules to prevent applications from mistakenly omitting a nickname entirely, because when internationalized strings are accepted a non-empty sequence of characters can result in a zero-length nickname after canonicalization).

This seems to contradict the recommendation for RPs to set [`PublicKeyCredentialUserEntity.displayName`](https://www.w3.org/TR/webauthn-3/#dom-publickeycredentialuserentity-displayname) to the empty string when "no suitable or [human-palatable](https://www.w3.org/TR/webauthn-3/#human-palatability) name is available" while simultaneously recommending RPs and clients enforce the rule:

> [Relying Parties](https://www.w3.org/TR/webauthn-3/#relying-party) SHOULD perform enforcement, as prescribed in Section 2.3 of [[RFC8266]](https://www.w3.org/TR/webauthn-3/#biblio-rfc8266) for the Nickname Profile of the PRECIS FreeformClass [[RFC8264]](https://www.w3.org/TR/webauthn-3/#biblio-rfc8264), when setting [displayName](https://www.w3.org/TR/webauthn-3/#dom-publickeycredentialuserentity-displayname)'s value, or displaying the value to the user.

> [Clients](https://www.w3.org/TR/webauthn-3/#client) SHOULD perform enforcement, as prescribed in Section 2.3 of [[RFC8266]](https://www.w3.org/TR/webauthn-3/#biblio-rfc8266) for the Nickname Profile of the PRECIS FreeformClass [[RFC8264]](https://www.w3.org/TR/webauthn-3/#biblio-rfc8264), on [displayName](https://www.w3.org/TR/webauthn-3/#dom-publickeycredentialuserentity-displayname)'s value prior to displaying the value to the user or including the value as a parameter of the [authenticatorMakeCredential](https://www.w3.org/TR/webauthn-3/#authenticatormakecredential) operation.

Should the spec be changed to state rule enforcement SHOULD only occur when `displayName` is not empty? I was personally bit by this omission in the RP library I am writing.

selfissued:
Fixes #2068 

Cc @zacknewman

***
<a href="https://pr-preview.s3.amazonaws.com/selfissued/webauthn/pull/2073.html" title="Last updated on Jul 17, 2024, 5:25 PM UTC (787585f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/2073/62b069e...selfissued:787585f.html" title="Last updated on Jul 17, 2024, 5:25 PM UTC (787585f)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -3278,8 +3278,8 @@ credential.
     :   <dfn>displayName</dfn>
     ::  A [=human palatability|human-palatable=] name for the [=user account=], intended only for
         display. The [=[RP]=] SHOULD let the user choose this, and SHOULD NOT restrict the choice
-        more than necessary. If no suitable or [=human palatability|human-palatable=] name is
-        available, the [=[RP]=] SHOULD set this value to an empty string.
+        more than necessary.
+        Its value MUST NOT be the empty string.
 
         Examples of suitable values for this identifier include, "Alex Müller", "Alex Müller (ACME Co.)" or "田中倫".
 
--- a/index.bs
+++ b/index.bs
@@ -3278,21 +3278,22 @@ credential.
     :   <dfn>displayName</dfn>
     ::  A [=human palatability|human-palatable=] name for the [=user account=], intended only for
         display. The [=[RP]=] SHOULD let the user choose this, and SHOULD NOT restrict the choice
-        more than necessary.
-        Its value MUST NOT be the empty string.
+        more than necessary. If no suitable or [=human palatability|human-palatable=] name is
+        available, the [=[RP]=] SHOULD set this value to an empty string.
 
         Examples of suitable values for this identifier include, "Alex Müller", "Alex Müller (ACME Co.)" or "田中倫".
 
         - [=[RPS]=] SHOULD perform enforcement, as prescribed in Section 2.3 of
             [[!RFC8266]] for the Nickname Profile of the PRECIS FreeformClass [[!RFC8264]],
-            when setting {{PublicKeyCredentialUserEntity/displayName}}'s value, or displaying the value to the user.
+            when setting {{PublicKeyCredentialUserEntity/displayName}}'s value to a non-empty string,
+            or displaying a non-empty value to the user.
 
         - This string MAY contain language and direction metadata. [=[RPS]=] SHOULD consider providing this information. See [[#sctn-strings-langdir]] about how this metadata is encoded.
 
         - [=Clients=] SHOULD perform enforcement, as prescribed in Section 2.3 of
             [[!RFC8266]] for the Nickname Profile of the PRECIS FreeformClass [[!RFC8264]],
-            on {{PublicKeyCredentialUserEntity/displayName}}'s value prior to displaying the value to the user or
-            including the value as a parameter of the [=authenticatorMakeCredential=] operation.
+            on {{PublicKeyCredentialUserEntity/displayName}}'s value prior to displaying a non-empty value to the user or
+            including a non-empty value as a parameter of the [=authenticatorMakeCredential=] operation.
 
         When [=clients=], [=client platforms=], or [=authenticators=] display a {{PublicKeyCredentialUserEntity/displayName}}'s value, they should always use UI elements to provide a clear boundary around the displayed value, and not allow overflow into other elements [[css-overflow-3]].
 


index.bs
```diff
@@ -3278,8 +3278,8 @@ credential.
     :   <dfn>displayName</dfn>
     ::  A [=human palatability|human-palatable=] name for the [=user account=], intended only for
         display. The [=[RP]=] SHOULD let the user choose this, and SHOULD NOT restrict the choice
-        more than necessary. If no suitable or [=human palatability|human-palatable=] name is
-        available, the [=[RP]=] SHOULD set this value to an empty string.
+        more than necessary.
+        Its value MUST NOT be the empty string.
```
MasterKale, line 3282: Hmm, this looks like it could be "an" empty string instead:

```suggestion
        Its value MUST NOT be an empty string.
```

MasterKale (approved):

MasterKale:
I approved as-is, but I keep coming back to the idea that maybe we keep the existing preamble, but change the "SHOULD set to an empty string" guidance to something like "MUST omit the value" instead 🤔

selfissued:
We could do that.

timcappalli (approved):

emlun:
>maybe we keep the existing preamble, but change the "SHOULD set to an empty string" guidance to something like "MUST omit the value" instead 🤔

That won't work with the dictionary IDL as currently defined, as `displayName` is a required, non-nullable member.

But maybe we could make it optional? I think the only blocker could be compatibility with CTAP2. But it actually looks like CTAP2 would already be compatible with that change. I noticed this in particular in [§6.8.6. Updating user information](https://fidoalliance.org/specs/fido-v2.1-ps-20210615/fido-client-to-authenticator-protocol-v2.1-ps-errata-20220621.html#updateUserInformation):

>Replace the matching credential’s [`PublicKeyCredentialUserEntity`](https://w3c.github.io/webauthn/#dictdef-publickeycredentialuserentity)'s [`name`](https://w3c.github.io/webauthn/#dom-publickeycredentialentity-name), [`displayName`](https://w3c.github.io/webauthn/#dom-publickeycredentialuserentity-displayname) with the passed-in user details. If a field is not present in the passed `user` details, or it is present and empty, remove it from the matching credential’s [`PublicKeyCredentialUserEntity`](https://w3c.github.io/webauthn/#dictdef-publickeycredentialuserentity).

But one problem is that in CTAP2, `PublicKeyCredentialUserEntity` appears in both input (contravariant) and output (covariant) position. Making `displayName` optional is fine in input position, but could break CTAP2 clients if they read out a `PublicKeyCredentialUserEntity` structure for a discoverable credential and expect `displayName` to always be present. But on the other hand there's also this in the definition of the [authenticatorGetAssertion response structure](https://fidoalliance.org/specs/fido-v2.1-ps-20210615/fido-client-to-authenticator-protocol-v2.1-ps-errata-20220621.html#authenticatorgetassertion-response-structure):

>FIDO Devices - [discoverable](https://fidoalliance.org/specs/fido-v2.1-ps-20210615/fido-client-to-authenticator-protocol-v2.1-ps-errata-20220621.html#discoverable) credentials: For discoverable credentials on FIDO devices, at least [`user`](https://w3c.github.io/webauthn/#dictdef-publickeycredentialuserentity) "[`id`](https://w3c.github.io/webauthn/#dom-publickeycredentialuserentity-id)" is mandatory.

So maybe clients are expected to expect that not all member will always be present.

emlun (approved):
Should we spell out the same requirement for [`name`](https://w3c.github.io/webauthn/#dom-publickeycredentialentity-name) too?

selfissued:
@emlun, would you be willing to create a PR making `displayName` and `name` optional in the IDL?  I'll then update this PR to indicate that these fields MUST be omitted if they have no values.  Thanks.

emlun:
> @emlun, would you be willing to create a PR making `displayName` and `name` optional in the IDL?

We can't trivially declare `name` as `optional` in the WebIDL because it's inherited from [`PublicKeyCredentialEntity`](https://w3c.github.io/webauthn/#dictdef-publickeycredentialentity) and represents both `rp.name` and `user.name`. So we can't declare `user.name` as `optional` without doing the same with `rp.name`. Which perhaps we could, but we haven't decided to do that as of now.

Also, I think it's a bad idea to make both `name` and `displayName` optional. This was discussed at length in [issue #1915](https://github.com/w3c/webauthn/issues/1915#issuecomment-1617744129).

> Also, we can't make it `optional` without leaving a trap for sites which will then break with all clients that haven't been updated with that.

Oh right, this is also a very good point. In fact we had already come to this same conclusion in #2024.

> Merely amending the spec to explicitly state that Nickname enforcement SHOULD be done by both RPs and clients for _non-empty strings_ is fine with me.

This sounds like a reasonable compromise to me. The background for why the PRECIS enforcement was added is in https://github.com/w3c/webauthn/issues/593#issuecomment-369402225 and I think most of that probably still applies, so I don't think we should drop it completely.

On the other hand there is https://github.com/w3c/webauthn/issues/2068#issuecomment-2119839162:

> OpenID Connect also forbids empty strings as claim values in an attempt to preempt some interop problems. There, https://openid.net/specs/openid-connect-core-1_0.html#UserInfoResponse says:
> 
> > If a Claim is not returned, that Claim Name SHOULD be omitted from the JSON object representing the Claims; it SHOULD NOT be present with a null or empty string value.
> 
> We should do likewise.

But I think this probably overrules still:

> If the value can't be empty, what should sites who don't have a value set it to?

If it can't be optional or nullable, then I guess empty string is the second best alternative?

Firstyear:
> If it can't be optional or nullable, then I guess empty string is the second best alternative?

We specifically enforce it can't be an empty string in webauthn-rs because it's potentially confusing vs a null/none type. And given how many providers offer OIDC, it seems far more prudent to say "must be a string excluding empty string". 

Then you have to also consider how many resident key UI's would mishandle the empty string too in a device listing. So I think there are a fair number of potential ways for this to go sideways.

sbweeden (dismissed):
I don't like the idea that we define this parameter with "`MUST NOT` be the empty string". If we do that, its implied that the client should raise an error if the RP sets this value to the empty string, which would be a breaking change to RPs that might be doing that now. I think it would be better just to define what the client behaviour should be when the empty string is provided (which should be done if we think its an error anyway).

MasterKale:
I'm leaning towards tweaking guidance on how to process `displayName` so that it can be empty, but **when** it's not empty it should be validated according to RFC 8266.

nsatragno:
During the call, we decided: Let's state that the RFC only applies for non-empty values. This prevents us from breaking backwards compatibility, removes the contradiction, and keeps the value from specifying how the display name should be chosen.

The issue discussed during the call of compatibility issues due to empty display names is orthogonal, and likely an implementation issue. If we tried to fix it by making it optional, we'd risk the much worse compatibility risk of clients that haven't updated breaking. We'll figure out if there are any spec bugs leading to that offline with @MasterKale.

(this discussion was long, and I might have missed something worth summarizing, feel free to chime in!)

@selfissued said he'd be happy to update this PR to match.

emlun (approved):

nsatragno (approved):
Approved, thank you!

agl (approved):

akshayku (approved):

PR #2039, opened 2024-03-06
Update Dependencies section

selfissued:
Fixes #1797

It removes the HTML reference, which seems superfluous.  It also updates the RFC 2119 language to the current recommended text.

***
<a href="https://pr-preview.s3.amazonaws.com/selfissued/webauthn/pull/2039.html" title="Last updated on Mar 6, 2024, 8:42 PM UTC (b8986ac)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/2039/c75b69e...selfissued:b8986ac.html" title="Last updated on Mar 6, 2024, 8:42 PM UTC (b8986ac)">Diff</a>

Issue #1979, opened 2023-09-27
Are notes in webauthn normative or informative?


plehegar:
## Proposed Change

We should consider adding:
[[
As well as sections marked as non-normative, all authoring guidelines, diagrams, examples, and notes in this specification are non-normative. Everything else in this specification is normative.
]]

ensuring that all of the Notes in the document are actually non-normative when doing so.

nadalin:
There are 196 "notes" in the specification some are webauthn specification and some are properly formed NOTES and others are notes within the text, my assumption is that when there is a NOTE it should be a properly formed NOTE. These need to be corrected. I could not find any W3C proper definition of what a NOTE consists of, I suggest that we go with ISO/IEC: According to the ISO/IEC Directives, Part 2, informative elements in a specification are used for providing additional information intended to assist the understanding or use of the document. [They do not contain any requirements, recommendations, or permissions ](https://www.iso.org/files/live/sites/isoorg/files/developing_standards/docs/en/how-to-write-standards.pdf)[1](https://www.iso.org/files/live/sites/isoorg/files/developing_standards/docs/en/how-to-write-standards.pdf).

PR #2093, opened 2024-07-08
Add Signal API

### LINKED ISSUE #1967: [Superset] Updating credential metadata and requesting deletion of stale credentials
This issue is to consolidate a bunch of issues from over the years around deletion of stale credentials, updating properties of existing credentials, etc and provide a single issue to continue discussion. There was general consensus at TPAC 2023 that something like this is needed.

## Existing Issues

All of these issues have now been closed. **Please continue discussion here.**

### #1456: Personal information updates & webauthn
Ability to update things like `name` and `displayName` without creating a new credential.

### #1779: Facility for an RP to indicate a change of displayName to a discoverable credential
Similar to 1456, but specific to displayName

### #1560: Cleanup when creating discoverable credentials with attestations
Essentially allow for clean up of stale credentials. For example, a credential that was deleted RP-side, or a credential that doesn't meet RP requirements.

## High Level Requirements

Trends observed in these issues:

* Desire to signal to a client/authenticator that a specific credential is no longer valid and can be deleted
* Desire to signal to a client/authenticator that `name` and/or `displayName` has changed

### LINKED ISSUE #2038: How to guarantee created resident key is actually received by RP in adverse networking conditions?
I've noticed that under adverse network conditions (very common on mobile phones; especially in rural areas)  it can happen quite often that the `navigator.credentials.create` call succeeds but the credential never gets received by the RP and users end up with many "junk" resident keys in their Apple Keychain because of it. There seems to be no way to avoid this happening. `excludeCredentials` is not an option as the RP isn't even aware that the credential is created. And the user is frustrated because they half of the time select the wrong passkey.

Note this is an issue I have actually observed in the wild. (in this image only one of the two passkeys work. One was created whilst artificially throttling the network in Dev Tools, causing the credential to never be received by the RP)
image

How are RPs supposed to make this robust?   My only solution so far is to drop the whole idea of allowing people to create resident credentials. Instead always opt for using server-side credentials. 

There is also no way for the device that stores the credentials to know which credentials are valid and which not; so it can also not be fixed on that side.

Hypothetically I think perhaps a hybrid flow would be nice where a server-side credential can be "upgraded" to a resident one during the first `get()` call. Anything else seems to be prone to failure.

1.  With the initial `create()` call; the credential is initially stored as a server side credential as credential id `cid1`
2.  The first get() _MUST_ include `allowCredentials: [ cid1 ]`
3. Only after the first `get()` call does the authenticator actually store the credential as resident key. And maybe we can signal in the response from `get()` that the key is now stored as resident
4. subsequent calls to `get()` can be done without `allowCredentials`

_Originally posted by @arianvp in https://github.com/w3c/webauthn/issues/1569#issuecomment-1871971272_

nsatragno:
This PR adds a `PublicKeyCredential.signal*` set of methods that relying parties can call to notify authenticators of changes on the applicability or metadata of credentials.
* `PublicKeyCredential.signalUnknownCredentialId`
  This lets the relying party notify the authenticator that a request with a given credential id would be rejected.
* `PublicKeyCredential.signalAllAcceptedCredentialIds`
  This lets the relying party send a snapshot of all the credential IDs it will accept for a user, allowing the authenticator to hide or remove credentials not present.
* `PublicKeyCredential.signalCurrentUserDetails`
  This lets the relying party update a user's `name` and `displayName`.

Please see the [explainer](https://github.com/w3c/webauthn/wiki/Explainer:-WebAuthn-Signal-API-explainer) for more details.

Closes #1967. Closes #2038.

***
<a href="https://pr-preview.s3.amazonaws.com/nsatragno/webauthn/pull/2093.html" title="Last updated on Aug 14, 2024, 8:26 PM UTC (f08e115)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/2093/2b8e368...nsatragno:f08e115.html" title="Last updated on Aug 14, 2024, 8:26 PM UTC (f08e115)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2953,6 +2953,182 @@ value and terminate the operation.
 
 </div>
 
+### Signal Credential Changes to the Authenticator - PublicKeyCredential's `signal()` Method {#sctn-signal}
+
+<xmp class="idl">
+    partial interface PublicKeyCredential {
+        static Promise<undefined> signal(PublicKeyCredentialSignalOptions options);
+    };
+
+    dictionary PublicKeyCredentialSignalOptions {
+        USVString                              rpId;
+        Base64URLString                        unknownCredentialId;
+        PublicKeyCurrentCredentialDetails      currentCredentials;
+    };
+
+    dictionary PublicKeyCurrentCredentialDetails {
+        required Base64URLString               userId;
+        PublicKeyCurrentCredentialUserDetails  user;
+        sequence<Base64URLString>              allAcceptedCredentialIds;
+    };
+
+    dictionary PublicKeyCurrentCredentialUserDetails {
+        required DOMString                     name;
+        required DOMString                     displayName;
+    };
+</xmp>
+
+[=[WRPS]=] may use {{PublicKeyCredential/signal()}} to signal [=authenticators=]
+the state of [=public key credentials=], so that incorrect or revoked
+credentials may be updated or removed. [=User agents=] provide this
+functionality opportunistically, since an authenticator may not support updating
+its [=credentials map=] or it may not be attached at the time the request is
+made. Furthermore, in order to avoid revealing information about a user's
+credentials without [=user consent=], the {{PublicKeyCredential/signal()}}
+operation does indicate whether the operation succeeded. A successfully resolved
+promise only indicates that the {{PublicKeyCredentialSignalOptions}} object was
+well formed and accepted.
+
+A number of <dfn>credential reports</dfn> can be signalled by [=[WRPS]=]. Each
+[=credential report=] declares a <dfn>validation algorithm</dfn> and <dfn>
+authenticator actions</dfn>. The input for both these algorithms is the
+{{PublicKeyCredentialSignalOptions}} |options|.
+
+#### <dfn>unknownCredentialId</dfn> report #### {#sctn-unknownCredentialIdReport}
+
+Signals that a [=credential id=] was not recognized by the [=[WRP]=], e.g.
+because it was deleted by the user.
+
+    :  [=unknownCredentialId=] [=validation algorithm=]:
+    :: If the result of [=base64url encoding | base64url decoding=]
+         <code>|options|.{{PublicKeyCredentialSignalOptions/unknownCredentialId}}</code>
+         is an error, then throw a {{TypeError}}.
+
+    : [=unknownCredentialId=] [=authenticator actions=]:
+    ::  [=map/For each=] [=public key credential source=] |credential| in the
+        [=authenticator=]'s [=credential map=]:
+        * If the |credential|'s [=public key credential source/rpId=] does
+             not equal
+             <code>|options|.{{PublicKeyCredentialSignalOptions/rpId}}</code>,
+             [=continue=].
+        * If the |credential|'s [=public key credential source/id=] does
+             not equal the result of [=base64url encoding | base64url decoding=]
+             <code>|options|.{{PublicKeyCredentialSignalOptions/unknownCredentialId}}</code>,
+             [=continue=].
+        * [=map/Remove=] |credential| from the [=credentials map=] or hide it
+             from future [=authentication ceremonies=].
+
+<div class="example">
+
+A user deletes a [=credential=] on a [=[WRP]=] provided UI. Later, when trying
+to authenticate to the [=[WRP]=] with an [=list/empty=]
+{{PublicKeyCredentialRequestOptions/allowCredentials}}, the [=authenticator=] UI
+offers them the [=credential=] they previously deleted. If the user selects that
+[=credential=], after rejecting the sign-in attempt, the [=[WRP]=] can run:
+
+```
+PublicKeyCredential.signal({
+    rpId: "example.com",
+    unknownCredentialId: <id of the credential the user just tried, base64url>,
+});
+```
+
+The [=authenticator=] then deletes or hides the [=credential=] from future
+[=authentication ceremonies=].
+
+</div>
+
+#### <dfn>currentCredentials</dfn> report #### {#sctn-currentCredentialsReport}
+
+Signals the complete list of [=credential ids=] for a given user, as well as the
+user's current {{PublicKeyCredentialEntity/name}} and
+{{PublicKeyCredentialUserEntity/displayName}}.
+
+    : [=currentCredentials=] [=validation algorithm=]:
+    :: If the result of [=base64url encoding | base64url decoding=]
+         <code>|options|.{{PublicKeyCredentialSignalOptions/currentCredentials}}.{{PublicKeyCurrentCredentialDetails/userId}}</code>
+         is an error, then throw a {{TypeError}}.
+         If
+         <code>|options|.{{PublicKeyCredentialSignalOptions/currentCredentials}}.{{PublicKeyCurrentCredentialDetails/allAcceptedCredentialIds}}</code>
+         is present:
+         * [=list/For each=] |credentialId| in
+             <code>|options|.{{PublicKeyCredentialSignalOptions/currentCredentials}}.{{PublicKeyCurrentCredentialDetails/allAcceptedCredentialIds}}</code>:
+             * If the result of [=base64url encoding | base64url decoding=]
+                 |credentialId| is an error, then throw a {{TypeError}}.
+
+    : [=currentCredentials=] [=authenticator actions=]:
+    ::  Let |rpId| be |options|.{{PublicKeyCredentialSignalOptions/rpId}}.
+         Let |userId| be result of [=base64url encoding | base64url decoding=]
+         <code>|options|.{{PublicKeyCredentialSignalOptions/currentCredentials}}.{{PublicKeyCurrentCredentialDetails/userId}}</code>.
+         Assert: |userId| is not an error.
+         Let |credential| be <code>[=credentials map=]\[|rpId|, |userId|]</code>.
+         If |credential| does not exist, abort these steps.
+        * Let |userDetails| be
+             <code>|options|.{{PublicKeyCredentialSignalOptions/currentCredentials}}.{{PublicKeyCurrentCredentialDetails/user}}</code>.
+        * If |userDetails| is present, update the |credential|'s
+             [=public key credential source/otherUI=] to match
+             <code>|userDetails|.{{PublicKeyCurrentCredentialUserDetails/name}}</code>
+             and
+             <code>|userDetails|.{{PublicKeyCurrentCredentialUserDetails/displayName}}</code>
+        * Let |allAcceptedCredentialIds| be
+             <code>|options|.{{PublicKeyCredentialSignalOptions/currentCredentials}}.{{PublicKeyCurrentCredentialDetails/allAcceptedCredentialIds}}</code>.
+        * If |allAcceptedCredentialIds| is present and
+             |allAcceptedCredentialIds| does NOT [=list/contain=] the result of
+             [=base64url encoding=] the |credential|'s [=public key credential
+             source/id=], then [=map/remove=] |credential| from the
+             [=credentials map=] or hide it from future [=authentication
+             ceremonies=].
+
+<div class="example">
+
+A user updates their name on a [=[WRP]=] provided UI. The [=[WRP]=] runs:
+
+```
+PublicKeyCredential.signal({
+    rpId: "example.com",
+    currentCredentials: {
+        userId: <user handle, base64url>
+        user: {
+            name: <new user name>,
+            displayName: <new user display name>,
+        }
+    }
+});
+```
+
+The [=authenticator=] then updates the [=public key credential source/otherUI=]
+of the matching credential.
+
+</div>
+
+<div class="example">
+
+A user has two credentials with [=credential ids=] that [=base64url encode=] to
+`aa` and `bb`. The user deletes the credential `aa` on a [=[WRP]=] provided UI.
+The [=[WRP]=] runs:
+
+```
+PublicKeyCredential.signal({
+    rpId: "example.com",
+    currentCredentials: {
+        userId: <user handle, base64url>
+        allAcceptedCredentialIds: [
+            "bb",
+        ]
+    }
+});
+```
+
+If the [=authenticator=] is attached at the time of the report, it deletes or
+hides the [=credential=] corresponding to `aa` from future [=authentication
+ceremonies=].
+
+</div>
+
+Note: [=authenticators=] may not be attached at the time a report is signalled.
+Therefore, [=[WRPS]=] may choose to signal [=currentCredentials=] periodically,
+e.g. on every sign in.
+
 ## Authenticator Responses (interface <dfn interface>AuthenticatorResponse</dfn>) ## {#iface-authenticatorresponse}
 
 [=Authenticators=] respond to [=[RP]=] requests by returning an object derived from the
--- a/index.bs
+++ b/index.bs
@@ -2953,92 +2953,100 @@ value and terminate the operation.
 
 </div>
 
-### Signal Credential Changes to the Authenticator - PublicKeyCredential's `signal()` Method ### {#sctn-signal}
+### Signal Credential Changes to the Authenticator - PublicKeyCredential's [=signal methods=] ### {#sctn-signalMethods}
 
 <xmp class="idl">
     partial interface PublicKeyCredential {
-        static Promise<undefined> signal(optional PublicKeyCredentialSignalOptions options = {});
+        static Promise<undefined> signalUnknownCredentialId(UnknownCredentialIdOptions options);
+        static Promise<undefined> signalAllAcceptedCredentialIds(AllAcceptedCredentialIdsOptions options);
+        static Promise<undefined> signalCurrentUserDetails(CurrentUserDetailsOptions options);
     };
 
-    dictionary PublicKeyCredentialSignalOptions {
-        UnknownCredentialIdReport              unknownCredential;
-        CurrentCredentialsReport               currentCredentials;
-    };
-
-    dictionary UnknownCredentialIdReport {
+    dictionary UnknownCredentialIdOptions {
         required USVString                     rpId;
         required Base64URLString               credentialId;
     };
 
-    dictionary CurrentCredentialsReport {
+    dictionary AllAcceptedCredentialIdsOptions {
         required USVString                     rpId;
         required Base64URLString               userId;
-        CurrentCredentialUserDetails           user;
-        sequence<Base64URLString>              allAcceptedCredentialIds;
+        required sequence<Base64URLString>     allAcceptedCredentialIds;
     };
 
-    dictionary CurrentCredentialUserDetails {
+    dictionary CurrentUserDetailsOptions {
+        required USVString                     rpId;
+        required Base64URLString               userId;
         required DOMString                     name;
         required DOMString                     displayName;
     };
 </xmp>
 
-[=[WRPS]=] may use {{PublicKeyCredential/signal()}} to signal [=authenticators=]
+[=[WRPS]=] may use these <dfn>signal methods</dfn> to signal [=authenticators=]
 the state of [=public key credentials=], so that incorrect or revoked
 credentials may be updated, removed, or hidden. [=Clients=] provide this
 functionality opportunistically, since an authenticator may not support updating
 its [=credentials map=] or it may not be attached at the time the request is
 made. Furthermore, in order to avoid revealing information about a user's
-credentials without [=user consent=], {{PublicKeyCredential/signal()}} does not
-indicate whether the operation succeeded. A successfully resolved promise only
-means that the {{PublicKeyCredentialSignalOptions}} object was well formed.
+credentials without [=user consent=], [=signal methods=] do not indicate whether
+the operation succeeded. A successfully resolved promise only means that the
+<code>options</code> object was well formed.
 
-A number of <dfn>credential reports</dfn> can be signalled by [=[WRPS]=]. Each
-[=credential report=] declares a <dfn for="credential report">validation
-algorithm</dfn> and <dfn for="credential report">authenticator actions</dfn>.
-
-[=Authenticators=] may choose to deviate in their [=credential
-report/authenticator actions=] from the present specification, e.g. to ignore a
+Each [=signal method=] includes <dfn for="signal method">authenticator
+actions</dfn>. [=Authenticators=] may choose to deviate in their [=signal
+method/authenticator actions=] from the present specification, e.g. to ignore a
 change they have a reasonable belief would be contrary to the user's wish, or to
-ask the user before making some change. [=credential report/Authenticator
-actions=] are thus provided as the recommended way to handle [=credential
-reports=].
+ask the user before making some change. [=signal method/Authenticator actions=]
+are thus provided as the recommended way to handle [=signal methods=].
 
 Note: In cases where an [=authenticator=] does not have the capability to
-process [=credential report/authenticator actions=], [=clients=] may choose to
-use existing infrastructure such as [[!FIDO-CTAP]]'s
+process a [=signal method/authenticator action=], [=clients=] may choose to use
+existing infrastructure such as [[!FIDO-CTAP]]'s
 `authenticatorCredentialManagement` command to achieve an equivalent effect.
 
-#### <dfn>unknownCredential</dfn> report #### {#sctn-unknownCredentialIdReport}
+Note: [=signal methods] intentionally do not wait for the [=authenticators=] to
+finish executing the [=signal method/authenticator actions=] to protect users
+from [=[WRPS]=] gaining information about availability of their credentials
+without [=user consent=] from the timing of the request.
+
+#### {{PublicKeyCredential/signalUnknownCredentialId(options)}} #### {#sctn-signalUnknownCredentialId}
 
 Signals that a [=credential id=] was not recognized by the [=[WRP]=], e.g.
 because it was deleted by the user.
 
-The [=unknownCredential=] [=validation algorithm=] takes an
-{{UnknownCredentialIdReport}} |options| and executes these steps:
+Upon execution of {{PublicKeyCredential/signalUnknownCredentialId(options)}},
+the [=client=] executes these steps:
 
 1. Let |effectiveDomain| be the |callerOrigin|'s [=effective domain=]. If
     [=effective domain=] is not a [=valid domain=], then throw a
     "{{SecurityError}}" {{DOMException}}.
-1. If <code>|options|.{{UnknownCredentialIdReport/rpId}}</code> [=is not a
+1. If <code>|options|.{{UnknownCredentialIdOptions/rpId}}</code> [=is not a
     registrable domain suffix of and is not equal to=] |effectiveDomain|, then
     throw a "{{SecurityError}}" {{DOMException}}.
 
     TODO: this may require updating once #2040 lands.
 1. If the result of [=base64url encoding | base64url decoding=]
-    <code>|options|.{{UnknownCredentialIdReport/credentialId}}</code> is an
+    <code>|options|.{{UnknownCredentialIdOptions/credentialId}}</code> is an
     error, then throw a {{TypeError}}.
+
+1. Run the following steps [=in parallel=]:
+        1. For every [=authenticator=] presently available on this [=client
+            platform=], invoke the [=signal method/authenticator
+            action/unknownCredentialId=] [=authenticator action=] with |options|
+            as input.
+
+1. Return [=a promise resolved with=] `undefined`.
     
-The [=unknownCredential=] [=authenticator actions=] take an
-{{UnknownCredentialIdReport}} |options| and are as follows:
+The <dfn for="signal method/authenticator action">unknownCredentialId</dfn>
+[=signal method/authenticator action=] takes an {{UnknownCredentialIdOptions}}
+|options| and is as follows:
 1. [=map/For each=] [=public key credential source=] |credential| in the
     [=authenticator=]'s [=credential map=]:
     1. If the |credential|'s [=public key credential source/rpId=] does not
-        equal <code>|options|.{{UnknownCredentialIdReport/rpId}}</code>,
+        equal <code>|options|.{{UnknownCredentialIdOptions/rpId}}</code>,
         [=continue=].
     1. If the |credential|'s [=public key credential source/id=] does
         not equal the result of [=base64url encoding | base64url decoding=]
-        <code>|options|.{{UnknownCredentialIdReport/credentialId}}</code>,
+        <code>|options|.{{UnknownCredentialIdOptions/credentialId}}</code>,
         [=continue=].
     1. [=map/Remove=] |credential| from the [=credentials map=] or employ an
         [=authenticator=]-specific procedure to hide it from future
@@ -3053,11 +3061,9 @@ offers them the [=credential=] they previously deleted. The user selects that
 [=credential=]. After rejecting the sign-in attempt, the [=[WRP]=] runs:
 
 ```javascript
-PublicKeyCredential.signal({
-    unknownCredential: {
-        rpId: "example.com",
-        credentialId: "aabbcc"  // credential id the user just tried, base64url
-    }
+PublicKeyCredential.signalUnknownCredentialId({
+    rpId: "example.com",
+    credentialId: "aabbcc"  // credential id the user just tried, base64url
 });
 ```
 
@@ -3066,50 +3072,48 @@ The [=authenticator=] then deletes or hides the [=credential=] from future
 
 </div>
 
-#### <dfn>currentCredentials</dfn> report #### {#sctn-currentCredentialsReport}
+#### {{PublicKeyCredential/signalAllAcceptedCredentialIds(options)}} #### {#sctn-signalAllAcceptedCredentialIds}
 
-Signals the complete list of [=credential ids=] for a given user, as well as the
-user's current {{PublicKeyCredentialEntity/name}} and
-{{PublicKeyCredentialUserEntity/displayName}}.
+Signals the complete list of [=credential ids=] for a given user.
 
-The [=currentCredentials=] [=validation algorithm=] takes a
-{{CurrentCredentialsReport}} |options| and executes these steps:
+Upon execution of
+{{PublicKeyCredential/signalAllAcceptedCredentialIds(options)}}, the [=client=]
+executes these steps:
+
 1. Let |effectiveDomain| be the |callerOrigin|'s [=effective domain=].  If
     [=effective domain=] is not a [=valid domain=], then throw a
     "{{SecurityError}}" {{DOMException}}.
-1. If <code>|options|.{{CurrentCredentialsReport/rpId}}</code> [=is not a
+1. If <code>|options|.{{AllAcceptedCredentialIdsOptions/rpId}}</code> [=is not a
     registrable domain suffix of and is not equal to=] |effectiveDomain|, then
     throw a "{{SecurityError}}" {{DOMException}}.
 
     TODO: this may require updating once #2040 lands.
 1. If the result of [=base64url encoding | base64url decoding=]
-    <code>|options|.{{CurrentCredentialsReport/userId}}</code> is an error, then
-    throw a {{TypeError}}.
-1. If
-    <code>|options|.{{CurrentCredentialsReport/allAcceptedCredentialIds}}</code>
-    is present:
-    1. [=list/For each=] |credentialId| in
-        <code>|options|.{{CurrentCredentialsReport/allAcceptedCredentialIds}}</code>:
-       1. If the result of [=base64url encoding | base64url decoding=]
-           |credentialId| is an error, then throw a {{TypeError}}.
+    <code>|options|.{{AllAcceptedCredentialIdsOptions/userId}}</code> is an
+    error, then throw a {{TypeError}}.
+1. [=list/For each=] |credentialId| in
+    <code>|options|.{{AllAcceptedCredentialIdsOptions/allAcceptedCredentialIds}}</code>:
+    1. If the result of [=base64url encoding | base64url decoding=]
+        |credentialId| is an error, then throw a {{TypeError}}.
 
-The [=currentCredentials=] [=credential report/authenticator actions=] take a
-{{CurrentCredentialsReport}} |options| and are as follows:
+1. Run the following steps [=in parallel=]:
+        1. For every [=authenticator=] presently available on this [=client
+            platform=], invoke the [=signal method/authenticator
+            actions/allAcceptedCredentialIds=] [=authenticator action=] with
+            |options| as input.
+
+The <dfn for="signal method/authenticator
+actions">allAcceptedCredentialIds</dfn> [=signal method/authenticator
+actions=] take an {{AllAcceptedCredentialIdsOptions}} |options| and are as
+follows:
 1. Let |userId| be result of [=base64url encoding | base64url decoding=]
-    <code>|options|.{{CurrentCredentialsReport/userId}}</code>.
+    <code>|options|.{{AllAcceptedCredentialIdsOptions/userId}}</code>.
 1. Assertion: |userId| is not an error.
 1. Let |credential| be
-    <code>[=credentials map=][|options|.{{CurrentCredentialsReport/rpId}}, |userId|]</code>.
+    <code>[=credentials map=][|options|.{{AllAcceptedCredentialIdsOptions/rpId}}, |userId|]</code>.
 1. If |credential| does not exist, abort these steps.
-1. If <code>|options|.{{CurrentCredentialsReport/user}}</code> is present,
-    update the |credential|'s [=public key credential source/otherUI=] to match
-    <code>|options|.{{CurrentCredentialsReport/user}}.{{CurrentCredentialUserDetails/name}}</code>
-    and
-    <code>|options|.{{CurrentCredentialsReport/user}}.{{CurrentCredentialUserDetails/displayName}}</code>
 1. If
-    <code>|options|.{{CurrentCredentialsReport/allAcceptedCredentialIds}}</code>
-    is present and
-    <code>|options|.{{CurrentCredentialsReport/allAcceptedCredentialIds}}</code>
+    <code>|options|.{{AllAcceptedCredentialIdsOptions/allAcceptedCredentialIds}}</code>
     does NOT [=list/contain=] the result of [=base64url encoding=] the
     |credential|'s [=public key credential source/id=], then [=map/remove=]
     |credential| from the [=credentials map=] or employ an
@@ -3118,41 +3122,17 @@ The [=currentCredentials=] [=credential report/authenticator actions=] take a
 
 <div class="example">
 
-A user updates their name on a [=[WRP]=] provided UI. The [=[WRP]=] runs:
-
-```javascript
-PublicKeyCredential.signal({
-    currentCredentials: {
-        rpId: "example.com",
-        userId: "aabbcc",  // user handle, base64url.
-        user: {
-            name: "New user name",
-            displayName: "New display name"
-        }
-    }
-});
-```
-
-The [=authenticator=] then updates the [=public key credential source/otherUI=]
-of the matching credential.
-
-</div>
-
-<div class="example">
-
 A user has two credentials with [=credential ids=] that [=base64url encode=] to
 `aa` and `bb`. The user deletes the credential `aa` on a [=[WRP]=] provided UI.
 The [=[WRP]=] runs:
 
 ```javascript
-PublicKeyCredential.signal({
-    currentCredentials: {
-        rpId: "example.com",
-        userId: "aabbcc",  // user handle, base64url.
-        allAcceptedCredentialIds: [
-            "bb",
-        ]
-    }
+PublicKeyCredential.signalAllAcceptedCredentialIds({
+    rpId: "example.com",
+    userId: "aabbcc",  // user handle, base64url.
+    allAcceptedCredentialIds: [
+        "bb",
+    ]
 });
 ```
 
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -2916,6 +2916,214 @@ value and terminate the operation.
...
+#### <dfn>unknownCredentialId</dfn> report #### {#sctn-unknownCredentialIdReport}
+
+Signals that a [=credential id=] was not recognized by the [=[WRP]=], e.g.
+because it was deleted by the user.
+
+The [=unknownCredentialId=] [=validation algorithm=] is as follows:
+1. If the result of [=base64url encoding | base64url decoding=]
+    <code>|options|.{{PublicKeyCredentialSignalOptions/unknownCredentialId}}</code>
+    is an error, then throw a {{TypeError}}.
+
```
MasterKale, line 2969: Do we need anything here, in the name of avoiding abuse of this call (either by RP or malicious JS), to have the client remember the credential ID in the last resolved `.get()` response and only allow the RP to specify that ID in a subsequent `unknownCredentialId` report? The idea being, if an RP receives a response with an `id` of "credA", they shouldn't ever set anything but `"credA"` in their unknown credential Report.

Thinking one step further, I'd say if that idea has merit then what's the point in the RP specifying a credential ID? The RP could more simply signal "I didn't recognize that credential" and the client uses its own knowledge of the last used credential's ID to deprecate that credential. No value validation needs to go on, the RP passes a simpler value for this specific report... 🤔

index.bs
```diff
@@ -2916,6 +2916,214 @@ value and terminate the operation.
...
+    <code>|options|.{{PublicKeyCredentialSignalOptions/unknownCredentialId}}</code>
+    is an error, then throw a {{TypeError}}.
+    
+The [=unknownCredentialId=] [=authenticator actions=] are as follows:
+1. [=map/For each=] [=public key credential source=] |credential| in the
+    [=authenticator=]'s [=credential map=]:
+    1. If the |credential|'s [=public key credential source/rpId=] does
+        not equal
+        <code>|options|.{{PublicKeyCredentialSignalOptions/rpId}}</code>,
+        [=continue=].
```
MasterKale, line 2976: Could we drop this `rpId` for the `unknownCredentialId` report? Does the client need it when it finds the authenticator that recognizes the credential ID so that the authenticator can actually delete the right credential? If this is about dealing with potential collisions in credential IDs for different RP IDs I can see a benefit to keeping `rpId`, but the way I see it if we can embrace [the "probabilistically-unique" property of credential IDs](https://pr-preview.s3.amazonaws.com/nsatragno/webauthn/pull/2093.html#credential-id) then could we make it simpler for RPs to signal this kind of report.

index.bs
```diff
@@ -2916,6 +2916,214 @@ value and terminate the operation.
...
+<xmp class="idl">
+    partial interface PublicKeyCredential {
+        static Promise<undefined> signal(PublicKeyCredentialSignalOptions options);
+    };
+
+    dictionary PublicKeyCredentialSignalOptions {
+        USVString                              rpId;
+        Base64URLString                        unknownCredentialId;
+        PublicKeyCurrentCredentialDetails      currentCredentials;
+    };
```
MasterKale, line 2930: What if we namespaced reports here, so that every type of report is encapsulated in some way? Sure, a report might only have one value. But I imagine it'd go more smoothly for future-us, when we inevitably want to tweak values to existing reports, if we start defining all reports now as dictionaries.

e.g. I might suggest rearranging these two reports like this:

```
const signalOptions = {
  unknownCredential: { rpId: "...", id: "..." },
  currentCredentials: { rpId: "...", userId: "...", ... },
};
```

It's a little bit verbose, but I think it'd be easier for RPs to drill down to the specific report they want and define only that report's values within a dictionary.

index.bs
```diff
@@ -2916,6 +2916,214 @@ value and terminate the operation.
 
 </div>
 
+### Signal Credential Changes to the Authenticator - PublicKeyCredential's `signal()` Method {#sctn-signal}
+
+<xmp class="idl">
+    partial interface PublicKeyCredential {
+        static Promise<undefined> signal(PublicKeyCredentialSignalOptions options);
```
MasterKale, line 2923: Does "`signal()`" leave the door open for a wider range of functionality? I kinda liked "`report()`" (maybe "`sendReport()`" instead though tbh 🤔) as [initially proposed via the explainer](https://github.com/w3c/webauthn/wiki/Explainer:-WebAuthn-Report-API-explainer). But I wonder if the current proposal was born out of subsequent conversations around naming how a "signal" might be more flexible than "report".

MasterKale (commented):
Also thank you very much for applying `Base64URLString` as abundantly as you have, this will save RPs so much time and effort 🙏

nsatragno:
Just to clarify: this isn't quite ready for a review yet! I'll move it from "draft" when ready.

index.bs
nsatragno: We need some form of `rpId` validation, otherwise a relying party can remove passkeys from another relying party that is unaffiliated with it if it knows the (~public) credential id.

index.bs
nsatragno: In terms of abuse, malicious JS can do way more nasty things, such as steal assertions and intercept credential creation requests that we can't meaningfully defend against so I would consider it outside the threat model. I lack the imagination to see what a malicious RP could do with this endpoint as specified, do you have an example?

> The RP could more simply signal "I didn't recognize that credential" and the client uses its own knowledge of the last used credential's ID to deprecate that credential. No value validation needs to go on, the RP passes a simpler value for this specific report... 🤔

I think this is conceptually fine, but significantly more complicated in terms of implementation for no discernible benefit. We'd have to specify how long we remember the credential id for, whether we do it across navigations, and what the remembered credential id is keyed to (the frame? the global browser object?). I would prefer not to go this route unless we find a good reason to avoid RPs from specifying arbitrary credential ids.

index.bs
nsatragno: I like this! Done.

index.bs
nsatragno: Our devrel team asked it to be renamed to avoid confusion with the [Reporting API](https://w3c.github.io/reporting/).

timcappalli: One other name suggestion would be "cleanup"

arnar: I'd leave it more generic (I preferred "report" as well but "signal" seems fine) since e.g. just a metadata update signal is not actually cleanup.

I also think we may want non-cleanup signals in the future, such as "I the RP just accepted this credential" - which providers may find useful to simplify UI with MRU logic, for example.

index.bs
```diff
@@ -2916,6 +2916,252 @@ value and terminate the operation.
...
+        static Promise<undefined> signal(optional PublicKeyCredentialSignalOptions options = {});
+    };
+
+    dictionary PublicKeyCredentialSignalOptions {
+        UnknownCredentialIdReport              unknownCredential;
+        CurrentCredentialsReport               currentCredentials;
+    };
+
+    dictionary UnknownCredentialIdReport {
+        USVString                              rpId;
```
MasterKale, line 2932: I think it'd be easier for an RP dev to understand required vs optional values if we went ahead and explicitly made this required:
```suggestion
        required USVString                              rpId;
```

index.bs
```diff
@@ -2916,6 +2916,252 @@ value and terminate the operation.
...
+        required Base64URLString               credentialId;
+    };
+
+    dictionary CurrentCredentialsReport {
+        USVString                              rpId;
```
MasterKale, line 2937: Same thing here, better to make it explicit than allow for any potential footguns the optionality might introduce:
```suggestion
        required USVString                              rpId;
```

index.bs

index.bs

index.bs
nsatragno: `signal()` it is then. On to bikeshedding other things, marking as resolved.

index.bs
```diff
@@ -2916,6 +2916,246 @@ value and terminate the operation.
...
+        required USVString                     rpId;
+        required Base64URLString               credentialId;
+    };
+
+    dictionary CurrentCredentialsReport {
+        required USVString                     rpId;
+        required Base64URLString               userId;
+        CurrentCredentialUserDetails           user;
+        sequence<Base64URLString>              allAcceptedCredentialIds;
+    };
```
MasterKale, line 2941: It could be argued that this report identifies itself as an **exhaustive** list of **recognized** credential IDs. If we accept that, could we shorten `allAcceptedCredentialIds` to `credentialIds`? Make it look more like `UnknownCredentialIdReport.credentialId` for brevity 🤔 

```suggestion
    dictionary CurrentCredentialsReport {
        required USVString                     rpId;
        required Base64URLString               userId;
        CurrentCredentialUserDetails           user;
        sequence<Base64URLString>              credentialIds;
    };
```

arnar: Are you suggesting that "Current" in "CurrentCredentialsReport" is enough of an indication that the RP must put all credential ids in here, and that it's superfluous to reinforce that with "allAccepted"?

My initial reaction is to keep it as is, because the consequences can be drastic: A provider may, and likely will delete any credential it has that is not included on the list. I.e. the meaning of the field here is very different from the credentialId field in the UnknownCredentialIdReport.

index.bs
```diff
@@ -2916,6 +2916,246 @@ value and terminate the operation.
...
 
+### Signal Credential Changes to the Authenticator - PublicKeyCredential's `signal()` Method ### {#sctn-signal}
+
+<xmp class="idl">
+    partial interface PublicKeyCredential {
+        static Promise<undefined> signal(optional PublicKeyCredentialSignalOptions options = {});
+    };
+
+    dictionary PublicKeyCredentialSignalOptions {
+        UnknownCredentialIdReport              unknownCredential;
```
arnar, line 2927: Nit: I would either name the type UnknownCredentialReport or the field unknownCredentialId

index.bs
```diff
@@ -2916,6 +2916,246 @@ value and terminate the operation.
...
+
+    dictionary UnknownCredentialIdReport {
+        required USVString                     rpId;
+        required Base64URLString               credentialId;
+    };
+
+    dictionary CurrentCredentialsReport {
+        required USVString                     rpId;
+        required Base64URLString               userId;
+        CurrentCredentialUserDetails           user;
```
arnar, line 2939: Also just a nit: Should this simply be a separate report type, since they can be combined, say CurrentUserDetailsReport? The name "CurrentCredentialsReport" is slightly too specific otherwise.

arnar (commented):
My only high level comment is that this structure allows combination of the two defined report types, which there is probably never a reason to do, but not a combination say two CurrentCredentialsReports with different RP-Ids, which there may be a reason to do (e.g. 'www.example.com' and 'example.com').

I don't think that's a major problem though.

index.bs
MasterKale: Hmm, we could go a slightly different route here and rename `CurrentCredentialsReport` to `AllAcceptedCredentialsReport`? Paired with a rename of `allAcceptedCredentialIds` to `credentialIds` it might be more explicit 🤔 

To keep the bikeshedding going, there might also be a possibility of leaving it called `CurrentCredentialsReport` but renaming `allAcceptedCredentialIds` to `allCurrentCredentialIds`.

Having said this I think what I'm trying to do is unify the wording here. I don't particularly care if we go with "current" or "accepted", but I do feel as though there should be one word used for consistency.

index.bs
emlun: I agree this must be restricted to the caller's own credentials. I suppose RP ID _could_ be optional with a default like for `PublicKeyCredentialRequestOptions.rpId`, but it's good practice to set that `rpId` explicitly anyway and the same would apply here, so making it optional is only mildly useful.

index.bs
emlun: > We'd have to specify how long we remember the credential id for, whether we do it across navigations, and what the remembered credential id is keyed to

Maybe it could be tied by exposing the signal as a callback method exposed on the `PublicKeyCredential` instance, but I agree this seems overcomplicated for little benefit.

index.bs
```diff
@@ -2916,6 +2916,246 @@ value and terminate the operation.
...
+1. If |credential| does not exist, abort these steps.
+1. If <code>|options|.{{CurrentCredentialsReport/user}}</code> is present,
+    update the |credential|'s [=public key credential source/otherUI=] to match
+    <code>|options|.{{CurrentCredentialsReport/user}}.{{CurrentCredentialUserDetails/name}}</code>
+    and
+    <code>|options|.{{CurrentCredentialsReport/user}}.{{CurrentCredentialUserDetails/displayName}}</code>
+1. If
+    <code>|options|.{{CurrentCredentialsReport/allAcceptedCredentialIds}}</code>
+    is present and
+    <code>|options|.{{CurrentCredentialsReport/allAcceptedCredentialIds}}</code>
```
emlun, line 3075: I think this line can be omitted. :slightly_smiling_face:

```suggestion
```

index.bs
```diff
@@ -2916,6 +2916,246 @@ value and terminate the operation.
...
+hides the [=credential=] corresponding to `aa` from future [=authentication
+ceremonies=].
+
+</div>
+
+Note: [=Authenticators=] may not be attached at the time a report is signalled.
+Therefore, [=[WRPS]=] may choose to signal [=currentCredentials=] periodically,
+e.g. on every sign in.
+
+#### <code>{{PublicKeyCredential/signal()}}</code> algorithm #### {#sctn-signalAlgorithm}
```
emlun, line 3132: This seems like it should be more prominent, as it is technically the core of the function specification. I think it should at least come before the other two subsections, but probably even be moved into the parent section? (Just before §5.1.10.1. unknownCredential report)

emlun (commented):
Looks pretty good to me!

index.bs
```diff
@@ -2916,6 +2916,246 @@ value and terminate the operation.
 
 </div>
 
+### Signal Credential Changes to the Authenticator - PublicKeyCredential's `signal()` Method ### {#sctn-signal}
```
pascoej, line 2919: I don't necessary have a better suggestion other than the previously attempted `report,` but `signal` as a name is already used in CredentialRequestOptions/CredentialCreationOptions in Credential Management which makes this a little odd. I almost like "Stale Credentials API," but that might be too specific to the current report types.

index.bs
```diff
@@ -2916,6 +2916,246 @@ value and terminate the operation.
...
+
+<xmp class="idl">
+    partial interface PublicKeyCredential {
+        static Promise<undefined> signal(optional PublicKeyCredentialSignalOptions options = {});
```
pascoej, line 2923: Could each report type have a separate method? ex `signalUnknownCredential` and `signalCurrentCredentials.`

This would make the autocomplete easier and explain to the developer what each call does more directly.

I don't think these two types of reports will actually be often sent together. A `UnknownCredentialIdReport` seems to mostly get sent after a non-valid (to the RP) `.get()` while `CurrentCredentialsReport` gets sent after a valid (to the RP) `.get().`

The client could also coalesce reports sent within a small time period together if modifying credentials on an authenticator is a heavy action.

index.bs
nsatragno: > It could also be a separate parameter (credentialIds and deleteOtherCredentials)

It should not be an explicit list of credentials to be deleted because otherwise the relying party has to keep that list ~forever.

> Hmm, we could go a slightly different route here and rename CurrentCredentialsReport to AllAcceptedCredentialsReport?

How about `AllAcceptedCredentialsReport` and keep `allAcceptedCredentialIds`? That way we get consistency and an explicit string. I'll also add a warning that this will remove or hide all credentials not in the list.

index.bs
nsatragno: I agree with your reasoning. I split the PR into three signal* methods, which should hopefully alleviate the confusion with the "signal" attribute and improve autocomplete. Even if a developer wanted to send multiple reports together, nothing is stopping them from calling multiple signal* methods!

index.bs
nsatragno: (renamed methods to signalUnknownCredentialId, etc)

index.bs
nsatragno: This section is gone now that the PR has been split into three methods.

index.bs
nsatragno: This parameter is now required.

index.bs
```diff
@@ -2916,6 +2916,246 @@ value and terminate the operation.
...
+    <code>|options|.{{CurrentCredentialsReport/userId}}</code>.
+1. Assertion: |userId| is not an error.
+1. Let |credential| be
+    <code>[=credentials map=][|options|.{{CurrentCredentialsReport/rpId}}, |userId|]</code>.
+1. If |credential| does not exist, abort these steps.
+1. If <code>|options|.{{CurrentCredentialsReport/user}}</code> is present,
+    update the |credential|'s [=public key credential source/otherUI=] to match
+    <code>|options|.{{CurrentCredentialsReport/user}}.{{CurrentCredentialUserDetails/name}}</code>
+    and
+    <code>|options|.{{CurrentCredentialsReport/user}}.{{CurrentCredentialUserDetails/displayName}}</code>
```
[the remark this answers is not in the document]

index.bs
```diff
@@ -2916,6 +2916,246 @@ value and terminate the operation.
...
+    1. If the |credential|'s [=public key credential source/rpId=] does not
+        equal <code>|options|.{{UnknownCredentialIdReport/rpId}}</code>,
+        [=continue=].
+    1. If the |credential|'s [=public key credential source/id=] does
+        not equal the result of [=base64url encoding | base64url decoding=]
+        <code>|options|.{{UnknownCredentialIdReport/credentialId}}</code>,
+        [=continue=].
+    1. [=map/Remove=] |credential| from the [=credentials map=] or employ an
+        [=authenticator=]-specific procedure to hide it from future
+        [=authentication ceremonies=].
```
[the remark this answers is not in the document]

index.bs
nsatragno: I split this into two reports.

index.bs
nsatragno: Both the method, dictionary, and attribute have the "id" on the name now.

index.bs
nsatragno: I added a note warning of deletion. Splitting the username / displayname updates should help avoid the deletion footgun when a developer is only interested in updating that.

The way to invoke the API now is:

```javascript
PublicKeyCredential.signalAllAcceptedCredentialIds({
    rpId: "example.com",
    userId: "aabbcc",  // user handle, base64url.
    allAcceptedCredentialIds: [
        "bb",
    ]
});
```

I cannot think of something more obvious short of `credentialsNotPresentInTheFollowingListShouldBeDeletedIKnowWhatImDoing`.

nsatragno:
This is ready for another pass.

Firehed (commented):
RP perspective: This looks great overall and should solve some real usability needs in some areas that are a little clumsy today. Looking forward to adding support here!

Thought: is it worth adding anything into `getClientCapabilities()` output, or is it enough to feature-detect on the new methods directly?

index.bs
```diff
@@ -2953,6 +2953,282 @@ value and terminate the operation.
...
+<xmp class="idl">
+    partial interface PublicKeyCredential {
+        static Promise<undefined> signalUnknownCredentialId(UnknownCredentialIdOptions options);
+        static Promise<undefined> signalAllAcceptedCredentialIds(AllAcceptedCredentialIdsOptions options);
+        static Promise<undefined> signalCurrentUserDetails(CurrentUserDetailsOptions options);
+    };
+
+    dictionary UnknownCredentialIdOptions {
+        required DOMString                     rpId;
+        required Base64URLString               credentialId;
```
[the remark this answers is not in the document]
arnar: Ah good catch. If the get request was in a non-discoverable flow, this can indeed happen depending on how the RP is implemented. (It would happen and be a problem on Google.com with our current implementation)

If a credential is rejected after a discoverable flow, should the RP just pass the userId would just come from the assertion? Or do we need this report both with and without userId?

(Semi-related: In initial drafts we had a whole other report that said "this userId is invalid or gone" - implying that all credentials for that userId could be deleted. But that's probably less useful.)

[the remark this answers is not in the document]
nsatragno: > Scenario: A user has multiple accounts on a service, perhaps protected by two different physical authenticators. When authenticating to account `a`, they try to use the authenticator for account `b`

> If the get request was in a non-discoverable flow, this can indeed happen depending on how the RP is implemented. 

If the relying party knows which account is expecting an assertion from, the request should have an `allowedCredentials` list populated with that user's list of credential ids. The user trying to use the wrong authenticator will not result in an assertion being passed to the relying party, so the relying party doesn't have any reason to call `signalUnknownCredentialId`.

This scenario would only be possible if the relying party chose to send an empty allow-list. But why would any relying party risk having the user be shown passkeys for a different account? Non-discoverable credentials are not even required to store the `userId`.

index.bs
```diff
@@ -2916,6 +2916,246 @@ value and terminate the operation.
...
+    update the |credential|'s [=public key credential source/otherUI=] to match
+    <code>|options|.{{CurrentCredentialsReport/user}}.{{CurrentCredentialUserDetails/name}}</code>
+    and
+    <code>|options|.{{CurrentCredentialsReport/user}}.{{CurrentCredentialUserDetails/displayName}}</code>
+1. If
+    <code>|options|.{{CurrentCredentialsReport/allAcceptedCredentialIds}}</code>
+    is present and
+    <code>|options|.{{CurrentCredentialsReport/allAcceptedCredentialIds}}</code>
+    does NOT [=list/contain=] the result of [=base64url encoding=] the
+    |credential|'s [=public key credential source/id=], then [=map/remove=]
```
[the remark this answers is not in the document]
nsatragno: >  might it be worth adding some guidance here around putting the credential into a recoverable state for a period of time before a full removal?

What would you say is the difference between putting a credential in a "recoverable" state and hiding it? The way I see it both have these two requirements:
* Make the credential disappear from the sign-in UI.
* The user must have some way to "recover" ("unhide") the credential, presumably in the settings.

If you develop an authenticator and you can do these things, you should hide the credentials, for recoverability reasons. But as you say, some authenticators will never be able to do that (~any authenticator deployed today that cannot be easily updated). I can add a little more text recommending authenticators hide for recoverability reasons, though it should be clear that's the best option from the other notes in the PR.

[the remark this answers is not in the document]
nsatragno: > Aside: it looks like the list could be populated using the same logic (and, likely, the same actual endpoint) for allowCredentials, including any "fake" credentials added as a privacy-preserving countermeasure 

Unfortunately the "fake" credentials approach is more complex than appears at first glance:
* You have to produce the same list of fakes for the same user every time...
* ...but then it's possible to tell when a user's list is real if they add another passkey.

I would consider it a mitigation more than a solution, and we have the chance to design around not needing that mitigation for the Signal API.

> (it's not explicitly in the procedure about how an authenticator would handle an unknown credential id, but to me that implies ignore it which should be fine). 

It should be stated via the authenticator action algorithm, it's indeed ignored.

> This should make using the API a bit easier!

If the relying party is only interested in empty allow-list requests, then implementing the "fake" credentials endpoint just for this signal is certainly more complex than calling a different function for this case, no?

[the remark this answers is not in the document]
Firehed: > If you develop an authenticator and you can do these things, you should hide the credentials, for recoverability reasons. But as you say, some authenticators will never be able to do that (~any authenticator deployed today that cannot be easily updated). I can add a little more text recommending authenticators hide for recoverability reasons, though it should be clear that's the best option from the other notes in the PR.

That was mostly what I was getting at - not all authenticators will have read _this PR_, but certainly should have read the actual spec. Granted, I expect they all _would_ put them into a recoverable state by default when possible (it seems the obvious choice to make when it's technically possible), but it seems prudent to be explicit in the spec about either the suggestion and/or the implications of not doing this.

index.bs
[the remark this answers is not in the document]
arnar: @nsatragno is right here, I forgot that in a non-discoverable flow an unrecognized credential wouldn't happen without some other kind of error on the RP's behalf, because otherwise that credentialId shouldn't have been on the allow list.

So an unrecognized credential can only happen if the RP doesn't know what account they're expecting. So no need for scoping the credentialId in the report.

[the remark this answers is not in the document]
Firehed: Agreed. While the wording does technically state that allowCredentials SHOULD rather than MUST be populated, I find it hard to think of a reasonable situation that would both be empty and result in an assertion being generated (I _think_ it can happen on some stateless authenticators, but by their nature they'd ignore this signal anyway).

Thanks for the added perspective. Consider this resolved (I don't seem to have the option to do so myself, despite starting the thread)

index.bs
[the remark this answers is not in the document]
nsatragno: Added a note to the effect, right under the note warning RPs not to shoot themselves in the foot.

nsatragno:
> Thought: is it worth adding anything into getClientCapabilities() output, or is it enough to feature-detect on the new methods directly?

Great question actually. We put other things there that can be feature-detected in other ways. I wouldn't be opposed.

index.bs
MasterKale: > How about `AllAcceptedCredentialsReport` and keep `allAcceptedCredentialIds`? That way we get consistency and an explicit string. I'll also add a warning that this will remove or hide all credentials not in the list.

I think this is probably the way to go. An issue with potential names like `RemoveUnlistedCredentialIds` and `DeleteNonListedCredentialIds` could be that different providers do different things. As suggested earlier, maybe a provider wants to tag a not in the list passkey vs delete it outright. "Accepted" vs "Remove" or "Delete" gives providers greater flexibility to handle such a signal without making it confusing to explain to RP's the impact of making such a call.

MasterKale:
@nsatragno I finally caught up on the latest shape of the API. It's almost `isUserVerifyingPlatformAuthenticatorAvailable` again but honestly I find it hard to disagree with the reasoning for it:

- @pascoej made a good point that the two primary reports, `signalAllAcceptedCredentialIds()` and `signalUnknownCredentialId` have diametrically opposite use cases (the former would be sent after a **successful** auth ceremony, while the latter would be sent after **failed** auth ceremony)
- The three discrete methods also addresses @arnar's concern that the old structure _technically_ allowed for multiple reports of the same type to be sent. The new structure definitely reduces edge cases related to this.

I would say I am generally in agreement with the latest shape of the API. I've got some comments I'll leave as Review comments in a few.

index.bs
```diff
@@ -2953,6 +2953,287 @@ value and terminate the operation.
 
 </div>
 
+### Signal Credential Changes to the Authenticator - PublicKeyCredential's [=signal methods=] ### {#sctn-signal-methods}
+
+<xmp class="idl">
+    partial interface PublicKeyCredential {
+        static Promise<undefined> signalUnknownCredentialId(UnknownCredentialIdOptions options);
+        static Promise<undefined> signalAllAcceptedCredentialIds(AllAcceptedCredentialIdsOptions options);
```
MasterKale, line 2961: What do you think about dropping off `Id` from just the method names? It makes the names a _little_ shorter which seems doable because of an implication that an ID will be specified within the options (e.g. "because how else would you identify the credential?").

So `signalAllAcceptedCredentialIds` becomes `signalAllAcceptedCredentials`:

```
await PublicKeyCredential.signalAllAcceptedCredentials({
  rpId: "example.com",
  userId: "M2YPl-KGnA8", // same as user.id at creation time
  allAcceptedCredentalIds: [
     // IDs of all accepted credentials
    "vI0qOggiE3OT01ZRWBYz5l4MEgU0c7PmAA",
    "Bq43BPs"
  ]
});
```

...and `signalUnknownCredentialId` becomes `signalUnknownCredential`:

```
await PublicKeyCredential.signalUnknownCredential({
  rpId: "example.com",
  credentialId: "vI0qOggiE3OT01ZRWBYz5l4MEgU0c7PmAA"  // b64-url
});
```

I think in aggregate it sets a precedent for slightly more human-looking API names especially with `signalCurrentUserDetails` in the mix.

Note that I didn't change any of the names of the options. I'm only proposing alternative names for the methods.

index.bs
```diff
@@ -2953,6 +2953,287 @@ value and terminate the operation.
...
+1. [=map/For each=] [=public key credential source=] |credential| in the
+    [=authenticator=]'s [=credential map=]:
+    1. If the |credential|'s [=public key credential source/rpId=] equals
+        <code>|options|.{{UnknownCredentialIdOptions/rpId}}</code> and the
+        |credential|'s [=public key credential source/id=] equals the result of
+        [=base64url encoding | base64url decoding=]
+        <code>|options|.{{UnknownCredentialIdOptions/credentialId}}</code>,
+        [=map/remove=] |credential| from the [=credentials map=] or employ an
+        [=authenticator=]-specific procedure to hide it from future
+        [=authentication ceremonies=].
```
MasterKale, line 3069: The [explainer mentions](https://github.com/w3c/webauthn/wiki/Explainer:-WebAuthn-Signal-API-explainer#examples) that there are times when one signal might be preferable over another:

> If the user revokes or deletes a credential, e.g. in an account settings UI on the relying party's website, the relying party can opportunistically report this at that time with `signalUnknownCredentialId`. However this will only have effect if the user agent is able to route the report to the same credential provider that created this credential. It may be better to send a `signalAllAcceptedCredentialIds` report instead, with a complete list of valid credential IDs.

However after reviewing the two authenticator actions for these signals, it's unclear to me if the preference for one method over the other (in the use case of a user deleting a credential on the RP's side of things) is one that is reflected in this PR.

Is the main benefit of going with `signalAllAcceptedCredentialIds` after RP-side credential deletion to unhide the credential for some reason later? `signalUnknownCredentialId` doesn't have an affordance for reversing this signal like how the `signalAllAcceptedCredentialIds` authenticator action allows for. Otherwise I can't quite tell how `signalAllAcceptedCredentialIds` might be better for routing the signal to authenticators that may or may not be available when the user deleted the credential.

arnar: > Otherwise I can't quite tell how signalAllAcceptedCredentialIds might be better for routing the signal to authenticators that may or may not be available when the user deleted the credential.

The credential you just deleted might not be in any reachable provider at that time, but a credential you delete perhaps yesterday might be. SignalAllAcceptedCredentialIds would propagate that information. 

I think my recommendation to an RP would be to always use signalAll... if the user is signed in and so it is safe to do so, and only use signalInvalid... when it isn't.

index.bs

index.bs
nsatragno: I added a note stating that relying parties should use `signalAllAcceptableCredentials` instead of `signalUnknownCredential` when the user is authenticated to reflect changes that may not have been picked up by the authenticators that are currently attached.

I think there's enough guidance on both methods for RPs now.

index.bs
```diff
@@ -4036,7 +4322,10 @@ Note: The {{UserVerificationRequirement}} enumeration is deliberately not refere
         "hybridTransport",
         "passkeyPlatformAuthenticator",
         "userVerifyingPlatformAuthenticator",
-        "relatedOrigins"
+        "relatedOrigins",
+        "signalAllAcceptedCredentials",
+        "signalCurrentUserDetails",
+        "signalUnknownCredential"
```
emlun, line 4328: (Continuing the discussion from https://github.com/w3c/webauthn/pull/2093#issuecomment-2253419585 and https://github.com/w3c/webauthn/pull/2093#pullrequestreview-2197576198)

I can see one argument against these: it adds an opportunity for implementation bugs. A client _could_ implement the functions but forget to report (some of) these capabilities, or vice versa. It's probably not very _likely_ that that would happen, but I think RPs can feature-detect the functions perfectly well without these capabilities - checking for the existence of types and functions is fairly normalized among web APIs. So why not eliminate the risk entirely when we can?

nsatragno: I'm good either way. We already have `conditionalGet` and `userVerifyingPlatformAuthenticator` which have their own way to be feature detected as well. I don't think it's terribly important whether we add these or not.

index.bs
```diff
@@ -2953,6 +2953,292 @@ value and terminate the operation.
...
+credentials may be updated, removed, or hidden. [=Clients=] provide this
+functionality opportunistically, since an authenticator may not support updating
+its [=credentials map=] or it may not be attached at the time the request is
+made. Furthermore, in order to avoid revealing information about a user's
+credentials without [=user consent=], [=signal methods=] do not indicate whether
+the operation succeeded. A successfully resolved promise only means that the
+<code>options</code> object was well formed.
+
+Each [=signal method=] includes <dfn for="signal method">authenticator
+actions</dfn>. [=Authenticators=] may choose to deviate in their [=signal
```
emlun, line 2995: This sounds normative to me, so let's make it all-capitals as [clarified in PR #2039](https://github.com/w3c/webauthn/pull/2039/files#diff-5e793325cd2bfc452e268a4aa2f02b4024dd9584bd1db3c2595f61f1ecf7b985R901).

```suggestion
actions</dfn>. [=Authenticators=] MAY choose to deviate in their [=signal
```

index.bs
```diff
@@ -2953,6 +2953,292 @@ value and terminate the operation.
...
+method/authenticator actions=] from the present specification, e.g. to ignore a
+change they have a reasonable belief would be contrary to the user's wish, or to
+ask the user before making some change. [=signal method/Authenticator actions=]
+are thus provided as the recommended way to handle [=signal methods=].
+
+Note: In cases where an [=authenticator=] does not have the capability to
+process a [=signal method/authenticator action=], [=clients=] may choose to use
```
emlun, line 3002:
```suggestion
process a [=signal method/authenticator action=], [=clients=] MAY choose to use
```

index.bs
```diff
@@ -2953,6 +2953,292 @@ value and terminate the operation.
...
+hidden, potentially irreversibly. [=Relying parties=] must exercise care that
+valid credential IDs are never omitted from the list. If a valid [=credential
+ID=] were accidentally omitted, [=relying parties=] SHOULD immediately include
+them in {{PublicKeyCredential/signalAllAcceptedCredentials(options)}} as soon
+as possible to "unhide" them, if supported by the [=authenticator=].
+
+Note: [=Authenticators=] SHOULD prefer hiding [=public key credentials=] instead
+of permanently removing them whenever possible to aid recovery if a [=[WRP]=]
+accidentally omits valid [=credential IDs=] from
+{{AllAcceptedCredentialsOptions/allAcceptedCredentialIds}}.
```
emlun, line 3182: `Note:`s should not be normative (#1979). If these directives should be normative, they should be moved outside of the `Note:`s.

```suggestion
Note: [=Authenticators=] may not be attached at the time
{{PublicKeyCredential/signalAllAcceptedCredentials(options)}} is executed.
Therefore, [=[WRPS]=] may choose to run
{{PublicKeyCredential/signalAllAcceptedCredentials(options)}} periodically,
e.g. on every sign in.

Note: Credentials not present in
{{AllAcceptedCredentialsOptions/allAcceptedCredentialIds}} will be removed or
hidden, potentially irreversibly. [=Relying parties=] must exercise care that
valid credential IDs are never omitted from the list. If a valid [=credential
ID=] were accidentally omitted, [=relying parties=] should immediately include
them in {{PublicKeyCredential/signalAllAcceptedCredentials(options)}} as soon
as possible to "unhide" them, if supported by the [=authenticator=].

Note: [=Authenticators=] should prefer hiding [=public key credentials=] instead
of permanently removing them whenever possible to aid recovery if a [=[WRP]=]
accidentally omits valid [=credential IDs=] from
{{AllAcceptedCredentialsOptions/allAcceptedCredentialIds}}.
```

index.bs
```diff
@@ -9039,6 +9337,11 @@ i.e., if {{PublicKeyCredentialRequestOptions/allowCredentials}} needs to be expo
 the [=[RP]=] could mitigate the privacy leak using the same approach of returning imaginary [=credential IDs=]
 as discussed in [[#sctn-username-enumeration]].
 
+When [=signal methods|signalling=] that a [=credential id=] was not recognized,
+the [=[WRP]=] should use the
```
emlun, line 9341: Should this be normative too?

```suggestion
the [=[WRP]=] SHOULD use the
```

emlun (changes requested):

index.bs
nsatragno: Done (this is no longer a note, too).

index.bs
MasterKale: Let's try and keep `getClientCapabilities()` up-to-date as we continue adding features, it'd be a shame if it falls out of sync so early after getting into L3.

In this PR I'd recommend adding three new capabilities, one for each of the new `signal*` methods.

index.bs
```diff
@@ -2953,6 +2953,292 @@ value and terminate the operation.
...
+from [=[WRPS]=] gaining information about availability of their credentials
+without [=user consent=] from the timing of the request.
+
+#### <dfn abstract-op>Asynchronous RP ID validation algorithm</dfn> #### {#sctn-signal-methods-async-rp-id-validation}
+
+The [$Asynchronous RP ID validation algorithm$] lets [=signal methods=] validate
+[=RP IDs=] [=in parallel=]. The algorithm takes a {{DOMString}} |rpId| as input
+and returns a promise that rejects if the validation fails. The steps are:
+
+1. Let |effectiveDomain| be the |callerOrigin|'s [=effective domain=]. If
```
emlun, line 3017: _callerOrigin_ is not defined here. In `create()` and `get()` it resolves indirectly to "the [relevant settings object](https://html.spec.whatwg.org/multipage/webappapis.html#relevant-settings-object)'s [origin](https://html.spec.whatwg.org/multipage/webappapis.html#concept-settings-object-origin), as determined by the calling [create()](https://w3c.github.io/webappsec-credential-management/#dom-credentialscontainer-create) implementation". I'm not sure if it's appropriate to put that resolution directly into the asynchronous RP ID validation algorithm, or if this needs to be an initial step in each of the signal methods.

index.bs
```diff
@@ -2953,6 +2953,292 @@ value and terminate the operation.
...
+    [=effective domain=] is not a [=valid domain=], then return [=a promise
```
emlun, line 3018: I think?
```suggestion
    |effective domain| is not a [=valid domain=], then return [=a promise
```

index.bs
```diff
@@ -2953,6 +2953,292 @@ value and terminate the operation.
...
+Therefore, [=[WRPS]=] may choose to run
+{{PublicKeyCredential/signalAllAcceptedCredentials(options)}} periodically,
+e.g. on every sign in.
+
+Note: Credentials not present in
+{{AllAcceptedCredentialsOptions/allAcceptedCredentialIds}} will be removed or
+hidden, potentially irreversibly. [=Relying parties=] must exercise care that
+valid credential IDs are never omitted from the list. If a valid [=credential
+ID=] were accidentally omitted, [=relying parties=] should immediately include
+them in {{PublicKeyCredential/signalAllAcceptedCredentials(options)}} as soon
```
emlun, line 3176: This says "a valid credential ID" (singular) and then "include them" and "unhide them" (plural), that should be fixed.

index.bs
```diff
@@ -2953,6 +2953,292 @@ value and terminate the operation.
...
+#### {{PublicKeyCredential/signalUnknownCredential(options)}} #### {#sctn-signalUnknownCredential}
+
+Signals that a [=credential id=] was not recognized by the [=[WRP]=], e.g.
+because it was deleted by the user. Unlike
+{{PublicKeyCredential/signalAllAcceptedCredentials(options)}}, this
+method does not require passing the entire list of accepted [=credential IDs=]
+and the [=userHandle=], avoiding a privacy leak to an unauthenticated caller
+(see [[#sctn-credential-id-privacy-leak]]).
+
+Upon execution of {{PublicKeyCredential/signalUnknownCredential(options)}},
```
emlun, line 3042: Tiny nit: I _think_ "upon invocation" might be slightly more appropriate? Feel free to ignore this if you disagree.

index.bs
nsatragno: I fixed this. I think it's good to have this here: if the domain is not a valid origin, it makes no sense to try to match the rp id to it.

index.bs
nsatragno: Addressed.

index.bs

index.bs
nsatragno: Fixed, thanks.

emlun (approved):
Thanks @nsatragno!

pascoej (approved):

index.bs
MasterKale: Oh, you've already implemented these three new capabilities. Thank you! 🎉

MasterKale (approved):

emlun:
Right now we have signals for updating `user.name` and `user.displayName`, but there is also an [`rp.name`](https://w3c.github.io/webauthn/#dom-publickeycredentialentity-name) property (display name for the RP) whose value might be stored in [otherUI](https://w3c.github.io/webauthn/#public-key-credential-source-otherui). Should there also be a signal for updating that?

index.bs
```diff
@@ -2953,6 +2953,293 @@ value and terminate the operation.
...
+    dictionary CurrentUserDetailsOptions {
+        required DOMString                     rpId;
+        required Base64URLString               userId;
+        required DOMString                     name;
+        required DOMString                     displayName;
+    };
+</xmp>
+
+[=[WRPS]=] may use these <dfn>signal methods</dfn> to signal [=authenticators=]
+the state of [=public key credentials=], so that incorrect or revoked
```
timcappalli, line 2985:
```suggestion
[=[WRPS]=] may use these <dfn>signal methods</dfn> to inform [=authenticators=]
of the state of [=public key credentials=], so that incorrect or revoked
```

index.bs
```diff
@@ -2953,6 +2953,293 @@ value and terminate the operation.
...
+credentials may be updated, removed, or hidden. [=Clients=] provide this
+functionality opportunistically, since an authenticator may not support updating
+its [=credentials map=] or it may not be attached at the time the request is
```
timcappalli, line 2988:
```suggestion
its [=credentials map=] or may not be attached at the time the request is
```

index.bs
```diff
@@ -2953,6 +2953,293 @@ value and terminate the operation.
...
+
+In cases where an [=authenticator=] does not have the capability to process an
+[=signal method/authenticator action=], [=clients=] MAY choose to use existing
+infrastructure such as [[!FIDO-CTAP]]'s `authenticatorCredentialManagement`
+command to achieve an equivalent effect.
+
+Note: [=Signal methods=] intentionally do not wait for the [=authenticators=] to
+finish executing the [=signal method/authenticator actions=] to protect users
+from [=[WRPS]=] gaining information about availability of their credentials
+without [=user consent=] from the timing of the request.
```
timcappalli, line 3009:
```suggestion
Note: [=Signal methods=] intentionally avoid waiting for [=authenticators=] to
complete executing the [=signal method/authenticator actions=]. This measure protects users
from [=[WRPS]=] gaining information about availability of their credentials
without [=user consent=] based on the timing of the request.
```

index.bs
```diff
@@ -2953,6 +2953,293 @@ value and terminate the operation.
...
+1. Return |p|.
+
+#### {{PublicKeyCredential/signalUnknownCredential(options)}} #### {#sctn-signalUnknownCredential}
+
+Signals that a [=credential id=] was not recognized by the [=[WRP]=], e.g.
+because it was deleted by the user. Unlike
+{{PublicKeyCredential/signalAllAcceptedCredentials(options)}}, this
+method does not require passing the entire list of accepted [=credential IDs=]
+and the [=userHandle=], avoiding a privacy leak to an unauthenticated caller
+(see [[#sctn-credential-id-privacy-leak]]).
```
timcappalli, line 3041:
```suggestion
The {{PublicKeyCredential/signalUnknownCredential(options)|signalUnknownCredential}} method signals that a [=credential id=] was not recognized by the [=[WRP]=],
e.g. because it was deleted by the user. Unlike {{PublicKeyCredential/signalAllAcceptedCredentials(options)}}, this
method does not require passing the entire list of accepted [=credential IDs=]
and the [=userHandle=], avoiding a privacy leak to an unauthenticated caller
(see [[#sctn-credential-id-privacy-leak]]).
```

index.bs
```diff
@@ -2953,6 +2953,293 @@ value and terminate the operation.
...
+hidden, potentially irreversibly. [=Relying parties=] must exercise care that
+valid credential IDs are never omitted from the list. If a valid [=credential
+ID=] were accidentally omitted, the [=relying party=] should immediately include
+it in {{PublicKeyCredential/signalAllAcceptedCredentials(options)}} as soon as
+possible to "unhide" it, if supported by the [=authenticator=].
+
+Note: [=Authenticators=] should prefer hiding [=public key credentials=] instead
+of permanently removing them whenever possible to aid recovery if a [=[WRP]=]
+accidentally omits valid [=credential IDs=] from
+{{AllAcceptedCredentialsOptions/allAcceptedCredentialIds}}.
```
timcappalli, line 3183:
```suggestion
Note: [=Authenticators=] should prefer hiding [=public key credentials=] for a period of time instead
of permanently removing them whenever possible to aid recovery if a [=[WRP]=]
accidentally omits valid [=credential IDs=] from
{{AllAcceptedCredentialsOptions/allAcceptedCredentialIds}}.
```

index.bs
```diff
@@ -2953,6 +2953,293 @@ value and terminate the operation.
...
+
+#### {{PublicKeyCredential/signalCurrentUserDetails(options)}} #### {#sctn-signalCurrentUserDetails}
+
+Signals the user's current {{PublicKeyCredentialEntity/name}} and
+{{PublicKeyCredentialUserEntity/displayName}}.
```
timcappalli, line 3188:
```suggestion
The {{PublicKeyCredential/signalCurrentUserDetails(options)|signalCurrentUserDetails}} method signals the user's
current {{PublicKeyCredentialEntity/name}} and {{PublicKeyCredentialUserEntity/displayName}}.
```

timcappalli (approved):
mostly editorial suggestions

nsatragno:
`rp.name` isn't stored or displayed by any authenticator or credential provider that I'm aware of. I would prefer not to spec something for an attribute that is ignored in practice.

I've filed https://github.com/w3c/webauthn/issues/2121 to track whether we want to deprecate `rp.name`, as discussed in the w3c meeting.

@akshayku, please take a look!

Naeempopalzai (approved):

Naeempopalzai (commented):
H

pascoej (approved):

akshayku (approved):

arnar (approved):
Looks great to me!

timcappalli:
> Is this also applicable in reverse will an authenticator / Credentials manager be able to inform of updates to metadata and revocation (deletion) of a credential to the RP?
> 
> Will the RP then update its stored records and or revoke the removed credentials it got informed of by the authenticator / Credentials manager for example in some way confirming the private keys it currently has in possession and available and valid for the rpid?
> 
> Can the RP confirm trustfully it has updated or revoked deleted the public keys affected to the authenticator / credentials mangager?

@maronsen please use passkeys.dev/discuss for these types of questions.

PR #2095, opened 2024-07-11
Alternative error codes

### LINKED ISSUE #2062: Support NoCredentialsError and UserCancelledError codes
## Related

An alternate solution to [https://github.com/w3c/webauthn/issues/1568](https://github.com/w3c/webauthn/issues/1568) / the issues described in [https://github.com/w3c/webauthn/issues/1749](https://github.com/w3c/webauthn/issues/1749).

## To Sum Up

The current paradigm creates a bad UX, because we have no way of knowing if a user has connected their current device to web auth, and so we either need to show them UI for both Create and Get paths, relying on their memory of what they have done, or we need to conditionally only show them the Get path if we "think" the user has connected their current device (based on some information passed to us from our backend), potentially causing the user to attempt web auth and fail if their connection has been destroyed for some reason.

Currently there are 2 existing proposed solutions to this problem out there:

Proposed Solution A is to implement a "Get OR Create" mechanism that handles the details under the hood. This would solve the primary issue but introduces other concerns around state / the desire for frontends to provide a bit of a custom Create path (eg; entering a custom "device name" to associate with the key).

Proposed Solution B is to implement a "does key exist" function that returns a simple boolean depending on whether or not the current device is indeed found in the list of allowed credentials. This would solve the primary issue but introduces possible anonymity / privacy concerns.

I would like to offer a third proposal.

## Proposed Change

Currently, if a user attempts to Get credentials for a device that does not have any, they see some form of automatic "Passkey does not exist" UI, requiring the user to click a Cancel button to back out of the procedure. The error thrown from clicking this Cancel button is a DOMException, with a name of "**NotAllowedError**". This is the same error that is thrown if a user DOES have credentials for their device but still opts to manually cancel out of the procedure.

### Change 1 ###

Throw "NoCredentialsError" if the user cancels out of a credentials.get() flow AND the reason for the cancellation is due to the device not having any credentials.

### Change 2 ###

Throw "UserCancelledError" if the user cancels out of a credentials.get() flow AND the reason for the cancellation is NOT due to the device not having any credentials.

## Benefits

- NotAllowedError can be used to inform the user that something is actually wrong.
- We can detect UserCancelledError and just silently back the user out of the process, instead of worrying about whether or not we need to alert them to some issue. They chose to cancel, we don't need to explain it to them.
- We can detect NoCredentialsError and store a cookie on the device in order to avoid showing the Get flow and instead show the Create flow on next-launch, until the user goes through with the creation. Meaning if they close the website entirely and come back, they won't need to go through this unhappy path again.
- We can automatically push the user through to the Create flow as a result of this error, now that we know they don't have any credentials on this device.
- The Get / Create flows remain separate instead of a possibly over-controlling "Get OR Create" flow that would make it more difficult for custom state-based interactions to take place (eg; entering a custom device name in the case of Create to be passed up to your backend)
- Although this is similar to Solution B ("does key exist") in the sense it exposes some form of anonymous information, the information is obtained as a result of the user attempting to get them, instead of hidden from the user as a preliminary background step.
- I imagine the required code change would be minimal as these are just error codes, perhaps with a bit of state tracking to determine which to throw

## Risks

- This could affect backwards compatibility, assuming people are already looking for the NoCredentialsError explicitly.

### LINKED ISSUE #2096: Return more nuanced errors
These suggestions were pulled out of a comment I made in https://github.com/w3c/webauthn/issues/2062#issuecomment-2093823953. This new issue pulls out the five new errors I'm proposing we add to WebAuthn to help restart conversation with a smaller amount of context to have to catch up on.

## Proposed Change

Earlier this week I met with some colleagues internally and brainstormed some new error messages that would be nice to have as an RP that wants more visibility into why users are having issues signing in, and to potentially offer inline remediation advice:

## `UserCancelledError`

For `UserCancelledError`, it would be great to pull this out of `NotAllowedError` as a new default error that gets returned **after the user specifically cancels out** of the modal experience. `NotAllowedError` is currently so overloaded it's difficult to understand whether the user experienced a legitimate issue with their browser + OS + authenticator, or simply decided to cancel out.

## `HybridPrerequisitesError`

The thinking here is that there are prerequisites that must be met for the browser and/or platform to facilitate successful hybrid registration and auth. If these prerequisites aren't met, and the browser notifies the user of it...

Screenshot 2024-05-03 at 1 54 06 PM

Screenshot 2024-05-03 at 1 55 44 PM

Screenshot 2024-05-03 at 1 54 32 PM

...why not let this propagate through to the RP **after the user cancels out**? In response to this the RP could e.g. show custom UI to guide the user through setting everything up to try again.

## `UserHybridCancelError`

Currently, browsers typically show the hybrid QR code to users when `allowCredentials` contains no locally usable credentials. An RP that could catch this kind of error **after the user cancels out** could better understand that the user cancelled out of the auth because they didn't know how to proceed, and guide the user accordingly.

Right now there's no way to understand this because an RP would receive a `NotAllowedError` in response to the user cancelling the modal experience.

## `UserVerificationError`

Passkeys are typically paired with a requirement to perform UV, both during registration and auth. If **the user clicks through the modal experience** to create a passkey or try to auth, but has no authenticator available that can perform UV or can't enter the correct PIN or scan the correct biometric, can WebAuthn not let the RP know that that's why the user failed the ceremony? The browser definitely tells the user that this is the case...

Screenshot 2024-05-03 at 2 57 04 PM
_U2F token in a `userVerification:required` registration ceremony_

Screenshot 2024-05-03 at 2 56 04 PM
_Security key PIN entry issue during auth ceremony_

...so why not propagate this through to the RP?

I also pondered breaking this one up, into two separate errors:

1. `UserVerificationSupportError` (e.g. U2F token when UV is required)
2. `UserVerificationCollectionError` (e.g. User can't enter correct PIN or scan correct biometric)

...but maybe these are less useful/too complex for a single "`UserVerificationError`" 🤔

## `TimeoutError`

If the user clicks to start a WebAuthn ceremony but gets distracted and walks away to the point that the ceremony times out, couldn't that signal be sent to the RP? It's not quite a true "error" that happened, which it can currently appear to be due to the currently ambiguous `NotAllowedError`, but would be helpful nonetheless in understanding that nothing actually went wrong during the ceremony.

---

Anyway as I mentioned earlier these were new errors that came out of a brainstorming session. I'd love to hear your thoughts, and welcome any suggestions from other RP's on other types of errors that we could try to incorporate into WebAuthn for sake of better understanding of the user's interaction with **WebAuthn.**

MasterKale:
This PR proposes alternative [WHATWG `DOMException` errors](https://webidl.spec.whatwg.org/#idl-DOMException-error-names) to be raised across various WebAuthn interactions. There is an assumption that the user has meaningfully interacted with some part of the ceremony to consent to informing RP's why the ceremony failed.

Fixes #2096.

An explainer is available here: https://github.com/w3c/webauthn/wiki/Explainer:-New-Error-Codes-(2024-Edition)

Note: This PR is targeting #2047 and should not be merged until that PR has been merged.

- [x] OptOutError
- [x] ConstraintError
- [x] TimeoutError

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/pull/2095.html" title="Last updated on Dec 18, 2025, 5:50 PM UTC (afa602b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/2095/b668301...afa602b.html" title="Last updated on Dec 18, 2025, 5:50 PM UTC (afa602b)">Diff</a>


nsatragno (commented):
Some general thoughts:
* I am in favour of adding more specificity for error states after the user has chosen to go ahead with the authentication ceremony. This would not result in a significant impact to the privacy properties of WebAuthn and if it solves a pain point for RPs, we should do it. Essentially, most circumstances where the user indicates their consent to sign in, an error is shown, and then the user taps "Okay" or "Cancel" could result in more specific errors.
* However, it's not clear to me that cancelling out of a ceremony constitutes a form of user consent. This PR will allow RPs to distinguish between "the user's authenticator has no passkeys or the authenticator decided the user denied consent" (NotAllowedError) and "the user cancelled out of the ceremony (likely because they _did_ have passkeys but did not consent)" (UserCancelledError). This is especially tricky because the actual browser implementation differs from the letter of the standard with respect to dispatching requests to authenticators, e.g. Chrome will default to the platform authenticator if it knows there's a credential there.

index.bs
```diff
@@ -1946,11 +1946,17 @@ a numbered step. If outdented, it (today) is rendered as a bullet in the midst o
         <dl class="switch">
             :   If |lifetimeTimer| expires,
             ::  [=set/For each=] |authenticator| in |issuedRequests| invoke the [=authenticatorCancel=] operation on |authenticator|
-                and [=set/remove=] |authenticator| from |issuedRequests|.
+                and [=set/remove=] |authenticator| from |issuedRequests|. Throw a "[=create/TimeoutError=]" {{DOMException}}.
```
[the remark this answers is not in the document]
MasterKale: Not a bad idea, I've created #2096 to pull all of these new errors into a fresh context.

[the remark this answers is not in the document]
MasterKale: To make sure I'm understanding, the final `NotAllowedError` that's currently Step 23 in Create...

Screenshot 2024-07-12 at 3 56 37 PM

...and Step 21 in Get...

Screenshot 2024-07-12 at 3 57 36 PM


...are what need to be replaced by the new `TimeoutError`?

index.bs
```diff
@@ -2254,13 +2264,21 @@ authorizing an authenticator. When <code>|options|.{{CredentialCreationOptions/m
...
     :   {{UnknownError}}
     ::  The [=authenticator=] could not process the supplied options,
         or encountered an error while creating the new credential.
+    
+    :   <dfn for="create">UserCancelledError</dfn>
+    ::  The user has exercised a user agent user-interface option
+        to end the ceremony.
 
     :   {{NotAllowedError}}
     ::  A catch-all error covering a wide range of possible reasons,
```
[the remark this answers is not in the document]
MasterKale: Thanks, addressed in https://github.com/w3c/webauthn/pull/2095/commits/0906982ffc4320eebf21cb75caee87815a5bc434

index.bs
```diff
@@ -2783,16 +2807,28 @@ The steps for [=issuing a credential request to an authenticator=] are as follow
...
     :   {{UnknownError}}
     ::  The [=authenticator=] could not process the supplied options,
         or encountered an error while generating an [=assertion signature=].
+    
+    :   <dfn for="get">UserCancelledError</dfn>
+    ::  The user has exercised a user agent user-interface option
+        to end the ceremony.
 
     :   {{NotAllowedError}}
     ::  A catch-all error covering a wide range of possible reasons,
```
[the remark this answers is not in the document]
MasterKale: Good catch, addressed in https://github.com/w3c/webauthn/pull/2095/commits/0906982ffc4320eebf21cb75caee87815a5bc434

index.bs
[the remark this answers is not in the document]
MasterKale: I created a new explainer here for your consideration:

https://github.com/w3c/webauthn/wiki/Explainer:-New-Error-Codes-(2024-Edition)

emlun (changes requested):
Before getting into specifics: I noted in https://github.com/w3c/webauthn/issues/2096#issuecomment-2233307689 that we can't simply use any value as the [`name` of a `DOMException`](https://webidl.spec.whatwg.org/#dfn-error-names-table):

>When [creating](https://webidl.spec.whatwg.org/#dfn-create-exception) or [throwing](https://webidl.spec.whatwg.org/#dfn-throw) a [DOMException](https://webidl.spec.whatwg.org/#idl-DOMException), specifications must use one of these names. If a specification author believes none of these names are a good fit for their case, they must [file an issue](https://github.com/whatwg/webidl/issues/new?title=DOMException%20name%20proposal) to discuss adding a new name to the shared namespace [...]

So we would need to upstream these new name definitions, or use [`DOMException` derived interfaces](https://webidl.spec.whatwg.org/#idl-DOMException-derived-interfaces) instead.

index.bs
```diff
@@ -3726,6 +3788,38 @@ SHOULD be aborted.
...
+[Exposed=Window, Serializable]
+interface UserHybridCancellationError : DOMException {
+};
+
+[Exposed=Window, Serializable]
+interface UserVerificationError : DOMException {
+};
+</xmp>
+
+
```
MasterKale, line 3822: @emlun I've defined custom interfaces for the new errors so they're derived from `DOMException`. Thoughts?

index.bs
```diff
@@ -3726,6 +3788,38 @@ SHOULD be aborted.
...
+please see [[#sctn-create-request-exceptions]] and [[#sctn-get-request-exceptions]].
+
+<xmp class="idl">
+[Exposed=Window, Serializable]
+interface HybridPrerequisitesError : DOMException {
+};
+
+[Exposed=Window, Serializable]
+interface TimeoutError : DOMException {
+};
```
emlun, line 3807: [Web IDL](https://webidl.spec.whatwg.org/#idl-DOMException-derived-interfaces) says:

>The [identifier](https://webidl.spec.whatwg.org/#dfn-identifier) of the [interface](https://webidl.spec.whatwg.org/#dfn-interface) must end with `Error`, and must not be any of the names in the [`DOMException` names table](https://webidl.spec.whatwg.org/#dfn-error-names-table).

So I don't think we need a separate `TimeoutError` since that's already covered by `DOMException`?

index.bs
```diff
@@ -3726,6 +3788,38 @@ SHOULD be aborted.
...
+
+### Custom WebAuthn Exceptions ### {#iface-custom-webauthn-exceptions}
+
+For descriptions of these exceptions,
+please see [[#sctn-create-request-exceptions]] and [[#sctn-get-request-exceptions]].
+
+<xmp class="idl">
+[Exposed=Window, Serializable]
+interface HybridPrerequisitesError : DOMException {
+};
```
emlun, line 3803: I think all of these [need to have explicit constructor operations and (de)serialization steps](https://webidl.spec.whatwg.org/#idl-DOMException-derived-interfaces):

>- The [interface](https://webidl.spec.whatwg.org/#dfn-interface) must have a [constructor operation](https://webidl.spec.whatwg.org/#idl-constructors) which sets the instance’s [name](https://webidl.spec.whatwg.org/#domexception-name) to the interface’s [identifier](https://webidl.spec.whatwg.org/#dfn-identifier).
>- Their [constructor operation](https://webidl.spec.whatwg.org/#idl-constructors) must take as its first parameter an [optional](https://webidl.spec.whatwg.org/#dfn-optional-argument) [DOMString](https://webidl.spec.whatwg.org/#idl-DOMString) named message defaulting to the empty string, and must set the instance’s [message](https://webidl.spec.whatwg.org/#domexception-message) to message.
- [...]
- They should be [serializable objects](https://html.spec.whatwg.org/multipage/structured-data.html#serializable-objects), whose [serialization steps](https://html.spec.whatwg.org/multipage/structured-data.html#serialization-steps) and [deserialization steps](https://html.spec.whatwg.org/multipage/structured-data.html#deserialization-steps) preserve the additional information.

It doesn't look like those can be inherited. So this would need to be something like this:

```suggestion
interface HybridPrerequisitesError : DOMException {
    constructor(optional DOMString message = "");
};

The <code>new HybridPrerequisitesError(|message|)</code> constructor steps are to set [=this=]'s {{DOMException/name}} to `"HybridPrerequisitesError"`.

{{HybridPrerequisitesError}} objects are [=serializable objects=].

Their [=serialization steps=], given |value| and |serialized|, are to run the {{DOMException}} [=serialization steps=] given |value| and |serialized|.

Their [=deserialization steps=], given |serialized| and |value|, are to run the {{DOMException}} [=deserialization steps=] given |serialized| and |value|.
```

...and repeat that for each derived interface. :face_with_diagonal_mouth:

index.bs
```diff
@@ -2244,7 +2260,7 @@ a numbered step. If outdented, it (today) is rendered as a bullet in the midst o
         </dl>
     </li>
 
-1. Throw a "{{NotAllowedError}}" {{DOMException}}. In order to prevent information leak that could identify the
+1. Throw a "{{TimeoutError}}" {{DOMException}}. In order to prevent information leak that could identify the
```
emlun, line 2263: I'm not entirely sure we can just change this error type. The point of the caveats in this paragraph is to prevent an information leak via distinguishable error conditions, so changing the type could reintroduce that information leak.

This _might_ be safe now as other pieces of the spec have changed since this was introduced - namely that clients are now expected to ask for user consent before returning an `IllegalStateError`, thus "authorizing" the information leak. But this needs some scrutiny before we can merge the change (and possibly even drop the caveat about the information leak).

emlun (changes requested):

index.bs
emlun: Replied in https://github.com/w3c/webauthn/pull/2095#pullrequestreview-2208153548

index.bs
MasterKale: Oh, then I misread [the example](https://webidl.spec.whatwg.org/#example-domexception-derived-interface) - I thought it was saying that stuff came along for free, not that it was spelling out exactly what had to be written out as per the steps you highlighted. That's...suboptimal, but if it's the reality then I suppose I'll expand all this out.

index.bs
MasterKale: Well that's unfortunate. If we can't throw `TimeoutError` here then that completely eliminates the possibility of defining a new `TimeoutError`. I'd definitely like to explore this at the next meeting.

index.bs
MasterKale: Well that's convenient, I'll go ahead and take out the custom one I defined in this PR.

index.bs
emlun: Yeah, we can probably make this work, we just need to do some due diligence to make sure we're not breaking something. This was written way back in 2017 (commit https://github.com/w3c/webauthn/commit/38e2c4fc25f5622d3a28152951471b15cc81e875 of PR #687), two years before L1, so it's quite likely that it's no longer relevant. Very early on in its life, these procedures were not the asynchronous "wait a while and negotiate with the user before returning a response" we know today, but instead would immediately attempt using any or all currently connected authenticators and then return once all or enough of those attempts were finished. The asynchronous model removes most of the concerns that an "early return" could indicate that a particular credential is present without user consent.

Actually, on review, this may have been mostly redundant [even since that same PR](https://github.com/w3c/webauthn/pull/687#issuecomment-359087435). I think the concern for `makeCredential` was most likely eliminated in PR #809. And I think both operations were already asynchronous even in the early commits of PR #687, but they had very recently become so (part of that change is in #687 itself). Maybe this concern was just because our thinking was still oriented around the semi-synchronous model.

MasterKale: From WG Meeting (7/31): We confirmed it's okay to to continue incorporating TimeoutError here. I'll go back over the surrounding language here.

selfissued (approved):
I support adding these error codes after the cleanups suggested by other commenters.

index.bs
```diff
@@ -2294,10 +2314,21 @@ The following {{DOMException}} exceptions can be raised:
     :   {{UnknownError}}
     ::  The [=authenticator=] could not process the supplied options,
         or encountered an error while creating the new credential.
+    
+    :   {{UserCancellationError}}
+    ::  The user has exercised a user agent user-interface option
+        to end the ceremony.
+
+    :   {{UserHybridCancellationError}}
+    ::  The user has exercised a user agent user-interface option
```
timcappalli, line 2323: In most cases, hybrid is not handled by the user agent. Is this error only for the cases where the user agent implements hybrid directly?

index.bs
```diff
@@ -2519,7 +2550,23 @@ When this method is invoked, the user agent MUST execute the following algorithm
...
+            
+            If the user agent is informing the user of an inability to continue the ceremony
+            due to missing {{AuthenticatorTransport/hybrid}} prerequisites,
+            throw a "{{HybridPrerequisitesError}}" {{DOMException}}.
+
+            If the user agent is prompting the user to complete the ceremony using an authenticator
+            that may be available over the {{AuthenticatorTransport/hybrid}} transport,
+            throw a "{{UserHybridCancellationError}}" {{DOMException}}.
+
+            If the user agent is informing the user that
```
timcappalli, line 2563: nit: does the user agent typically inform the user of any of this?

index.bs
```diff
@@ -2846,6 +2893,10 @@ The following {{DOMException}} exceptions can be raised:
     ::  The ceremony was cancelled by an {{AbortController}}.
         See [[#sctn-abortoperation]] and [[#sctn-sample-aborting]].
 
+    :   {{HybridPrerequisitesError}}
```
timcappalli, line 2896: I don't really understand where these hybrid errors would be used. There is no way for an RP to mandate a specific transport without an allow list, and even with an allowlist, all credential IDs with hybrid would also have internal.

We also don't provide this level of granularity for other transports. What if I don't have an NFC reader? What if it is the Bluetooth transport and Bluetooth is not available?

IMO, this level of detail should not be disclosed to the RP. Is the WebAuthn client's responsibility to remediate transport issues and/or guide the user to an alternative option.

Hybrid availability is also advertised in `getClientCapabilities()`.

index.bs
MasterKale: I suppose yes, Chrome's prompt is what I had in mind when I was thinking about this:

Screenshot 2024-08-07 at 11 43 36 AM

Could a blocker to codifying an error like `UserHybridCancellationError` be that when the platform handles hybrid it might not tell the user agent that the user cancelled out of prompt to try hybrid? 🤔

index.bs
MasterKale: Chrome can when it's handling hybrid `HybridPrerequisitesError`:

image

image

The macOS platform authenticator (obviously not a user agent, to another comment you left earlier) also shows something similar:

image

The fact that we now have platforms able to handle hybrid is complicating these new hybrid errors...

index.bs
MasterKale: > I don't really understand where these hybrid errors would be used.

I tried to capture my thinking [in the explainer](https://github.com/w3c/webauthn/wiki/Explainer:-New-Error-Codes-(2024-Edition)#hybridprerequisiteserror):

> The browser wanted to prompt the user to try for a hybrid authentication, but found that it lacked OS-level permission to use Bluetooth. After notifying the user of this, the user cancels the ceremony. The browser raises a `HybridPrerequisitesError` error, **which the Relying Party detects and shows the user how to grant the browser Bluetooth permission.** The user is able to turn on Bluetooth on their own and then successfully complete auth.

These errors can enable RPs to offer additional remediation guidance over top of whatever a user agent may wish to implement.

> There is no way for an RP to mandate a specific cross-platform transport without an allow list, and even with an allowlist, all credential IDs with hybrid would also have internal.

An assumption here is that the `"internal"` transport was a miss so the user agent proposes an attempt via `"hybrid"`.

> We also don't provide this level of granularity for other transports. What if I don't have an NFC reader? What if it is the Bluetooth transport and Bluetooth is not available?

To this I ask, could we not expose more granular errors for other transports?

timcappalli: > These errors can enable RPs to offer additional remediation guidance over top of whatever a user agent may wish to implement.

This is not the RP's responsibility, and honestly if I were a user, I would be suspicious of a website asking me to change system settings.

> An assumption here is that the "internal" transport was a miss so the user agent proposes an attempt via "hybrid".

But the client has no way to know whether the request will be successful even if hybrid is possible. This error can lead the RP to believe that if the hybrid protocol issues are resolved, there will be a successful ceremony, which is not always the case. So then you end up with two failed transactions. 

> To this I ask, could we not expose more granular errors for other transports?

Because it is the WebAuthn client's responsibility to provide all of this, as it is responsible for interfacing with authenticators.

index.bs
timcappalli: Right, but in most cases, it is not the user agent. Chrome on Mac and Chrome on Linux are the exceptions.

MasterKale:
I'm inclined to break this PR up to try and get some of the less contentious new error codes into L3 before the upcoming deadline (which those are, I'll try and identify before today's WG meeting.)

index.bs
MasterKale: I'm taking `UserHybridCancellationError` out of the running for now, I don't want it to block some of the other errors.

index.bs
MasterKale: I'm taking `HybridPrerequisitesError ` out of the running for now, I don't want it to block some of the other errors.

index.bs
MasterKale: This error is being considered for a different PR than this one now.

index.bs
MasterKale: @emlun I've made these changes in a way that I think satisfies all of what https://webidl.spec.whatwg.org/#idl-DOMException-derived-interfaces says about adding custom DOMErrors:

Screenshot 2024-08-28 at 11 02 06 AM

Screenshot 2024-08-28 at 11 02 11 AM

index.bs
```diff
@@ -1946,11 +1946,17 @@ a numbered step. If outdented, it (today) is rendered as a bullet in the midst o
...
             ::  [=set/For each=] |authenticator| in |issuedRequests| invoke the [=authenticatorCancel=] operation on |authenticator|
-                and [=set/remove=] |authenticator| from |issuedRequests|.
+                and [=set/remove=] |authenticator| from |issuedRequests|. Throw a "[=create/TimeoutError=]" {{DOMException}}.
 
             :   If the user exercises a user agent user-interface option to cancel the process,
             ::  [=set/For each=] |authenticator| in |issuedRequests| invoke the [=authenticatorCancel=] operation on |authenticator|
-                and [=set/remove=] |authenticator| from |issuedRequests|. Throw a "{{NotAllowedError}}" {{DOMException}}.
+                and [=set/remove=] |authenticator| from |issuedRequests|.
+
+                If the user agent is informing the user of an inability to continue the ceremony
```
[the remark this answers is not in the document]
MasterKale: I've punted on these errors for now. Let's revisit this when a "New Error Codes Part 2" becomes feasible.

index.bs
```diff
@@ -2460,11 +2478,17 @@ When this method is invoked, the user agent MUST execute the following algorithm
...
         ::  [=set/For each=] |authenticator| in |issuedRequests| invoke the [=authenticatorCancel=] operation on
-            |authenticator| and [=set/remove=] |authenticator| from |issuedRequests|.
+            |authenticator| and [=set/remove=] |authenticator| from |issuedRequests|. Throw a "[=get/TimeoutError=]" {{DOMException}}.
 
         :   If the user exercises a user agent user-interface option to cancel the process,
         ::  [=set/For each=] |authenticator| in |issuedRequests| invoke the [=authenticatorCancel=] operation on |authenticator|
-            and [=set/remove=] |authenticator| from |issuedRequests|. Throw a "{{NotAllowedError}}" {{DOMException}}.
+            and [=set/remove=] |authenticator| from |issuedRequests|.
+            
+            If the user agent is informing the user of an inability to continue the ceremony
```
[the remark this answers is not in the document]
MasterKale: I've punted on these errors for now. Let's revisit this when a "New Error Codes Part 2" becomes feasible.

MasterKale:
I've taken `HybridPrerequisitesError` and `UserHybridCancellationError` out of the running for now for sake of pursuing incremental improvement in WebAuthn's error messages instead.

index.bs
emlun: Discussion moved to #2132 (please feel free to continue under the assumption that a `TimeoutError` is okay).

index.bs
```diff
@@ -2698,7 +2745,7 @@ When this method is invoked, the user agent MUST execute the following algorithm
             1.  Return |constructAssertionAlg| and terminate this algorithm.
     </dl>
 
-1. Throw a "{{NotAllowedError}}" {{DOMException}}. In order to prevent information leak that could identify the
+1. Throw a "{{TimeoutError}}" {{DOMException}}. In order to prevent information leak that could identify the
```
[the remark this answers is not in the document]
emlun: Discussion moved to #2132.

index.bs
```diff
@@ -2280,9 +2288,23 @@ The following {{DOMException}} exceptions can be raised:
         the [=client=] does not support [[#sctn-related-origins|related origin requests]]
         or the [$related origins validation procedure$] failed.
 
+    :   {{TimeoutError}}
+    ::  The ceremony was cancelled by the user agent after exceeding the time limit permitted for the ceremony.
+        See [[#sctn-timeout-recommended-range]] for more information.
+
+    :   {{UnknownError}}
+    ::  The [=authenticator=] could not process the supplied options,
+        or encountered an error while creating the new credential.
```
emlun, line 2297: I don't think `error code equivalent to "UnknownError"` is actually passed through by the client operation, is it? I think it's currently only `InvalidStateError` that's passed through like that.

index.bs
```diff
@@ -1965,7 +1965,15 @@ a numbered step. If outdented, it (today) is rendered as a bullet in the midst o
...
-                and [=set/remove=] |authenticator| from |issuedRequests|. Throw a "{{NotAllowedError}}" {{DOMException}}.
+                and [=set/remove=] |authenticator| from |issuedRequests|.
+
+                If the user agent is informing the user that
+                the last used |authenticator| cannot collect [=user verification=] when
+                <code>|pkOptions|.{{PublicKeyCredentialCreationOptions/authenticatorSelection}}.{{AuthenticatorSelectionCriteria/userVerification}}</code>
+                is set to {{UserVerificationRequirement/required}},
+                throw a "{{UserVerificationError}}" {{DOMException}}.
+
+                Otherwise, throw a "{{UserCancellationError}}" {{DOMException}}.
```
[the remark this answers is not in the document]
MasterKale: Yeah, alright, nice suggestion. I'll switch to this instead.

varunbgit:
@agl @pascoej @jschanck 
@MasterKale 
What is the plan to merge this PR and release it to public?

mitchgl:
Are there any updates on this? Seems useful for context for RP's

MasterKale:
This work is on the back burner for now while we continue to work on getting L3 through the process to get it to Recommendation so that it officially supplants L2. I have every intention of picking this back up when we're further along this process.

index.bs
```diff
@@ -1965,7 +1965,15 @@ a numbered step. If outdented, it (today) is rendered as a bullet in the midst o
...
             :   If the user exercises a user agent user-interface option to cancel the process,
             ::  [=set/For each=] |authenticator| in |issuedRequests| invoke the [=authenticatorCancel=] operation on |authenticator|
-                and [=set/remove=] |authenticator| from |issuedRequests|. Throw a "{{NotAllowedError}}" {{DOMException}}.
+                and [=set/remove=] |authenticator| from |issuedRequests|.
+
+                If the user agent is informing the user that
+                the last used |authenticator| cannot collect [=user verification=] when
+                <code>|pkOptions|.{{PublicKeyCredentialCreationOptions/authenticatorSelection}}.{{AuthenticatorSelectionCriteria/userVerification}}</code>
+                is set to {{UserVerificationRequirement/required}},
+                throw a "{{UserVerificationError}}" {{DOMException}}.
```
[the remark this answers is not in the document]
MasterKale: Coming back to this comment, `ConstraintError` is currently used to communicate a couple of issues **during registration**:

- A resident key was required but the authenticator doesn't support it
- UV was required but the authenticator couldn't collect it

I'm kinda bummed out by the prospect of mashing "no rkey support" and "can't do UV" into the same bucket, even though I have to admit it'd make sense to make the change you suggest and turn this proposed `UserVerificationError` into a `ConstraintError` because an existing similar error state already sets precedent.

We're probably well past the time now where lack of rkey support is the minority cause of a `ConstraintError`; put another way, my concern may not hold as much water because the current state of passkeys will have RPs implying that a `ConstraintError` is more likely "can't do UV" than "can't create an rkey."

I'd like to discuss this at the next WG meeting before I make any changes.

index.bs
```diff
@@ -2450,7 +2470,15 @@ When this method is invoked, the user agent MUST execute the following algorithm
...
         :   If the user exercises a user agent user-interface option to cancel the process,
         ::  [=set/For each=] |authenticator| in |issuedRequests| invoke the [=authenticatorCancel=] operation on |authenticator|
-            and [=set/remove=] |authenticator| from |issuedRequests|. Throw a "{{NotAllowedError}}" {{DOMException}}.
+            and [=set/remove=] |authenticator| from |issuedRequests|.
+            
+            If the user agent is informing the user that
+            the last used |authenticator| cannot collect [=user verification=] when
+            <code>|pkOptions|.{{PublicKeyCredentialCreationOptions/authenticatorSelection}}.{{AuthenticatorSelectionCriteria/userVerification}}</code>
+            is set to {{UserVerificationRequirement/required}},
+            throw a "{{UserVerificationError}}" {{DOMException}}.
```
MasterKale: If we have to replace the `UserVerificationError` during registration with `ConstraintError`, then we'll have to replace this one during authentication as well.

index.bs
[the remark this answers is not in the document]
emlun: Hm... it looks like we can!

[WebIDL §2.8 Exceptions](https://webidl.spec.whatwg.org/#idl-exceptions) reads:
>An example of including additional information used to construct the exception message would be:
>
>>[Throw](https://webidl.spec.whatwg.org/#dfn-throw) a "[SyntaxError](https://webidl.spec.whatwg.org/#syntaxerror)" [DOMException](https://webidl.spec.whatwg.org/#idl-DOMException) indicating that the given value had disallowed trailing spaces.

This refers to one of the steps in [§3.14.3. Creating and throwing exceptions](https://webidl.spec.whatwg.org/#js-creating-throwing-exceptions) which is linked just afterwards:

>To [create](https://webidl.spec.whatwg.org/#dfn-create-exception) a [DOMException](https://webidl.spec.whatwg.org/#idl-DOMException) given a string _name_:
>[...]
>4. Set _ex_’s [message](https://webidl.spec.whatwg.org/#domexception-message) to an [implementation-defined](https://infra.spec.whatwg.org/#implementation-defined) message appropriate for the exceptional situation. The calling specification may contain information to to help implementations construct this message.
>  Implementations need to be cautious not to leak sensitive or secured information when constructing this message, e.g., by including the URL of a cross-origin frame, or information which could identify the user.

[2.8.1. Base DOMException error names](https://webidl.spec.whatwg.org/#idl-DOMException-error-names) does state that "When creating or throwing a DOMException, specifications must use one of these names [listed in the `DOMException` names table].", but I can't see anything like that for `message`.

So yes, I think it seems like "The calling specification [WebAuthn] may contain information to to help implementations construct [_ex_'s `message`]".

PR #2097, opened 2024-07-17
Structured confirmation

rlin1:
This is an alternative to PR #2020 and does not rely on a single arbitrary string provided by the relying party. It instead allows the platform and the authenticator to differentiate "moving money from account A to account B" from "sharing health data of John Doe with hospital X".

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -3655,11 +3655,12 @@ Note: The {{CollectedClientData}} may be extended in the future. Therefore it's 
 
 <xmp class="idl">
     dictionary CollectedClientData {
-        required DOMString           type;
-        required DOMString           challenge;
-        required DOMString           origin;
-        DOMString                    topOrigin;
-        boolean                      crossOrigin;
+        required DOMString            type;
+        required DOMString            challenge;
+        required DOMString            origin;
+        DOMString                     topOrigin;
+        boolean                       crossOrigin;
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -7685,6 +7689,181 @@ To <dfn abstract-op>Create a new supplemental public key record</dfn>, perform t
         [=set/append=] this [=supplemental public key record=] to |credentialRecord|.[$credential record/supplementalPubKeys$].
 
 
+### Confirmation Extension (confirmation) ### {#sctn-confirmation-extension}
+
+This <dfn>confirmation extension</dfn> allows for capturing user confirmation. A
+   [=[RP]=] can specify the structured details of a specific confirmation type. The confirmation tpe specific structured details are intended for display
```
nicksteele, line 7695:
```suggestion
   [=[RP]=] can specify the structured details of a specific confirmation type. The confirmation type specific structured details are intended for display
```

timcappalli:
2025-03-26 WG call: this hasn't been discussed in quite some time, and the last time it was discussed there were concerns around untrusted text in a trusted UI. Planning to close this at the next meeting unless there is activity.

rlin1:
What kind of activity are you expecting here?

index.bs
rlin1: addressed.

PR #2074, opened 2024-05-20
Align RP ID string types

emlun:
`rp.id` in `PublicKeyCredentialCreationOptions` and `rpId` in `PublicKeyCredentialRequestOptions` represent the same thing, but with different types. The WG agreed on the 2024-05-15 call that both should be `USVString`. Strictly speaking this is a breaking change (changing a type bound in input (contravariant) position to be more restrictive), but in practice this shouldn't be able to break any applications since then those credentials wouldn't have worked in `get()` anyway.

Fixes #2066.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/pull/2074.html" title="Last updated on Jul 17, 2024, 12:25 PM UTC (3618093)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/2074/62b069e...3618093.html" title="Last updated on Jul 17, 2024, 12:25 PM UTC (3618093)">Diff</a>

PR #2098, opened 2024-07-17
Relax PublicKeyCredentialRequestOptions.rpId to DOMString

### LINKED ISSUE #2066: Make PublicKeyCredentialRequestOptions.rpId a DOMString
Both [`PublicKeyCredentialRpEntity.id`](https://www.w3.org/TR/webauthn-3/#dom-publickeycredentialrpentity-id) and [`PublicKeyCredentialRequestOptions.rpId`](https://www.w3.org/TR/webauthn-3/#dom-publickeycredentialrequestoptions-rpid) represent the same thing (i.e., [RP ID](https://www.w3.org/TR/webauthn-3/#rp-id)); however the former is modeled as a [`DOMString`](https://webidl.spec.whatwg.org/#idl-DOMString) while the latter is modeled as a [`USVString`](https://webidl.spec.whatwg.org/#idl-USVString). These should be the same type. Specifically `PublicKeyCredentialRequestOptions.rpId` should be a `DOMString` as well or `PublicKeyCredentialRpEntity.id` should be a `USVString`. Should be noted that [`PublicKeyCredentialRequestOptionsJSON.rpId`](https://www.w3.org/TR/webauthn-3/#dom-publickeycredentialrequestoptionsjson-rpid)  is already a `DOMString`.

emlun:
Alternative resolution to issue #2066 and PR #2074.

Fixes #2066.

This is a backwards-compatible change (changing a type bound in input (contravariant) position to be more permissive). This could have risked that WebAuthn calls in new implementations fail if run in clients that were built to the older spec and still enforce the `USVString` type in `get()`, but WebAuthn already defines that an RP ID must be a _valid domain string_:

- A [valid domain string][1] must pass the _domain to ASCII_ procedure without error.
- [Domain to ASCII][2] invokes the _processing steps in section 4_ of the same document.
- Those processing steps begin with looking up each code point in the _IDNA mapping table_ and then returning an error if any code point was mapped to **disallowed**.
- The [IDNA mapping table][3] notes that "Each table for a version of the Unicode Standard will always be backward compatible with previous versions of the table: only characters with the Status value **disallowed** may change in Status or Mapping value, [...]".
- [Version 15.1.0 of the mapping table][4] maps the range U+D800..U+DFFF to **disallowed**. This is the range of surrogate code units.
- Since this range is **disallowed** in version 15.1.0, it must also be **disallowed** in all previous versions.

Conclusion: WebAuthn RP IDs are already always scalar value strings, because valid domain strings MUST NOT contain surrogate code units. Therefore this type relaxation is backwards compatible with WebAuthn clients that still enforce `USVString` type in `get()`.

[1]: https://url.spec.whatwg.org/#valid-domain-string
[2]: https://www.unicode.org/reports/tr46/#ToASCII
[3]: https://www.unicode.org/Public/idna/
[4]: https://www.unicode.org/Public/idna/15.1.0/IdnaMappingTable.txt

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/pull/2098.html" title="Last updated on Jul 17, 2024, 12:24 PM UTC (e3603f3)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/2098/62b069e...e3603f3.html" title="Last updated on Jul 17, 2024, 12:24 PM UTC (e3603f3)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -3472,7 +3472,7 @@ an assertion. Its {{PublicKeyCredentialRequestOptions/challenge}} member MUST be
     dictionary PublicKeyCredentialRequestOptions {
         required BufferSource                challenge;
         unsigned long                        timeout;
-        USVString                            rpId;
+        DOMString                            rpId;
         sequence<PublicKeyCredentialDescriptor> allowCredentials = [];
         DOMString                            userVerification = "preferred";
         sequence<DOMString>                  hints = [];


pascoej (approved):

selfissued (approved):

MasterKale (approved):

timcappalli (approved):

nsatragno (approved):
Approved, thank you.

agl (approved):

akshayku (approved):

nsatragno:
Chromium bug: https://crbug.com/356565292

Issue #2102, opened 2024-07-22
Add `topOrigin` to the limited verification algorithm


zacknewman:
I rely on the [limited verification algorithm](https://www.w3.org/TR/webauthn-3/#clientdatajson-verification) for my WebAuthn RP library, but I had to amend it to validate [`topOrigin`](https://www.w3.org/TR/webauthn-3/#dom-collectedclientdata-toporigin).

Is there no way this can be added to the spec? I realize backward compatibility is likely the reason it was not added, but could there be a step mentioning optional steps? Or perhaps a section for an alternative "limited verification algorithm" that _doesn't_ guarantee backward compatibility?

In my implementation, I added sub-step 6 to step 1 and sub-step 2 (and two sub-steps of it) to step 11:

> 1\. The inputs to the algorithm are:
> &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;1\. A bytestring, _clientDataJSON_, that contains `clientDataJSON`— the serialized `CollectedClientData` that is to be verified.
> &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;⋮
> &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;6\. A string, _topOrigin_, that contains the expected `topOrigin` that issued the request to the user agent if _crossOrigin_ is true and if available.
> ⋮
> 11\. If _crossOrigin_ is true:
> &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;1. Append 0x74727565 (`true`) to _expected_.
> &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;2. If _topOrigin_ exists:
> &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;1. Append 0x2c22746f704f726967696e223a (`,"topOrigin":`) to _expected_.
> &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;2. Append CCDToString(`topOrigin`) to _expected_.
>12\. Otherwise, i.e. _crossOrigin_ is false:

agl:
Obviously there's a tension around how "limited" the limited algorithm is. If it keeps gaining things it'll be no better than writing a small JSON parser!

[...]

Issue #1631, opened 2021-06-19
CollectedClientData.crossOrigin default value and whether it is required


equalsJeffH:
`CollectedClientData.crossOrigin` is defined like so:
```
dictionary CollectedClientData {
    [ ... ]
    boolean                      crossOrigin;
    [ ... ]
};
```
In examining both the `[[Create]]()` and `[[DiscoverFromExternalSource]]()` methods, as well as the CollectedClientData serialization, it appears that `crossOrigin` is invariably added to `collectedClientData`, and that it's default value is (effectively) False.

Did we _not_ stipulate it as `required` in the above WebIDL declaration for backwards compatiblity reasons?

In any case, I think it'd be good to add a note along these lines (to help obviate grovelling thru the spec to figure this out):
> crossOrigin, of type boolean
>         This member contains the inverse of the sameOriginWithAncestors argument value that was passed into the internal method.
>       Note: This member is always added to `CollectedClientData` by the `[[Create]]()` and `[[DiscoverFromExternalSource]]()` methods. If for some reason it is not present in `CollectedClientData`, and the Section 5.8.1.1 Serialization is used, it defaults to False.

emlun:
It doesn't appear to be explicitly documented in #1276, but I think it seems likely that backwards compatibility is the reason we made it not `required` - RPs cannot rely on it being present in case the client is an L1 implementation.

MasterKale:
From WG call: @akshayku will create a PR to add a note similar to what was proposed above.

PR #2104, opened 2024-07-22
Add topOrigin to the limited verification algorithm.

zacknewman:
Add `topOrigin` to the limited verification algorithm.
Fixes https://github.com/w3c/webauthn/issues/2102.

***
<a href="https://pr-preview.s3.amazonaws.com/zacknewman/webauthn/pull/2104.html" title="Last updated on Aug 14, 2024, 4:42 PM UTC (60fc0e8)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/2104/2b8e368...zacknewman:60fc0e8.html" title="Last updated on Aug 14, 2024, 4:42 PM UTC (60fc0e8)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -3833,6 +3833,14 @@ Verifiers may use the following algorithm to verify an encoded {{CollectedClient
     1. A string, |origin|, that contains the expected {{CollectedClientData/origin}} that issued the request to the user agent.
     1. A boolean, |crossOrigin|, that is true if, and only if, the request should have been performed within a cross-origin <{iframe}>.
     1. An optional string, |topOrigin|, that contains the expected {{CollectedClientData/topOrigin}} that issued the request to the user agent if |crossOrigin| is true and if available.
+    1. A boolean, |requireTopOrigin|, that is true if, and only if, the verification should fail
+        if |topOrigin| is defined and the {{CollectedClientData/topOrigin}} attribute is not present in |clientDataJSON|.
+
+        This means that the verification algorithm is backwards compatible
+        with the [[webauthn-2-20210408#clientdatajson-serialization|JSON-compatible serialization algorithm]]
+        in Web Authentication Level 2 [[webauthn-2-20210408]]
+        if, and only if, |requireTopOrigin| is [FALSE].
+
 1. Let |expected| be an empty byte string.
 1. Append 0x7b2274797065223a (`{"type":`) to |expected|.
 1. Append [=CCDToString=](|type|) to |expected|.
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -3843,6 +3844,9 @@ Verifiers may use the following algorithm to verify an encoded {{CollectedClient
 1. Append 0x2c2263726f73734f726967696e223a (`,"crossOrigin":`) to |expected|.
 1. If |crossOrigin| is true:
     1. Append 0x74727565 (`true`) to |expected|.
+    1. If |topOrigin| is defined:
+        1. Append 0x2c22746f704f726967696e223a (`,"topOrigin":`) to |expected|.
+        1. Append [=CCDToString=](|topOrigin|) to |expected|.
```
emlun, line 3849: This is complicated a bit by there existing both L2 and L3 client implementations at the same time, where L2 clients may include `"crossOrigin":true` but not include `"topOrigin"`. This logic as currently written handles the cases asymmetrically:

- It is probably correct that an "L2" RP, expecting `crossOrigin: true, topOrigin: undefined`, should **accept** an L3 client data with `"crossOrigin":true,"topOrigin":"<something>"`, since `"topOrigin"` is an unknown attribute to that RP and so it falls under the "allow any trailing attributes" rule in the final step.

- The inverse is not necessarily correct, though: an "L3" RP expecting `crossOrigin: true, topOrigin: "<something>"` might still want to accept L2 client data with `"crossOrigin":true` (and `"topOrigin"` not present). That currently **fails** this rule.

So I think it's at least worth to somehow call out that the RP may want to retry the verification with expecting `topOrigin: undefined`, perhaps as an "If the RP wants to support L2 client implementations, retry the verification procedure with the |topOrigin| input undefined" follow-on to step 13. I don't really like the way that's going if we continue adding more attributes, but I think it's necessary.

emlun: Hm, I just noticed that another wrinkle to this is that `crossOrigin` is not backwards-compatible in the same way in the limited verification procedure.

As noted in #1631, `crossOrigin` is defined as `optional` in the Web IDL, probably for backwards compatibility with L1 clients. But the limited verification procedure does not maintain this backwards compatibility with `"crossOrigin"` being absent in the client data.

So maybe we shouldn't attempt to keep `topOrigin` backwards-compatible either?

emlun: Oh, right. That is an excellent point.

emlun: I added a _requireTopOrigin_ parameter and extended the verification steps to use it to decide backwards compatibility, that resolves this review comment. I also deleted the _crossOrigin_ parameter as its value can now equivalent to _topOrigin_ being defined.

emlun:
I added a _requireTopOrigin_ parameter and extended the verification steps to use it to decide backwards compatibility with L2. I also deleted the _crossOrigin_ parameter as its value is now equivalent to _topOrigin_ being defined.

timcappalli (approved):

agl (approved):

pascoej (approved):

agl:
From the call of 2024-08-28: we would merge this if the IPR bot would let us. Simone is going to try running the IPR approval process to get this merged.

emlun (approved):

plehegar:
Note: the IPR approval process was applied and [concluded successfully](https://lists.w3.org/Archives/Team/team-ipr-archive/2024Sep/0019.html).

PR #2107, opened 2024-07-27
Make `AuthenticatorAttestationResponseJSON.publicKeyAlgorithm` optional.

### LINKED ISSUE #2106: Make `AuthenticatorAttestationResponseJSON.publicKeyAlgorithm` optional
The motivation behind both [`AuthenticatorAttestationResponseJSON.publicKey`](https://w3c.github.io/webauthn/#dom-authenticatorattestationresponsejson-publickey) and [`AuthenticatorAttestationResponseJSON.publicKeyAlgorithm`](https://w3c.github.io/webauthn/#dom-authenticatorattestationresponsejson-publickeyalgorithm) is the same: [easy access to credential data](https://w3c.github.io/webauthn/#sctn-public-key-easy). For good reason though,  `AuthenticatorAttestationResponseJSON.publicKey` is not required since technically such data exists in the required [`AuthenticatorAttestationResponseJSON.attestationObject`](https://w3c.github.io/webauthn/#dom-authenticatorattestationresponsejson-attestationobject). I believe the same should be true for `AuthenticatorAttestationResponseJSON.publicKeyAlgorithm` since it doesn't really serve purpose without `AuthenticatorAttestationResponseJSON.publicKey` also existing.

zacknewman:
`AuthenticatorAttestationResponseJSON.publicKeyAlgorithm` serves no purpose without `AuthenticatorAttestationResponseJSON.publicKey` also existing. The latter is wisely optional, so this should be too.
Fixes #2106.

***
<a href="https://pr-preview.s3.amazonaws.com/zacknewman/webauthn/pull/2107.html" title="Last updated on Jul 27, 2024, 1:53 AM UTC (68a3709)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/2107/2b8e368...zacknewman:68a3709.html" title="Last updated on Jul 27, 2024, 1:53 AM UTC (68a3709)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1657,7 +1657,7 @@ that are returned to the caller when a new credential is created, or a new asser
         // algorithm then the public key must be parsed directly from
         // attestationObject or authenticatorData.
         Base64URLString publicKey;
-        required COSEAlgorithmIdentifier publicKeyAlgorithm;
+        COSEAlgorithmIdentifier publicKeyAlgorithm;
         // This value contains copies of some of the fields above. See
         // section “Easily accessing credential data”.
         required Base64URLString attestationObject;


agl:
Backwards compatibility, and aspects of browser/security key interaction require the non-JSON field to be optional. But those cases don't apply here.

agl:
Hmm, having just said that, it does beg the question of what a browser is to do when converting to JSON a response for which it does not recognise the public key algorithm. Still, more nuance than simply making them optional is called for: the point is that COSE is an obscure standard and forcing people to decode it is an impediment to implementations, which as much as possible should be able to depend on these fields being populated.

PR #2127, opened 2024-08-19
Codify semantic line breaks as editorial convention

### LINKED ISSUE #2045: Editorial convention: Semantic line breaks
[Semantic line breaks](https://sembr.org/) is an editorial convention that may facilitate more efficient work on the text, while making no difference to the rendered output.

Currently, the WebAuthn spec source is a mix of at least 3 styles of line breaks:

- No line breaks: a whole paragraph of text may be a single line in the source. [Example in spec](https://github.com/w3c/webauthn/blob/89f204f4b1b02a147dfcdee4fbd88319cab5e104/index.bs#L1215)
- Hard-wrapped text: add a line break before each word that would extend past some particular column width. The column limit also varies throughout the document. [Example in spec](https://github.com/w3c/webauthn/blob/89f204f4b1b02a147dfcdee4fbd88319cab5e104/index.bs#L1253-L1258)
- [Semantic line breaks](https://sembr.org/): add line breaks at semantically meaningful points, such as after punctuation, before subordinate clauses, etc. [Example in spec](https://github.com/w3c/webauthn/blob/89f204f4b1b02a147dfcdee4fbd88319cab5e104/index.bs#L1393-L1402)

Of the three, I find that semantic line breaks makes text easiest to work with, in several ways:

- It typically results in a text width that's comfortable to work with in a text editor as well as in GitHub diffs and commandline Git diffs.
- Small changes to a paragraph are easy to identify and do not "inflate" diffs:
  - With no line breaks, the [entire line shows as changed](https://github.com/w3c/webauthn/commit/39f4984f9fe7f5bf2165f568071698df9a8a59b6#diff-5e793325cd2bfc452e268a4aa2f02b4024dd9584bd1db3c2595f61f1ecf7b985R7281) unless word diff highlighting is used. GitHub diffs do usually highlight word differences, but only up to some limited line length - very long lines do not seem to get word diffs.
  - With hard-wrapped text, the whole rest of the paragraph shows as changed if the word wrapping changes. This works okay with word diff in commandline Git, but [not very well in GitHub diffs](https://github.com/w3c/webauthn/commit/2de6bfbb93a35216667efff9a7f6755de2c10b5a#diff-5e793325cd2bfc452e268a4aa2f02b4024dd9584bd1db3c2595f61f1ecf7b985R1928-R1932).
  - With semantic line breaks, [neighboring lines usually do not need to change](https://github.com/w3c/webauthn/commit/8dd71903a34b2d85de97ea23c0f71718ec58b568#diff-5e793325cd2bfc452e268a4aa2f02b4024dd9584bd1db3c2595f61f1ecf7b985R6636) since they are separated at semantically meaningful borders.

## Proposed Change

Adopt semantic line breaks as an editorial convention for future changes:

- Do not change existing text to use semantic line breaks.
- When changing text that is not formatted with semantic line breaks, reformat changing lines to use semantic line breaks. Do not reformat unchanged neighboring lines to use semantic line breaks unless it's very few lines, or it otherwise helps readability, or it otherwise seems like a good idea.
- When adding new text, use semantic line breaks.
- Take the [Semantic Line Breaks specification](https://sembr.org/) as a set of guidelines, not a rigid set of rules. If a contributor proposes changes that clearly do not use semantic line breaks, inform or remind them of the convention and ask them to reformat, but respect their choice of where to place line breaks and do not "bikeshed" the details. The point is to make the text easier to work with, not to enforce a style.
  - In particular, I think the recommended max line length of 80 characters is too restrictive with how much indentation and markup syntax we use. Something more like ~100 characters, or ~80 characters excluding indentation, may be more appropriate for this document. But I don't think we should set a hard limit - use good judgement of where to draw the line in each case.
- Add a notice of this convention to [`CONTRIBUTING.md`](https://github.com/w3c/webauthn/blob/main/CONTRIBUTING.md).

This should apply to changes made after the decision and to pull requests opened after the decision. Existing pull requests MAY be updated to the new convention, but are not required to apply this convention retroactively.

emlun:
Fixes #2045.

This should take effect for pull requests opened after this PR is merged. Existing pull requests MAY be updated to this new convention, but are not required to apply this convention retroactively.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -23,3 +23,51 @@ contributors in the pull request's body or in subsequent comments.
 
  If you are making a pull request on behalf of someone else but you had no part in designing the 
  feature, you can remove yourself with the above syntax.
+
+
+## Editorial conventions
+
+When editing the spec, please follow these editorial conventions.
+
+
+### Text macros
+
+Where appropriate, use the
+[text macros defined near the beginning of the document](https://github.com/w3c/webauthn/blob/93193a1dcfddf6a6daa4726fafa4f556bff203ca/index.bs#L51-L60).
+For example, use `[TRUE]` to render `<code>true</code>`
+and use `[=[RP]=]` to render a linked reference to the term "Relying Party".
+
+
+### Semantic line breaks
+
+Use [semantic line breaks][sembr].
+To summarize, this means adding line breaks at semantically meaningful points,
+such as after punctuation, before subordinate clauses, etc.
+This makes changes easier to review and helps automatic conflict resolution,
+as small changes are more likely to affect only one or a few lines
+rather than an entire paragraph.
+For more examples and rationale, see: https://github.com/w3c/webauthn/issues/2045.
+
+Use semantic line breaks along these additional guidelines:
+
+- Do not change existing text just for the sake of introducing semantic line breaks.
+- When changing text that is not formatted with semantic line breaks,
+  reformat changing lines to use semantic line breaks.
+  Do not reformat unchanged neighboring lines to use semantic line breaks unless it's very few lines,
+  or it otherwise helps readability,
+  or it otherwise seems like a good idea.
+- When changing text that already uses semantic line breaks,
+  add and remove semantic line breaks as appropriate but only as needed for the change.
+  Do not break or join lines that would otherwise remain unchanged.
+- When adding new text, use semantic line breaks.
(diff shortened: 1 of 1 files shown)


timcappalli (approved):

MasterKale (approved):

akshayku (approved):

selfissued (approved):
Makes sense.  For what it's worth, I was already doing this.

PR #2138, opened 2024-09-06
Adds dfn for passkey in passkey platform authenticator and exports

timcappalli:
Closes #2136

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/pull/2138.html" title="Last updated on Sep 6, 2024, 4:28 PM UTC (ee25bae)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/2138/a871f79...ee25bae.html" title="Last updated on Sep 6, 2024, 4:28 PM UTC (ee25bae)">Diff</a>

PR #2140, opened 2024-09-06
Rename first-factor roaming authenticator and integrate passkey term …

emlun:
This is a meta-PR into #2138. An update pass over this text is probably long overdue, and as part of that I think we should rename "first-factor roaming authenticator" to "passkey roaming authenticator".


timcappalli (approved):
Consistency is good! LGTM.

PR #2148, opened 2024-09-18
Add userName and userDisplayName to webdriver

### LINKED ISSUE #2143: Add `userName` and `userDisplayName` to WebDriver's `Credential Parameters` JSON object
## Proposed Change

Add `userName` and `userDisplayName` to WebDriver's [`Credential Parameters` JSON object](https://w3c.github.io/webauthn/#credential-parameters). It's not currently possible to obtain those values from either a `get` call or the automation API, and we need that to write WPTs for the signal* methods.

nsatragno:
This PR adds the userName and userDisplayName properties to the webdriver's credential parameters. These properties are useful to test the new signalCurrentUserDetails method, both on WPTs and for web developers.

Closes #2143

 -- This will be done as a follow-up

Implementation commitment:
- [ X ] Chromium ([link to issue](https://crbug.com/367915823))

***
<a href="https://pr-preview.s3.amazonaws.com/nsatragno/webauthn/pull/2148.html" title="Last updated on Oct 1, 2024, 3:14 PM UTC (146a63e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/2148/efdf948...nsatragno:146a63e.html" title="Last updated on Oct 1, 2024, 3:14 PM UTC (146a63e)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -8013,6 +8013,22 @@ The <dfn>Credential Parameters</dfn> is a JSON [=Object=] passed to the [=remote
                 </td>
                 <td>boolean</td>
             </tr>
+            <tr>
+                <td>|userName|</td>
+                <td>
+                    The {{PublicKeyCredentialUserEntity|user}}'s {{PublicKeyCredentialEntity/name}} associated to the credential.
+                    If unset, the value will default to the empty string.
+                </td>
+                <td>string</td>
+            </tr>
+            <tr>
+                <td>|userDisplayName|</td>
+                <td>
+                    The {{PublicKeyCredentialUserEntity|user}}'s {{PublicKeyCredentialUserEntity/displayName}} associated to the credential.
+                    If unset, the value will default to the empty string.
+                </td>
+                <td>string</td>
+            </tr>
         </tbody>
     </table>
 </figure>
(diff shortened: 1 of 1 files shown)


nicksteele (approved):

timcappalli (approved):
LGTM

nsatragno:
Nina to fix the empty name / display name problem

index.bs
```diff
@@ -8011,6 +8011,22 @@ The <dfn>Credential Parameters</dfn> is a JSON [=Object=] passed to the [=remote
                 </td>
                 <td>boolean</td>
             </tr>
+            <tr>
+                <td>|userName|</td>
+                <td>
+                    The {{PublicKeyCredentialUserEntity|user}}'s {{PublicKeyCredentialEntity/name}} associated to the credential.
+                    If unset, the value will default to the empty string.
```
MasterKale, line 8018: Hmm, thinking about ways in which `PublicKeyCredentialUserEntity.name` can be set, [`user.name` is a required value when calling `.create()`](https://w3c.github.io/webauthn/#dom-publickeycredentialentity-name), and [`signalCurrentUserDetails()` requires `name`](https://w3c.github.io/webauthn/#dom-currentuserdetailsoptions-name) be set as well. It feels like a potential foot gun to allow `user.name` to end up as an empty string when using the virtual authenticator API 🤔

MasterKale (changes requested):

pascoej (approved):

index.bs
emlun: On the other hand, I assume that requiring that these values be set would make the WebDriver method(s) incompatible with callers that haven't yet updated to set them?

emlun (approved):

index.bs
nsatragno: We cannot make it required. But we can make it so that if present, it has to be set. If not present, it will be left unchanged.

This means that it is still possible to create a credential with an empty name through the virtual authenticator, but that was already a problem before this PR and nobody ever complained, so it's probably not a big deal.

emlun: > if present, it has to be set

As in not `null`? That seems reasonable.

nsatragno: Wait a minute, it's required but it can be empty :thinking: 

That means that the current text is fine. If left unset, it defaults to the empty string. RPs can make credentials with empty names through regular webauthn already.

MasterKale (approved):
Alright, I think Nina's addressed my concerns. lgtm :shipit:

PR #2174, opened 2024-10-01
Add test vectors for PRF extension

### LINKED ISSUE #2088: Add examples for PRF extension
## Proposed Change

Multiple people who are implementing PRF extensions have got the implementation wrong regarding extension fields in request and response. 

We have to add some examples for this extension.

emlun:
Closes #2088.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/pull/2174.html" title="Last updated on Oct 2, 2024, 9:32 PM UTC (a231517)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/2174/efdf948...a231517.html" title="Last updated on Oct 2, 2024, 9:32 PM UTC (a231517)">Diff</a>


MasterKale (approved):

emlun:
> Do you think it makes sense to add something about the client _not_ sending this data to the server?
>[...]
>Is it "obvious" that the client should not send this data seeing how if the server had this info it's now at best devolved into a shared secret?

Hm. I wanted to say that yes, this should be obvious enough in the use cases where this is relevant, and that there are other use cases where you actually do want to send the PRF outputs to the server. But we do since recently have this step in both [§7.1. Registering a New Credential](https://w3c.github.io/webauthn/#sctn-registering-a-new-credential) and [§7.2. Verifying an Authentication Assertion](https://w3c.github.io/webauthn/#sctn-verifying-assertion):

>4. Let _clientExtensionResults_ be the result of calling _credential_.[`getClientExtensionResults()`](https://w3c.github.io/webauthn/#dom-publickeycredential-getclientextensionresults).

which perhaps complicates the matter a bit. The Relying Party Operations sections also don't really make a distinction between client-side and server-side parts of the RP, rather treating both as one entity, so it also doesn't seem quite appropriate to simply add something simple like ", omitting any properties that should not be sent to the server" either.

Maybe the right way to go is to just call out the client-side-only use case in the description of [`AuthenticationExtensionsPRFOutputs.results`](https://w3c.github.io/webauthn/#dom-authenticationextensionsprfoutputs-results), and warn to not accidentally send them back to the server if that is undesired for the use case.

index.bs
```diff
@@ -7504,6 +7504,218 @@ Note: this extension may be implemented for [=authenticators=] that do not use [
...
+          second: new Uint32Array([0x982e17c4, 0xc397902e, 0xb70c6c9a, 0x5b37cb20, 0xadfce392, 0xe4634a15, 0x09f1933a, 0x73191e6b]),
+        }
+      }
+    }
+    </xmp>
+
+Pseudo-random values used in this section were generated as follows:
+
+- <code>"e02eZ9lPp0UdkF4vGRO4-NxlhWBkL1FCmsmb1tTfRyE" = Base64Url(SHA-256(UTF-8("WebAuthn PRF test vectors") || 0x00))</code>
+- <code>h&apos;c4172e982e9097c39a6c0cb720cb375b92e3fcad154a63e43a93f1096b1e1973' = SHA-256(UTF-8("WebAuthn PRF test vectors") || 0x01)</code>
```
agl, line 7608:
```suggestion
- <code>h'c4172e982e9097c39a6c0cb720cb375b92e3fcad154a63e43a93f1096b1e1973' = SHA-256(UTF-8("WebAuthn PRF test vectors") || 0x01)</code>
```

agl (approved):
(I did not confirm the hmac_secret values in detail; i.e. by recomputing the encryptions and decryptions.)

Firehed:
> Is it "obvious" that the client should not send this data seeing how if the server had this info it's now at best devolved into a shared secret?

From my eyes, it is not - though I'll readily admit that's more due to unfamiliarity with the use-cases than going through the actual spec. While it's out of scope on _this_ PR to add more info about use-cases, I think it would be a worthwhile addition to the spec to go into more details there. [MasterKale's gist](https://gist.github.com/MasterKale/dbe39a01438251f0cbd55576304731fd) provides a bit more context, but I think more detail about what party manages and/or retains what data would go a long way towards keeping implementations on the successful path.

When it comes to cryptography, and especially handling of key material (or inputs to derive it), I'd recommend erring strongly on the side of overly-explicit.

> That's where I am leaning too: a brief cautionary note that mentions one may want to omit this information when sending getClientExtensionResults or toJSON to the server when the server should not have this sensitive data.

It may be too late to materially change the mechanics, but if this is the case, it seems to me that the sensitive data should have a different access path that's outside of those APIs. In particular, `toJSON` is really intended to dump the whole thing to the server for processing (or, at least, that's how I plan on using it once there's enough browser support); default-including data in there that should not leave the client is asking for problematic implementations.

index.bs
emlun: Thanks, but no - this renders incorrectly for similar reasons as noted in PR #2175. Bikeshed replaces `&apos;` with `'` in the output, but replaces `'` with "’" (U+2019) here.

Although now that you pointed it out, maybe it should be `&apos;` on both sides in that case...

Firehed:
> A harsh rebuttal would be that any RP that "accidentally" sends sensitive data is likely not qualified to do whatever it is they are trying to do (e.g., password manager). 

I agree, but that doesn't tend to stop it in practice. Mistakes happen.

>  I think there are other reasons one may want to remove certain data from toJSON anyway. [...]

Indeed! However, once you go down that path, you can rapidly end up at a point where you're doing enough customization to the format that you're better off not bothering with `toJSON` at all and manually assembling the parts of the format you _do_ want (as you have to do in practice today).

Which is in no way to suggest that your approach is wrong or invalid, but I want to keep in mind the common-case usage of that API. If there's a chance to avoid a pretty major security footgun before that API has widespread support, it's probably wise to do so.

Maybe my concerns are overstated. I suspect that most parties will ignore the extension entirely, and (as you suggest) it'll probably only get used by RPs that actually do need it and would understand the security implications.

In any case, the addition of test vectors for this looks great and I'm excited to see them added. Maybe it'd be better to spin this aspect off into a separate issue?

emlun:
Yes, let's move that discussion to #2177 (thanks @zacknewman!).

akshayku (approved):

selfissued (approved):
Approved during the 9-Oct-24 call

PR #2176, opened 2024-10-03
Non-incrementing signature counters could be due to race condition

### LINKED ISSUE #2172: Should race condition be added as a reason for a signature counter not increasing?
Currently [§ 6.1.1.](https://w3c.github.io/webauthn/#sctn-sign-counter) only states the following as reasons for why a signature counter does not increase:

> If either is non-zero, and the new [`signCount`](https://w3c.github.io/webauthn/#authdata-signcount) value is less than or equal to the stored value, a cloned authenticator may exist, or the authenticator may be malfunctioning.

However it's possible an older response—from the perspective of the authenticator—is processed after a newer one since there is no guarantee that data that is sent from the client before other data sent from the same client will be received let alone processed before the other. This primarily affects passkey flows and not second-factor ones; since for the latter, RPs can either force at most one active ceremony per credential or use the `signCount` at the time the ceremony began to compare to.

Is this deemed too unlikely to warrant mention?

As an explicit example:

User starts a passkey authentication ceremony and sends the updated signature counter, _C1_. Same user starts another passkey authentication ceremony and sends a newer signature counter, _C2_. Before the server receives the response containing _C1_, it receives and processes the response containing _C2_. Finally the server receives and processes the response containing _C1_ which is less than the current counter, _C2_.

There are many reasons for why such a thing happens: BGP routes messed up, wonky load balancer, black hole causing time dilation[^1]. The point is that there are technically legitimate reasons for a counter not increasing.

[^1]: I'm obviously being facetious about this one

sbweeden:
Closes #2172 

Includes observation that an RP processing order race condition may be a cause of non-incrementing signature counters during validation.

***
<a href="https://pr-preview.s3.amazonaws.com/sbweeden/webauthn/pull/2176.html" title="Last updated on Oct 23, 2024, 6:29 PM UTC (8c2cfdd)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/2176/5831a2c...sbweeden:8c2cfdd.html" title="Last updated on Oct 23, 2024, 6:29 PM UTC (8c2cfdd)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -4811,9 +4811,9 @@ leave the <code>[=authData/signCount=]</code> in the [=authenticator data=] cons
 
 A [=[RP]=] stores the [=signature counter=] of the most recent [=authenticatorGetAssertion=] operation. (Or the counter from the [=authenticatorMakeCredential=] operation if no [=authenticatorGetAssertion=] has ever been performed on a credential.) In subsequent
 [=authenticatorGetAssertion=] operations, the [=[RP]=] compares the stored [=signature counter=] value with the new
-<code>[=authData/signCount=]</code> value returned in the assertion's [=authenticator data=]. If either is non-zero, and the new <code>[=authData/signCount=]</code> value is less than or equal to the stored value, a cloned authenticator may exist, or the authenticator may be malfunctioning.
+<code>[=authData/signCount=]</code> value returned in the assertion's [=authenticator data=]. If either is non-zero, and the new <code>[=authData/signCount=]</code> value is less than or equal to the stored value, a cloned authenticator may exist, or the authenticator may be malfunctioning, or a race condition might exist where the relying party is receiving and processing assertions in an order other than the order they were generated at the authenticator.
 
-Detecting a [=signature counter=] mismatch does not indicate whether the current operation was performed by a cloned authenticator or the original authenticator.  [=[RPS]=] should address this situation appropriately relative to their individual situations, i.e., their risk tolerance.
+Detecting a [=signature counter=] mismatch does not indicate whether the current operation was performed by a cloned authenticator or the original authenticator.  [=[RPS]=] should address this situation appropriately relative to their individual situations, i.e., their risk tolerance or operational factors that might result in an acceptable reason for non-increasing values.
 
 Authenticators:
 - SHOULD implement per credential [=signature counters=].  This prevents the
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -4811,9 +4811,9 @@ leave the <code>[=authData/signCount=]</code> in the [=authenticator data=] cons
 
 A [=[RP]=] stores the [=signature counter=] of the most recent [=authenticatorGetAssertion=] operation. (Or the counter from the [=authenticatorMakeCredential=] operation if no [=authenticatorGetAssertion=] has ever been performed on a credential.) In subsequent
 [=authenticatorGetAssertion=] operations, the [=[RP]=] compares the stored [=signature counter=] value with the new
-<code>[=authData/signCount=]</code> value returned in the assertion's [=authenticator data=]. If either is non-zero, and the new <code>[=authData/signCount=]</code> value is less than or equal to the stored value, a cloned authenticator may exist, or the authenticator may be malfunctioning.
+<code>[=authData/signCount=]</code> value returned in the assertion's [=authenticator data=]. If either is non-zero, and the new <code>[=authData/signCount=]</code> value is less than or equal to the stored value, a cloned authenticator may exist, or the authenticator may be malfunctioning, or a race condition might exist where the relying party is receiving and processing assertions in an order other than the order they were generated at the authenticator.
```
emlun, line 4814: Per the recently added [editorial convention](https://github.com/w3c/webauthn/blob/main/CONTRIBUTING.md#semantic-line-breaks), please use semantic line breaks for at least the new text here and just below. For example:

```suggestion
<code>[=authData/signCount=]</code> value returned in the assertion's [=authenticator data=]. If either is non-zero, and the new <code>[=authData/signCount=]</code> value is less than or equal to the stored value, a cloned authenticator may exist, or the authenticator may be malfunctioning,
or a race condition might exist where the relying party is receiving and processing assertions
in an order other than the order they were generated at the authenticator.
```

If you like you could break up the original line too, since it's already in the diff due to the punctuation change, but I would personally leave it as is to minimize the diff.

MasterKale (changes requested):
WebAuthn is a human-centric API with ceremony times long enough to prevent most race conditions related to counters not incrementing between responses. What types of RP use cases is this updated verbiage intended to address? I can only think of convoluted scenarios like "a single passkey used by multiple users across multiple kiosks not syncing counter state" that this might unblock. But otherwise why would we weaken the counter check like this?

MasterKale (approved):
We had a good talk about this at today's WG call. I've come around and think this is fine **guidance** to include in the spec (since it doesn't change anything about actual counter-related response verification logic)

nsatragno:
What if the RP stores a snapshot of the counter of all authenticators for a user at the time a challenge is vended? That way you'd be able to compare the counter against the value at the time you requested an assertion.

sbweeden:
> What if the RP stores a snapshot of the counter of all authenticators for a user at the time a challenge is vended? That way you'd be able to compare the counter against the value at the time you requested an assertion.

That's a clever remediation, but the focus of this PR is simply to describe some possible causes for counter mismatch rather than suggest how an RP should deal with it.

PR #2178, opened 2024-10-04
Add note about possible sensitive data in extensions

### LINKED ISSUE #2177: Add cautionary note about extension data in the ceremony criteria
In #2174 it was mentioned that a cautionary note about _not_ sending PRF data to the server may be appropriate for use cases where the output is used as a decryption key that should always remain client-side. I propose adding notes to the registration and authentication ceremony sections that express something like below:

> Note: Since some extension data may need to remain client-side, the [Relying Party](https://w3c.github.io/webauthn/#relying-party) MUST be prepared to remove data in _clientExtensionResults_ client-side before sending data to the server for ceremony completion. Since signatures are based on _authData_, the Relying Party MUST NOT rely on extensions whose corresponding [authenticator extension outputs](https://w3c.github.io/webauthn/#authenticator-extension-output) in the [`extensions`](https://w3c.github.io/webauthn/#authdata-extensions) in _authData_ contains data that should remain client-side when relying on the server to complete the ceremony.

Should be noted that there already exist two notes about extension processing in Steps 20 and 19 of the registration and authentication ceremonies respectively; thus this does not seem out of place. By being placed in the ceremony criteria, we don't have to worry about applying disclaimers for specific extensions (e.g., PRF).

zacknewman:
Closes #2177.

***
<a href="https://pr-preview.s3.amazonaws.com/zacknewman/webauthn/pull/2178.html" title="Last updated on Oct 4, 2024, 8:22 PM UTC (6885f04)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/2178/1e2256d...zacknewman:6885f04.html" title="Last updated on Oct 4, 2024, 8:22 PM UTC (6885f04)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -5997,6 +5997,12 @@ a numbered step. If outdented, it (today) is rendered as a bullet in the midst o
 
         Note: Since all extensions are OPTIONAL for both the [=client=] and the [=authenticator=], the [=[RP]=] MUST also be
         prepared to handle cases where none or not all of the requested extensions were acted upon.
+
+        Note: Since some extension data may need to remain client-side, [=[RPS]=] MUST be prepared to remove data in
+        |clientExtensionResults| client-side before sending data to the server for ceremony completion. Since signatures are
+        based on |authData|, the [=[RP]=] MUST NOT rely on extensions whose corresponding [=authenticator extension outputs=]
+        in the <code>[=authData/extensions=]</code> in |authData| contains data that should remain client-side when relying on
+        the server to complete the ceremony.
     </li>
 
 1. Determine the attestation statement format by performing a USASCII case-sensitive match on |fmt| against the set of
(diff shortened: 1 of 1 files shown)


emlun (changes requested):
Thanks, but I don't think this quite hits the mark. The approach suggested in https://github.com/w3c/webauthn/pull/2174#issuecomment-2389695303 and seemingly agreed in https://github.com/w3c/webauthn/pull/2174#issuecomment-2389714033 was rather to call this out in the definition of the `results` output of the PRF extension. The approach here has a couple of problems in my opinion:

- This instead adds the warning to the RP Operations processing steps; at that point of processing it is likely too late as the data has usually already been sent to the server and these steps are run on the server side.
- I don't think the generalization to "[any] extension data [that] may need to remain client-side" seems all that useful, especially since this avoids pointing out PRF as the one (currently) relevant example. Similarly, the "[=[RP]=] MUST NOT rely on extensions [...]" part doesn't really seem actionable and rather sounds to me like such an extension is ill-specified.
- As noted in #1979 we shouldn't have normative language ("MUST" etc.) in notes. Admittedly we have a lot of them already, but we shouldn't add more.

I therefore propose PR #2183 to supersede this.

PR #2216, opened 2024-11-27
Fix CDDL inconsistencies in attStmtType and compound format

### LINKED ISSUE #2210: Compound attestation statement format is incompatible with attStmtTemplate
[6.5.4. Generating an Attestation Object](https://w3c.github.io/webauthn/#sctn-generating-an-attestation-object) defines a CDDL template for attestation statement formats:

```
attObj = {
            authData: bytes,
            $$attStmtType
         }

attStmtTemplate = (
                      fmt: text,
                      attStmt: { * tstr => any } ; Map is filled in by each concrete attStmtType
                  )

; Every attestation statement format must have the above fields
attStmtTemplate .within $$attStmtType
```

Note that the `attStmt` member is defined as a CBOR map.

[§8.9. Compound Attestation Statement Format](https://w3c.github.io/webauthn/#sctn-compound-attestation) defines a choice for the `$$attStmtType` group socket:

```
$$attStmtType //= (
                      fmt: "compound",
                      attStmt: [2* nonCompoundAttStmt]
                  )

nonCompoundAttStmt = { $$attStmtType } .within { fmt: text .ne "compound" }
```

Note that the `attStmt` member here is a CBOR array.

I also don't think the expression `attStmtTemplate .within $$attStmtType` successfully encodes the intent `Every attestation statement format must have the above fields`, for two reasons: it does not define a CDDL _rule_ since it contains no `=` sign, and even if it did, the `.within` control operator would apply only to the new type defined by that rule, but not to the `attObj` type.

On the 2024-11-20 WG call it was said that there are implementations of compound attestation shipping, so the preferred resolution to this is to relax the template to allow array-based attestation statements.

## Proposed Change

1. Inline the `.within` control operator into the `attObj` definition:

    ```
    attObj = {
        authData: bytes,
        $$attStmtType
    } .within attStmtTemplate  ; Every attestation statement format must have the fields below
    ```

2. Add a choice to `attStmtTemplate` to allow an array for `attStmt`:

    ```
    attStmtTemplate = {
        authData: bytes,
        fmt: text,
        attStmt: (
            { * tstr => any } ; Map is filled in by each concrete attStmtType
            //
            [ * any ]         ; attStmt may also be an array
        ),
    }
    ```

emlun:
Closes #2210. Also fixes a couple of other issues I noticed while working on this:

- [§6.5.2. Attestation Statement Formats](https://w3c.github.io/webauthn/#sctn-attestation-formats) refers to a CDDL extension point `$attStmtFormat` that (no longer?) exists, it should be `$$attStmtType`.
- The `.within { fmt: text .ne "compound" }` control on `nonCompoundAttStmt` inadvertently makes `nonCompoundAttStmt` unable to match anything, because maps are matched exhaustively (see [CDDL](https://datatracker.ietf.org/doc/html/rfc8610#section-2.1), so the right-hand-side of `.within` forbids additional fields beyond `fmt`.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/pull/2216.html" title="Last updated on Nov 27, 2024, 12:37 PM UTC (8b29bec)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/2216/3bba180...8b29bec.html" title="Last updated on Nov 27, 2024, 12:37 PM UTC (8b29bec)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -5736,7 +5736,7 @@ template:
 
 - <strong>Syntax:</strong>
     The syntax of an [=attestation statement=] produced in this format, defined using CDDL [[!RFC8610]] for the extension point
-    `$attStmtFormat` defined in [[#sctn-generating-an-attestation-object]].
+    `$$attStmtType` defined in [[#sctn-generating-an-attestation-object]].
 
 - <dfn>Signing procedure</dfn>:
     The [=signing procedure=] for computing an [=attestation statement=] in this [=attestation statement format|format=] given
--- a/index.bs
+++ b/index.bs
@@ -5832,17 +5832,17 @@ the [=authenticator=] MUST:
 
     ```
     attObj = {
-                authData: bytes,
-                $$attStmtType
-             }
+        authData: bytes,
 
-    attStmtTemplate = (
-                          fmt: text,
-                          attStmt: { * tstr => any } ; Map is filled in by each concrete attStmtType
-                      )
+        ; Each choice in $$attStmtType defines the fmt value and attStmt structure
+        $$attStmtType
+    } .within attStmtTemplate
 
-    ; Every attestation statement format must have the above fields
-    attStmtTemplate .within $$attStmtType
+    attStmtTemplate = {
+        authData: bytes,
+        fmt: text,
+        attStmt: { * tstr => any } ; Map is filled in by each concrete attStmtType
+    }
     ```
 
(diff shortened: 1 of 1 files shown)


timcappalli (approved):
Thanks for fixing this!

MasterKale (approved):

PR #2217, opened 2024-11-27
Make remaining `USVString`s `DOMString`s

### LINKED ISSUE #2203: Replace `USVString` with `DOMString`
It would appear that the same justification that was used for #2115 applies to the remaining areas where `USVString` is used:

> As recommended by the [Web IDL spec](https://webidl.spec.whatwg.org/#idl-USVString):
>
>> Specifications should only use `USVString` for APIs that perform text processing and need a string of scalar values to operate on. Most APIs that use strings should instead be using `DOMString`, which does not make any interpretations of the code units in the string. When in doubt, use `DOMString`.

Currently the only places where `USVString` is used are the following extensions:
* [`appid`](https://w3c.github.io/webauthn/#dom-authenticationextensionsclientinputs-appid)
* [`appidExclude`](https://w3c.github.io/webauthn/#dom-authenticationextensionsclientinputs-appidexclude)
* [`prf`](https://w3c.github.io/webauthn/#dictdef-authenticationextensionsprfinputs)

Should these all be changed to `DOMString`?

zacknewman:
Fixes #2203.

***
<a href="https://pr-preview.s3.amazonaws.com/zacknewman/webauthn/pull/2217.html" title="Last updated on Nov 27, 2024, 10:19 PM UTC (95fb560)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/2217/3bba180...zacknewman:95fb560.html" title="Last updated on Nov 27, 2024, 10:19 PM UTC (95fb560)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -311,7 +311,6 @@ spec:url; type:dfn; text:domain
 spec:url; type:dfn; for:url; text:host
 spec:url; type:dfn; text:valid domain;
 spec:webidl; type:dfn; text:DOMString
-spec:webidl; type:dfn; text:USVString
 spec:webidl; type:interface; text:Promise
 </pre>
 
@@ -7216,10 +7215,10 @@ but the currently used [=RP ID=] might be "example.com".
 :: [=authentication extension|Authentication=]
 
 : Client extension input
-:: A single USVString specifying a FIDO |AppID|.
+:: A single DOMString specifying a FIDO |AppID|.
     <xmp class="idl">
     partial dictionary AuthenticationExtensionsClientInputs {
-      USVString appid;
+      DOMString appid;
     };
     </xmp>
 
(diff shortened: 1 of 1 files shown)


emlun (approved):
Thank you!

I double-checked that this won't be a backwards compatibility trap, and concluded that:

- `evalByCredential` further restricts the permitted character set to just the Base64Url alphabet, so it is not possible for a `DOMString` value to be a valid `evalByCredential` key but not a valid `USVString`.
- For FIDO AppID we refer to the [FIDO AppID and Facet Specification](https://fidoalliance.org/specs/fido-v2.0-id-20180227/fido-appid-and-facets-v2.0-id-20180227.html), and the reference chain ultimately leads to [section 4.2 of RFC 5891](https://www.rfc-editor.org/rfc/rfc5891#section-4.2) and a table in [RFC 5892](https://www.rfc-editor.org/rfc/rfc5892) which lists `D800..FA0D  ; DISALLOWED  # <Non Private Use High Surrogate>..CJK COMPAT`. [`USVString`](https://webidl.spec.whatwg.org/#idl-USVString) is defined as a string containing no code points in the range U+D800 to U+DFFF, inclusive. Thus it is not possible for a `DOMString` value to be a valid FIDO AppID but not a valid `USVString`.

So this change seems safe to me.

timcappalli (approved):

agl (approved):

akshayku (approved):

pascoej (approved):

PR #2219, opened 2024-12-03
Fix AuthenticationExtensionsAuthenticatorInputs/Outputs CDDL

tidoust:
According to the [CDDL grammar](https://datatracker.ietf.org/doc/html/rfc8610#appendix-B), after a control operator (called `ctlop` in the grammar), there can only be a `type2` production. In a `type2` production, wrapping parentheses can only be used to wrap a `type` production. `tstr => any` is a `group` production. As such, unless I missed something, it cannot be wrapped in parentheses to produce a `type2` production. It rather needs to be wrapped in curly braces or brackets.

In other words, from a CDDL grammar perspective, this is an invalid type:
```cddl
foo .within ( tstr => any )
```

This is a valid type:
```cddl
foo .within { tstr => any }
```

This update fixes the CDDL type definitions that used the `.within` operator with an invalid type2.

I don't believe the change introduces any semantic change (so more an editorial update than anything else) but note I'm no CDDL expert, I just happened to have been playing with the CDDL grammar recently... If the semantics change, well, the problem still stands: the current definition is not valid CDDL.

***
<a href="https://pr-preview.s3.amazonaws.com/tidoust/webauthn/pull/2219.html" title="Last updated on Dec 18, 2024, 5:30 PM UTC (a9b57fa)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/2219/3bba180...tidoust:a9b57fa.html" title="Last updated on Dec 18, 2024, 5:30 PM UTC (a9b57fa)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -4013,7 +4013,7 @@ This is a dictionary containing the [=client extension output=] values for zero 
 
 ```
 AuthenticationExtensionsAuthenticatorInputs = {
-  * $$extensionInput .within ( tstr => any )
+  * $$extensionInput .within { tstr => any }
 }
 ```
 
@@ -4028,7 +4028,7 @@ This type is not exposed to the [=[RP]=], but is used by the [=client=] and [=au
 
 ```
 AuthenticationExtensionsAuthenticatorOutputs = {
-  * $$extensionOutput .within ( tstr => any )
+  * $$extensionOutput .within { tstr => any }
 }
 ```
 
--- a/index.bs
+++ b/index.bs
@@ -4013,8 +4013,8 @@ This is a dictionary containing the [=client extension output=] values for zero 
 
 ```
 AuthenticationExtensionsAuthenticatorInputs = {
-  * $$extensionInput .within { tstr => any }
-}
+  * $$extensionInput
+} .within { * tstr => any }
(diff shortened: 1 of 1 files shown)


emlun (changes requested):
Well spotted, thanks! Unfortunately this still isn't valid CDDL, though. As I [recently had clarified on the CDDL mail list](https://mailarchive.ietf.org/arch/msg/cbor/lz2snfZHg0oGie8n04fcbCzkNFo/), control operators only applies to types, not groups. This PR fixes the right-hand side, but the left-hand side is still a group, so the grammar is still invalid. I will open a meta-PR to complete the change.

emlun:
This was effectively merged in #2221, but didn't get automatically closed because https://github.com/tidoust/webauthn/pull/1 was merged with a rebase-merge and therefore created a new commit https://github.com/tidoust/webauthn/commit/a9b57fa91f4014f9697e369e76736660b91d3931 that is not the same as commit https://github.com/w3c/webauthn/pull/2221/commits/5d855e79c4242c330c737f00a1f6b15c85c1fd1a in #2221.

Thanks!

PR #2253, opened 2025-02-03
Backup preference 

### LINKED ISSUE #2252: Provide a way for an RP to indicate backup preference during credential registration
## Description

Passkey Providers/Authenticators now support backed-up and non-backup credentials. It varies from one provider to the other the choices they provide.

For an RP, currently they have no way to express their preference regarding backed credentials. For example, some enterprises and high security consumer RPs may want a non-backed up credential. Other RPs can prefer backed-up credential for their use cases regarding availability everywhere w.r.t current and future devices.

For the providers who support both backed-up and non-backed-up credential, RP's preference helps guide the user. 

Hence, we need a way for an RP to indicate their backup preference in WebAuthn spec.

akshayku:
Closes #2252 

This PR provides a way for an RP to indicate their backup preference regarding the credential being created at registration time via PublicKeyCredentialHint.

Note: Given the nature of different options provided by the providers/authenticators, their capabilities, user choices etc., RP must expect both backed-up and non-backed-up credentials in the registration responses.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -4428,6 +4428,8 @@ Note: The {{ClientCapability}} enumeration is deliberately not referenced, see [
         "security-key",
         "client-device",
         "hybrid",
+        "backed-up",
+        "non-backed-up",
     };
 </xmp>
 
@@ -4453,6 +4455,12 @@ Note: The {{PublicKeyCredentialHint}} enumeration is deliberately not referenced
     ::  Indicates that the [=[RP]=] believes that users will satisfy this request with general-purpose [=authenticators=] such as smartphones. For example, a consumer [=[RP]=] may believe that only a small fraction of their customers possesses dedicated security keys. This option also implies that the local [=platform authenticator=] should not be promoted in the UI.
 
         For compatibility with older user agents, when this hint is used in {{PublicKeyCredentialCreationOptions}}, the {{AuthenticatorSelectionCriteria/authenticatorAttachment}} SHOULD be set to {{AuthenticatorAttachment/cross-platform}}.
+
+    :   <dfn>backed-up</dfn>
+    ::  Indicates that the [=[RP]=] prefers a [=backed up=] credential.
+
+    :   <dfn>non-backed-up</dfn>
+    ::  Indicates that the [=[RP]=] prefers a non [=backed up=] credential.
 
 </div>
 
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -4428,6 +4428,8 @@ Note: The {{ClientCapability}} enumeration is deliberately not referenced, see [
         "security-key",
         "client-device",
         "hybrid",
+        "backed-up-cred,
```
sbweeden, line 4431: Nit - missing trailing quotes

sbweeden (changes requested):
Only a minor editorial.

sbweeden:
A more wholisitc comment about this approach - keep in mind that there are two flags - BE (backup eligible) and BS (backup state). I think the PR should explicitly state that the hint is in regards to BE = true/false, not BS.

MasterKale (changes requested):
I think I'm fine with the high level concept, but I disagree on the semantics of specifying it via `hints`. `PublicKeyCredentialHint` are **authenticator** properties, while this proposed backup eligibility hint is **credential** property.

The main issue is that `hints` is ordered by preference; I can imagine that hints processing rules would get quite complicated (and RPs would easily get confused too) if they had to accommodate all the permutations of these values and their positions in the array.

Maybe we define a new "`credentialHints`"?

akshayku:
> I think I'm fine with the high level concept, but I disagree on the semantics of specifying it via `hints`. `PublicKeyCredentialHint` communicate a preference for **authenticator** properties, while this proposed backup eligibility hint represents a desired **credential** property.
> 
> The main issue is that values in `hints` are ordered by preference; I can imagine that hints processing rules would get quite complicated (and RPs would easily get confused too) if they had to accommodate all the permutations of these values and their positions in the array.
> 
> Maybe we define a new "`credentialHints`"?

When we added hints to the spec, I thought the purpose of naming it as a general purpose "hint" name is that we can add more things to this array in the future without changing the IDL. I was also not thinking about Authenticator hint vs Credential hint and interestingly we named it as PublicKeyCredentialHint instead of PublicKeyAuthenticatorHint. 

Regarding the complication of processing rules, anything we add to hints array in the future which are not conveying already present transport info (either via AuthenticatorHints or CredentialHints), will arise in that processing complication anyway. So in effect, it looks like that hint is now effectively a static transport specific option. 

I have opened an alternative backup preference specific option as part of PublicKeyCredentialCreationOptions in #2259.

emlun:
I don't see a problem with having this in `hints`. Nothing in their definition reserves them for only authenticator properties, it just happens that only authenticator properties are defined at the moment.

PR #2265, opened 2025-02-21
chore: try 3 for fixing the issue template

timcappalli:
Sorry, there is no way to preview these without pushing to main.


james-d-elliott:
As a suggestion you can actually preview them in a dummy repo, it's how I usually make changes personally.

emlun (approved):

Issue #1643, opened 2021-07-09
Use of in-field metadata not preferred


aphillips:
6.4.2. Language and Direction Encoding
https://www.w3.org/TR/webauthn-2/#sctn-strings-langdir

> The first encodes a language tag with the code point U+E0001 followed by the ASCII values of the language tag each shifted up by U+E0000. For example, the language tag “en-US” becomes the code points U+E0001, U+E0065, U+E006E, U+E002D, U+E0055, U+E0053.

The above mechanism was introduced in this version of webauthn to address I18N's comment (w3c/webauthn#1526) about the need for language/direction metadata. It would be preferable to introduce separate metadata fields. It's unclear if this mechanism is a good idea vs. doing nothing, since presumably implementations might be created and thus constitute a backwards compatibility issue.

equalsJeffH:
See also issue #1646

aphillips:
In our visit to your meeting at TPAC 2024, we discussed the various open I18N issues. This comment is to record what we said the resolution would likely be and I18N's reaction to it. I will add comments linking to this one to the other issues we have open, all of which relate to the same general problem.

---

In our meeting we discussed that inventing a new language/direction metadata serialization was undesirable, especially since no implementations currently have coded it or support decoding it. If any implementation did start to use this encoding, the result would be the addition of gibberish to values displayed in all other implementations on the Web.

Instead, we discussed that you would remove the additional serialization. For existing fields that antedate standardization, there appears to be no way to introduce field-specific metadata. For new fields that you add to Webauthn, you will follow the [single language localizable text field](https://www.w3.org/TR/string-meta/#single-linguistic-field) best practices in String-Meta.

It occurred to me after the meeting to suggest providing a [resource-wide default](https://www.w3.org/TR/string-meta/#resource_wide_default) would be a way to possibly address the existing fields. These would be additional optional new fields in authentication request/response structures such as `AuthenticatorResponse` that allow the locale and direction of any contained natural language strings to be indicated without placing a burden on existing systems.

In any case, if you remove the lang/dir serialization from the spec and close this (and related issues), I18N will not object to your progressing.

aphillips:
[...]

PR #2280, opened 2025-04-09
Replace in-field string metadata with resource-level default fields

emlun:
This addresses #1643:

> We request that you do the following:
> 
> * Remove the in-field encoding of language metadata using Unicode tag characters
> * Remove the in-field encoding of string direction using strongly directional marks
> * Add a document-level field for the document language and a document-level field for default direction; such fields would be optional, since you have an installed base

New `language` and `direction` attributes on `PublicKeyCredentialUserEntity` seems like the most appropriate implementation of a "document-level field", since these are the only natural language string parameters we have. The new attributes won't work with currently existing CTAP authenticators, but their processing is specified as "best-effort" (at least that's the intent) so RPs should be aware that they might not work.

---

Implementation commitment:

Documentation and checks

- N/A ~~Affects privacy~~
- N/A ~~Affects security~~
- N/A ~~Updated explainer~~

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -240,6 +240,10 @@ spec: RFC8610; urlPrefix: https://tools.ietf.org/html/rfc8610
     type: dfn
         text: group sockets; url: section-3.9
 
+spec: String-Meta; urlPrefix: https://www.w3.org/TR/2024/NOTE-string-meta-20241017
+    type: dfn
+        text: block direction; url: dfn-block-direction
+        text: resource-wide default; url: dfn-resource-wide-default
 </pre> <!-- class=anchors -->
 
 <!-- L128 spec:webappsec-credential-management-1; type:dictionary; for:/; text:CredentialRequestOptions -->
@@ -1767,6 +1771,9 @@ When this method is invoked, the user agent MUST execute the following algorithm
 
 1. If the length of <code>|pkOptions|.{{PublicKeyCredentialCreationOptions/user}}.{{PublicKeyCredentialUserEntity/id}}</code> is not between 1 and 64 bytes (inclusive) then throw a {{TypeError}}.
 
+1. If <code>|pkOptions|.{{PublicKeyCredentialCreationOptions/user}}.{{PublicKeyCredentialUserEntity/language}}</code> is present
+    and is not a [=well-formed language tag=], throw a "{{SyntaxError}}" {{DOMException}}.
+
 1. Let |callerOrigin| be {{PublicKeyCredential/[CREATE-METHOD]/origin}}. If |callerOrigin| is an [=opaque origin=], throw a "{{NotAllowedError}}" {{DOMException}}.
 
 1. Let |effectiveDomain| be the |callerOrigin|'s [=effective domain=].
@@ -2224,6 +2231,10 @@ The following {{DOMException}} exceptions can be raised:
         the [=client=] does not support [[#sctn-related-origins|related origin requests]]
         or the [$related origins validation procedure$] failed.
 
+    :   {{SyntaxError}}
+    ::  <code>{{PublicKeyCredentialCreationOptions/user}}.{{PublicKeyCredentialUserEntity/language}}</code>
+        was present and was not a [=well-formed language tag=].
+
     :   {{NotAllowedError}}
     ::  A catch-all error covering a wide range of possible reasons,
         including common ones like the user canceling out of the ceremony.
@@ -2476,7 +2487,8 @@ When this method is invoked, the user agent MUST execute the following algorithm
 
                 1. Prompt the user to optionally select a [=DiscoverableCredentialMetadata=] from |silentlyDiscoveredCredentials|.
                     The prompt SHOULD display values from the [=DiscoverableCredentialMetadata/otherUI=] of each [=DiscoverableCredentialMetadata=],
-                    such as {{PublicKeyCredentialEntity/name}} and {{PublicKeyCredentialUserEntity/displayName}}.
+                    such as {{PublicKeyCredentialEntity/name}} and {{PublicKeyCredentialUserEntity/displayName}},
+                    taking into account {{PublicKeyCredentialUserEntity/language}} and {{PublicKeyCredentialUserEntity/direction}} if available.
 
                     Let |credentialMetadata| be the [=DiscoverableCredentialMetadata=] chosen by the user, if any.
 
@@ -2926,6 +2938,8 @@ value and terminate the operation.
         required Base64URLString        id;
         required DOMString              name;
         required DOMString              displayName;
+        DOMString                       language;
+        DOMString                       direction;
     };
 
     dictionary PublicKeyCredentialDescriptorJSON {
@@ -3006,6 +3020,8 @@ value and terminate the operation.
         required Base64URLString               userId;
         required DOMString                     name;
         required DOMString                     displayName;
+        DOMString                              language;
+        DOMString                              direction;
     };
 </xmp>
 
@@ -3212,7 +3228,9 @@ accidentally omits valid [=credential IDs=] from
 #### {{PublicKeyCredential/signalCurrentUserDetails(options)}} #### {#sctn-signalCurrentUserDetails}
 
 The {{PublicKeyCredential/signalCurrentUserDetails(options)|signalCurrentUserDetails}} method signals the user's
-current {{PublicKeyCredentialEntity/name}} and {{PublicKeyCredentialUserEntity/displayName}}.
+current {{PublicKeyCredentialEntity/name}} and {{PublicKeyCredentialUserEntity/displayName}},
+and accompanying {{PublicKeyCredentialUserEntity/language}}
+and {{PublicKeyCredentialUserEntity/direction}} metadata if applicable.
 
 Upon invocation of {{PublicKeyCredential/signalCurrentUserDetails(options)}}, the
 [=client=] executes these steps:
@@ -3220,6 +3238,9 @@ Upon invocation of {{PublicKeyCredential/signalCurrentUserDetails(options)}}, th
 1. If the result of [=base64url encoding | base64url decoding=]
     <code>|options|.{{CurrentUserDetailsOptions/userId}}</code> is an error,
     then return [=a promise rejected with=] a {{TypeError}}.
+1. If <code>|options|.{{CurrentUserDetailsOptions/language}}</code> is present
+    and is not a [=well-formed language tag=],
+    then return [=a promise rejected with=] a "{{SyntaxError}}" {{DOMException}}.
 1. Let |p| be the result of executing the [$Asynchronous RP ID validation
     algorithm$] with
     <code>|options|.{{CurrentUserDetailsOptions/rpId}}</code>.
@@ -3240,8 +3261,10 @@ The <dfn for="signal method/authenticator actions">currentUserDetails</dfn>
     <code>[=credentials map=][|options|.{{CurrentUserDetailsOptions/rpId}}, |userId|]</code>.
 1. If |credential| does not exist, abort these steps.
 1. Update the |credential|'s [=public key credential source/otherUI=] to match
-    <code>|options|.{{CurrentUserDetailsOptions/name}}</code> and
-    <code>|options|.{{CurrentUserDetailsOptions/displayName}}</code>.
+    <code>|options|.{{CurrentUserDetailsOptions/name}}</code>,
+    <code>|options|.{{CurrentUserDetailsOptions/displayName}}</code>,
+    <code>|options|.{{CurrentUserDetailsOptions/language}}</code>
+    and <code>|options|.{{CurrentUserDetailsOptions/direction}}</code>.
 
 <div class="example">
 
@@ -3252,7 +3275,8 @@ PublicKeyCredential.signalCurrentUserDetails({
     rpId: "example.com",
     userId: "aabbcc",  // user handle, base64url.
     name: "New user name",
-    displayName: "New display name"
+    displayName: "New display name",
+    language: "en",
 });
 ```
 
@@ -3556,10 +3580,6 @@ associated with or [=scoped=] to, respectively.
                 [[!RFC8266]] for the Nickname Profile of the PRECIS FreeformClass [[!RFC8264]],
                 when setting {{PublicKeyCredentialEntity/name}}'s value, or displaying the value to the user.
 
-              - This string MAY contain language and direction metadata.
-                [=[RPS]=] SHOULD consider providing this information if setting the member to a value other than the [=RP ID=].
-                See [[#sctn-strings-langdir]] about how this metadata is encoded.
-
               - [=Clients=] SHOULD perform enforcement, as prescribed in Section 2.3 of
                 [[!RFC8266]] for the Nickname Profile of the PRECIS FreeformClass [[!RFC8264]],
                 on {{PublicKeyCredentialEntity/name}}'s value prior to displaying the value to the user or
@@ -3578,8 +3598,6 @@ associated with or [=scoped=] to, respectively.
                 as prescribed in Section 3.4.3 of [[!RFC8265]] for the UsernameCasePreserved Profile of the PRECIS
                 IdentifierClass [[!RFC8264]], when setting {{PublicKeyCredentialEntity/name}}'s value, or displaying the value
                 to the user.
-
-              - This string MAY contain language and direction metadata. [=[RPS]=] SHOULD consider providing this information. See [[#sctn-strings-langdir]] about how this metadata is encoded.
 
               - [=Clients=] SHOULD perform enforcement, as prescribed in Section 3.4.3 of [[!RFC8265]]
                 for the UsernameCasePreserved Profile of the PRECIS IdentifierClass [[!RFC8264]],
@@ -3620,6 +3638,8 @@ credential.
     dictionary PublicKeyCredentialUserEntity : PublicKeyCredentialEntity {
         required BufferSource   id;
         required DOMString      displayName;
+        DOMString               language;
+        DOMString               direction;
     };
 </xmp>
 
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -3252,7 +3275,8 @@ PublicKeyCredential.signalCurrentUserDetails({
     rpId: "example.com",
     userId: "aabbcc",  // user handle, base64url.
     name: "New user name",
-    displayName: "New display name"
+    displayName: "New display name",
+    language: "en",
```
aphillips, line 3279: You might wish to include direction here also, as an example. In this case, the direction would be `ltr`

index.bs
```diff
@@ -3666,6 +3684,21 @@ credential.
...
+
+    :   <dfn>language</dfn>
+    ::  An OPTIONAL [=resource-wide default=] language field [[!String-Meta]] for the entity.
+
+        The [=Client=] SHOULD verify that the value is a [=well-formed language tag=].
+
+        This member MAY be ignored if the combination of [=client=] and [=authenticator=] cannot suitably process and store it.
+
+    :   <dfn>direction</dfn>
+    ::  An OPTIONAL [=resource-wide default=] [=block direction=] field [[!String-Meta]] for the entity.
```
aphillips, line 3696: The term here should be `string direction` instead of block direction.

```suggestion
    ::  An OPTIONAL [=resource-wide default=] [=string direction=] field [[!String-Meta]] for the entity.
```

index.bs
```diff
@@ -3666,6 +3684,21 @@ credential.
         When storing a {{PublicKeyCredentialUserEntity/displayName}} member's value,
         the value MAY be truncated as described in [[#sctn-strings-truncation]]
         using a size limit greater than or equal to 64 bytes.
+
+    :   <dfn>language</dfn>
+    ::  An OPTIONAL [=resource-wide default=] language field [[!String-Meta]] for the entity.
```
aphillips, line 3689: Does the word "entity" clearly mean the entire `PublicKeyCredentialUserEntity`? Or should this say something like "... for all natural language strings contained by the entity" to clarify?

index.bs
```diff
@@ -3666,6 +3684,21 @@ credential.
...
+
+        The [=Client=] SHOULD verify that the value is a [=well-formed language tag=].
+
+        This member MAY be ignored if the combination of [=client=] and [=authenticator=] cannot suitably process and store it.
+
+    :   <dfn>direction</dfn>
+    ::  An OPTIONAL [=resource-wide default=] [=block direction=] field [[!String-Meta]] for the entity.
+
+        This SHOULD be one of the values `"ltr"`, `"rtl"` or `"auto"`.
```
aphillips, line 3698: Shouldn't this SHOULD be MUST? There is no utility in allowing other strings here and you permit clients to ignore unknown values later.

index.bs
```diff
@@ -5496,20 +5531,17 @@ but MUST NOT be shorter than the longest prefix substring that satisfies the siz
...
 
-The second consists of a single code point which is either U+200E (“LEFT-TO-RIGHT MARK”), U+200F (“RIGHT-TO-LEFT MARK”), or U+E007F (“CANCEL TAG”). The first two can be used to indicate directionality but SHOULD only be used when neccessary to produce the correct result. (E.g. an RTL string that starts with LTR-strong characters.) The value U+E007F is a direction-agnostic indication of the end of the [=language tag=].
+For compatibility with existing fixed-function [=authenticators=] without support for dedicated language and direction metadata fields,
+Web Authentication Level 2 included provisions for embedding such metadata in the string itself to ensure that it is transported atomically.
+This encoding is now NOT RECOMMENDED; [=clients=] and [=authenticators=] MAY ignore such encoding in new values.
+[=Clients=] and [=authenticators=] MAY detect and process language and direction metadata encoded in existing strings
```
aphillips, line 5539: You might remove "now" from "now NOT RECOMMENDED".

You might say that the new resource-level fields override the old mechanism.

aphillips (commented):
Looks good. Thank you for this!

I made a few comments.

index.bs
emlun: I considered it but decided to omit `direction` to indicate that the fields are optional, but fair enough. We can indicate that with comments instead.

index.bs
emlun: Yes, that was the intent. Does 1bcd085855738204987ee00dae5430e67c701a85 make this clearer?1

index.bs
emlun: [String-Meta](https://w3c.github.io/string-meta/#resource_wide_default) specifies:

>If your specification defines its own document level defaults, provide two optional fields:
>
>- A resource-wide default language field SHOULD be called `language` [...]
>- A resource-wide default [block direction](https://w3c.github.io/string-meta/#dfn-block-direction) field SHOULD be called `direction` and support the values `ltr`, `rtl`, or `auto`.

So is "block direction" really not the appropriate term then?

index.bs
emlun: I made it like this for consistency with our other [enumerations as DOMString types](https://w3c.github.io/webauthn/#sct-domstring-backwards-compatibility), for example the [`attestation`](https://w3c.github.io/webauthn/#dom-publickeycredentialcreationoptions-attestation) parameter. "MUST" would rather contradict the later statement that unknown values are allowed but ignored, while "SHOULD" expresses some more flexibility in case new valid values are added in the future. That is perhaps unlikely for `direction`, but you never know - maybe a `"ttb-ltr"` value could at some point make sense for some Asian scripts, for example.

index.bs
emlun: Thanks, done in 8426a007467fddbc1f9858340005e198fec3efb0.

index.bs
aphillips: No, that won't happen for direction. Vertical vs. horizontal is a presentational variation (often called the "writing mode"). Direction pertains the Unicode Bidi Algorithm and the values have been stable for a long time. But the consistency argument is reasonable.

index.bs
```diff
@@ -1767,6 +1771,9 @@ When this method is invoked, the user agent MUST execute the following algorithm
 
 1. If the length of <code>|pkOptions|.{{PublicKeyCredentialCreationOptions/user}}.{{PublicKeyCredentialUserEntity/id}}</code> is not between 1 and 64 bytes (inclusive) then throw a {{TypeError}}.
 
+1. If <code>|pkOptions|.{{PublicKeyCredentialCreationOptions/user}}.{{PublicKeyCredentialUserEntity/language}}</code> is present
+    and is not a [=well-formed language tag=], throw a "{{SyntaxError}}" {{DOMException}}.
```
aphillips, line 1775: One final thought: this requirement feels like it might push implementations that don't support the language field to do extra work. The field is optional and supporting the value is optional, but this clearly says that implementations need to check the tag's structure if the field is present.

String-Meta recommends but does not require that you emit an error here. I'd suggest handling `language` similar to `direction` and ignoring the field if the tag is not well-formed.

aphillips (approved):
Looks good. Thanks!!!

index.bs
aphillips: That might be a slight bug in our spec. The term isn't outright wrong, although I tend to think of "block direction" as the direction of text where the string is inserted and "string direction" as the base direction of the string (or, in this case, strings). I tend to think that "string direction" is clearer for implementers.

emlun:
From 2025-04-11 face-to-face WG meeting:

- It's unclear what clients would do differently for different `language` values, especially considering these values (`user.name` and `user.displayName`) are likely chosen by the user in some way. Is the `language` field needed here?

index.bs
emlun: I see what you mean, but I'm still more inclined to keep the error in the interest of "fail faster". Of course no error will be thrown here if the client implementation predates WebAuthn L3, so there will always be that possibility, but I still think that clients that know the field exists should verify its well-formedness when present - even if the client otherwise ignores the value.

I think this position is also supported by the [W3C priority of consituencies](https://www.w3.org/TR/design-principles/#priority-of-constituencies): "[...] the needs of web page authors [...] come before the needs of user agent implementors, [...]".

index.bs
emlun: Ok. Fixed in bc361f81940078bb0c023d90bc2f88c6cf91ab24, thanks!

pascoej (approved):

index.bs
```diff
@@ -8021,6 +8056,22 @@ The <dfn>Credential Parameters</dfn> is a JSON [=Object=] passed to the [=remote
                 </td>
                 <td>string</td>
             </tr>
+            <tr>
+                <td>|userLanguage|</td>
+                <td>
+                    The [=resource-wide default=] language tag for the {{PublicKeyCredentialUserEntity|user}} entity associated to the credential.
+                    This property may not be defined.
```
selfissued, line 8063:
```suggestion
                    Note that this property might not be present.
```

index.bs
```diff
@@ -8021,6 +8056,22 @@ The <dfn>Credential Parameters</dfn> is a JSON [=Object=] passed to the [=remote
...
+                </td>
+                <td>string</td>
+            </tr>
+            <tr>
+                <td>|userDirection|</td>
+                <td>
+                    The [=resource-wide default=] [=string direction=] for the {{PublicKeyCredentialUserEntity|user}} entity associated to the credential.
+                    This property may not be defined.
```
selfissued, line 8071:
```suggestion
                     Note that this property might not be present.
```

index.bs
```diff
@@ -8078,7 +8131,7 @@ The [=remote end steps=] are:
     : [=public key credential source/userHandle=]
     :: |userHandle|
     : [=public key credential source/otherUI=]
-    :: Construct from |userName| and |userDisplayName|.
+    :: Construct from |userName|, |userDisplayName|, |userLanguage| and |userDirection|.
```
selfissued, line 8134:
```suggestion
    :: Construct from |userName|, |userDisplayName|, |userLanguage|, and |userDirection|.
```

index.bs
```diff
@@ -9874,6 +9927,11 @@ Deprecations:
     [[#dictionary-pkcredentialentity]]
 - [[#sctn-android-safetynet-attestation]]
 - [=CollectedClientData/tokenBinding=] was changed to \[RESERVED].
+- In-field language and direction metadata no longer recommended:
```
selfissued, line 9930:
```suggestion
- In-field language and direction metadata are no longer recommended:
```

index.bs
```diff
@@ -9900,6 +9958,8 @@ New features:
     - [[#sctn-automation-set-credential-properties]]
 - [[#sctn-compound-attestation]]
 - [[#prf-extension]]
+- New optional {{PublicKeyCredentialUserEntity}} attributes
```
selfissued, line 9961:
```suggestion
- New optional {{PublicKeyCredentialUserEntity}} attributes:
```

index.bs
```diff
@@ -8065,6 +8116,8 @@ The [=remote end steps=] are:
  1. If |userName| is not defined, set |userName| to the empty string.
  1. Let |userDisplayName| be the |parameters|' |userDisplayName| property.
  1. If |userDisplayName| is not defined, set |userDisplayName| to the empty string.
+ 1. Let |userLanguage| be the |parameters|' |userLanguage| property.
+ 1. Let |userDirection| be the |parameters|' |userDirection| property.
```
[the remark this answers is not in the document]
emlun: Yes, judging by the processing of the similarly defined _userHandle_ property. Fixed in 6507c4c49cd8919d2745b35f016abe2729a79a72, thanks!

index.bs
emlun: Thanks, but I'm going to reject this proposal as "This property may not be defined" is the formulation already used for `userHandle` and `largeBlob`:

https://github.com/w3c/webauthn/blob/6507c4c49cd8919d2745b35f016abe2729a79a72/index.bs#L8003-L8006

https://github.com/w3c/webauthn/blob/6507c4c49cd8919d2745b35f016abe2729a79a72/index.bs#L8016-L8019

index.bs
emlun: See https://github.com/w3c/webauthn/pull/2280#discussion_r2063431989

timcappalli (approved):

index.bs
MasterKale, line 1775: WebAuthn typically raises `TypeError`'s when there's an issue with a malformed value, as documented in L3:

- https://w3c.github.io/webauthn/#sctn-create-request-exceptions
- https://w3c.github.io/webauthn/#sctn-get-request-exceptions

We should probably maintain this behavior for this exception too.

index.bs
```diff
@@ -2224,6 +2231,10 @@ The following {{DOMException}} exceptions can be raised:
         the [=client=] does not support [[#sctn-related-origins|related origin requests]]
         or the [$related origins validation procedure$] failed.
 
+    :   {{SyntaxError}}
+    ::  <code>{{PublicKeyCredentialCreationOptions/user}}.{{PublicKeyCredentialUserEntity/language}}</code>
+        was present and was not a [=well-formed language tag=].
+
```
MasterKale, line 2237: If we agree to keep things consistent with `TypeError` as I call out above, we'll want to take this out.

index.bs
```diff
@@ -3212,14 +3228,19 @@ accidentally omits valid [=credential IDs=] from
...
 
 Upon invocation of {{PublicKeyCredential/signalCurrentUserDetails(options)}}, the
 [=client=] executes these steps:
 
 1. If the result of [=base64url encoding | base64url decoding=]
     <code>|options|.{{CurrentUserDetailsOptions/userId}}</code> is an error,
     then return [=a promise rejected with=] a {{TypeError}}.
+1. If <code>|options|.{{CurrentUserDetailsOptions/language}}</code> is present
+    and is not a [=well-formed language tag=],
+    then return [=a promise rejected with=] a "{{SyntaxError}}" {{DOMException}}.
```
MasterKale, line 3243: Tagging another `SyntaxError` pending a decision about sticking with `TypeError` as called out above.

index.bs
```diff
@@ -5496,20 +5534,17 @@ but MUST NOT be shorter than the longest prefix substring that satisfies the siz
...
+For compatibility with existing fixed-function [=authenticators=] without support for dedicated language and direction metadata fields,
+Web Authentication Level 2 included provisions for embedding such metadata in the string itself to ensure that it is transported atomically.
+This encoding is NOT RECOMMENDED; [=clients=] and [=authenticators=] MAY ignore such encoding in new values.
+[=Clients=] and [=authenticators=] MAY detect and process language and direction metadata encoded in existing strings
+as described in [[webauthn-2-20210408#sctn-strings-langdir|Web Authentication Level 2 §6.4.2. Language and Direction Encoding]].
 
-So the string “حبیب الرحمان” could have two different DOMString values, depending on whether the language was encoded or not. (Since the direction is unambiguous a directionality marker is not needed in this example.)
+Instead, the Level 3 [=Web Authentication API=] provides [=resource-wide default=] fields for encoding language and direction metadata
+which SHOULD be used when supported by the [=client=] and [=authenticator=].
```
nsatragno, line 5546: How does a relying party know if the client/authenticator support the new language tags?

index.bs
```diff
@@ -1767,6 +1771,9 @@ When this method is invoked, the user agent MUST execute the following algorithm
 
 1. If the length of <code>|pkOptions|.{{PublicKeyCredentialCreationOptions/user}}.{{PublicKeyCredentialUserEntity/id}}</code> is not between 1 and 64 bytes (inclusive) then throw a {{TypeError}}.
 
+1. If <code>|pkOptions|.{{PublicKeyCredentialCreationOptions/user}}.{{PublicKeyCredentialUserEntity/language}}</code> is present
```
nsatragno, line 1774: Should we require `language` and `direction` be present if the other one is? i.e. if `language` is present, `direction` cannot be undefined.

index.bs
```diff
@@ -240,6 +240,10 @@ spec: RFC8610; urlPrefix: https://tools.ietf.org/html/rfc8610
     type: dfn
         text: group sockets; url: section-3.9
 
+spec: String-Meta; urlPrefix: https://www.w3.org/TR/2024/NOTE-string-meta-20241017
```
nsatragno, line 243: General comment:

I understand this is the i18n wg proposal, but this approach feels strictly worse, since authenticators have to support it for browsers to take advantage of it.

index.bs
```diff
@@ -2926,6 +2938,8 @@ value and terminate the operation.
         required Base64URLString        id;
         required DOMString              name;
         required DOMString              displayName;
+        DOMString                       language;
```
nsatragno, line 2941: Was it decided whether we needed `language` or not?

nsatragno (changes requested):

MasterKale (commented):
A language map might be a better pattern here:

https://www.w3.org/TR/2024/NOTE-string-meta-20241017/#language-maps

For example, if we added a new "`user.localizedName`" property, that looked like this...

```
localizedName: {
  "en-US": {
    value: "...",
    dir: "ltr"
  }
}
```

...then processing rules could say clients should use a value in `localizedName` when present, otherwise fall back to `name`.

index.bs
emlun: It doesn't... but I think I see what you mean. When I wrote "SHOULD be used" I meant "[the **client**] should use them [when provided by the RP and supported by the authenticator]", but your reading was "[the **RP**] should use them [when supported by the client and authenticator]"? That's definitely an ambiguity I need to fix.

Anything else to do here after fixing that ambiguity?

index.bs
emlun: I don't think so, String-Meta specifically describes these as separate values. It doesn't look like either should require the other.

selfissued (approved):
Ready to go, as far as I'm concerned

index.bs
emlun: Hm. I see what you mean, but `TypeError` is not semantically correct for this error case. The `user.id` error ideally should have been a `SyntaxError` too, but that's probably too late to change now. [`SyntaxError` is also used by the PRF extension when base64 decoding fails](https://www.w3.org/TR/2025/WD-webauthn-3-20250127/#ref-for-dom-authenticationextensionsprfinputs-evalbycredential%E2%91%A2). So I'd prefer to keep this as `SyntaxError` rather than `TypeError`.

MasterKale:
From WG meeting @ May 28: We'll wait till next week's meeting when we can have a more direct discussion on the "best" path forward

index.bs
emlun: Does 5d2e401fc80c4f021d161abd23e1faafbed06768 resolve this ambiguity?

index.bs
MasterKale: `SyntaxError` will stick around for now

emlun:
Closing in favour of #2308, per https://github.com/w3c/webauthn/issues/1643#issuecomment-2985299304.

PR #2283, opened 2025-04-25
Account for new fully-specified ECDSA and Ed448 COSEAlgorithmIdentifiers

### LINKED ISSUE #2282: Update COSEAlgorithmIdentifier examples and recommendations to prefer fully-specified alg IDs
This is broken out from issue #2276:

## Proposed Changes for WebAuthn L3

Update non-normative examples, and update recommendations for RP to prefer the new values. These are not yet implemented by clients and authenticators, but will gracefully fall back to the legacy values until implemented.

- [§1.3.1. Registration](https://w3c.github.io/webauthn/#sctn-sample-registration): change example from using -7 (ES256) to -9 (ESP256) and from -8 (EdDSA) to -50 (Ed25519)

- [`pubKeyCredParams`](https://w3c.github.io/webauthn/#dom-publickeycredentialcreationoptions-pubkeycredparams): update algoritms recommended to RPS:

  >[Relying Parties](https://w3c.github.io/webauthn/#relying-party) that wish to support a wide range of [authenticators](https://w3c.github.io/webauthn/#authenticator) SHOULD include at least the following [COSEAlgorithmIdentifier](https://w3c.github.io/webauthn/#typedefdef-cosealgorithmidentifier) values:
  >
  >- -8 (Ed25519)
  >- -7 (ES256)
  >- -257 (RS256)
  >
  >Additional signature algorithms can be included as needed.

- [§5.8.5. Cryptographic Algorithm Identifier (typedef `COSEAlgorithmIdentifier`)](https://w3c.github.io/webauthn/#sctn-alg-identifier): update examples:
  >A [COSEAlgorithmIdentifier](https://w3c.github.io/webauthn/#typedefdef-cosealgorithmidentifier)’s value is a number identifying a cryptographic algorithm. The algorithm identifiers SHOULD be values registered in the IANA COSE Algorithms registry [[IANA-COSE-ALGS-REG]](https://w3c.github.io/webauthn/#biblio-iana-cose-algs-reg), for instance, -7 for "ES256" and -257 for "RS256".

- [§6.5.1.1. Examples of credentialPublicKey Values Encoded in COSE_Key Format](https://w3c.github.io/webauthn/#sctn-encoded-credPubKey-examples): add example(s) using fully-specified ID(s)

- [§6.5.5. Signature Formats for Packed Attestation, FIDO U2F Attestation, and Assertion Signatures](https://w3c.github.io/webauthn/#sctn-signature-attestation-types): add -9 (ESP256) in addition to -7 (ES256)

emlun:
Fixes #2282.

This: 

- Adds a normative recommendation against using the identifiers -9 (ESP256), -51 (ESP384), -52 (ESP512) and -19 (Ed25519), recommending to continue using -7 (ES256), -35 (ES384), -36 (ES512) and -8 (EdDSA) instead.
- Carries forward the restriction against compressed point format from the `ES*` algorithms to the new `ESP*` algorithms.
- Adds Ed448 test vectors.
- Changes the -8 (EdDSA) test vectors to using the seed `'packed.EdDSA'` and link anchor `#sctn-test-vectors-packed-eddsa` instead of `'packed.Ed25519'` and `#sctn-test-vectors-packed-ed25519`, in case we want to add test vectors with COSE alg -19 (Ed25519) in the future.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/pull/2283.html" title="Last updated on May 22, 2025, 3:40 PM UTC (472e8d0)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/2283/86e50a9...472e8d0.html" title="Last updated on May 22, 2025, 3:40 PM UTC (472e8d0)">Diff</a>


timcappalli (approved):

MasterKale (approved):

nadalin:
@selfissued can you review ?

selfissued (approved):

emlun:
@selfissued Thanks, this is resolved in meta-PR #2290 which in turn depends on #2289. I'll mark this PR as blocked waiting on #2289 so we can update the test vectors along with the generator script.

emlun:
@zacknewman Thanks, I had fixed this (thanks for pointing it out!) but apparently forgot to push it. Fixed now in 928c6684a87c0a1e1efc7c0506009fac3e9a1b76.

index.bs
```diff
@@ -3479,7 +3487,9 @@ optionally evidence of [=user consent=] to a specific transaction.
 
         [=[RPS]=] that wish to support a wide range of [=authenticators=] SHOULD include at least the following {{COSEAlgorithmIdentifier}} values:
 
-        * -8 (Ed25519)
+        * -50 (Ed25519)
+        * -8 (EdDSA)
+        * -9 (ESP256)
```
akshayku, line 3492: IANA is showing that it is unassigned as of now. -7 is for ECDSA with P-256 with SHA256.

index.bs
```diff
@@ -510,10 +510,18 @@ The sample code for generating and registering a new key follows:
       // This Relying Party will accept either an EdDSA, ES256 or RS256 credential, but
       // prefers an EdDSA credential.
       pubKeyCredParams: [
+        {
+          type: "public-key",
+          alg: -50 // "Ed25519" as registered in the IANA COSE Algorithms registry
```
akshayku, line 515: This is assigned to ML-DSA-87 in IANA https://www.iana.org/assignments/cose/cose.xhtml

index.bs
```diff
@@ -4322,12 +4332,15 @@ Note: The {{AuthenticatorTransport}} enumeration is deliberately not referenced,
...
     A {{COSEAlgorithmIdentifier}}'s value is a number identifying a cryptographic algorithm.
     The algorithm identifiers SHOULD be values registered in the IANA COSE Algorithms registry [[!IANA-COSE-ALGS-REG]],
-    for instance, <code>-7</code> for "ES256" and <code>-257</code> for "RS256".
+    for instance, <code>-9</code> for "ESP256" and <code>-257</code> for "RS256".
 
     The COSE algorithms registry leaves degrees of freedom to be specified by other parameters in a [=COSE key=]. In order to promote interoperability, this specification makes the following additional guarantees of [=credential public keys=]:
         1. Keys with algorithm ES256 (-7) MUST specify P-256 (1) as the [=crv=] parameter and MUST NOT use the compressed point form.
+        1. Keys with algorithm ESP256 (-9) MUST NOT use the compressed point form.
         1. Keys with algorithm ES384 (-35) MUST specify P-384 (2) as the [=crv=] parameter and MUST NOT use the compressed point form.
+        1. Keys with algorithm ESP384 (-48) MUST NOT use the compressed point form.
```
akshayku, line 4341: ECDSA with P384 is -35 I think.

index.bs
```diff
@@ -4322,12 +4332,15 @@ Note: The {{AuthenticatorTransport}} enumeration is deliberately not referenced,
...
         1. Keys with algorithm ES512 (-36) MUST specify P-521 (3) as the [=crv=] parameter and MUST NOT use the compressed point form.
+        1. Keys with algorithm ESP512 (-49) MUST NOT use the compressed point form.
```
akshayku, line 4343: ECDSA with P521 with SHA512 is -36

index.bs
emlun: -35 is ECDSA with SHA-384 and unspecified curve (this is why WebAuthn adds the restriction to the P-384 curve on line 4340). -48 is the new identifier that fully specifies ECDSA with SHA-384 and curve P-384.

emlun:
@selfissued As you can see in [Akshay's comments above](https://github.com/w3c/webauthn/pull/2283#pullrequestreview-2822828036), we discovered on today's WG call that some of the COSE algorithm identifiers in https://www.ietf.org/archive/id/draft-ietf-jose-fully-specified-algorithms-11.html have also been requested for assignment in https://www.ietf.org/archive/id/draft-ietf-cose-dilithium-06.html:

* -48 is requested for both ESP384 and ML-DSA-44
* -49 is requested for both ESP512 and ML-DSA-65
* -50 is requested for both Ed25519 and ML-DSA-87

Looks like the ML-DSA ones have already landed in the [COSE registry](https://www.iana.org/assignments/cose/cose.xhtml#algorithms). So I assume jose-fully-specified-algorithms will need to change these assignment requests to other values?

emlun:
#2289 is now merged, but I'll keep the "blocked" label until https://github.com/w3c/webauthn/pull/2283#issuecomment-2859975962 is resolved.

index.bs
akshayku: If they are logically equivalent, I would actually advice against using these new numbers by the RPs.

Older platforms and Older Authenticators do not understand these numbers and will fail when specified with these numbers.  

These new numbers do not add any value IMO vs specifying existing values and will confuse RPs and add support costs.

akshayku: For example, one such unupgradable security keys shows supporting these algos in their getInfo. [{"alg": -7, "type": "public-key"}, {"alg": -8, "type": "public-key"}, {"alg": -35, "type": "public-key"}]. Newer clients may be able to do the translation to older value from newer values, but not every user agent and OS is upgradable.

akshayku: Looked at earlier discussions in the issue. Fallback mechanism is easy to get it wrong. We have spent years to get these values supported across the ecosystem by every RP, user agent and authenticator.

akshayku: Also, let's assume that -7 and -9 are equivalent (ECC-P256/SHA256) and -35 and -1001 are equivalent (ECC-P384/SHA384). 

Let's say
- Authenticator-1 supports -7 (older value) and -1001 (newer value). 
- Authenticator-2 supports -7 (older value) and -35 (older value). 

Technically both Authenticators support (ECC-P256/SHA256) and (ECC-P384/SHA384)

RP wants to prioritize (ECC-P384/SHA384) first if supported and fallback to (ECC-P256/SHA256) if necessary.

RPs Algorithm preferred list of
- (-35, -1001, -7 -9) will result in both Authenticator-1 and Authenticator-2 picking up (ECC-P384/SHA384)
- (-35, -7, -1001, -9) will result in Authenticator-1 picking up (ECC-P384/SHA384) and Authenticator-2 picking up (ECC-P256/SHA256)

Having two different values with same meaning in a priority list can become tricky and RP needs to be careful. 

I am not seeing the benefit of using the newer values. I would recommend that WebAuthn specify that RPs should use older values as we have already reserved curves for those numbers in WebAuthn.

akshayku: Also, let's assume that -7 and -9 are equivalent (ECC-P256/SHA256) and -35 and -1001 (some random number for discussion sake) are equivalent (ECC-P384/SHA384). 

Let's say
- Authenticator-1 supports -7 and -1001. 
- Authenticator-2 supports -7 and -35. 

RPs Algorithm preferred list of (-35, -7, -1001, -9) will result in Authenticator-1 picking up (ECC-P256/SHA256) and Authenticator-2 picking up (ECC-P384/SHA384), although technically both support (ECC-P256/SHA256) and (ECC-P384/SHA384)

Having two different values with same meaning in a priority list can become tricky and RP needs to be careful. 

I am not seeing the benefit of using the newer values. I would recommend that WebAuthn specify that RPs should use older values as we have already reserved curves for those numbers in WebAuthn.

selfissued:
Thanks for catching the conflict with the new ML-DSA assignments.  I'll update Fully-Specified Algorithms today.

This is particularly timely, because Fully-Specified Algorithms is going to the IESG for approval on their "telechat" in 10 hours.

selfissued:
https://www.ietf.org/archive/id/draft-ietf-jose-fully-specified-algorithms-12.html updates the requested assignments to avoid the conflicts.  It also requests assignment of Ed25519 in the one-byte range, matching the size of EdDSA, per @akshayku's observation.

index.bs
emlun: Thanks for catching the conflict! Draft 12 changed the assignment request to -19: https://www.ietf.org/archive/id/draft-ietf-jose-fully-specified-algorithms-12.html#name-edwards-curve-digital-signa . Fixed in c129a055c23e97b01caf513aa130a48d3904afe2.

index.bs
emlun: Thanks for catching the conflict! [Draft 12](https://www.ietf.org/archive/id/draft-ietf-jose-fully-specified-algorithms-12.html#name-elliptic-curve-digital-sign) changed the assignment request to -51. Fixed in c129a055c23e97b01caf513aa130a48d3904afe2.

index.bs
emlun: Thanks for catching the conflict! [Draft 12](https://www.ietf.org/archive/id/draft-ietf-jose-fully-specified-algorithms-12.html#name-elliptic-curve-digital-sign) changed the assignment request to -52. Fixed in c129a055c23e97b01caf513aa130a48d3904afe2.

index.bs
emlun: I agree that there's little value in these particular identifiers (except the addition of Ed448) within the current context of WebAuthn, where we've explicitly patched the ambiguity in the polymorphic identifiers, but these new identifiers are defined for a wider context that covers more than just WebAuthn. In particular, they patch the gap between JOSE and COSE where, for example, the JOSE identifier `"ES256"` _does_ fully specify "ECDSA with SHA-256 on the P-256 curve" but the "equivalent" COSE identifier `-7` _does not_ (evidently to the WebAuthn WG's surprise earlier) specify the curve, only "ECDSA with SHA-256". For other COSE use cases, especially new ones, it makes a lot of sense to be able to fully describe the crypto suite with a single identifier. So the identifiers definitely have merit in the wider context outside WebAuthn.

So in WebAuthn we are faced with a choice, then: should we support these new identifiers, or not?

If we do not, we'll have to declare that our `COSEAlgorithmIdentifier` parameter only supports a subset of `COSEAlgorithmIdentifier` values, and continue updating our exclusion list as new identifiers are added to the registry. And it would need to be an exclusion list rather than an allow list, because with an allow list there'd be no point in referring to an external registry in the first place. I don't think this is a good approach. It's already unfortunate that we add the restrictions against compressed point formats, I don't think we should add even more constraints on our interop with COSE.

If we do support the new identifiers (which WebAuthn L1 and L2 do implicitly!), then I agree that it's unfortunate to have values that are semantically equivalent (in our context) but need to be treated as distinct in order to not confuse implementations. But I think that's a lesser problem. Authenticators are incentivized to support both identifiers in case an RP only requests one (including old RPs that only request the old identifier), and RPs are also incentivized to request both identifiers in case an authenticator only supports one (including existing authenticators that only support the old identifier). But as you say, the old values are likely to continue to dominate for the foreseeable future, so it is unlikely that the new values will disadvantage RPs or authenticators that have not or cannot update to support the new values.

To summarize: the new COSE identifiers apply to a wider context than WebAuthn, and I think that making WebAuthn align more with COSE is better than making WebAuthn diverge further from COSE.

selfissued:
FYI, the fully-specified algorithm numbers are now registered at https://www.iana.org/assignments/cose/cose.xhtml#algorithms.

selfissued:
@zacknewman the descriptions in the Fully-Specified Algorithms draft can still be updated at this point to be more prescriptive.  I'd be glad to do that.  Would these updated descriptions fit the bill, as far as you're concerned?

- Ed25519: EdDSA using Ed25519 curve, as described in Section 5.1 of RFC 8032
- Ed448: EdDSA using Ed448 curve, as described in Section 5.2 of RFC 8032

Of course, that's what was always meant, but it's fine to be more precise.

selfissued:
OK, I understand and agree with this point you're making:
> Similar descriptions for Ed448 follow. The key is to not phrase "Ed25519" in the context of a _curve_ but instead as the specific signature scheme which of course implies curve ed25519 in addition to 10 other things.

Here, it sounds to me like you are talking about key representations:
> Since we are here, it would also be nice for algorithms like ESP256 to require the uncompressed form since this would entirely eliminate the need for further refinement in WebAuthn for all of these new IDs.

By design, the cryptographic algorithms are independent of the key representations used.  Everything works no matter what key representation is used, provided the implementation understands and correctly interprets the key representation(s) used.  Therefore, if WebAuthn wants to make statements about key representations it needs to do so itself.  I would file an issue about that if you believe that something is missing from WebAuthn.

For example, the equivalence of key representations is discussed in another context at https://www.rfc-editor.org/rfc/rfc7638.html#section-3.5 :
>   Note that a key need not be in JWK format to create a JWK Thumbprint
>   of it.  The only prerequisites are that the JWK representation of the
>   key be defined and the party creating the JWK Thumbprint be in
>   possession of the necessary key material.  These are sufficient to
>   create the hash input from the JWK representation of the key, as
>   described in [Section 3.3](https://www.rfc-editor.org/rfc/rfc7638.html#section-3.3).

It may not be OK for WebAuthn, but for the cryptographic algorithm, it doesn't matter if the key is represented as a JWK, a COSE_Key, in a X.509 certificate, or in an XML data structure, provided the implementation understands and correctly uses it.

cc: @OR13

selfissued:
> Does Ed25519 only mean any EdDSA that uses curve ed25519 (e.g., Ed25519ph), or is the intent for it to only mean the Ed25519 scheme as defined in RFC 8032 § 5.1?

I'd already clarified that the intent is for these identifiers to refer to the algorithms as specified in RFC 8032.  I will update the Fully-Specified Algorithms draft to say that explicitly during the AUTH48 process.  In the meantime, WebAuthn can count on that.

Thanks again for bringing the possibly ambiguous interpretation of the existing language to our attention.  I will close that ambiguity as part of the process of publishing Fully-Specified Algorithms as an RFC.

emlun:
Thanks @selfissued!

@zacknewman as for the inline additional requirements for `ESP*`, I'm conflicted. I would prefer not to have the restriction to uncompressed representations, but since we have it for `ES*` I figure the "path of least surprise" is to have them for `ESP*` too. But you'll notice there's no additional requirements like that for `Ed25519` or `Ed448` since none is needed - the only explicit mentions of them are in recommendations, examples and test vectors.

With that I think this is ready to merge as far as I'm concerned, unless we want to debate the restriction to uncompressed `ESP*` key representations.

akshayku:
Sorry, but I have objections to the PR itself for existing algorithms (-7, -35, -36, -8). 

It has cons and is a lot of make work for lot of companies without realistically RP/Platform/Authenticator ever being able to deprecate these existing COSE end points. 

From Windows perspective, we can't make changes to older platforms. And it is not planning to make changes for newer code points. We don't see any benefit. 
From security key perspective, they have unupgradable security keys with existing algorithm code points burned in. 
From our Entra/MSA as an RP perspective, we cannot remove/deprecate  exisiting

I am not commenting about whether these newer code points are beneficial for other specs outside of WebAuthn. But I don't see any benefit out of these new code points. 

I am fine adding new algorithm for Ed448 if you care because that doesn't exist today from WebAuthn perspective.

OR13:
I understand the desire not to use the new algorithms if the existing ones are burned into devices.

If I understand correctly, the existing webauthn behavior is that in cases where an algorithm worked with multiple curves, webauthn chose a curve, and does not support all the curves for that algorithm, is that correct?

In other words, in webauthn:

-7 means ECDSA with SHA256 and P-256 ... this now has an official name ESP256
-8 means EdDSA with Ed25519 ... this now has an official name Ed25519
...

Just for my own clarity are there any hardware devices out there that use -7 with a curve other than P-256?

In other words, does webauthn make use of the "feature" that -7 works with P-384 and P-521, or does it just ignore that possibility, and assume that -7 is always with P-256? (and my webauthn, I really mean devices that can't be updated, as opposed to the spec itself)

akshayku:
> If I understand correctly, the existing webauthn behavior is that in cases where an algorithm worked with multiple curves, webauthn chose a curve, and does not support all the curves for that algorithm, is that correct?

Yes

> Just for my own clarity are there any hardware devices out there that use -7 with a curve other than P-256?

No. AFAIK from over the years. 

> In other words, does webauthn make use of the "feature" that -7 works with P-384 and P-521, or does it just ignore that possibility, and assume that -7 is always with P-256? (and my webauthn, I really mean devices that can't be updated, as opposed to the spec itself)

We always assume that -7 is ECDSA with P-256. Similarly with others (-8, -35, -36). We have put below descriptions in the spec. 

> Keys with algorithm ES256 (-7) MUST specify P-256 (1) as the [crv](https://tools.ietf.org/html/rfc9053#name-double-coordinate-curves) parameter and MUST NOT use the compressed point form.

>Keys with algorithm ES384 (-35) MUST specify P-384 (2) as the [crv](https://tools.ietf.org/html/rfc9053#name-double-coordinate-curves) parameter and MUST NOT use the compressed point form.

>Keys with algorithm ES512 (-36) MUST specify P-521 (3) as the [crv](https://tools.ietf.org/html/rfc9053#name-double-coordinate-curves) parameter and MUST NOT use the compressed point form.

>Keys with algorithm EdDSA (-8) MUST specify Ed25519 (6) as the [crv](https://tools.ietf.org/html/rfc9053#name-double-coordinate-curves) parameter. (These always use a compressed form in COSE.)

akshayku:
> @akshayku Ok, but I don't understand what part of this PR would mean more work for any implementations? We are not replacing the old IDs, only adding the new ones to the set of recommended arguments.

Adding it to the WebAuthn means that we are endorsing those new algorithms for WebAuthn. Some algorithms just being registered as COSEAlgorithmIdentifier does not necessarily means that it works in WebAuthn across browser/OS/Authenticator. There are many more COSEAlgorithmIdentifier that does not work for WebAuthn. And I think, Realisticly, the current state is that only algorithms mentioned in the webAuthn spec works in reality across the ecosystem. Actually that also is not true across the ecosystem for -8. 

> 
> * Current implementations - of authenticators, of clients and of RPs - that only support/request the old IDs will continue to work with the old identifiers.

Yes. 

> * New implementations that support/request both IDs will work with both.

Yes, but they don't have to. There is no need, and this is extra work for Authenticators for no benefit. And new code means new potential bugs. And this is again guidance for the authenticators. May be there is some authenticator vendor who has not read all this conversation to only implement newer algorithms. And suddenly their release devices can't be fixed. 

> * New RP implementations that request only the new IDs will not work with current authenticator implementations that only support the old IDs, so they are incentivized to also request the old IDs.

RPs will get it wrong. 

Let's have one such example. Let's say that these newer Ecdsa algorithms gets supported on certain combination of browser/OS and RP didn't test all the platforms/Browser/OS combination. Just because they only tested major OS/Browser combination. I have seen this happen over the years. Also, we have already established that these new algorithms will not be supported on all authenticators/platforms. So, these RPs will have support costs for these unsupported combination situations down the road. Suddenly there is a situation that existing security keys for example does not work. Is RP staffed to add those? May be someone from RP says it is not the worth given their large population may not have security keys (I have seen this happen too). Then the user who has gone out of its way to buy such security key, will be left stranded. 

You can add more guidance for the RPs. But even with current guidance, people get it wrong. Unless there was any issue with security/incompatibility/privacy, I don't think we should just add new definitions that can cause those incompatibilities.

Again, what is the need to introduce these new algorithms in WebAuthn? Agreed that they are not fully specified yet in COSE registry. But we have already made it clear in our spec about what those algorithms means. 

>   
>   * Consequently, existing authenticator implementations do not need to update (even in new versions) and can continue supporting only the old IDs if they wish.

Provided that RP does the right thing. We shouldn't even open the possibility for RPs to get it wrong, IMO. There is no benefit. For example, if an RP wants ECDSA with ECC P-256 with SHA256, they use -7 today. There is no need to redefine it to something else. 

> * New authenticator implementations that support only the new IDs will not work with current RP implementations that only request the old IDs, so they are incentivized to also support the old IDs.

Again, this is extra work for Authenticator for no benefit. All the cons of these proposals are because of various combinations of browsers/OS/Authenticators. Some are upgradable, some are not. And I don't think this is even a problem.  

>   
>   * Consequently, existing RP implementations do not need to update (even in new versions) and can continue requesting only the old IDs if they wish.

Yes. That is the ideal state. Newer RPs shouldn't be thinking different from existing implementation that just works. We don't know how  

> 
> So I agree, we're not going to realistically _eliminate_ the old IDs short of a complete ecosystem-wide migration to PQC/etc. But eliminating the old IDs was never a goal either, so I don't see how that is a problem?
> 
> Note also that the new IDs will be available for RPs to request and for authenticators to implement _regardless_ of whether we make any change to WebAuthn. The `COSEAlgorithmIdentifier`s are registered, and WebAuthn L1 and L2 allow any `COSEAlgorithmIdentifier` to be used. WebAuthn L3 will too unless we explicitly forbid these specific values ([which would in my opinion be worse](https://github.com/w3c/webauthn/pull/2283#discussion_r2079594682)). All this PR does is inform RPs of this new reality: that for maximum compatibility they should request both the old and new IDs (well ok, it does also carry the format restriction on from `ES*` to `ESP*`, but that is not what introduces the new alg IDs).

Overall, there is no benefit of adding these new algorithms for existing algorithms defined in WebAuthn spec.

emlun:
@akshayku 

> [...] There is no need, and this is extra work for Authenticators for no benefit. And new code means new potential bugs. And this is again guidance for the authenticators. [...]

I was about to say there is no change in guidance for authenticators here, but on second thought I guess the changes to [§6.5.1.1. Examples of credentialPublicKey Values Encoded in COSE_Key Format](https://pr-preview.s3.amazonaws.com/w3c/webauthn/pull/2283.html#sctn-encoded-credPubKey-examples) do qualify, as do the new/updated test vectors. Good point!

> Again, what is the need to introduce these new algorithms in WebAuthn? [...]

Without it, authenticators that choose to implement support for ESP256 (-9) would be free to emit ESP256 COSE keys in compressed format since that is not forbidden for ESP256. Would that be a problem? RPs likely won't expect the compressed format and would have to update their code to support those authenticators that use the compressed format.

On the other hand I argued earlier that some incompatibility should be expected with new features, so maybe that's fine. Honestly I _would_ prefer to drop the restriction against compressed format for `ESP*`, I just figured we would cause fewer problems (i.e., less work for RPs) continuing that restriction for `ESP*` than we would by having it for `ES*` but not for `ESP*`.


@zacknewman 

> Another way to view this is now with these new IDs there are two equivalent IDs, from the perspective of WebAuthn, so we arbitrarily pick one (i.e., the legacy ID) to recommend.

No, not either one - we should _always_ recommend the legacy ID, the only question is whether we should _also_ recommend the new one. I believe there is no harm in recommending the new ID _and_ the old ID, and little or no harm in continuing to recommend only the old ID, but there certainly is harm in recommending the new ID _instead of_ the old ID.

So okay, I guess "little or no harm in continuing to recommend only the old ID" may be a signal that we should do just that.


> [...] about `getPublicKey`. Currently user agents are _required_ to return SPKI data from that function when the corresponding ID is EdDSA, ES256, or RS256. Presumably we add that user agents are required to support the new IDs as well?

We're not adding that requirement in L3. `getPublicKey()` is still required to support EdDSA (-8), ES256 (-7) and RS256 (-257), but is _not_ required to support Ed25519 (-19) or ESP256 (-9). Ed25519 (-19) and ESP256 (-9) should be "supported" in `pubKeyCredParams` just like they are today - in that the client doesn't throw an exception if you pass `pubKeyCredParams: [{ alg: -9, type: "public-key" }]`), but not necessarily in `getPublicKey()`. This is what's defined in the current L3 draft (without this PR) and I do not intend to change that in L3 (nor necessarily in L4+ if there are objections to it).

---

Okay, I've been convinced that we should at least reduce the scope of this change to keep the old IDs the "preferred" ones in examples and test vectors. I've updated the PR with that (I also changed the Ed25519 test vector to use the seed `'packed.EdDSA'` instead of `'packed.Ed25519'` since it is using the COSE identifier `EdDSA` (-8), in case we add an `Ed25519` (-19) test vector in the future).

I could go either way on recommending that RPs request the new IDs (_in addition_ to the respective old ones) in `pubKeyCredParams`. I've kept it in the PR for now. I think we should at the very least have some note saying if an RP requests one of the new IDs they really SHOULD also request the old one.

For now I've also kept the restriction against compressed form for `ESP*` public keys, but I'm happy to drop that too if we think it's a good idea for both compressed and uncompressed formats to coexist (but only with the new `ESP*` IDs, not with the old `ES*` IDs).

index.bs
emlun: -9 and -19 are registered now: https://www.iana.org/assignments/cose/cose.xhtml#algorithms

selfissued (approved):
This looks ready to merge.

emlun:
> I will say that removing the restriction for `ESP*` IDs to allow them to be encoded as octet-key pairs (OKPs) [...]

COSE does not allow encoding `ESP*` keys with the OKP type. ECDSA public keys are required to use the EC2 key type; see [RFC 9053 §7.1. Elliptic Curve Keys](https://www.rfc-editor.org/rfc/rfc9053.html#name-elliptic-curve-keys) and [RFC 9053 §2.1. ECDSA](https://www.rfc-editor.org/rfc/rfc9053.html#section-2.1):

>When using a COSE key for this algorithm , the following checks are made:
>- The "kty" field MUST be present, and it MUST be "EC2".
>[...]

Only Edwards curve keys (currently) are allowed (and indeed required) to use OKP, and yes, [RFC 8032](https://www.rfc-editor.org/rfc/rfc8032.html) defines a single canonical encoding for EdDSA keys.

Rather, it's the EC2 key type that has two variants: compressed or uncompressed y coordinate.

emlun:
I've now also reverted adding the new algs to recommendations and examples, per @akshayku's argument that if RPs should continue using the old values anyway, we should not recommend using the new ones.

I've still kept the restriction against point compression for `ESP*`, for consistency with `ES*`.

index.bs
```diff
@@ -4326,8 +4326,11 @@ Note: The {{AuthenticatorTransport}} enumeration is deliberately not referenced,
 
     The COSE algorithms registry leaves degrees of freedom to be specified by other parameters in a [=COSE key=]. In order to promote interoperability, this specification makes the following additional guarantees of [=credential public keys=]:
         1. Keys with algorithm ES256 (-7) MUST specify P-256 (1) as the [=crv=] parameter and MUST NOT use the compressed point form.
+        1. Keys with algorithm ESP256 (-9) MUST NOT use the compressed point form.
```
akshayku, line 4329: This kind of says that WebAuthn support these newer algorithms. Can we put somewhere in the spec that RPs should use older values instead of new ones? Or atleast should always specify older values first for backwards compatibility.

selfissued: > This kind of says that WebAuthn support these newer algorithms.

The intent, I think, is to constrain how the new algorithms are used within WebAuthn.

> Can we put somewhere in the spec that RPs should use older values instead of new ones? Or atleast should always specify older values first for backwards compatibility.

I'm OK with us saying that the older algorithms that we constrained the meanings of in WebAuthn continue to have those meanings and continue to be usable.  And that to promote interoperability, where old algorithms exist that we constrained the meanings of, that implementations are encouraged to keep using them with those meanings to enhance interoperability with implementations that were created before the new algorithms came into existence.

emlun: @akshayku sure, I'll add a note.

> for RPs that do support the new `ESP*` IDs they are required to represent the keys in the uncompressed form

This requirement applies more to authenticators than RPs. Authenticators are the ones that emit public keys, and RPs are the ones that consume them. WebAuthn does not care how RPs represent keys internally.

> why don't we require the uncompressed form for COSE IDs like `ES256K`?

Good question! Probably because there weren't (and still aren't?) many authenticators that support `ES256K`, so we haven't observed interop problems with its key encoding. (Much to @akshayku's point: even if WebAuthn _theoretically_ supports any applicable COSE alg, in practice only a few are commonly used.) I wouldn't add a restriction for `ES256K` at this point.

akshayku (approved):

selfissued (approved):

PR #2289, opened 2025-05-06
Add scripts used to generate test vectors

emlun:
This replaces PR #2281.

#2209 added test vectors, and I said that:

>I plan to also share the code used to generate these, but I needed to patch some of the libraries I used, so I need to resolve that first.

Most of this upstreaming is now resolved, so this PR adds a cleaned-up version of the scripts used to generate test vectors.

This also revealed a mistake in the documentation of how the `salt_enc` IV values were generated: the were generated with seed suffixes `0x06` and `0x07`, not `0x07` and `0x08`. Commit 946490376a44df546c6745f2d17d5dbe5aa3089c first fixes the description to match the test vectors, and commit 87cab44efad607ee370416f4cf517eb77f28d632 then reverts to the original description and instead updates the test vectors to match.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/pull/2289.html" title="Last updated on May 6, 2025, 2:37 PM UTC (00542a6)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/2289/43b55de...00542a6.html" title="Last updated on May 6, 2025, 2:37 PM UTC (00542a6)">Diff</a>

PR #2290, opened 2025-05-06
Add and prefer fully-specified COSE algorithm identifers in test vectors

emlun:
(This meta-PR would merge into #2283, not directly to main)

As requested in https://github.com/w3c/webauthn/pull/2283#issuecomment-2845669597, this updates the test vectors section to primarily use fully-specified `COSEAlgorithmIdentifier`s (but still include examples of polymorphic ones). This also adds an Ed448 case since that was not previously expressible.

This relies on an unofficial version of [`python-fido2`](https://pypi.org/project/fido2/) since its latest official version (`1.2.0`) does not include support for the new fully-specified `COSEAlgorithmIdentifier`s.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/pull/2290.html" title="Last updated on May 7, 2025, 7:40 PM UTC (b4d412c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/2290/d51a7ad...b4d412c.html" title="Last updated on May 7, 2025, 7:40 PM UTC (b4d412c)">Diff</a>


selfissued (approved):

emlun:
Thanks @zacknewman, good catch! It certainly makes sense that the self attestation case should use the same `alg` for the attestation object as in the credential key. But note that in general, the `alg` attribute of the attestation (when present) may be different from the `alg` of the credential public key. For example, all attestation statements in these test vectors use ESP256 (previously ES256) regardless of the credential key algorithm (this is mostly because EC keys are much easier to generate than RSA keys).

emlun:
This is a meta-PR that doesn't merge directly to main, so now with PR #2289 merged I'll merge this into PR #2283.

PR #2304, opened 2025-06-16
Fix COSEAlgorithmIdentifier in section 5.4

lennartkloock:
This PR fixes the description for `pubKeyCredParams` in section 5.4. The number `-8` does not identify the `Ed25519` algorithm but `EdDSA` (see the [IANA document](https://www.iana.org/assignments/cose/cose.xhtml#algorithms)).
`Ed25519` has number -19.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -3479,7 +3479,7 @@ optionally evidence of [=user consent=] to a specific transaction.
 
         [=[RPS]=] that wish to support a wide range of [=authenticators=] SHOULD include at least the following {{COSEAlgorithmIdentifier}} values:
 
-        * -8 (Ed25519)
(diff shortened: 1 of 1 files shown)


emlun (approved):
Thanks! At first I'd say this is correct as is, because -8 is specialized to mean Ed25519 within the WebAuthn context (see [§5.8.5. Cryptographic Algorithm Identifier (typedef COSEAlgorithmIdentifier)](https://www.w3.org/TR/2025/WD-webauthn-3-20250127/#sctn-alg-identifier)), but I agree that it's inconsistent when we use the labels "-8 (EdDSA)" and "-19 (Ed25519)" further down in the same paragraph.

PR #2315, opened 2025-07-15
Clarify that DER lengths vary with INTEGER magnitude and curve size

### LINKED ISSUE #2314: Clarify DER encoding rules apply to signature format encoding in section 6.5.5
## Background
In section 6.5.5 "Signature Formats for Packed Attestation, FIDO U2F Attestation, and Assertion Signatures", it is stated that "the sig value MUST be encoded as an ASN.1 DER Ecdsa-Sig-Value".

However, the example states "02 20      ; INTEGER (32 Bytes)" - suggesting that the encoding would always lead to 32 bytes.

Some real world implementations don't follow DER encoding rules to "ensure that an integer value is always encoded in the smallest possible number of octets".

## Proposed Change
Add a (non-normative) comment to the example that the ASN.1 DER encoding could lead to a different number of encoded bytes.

emlun:
Closes #2314.

The new example is copied from the ["ES256 Credential with very long credential ID" test vector](https://w3c.github.io/webauthn/#sctn-test-vectors-none-es256-long-credential-id) whose [signature](https://github.com/w3c/webauthn/blob/96cc62b627f0b17114931e1bbc546f376f372c5d/index.bs#L9387) happens to have components of different lengths.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/pull/2315.html" title="Last updated on Jul 15, 2025, 4:54 PM UTC (81514c7)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/2315/96cc62b...81514c7.html" title="Last updated on Jul 15, 2025, 4:54 PM UTC (81514c7)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -5847,13 +5847,14 @@ the [=authenticator=] MUST:
         ```
         Example:
         Note: Encoding lengths vary with INTEGER magnitude and curve size.
-        30 44                                ; SEQUENCE (68 Bytes)
+        30 45                                ; SEQUENCE (69 Bytes)
             02 20                            ; INTEGER (32 Bytes)
-            |  3d 46 28 7b 8c 6e 8c 8c  26 1c 1b 88 f2 73 b0 9a
-            |  32 a6 cf 28 09 fd 6e 30  d5 a7 9f 26 37 00 8f 54
-            02 20                            ; INTEGER (32 Bytes)
-            |  4e 72 23 6e a3 90 a9 a1  7b cf 5f 7a 09 d6 3a b2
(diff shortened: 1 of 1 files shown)


akshayku (approved):

timcappalli (approved):

rlin1:
Could you also make one value *shorter* than 32 bytes? Because this is the other edge-case.

emlun:
I considered it, but figured it's good to represent the common case too. But fair enough, done in 488684b04188d4ac650285f3598f1752f8bf25b2.

rlin1 (approved):

PR #2372, opened 2025-12-10
Forbid authenticator data from containing cleartext PRF outputs

### LINKED ISSUE #2359: Forbid authenticator data from containing plaintext PRF outputs
There are password managers (e.g., Bitwarden) where one of the primary security guarantees is that the vault key is _never_ sent to the server. When unknown CBOR key-value pairs are ignored in the authenticator data and an authenticator is used that includes the PRF outputs in the authenticator data, one cannot simply scrub out such key-value pairs since the authenticator data is part of the signature. This causes the vault key to be sent to the server violating one of the security guarantees.

I propose that the spec should explicitly forbid authenticators from including the plaintext PRF outputs in the authenticator data. It's less important for CTAP-based authenticators to have this requirement since `hmac-secret` and `hmac-secret-mc` are encrypted; however nothing forbids such an authenticator to embed some other key-value pair and still be FIDO and WebAuthn compliant. At the very least though, non-CTAP-based authenticators should be required to not send the plaintext PRF outputs instead of simply saying:

> **Authenticator extension input / output**
>
> [This extension](https://w3c.github.io/webauthn/#prf) is abstract over the authenticator implementation, using either the [[FIDO-CTAP]](https://w3c.github.io/webauthn/#biblio-fido-ctap) `hmac-secret` extension or an unspecified interface for communication between the client and authenticator. It thus does not specify a CBOR interface for inputs and outputs.

I think it should say something like (emphasis added):

> **Authenticator extension input / output**
>
> [This extension](https://w3c.github.io/webauthn/#prf) is abstract over the authenticator implementation, using either the [[FIDO-CTAP]](https://w3c.github.io/webauthn/#biblio-fido-ctap) `hmac-secret` extension or an unspecified interface for communication between the client and authenticator. It thus does not specify a CBOR interface for inputs and outputs _except that the CBOR outputs must not contain the plaintext PRF outputs._

Recently Google Password Manager (GPM) [was changed](https://issues.chromium.org/issues/430804950) since it was sending the following CBOR map in the authenticator data when PRF was requested:

```js
{
  "prf": {
    "eval": {
      "first": <first input passed to PRF>,
      "second": <second input passed to PRF>
    },
    "enabled": true
  }
}
```

This was changed so that it no longer embedded `prf` into the authenticator data. In this case it wasn't a problem since the data that was being sent were the PRF _inputs_ (and `AuthenticationExtensionsPRFOutputs.enabled`); but even if it were sending the PRF outputs, it wouldn't violate the spec. Should be noted that the spec explicitly mentions how PRF can be used for "encryption keys" but only states the importance of removing the PRF outputs from the client extensions:

> For some use cases, for example if PRF outputs are used to derive encryption keys to use only on the client side, it may be necessary to omit this [results](https://w3c.github.io/webauthn/#dom-authenticationextensionsprfoutputs-results) output if the [PublicKeyCredential](https://w3c.github.io/webauthn/#publickeycredential) is sent to a remote server, for example to perform the procedures in [§ 7 WebAuthn Relying Party Operations](https://w3c.github.io/webauthn/#sctn-rp-operations). Note in particular that the [RegistrationResponseJSON](https://w3c.github.io/webauthn/#dictdef-registrationresponsejson) and [AuthenticationResponseJSON](https://w3c.github.io/webauthn/#dictdef-authenticationresponsejson) returned by [PublicKeyCredential](https://w3c.github.io/webauthn/#publickeycredential).[toJSON()](https://w3c.github.io/webauthn/#dom-publickeycredential-tojson) will include this [results](https://w3c.github.io/webauthn/#dom-authenticationextensionsprfoutputs-results) output if present.

Without the spec forbidding plaintext PRF outputs from being sent in the authenticator data, password managers can't realistically use PRF for vault encryption unless they only allow PRF for CTAP-based authenticators or use attestation and enforce that only certain authenticators are used that have been tested to not include such data in the authenticator data.

emlun:
Closes #2359.

- ~~[ ] Modified Web platform tests ([link](https://github.com/web-platform-tests/wpt/))~~ N/A - abstract requirement

Implementation commitment:

- ~~[ ] WebKit ([link to issue](https://bugs.webkit.org/))~~ N/A - abstract requirement
- ~~[ ] Chromium ([link to issue](https://issues.chromium.org/issues/new?component=1456855&template=0))~~ N/A - abstract requirement
- ~~[ ] Gecko ([link to issue](https://bugzilla.mozilla.org/home))~~ N/A - abstract requirement

Documentation and checks

- [x] Affects privacy
- [x] Affects security
- ~~[ ] Updated explainer ([link](https://github.com/w3c/webauthn/wiki))~~ N/A - abstract requirement

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/pull/2372.html" title="Last updated on Dec 10, 2025, 8:52 PM UTC (5481e3b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/2372/33f5b27...5481e3b.html" title="Last updated on Dec 10, 2025, 8:52 PM UTC (5481e3b)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -7538,7 +7538,14 @@ but also cannot be deleted by the client since the [=authenticator data=] is sig
 : Authenticator extension input / output
 :: [=prf|This extension=] is abstract over the authenticator implementation,
     using either the [[FIDO-CTAP]] `hmac-secret` extension or an unspecified interface for communication between the client and authenticator.
-    It thus does not specify a CBOR interface for inputs and outputs.
+    It thus does not specify a CBOR interface for inputs and outputs, except for the following requirement:
+
+    - [=Authenticator extension outputs=] MUST NOT contain cleartext PRF outputs.
+
+        Note: This is because the [=authenticator data=] is signed,
+        so authenticator extension outputs cannot be omitted in case the {{PublicKeyCredential}} needs to be sent to the [=[RP]=] server.
+        This is important for use cases where PRF outputs should remain private to the client side,
+        such as using PRF outputs to derive encryption keys.
 
 : Authenticator extension processing
 :: [=Authenticators=] that support the [[FIDO-CTAP]] `hmac-secret` extension implement authenticator processing as defined in that extension.


arnar (approved):
Looks great to me, thank you!

zacknewman (approved):
LGTM

harshlal028 (approved):

timcappalli (approved):

PR #2196, opened 2024-11-13
Make NOTEs non-normative

emlun:
Closes #1979.

This changes some NOTEs into normative text, and "promotes" some occurrences of "must" and "should" to their RFC 8174-enforced uppercase forms; thus formally speaking this introduces new normative language to the spec. But I have done my best to preserve what I interpret as the spirit of the existing language, so I think one can argue that these still count as "just" editorial changes.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/pull/2196.html" title="Last updated on Jan 8, 2025, 8:31 PM UTC (b6fde37)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/2196/b36a79a...b6fde37.html" title="Last updated on Jan 8, 2025, 8:31 PM UTC (b6fde37)">Diff</a>

PR #2375, opened 2026-01-05
remoteClientDataJSON Extension

### LINKED ISSUE #1577: Support for remote desktops

akshayku:
Closes #1577

Use Case:
Remote desktop web applications let users interact with the (native) desktop environment of a remote host from a local client (web) application.

A user may want to use a WebAuthn authenticator connected to their local client in order to sign into a website in a browser running on a remote host.

This is based on https://github.com/w3c/webauthn/wiki/Explainer:-Remote-Desktop-Support which is already present in some form in chromium code base. 

Based on feedback and use case scenario, this proposal solves the double clientData calculation issue and returns remoteDesktopClientOrigin to the RP as part of client/Authenticator extension.

Implementation commitment:

Documentation and checks

- [ ] Affects privacy
- [ ] Affects security
- [ ] Explainer ([link](https://github.com/w3c/webauthn/wiki/Explainer:-Remote-Desktop-Support))

***

### :boom: Error: 422 Unprocessable Entity :boom: ###

[PR Preview](https://github.com/tobie/pr-preview#pr-preview) failed to build. _(Last tried on Apr 24, 2026, 3:59 PM UTC)_.

<details>
<summary>More</summary>

:rotating_light: [Spec Generator](https://www.w3.org/publications/spec-generator/) - Spec Generator is the web service used to build bikeshed/ReSpec specs

:link: [Related URL](https://www.w3.org/publications/spec-generator/?type=bikeshed-spec&output=html&url=https%3A%2F%2Fraw.githubusercontent.com%2Fw3c%2Fwebauthn%2F52a5b1156bba3012711d9b1cf2d59e784033fe43%2Findex.bs&force=1&md-warning=not%20ready)

**Error output:**

```json
[
    {
        "lineNum": "2063:21",
        "messageType": "fatal",
        "text": "Tried to auto-close a <li>, but there were unclosed elements remaining before the nearest matching start tag (at 1919:5).\nOpen tags: <div> at 1697:1, <li> at 1919:5, <dl> at 1922:9"
    },
    {
        "lineNum": "8673",
        "messageType": "warning",
        "text": "The heading 'Well-Known URI Registration' needs a manually-specified ID."
    },
    {
        "lineNum": "8493:5",
        "messageType": "lint",
        "text": "Unexported dfn that's not referenced locally - did you mean to export it?\n<dfn bs-line-number=\"8493:5\" data-dfn-type=\"dfn\" id=\"set-user-verified-parameters\" data-lt=\"Set User Verified Parameters\" data-noexport=\"by-default\" class=\"dfn-paneled\">Set User Verified Parameters</dfn>"
    },
    {
        "lineNum": null,
        "messageType": "failure",
        "text": "Did not generate, due to errors exceeding the allowed error level."
    }
]
```

_This seems to be an issue with the [Spec Generator](https://www.w3.org/publications/spec-generator/) service. PR Preview doesn't manage this service and so has no control over it. If you've identified an issue with it, you can [report the issue to the maintainers of Spec Generator](https://github.com/w3c/spec-generator/issues/new) directly. Please be courteous. Thank you!_

_If you don't have enough information above to solve the error by yourself or if the issue doesn't seem related to Spec Generator, you can [file an issue with PR Preview](https://github.com/tobie/pr-preview/issues/new?title=Unidentified%20Error&body=See%20w3c/webauthn%232375.)._

</details>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -8779,6 +8779,9 @@ possible for [=[RPS]=] to trust any further [=attestation statements=] from the 
 
 See also the related security consideration for [=[RPS]=] in [[#sctn-revoked-attestation-certificates]].
 
+
+## Security considerations for [=clients=] ## {#sctn-security-considerations-client}
+
 ### The remoteClientDataJSON Extension ### {#sctn-remote-clientdatajson-security}
 
 The [=remoteClientDataJSON=] extension enables a [=Remote desktop web client=] to proxy
--- a/index.bs
+++ b/index.bs
@@ -1,4 +1,4 @@
-﻿<!--   ***********  Web Authentication - Level 3 - Spec source file ***********
+<!--   ***********  Web Authentication - Level 3 - Spec source file ***********
 Notes:
 * the h1 tag is spec title that is rendered within the document window. Wrapping
   may be controlled with break tags.  The spec 'Level' value is not appended.
@@ -1864,23 +1864,27 @@ a numbered step. If outdented, it (today) is rendered as a bullet in the midst o
         1. [=map/Set=] |authenticatorExtensions|[|extensionId|] to the [=base64url encoding=] of |authenticatorExtensionInput|.
     </li>
 
-1. Let |collectedClientData| be a new {{CollectedClientData}} instance whose fields are:
-    : {{CollectedClientData/type}}
-    :: The string "webauthn.create".
-    : {{CollectedClientData/challenge}}
-    :: The [=base64url encoding=] of |pkOptions|.{{PublicKeyCredentialCreationOptions/challenge}}.
-    : {{CollectedClientData/origin}}
-    :: The [=ascii serialization of an origin|serialization of=] |callerOrigin|.
-    : {{CollectedClientData/crossOrigin}}
-    :: The inverse of the value of the
-        {{PublicKeyCredential/[CREATE-METHOD]/sameOriginWithAncestors}}
-        argument passed to this [=internal method=].
-    : {{CollectedClientData/topOrigin}}
-    :: The [=ascii serialization of an origin|serialization of=] |callerOrigin|'s [=top-level origin=] if
-        the {{PublicKeyCredential/[CREATE-METHOD]/sameOriginWithAncestors}}
-        argument passed to this [=internal method=] is [FALSE], else `undefined`.
+    <li id="CreateCred-ConstructCollectedClientData">
+        Let |collectedClientData| be a new {{CollectedClientData}} instance whose fields are:
+            : {{CollectedClientData/type}}
+            :: The string "webauthn.create".
+            : {{CollectedClientData/challenge}}
+            :: The [=base64url encoding=] of |pkOptions|.{{PublicKeyCredentialCreationOptions/challenge}}.
+            : {{CollectedClientData/origin}}
+            :: The [=ascii serialization of an origin|serialization of=] |callerOrigin|.
+            : {{CollectedClientData/crossOrigin}}
+            :: The inverse of the value of the
+                {{PublicKeyCredential/[CREATE-METHOD]/sameOriginWithAncestors}}
+                argument passed to this [=internal method=].
+            : {{CollectedClientData/topOrigin}}
+            :: The [=ascii serialization of an origin|serialization of=] |callerOrigin|'s [=top-level origin=] if
+                the {{PublicKeyCredential/[CREATE-METHOD]/sameOriginWithAncestors}}
+                argument passed to this [=internal method=] is [FALSE], else `undefined`.
+    </li>
 
-1. Let |clientDataJSON| be the [=JSON-compatible serialization of client data=] constructed from |collectedClientData|.
+    <li id="CreateCred-SerializeClientDataJSON">
+        Let |clientDataJSON| be the [=JSON-compatible serialization of client data=] constructed from |collectedClientData|.
+    </li>
 
 1. Let |clientDataHash| be the [=hash of the serialized client data=] represented by |clientDataJSON|.
 
@@ -2408,23 +2412,27 @@ When this method is invoked, the user agent MUST execute the following algorithm
 
     1. [=map/Set=] |authenticatorExtensions|[|extensionId|] to the [=base64url encoding=] of |authenticatorExtensionInput|.
 
-1. Let |collectedClientData| be a new {{CollectedClientData}} instance whose fields are:
-    : {{CollectedClientData/type}}
-    :: The string "webauthn.get".
-    : {{CollectedClientData/challenge}}
-    :: The [=base64url encoding=] of |pkOptions|.{{PublicKeyCredentialRequestOptions/challenge}}
-    : {{CollectedClientData/origin}}
-    :: The [=ascii serialization of an origin|serialization of=] |callerOrigin|.
-    : {{CollectedClientData/crossOrigin}}
-    :: The inverse of the value of the
-        {{PublicKeyCredential/[DISCOVER-METHOD]/sameOriginWithAncestors}}
-        argument passed to this [=internal method=].
-    : {{CollectedClientData/topOrigin}}
-    :: The [=ascii serialization of an origin|serialization of=] |callerOrigin|'s [=top-level origin=] if
-        the {{PublicKeyCredential/[DISCOVER-METHOD]/sameOriginWithAncestors}}
-        argument passed to this [=internal method=] is [FALSE], else `undefined`.
+    <li id="GetAssn-ConstructCollectedClientData">
+        Let |collectedClientData| be a new {{CollectedClientData}} instance whose fields are:
+            : {{CollectedClientData/type}}
+            :: The string "webauthn.get".
+            : {{CollectedClientData/challenge}}
+            :: The [=base64url encoding=] of |pkOptions|.{{PublicKeyCredentialRequestOptions/challenge}}
+            : {{CollectedClientData/origin}}
+            :: The [=ascii serialization of an origin|serialization of=] |callerOrigin|.
+            : {{CollectedClientData/crossOrigin}}
+            :: The inverse of the value of the
+                {{PublicKeyCredential/[DISCOVER-METHOD]/sameOriginWithAncestors}}
+                argument passed to this [=internal method=].
+            : {{CollectedClientData/topOrigin}}
+            :: The [=ascii serialization of an origin|serialization of=] |callerOrigin|'s [=top-level origin=] if
+                the {{PublicKeyCredential/[DISCOVER-METHOD]/sameOriginWithAncestors}}
+                argument passed to this [=internal method=] is [FALSE], else `undefined`.
+    </li>
 
-1. Let |clientDataJSON| be the [=JSON-compatible serialization of client data=] constructed from |collectedClientData|.
+    <li id="GetAssn-SerializeClientDataJSON">
+        Let |clientDataJSON| be the [=JSON-compatible serialization of client data=] constructed from |collectedClientData|.
+    </li>
 
 1. Let |clientDataHash| be the [=hash of the serialized client data=] represented by |clientDataJSON|.
 
@@ -4476,7 +4484,7 @@ Note: The {{PublicKeyCredentialHint}} enumeration is deliberately not referenced
 
 ## Permissions Policy integration ## {#sctn-permissions-policy}
 
-This specification defines three [=policy-controlled features=]. [[!Permissions-Policy]]
+This specification defines three [=policy-controlled features=] [[!Permissions-Policy]].
 
 Two features are identified by the feature-identifier tokens "<code><dfn data-lt="publickey-credentials-create-feature" export>publickey-credentials-create</dfn></code>"
 and "<code><dfn data-lt="publickey-credentials-get-feature" export>publickey-credentials-get</dfn></code>".
@@ -7814,7 +7822,7 @@ This permission MUST be configured per-origin and MUST NOT offer the capability 
 :: When [[#sctn-createCredential|creating a new credential]], this extension injects the following
     steps at labeled points in that algorithm:
 
-    1. Just before [establishing the RP ID](#CreateCred-DetermineRpId):
+    1. Replace [establishing the RP ID](#CreateCred-DetermineRpId) with following procedure
         1. If the result of [=getting the current permission state=] for
             `"publickey-credentials-remote-client-data-json"` and the [=current settings object=]
             is not `"granted"`.
@@ -7826,49 +7834,58 @@ This permission MUST be configured per-origin and MUST NOT offer the capability 
             given |remoteClientDataJSON|.
             1. If above parsing step fails, throw a "{{NotSupportedError}}" {{DOMException}}.
         1. Let |remoteOrigin| be the value of the `"origin"` key of |collectedClientData|.
-
-    1. In [establishing the RP ID](#CreateCred-DetermineRpId):
         1. If <code>|pkOptions|.{{PublicKeyCredentialCreationOptions/rp}}.{{PublicKeyCredentialRpEntity/id}}</code>
-            [=is not a registrable domain suffix of and is not equal to=] |effectiveDomain|,
-            instead of running the [$related origins validation procedure$],
-            verify that <code>|pkOptions|.{{PublicKeyCredentialCreationOptions/rp}}.{{PublicKeyCredentialRpEntity/id}}</code>
-            [=is a registrable domain suffix of or is equal to=] |remoteOrigin|.
-            1. If this check fails, throw a "{{SecurityError}}" {{DOMException}}.
+            [=is not a registrable domain suffix of and is not equal to=] |remoteOrigin|, and if the client
+            1. does not support [[#sctn-related-origins|related origin requests]]
+                1. Throw a "{{SecurityError}}" {{DOMException}}.
+            1. supports [[#sctn-related-origins|related origin requests]]
+                1. Skip [$related origins validation procedure$] for this operation.
 
-    1. In the step that constructs |clientDataJSON| from |collectedClientData|,
+    1. Skip [constructing collectedClientData](#CreateCred-ConstructCollectedClientData) step.
+        1. |collectedClientData| has been parsed from given |remoteClientDataJSON| in [establishing the RP ID](#CreateCred-DetermineRpId).
+        1. The [=user agent=] MUST NOT add, remove, or modify any contents of |collectedClientData|,
+            as doing so would invalidate the remote machine's calculated |clientDataJSON|,
+            resulting in a signature mismatch at the [=[RP]=].
+
+    1. In [serializing client data](#CreateCred-SerializeClientDataJSON) step,
         set |clientDataJSON| to |remoteClientDataJSON| instead.
         1. The [=user agent=] MUST NOT add, remove, or modify any contents of |remoteClientDataJSON|,
-            as doing so would invalidate the remote machine's calculated |clientDataJSON|,
+            to construct |clientDataJSON| here as doing so would invalidate the remote machine's calculated |clientDataJSON|,
             resulting in a signature mismatch at the [=[RP]=].
 
 : Client extension processing ([=authentication extension|authentication=])
 :: When [[#sctn-getAssertion|generating an authentication assertion]], this extension injects the
     following steps at labeled points in that algorithm:
 
-    1. Just before [establishing the RP ID](#GetAssn-DetermineRpId):
-        1. If the result of [=getting the current permission state=] [[!Permissions]] for
+    1. Replace [establishing the RP ID](#GetAssn-DetermineRpId) with following procedure:
+        1. If the result of [=getting the current permission state=] for
             `"publickey-credentials-remote-client-data-json"` and the [=current settings object=]
-            is not `"granted"`
-            1. Throw a "{{NotAllowedError}}" {{DOMException}}
-        1. If <code>|pkOptions|.{{PublicKeyCredentialRequestOptions/rpId}}</code>
-            is not present, throw a "{{NotAllowedError}}" {{DOMException}}.
+            is not `"granted"`.
+            1. Throw a "{{NotAllowedError}}" {{DOMException}}.
+        1. If <code>|pkOptions|.{{PublicKeyCredentialCreationOptions/rp}}.{{PublicKeyCredentialRpEntity/id}}</code>
+            is not present
+            1. Throw a "{{NotAllowedError}}" {{DOMException}}.
         1. Let |collectedClientData| be the result of [=parse a JSON string to an Infra value=]
-            given |remoteClientDataJSON|. [[!infra]]
-            If this throws, throw a "{{NotSupportedError}}" {{DOMException}}.
+            given |remoteClientDataJSON|.
+            1. If above parsing step fails, throw a "{{NotSupportedError}}" {{DOMException}}.
         1. Let |remoteOrigin| be the value of the `"origin"` key of |collectedClientData|.
+        1. If <code>|pkOptions|.{{PublicKeyCredentialCreationOptions/rp}}.{{PublicKeyCredentialRpEntity/id}}</code>
+            [=is not a registrable domain suffix of and is not equal to=] |remoteOrigin|, and if the client
+            1. does not support [[#sctn-related-origins|related origin requests]]
+                1. Throw a "{{SecurityError}}" {{DOMException}}.
+            1. supports [[#sctn-related-origins|related origin requests]]
+                1. Skip [$related origins validation procedure$] for this operation.
 
-    1. In [establishing the RP ID](#GetAssn-DetermineRpId):
-        1. If <code>|pkOptions|.{{PublicKeyCredentialRequestOptions/rpId}}</code>
-            [=is not a registrable domain suffix of and is not equal to=] |effectiveDomain|,
-            instead of running the [$related origins validation procedure$],
-            verify that <code>|pkOptions|.{{PublicKeyCredentialRequestOptions/rpId}}</code>
-            [=is a registrable domain suffix of or is equal to=] |remoteOrigin|.
-            1. If this check fails, throw a "{{SecurityError}}" {{DOMException}}.
+    1. Skip [constructing collectedClientData](#GetAssn-ConstructCollectedClientData) step.
+        1. |collectedClientData| has been parsed from given |remoteClientDataJSON| in [establishing the RP ID](#GetAssn-DetermineRpId).
+        1. The [=user agent=] MUST NOT add, remove, or modify any contents of |collectedClientData|,
+            as doing so would invalidate the remote machine's calculated |clientDataJSON|,
+            resulting in a signature mismatch at the [=[RP]=].
 
-    1. In the step that constructs |clientDataJSON| from |collectedClientData|,
+    1. In [serializing client data](#GetAssn-SerializeClientDataJSON) step,
         set |clientDataJSON| to |remoteClientDataJSON| instead.
         1. The [=user agent=] MUST NOT add, remove, or modify any contents of |remoteClientDataJSON|,
-            as doing so would invalidate the remote machine's calculated |clientDataJSON|,
+            to construct |clientDataJSON| here as doing so would invalidate the remote machine's calculated |clientDataJSON|,
             resulting in a signature mismatch at the [=[RP]=].
 
 : Authenticator extension input
@@ -7881,7 +7898,7 @@ This permission MUST be configured per-origin and MUST NOT offer the capability 
 :: None
 
 : Client extension output
-:: Returns the boolean to indicate whether that the extension rules were checked and validated.
+:: Returns the boolean to indicate whether parsing all the extension steps resulted in a success or failure.
     <xmp class="idl">
     partial dictionary AuthenticationExtensionsClientOutputs {
       boolean remoteClientDataJson;
@@ -8764,8 +8781,6 @@ possible for [=[RPS]=] to trust any further [=attestation statements=] from the 
 
 See also the related security consideration for [=[RPS]=] in [[#sctn-revoked-attestation-certificates]].
 
-## Security considerations for [=clients=] ## {#sctn-security-considerations-client}
-
 ### The remoteClientDataJSON Extension ### {#sctn-remote-clientdatajson-security}
 
 The [=remoteClientDataJSON=] extension enables a [=Remote desktop web client=] to proxy
@@ -8789,6 +8804,11 @@ alters the trust model of the [=Web Authentication API=] in the following ways:
 [=User agents=] that grant this extension SHOULD clearly communicate to users when a
 remote-desktop-proxied WebAuthn operation is in progress, so that users understand
 that their authenticator is being used on behalf of a remotely hosted application.
+
+<!-- no sec cons for clients enumerated at this time
+## Security considerations for [=clients=] ## {#sctn-security-considerations-client}
+-->
+
 
 ## Security considerations for [=[RPS]=] ## {#sctn-security-considerations-rp}
 
--- a/index.bs
+++ b/index.bs
@@ -7771,7 +7771,7 @@ in [=Remote desktop web client=].
 This allows web-based remote desktop web clients to execute remote WebAuthn requests locally. Here the [=Remote desktop web client=] is calling WebAuthn request on behalf of
 RP present on the remote machine and have different origin than the calling RP in the remote machine.
 
-Note: This extension is a [=powerful feature=] that requires explicit per-origin permission before use.
+This extension is a [=powerful feature=] that requires explicit per-origin permission before use.
 [=User agents=] SHOULD NOT prompt the user for this permission at request time;
 instead, the grant is administered through device management policy (e.g., managed browser profiles, managed browser instances, or
 managed devices that enumerate specific [=Remote desktop web client=] origins) or
@@ -7785,8 +7785,6 @@ This permission MUST be configured per-origin and MUST NOT offer the capability 
     The [=default permission state=] for this feature is `"denied"`,
     meaning the extension is not available by default and must be explicitly enabled.
     Since no additional aspects are defined, the [=permission descriptor type=] defaults to {{PermissionDescriptor}}.
-
-    The [=permission descriptor type=] for this feature is:
 
     <div class="example" id="example-publickey-credentials-remote-client-data-json-permission-query">
     A [=Remote desktop web client=] may check whether the [=user agent=] has been configured
@@ -7898,7 +7896,7 @@ This permission MUST be configured per-origin and MUST NOT offer the capability 
 :: None
 
 : Client extension output
-:: Returns the boolean to indicate whether parsing all the extension steps resulted in a success or failure.
+:: Returns the value [TRUE] to indicate to the [=[RP]=] that the extension was acted upon.
     <xmp class="idl">
     partial dictionary AuthenticationExtensionsClientOutputs {
       boolean remoteClientDataJson;
@@ -8789,8 +8787,8 @@ alters the trust model of the [=Web Authentication API=] in the following ways:
 
 - **Origin substitution:** The [=user agent=] accepts a {{CollectedClientData/origin}} supplied by the calling
     [=Remote desktop web client=] rather than deriving it from the current execution context.
-    Any [=user agent=] granting this permission must therefore trust that the caller
-    accurately supplies the origin of the remote [=[RP]=].
+    Any [=user agent=] granting this permission MUST therefore trust that the caller
+    accurately and honestly supplies the origin of the remote [=[RP]=].
 
 - **RP ID validation:** The [=RP ID=] is validated against the origin present in |remoteClientDataJSON|
     rather than the [=effective domain=] of the local [=client=]'s origin.
@@ -8804,10 +8802,6 @@ alters the trust model of the [=Web Authentication API=] in the following ways:
 [=User agents=] that grant this extension SHOULD clearly communicate to users when a
 remote-desktop-proxied WebAuthn operation is in progress, so that users understand
 that their authenticator is being used on behalf of a remotely hosted application.
-
-<!-- no sec cons for clients enumerated at this time
-## Security considerations for [=clients=] ## {#sctn-security-considerations-client}
--->
 
 
 ## Security considerations for [=[RPS]=] ## {#sctn-security-considerations-rp}
--- a/index.bs
+++ b/index.bs
@@ -1,4 +1,4 @@
-﻿<!--   ***********  Web Authentication - Level 3 - Spec source file ***********
+<!--   ***********  Web Authentication - Level 3 - Spec source file ***********
 Notes:
 * the h1 tag is spec title that is rendered within the document window. Wrapping
   may be controlled with break tags.  The spec 'Level' value is not appended.
--- a/index.bs
+++ b/index.bs
@@ -7832,14 +7832,10 @@ This permission MUST be configured per-origin and MUST NOT offer the capability 
             given |remoteClientDataJSON|.
             1. If above parsing step fails, throw a "{{EncodingError}}" {{DOMException}}.
         1. Let |remoteOrigin| be the value of the `"origin"` key of |collectedClientData|.
-        1. If <code>|pkOptions|.{{PublicKeyCredentialCreationOptions/rp}}.{{PublicKeyCredentialRpEntity/id}}</code>
-            [=is not a registrable domain suffix of and is not equal to=] |remoteOrigin|, and if the client
-            1. does not support [[#sctn-related-origins|related origin requests]]
-                1. Throw a "{{SecurityError}}" {{DOMException}}.
-            1. supports [[#sctn-related-origins|related origin requests]]
-                1. Skip [$related origins validation procedure$] for this operation.
-                
-                    Note: This means the local client is delegating the [$related origins validation procedure$] to the remote client.
+        1. Skip checking whether <code>|pkOptions|.{{PublicKeyCredentialCreationOptions/rp}}.{{PublicKeyCredentialRpEntity/id}}</code>
+            [=is a registrable domain suffix of or is equal to=] |remoteOrigin| or not.
+
+            Note: This means the local client is delegating all the [=RP ID=] related origin checks to the remote client.
 
     1. Skip [constructing collectedClientData](#CreateCred-ConstructCollectedClientData) step.
         1. |collectedClientData| has been parsed from given |remoteClientDataJSON| in [establishing the RP ID](#CreateCred-DetermineRpId).
@@ -7869,12 +7865,10 @@ This permission MUST be configured per-origin and MUST NOT offer the capability 
             given |remoteClientDataJSON|.
             1. If above parsing step fails, throw a "{{NotSupportedError}}" {{DOMException}}.
         1. Let |remoteOrigin| be the value of the `"origin"` key of |collectedClientData|.
-        1. If <code>|pkOptions|.{{PublicKeyCredentialCreationOptions/rp}}.{{PublicKeyCredentialRpEntity/id}}</code>
-            [=is not a registrable domain suffix of and is not equal to=] |remoteOrigin|, and if the client
-            1. does not support [[#sctn-related-origins|related origin requests]]
-                1. Throw a "{{SecurityError}}" {{DOMException}}.
-            1. supports [[#sctn-related-origins|related origin requests]]
-                1. Skip [$related origins validation procedure$] for this operation.
+        1. Skip checking whether <code>|pkOptions|.{{PublicKeyCredentialCreationOptions/rp}}.{{PublicKeyCredentialRpEntity/id}}</code>
+            [=is a registrable domain suffix of or is equal to=] |remoteOrigin| or not.
+
+            Note: This means the local client is delegating all the [=RP ID=] related origin checks to the remote client.
 
     1. Skip [constructing collectedClientData](#GetAssn-ConstructCollectedClientData) step.
         1. |collectedClientData| has been parsed from given |remoteClientDataJSON| in [establishing the RP ID](#GetAssn-DetermineRpId).
@@ -8795,10 +8789,10 @@ alters the trust model of the [=Web Authentication API=] in the following ways:
     Any [=user agent=] granting this permission MUST therefore trust that the caller
     accurately and honestly supplies the origin of the remote [=[RP]=].
 
-- **RP ID validation:** The [=RP ID=] is validated against the origin present in |remoteClientDataJSON|
-    rather than the [=effective domain=] of the local [=client=]'s origin.
-    If content running in the [=Remote desktop web client=] origin can influence the
-    |remoteClientDataJSON| value, it could cause assertions valid for an unintended [=[RP]=] to be produced.
+- **RP ID validation:** The [=RP ID=] validatation is delegated to the remote client as remote client
+    has all the necessary context of validating the origin to account for related origins and platform specific app deployment models.
+    Any [=user agent=] granting this permission MUST therefore trust that the remote client has
+    accurately and honestly validated [=RP ID=] against all the origin checks.
 
 - **Per-origin configuration mechanism required:** As this extension's permissions are per origin,
     [=user agents=] MUST only provide configuration mechanisms which are per-origin basis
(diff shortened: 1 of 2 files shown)


index.bs
```diff
@@ -7740,6 +7740,59 @@ However, [=authenticators=] that do not utilize [[!FIDO-CTAP]] do not necessaril
...
+: Extension identifier
+:: `remoteDesktopClientOverrideJSON`
+
+: Operation applicability
+:: [=registration extension|Registration=] and [=authentication extension|authentication=]
+
+: Client extension input
+::  <xmp class="idl">
+    partial dictionary AuthenticationExtensionsClientInputs {
+        DOMString remoteClientDataJSON;
```
MasterKale: Is it not feasible to make `remoteClientDataJSON` a dictionary of the shape of `CollectedClientData`? It's not immediately clear what is gained from stringifying the JSON.

index.bs
```diff
@@ -7740,6 +7740,59 @@ However, [=authenticators=] that do not utilize [[!FIDO-CTAP]] do not necessaril
...
+    </xmp>
+
+: Client extension processing ([=registration extension|registration=] and [=authentication extension|authentication=])
+::
+       1. If browser support this extension:
+           1. Save current |origin| as remoteDesktopClientOrigin
+           1. JSON parse remoteClientDataJSON to construct |collectedClientData|.
+           1. If there is error parsing remoteClientDataJSON:
+               1. Return a {{DOMException}} whose name is “{{NotSupportedError}}”.
+           1. Let |clientDataJSON| be equal to remoteClientDataJSON.
```
MasterKale: This step would happen to overwrite what the remote browser collects for `origin`, `crossOrigin`, `topOrigin`, etc... too. How concerned are we that this extension could be used maliciously in non-remote scenarios? This is essentially telling the browser "parrot back what the RP wants to see" which seems to dilute the value of the browser being an active participant in the ceremony when it creates `clientDataJSON`. Maybe I'm missing something, I reviewed #1577 today and couldn't quite understand the intended benefit here.

index.bs
kreichgauer: Chromium adds an additional key to the CollectedClientData, to ensure websites validate the parsed JSON rather than compare it against a string template. https://chromium.googlesource.com/chromium/src/+/master/content/browser/webauth/client_data_json.md

index.bs
```diff
@@ -7740,6 +7740,59 @@ However, [=authenticators=] that do not utilize [[!FIDO-CTAP]] do not necessaril
...
+:: None
+
+: Client extension output
+:: <xmp class="idl">
+    partial dictionary AuthenticationExtensionsClientOutputs {
+        DOMString remoteDesktopClientOrigin;
+    };
+    partial dictionary AuthenticationExtensionsClientOutputsJSON {
+        DOMString remoteDesktopClientOrigin;
+    };
```
MasterKale: **From in-person chat @ Feb 3:** It's likely to be deemed a fingerprinting mechanism/privacy violation if we allow a non-workforce RP (e.g. github.com) to receive a signal that it's "being accessed from a browser on a remote machine being accessed via RDP." Typical use of WebAuthn wouldn't otherwise reveal such a signal to an RP so I don't think we can provide it here as extension output.

It _could_ be possible for use of this extension to be defined on a per-RP ID basis, but that would be platform-level policy outside the influence of this spec.

index.bs
```diff
@@ -7740,6 +7740,59 @@ However, [=authenticators=] that do not utilize [[!FIDO-CTAP]] do not necessaril
...
 
+### Remote Desktop Client Override JSON extension (<dfn>remoteDesktopClientOverrideJSON</dfn>) ### {#sctn-remote-desktop-client-override-json-extension}
+
+Remote desktop web applications let users interact with the (native) desktop environment of a remote host from a local client (web) application.
+
+This [=client extension|client=] [=registration extension=] and [=authentication extension=] allows user to use
+WebAuthn authenticator connected to their local client in order to sign into a website in a browser running on a remote host
+in remote desktop web application.
+
+This allows web-based remote desktop clients to execute remote WebAuthn requests locally.
```
MasterKale: Can we add a non-normative note here that this extension is intended to only be available on managed endpoints? There's general agreement that this extension would only be enabled when an endpoint is managed, but this point cannot be appreciated just by reading the spec. The lack of such a note could set the expectation that this extension would be widely available because "it's defined in the spec."

MasterKale (approved):
Pending the removal of `remoteDesktopClientOrigin` from `AuthenticationExtensionsClientOutputs` I'm fine with this extension.

MasterKale:
@akshayku I prepared this sequence diagram, based on our in-person conversation this week, to try and document an end-to-end use of passkeys that would benefit from this extension. Let me know if there's anything I can improve:

```mermaid
sequenceDiagram
    autonumber
    
    box Remote Machine
    participant RP.com
    participant Remote Browser
    participant RemotePlatform as Platform
    end

    box Client Machine
    participant MyRDC.com
    participant Client Browser
    participant ClientPlatform as Platform
    end

    MyRDC.com ->> MyRDC.com: Prepare JavaScript RDP client
    MyRDC.com ->> RemotePlatform: User initiates RDP connection
    Remote Browser ->> RP.com: User loads RP.com
    RP.com ->> RP.com: Ask user to log in with a passkey
    RP.com ->> Remote Browser: Call WebAuthn .create()/.get()
    Remote Browser ->> RemotePlatform: Send .create()/.get() options and clientDataJSON<br>to begin WebAuthn
    RemotePlatform ->> MyRDC.com: Redirect .create()/.get() options and clientDataJSON
    critical New extension usage
    MyRDC.com ->> MyRDC.com: Add redirected clientDataJSON<br>to .create()/.get() options<br>as remoteClientDataJSON extension
    MyRDC.com ->> Client Browser: Call WebAuthn .create()/.get()
    Client Browser ->> Client Browser: Use remoteClientDataJSON <br>as clientDataJSON
    end
    Client Browser ->> ClientPlatform: Send .create()/.get() options and clientDataJSON<br>to begin WebAuthn
    ClientPlatform ->> ClientPlatform: Collect WebAuthn response
    ClientPlatform ->> Client Browser: Return WebAuthn response
    Client Browser ->> MyRDC.com: Return WebAuthn response
    MyRDC.com ->> RemotePlatform: Redirect WebAuthn response
    RemotePlatform ->> Remote Browser: Return WebAuthn response
    Remote Browser ->> RP.com: Return WebAuthn response
    RP.com ->> RP.com: Verify WebAuthn response
    RP.com ->> RP.com: Log user in
```

index.bs
```diff
@@ -7740,6 +7740,59 @@ However, [=authenticators=] that do not utilize [[!FIDO-CTAP]] do not necessaril
 : Authenticator extension processing
 :: [=largeblob|This extension=] directs the user-agent to cause the large blob to be stored on, or retrieved from, the authenticator. It thus does not specify any direct authenticator interaction for [=[RPS]=].
 
+### Remote Desktop Client Override JSON extension (<dfn>remoteDesktopClientOverrideJSON</dfn>) ### {#sctn-remote-desktop-client-override-json-extension}
```
MasterKale: Is "Remote Desktop" too solution-specific? Couldn't a non-RDP-based remote connection scheme (e.g. VNC?) that's capable of redirecting WebAuthn requests also use this?

This got me thinking about a potentially different name for this extension. Maybe "Remote Client Data JSON Override?" `remoteClientDataJSONOverride`? 🤔

index.bs
akshayku: Added a note.

MasterKale (approved):

index.bs
```diff
@@ -7740,6 +7740,52 @@ However, [=authenticators=] that do not utilize [[!FIDO-CTAP]] do not necessaril
...
+::  <xmp class="idl">
+    partial dictionary AuthenticationExtensionsClientInputs {
+        DOMString remoteClientDataJSON;
+    };
+    partial dictionary AuthenticationExtensionsClientInputsJSON {
+        DOMString remoteClientDataJSON;
+    };
+    </xmp>
+
+: Client extension processing ([=registration extension|registration=] and [=authentication extension|authentication=])
```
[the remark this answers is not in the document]
akshayku: @kreichgauer , I have added explanation in the extension around Client Skipping RPID check against effective domain. Also, I have added a check that when this extension is present, RPID must be specified by the remote web client in the create()/get() options and it must match origin present in the remoteClientDataJSON. Remote side web client is already doing RPID check as per the rules of the spec. 

Can you review again?

akshayku:
@emlun , @timcappalli , @kreichgauer 

Please review the PR again.

index.bs
```diff
@@ -7740,6 +7740,76 @@ However, [=authenticators=] that do not utilize [[!FIDO-CTAP]] do not necessaril
...
+<dfn>Remote desktop web client</dfn> let users interact with the (native) desktop environment of a remote host from a local client (web) application.
+
+[=remoteClientDataJSON|This=] [=registration extension=] and [=authentication extension=] allows user to use a
+WebAuthn authenticator connected to the local [=client platform=] in order to sign into a website in a browser, or an app running on a remote host
+in [=Remote desktop web client=].
+
+This allows web-based remote desktop web clients to execute remote WebAuthn requests locally. Here the [=Remote desktop web client=] is calling WebAuthn request on behalf of
+RP present on the remote machine and have different origin than the calling RP in the remote machine.
+
+Note: This extension is intended to be available in managed contexts (managed browser profiles, managed browser instances, managed devices, etc) or via explicit user opt-in configuration of a [=WebAuthn Client=].
```
kreichgauer: I don't think this is quite strong enough. For example, it's not saying that user agents probably never want the extension to be available to all caller origins. Any opt-in, whether it's via device management or user configuration, has to be at least on a per-origin basis.  

I wonder if access to this extension should be formally declared as a [powerful feature](https://w3c.github.io/permissions/#dfn-powerful-feature), and we standardize a permission along with it. The expectation wouldn't be that any user agent actually prompts the user at request time for the permission, it would still be policy controlled. But that at least regulates the per-origin nature of the opt-in, provides permissions policy integration for iframes, and also a query API for RPs.

akshayku: I like this powerful feature approach as that introduces user opt-in formally like other similar features like geolocation. 

So, in my mind similar to geolocation powerful feature, there is a global OS level system settings for user to explicitly opt in/enable remote WebAuthn and then user has to opt-in explicitly again for individual page where remote web client can request this permission. 

Thoughts? 

Also, didn't understand your query API for RPs comment. This is on the client side, so how is this query going to work?

kreichgauer: > So, in my mind similar to geolocation powerful feature, there is a global OS level system settings for user to explicitly opt in/enable remote WebAuthn and then user has to opt-in explicitly again for individual page where remote web client can request this permission.

I don't think that's how we would do it in Chromium at least. We would probably want a device level management policy that takes a list of origins; and maybe some flag overrides that developers can set for testing. But different user agents might well come to different conclusions about this.

For reference, here is a recent example of a similar case where enterprise policy support in Chromium allows admins to exempt an individual origin from a security restriction (the user gesture requirement for fullscreen): https://github.com/explainers-by-googlers/html-fullscreen-without-a-gesture Though it looks like the PR for it never landed.

> Also, didn't understand your query API for RPs comment. This is on the client side, so how is this query going to work?

With a permission, the RDP client app could do something like 

`navigator.permissions.query({name: 'webauthn', allowRemoteDesktopClientDataJSON: true});`

to query whether the UA has indeed been configured to permit proxied WebAuthn requests from the RPD client origin.

akshayku: @kreichgauer , I have defined this feature as a powerful feature. Please re-review.

akshayku:
@emlun / @pascoej / @timcappalli / @ve7jtb / @nsatragno / @kreichgauer

 Can you all please review this PR?

index.bs
```diff
@@ -4475,10 +4476,14 @@ Note: The {{PublicKeyCredentialHint}} enumeration is deliberately not referenced
...
+This specification defines three [=policy-controlled features=]. [[!Permissions-Policy]]
+
+Two features are identified by the feature-identifier tokens "<code><dfn data-lt="publickey-credentials-create-feature" export>publickey-credentials-create</dfn></code>"
 and "<code><dfn data-lt="publickey-credentials-get-feature" export>publickey-credentials-get</dfn></code>".
-Their [=default allowlists=] are both '<code>self</code>'. [[!Permissions-Policy]]
+Their [=default allowlists=] are both '<code>self</code>'.
+
+A third feature is identified by the feature-identifier token "<code><dfn data-lt="remote-client-data-json-feature" export>remote-client-data-json</dfn></code>".
+Its [=default allowlist=] is '<code>none</code>', meaning it is disabled for all origins by default. This feature is also a [=powerful feature=] ([[!Permissions]]) identified by the name <code>"remote-client-data-json"</code>, with [=default permission state=] `"denied"`. Access is granted only via user agent device management policy that specifies a per-origin allowlist of trusted [=Remote desktop web client=] origins, or through explicit per-origin user opt-in in [=WebAuthn Client=] settings.
```
kreichgauer: Different user agents might arrive at different conclusions for what kind of mechanism is appropriate to grant the exception, so I think we can be less prescriptive here without much risk to interoperability. Mechanisms like enterprise policies/device management aren't really a thing in the domain of web specs anyway I think?

Somewhere else we should clearly calling out the risks involved, perhaps a subsection in "Security Considerations"?

```suggestion
Its [=default allowlist=] is '<code>none</code>', meaning it is disabled for all origins by default. This feature is also a [=powerful feature=] ([[!Permissions]]) identified by the name <code>"remote-client-data-json"</code>, with [=default permission state=] `"denied"`.
```

index.bs
```diff
@@ -4475,10 +4476,14 @@ Note: The {{PublicKeyCredentialHint}} enumeration is deliberately not referenced
...
-This specification defines two [=policy-controlled features=] identified by
-the feature-identifier tokens "<code><dfn data-lt="publickey-credentials-create-feature" export>publickey-credentials-create</dfn></code>"
+This specification defines three [=policy-controlled features=]. [[!Permissions-Policy]]
+
+Two features are identified by the feature-identifier tokens "<code><dfn data-lt="publickey-credentials-create-feature" export>publickey-credentials-create</dfn></code>"
 and "<code><dfn data-lt="publickey-credentials-get-feature" export>publickey-credentials-get</dfn></code>".
-Their [=default allowlists=] are both '<code>self</code>'. [[!Permissions-Policy]]
+Their [=default allowlists=] are both '<code>self</code>'.
+
+A third feature is identified by the feature-identifier token "<code><dfn data-lt="remote-client-data-json-feature" export>remote-client-data-json</dfn></code>".
```
kreichgauer: Should we use the same prefix as the other two features? `publickey-credentials-remote-client-data-json` or something like that?

index.bs
```diff
@@ -7740,6 +7745,107 @@ However, [=authenticators=] that do not utilize [[!FIDO-CTAP]] do not necessaril
...
+        DOMString remoteClientDataJSON;
+    };
+    partial dictionary AuthenticationExtensionsClientInputsJSON {
+        DOMString remoteClientDataJSON;
+    };
+    </xmp>
+
+: Client extension processing
+:: [=registration extension|registration=] processing changes
+    1. If the [=permission state=] of `"remote-client-data-json"` for the [=current settings object=] is not `"granted"`: [[!Permissions]]
```
kreichgauer: Should this reference some standard algorithm for checking permissions? Other specs seem to use this e.g. https://html.spec.whatwg.org/multipage/iframe-embed-object.html#allowed-to-use

(Same comment for assertion processing.)

index.bs
```diff
@@ -7740,6 +7745,107 @@ However, [=authenticators=] that do not utilize [[!FIDO-CTAP]] do not necessaril
...
+         1. Throw a "{{NotAllowedError}}" {{DOMException}}
+    1. If <code>|pkOptions|.{{PublicKeyCredentialRequestOptions/rpId}}</code> is not present
+         1. Throw a "{{NotAllowedError}}" {{DOMException}}
+    1. Skip <code>|pkOptions|.{{PublicKeyCredentialCreationOptions/rp}}.{{PublicKeyCredentialRpEntity/id}}</code>
```
kreichgauer: Should this also skip the related origin request?

(Same comment for assertion processing.)

index.bs
```diff
@@ -7740,6 +7745,107 @@ However, [=authenticators=] that do not utilize [[!FIDO-CTAP]] do not necessaril
...
+          check of being registrable domain suffix or equal to |effectiveDomain|.
+    1. JSON parse remoteClientDataJSON to construct |collectedClientData|.
```
kreichgauer: JSON parsing should probably reference a canonical spec. Perhaps https://infra.spec.whatwg.org/#parse-json-bytes-to-an-infra-value?

(Same comment for assertion processing.)

index.bs
```diff
@@ -7740,6 +7745,107 @@ However, [=authenticators=] that do not utilize [[!FIDO-CTAP]] do not necessaril
...
+         1. If there is error parsing remoteClientDataJSON:
+             1. Return a {{DOMException}} whose name is “{{NotSupportedError}}”.
+         1. If origin inside remoteClientDataJSON does not match <code>|pkOptions|.{{PublicKeyCredentialCreationOptions/rp}}.{{PublicKeyCredentialRpEntity/id}}</code>
```
kreichgauer: So this is supposed to do the RP ID validation that the step above said to skip? In that case, same question, should this do related origin validation?

Perhaps it would be better to to the JSON parsing as a first step, and then instead of skipping the initial RP ID validation, to just do it with the different origin as an input?

(Same comment for assertion processing.)

index.bs
```diff
@@ -7740,6 +7745,107 @@ However, [=authenticators=] that do not utilize [[!FIDO-CTAP]] do not necessaril
...
+managed devices that enumerate specific [=Remote desktop web client=] origins) or
+through explicit per-origin user opt-in in [=WebAuthn Client=] settings.
+This permission MUST be configured per-origin and MUST NOT offer the capability to permit all origins.
+
+: Permissions
+:: This extension defines a [=powerful feature=] ([[!Permissions]]) and a [=policy-controlled feature=] ([[!Permissions-Policy]]), both identified by the name <code>"remote-client-data-json"</code>.
+    The [=default allowlist=] for this [=policy-controlled feature=] is `'none'`.
+    The [=default permission state=] for this feature is `"denied"`, meaning the extension is not available by default and must be explicitly enabled.
+
+    The [=permission descriptor type=] for this feature is:
```
kreichgauer: Defining a permission descriptor is optional. Since the powerful feature doesn't define any aspects, I think we can omit it, and just defaults to PermissionDescriptor: 
https://www.w3.org/TR/permissions/#specifying-a-powerful-feature
https://www.w3.org/TR/permissions/#dfn-permission-descriptor-type

index.bs
```diff
@@ -7740,6 +7745,107 @@ However, [=authenticators=] that do not utilize [[!FIDO-CTAP]] do not necessaril
...
+
+    <xmp class="idl">
+    dictionary RemoteClientDataJSONPermissionDescriptor : PermissionDescriptor {
+        // name is "remote-client-data-json"
+    };
+    </xmp>
+
+    A [=Remote desktop web client=] may query whether the [=user agent=] has been configured to permit use of this extension via the {{Permissions}} API:
```
kreichgauer: We should add this to the explainer, but it isn't really the domain of WebAuthn to explain the navigator.permissions API. Perhaps we should we remove this? Or make it an "example".

kreichgauer (commented):
Thanks for adding the permission/permissions policy integration. I think this is useful to add.

index.bs
akshayku: Changed.

index.bs
akshayku: Made it as an example.

index.bs
akshayku: Defaulted to PermissionDescriptor.

index.bs
akshayku: Yes. It should skip related origins validation procedure. Added.

index.bs
akshayku: Yes. Added.

index.bs
akshayku: Referenced Infra spec.

index.bs
akshayku: Not sure if this addressed your comment. Can you check the new text? And propose the change here if needed?

akshayku:
> @emlun / @pascoej / @timcappalli / @ve7jtb / @nsatragno / @kreichgauer
> 
> Can you all please review this PR?

Folks, Please review.

index.bs
```diff
@@ -7747,6 +7752,107 @@ However, [=authenticators=] that do not utilize [[!FIDO-CTAP]] do not necessaril
...
+:: [=registration extension|registration=] processing changes
+    1. If the result of [=getting the current permission state=] for `"publickey-credentials-remote-client-data-json"` and the [=current settings object=] is not `"granted"`: [[!Permissions]]
+         1. Throw a "{{NotAllowedError}}" {{DOMException}}
+    1. If <code>|pkOptions|.{{PublicKeyCredentialRequestOptions/rpId}}</code> is not present
+         1. Throw a "{{NotAllowedError}}" {{DOMException}}
+    1. Let |collectedClientData| be the result of [=parse a JSON string to an Infra value=] given |remoteClientDataJSON|. [[!infra]]
+         1. If this throws, return a {{DOMException}} whose name is "{{NotSupportedError}}".
+    1. Let |remoteOrigin| be the value of the `"origin"` key of |collectedClientData|.
+    1. Verify that <code>|pkOptions|.{{PublicKeyCredentialCreationOptions/rp}}.{{PublicKeyCredentialRpEntity/id}}</code>
+          [=is a registrable domain suffix of or is equal to=] |remoteOrigin|, substituting |remoteOrigin| for |effectiveDomain| in that check.
```
emlun: There is no `|effectiveDomain|` in the [is a registrable domain suffix of or is equal to](https://html.spec.whatwg.org/multipage/browsers.html#is-a-registrable-domain-suffix-of-or-is-equal-to) procedure in Infra, and `...or is equal to |remoteOrigin|` already makes `|remoteOrigin|` the argument to the _originalHost_ parameter in that procedure.

I think this needs to modify the check in step 8 of `PublicKeyCredential.create()` and step 7 of `PublicKeyCredential.get()` instead.

index.bs
```diff
@@ -7747,6 +7752,107 @@ However, [=authenticators=] that do not utilize [[!FIDO-CTAP]] do not necessaril
...
+    partial dictionary AuthenticationExtensionsClientInputs {
+        DOMString remoteClientDataJSON;
+    };
+    partial dictionary AuthenticationExtensionsClientInputsJSON {
+        DOMString remoteClientDataJSON;
+    };
+    </xmp>
+
+: Client extension processing
+:: [=registration extension|registration=] processing changes
```
emlun: We have this formatting precedent now from the PRF extension:
```suggestion
: Client extension processing ([=registration extension|registration=])
::
```

index.bs
```diff
@@ -7747,6 +7752,107 @@ However, [=authenticators=] that do not utilize [[!FIDO-CTAP]] do not necessaril
...
+    1. Let |remoteOrigin| be the value of the `"origin"` key of |collectedClientData|.
+    1. Verify that <code>|pkOptions|.{{PublicKeyCredentialCreationOptions/rp}}.{{PublicKeyCredentialRpEntity/id}}</code>
+          [=is a registrable domain suffix of or is equal to=] |remoteOrigin|, substituting |remoteOrigin| for |effectiveDomain| in that check.
+         1. If this check fails, throw a "{{NotAllowedError}}" {{DOMException}}.
+    1. Skip the [$related origins validation procedure$] for |effectiveDomain|.
+    1. Let |clientDataJSON| be equal to |remoteClientDataJSON|.
+         1. The [=user agent=] MUST NOT add, remove, or modify any contents of |remoteClientDataJSON| when constructing |clientDataJSON|,
+               as doing so would invalidate the remote machine's calculated |clientDataJSON|, resulting in a signature mismatch at the [=[RP]=].
+
+:: [=authentication extension|authentication=] processing changes
```
emlun: We have this formatting precedent now from the PRF extension:
```suggestion
: Client extension processing ([=authentication extension|authentication=])
::
```

index.bs
```diff
@@ -7747,6 +7752,107 @@ However, [=authenticators=] that do not utilize [[!FIDO-CTAP]] do not necessaril
...
+    </xmp>
+
+: Client extension processing
+:: [=registration extension|registration=] processing changes
+    1. If the result of [=getting the current permission state=] for `"publickey-credentials-remote-client-data-json"` and the [=current settings object=] is not `"granted"`: [[!Permissions]]
+         1. Throw a "{{NotAllowedError}}" {{DOMException}}
+    1. If <code>|pkOptions|.{{PublicKeyCredentialRequestOptions/rpId}}</code> is not present
+         1. Throw a "{{NotAllowedError}}" {{DOMException}}
+    1. Let |collectedClientData| be the result of [=parse a JSON string to an Infra value=] given |remoteClientDataJSON|. [[!infra]]
+         1. If this throws, return a {{DOMException}} whose name is "{{NotSupportedError}}".
```
emlun: Let's be more consistent with the exception wording.

```suggestion
         1. If this throws, throw a "{{NotSupportedError}}" {{DOMException}}.
```

index.bs
```diff
@@ -7747,6 +7752,107 @@ However, [=authenticators=] that do not utilize [[!FIDO-CTAP]] do not necessaril
...
+    1. Let |remoteOrigin| be the value of the `"origin"` key of |collectedClientData|.
+    1. Verify that <code>|pkOptions|.{{PublicKeyCredentialCreationOptions/rp}}.{{PublicKeyCredentialRpEntity/id}}</code>
+          [=is a registrable domain suffix of or is equal to=] |remoteOrigin|, substituting |remoteOrigin| for |effectiveDomain| in that check.
+         1. If this check fails, throw a "{{NotAllowedError}}" {{DOMException}}.
+    1. Skip the [$related origins validation procedure$] for |effectiveDomain|.
+    1. Let |clientDataJSON| be equal to |remoteClientDataJSON|.
```
emlun: Similarly, I think these steps need to say something to the effect of "in step 8 of `PublicKeyCredential.create()`, do not support related origin requests" and "in step 14 of `PublicKeyCredential.create()`, set |clientDataJSON| to |remoteClientDataJSON| instead". Otherwise we're just creating a new unused `|clientDataJSON|` variable and instructing to skip a step that, strictly speaking, should already have executed before the extension processing does.

index.bs
```diff
@@ -7747,6 +7752,107 @@ However, [=authenticators=] that do not utilize [[!FIDO-CTAP]] do not necessaril
...
+    };
+    </xmp>
+
+: Client extension processing
+:: [=registration extension|registration=] processing changes
+    1. If the result of [=getting the current permission state=] for `"publickey-credentials-remote-client-data-json"` and the [=current settings object=] is not `"granted"`: [[!Permissions]]
+         1. Throw a "{{NotAllowedError}}" {{DOMException}}
+    1. If <code>|pkOptions|.{{PublicKeyCredentialRequestOptions/rpId}}</code> is not present
+         1. Throw a "{{NotAllowedError}}" {{DOMException}}
+    1. Let |collectedClientData| be the result of [=parse a JSON string to an Infra value=] given |remoteClientDataJSON|. [[!infra]]
```
emlun: I think the reference brackets belong adjacent to what's referring them, and it looks especially weird to put one after the colon.

I don't think either of these are really necessary since the autolinks already link to them and add the specs and terms to the reference index, but they don't hurt either. Perhaps `INFRA` should be capitalized to match how it appears in the index, though.

```suggestion
    1. If the result of [=getting the current permission state=] [[!Permissions]] for `"publickey-credentials-remote-client-data-json"` and the [=current settings object=] is not `"granted"`:
         1. Throw a "{{NotAllowedError}}" {{DOMException}}
    1. If <code>|pkOptions|.{{PublicKeyCredentialRequestOptions/rpId}}</code> is not present
         1. Throw a "{{NotAllowedError}}" {{DOMException}}
    1. Let |collectedClientData| be the result of [=parse a JSON string to an Infra value=] [[!INFRA]] given |remoteClientDataJSON|.
```

index.bs
```diff
@@ -7747,6 +7752,107 @@ However, [=authenticators=] that do not utilize [[!FIDO-CTAP]] do not necessaril
...
+:: None
+
+: Authenticator extension processing
+:: None
+
+: Authenticator extension input / output
+:: None
+
+: Client extension output
+:: None
```
emlun: [§9.2. Defining Extensions](https://w3c.github.io/webauthn/#sctn-extension-specification):

>Any [client extension](https://w3c.github.io/webauthn/#client-extension) that is processed by the client MUST return a [client extension output](https://w3c.github.io/webauthn/#client-extension-output) value so that the [WebAuthn Relying Party](https://w3c.github.io/webauthn/#webauthn-relying-party) knows that the extension was honored by the client. [...] If an extension does not otherwise require any result values, it SHOULD be defined as returning a JSON Boolean [client extension output](https://w3c.github.io/webauthn/#client-extension-output) result, set to true to signify that the extension was understood and processed.

:)

index.bs
```diff
@@ -8589,6 +8695,29 @@ or where [=client=] and [=authenticator=] do not communicate directly,
 designers SHOULD consider how this affects the enforcement of [=scope=] restrictions
 and the strength of the [=authenticator=] as a [=something you have=] authentication factor.
 
+## Security Considerations for the remoteClientDataJSON Extension ## {#sctn-remote-clientdatajson-security}
```
emlun: I would title this just "The remoteClientDataJSON Extension" since it's already a subsection of "Security Considerations". I would also move it to after its sibling subsections, since those concern the core API while this is an optional extension.

index.bs
```diff
@@ -7747,6 +7752,107 @@ However, [=authenticators=] that do not utilize [[!FIDO-CTAP]] do not necessaril
...
+instead, the grant is administered through device management policy (e.g., managed browser profiles, managed browser instances, or
+managed devices that enumerate specific [=Remote desktop web client=] origins) or
+through explicit per-origin user opt-in in [=WebAuthn Client=] settings.
+This permission MUST be configured per-origin and MUST NOT offer the capability to permit all origins.
+
+: Permissions
+:: This extension defines a [=powerful feature=] ([[!Permissions]]) and a [=policy-controlled feature=] ([[!Permissions-Policy]]), both identified by the name <code>"publickey-credentials-remote-client-data-json"</code>.
+    The [=default allowlist=] for this [=policy-controlled feature=] is `'none'`.
+    The [=default permission state=] for this feature is `"denied"`, meaning the extension is not available by default and must be explicitly enabled.
+    Since no additional aspects are defined, the [=permission descriptor type=] defaults to {{PermissionDescriptor}}.
```
emlun: Please [use semantic line breaks](https://github.com/w3c/webauthn/blob/44ad61ea568cf0b3121b1d1881f55a4a2e83abd1/CONTRIBUTING.md#semantic-line-breaks). For example, here's how I would do it:

```suggestion
:: This extension defines a [=powerful feature=] ([[!Permissions]]) and a [=policy-controlled feature=] ([[!Permissions-Policy]]),
    both identified by the name <code>"publickey-credentials-remote-client-data-json"</code>.
    The [=default allowlist=] for this [=policy-controlled feature=] is `'none'`.
    The [=default permission state=] for this feature is `"denied"`,
    meaning the extension is not available by default and must be explicitly enabled.
    Since no additional aspects are defined, the [=permission descriptor type=] defaults to {{PermissionDescriptor}}.
```

This applies throughout the PR but I won't point out everywhere I think it could be improved (it's a bit subjective, and not worth that much noise). I can submit a meta-PR with more exhaustive proposals, though.

index.bs
```diff
@@ -7747,6 +7752,107 @@ However, [=authenticators=] that do not utilize [[!FIDO-CTAP]] do not necessaril
...
+               as doing so would invalidate the remote machine's calculated |clientDataJSON|, resulting in a signature mismatch at the [=[RP]=].
+
+
+: Authenticator extension input
+:: None
+
+: Authenticator extension processing
+:: None
+
+: Authenticator extension input / output
```
emlun: This duplicates the input section just above.
```suggestion
: Authenticator extension output
```

index.bs
```diff
@@ -7747,6 +7752,107 @@ However, [=authenticators=] that do not utilize [[!FIDO-CTAP]] do not necessaril
...
+    <div class="example" id="example-publickey-credentials-remote-client-data-json-permission-query">
+    A [=Remote desktop web client=] may check whether the [=user agent=] has been configured
+    to permit this extension for the current origin via the {{Permissions}} API:
+
+    <xmp highlight="javascript">
+    const status = await navigator.permissions.query({
+        name: "publickey-credentials-remote-client-data-json"
+    });
+    if (status.state === "granted") {
+        // The UA is configured to allow the remoteClientDataJSON extension; proceed with the RDP flow.
```
emlun: "Remote desktop" is a generic enough term IMO, but RDP is commonly understood to refer to a particular proprietary protocol, right? That's my first association, at least.

```suggestion
        // The client is configured to allow the remoteClientDataJSON extension; proceed with the remote desktop flow.
```

emlun (changes requested):
Sorry for the delay! I've no objections to this on a technical level, but here are some editorial comments and change proposals. I can submit a meta-PR with these and some less-substantive polish if you'd like?

index.bs
akshayku: Added

index.bs
akshayku: Referenced.

akshayku:
@emlun / @kreichgauer , I have updated the steps based on WG discussion.  Can you take a look again?

index.bs
```diff
@@ -7747,6 +7760,153 @@ However, [=authenticators=] that do not utilize [[!FIDO-CTAP]] do not necessaril
...
+
+This allows web-based remote desktop web clients to execute remote WebAuthn requests locally. Here the [=Remote desktop web client=] is calling WebAuthn request on behalf of
+RP present on the remote machine and have different origin than the calling RP in the remote machine.
+
+Note: This extension is a [=powerful feature=] that requires explicit per-origin permission before use.
+[=User agents=] SHOULD NOT prompt the user for this permission at request time;
+instead, the grant is administered through device management policy (e.g., managed browser profiles, managed browser instances, or
+managed devices that enumerate specific [=Remote desktop web client=] origins) or
+through explicit per-origin user opt-in in [=WebAuthn Client=] settings.
+This permission MUST be configured per-origin and MUST NOT offer the capability to permit all origins.
```
emlun: Per #1979 and #2196, we shouldn't put normative requirements in `Note:`s. This text seems quite important, though. Would it make sense to just drop the `Note:` and keep the rest of the text as is?

index.bs
```diff
@@ -7747,6 +7760,153 @@ However, [=authenticators=] that do not utilize [[!FIDO-CTAP]] do not necessaril
...
+
+: Permissions
+:: This extension defines a [=powerful feature=] ([[!Permissions]]) and a [=policy-controlled feature=] ([[!Permissions-Policy]]),
+    both identified by the name <code>"publickey-credentials-remote-client-data-json"</code>.
+    The [=default allowlist=] for this [=policy-controlled feature=] is `'none'`.
+    The [=default permission state=] for this feature is `"denied"`,
+    meaning the extension is not available by default and must be explicitly enabled.
+    Since no additional aspects are defined, the [=permission descriptor type=] defaults to {{PermissionDescriptor}}.
+
+    The [=permission descriptor type=] for this feature is:
```
emlun: Is this supposed to be immediately followed by the example? I was expecting some form of normative definition here, not an example.

index.bs
```diff
@@ -7747,6 +7760,153 @@ However, [=authenticators=] that do not utilize [[!FIDO-CTAP]] do not necessaril
...
+        1. Let |collectedClientData| be the result of [=parse a JSON string to an Infra value=]
+            given |remoteClientDataJSON|.
+            1. If above parsing step fails, throw a "{{NotSupportedError}}" {{DOMException}}.
+        1. Let |remoteOrigin| be the value of the `"origin"` key of |collectedClientData|.
+        1. If <code>|pkOptions|.{{PublicKeyCredentialCreationOptions/rp}}.{{PublicKeyCredentialRpEntity/id}}</code>
+            [=is not a registrable domain suffix of and is not equal to=] |remoteOrigin|, and if the client
+            1. does not support [[#sctn-related-origins|related origin requests]]
+                1. Throw a "{{SecurityError}}" {{DOMException}}.
+            1. supports [[#sctn-related-origins|related origin requests]]
+                1. Skip [$related origins validation procedure$] for this operation.
```
emlun: Since this always skips the related origins validations procedure, do we need the "if the client [supports|does not support]" branch? Can we not just always throw `SecurityError` (and maybe also instruct to skip related origins for clarity)?

emlun: There was discussion on the 2026-04-15 WG call that we might need related origins to work for this, though. I think the scenario was:

If the remote RP `remote-example.org` sets `rp.id = "partner.com"` and `partner.com` declares `remote-example.org` as a related origin, then the remote client will be sending `rp.id = "partner.com"` and `remoteClientData = {"origin": "https://remote-example.org", ...}`. So then by the current logic, the local client will check that `partner.com` is a _registrable domain suffix of or is equal to_ `remote-example.org`, which will fail if the local client does not consider related origins.

index.bs
```diff
@@ -7747,6 +7760,153 @@ However, [=authenticators=] that do not utilize [[!FIDO-CTAP]] do not necessaril
...
+:: None
+
+: Authenticator extension processing
+:: None
+
+: Authenticator extension output
+:: None
+
+: Client extension output
+:: Returns the boolean to indicate whether parsing all the extension steps resulted in a success or failure.
```
emlun: The extension will always throw an exception before having a chance to return a `false` output, right? Let's use the same wording as in the `appidExclude` extension.

```suggestion
:: Returns the value [TRUE] to indicate to the [=[RP]=] that the extension was acted upon.
```

index.bs
```diff
@@ -8621,10 +8781,35 @@ possible for [=[RPS]=] to trust any further [=attestation statements=] from the
...
+### The remoteClientDataJSON Extension ### {#sctn-remote-clientdatajson-security}
+
+The [=remoteClientDataJSON=] extension enables a [=Remote desktop web client=] to proxy
+[=Web Authentication API=] requests on behalf of a remote [=[RP]=]. This fundamentally
+alters the trust model of the [=Web Authentication API=] in the following ways:
+
+- **Origin substitution:** The [=user agent=] accepts a {{CollectedClientData/origin}} supplied by the calling
+    [=Remote desktop web client=] rather than deriving it from the current execution context.
+    Any [=user agent=] granting this permission must therefore trust that the caller
+    accurately supplies the origin of the remote [=[RP]=].
```
emlun: Let's make this a BCP 14 "MUST", and also qualify that the caller must be trusted to be _honest_ in addition to correct.

```suggestion
    Any [=user agent=] granting this permission MUST therefore trust that the caller
    accurately and honestly supplies the origin of the remote [=[RP]=].
```

index.bs
```diff
@@ -8621,10 +8781,35 @@ possible for [=[RPS]=] to trust any further [=attestation statements=] from the
...
+    permission globally to all origins. The permission MUST be restricted on a per-origin basis
+    via [=user agent=] policy.
+
+[=User agents=] that grant this extension SHOULD clearly communicate to users when a
+remote-desktop-proxied WebAuthn operation is in progress, so that users understand
+that their authenticator is being used on behalf of a remotely hosted application.
+
 <!-- no sec cons for clients enumerated at this time
 ## Security considerations for [=clients=] ## {#sctn-security-considerations-client}
 -->
```
emlun: This can now be deleted! :tada:

```suggestion
```

emlun (changes requested):

index.bs
akshayku: Removing the Note.

index.bs
akshayku: The above line says that no additional aspects are defined resulting in default behavior. I think this line can be removed. Please let me know if this is incorrect. Not entirely sure what else needs to be said here.

index.bs
akshayku: Changed.

index.bs
akshayku: Changed.

index.bs
akshayku: Removed.

index.bs
akshayku: If such a case arises, remote client has already done the related origins checks. 

This condition is saying that if `partner.com` is **_NOT_** _a registrable domain suffix of is equal to_ `remote-example.org` and client supports related origins in general, accept the client data JSON as it is as that is already done on the remote side. 

IIRC, Nina/Martin were saying that we can't even do related origins checks on client side in certain contexts.

emlun: Oooh, right. That is indeed what this logic is saying. My bad!

Perhaps a note would be helpful hear to point that out, though? Maybe something like:
```suggestion
            [=is not a registrable domain suffix of and is not equal to=] |remoteOrigin|, and if the client
            1. does not support [[#sctn-related-origins|related origin requests]]
                1. Throw a "{{SecurityError}}" {{DOMException}}.
            1. supports [[#sctn-related-origins|related origin requests]]
                1. Skip [$related origins validation procedure$] for this operation.
                
                    Note: This means the local client is delegating the [$related origins validation procedure$] to the remote client.
```

Hm, that also gets me thinking: does there need to be a signal here, from the remote client to the local client, saying whether the remote client supports related origins? I'm thinking otherwise there's the possibility that the remote client _doesn't_ support related origins, but the local client blindly assumes that it does. Could that be a problem?

I suppose you wouldn't get this far with a mismatched origin if the remote client didn't support related origins, as then it would never forward to the local client in the first place. So probably this can indeed be inferred (assuming an honest remote client). I just wanted to call that out anyway to confirm - what do you think?

index.bs
emlun: Oh wait, sorry, we still need the heading. :facepalm: Fixed in 81003f4ec75e393a3a41e553efaafd45206e0966.

index.bs
emlun: Ok, that looks good to me. I'm not familiar with [Permissions], but that looks to me like it's sufficient.

index.bs
akshayku: There is no real verifiable way for client to confirm what kind of client is there on the remote side. 

Assuming an honest client, if remote client does not support related origins and there is a mismatch in origin, that call will never come to the client side.

akshayku: Added the note.

emlun: Yeah... and for example adding that signal in `remoteClientDataJSON` would also only work assuming the remote client is honest (otherwise it could just lie about that signal anyway). So yeah, that wouldn't do anything. Thanks for discussing!

pascoej (approved):

emlun (approved):
There's some more editorial polish (punctuation, wordsmithing and such) I'd like to make, but we can do that as separate follow-up PRs.

ve7jtb (approved):
LGTM

timcappalli (approved):
LGTM2

index.bs
kreichgauer: Doesn't the client need to make a fetch to determine whether the RP "supports related origin requests" though? And that fetch would potentially violate the CSP. So ISTM the client can't really impose any restrictions on the origin/RP ID tuple, i.e. the SecurityError case needs to go. But maybe I'm misunderstanding something.

index.bs
```diff
@@ -7747,6 +7760,153 @@ However, [=authenticators=] that do not utilize [[!FIDO-CTAP]] do not necessaril
...
+        1. If the result of [=getting the current permission state=] for
+            `"publickey-credentials-remote-client-data-json"` and the [=current settings object=]
+            is not `"granted"`.
+            1. Throw a "{{NotAllowedError}}" {{DOMException}}.
+        1. If <code>|pkOptions|.{{PublicKeyCredentialCreationOptions/rp}}.{{PublicKeyCredentialRpEntity/id}}</code>
+            is not present
+            1. Throw a "{{NotAllowedError}}" {{DOMException}}.
+        1. Let |collectedClientData| be the result of [=parse a JSON string to an Infra value=]
+            given |remoteClientDataJSON|.
+            1. If above parsing step fails, throw a "{{NotSupportedError}}" {{DOMException}}.
```
kreichgauer: Would EncodingError fit better here?

```suggestion
            1. If above parsing step fails, throw a "{{EncodingError}}" {{DOMException}}.
```

index.bs
```diff
@@ -8621,9 +8781,33 @@ possible for [=[RPS]=] to trust any further [=attestation statements=] from the
...
+    [=Remote desktop web client=] rather than deriving it from the current execution context.
+    Any [=user agent=] granting this permission MUST therefore trust that the caller
+    accurately and honestly supplies the origin of the remote [=[RP]=].
+
+- **RP ID validation:** The [=RP ID=] is validated against the origin present in |remoteClientDataJSON|
+    rather than the [=effective domain=] of the local [=client=]'s origin.
+    If content running in the [=Remote desktop web client=] origin can influence the
+    |remoteClientDataJSON| value, it could cause assertions valid for an unintended [=[RP]=] to be produced.
+
+- **Per-origin grant required:** For the above reasons, [=user agents=] MUST NOT grant this
```
kreichgauer: This is a bit confusing since permissions are [origin keyed]( https://www.w3.org/TR/permissions/#dfn-permission-key-type), so they're per-origin by their spec definition basically.

I think what this is probably meant to express is that when designing user agent configuration mechanisms (e.g. UA settings or enterprise policy) to grant this permissions, user agents should make sure that the mechanism is reasonably fine-grained.

index.bs
akshayku: You are right that client can't really impose any restrictions on the origin/RPID tuple.

But SecurityError is only for situation where Client does not support related origins (for whatever reason) and there is a mismatch between RPID is NOT a registrable domain suffix of and is not equal to remoteOrigin. Such clients never supported related origins in the first place and would have resulted in error on remote machine also with their client. Hence mismatch of origins on client side is an error to them also.

SecurityError is not for clients which supports related origins. In that case, client is supposed to trust that remote machine has done the due diligence. 

@kreichgauer The step it is replacing distinguishes between clients which supports related origins and clients which does not. Hence, I have mentioned that case here.

index.bs
akshayku: I thought stating it as configuration MUST NOT be for all origins needs to be clarified here. If it is already implicitly enforced, then probably we don't need to over-emphasize here. I changed the language. Let me know if that works.

index.bs
kreichgauer: Apologies, I misread the text to say "if the _RP_ supports related origin" instead of the client. This makes more sense now.

I'm still not entirely convinced this should throw a SecurityError though.

Ignoring related origins for a second, what if the local side is a Windows Desktop, and the remote side is an Android device, and the caller origin actually is a native Android app? The origin would be something like `android:apk-key-hash:<hash>` and the RP ID would be `acme.com`. The Android side would have determined that the caller app is authorized to claim acme.com credentials via asset linking, not related origins. But the Desktop obviously doesn't understand APK asset links and has no way to validate that relationship. So should whether this request completes successfully really depend on whether the client supports related origin requests?

akshayku: Yes, currently as this proposal stands today, technically this type of request will work only on clients who supports related origin requests. 

I don't see any asset related logic in the spec which clarifies how a general client on Android does this today in a non-remote scenario. Maybe we need some clarification in general in the spec for asset linking even for native non-remote case? But that is a separate issue all together. 

Thinking more on this, from the perspective of this remote scenario and this extension, as those checks have already happened on the remote side on honest client, client side really can't enforce any such checks similar to related origins checks and local client that implements this remoteClientDataJSON extension has appropriate permissions/managed policies in place, I am OK with local client skipping the check of `RPID` against `remoteOrigin` in all cases irrespective of whether client supports related origins or not. 

Will that work?

emlun: > I am OK with local client skipping the check of `RPID` against `remoteOrigin` in all cases

Hmyeah, I suppose that's the only way to make this work for all combinations of local and remote client. We'll need to also update the security considerations to clearly point out that this gives up _all_ control of RP ID/origin validation to the remote client.

akshayku: @kreichgauer I have updated the PR. Can you take a look?

akshayku:
@kreichgauer Can you review again?

kreichgauer (approved):

akshayku:
Thanks everyone.

PR #2377, opened 2026-01-08
Add "Credential Manager Trust Group (CMTG) Key" extension

### LINKED ISSUE #2338: Add Credential Manager Trust Group Key (CMTG) extension
Explainer: https://github.com/w3c/webauthn/wiki/Explainer:-Credential-Manager-Trust-Group-Key

timcappalli:
Closes #2338 

Implementation commitment:

Documentation and checks

- [ ] Affects privacy
- [ ] Affects security
- [x] Updated explainer ([link](https://github.com/w3c/webauthn/wiki/Explainer:-Credential-Manager-Trust-Group-Key))

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/pull/2377.html" title="Last updated on Jul 20, 2026, 8:32 PM UTC (c0a4bef)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/2377/e6f24bd...c0a4bef.html" title="Last updated on Jul 20, 2026, 8:32 PM UTC (c0a4bef)">Diff</a>


index.bs
```diff
@@ -7738,15 +7753,220 @@ However, [=authenticators=] that do not utilize [[!FIDO-CTAP]] do not necessaril
...
+[=Credential Manager Trust Group private key=] (`cmtgKey1`).
+
+A subsequent {{CredentialsContainer/get()|get()}} operation (with the `cmtgKey` extension) on `dev1`
+using the same credential and the same authenticator/credential manager (`cm1`), 
+will include a signature from the same [=Credential Manager Trust Group private key=], `cmtgKey1`.
+
+`cmtgKey1` is only synced to an authenticator/credential manager on another device (`dev2`)
+if a non-remote relationship exists between the source (`dev1`) and destination (`dev2`) device.
+
+If a third device (`device 3`) is introduced (with `cm1` configured)
```
nsatragno: nit: Let's pick "devN" or "device N" and run with that consistently.

index.bs
```diff
@@ -7738,15 +7753,220 @@ However, [=authenticators=] that do not utilize [[!FIDO-CTAP]] do not necessaril
...
+This is done each time this [=cmtgKey=] extension is included with either a 
+{{CredentialsContainer/create()|navigator.credentials.create()}} or {{CredentialsContainer/get()|navigator.credentials.get()}} call.
+
+#### Relying Party Usage #### {#sctn-cmtg-key-extension-usage}
+
+This extension is intended for use by those [=[RPS]=] employing risk-analysis systems informing their sign-in decisions.
+
+This extension provides a "trust group" signal <i>when used consistently</i> with both
+{{CredentialsContainer/create()|navigator.credentials.create()}} and {{CredentialsContainer/get()|navigator.credentials.get()}} operations.
+
```
nsatragno: Let's also add some javascript examples, both on the spec and the explainer. The explainer doesn't have any javascript for processing a response.

index.bs
```diff
@@ -7738,15 +7753,220 @@ However, [=authenticators=] that do not utilize [[!FIDO-CTAP]] do not necessaril
...
+If the `cmtgKey` extension was included on a {{CredentialsContainer/get()|navigator.credentials.get()}} call,
+then the below verification steps are performed in the context of <a href=#authn-ceremony-verify-extension-outputs>this step</a> of [[#sctn-verifying-assertion]]
+using these variables established therein: |credential|, |clientExtensionResults|, |authData|, and |hash|. 
+[=[RP]=] policy may specify whether a response without a `cmtgKey` is acceptable.
+
+1. Let |attObjForCmtgKey| be the value of the `cmtgKey` member of |clientExtensionResults|.
+
+1. Verify that |attObjForCmtgKey| is valid CBOR conforming to the syntax defined above
+    and perform CBOR decoding on it to extract the contained fields: |signature|, |cmtgKey|.
```
nsatragno: Related to the comment above, |signature| and |cmtgKey| are not CBOR members, they're javascript members. Perhaps you wanted to have the RP verify the integrity of the CBOR encoded COSE public key?

index.bs
```diff
@@ -7738,15 +7753,220 @@ However, [=authenticators=] that do not utilize [[!FIDO-CTAP]] do not necessaril
...
+
+    Note: The latter |attObjForCmtgKey| fields are referenced exclusively in the below steps 
+    and are not to be confused with other fields with the same names in other portions of [=authenticator data=].
+
+1. Verify that {{AuthenticationExtensionsCmtgKeyOutputs/signature}} is a valid signature over the [=assertion signature=] [input](#fig-signature)
+    (i.e. `authData` and `hash`) by the [=Credential Manager Trust Group public key=] |cmtgKey|.
+    (The signature algorithm is the same as for the [=user credential=].)
+
+1. If the [=[RP]=]'s [=user account=] mapped to the <code>|credential|.{{Credential/id}}</code> in play (i.e., for the user being authenticated)
+    hold a `cmtgKey` value corresponding to the extracted |attObjForCmtgKey| fields,
```
nsatragno:
```suggestion
    holds a `cmtgKey` value corresponding to the extracted |attObjForCmtgKey| fields,
```

index.bs
```diff
@@ -7738,15 +7753,220 @@ However, [=authenticators=] that do not utilize [[!FIDO-CTAP]] do not necessaril
...
+    and perform CBOR decoding on it to extract the contained fields: |signature|, |cmtgKey|.
+
+    Note: The latter |attObjForCmtgKey| fields are referenced exclusively in the below steps 
+    and are not to be confused with other fields with the same names in other portions of [=authenticator data=].
+
+1. Verify that {{AuthenticationExtensionsCmtgKeyOutputs/signature}} is a valid signature over the [=assertion signature=] [input](#fig-signature)
+    (i.e. `authData` and `hash`) by the [=Credential Manager Trust Group public key=] |cmtgKey|.
+    (The signature algorithm is the same as for the [=user credential=].)
+
+1. If the [=[RP]=]'s [=user account=] mapped to the <code>|credential|.{{Credential/id}}</code> in play (i.e., for the user being authenticated)
```
nsatragno: You should refer to the credential record object here as well.

index.bs
```diff
@@ -7738,15 +7753,220 @@ However, [=authenticators=] that do not utilize [[!FIDO-CTAP]] do not necessaril
...
+
+    Note: The latter |attObjForCmtgKey| fields are referenced exclusively in the below steps
+    and are not to be confused with other fields with the same names in other portions of the top-level [=attestation object=].
+
+1. Verify that {{AuthenticationExtensionsCmtgKeyOutputs/signature}} is a valid signature over the [=assertion signature=] [input](#fig-signature)
+    (i.e. `authData` and `hash`) by the [=Credential Manager Trust Group Key=] |cmtgKey|.
+    (The signature algorithm is the same as for the [=user credential=].)
+
+1. Complete the steps from [[#sctn-registering-a-new-credential]] and, if those steps are successful, 
+    store the |cmtgKey| value indexed to the <code>|credential|.{{Credential/id}}</code> in the [=user account=].
```
nsatragno: You should refer to the credential record object here instead.

In fact, the credential record object should be updated to add CMTG public keys.

index.bs
```diff
@@ -7738,15 +7753,220 @@ However, [=authenticators=] that do not utilize [[!FIDO-CTAP]] do not necessaril
...
+
+    </dl>
+
+1. Otherwise, the [=[RP]=] does not have |attObjForCmtgKey| fields presently mapped to this [=user account=]
+    and <code>|credential|.{{Credential/id}}</code> pair:
+
+    1. Store the extracted |cmtgKey| value indexed to the <code>|credential|.{{Credential/id}}</code> in the [=user account=].
+        Terminate these verification steps.
 
+See also [[#sctn-cmtg-key-extension-usage]].
```
nsatragno: Would it be possible to add virtual authenticator support for this feature? It feels it should be pretty simple, and it'll be required to write WPTs.

nsatragno (commented):
Thank you so much for doing this work! Here's a first pass at a review.

index.bs
timcappalli: Issue created. I think we can handle that as a separate PR.

https://github.com/w3c/webauthn/issues/2395

index.bs
```diff
@@ -7738,15 +7753,220 @@ However, [=authenticators=] that do not utilize [[!FIDO-CTAP]] do not necessaril
...
+This is done by creating an additional [=public key credential source=]-specific key pair
+in the authenticator/credential manager, if such a key pair does not already exist for 
+the [=public key credential source=] being created or exercised,
+and returning the [=Credential Manager Trust Group public key=]
+along with a signature by the [=Credential Manager Trust Group private key=] to the [=[RP]=].
+
+This is done each time this [=cmtgKey=] extension is included with either a 
+{{CredentialsContainer/create()|navigator.credentials.create()}} or {{CredentialsContainer/get()|navigator.credentials.get()}} call.
+
+#### Relying Party Usage #### {#sctn-cmtg-key-extension-usage}
```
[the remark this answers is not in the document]
timcappalli: Hm. Not sure how to really split it. Its not specifying what the authenticator should do, its describing the scenarios. I think that is important for an RP to understand.

index.bs
timcappalli: We don't have any inline code samples anywhere else in the spec (besides section 1).

nsatragno: There are some, [e.g. for the signal API](https://w3c.github.io/webauthn/#sctn-signalAllAcceptedCredentials) and [related origins](https://w3c.github.io/webauthn/#sctn-related-origins). As an implementer I think they're useful, but your call whether to add an example to the spec or not.

Please do make sure the [explainer](https://github.com/w3c/webauthn/wiki/Explainer:-Credential-Manager-Trust-Group-Key) has code examples. They're necessary for review down the line.

index.bs
```diff
@@ -7745,8 +7761,251 @@ However, [=authenticators=] that do not utilize [[!FIDO-CTAP]] do not necessaril
...
+will include a signature from the same [=Credential Manager Trust Group private key=], `cmtgKey1`.
+
+`cmtgKey1` is only synced to an authenticator/credential manager on another device (`dev2`)
+if a non-remote relationship exists between the source (`dev1`) and destination (`dev2`) device.
+
+If a third device (`dev3`) is introduced (with `cm1` configured),
+which does not share a non-remote relationship with `dev1` or `dev2`
+(as evaluated by the authenticator/credential manager, `cm1`), 
+a new [=Credential Manager Trust Group private key=] (`cmtgKey2`) is generated and its public key
+is returned to the [=[RP]=], signaling the start of a separate trust group.
```
nsatragno, line 8015: One thing you might want to note for RPs is that it's possible for CMTG keys to be synced after the first assertion on each device.

Suppose dev1 and dev2 initially aren't within the same trust group. The RP collects cmtgKey1 and cmtgKey2. Later, these two devices are considered to be within the same trust group (e.g. because they've been in close proximity). The credential manager may elect to use cmtgKey1 when running an assertion on dev2 (or vice versa).

index.bs
```diff
@@ -7745,8 +7761,251 @@ However, [=authenticators=] that do not utilize [[!FIDO-CTAP]] do not necessaril
...
+If the `cmtgKey` extension was included on a {{CredentialsContainer/create()|navigator.credentials.create()}} call
+and a [=Credential Manager Trust Group Key=] is returned by an [=authenticator=],
+the below verification steps are performed in the context of <a href=#reg-ceremony-verify-extension-outputs>this step</a> of [[#sctn-registering-a-new-credential]]
+using these variables established therein: |credential|, |clientExtensionResults|, |authData|, and |hash|. 
+
+1. Let |credentialRecord| be the [=credential record=] for the [=credential=].
+
+1. Let |extensionOutput| be the value of the {{AuthenticationExtensionsClientOutputs/cmtgKey}} member of |clientExtensionResults|.
+
+1. Verify that |extensionOutput| is a valid {{AuthenticationExtensionsCmtgKeyOutputs}} dictionary.
```
nsatragno: Is "|var| is a valid {{type}} dictionary" well defined?

In particular, is

```
{
  cmtgKey: ArrayBuffer([...]),
  signature: ArrayBuffer([...]),
  extraKey: ArrayBuffer([...])
}
```

a valid `AuthenticationExtensionsCmtgKeyOutputs` dictionary? We don't want RPs rejecting the CMTG key if there are extra keys.

I suggest we are more specific about what RPs should be validating. You can also omit the need to individually validate that `cmtgKey` and `signature` are present and an ArrayBuffer by making them required on the IDL.

This applies below as well.

index.bs
```diff
@@ -7745,8 +7761,251 @@ However, [=authenticators=] that do not utilize [[!FIDO-CTAP]] do not necessaril
...
+    (The signature algorithm is the same as for the [=user credential=].)
+
+1. If |credentialRecord|.[$credential record/cmtgKeyRecords$] contains an entry |e|
+    where |e|.[=CMTG Key Record/publicKey=] matches |cmtgKey|:
+
+    The [=authenticator=]/[=credential manager=] is part of a previously seen trust group.
+
+    1. Update |e|.[=CMTG Key Record/lastUsed=] to the current time.
+
+    1. Terminate these verification steps.
```
nsatragno: This could be taken to mean to terminate the relying party assertion verification steps are terminated. Remove?

index.bs
```diff
@@ -7745,8 +7761,251 @@ However, [=authenticators=] that do not utilize [[!FIDO-CTAP]] do not necessaril
...
+    This is possibly a new [=Credential Manager Trust Group Key=], signifying a new trust group.
+
+    1. Optionally, perform additional policy checks.
+    
+    1. Let |newEntry| be a new [=CMTG Key Record=] where its [=CMTG Key Record/publicKey=] is |cmtgKey|
+        and its [=CMTG Key Record/lastUsed=] is the current time.
+
+    1. Append |newEntry| to |credentialRecord|.[$credential record/cmtgKeyRecords$].
+
+    1. Terminate these verification steps.
```
nsatragno: Same here.

index.bs
```diff
@@ -7738,15 +7753,220 @@ However, [=authenticators=] that do not utilize [[!FIDO-CTAP]] do not necessaril
...
+
+This is done each time this [=cmtgKey=] extension is included with either a 
+{{CredentialsContainer/create()|navigator.credentials.create()}} or {{CredentialsContainer/get()|navigator.credentials.get()}} call.
+
+#### Relying Party Usage #### {#sctn-cmtg-key-extension-usage}
+
+This extension is intended for use by those [=[RPS]=] employing risk-analysis systems informing their sign-in decisions.
+
+This extension provides a "trust group" signal <i>when used consistently</i> with both
+{{CredentialsContainer/create()|navigator.credentials.create()}} and {{CredentialsContainer/get()|navigator.credentials.get()}} operations.
```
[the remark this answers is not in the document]
nsatragno: This comment is no longer relevant.

index.bs
```diff
@@ -7745,8 +7761,251 @@ However, [=authenticators=] that do not utilize [[!FIDO-CTAP]] do not necessaril
...
+If the `cmtgKey` extension was included on a {{CredentialsContainer/get()|navigator.credentials.get()}} call,
+then the below verification steps are performed in the context of <a href=#authn-ceremony-verify-extension-outputs>this step</a> of [[#sctn-verifying-assertion]]
+using these variables established therein: |credential|, |clientExtensionResults|, |authData|, and |hash|. 
+[=[RP]=] policy may specify whether a response without a `cmtgKey` is acceptable.
+
+1. Let |credentialRecord| be the [=credential record=] for the [=credential=].
+
+1. Let |extensionOutput| be the value of the {{AuthenticationExtensionsClientOutputs/cmtgKey}} member of |clientExtensionResults|.
+
+1. Verify that |extensionOutput| is a valid {{AuthenticationExtensionsCmtgKeyOutputs}} dictionary.
```
nsatragno: An authenticator may not return a CMTG key, in which case this would fail to validate.

Why not:

```
1. If |extensionOutput| is not present,
  1. Optionally, perform additional policy checks.
  1. Terminate the `cmtgKey` validation algorithm. (or similar)
```

nsatragno (commented):
Thank you for your changes. A few smaller comments on the new text:

nsatragno (approved):
Fantastic work, thank you! Approved.

akshayku (approved):

nsatragno:
@timcappalli did we forget to add a client capability for this extension?

timcappalli:
> @timcappalli did we forget to add a client capability for this extension?

@nsatragno No. Extensions don't have explicit capability entries in the spec.

> The set of [keys](https://infra.spec.whatwg.org/#map-getting-the-keys) SHOULD also contain a key for each [extension](https://www.w3.org/TR/webauthn-3/#sctn-extensions) implemented by the client, where the key is formed by prefixing the string extension: to the [extension identifier](https://www.w3.org/TR/webauthn-3/#extension-identifier). The associated value for each implemented extension SHOULD be true.

nsatragno:
@timcappalli at the risk of being wrong again again: did we forget to add JSON serialization for CMTG? We don't seem to have a partial [AuthenticationExtensionsClientInputsJSON](https://w3c.github.io/webauthn/#dictdef-authenticationextensionsclientinputsjson) dictionary defined for this extension (or [AuthenticationExtensionsClientOutputsJSON](https://w3c.github.io/webauthn/#dictdef-authenticationextensionsclientoutputsjson)).

index.bs
```diff
@@ -7745,8 +7761,370 @@ However, [=authenticators=] that do not utilize [[!FIDO-CTAP]] do not necessaril
...
+and a [=Credential Manager Trust Group Key=] is returned by an [=authenticator=],
+the below verification steps are performed in the context of <a href=#reg-ceremony-verify-extension-outputs>this step</a> of [[#sctn-registering-a-new-credential]]
+using these variables established therein: |credential|, |clientExtensionResults|, |authData|, and |hash|. 
+
+1. Let |credentialRecord| be the [=credential record=] for the [=credential=].
+
+1. Let |extensionOutput| be the value of the {{AuthenticationExtensionsClientOutputs/cmtgKey}} member of |clientExtensionResults|.
+
+1. If |extensionOutput| does not contain the required members of {{AuthenticationExtensionsCmtgKeyOutputs}},
+    Terminate this `cmtgKey` validation algorithm.
```
emlun: Should we clarify here to terminate with a failure result?

index.bs
```diff
@@ -7745,8 +7761,370 @@ However, [=authenticators=] that do not utilize [[!FIDO-CTAP]] do not necessaril
...
+
+    NOTE: [=[RP]=]s MUST ignore any keys in |extensionOutput| that are not defined in the {{AuthenticationExtensionsCmtgKeyOutputs}} dictionary.
```
emlun: Per #1979 and #2196, we should not have normative language in NOTEs.

index.bs
```diff
@@ -7745,8 +7761,370 @@ However, [=authenticators=] that do not utilize [[!FIDO-CTAP]] do not necessaril
...
+
+1. Let |signature| be |extensionOutput|.{{AuthenticationExtensionsCmtgKeyOutputs/signature}}.
+    Verify that |signature| is a valid ArrayBuffer.
```
emlun: This verification doesn't seem actionable to me - how would the RP verify whether it's a valid ArrayBuffer? Presumably the RP JavaScript would never reach this step with `|signature|` being anything else than a an `ArrayBuffer` instance, unless the client is malfunctioning horribly.

I think we can just cut this line altogether.

index.bs
```diff
@@ -7745,8 +7761,370 @@ However, [=authenticators=] that do not utilize [[!FIDO-CTAP]] do not necessaril
...
+
+1. Let |cmtgKey| be |extensionOutput|.{{AuthenticationExtensionsCmtgKeyOutputs/cmtgKey}}.
+    Verify that |cmtgKey| is a valid ArrayBuffer containing a COSE_Key object.
```
emlun: Ditto, I don't think verifying "valid ArrayBuffer" seems actionable.

```suggestion
    Verify that |cmtgKey| contains a well-formed COSE_Key object.
```

emlun: Or perhaps this should already instruct to parse the COSE_Key, so that step 7 below has `|cmtgKey|` to work with as a public key instead of an `ArrayBuffer`.

```suggestion
    Parse |cmtgKey| as a COSE_Key.
```

index.bs
```diff
@@ -7745,8 +7761,370 @@ However, [=authenticators=] that do not utilize [[!FIDO-CTAP]] do not necessaril
...
+
+1. Let |credentialRecord| be the [=credential record=] for the [=credential=].
+
+1. Let |extensionOutput| be the value of the {{AuthenticationExtensionsClientOutputs/cmtgKey}} member of |clientExtensionResults|.
+
+1. If |extensionOutput| is not present or does not contain the required members of {{AuthenticationExtensionsCmtgKeyOutputs}}
+
+    1. Optionally, perform additional policy checks.
+
+    1. Terminate this `cmtgKey` validation algorithm.
```
emlun: At first glance I interpreted this as "terminate the whole assertion verification as a failure", but now realize this is meant to just terminate the `cmtgKey` validation and defer to those other policy checks instead. Would it make sense to somehow highlight here to "terminate _these_ validation steps as inconclusive", to pre-empt that misreading?

index.bs
```diff
@@ -7745,8 +7761,370 @@ However, [=authenticators=] that do not utilize [[!FIDO-CTAP]] do not necessaril
...
+:: The Boolean value [TRUE] to indicate that this extension is requested by the [=[RP]=].
+    <xmp class="idl">
+    partial dictionary AuthenticationExtensionsClientInputs {
+      boolean cmtgKey;
+    };
+    </xmp>
+
+: Client extension processing
+:: For both [=authenticatorMakeCredential=] and [=authenticatorGetAssertion=] operations:
+    1. Create the authenticator extension input from the client extension input.
```
emlun, line 8120: What if the input is `cmtgKey: false`? Should that be interpreted the same as no input, or throw an exception? (I notice now that we also have the same issue in `credProps`.)

arnar: Yes, in general "parameterless" extensions that are requested with just a boolean should not be processed if that boolean is set to false.

[hmacSecret](https://fidoalliance.org/specs/fido-v2.0-id-20180227/fido-client-to-authenticator-protocol-v2.0-id-20180227.html#sctn-hmac-secret-extension) in CTAP has the same form also, and does specify the above.

timcappalli: Do we need to be more explicit about this or is it good as is?

index.bs
timcappalli: Added further down in this section.

index.bs
nsatragno: Explicit > Implicit every time in specs IMO.

nsatragno (approved):
Approved, thank you!

index.bs
```diff
@@ -7745,8 +7761,377 @@ However, [=authenticators=] that do not utilize [[!FIDO-CTAP]] do not necessaril
...
+: Authenticator extension processing
+:: For both [=authenticatorMakeCredential=] and [=authenticatorGetAssertion=] operations:
+    1. Create or select the [=public key credential source=] as usual (see [[#sctn-op-make-cred]],
+        or [[#sctn-op-get-assertion]] as appropriate).
+
+    1. If the [=public key credential source=] is not [=backup eligible=],
+        terminate these processing steps as this extension only applies to [=multi-device credentials=].
+
+    1. If a [=Credential Manager Trust Group Key=] does not already exist for this ([=public key credential source/id|Credential ID=],
+        [=public key credential source/rpId|RP ID=], [=public key credential source/userHandle|User Handle=]) tuple in the [=authenticator=]/[=credential manager=],
```
MasterKale, line 8185: We should probably update https://w3c.github.io/webauthn/#authenticatormakecredential to note that, if Step 4 here...

Image

...overwrites an existing credential `cred1` with `cred2`, such that `(cred1, rpID, userHandle)` will no longer ever be a valid tuple for the authenticator, the authenticator should clean up any existing CMTG keys for `cred1`.

MasterKale (approved):

index.bs
```diff
@@ -7745,8 +7761,377 @@ However, [=authenticators=] that do not utilize [[!FIDO-CTAP]] do not necessaril
...
+        create it using the same public key algorithm as that used by the [=user credential=]'s [=credential key pair=],
+        otherwise locate the existing [=Credential Manager Trust Group Key=].
+
+    1. Let |cmtgPublicKey| be the newly created or existing [=Credential Manager Trust Group Key=],
+        in COSE_Key format [=credentialPublicKey|in the same fashion as for the user credential's credentialPublicKey=].
+
+    1. Let |cmtgPrivateKey| be the newly created or existing [=Credential Manager Trust Group private key=].
+
+    1. Let |cmtgKeySig| be the result of signing the [=assertion signature=] [input](#fig-signature) with |cmtgPrivateKey|,
```
nsatragno: This is ambiguous in a subtle way. The assertion signature input is the `authenticatorData + clientDataHash`. However, the `authenticatorData` is not the same for the authenticator get assertion and make credential operations.

I think for make credential, you mean to use the computed `authenticatorData` with the `attestedCredentialData`. This makes the most sense to me.

So, saying "assertion signature input" for make credential seems subtly wrong. Why not spell out what is signed exactly?

index.bs
```diff
@@ -353,14 +353,14 @@ Beyond that, the intended audiences for this document are the following main gro
 
 ## Use Cases ## {#sctn-use-cases}
 
-The below use case scenarios illustrate use of two very different types of [=authenticators=] and credentials across two common deployment types, 
-as well as outline further scenarios. Additional scenarios, including sample code, are given later in [[#sctn-sample-scenarios]]. 
+The below use case scenarios illustrate use of two very different types of [=authenticators=] and credentials across two common deployment types,
```
nsatragno, line 358: Adding this as a comment to remind ourselves to fix the build issues as well, since it seems that all errors left are specific to this PR.

MasterKale: Did you happen to pull in the latest master? #2442 got merged last week that got the spec able to use the latest Bikeshed 7 so builds started working again. There were changes to an upstream Bikeshed data repo that prevented our older build setup from working at all. I'll bet if you merge in master to this branch you might see stuff start building again.

index.bs
timcappalli: Is it weird to put text about an optional extension in these main operations?

index.bs
```diff
@@ -7903,8 +7919,385 @@ This permission MUST be configured per-origin and MUST NOT offer the capability
...
+
+: Client extension input
+:: The Boolean value [TRUE] to indicate that this extension is requested by the [=[RP]=].
+    <xmp class="idl">
+    partial dictionary AuthenticationExtensionsClientInputs {
+      boolean cmtgKey;
+    };
+    partial dictionary AuthenticationExtensionsClientInputsJSON {
+      boolean cmtgKey;
+}   ;
```
nsatragno: nit: remove the extraneous whitespace.

index.bs
```diff
@@ -7903,8 +7919,385 @@ This permission MUST be configured per-origin and MUST NOT offer the capability
...
+
+If a {{AuthenticationExtensionsClientOutputs/cmtgKey}} extension output is present in a {{CredentialsContainer/create()|navigator.credentials.create()}} call,
+the below verification steps are performed in the context of <a href=#reg-ceremony-verify-extension-outputs>this step</a> of [[#sctn-registering-a-new-credential]]
+using these variables established therein: |credential|, |clientExtensionResults|, |authData|, and |hash|.
+
+1. Let |credentialRecord| be the [=credential record=] for |credential|.
+
+1. Let |extensionOutput| be the value of <code>|clientExtensionResults|.{{AuthenticationExtensionsClientOutputs/cmtgKey}}</code>.
+
+1. If |extensionOutput| does not contain the required members of {{AuthenticationExtensionsCmtgKeyOutputs}},
```
nsatragno: nit: `extensionOutput` may not be present.

1. If |extensionOutput| is not present or |extensionOutput| does not contain...

nsatragno (approved):
Approved with two nits.

Thank you so much for taking on this feature. I really appreciate it!

index.bs
MasterKale: Okay, I see the concern. Maybe we add a garbage collection step to CMTG processing instead, so CMTG keys with an invalid/replaced credential source get deleted the next time the authenticator creates or exercises a CMTG key 🤔

index.bs
```diff
@@ -7903,8 +7919,385 @@ This permission MUST be configured per-origin and MUST NOT offer the capability
...
+        in COSE_Key format [=credentialPublicKey|in the same fashion as for the user credential's credentialPublicKey=].
+
+    1. Let |cmtgPrivateKey| be the newly created or existing [=Credential Manager Trust Group private key=].
+
+    1. If this algorithm is being run as part of [=authenticatorMakeCredential=]:
+
+        1. Let |cmtgKeySig| be the result of signing the concatenation of |authenticatorData| (as constructed in [[#sctn-generating-an-attestation-object]],
+            which includes |attestedCredentialData|) and |clientDataHash|, with |cmtgPrivateKey|,
+            using the same signature algorithm as that used by the [=user credential=]'s [=credential key pair=].
+
```
MasterKale, line 8199: Related to https://github.com/w3c/webauthn/pull/2377/changes#r3624840460, this might be a good place to add an additional step, after this existing one, to iterate over existing CMTG keys and ensure that the authenticator is still managing a credential source for each CMTG key's corresponding `([=public key credential source/id|Credential ID=], [=public key credential source/rpId|RP ID=], [=public key credential source/userHandle|User Handle=])` tuple. If not, delete the CMTG key.

index.bs
MasterKale: I think maybe here: https://github.com/w3c/webauthn/pull/2377/changes#r3624947487

bjorn3:
Why does the PR description not have "Affects privacy" checked? This would allow distinguishing between multiple devices of the same user without the user explicitly consenting to it. That is a fingerprintable signal, right?

index.bs
```diff
@@ -7903,8 +7919,385 @@ This permission MUST be configured per-origin and MUST NOT offer the capability
...
+1. Let |credentialRecord| be the [=credential record=] for |credential|.
+
+1. Let |extensionOutput| be the value of <code>|clientExtensionResults|.{{AuthenticationExtensionsClientOutputs/cmtgKey}}</code>.
+
+1. If |extensionOutput| is not present or does not contain the required members of
+    {{AuthenticationExtensionsCmtgKeyOutputs}}, terminate this `cmtgKey` validation algorithm.
+
+    NOTE: [=[RPS]=] MUST ignore any keys in |extensionOutput| that are not defined in the {{AuthenticationExtensionsCmtgKeyOutputs}} dictionary.
+
+1. Parse |cmtgKey| as a COSE_Key.
```
harshlal028, line 8234: [nit] Add explicit extraction steps before step 4:

Let |cmtgKey| be |extensionOutput|.cmtgKey.
Let |signature| be |extensionOutput|.signature.
Verify |cmtgKey| and |signature| are valid ArrayBuffers.

index.bs
```diff
@@ -7903,8 +7919,385 @@ This permission MUST be configured per-origin and MUST NOT offer the capability
...
+
+1. Verify that the algorithm used for |signature| matches the COSE algorithm identifier of the [=credentialPublicKey=]
+    within the [=attestedCredentialData=] of |authData|.
+
+1. Verify that |signature| is a valid signature over the [=assertion signature=] [input](#fig-signature)
```
harshlal028, line 8239: Registration produces an attestation object, not an assertion. Referring to the "assertion signature input" during registration might be confusing.

index.bs
```diff
@@ -7903,8 +7919,385 @@ This permission MUST be configured per-origin and MUST NOT offer the capability
...
+    (i.e. |authData| and |hash|) by the [=Credential Manager Trust Group Key=] |cmtgKey|.
+
```
harshlal028, line 8241: Maybe it would be worth explicitly calling out the case if verification fails.

index.bs
```diff
@@ -7903,8 +7919,385 @@ This permission MUST be configured per-origin and MUST NOT offer the capability
...
+1. Let |signature| be |extensionOutput|.{{AuthenticationExtensionsCmtgKeyOutputs/signature}}.
+    Verify that |signature| is a valid ArrayBuffer.
+
+1. Let |cmtgKey| be |extensionOutput|.{{AuthenticationExtensionsCmtgKeyOutputs/cmtgKey}}.
+    Verify that |cmtgKey| is a valid ArrayBuffer containing a COSE_Key object.
+
+1. Verify that the algorithm used for |signature| matches the algorithm of the public key
+    previously stored for the |credentialRecord|.
+
+1. Verify that |signature| is a valid signature over the [=assertion signature=] [input](#fig-signature)
```
harshlal028, line 8278: Maybe it would be worth explicitly calling out the case if verification fails.

index.bs
```diff
@@ -7903,8 +7919,385 @@ This permission MUST be configured per-origin and MUST NOT offer the capability
...
+1. Otherwise:
+
+    This is possibly a new [=Credential Manager Trust Group Key=], signifying a new trust group.
+
+    1. Optionally, perform additional policy checks.
+
+    1. Let |newEntry| be a new [=CMTG Key Record=] where its [$CMTG Key Record/publicKey$] is |cmtgKey|
+        and its [$CMTG Key Record/lastUsed$] is the current time.
+
+    1. Append |newEntry| to |credentialRecord|.[$credential record/cmtgKeyRecords$].
```
harshlal028, line 8298: Would it make sense to add that appending |newEntry| to `cmtgKeyRecords` should only occur if the RP's policy allows accepting the new trust group (e.g., following successful step-up or risk checks) is passed.

index.bs
```diff
@@ -7903,8 +7919,385 @@ This permission MUST be configured per-origin and MUST NOT offer the capability
...
+        1. Let |cmtgKeySig| be the result of signing the concatenation of |authenticatorData| (as constructed in [[#sctn-generating-an-attestation-object]],
+            which includes |attestedCredentialData|) and |clientDataHash|, with |cmtgPrivateKey|,
+            using the same signature algorithm as that used by the [=user credential=]'s [=credential key pair=].
+
+    1. If this algorithm is being run as part of [=authenticatorGetAssertion=]:
+
+        1. Let |cmtgKeySig| be the result of signing the [=assertion signature|assertion signature input=] with |cmtgPrivateKey|,
+            using the same signature algorithm as that used by the [=user credential=]'s [=credential key pair=].
+
+        Note: The [=assertion signature=] [input](#fig-signature) and |cmtgKeySig| covers the [=[RP]=]'s {{PublicKeyCredentialCreationOptions/challenge}}
```
harshlal028, line 8205: Maybe it should be `PublicKeyCredentialRequestOptions/challenge` given it is assertion.

index.bs
```diff
@@ -1062,6 +1063,21 @@ BCP 14 [[!RFC2119]] [[!RFC8174]] when, and only when, they appear in all capital
     Note: The [=credential public key=] is referred to as the [=user public key=] in FIDO UAF [[UAFProtocol]], and in FIDO U2F
     [[FIDO-U2F-Message-Formats]] and some parts of this specification that relate to it.
 
+: <dfn>Credential Manager Trust Group Key</dfn>
+: <dfn>Credential Manager Trust Group Private Key</dfn>
+: <dfn>Credential Manager Trust Group Public Key</dfn>
+: <dfn>Credential Manager Trust Group Key Pair</dfn>
+:: A [=Credential Manager Trust Group Key=], is a [=credential manager=]-, [=[RP]=]-, and [=user credential=]-specific
```
harshlal028, line 1070: [nit] Extra comma "," can be deleted on line 1070 after [=Credential Manager Trust Group Key=]

index.bs
```diff
@@ -7903,8 +7919,385 @@ This permission MUST be configured per-origin and MUST NOT offer the capability
...
+#### Relying Party Usage #### {#sctn-cmtg-key-extension-usage}
+
+This extension is intended for use by those [=[RPS]=] employing risk-analysis systems informing their sign-in decisions.
+
+This extension provides a "trust group" signal <i>when used consistently</i> with both
+{{CredentialsContainer/create()|navigator.credentials.create()}} and {{CredentialsContainer/get()|navigator.credentials.get()}} operations.
+
+When a [=[RP]=] uses the `cmtgKey` extension during a {{CredentialsContainer/create()|create()}} option on the first device (`dev1`),
+the selected [=authenticator=]/[=credential manager=] (`cm1`) generates a new [=Credential Manager Trust Group Key=] pair.
+The extension output includes the public key and a signature generated using the
```
harshlal028, line 7948: If an RP includes `cmtgKey: true` in extensions, but the client or authenticator does not support CMTG then no cmtgKey will be returned. Might be worth making it explicit. wdyt?

harshlal028 (approved):

index.bs
```diff
@@ -7903,8 +7919,385 @@ This permission MUST be configured per-origin and MUST NOT offer the capability
...
+        otherwise locate the existing [=Credential Manager Trust Group Key=].
+
+    1. Let |cmtgPublicKey| be the newly created or existing [=Credential Manager Trust Group Key=],
+        in COSE_Key format [=credentialPublicKey|in the same fashion as for the user credential's credentialPublicKey=].
+
+    1. Let |cmtgPrivateKey| be the newly created or existing [=Credential Manager Trust Group private key=].
+
+    1. If this algorithm is being run as part of [=authenticatorMakeCredential=]:
+
+        1. Let |cmtgKeySig| be the result of signing the concatenation of |authenticatorData| (as constructed in [[#sctn-generating-an-attestation-object]],
```
emlun, line 8196: Is `|authenticatorData|` here meant to be the attestation object? Otherwise this seems to be to be identical to the corresponding step for getAssertion.

index.bs
```diff
@@ -7903,8 +7919,385 @@ This permission MUST be configured per-origin and MUST NOT offer the capability
...
+         and that would imply that the signature signs over itself.
+
+#### `cmtgKey` Extension Output Verification Procedures #### {#sctn-cmtg-key-extension-verification}
+
+Verifying the <code>[=cmtgKey=]</code> extension output is performed by the [=[RP]=] whenever a [=Credential Manager Trust Group Key=] is returned within the extension output.
+
+##### Registration (`create()`) ##### {#sctn-cmtg-key-extension-verification-create}
+
+If a {{AuthenticationExtensionsClientOutputs/cmtgKey}} extension output is present in a {{CredentialsContainer/create()|navigator.credentials.create()}} call,
+the below verification steps are performed in the context of <a href=#reg-ceremony-verify-extension-outputs>this step</a> of [[#sctn-registering-a-new-credential]]
```
emlun, line 8222: For the sake of the text making sense even if the hyperlinks get dropped (perhaps unlikely, but might be relevant for plain text quotations if nothing else), I think we should refer to this as a "step X" number. That's also the style currently used throughout the document.

index.bs
```diff
@@ -7903,8 +7919,385 @@ This permission MUST be configured per-origin and MUST NOT offer the capability
...
+using these variables established therein: |credential|, |clientExtensionResults|, |authData|, and |hash|.
+
+1. Let |credentialRecord| be the [=credential record=] for |credential|.
+
+1. Let |extensionOutput| be the value of <code>|clientExtensionResults|.{{AuthenticationExtensionsClientOutputs/cmtgKey}}</code>.
+
+1. If |extensionOutput| is not present or does not contain the required members of
+    {{AuthenticationExtensionsCmtgKeyOutputs}}, terminate this `cmtgKey` validation algorithm.
+
+    NOTE: [=[RPS]=] MUST ignore any keys in |extensionOutput| that are not defined in the {{AuthenticationExtensionsCmtgKeyOutputs}} dictionary.
```
emlun, line 8232: Per #1979, the normative MUST should not be in a NOTE block.

index.bs
```diff
@@ -7903,8 +7919,385 @@ This permission MUST be configured per-origin and MUST NOT offer the capability
...
+
+1. Parse |cmtgKey| as a COSE_Key.
+
+1. Verify that the algorithm used for |signature| matches the COSE algorithm identifier of the [=credentialPublicKey=]
```
emlun, line 8236: `|signature|` is undefined here. I think this is meant to be `<code>|extensionOutput|.{{AuthenticationExtensionsCmtgKeyOutputs/signature}}</code>`?

```suggestion
1. Verify that the algorithm used for <code>|extensionOutput|.{{AuthenticationExtensionsCmtgKeyOutputs/signature}}</code> matches the COSE algorithm identifier of the [=credentialPublicKey=]
```
(This should then also apply to further references to `|signature|` below.)

index.bs
```diff
@@ -7903,8 +7919,385 @@ This permission MUST be configured per-origin and MUST NOT offer the capability
...
+
+1. Let |extensionOutput| be the value of <code>|clientExtensionResults|.{{AuthenticationExtensionsClientOutputs/cmtgKey}}</code>.
+
+1. If |extensionOutput| is not present or does not contain the required members of {{AuthenticationExtensionsCmtgKeyOutputs}}:
+
+    1. Optionally, perform additional policy checks.
+
+    1. Terminate these validation steps as inconclusive.
+
+    NOTE: [=[RPS]=] MUST ignore any keys in |extensionOutput| that are not defined in the {{AuthenticationExtensionsCmtgKeyOutputs}} dictionary.
```
emlun, line 8267: Per #1979, the normative MUST should not be in a NOTE block.

index.bs
```diff
@@ -7903,8 +7919,385 @@ This permission MUST be configured per-origin and MUST NOT offer the capability
...
+
+1. Verify that the algorithm used for |signature| matches the algorithm of the public key
+    previously stored for the |credentialRecord|.
+
+1. Verify that |signature| is a valid signature over the [=assertion signature=] [input](#fig-signature)
+    (i.e. `authData` and `hash`) by the [=Credential Manager Trust Group public key=] |cmtgKey|.
+    (The signature algorithm is the same as for the [=user credential=].)
+
+1. If |credentialRecord|.[$credential record/cmtgKeyRecords$] contains an entry |e|
+    where |e|.[$CMTG Key Record/publicKey$] matches |cmtgKey|:
```
emlun, line 8283: "matches" might be a dangerous word here as it could be interpreted several ways (for example, a COSE_Key with `kid` might "match" a COSE_Key without `kid` but other members equal). But since [§2.4](https://www.w3.org/TR/2026/REC-webauthn-3-20260825/#sctn-conforming-all-classes) requires CTAP2 canonical CBOR for everything, and [§6.5.1](https://www.w3.org/TR/2026/REC-webauthn-3-20260825/#sctn-attested-credential-data) further restricts the encoding of these public keys to forbid optional values (like `kid`), it should be safe to assume that all encodings of the same key should be byte-identical. So it should be safe to specify here that `publicKey` and _cmtgKey_ should be equal byte arrays, which eliminates any ambiguity.

index.bs
```diff
@@ -7903,8 +7919,385 @@ This permission MUST be configured per-origin and MUST NOT offer the capability
...
+
+This extension provides a "trust group" signal <i>when used consistently</i> with both
+{{CredentialsContainer/create()|navigator.credentials.create()}} and {{CredentialsContainer/get()|navigator.credentials.get()}} operations.
+
+When a [=[RP]=] uses the `cmtgKey` extension during a {{CredentialsContainer/create()|create()}} option on the first device (`dev1`),
+the selected [=authenticator=]/[=credential manager=] (`cm1`) generates a new [=Credential Manager Trust Group Key=] pair.
+The extension output includes the public key and a signature generated using the
+[=Credential Manager Trust Group private key=] (`cmtgKey1`).
+
+<div class="example">
```
emlun, line 7951: Both of these two examples look like most of it is the usual base WebAuthn stuff, just with the addition of a `cmtgKey: true` extension input, right? If so, then I don't think these examples seem very helpful as is. They give an initial impression that you need a lot of code to use this extension, while actually the `cmtgKey` part is a very small fraction of the example volume (at least on the client side). It might be more helpful to reduce the example to concentrate on the extension parts specifically, and perhaps also include some abbreviated pseudocode for what needs to happen on the server side too.

index.bs
timcappalli: This is implied as extension support is optional, but I can add an explicit call out.

index.bs
timcappalli: @emlun what do you think of the updated example?

timcappalli: @emlun what do you think of the updated example?

PR #2380, opened 2026-01-26
Add Cross Device Fallback URL extension for hybrid transport authentication

### LINKED ISSUE #2341: Add hybrid fallback URL extension
## Description

Add a fallbackURLextension to support Hybrid Cross device Passkeys not found use cases.

## Related Links

https://github.com/fido-alliance/fido-2-specs/issues/1543

harshlal028:
Closes #2341

Implementation commitment:

- [ ] Chromium ([509934168](https://issues.chromium.org/issues/509934168))

Documentation and checks

- [x] Affects privacy
- [x] Affects security
- [x] Updated explainer ([link](https://github.com/w3c/webauthn/wiki))

***
<a href="https://pr-preview.s3.amazonaws.com/harshlal028/webauthn/pull/2380.html" title="Last updated on Aug 26, 2026, 5:53 PM UTC (757653e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/2380/146d0db...harshlal028:757653e.html" title="Last updated on Aug 26, 2026, 5:53 PM UTC (757653e)">Diff</a>


index.bs
```diff
@@ -7740,6 +7740,55 @@ However, [=authenticators=] that do not utilize [[!FIDO-CTAP]] do not necessaril
...
+
+Note: This extension is restricted to {{AuthenticatorTransport/hybrid}}
+
+: Extension identifier
+:: `fallbackUrl`
+
+: Operation applicability
+:: [=authentication extension|authentication=]
+
+: Client extension input
```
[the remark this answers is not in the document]
harshlal028: Done.

index.bs
```diff
@@ -7740,6 +7740,55 @@ However, [=authenticators=] that do not utilize [[!FIDO-CTAP]] do not necessaril
...
+    };
+    </xmp>
+
+: Client extension processing
+:: When [[##sctn-getAssertion|sending an assertion request over {{AuthenticatorTransport/hybrid}}]]:
+        1. If the [=authenticator=] does not find a credential:
+            1. Run the [$related origins validation procedure$] with arguments |[=RP ID=]| and |[=fallbackUrl=]|. 
+                1. If the result is [FALSE], throw a "{{SecurityError}}" {{DOMException}}.
+                1. Otherwise Deploy [=fallbackUrl=] on the [=authenticator=] device
+        1. Otherwise ignore the [=fallbackUrl=] and continue with normal processing.
```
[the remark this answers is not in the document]
harshlal028: Done.

index.bs
```diff
@@ -7740,6 +7740,55 @@ However, [=authenticators=] that do not utilize [[!FIDO-CTAP]] do not necessaril
...
+
+: Client extension output
+:: Returns the value [TRUE] to indicate to the [=[RP]=] that the extension was acted upon.
```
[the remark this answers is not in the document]
harshlal028: There is a generic effort in this regards: https://github.com/w3c/webauthn/issues/2096

index.bs
```diff
@@ -7745,7 +7745,76 @@ However, [=authenticators=] that do not utilize [[!FIDO-CTAP]] do not necessaril
 
 This section defines extensions that are both [=client extensions=] and [=authenticator extensions=].
 
-This section is currently empty.
+### Cross Device Fallback URL extension (<dfn>crossDeviceFallbackUrl</dfn>) ### {#sctn-cross-device-fallback-url-extension}
+This [=client extension|client=] [=authentication extension=] allows a [=[RP]=] to include a URL to fallback to for authentication in case a valid credential is not found on the local [=authenticator=].
```
timcappalli:
```suggestion
This [=authenticator extension|authenticator=] [=authentication extension=] allows a [=[RP]=] to include a fallback sign in URL for cases where a valid credential is not found on the remote [=authenticator=].
```

index.bs
```diff
@@ -7745,7 +7745,76 @@ However, [=authenticators=] that do not utilize [[!FIDO-CTAP]] do not necessaril
...
+
+Note: This extension is restricted to {{AuthenticatorTransport/hybrid}}
```
timcappalli:
```suggestion
This extension is restricted to {{AuthenticatorTransport/hybrid}}
```

index.bs
nsatragno: > the local [=authenticator=].

The authenticator is not local to the client device (it's the opposite, no?). This should say that the extension is useful if a valid credential is not found on a roaming authenticator contacted via hybrid.

index.bs
nsatragno: nit: full stop at the end of the sentence, please.

index.bs
```diff
@@ -7745,7 +7745,76 @@ However, [=authenticators=] that do not utilize [[!FIDO-CTAP]] do not necessaril
...
+      DOMString crossDeviceFallbackUrl;
+    };
+    partial dictionary AuthenticationExtensionsClientInputsJSON {
+      DOMString crossDeviceFallbackUrl;
+    };
+    </xmp>
+
+: Client extension processing
+:: When [[##sctn-getAssertion|receives an assertion request]]:
+        1. The [=client platform=] strips out the [=crossDeviceFallbackUrl=], and then passes the request to enabled credential managers.
```
nsatragno: "enabled credential managers" is not defined.

Note also that the "client platform" is the _browser_ side, not the authenticator side. But this extension is intended to do something on the "authenticator" (which to be fair, sorta works like another client platform. But in the context of webauthn, it's not one).

This needs to be rephrased in terms of the authenticator checking for its credentials.

index.bs
```diff
@@ -7745,7 +7745,76 @@ However, [=authenticators=] that do not utilize [[!FIDO-CTAP]] do not necessaril
...
+        1. If a [=credential=] is found:
+            1. Continue with normal processing.
+        1. Otherwise:
+            1. Run the [$related origins validation procedure$] with arguments |[=RP ID=]| and |[=crossDeviceFallbackUrl=]|.
```
nsatragno: Shouldn't you also check if it's a registrable domain suffix first? Why jump to remote validation?

nsatragno: Using the RP ID and not the origin here is tricky. The algorithm expects origins, not RPIDs, and these are not the same. For example, there's a distinction between https and http origins. Consider transforming the RP ID into an origin through some algorithm.

Also, please add a note or something that you're using the algorithm backwards intentionally. I had to do a double take.

nsatragno: Do we need remote validation? Is there a specific problem we want to solve with it?

This can lead to very confusing behaviour for RPs.

Suppose an RP has three domains:
legacy.com
site.com
site-auth.com

The RP ID is legacy.com. The site supports WebAuthn on both site.com and site-auth.com.

legacy.com has a related origins file containing site.com and site-auth.com. This is enough for calls coming from site.com and site-auth.com to succeed. Great!

Now, the RP adds fallbackUrl to all their webauthn invocations. The fallback is to site-auth.com. When site.com uses fallbackUrl, the hybrid authenticator will make a request to site-auth.com, _not_ legacy.com. I suspect besides being confusing, this might have other consequences, but I can't readily think about what they are.

index.bs
```diff
@@ -7745,7 +7745,76 @@ However, [=authenticators=] that do not utilize [[!FIDO-CTAP]] do not necessaril
...
+
+: Client extension processing
+:: When [[##sctn-getAssertion|receives an assertion request]]:
+        1. The [=client platform=] strips out the [=crossDeviceFallbackUrl=], and then passes the request to enabled credential managers.
+        1. If a [=credential=] is found:
+            1. Continue with normal processing.
+        1. Otherwise:
+            1. Run the [$related origins validation procedure$] with arguments |[=RP ID=]| and |[=crossDeviceFallbackUrl=]|. 
+                1. If the result is [FALSE]:
+                    1. Throw a "{{SecurityError}}" {{DOMException}}.
```
nsatragno: Note authenticators don't throw errors.

index.bs
```diff
@@ -7745,7 +7745,76 @@ However, [=authenticators=] that do not utilize [[!FIDO-CTAP]] do not necessaril
 
 This section defines extensions that are both [=client extensions=] and [=authenticator extensions=].
 
-This section is currently empty.
+### Cross Device Fallback URL extension (<dfn>crossDeviceFallbackUrl</dfn>) ### {#sctn-cross-device-fallback-url-extension}
```
nsatragno: Your PR name says this is the fallbackUrl extension. The PR name should match what's here.

index.bs
```diff
@@ -7745,7 +7745,76 @@ However, [=authenticators=] that do not utilize [[!FIDO-CTAP]] do not necessaril
...
+:: A single DOMString specifying a fallback URL.
+    <xmp class="idl">
+    partial dictionary AuthenticationExtensionsAuthenticatorInputs {
+      DOMString crossDeviceFallbackUrl;
+    };
+    </xmp>
+
+: Authenticator extension processing
+:: When [[##sctn-getAssertion|sending an assertion request over {{AuthenticatorTransport/hybrid}}]]:
+        1. The remote [=client platform=] strips out the [=crossDeviceFallbackUrl=], and then passes the request to enable credential managers.
```
nsatragno: What does "remote" mean here? The authenticator itself is remote. This seems incorrect?

nsatragno (changes requested):

index.bs
sbweeden: How does `[=client extension|client=] [=authentication extension=]` resolve in human readable text when expanded?

index.bs
```diff
@@ -7745,7 +7745,76 @@ However, [=authenticators=] that do not utilize [[!FIDO-CTAP]] do not necessaril
...
+Note: This extension is restricted to {{AuthenticatorTransport/hybrid}}
+
+: Extension identifier
+:: `crossDeviceFallbackUrl`
+
+: Operation applicability
+:: [=authentication extension|authentication=]
+
+: Client extension input
+:: A single DOMString specifying a fallback URL.
```
sbweeden: Is this fallbackURL necessarily absolute, or could server-relative (say relative to the web origin of the current request?) be permitted?

index.bs
sbweeden: Would allowing a server-relative URL format (relative to the web origin of the current request) help with this use case?

sbweeden (changes requested):
Would like to see clarification on whether the fallbackURL can be server-relative or not.

index.bs
nsatragno: One more issue: the URL should be checked against [CSP](https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/CSP) / LNA.

We need to make sure this doesn't become a mechanism to exfiltrate data to third parties.

nsatragno (changes requested):

index.bs
harshlal028: Done.

index.bs
harshlal028: Done

index.bs
harshlal028: Done.

index.bs
harshlal028: Changed "enabled credential managers" to "credential manager".

Also added authenticator to the if condition text.

index.bs
harshlal028: Discussed this offline. Given the request over Hybrid is - doing this check doesn't make much sense. Removing the section for remote validation check.

harshlal028: Removed the section for remote validation checks.

index.bs
harshlal028: Section removed.

index.bs
harshlal028: Removed "remote"

index.bs
harshlal028: This section has been removed as per later comments.

thx

index.bs
harshlal028: This is absolute, not server-relative. Ref: https://github.com/w3c/webauthn/blob/main/explainers/cross-device-fallback-url.md

index.bs
harshlal028: - [=client extension|client=] links to https://www.w3.org/TR/webauthn-3/#client-extension. An example can be seen in [section 12.4](https://www.w3.org/TR/webauthn-3/#sctn-extensions-reg) in the "Description" section of "WebAuthn Extension Identifier: credProps"

- [[=authentication extension=]] links to https://www.w3.org/TR/webauthn-3/#authentication-extension. An example can be seen in [section 12.3](https://www.w3.org/TR/webauthn-3/#sctn-extensions-reg-update) in the "Description" section of "WebAuthn Extension Identifier: appid"

index.bs
harshlal028: Can you provide an example of data exfiltration you are thinking about?

The fallbackURL would typically look like `https://login.example.com/<session_identifier>` which authenticator would wrap into a view intent

```
Intent i = new Intent(Intent.ACTION_VIEW);
i.setData(Uri.parse(fallbackURL));
startActivity(i);
```

This can resolve to a webpage on browser or an android deep link which is auto verified by android [ref](https://developer.android.com/training/app-links/verify-applinks)

Are there concerns with approach.

nsatragno: You can learn more about [CSP](https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/CSP) from the excellent MDN page. Essentially, it's a mechanism for sites to restrict which actions they can take, and among those which pages they are allowed to fetch. CSP reduces the blast radius of XSS vulns by e.g. preventing an attacker from exfiltrating cookies through a `fetch`. Similarly, [Local Network Access](https://developer.chrome.com/blog/local-network-access) protects devices on a local network.

We have https://github.com/w3c/webauthn/issues/2408 open to fix this in the spec for related origins. We should consider doing the same here: specify that the cross device fallback URL must be allowed to be fetched by CSP before the request is allowed to proceed to the authenticator. Obviously we can't just have the authenticator enforce this since it doesn't have access to the page context.

Happy to talk more offline.

harshlal028: Added CSP check. Please take a look.

index.bs
```diff
@@ -2806,6 +2810,45 @@ The following [=simple exceptions=] can be raised:
 
 </dl>
 
+The following custom {{DOMException}} can be raised:
```
nsatragno: This text is misplaced here, as it interrupts the list above. Also, the text you're adding is normative, but it's under a non-normative section.

Move the definition of the CrossDeviceFallbackError somewhere else. Section 5.8 supporting data structures looks like the right place.

index.bs
```diff
@@ -2806,6 +2810,45 @@ The following [=simple exceptions=] can be raised:
...
+
+<xmp class="idl">
+    [Exposed=Window, Serializable]
+    interface CrossDeviceFallbackError : DOMException {
```
nsatragno: Is this really an error if it's expected? `CrossDeviceFallbackException` seems more appropriate.

index.bs
```diff
@@ -2806,6 +2810,45 @@ The following [=simple exceptions=] can be raised:
...
+        constructor(optional DOMString message = "", CrossDeviceFallbackErrorOptions options);
```
nsatragno: Why is the message optional? Might as well make it required.

If you really want this to be optional, you should flip the parameter order. It's very cursed to have the first parameter be optional, although it seems like this is allowed by web IDL (!!!).

index.bs
```diff
@@ -2806,6 +2810,45 @@ The following [=simple exceptions=] can be raised:
...
+        readonly attribute boolean deployed;
+    };
+
+    dictionary CrossDeviceFallbackErrorOptions {
+        required boolean deployed;
+    };
+
+</xmp>
+
+The <code>new CrossDeviceFallbackError(|message|)</code> constructor steps are:
```
nsatragno: Are you not forgetting to take `options` here?

index.bs
```diff
@@ -2806,6 +2810,45 @@ The following [=simple exceptions=] can be raised:
...
+
+1. Set [=this=]'s {{DOMException/name}} to `"CrossDeviceFallbackError"`.
+
+1. Set [=this=]'s {{DOMException/message}} to message.
+
+1. If options["deployed"] is present, set [=this=]'s {{CrossDeviceFallbackError/deployed}} to options["deployed"].
```
nsatragno: \`options["deployed"]\`. (backticks)

nsatragno: Why check if `options["deployed"]` is present? You made it required above.

index.bs
```diff
@@ -2806,6 +2810,45 @@ The following [=simple exceptions=] can be raised:
...
+
+
+{{CrossDeviceFallbackError}} objects are [=serializable objects=].
+
+Their [=serialization steps=], given |value| and |serialized|, are to run the {{DOMException}} [=serialization steps=] given |value| and |serialized|.
```
nsatragno: Doesn't this mean that `deployed` is lost when serialized / deserialized?

index.bs
```diff
@@ -2806,6 +2810,45 @@ The following [=simple exceptions=] can be raised:
...
+
+Their [=deserialization steps=], given |serialized| and |value|, are to run the {{DOMException}} [=deserialization steps=] given |serialized| and |value|.
+
+
+<dl>
+
+    :   {{CrossDeviceFallbackError}}
+    ::  The error is returned when a valid credential is not found and <code>[[#sctn-cross-device-fallback-url-extension]]</code> was processed.
```
nsatragno: >  <code>[[#sctn-cross-device-fallback-url-extension]]</code> was processed

This needs to be expanded to say that the extension was requested, _and the user consented to navigating to the fallback URL_. Otherwise it seems you're making an oracle to determine if a user has credentials or not.

index.bs
```diff
@@ -7745,7 +7788,74 @@ However, [=authenticators=] that do not utilize [[!FIDO-CTAP]] do not necessaril
 
 This section defines extensions that are both [=client extensions=] and [=authenticator extensions=].
 
-This section is currently empty.
+### Cross Device Fallback URL extension (<dfn>crossDeviceFallbackUrl</dfn>) ### {#sctn-cross-device-fallback-url-extension}
+This [=client extension|client=] [=authentication extension=] allows a [=[RP]=] to include a URL to fallback to for authentication in case a valid credential is not found on a [=roaming authenticator=] via {{AuthenticatorTransport/hybrid}}.
```
nsatragno:
```suggestion
This [=client extension|client=] [=authentication extension=] allows a [=[RP]=] to include a URL to fall back to for authentication in case a valid credential is not found on a [=roaming authenticator=] via {{AuthenticatorTransport/hybrid}}.
```

index.bs
```diff
@@ -7745,7 +7788,74 @@ However, [=authenticators=] that do not utilize [[!FIDO-CTAP]] do not necessaril
...
+
+Note: This extension is restricted to {{AuthenticatorTransport/hybrid}}.
+
+: Extension identifier
+:: `crossDeviceFallbackUrl`
+
+: Operation applicability
+:: [=authentication extension|authentication=]
```
nsatragno:
```suggestion
:: [=authentication extension|Authentication=]
```

index.bs
```diff
@@ -7745,7 +7788,74 @@ However, [=authenticators=] that do not utilize [[!FIDO-CTAP]] do not necessaril
...
+
+: Client extension input
+:: A single DOMString specifying a fallback URL.
```
nsatragno: Autolink all types, please.

```suggestion
:: A single {{DOMString}} specifying a fallback URL.
```

index.bs
```diff
@@ -7745,7 +7788,74 @@ However, [=authenticators=] that do not utilize [[!FIDO-CTAP]] do not necessaril
...
+    partial dictionary AuthenticationExtensionsClientInputs {
+      DOMString crossDeviceFallbackUrl;
+    };
+    partial dictionary AuthenticationExtensionsClientInputsJSON {
+      DOMString crossDeviceFallbackUrl;
+    };
+    </xmp>
+
+: Client extension processing
+:: When [[##sctn-getAssertion|receives an assertion request]]:
```
nsatragno: This renders like this:

"When [[##sctn-getAssertion|receives an assertion request]]:"

This is definitely not what you wanted, fix it?

index.bs
```diff
@@ -7745,7 +7788,74 @@ However, [=authenticators=] that do not utilize [[!FIDO-CTAP]] do not necessaril
...
+        1. The [=client platform=] strips out the [=crossDeviceFallbackUrl=], and then passes the request to credential manager.
```
nsatragno: Imagine an idealized WebAuthn ceremony where a user is authenticating with a discoverable credential accessible on their phone. In this case:

* The client platform is the laptop + user agent (browser).
* The phone is the authenticator.

It makes no sense for the UA to strip out the URL. It should really be the authenticator (the phone) doing this.

You might be thinking that it's strange that the phone is an "authenticator" when on Android it usually has a credential manager, and potentially multiple passkey providers. But WebAuthn does not know or handle those!

nsatragno: "Credential manager" is not defined on this document. What does it mean? Did you mean https://w3c.github.io/webappsec-credential-management/#credential-manager? Please define or link to the correct definition.

index.bs
```diff
@@ -7745,7 +7788,74 @@ However, [=authenticators=] that do not utilize [[!FIDO-CTAP]] do not necessaril
...
+        1. If a [=credential=] is found on the [=roaming authenticator=]:
+            1. Continue with normal processing.
+        1. Otherwise:
+            1. [=client platform=] crafts a fake [=request=] to fetch the [=crossDeviceFallbackUrl=] and executes the [=url policy violation=] procedure.
```
nsatragno:
```suggestion
            1. The [=client platform=] crafts a fake [=request=] to fetch the [=crossDeviceFallbackUrl=] and executes the [=url policy violation=] procedure.
```

But what does it mean for a request to be fake? Can you not craft a "real" request and simply not fetch it?

nsatragno: [=request=] points to the wrong definition (render the spec and click it to check), please disambiguate to the right one.

nsatragno: [=crossDeviceFallbackUrl=] here is a concept, not an identifier. You probably meant {{AuthenticationExtensionsClientInputs.crossDeviceFallbackUrl}}.

nsatragno: What is `[=url policy violation=]`? It's not autolinking to anything and I can't find it anywhere. Is it perhaps a hallucination?

index.bs
```diff
@@ -7745,7 +7788,74 @@ However, [=authenticators=] that do not utilize [[!FIDO-CTAP]] do not necessaril
...
+:: A single DOMString specifying a fallback URL.
+    <xmp class="idl">
+    partial dictionary AuthenticationExtensionsAuthenticatorInputs {
+      DOMString crossDeviceFallbackUrl;
+    };
+    </xmp>
+
+: Authenticator extension processing
+:: When [[##sctn-getAssertion|sending an assertion request over {{AuthenticatorTransport/hybrid}}]]:
+        1. The remote [=client platform=] strips out the [=crossDeviceFallbackUrl=], and then passes the request to enable credential managers.
```
nsatragno: This is not an authenticator step?

index.bs
```diff
@@ -7745,7 +7788,74 @@ However, [=authenticators=] that do not utilize [[!FIDO-CTAP]] do not necessaril
...
+        1. If the [=authenticator=] finds a credential that can satisfy the request:
+            1. Ignore the [=crossDeviceFallbackUrl=] extension and continue with normal processing.
+        1. Otherwise:
+            1. Deploy [=crossDeviceFallbackUrl=] on the [=authenticator=] device
```
nsatragno: What does "Deploy [=crossDeviceFallbackUrl=] " mean? This needs to be specified in detail. Basically 90% of the logic of your extension is here.

index.bs
```diff
@@ -7745,7 +7788,74 @@ However, [=authenticators=] that do not utilize [[!FIDO-CTAP]] do not necessaril
...
+            1. Include [=authenticator extension output=] in the response
```
nsatragno: But is there going to be a response? I thought the authenticator would send an error code, in which case there is no authenticator output!

nsatragno (changes requested):

index.bs
harshlal028: Done. Moved to section 5.8

index.bs
harshlal028: Renamed to `CrossDeviceFallbackException` to aligns with the conceptual nature of this as a control-flow exception (handling an expected fallback path) rather than a critical failure.

index.bs
harshlal028: Flipped the order.

index.bs
harshlal028: Added procedure to fix this.

thx

index.bs
harshlal028: Updated.

thx

index.bs
harshlal028: Done.

index.bs
harshlal028: Done.

index.bs
harshlal028: Done.

index.bs
harshlal028: Done.

index.bs
harshlal028: Done.

harshlal028: Updated the verbiage to reflect this.

thx.

harshlal028: Done.

harshlal028: Fixed linking.

index.bs
harshlal028: Updated to return `CrossDeviceFallbackException`

thx

index.bs
harshlal028: Added detail. PTAL.

thx

index.bs
harshlal028: I grouped it under the current section as per (https://github.com/w3c/webauthn/pull/2380#discussion_r2728896029). As this is something that needs to be happen on authenticator device (by the hybrid implementation)-  and there were only two options client and authenticator. Hence Is used the term remote client platform`. Lmk if you have a different suggestion and I can make the change.

index.bs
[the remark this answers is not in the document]
harshlal028: Added `CrossDeviceFallbackException`

index.bs
harshlal028: Done.

index.bs
harshlal028: Fixed. PTAL.

harshlal028: Done.

index.bs
harshlal028: Fixed. PTAL.

thx

index.bs
harshlal028: Addressed. PTAL.

thx

index.bs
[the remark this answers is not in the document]
kreichgauer: You also need to define an extension output value (probably just a bool indicating that the extension was processed).
https://w3c.github.io/webauthn/#sctn-extension-specification

index.bs
```diff
@@ -7745,7 +7802,61 @@ However, [=authenticators=] that do not utilize [[!FIDO-CTAP]] do not necessaril
...
+    };
+    </xmp>
+
+: Client extension processing
+:: When [[#sctn-getAssertion|receives an assertion request]]:
+        1. If the value of {{AuthenticationExtensionsClientInputs/crossDeviceFallbackUrl}} is not an [=absolute-URL string=], ignore the extension and continue with normal processing.
+        1. Otherwise:
+            1. The [=client platform=] constructs a [=request=] with its URL set to the URL specified by {{AuthenticationExtensionsClientInputs/crossDeviceFallbackUrl}} and executes the [=url policy violation=] procedure.
+            1. If the response does not violate, pass the {{AuthenticationExtensionsClientInputs/crossDeviceFallbackUrl}} extension input to the [=authenticator=] and continue processing.
+                1. If the [=authenticator=] returns an error code indicating that a fallback URL is processed, throw a new {{CrossDeviceFallbackException}} with an options dictionary whose {{CrossDeviceFallbackExceptionOptions/deployed}} member is set to true.
```
kreichgauer: The deployed field is only ever set to true. Do we need it at all?

index.bs
```diff
@@ -7745,7 +7802,61 @@ However, [=authenticators=] that do not utilize [[!FIDO-CTAP]] do not necessaril
...
+:: A single text string specifying a fallback URL.
+    <xmp class="example" highlight="cddl">
+    $$extensionInput //= (
+      crossDeviceFallbackUrl: tstr
+    )
+    </xmp>
+
+: Authenticator extension processing
+:: When [[#sctn-getAssertion|sending an assertion request over <code>AuthenticatorTransport/hybrid</code>]]:
+        1. The remote [=authenticator=]'s platform strips out the {{AuthenticationExtensionsClientInputs/crossDeviceFallbackUrl}}, and then passes the request to enable [=credential manager=]s.
```
MasterKale: Maybe this should be "enabled"?

```suggestion
        1. The remote [=authenticator=]'s platform strips out the {{AuthenticationExtensionsClientInputs/crossDeviceFallbackUrl}}, and then passes the request to enabled [=credential manager=]s.
```

index.bs
```diff
@@ -7745,7 +7802,61 @@ However, [=authenticators=] that do not utilize [[!FIDO-CTAP]] do not necessaril
...
+        1. If the [=authenticator=] finds a credential that can satisfy the request:
```
MasterKale: There's two authenticators in the mix with this feature, I suggest the text be more explicit about which authenticator is the focus for a given step of the process:

```suggestion
        1. If the remote [=authenticator=] finds a credential that can satisfy the request:
```

index.bs
```diff
@@ -2806,6 +2815,13 @@ The following [=simple exceptions=] can be raised:
 
 </dl>
 
+<dl>
+
+    :   {{CrossDeviceFallbackException}}
```
kreichgauer: This is essentially a global namespace name, so I think we should probably prefix this with "WebAuthentication" or something similar. 

More generally, would it make sense for WebAuthn to define a single WebAuthnOperationError with a custom operationErrorCode or something along those lines? I know there have been requests to provide better error distinction in WebAuthn even for existing features in #2096, and @emlun suggested something similar there (https://github.com/w3c/webauthn/issues/2096#issuecomment-2233307689). So perhaps we can do something more extensible here.

harshlal028: Does that have to be a part of this PR. Seems like it would be conflating the current PR. 

@kreichgauer I am open to name change tho. Please let me know a name you would prefer and I can update the PR to use that.

kreichgauer: It would need to be part of this PR or one that lands ahead of this PR. We can't really change it after the fact.

harshlal028: Changed name to "WebAuthnCrossDeviceFallbackException".

index.bs
harshlal028: Thanks for the catch. Done.

index.bs
harshlal028: Applied suggestion. Thanks.

index.bs
harshlal028: Removed.

timcappalli:
@harshlal028 can you merge in `main` so that the PR preview starts working again?

harshlal028:
> @harshlal028 can you merge in `main` so that the PR preview starts working again?

Hi @timcappalli,
Here is the `main` branch. Git log after latest git pull

commit 3cf21432001065489e5e6ad904467b3c44741f04 (HEAD -> main, origin/main, origin/HEAD)
Merge: b668301 ddde943
Author: David Waite <david@alkaline-solutions.com>
Date:   Wed Jan 7 15:26:43 2026 -0700

---

Here is my branch `add-fallback-url-extension` which is already on top of main. Git log from my branch

<other commits on my branch>

commit 5c15306777b77e07b4207aaec8ed08bcd29a88a9
Author: Harsh Lal <32895139+harshlal028@users.noreply.github.com>
Date:   Mon Jan 26 18:42:14 2026 +0000

    Add fallbackUrl client extension for hybrid transport authentication

commit 3cf21432001065489e5e6ad904467b3c44741f04 (origin/main, origin/HEAD, main)
Merge: b668301 ddde943
Author: David Waite <david@alkaline-solutions.com>
Date:   Wed Jan 7 15:26:43 2026 -0700


Is there something else you would like me to do.

---

PS: I did build the spec using following commands and I am able to see the generated HTML with changes

```
$ python3 -m venv .python-venv
$ source .python-venv/bin/activate
$ pip install bikeshed==5.4.2
$ bikeshed spec
```

MasterKale:
@harshlal028 You're going to need to update your fork's `main` branch, it's a bit behind the `main` branch here:

Screenshot 2026-07-29 at 11 01 57 AM

Once you do that you can merge your `harshlal028:add-fallback-url-extension` branch with `harshlal028:main` and it should allow previews over here to work again.

harshlal028:
Thanks for the assist Matt. Updated the branch. Preview is working now. :)

christiaanbrand:
Since all the changes are addressed, are we okay to merge this now?

index.bs
```diff
@@ -2812,6 +2821,13 @@ The following [=simple exceptions=] can be raised:
 
 </dl>
 
+<dl>
+
+    :   {{WebAuthnCrossDeviceFallbackException}}
+    ::  The exception is returned when a valid credential is not found on a {{AuthenticatorTransport/hybrid}} device when {{AuthenticationExtensionsClientInputs/crossDeviceFallbackUrl}} extension was requested, and the user consented to navigating to the fallback URL.
```
timcappalli: This text doesn't really make sense. 

It is called an exception, but the description essentially says that the happy path for the fallback URL has been met (e.g. the user said yes).

index.bs
```diff
@@ -4478,6 +4494,34 @@ Note: The {{PublicKeyCredentialHint}} enumeration is deliberately not referenced
...
+{{WebAuthnCrossDeviceFallbackException}} objects are [=serializable objects=].
+
+Their [=serialization steps=], given |value| and |serialized|, are:
+
+1. Run the {{DOMException}} [=serialization steps=] given |value| and |serialized|.
+
+Their [=deserialization steps=], given |serialized| and |value|, are:
+
+1. Run the {{DOMException}} [=deserialization steps=] given |serialized| and |value|.
+
```
timcappalli: This is out of place in the spec. A custom DOM Exception isn't really a supporting data structure. 

Let's create a new section called "Custom DOM Exceptions" under section 5 (Web Authentication API) and call this section Cross-Device Fallback Exception.

index.bs
```diff
@@ -7903,7 +7947,97 @@ This permission MUST be configured per-origin and MUST NOT offer the capability
 
 This section defines extensions that are both [=client extensions=] and [=authenticator extensions=].
 
-This section is currently empty.
+### Cross Device Fallback URL extension ### {#sctn-cross-device-fallback-url-extension}
+This [=client extension|client=] [=authentication extension=] allows a [=[RP]=] to include a URL to fall back to for authentication in case a valid credential is not found on a [=roaming authenticator=] via {{AuthenticatorTransport/hybrid}}.
```
timcappalli:
```suggestion
This [=client extension|client=] [=authentication extension=] allows a [=[RP]=] to include a fallback authentication URL for cases where a valid credential cannot be found on a [=roaming authenticator=] via {{AuthenticatorTransport/hybrid}}.
```

index.bs
```diff
@@ -7903,7 +7947,97 @@ This permission MUST be configured per-origin and MUST NOT offer the capability
...
+
+Note: This extension is restricted to {{AuthenticatorTransport/hybrid}}.
```
timcappalli:
```suggestion
This extension MUST only be used with the {{AuthenticatorTransport/hybrid}} transport.
```

This should be normative

index.bs
timcappalli, line 270: add: 

`spec:html; type:dfn; text:parsing a URL`

index.bs
```diff
@@ -7903,7 +7947,97 @@ This permission MUST be configured per-origin and MUST NOT offer the capability
...
+
+: Client extension processing
+:: When [[#sctn-getAssertion|receives an assertion request]]:
+        1. If the value of {{AuthenticationExtensionsClientInputs/crossDeviceFallbackUrl}} is not an [=absolute-URL string=], ignore the extension and continue with normal processing.
+        1. Otherwise:
+            1. The [=client platform=] constructs a [=request=] with its URL set to the URL specified by {{AuthenticationExtensionsClientInputs/crossDeviceFallbackUrl}} and executes the [=url policy violation=] procedure.
+            1. If the response does not violate, pass the {{AuthenticationExtensionsClientInputs/crossDeviceFallbackUrl}} extension input to the [=authenticator=] and continue processing.
+                1. If the [=authenticator=] returns an error code indicating that a fallback URL is processed, throw a new {{WebAuthnCrossDeviceFallbackException}}.
+                1. Otherwise, continue with normal processing.
+            1. Otherwise, ignore the extension and continue with normal processing.
```
timcappalli:
```suggestion
:: 1. Let |fallbackUrl| be the result of [=parsing a URL=] the value of {{AuthenticationExtensionsClientInputs/crossDeviceFallbackUrl}}.
    1. If |fallbackUrl| is a failure, or if |fallbackUrl|'s [=url/scheme=] is not "https",
        then ignore this extension and continue with normal processing.
    1. Let |effectiveDomain| be the [=effective domain=] of |fallbackUrl|.
    1. If |effectiveDomain| is not a [=valid domain=], ignore this extension and continue with normal processing.
    1. Let |rpId| be the [=Relying Party identifier=] for the current operation.
    1. Let |fallbackUrlOrigin| be the [=origin=] of |fallbackUrl|.
    1. If |rpId| [=is not a registrable domain suffix of and is not equal to=] |effectiveDomain|, and if the client:
        <dl class="switch">
            :   supports [[#sctn-related-origins|related origin requests]]
            ::  1. Run the [$related origins validation procedure$] with arguments |fallbackUrlOrigin| and |rpId|.
                    If the result is [FALSE], ignore this extension and continue with normal processing.
            :   does not support [[#sctn-related-origins|related origin requests]]
            ::  ignore this extension and continue with normal processing.
        </dl>
    1. Construct a [=request=] with its URL set to |fallbackUrl| and execute the [=url policy violation=] procedure.
    1. If the response violates policy, ignore this extension and continue with normal processing.
    1. Pass the {{AuthenticationExtensionsClientInputs/crossDeviceFallbackUrl}} extension input to the [=authenticator=] and continue processing.
    1. If the [=authenticator=] returns an error code indicating that a fallback URL is processed, throw a new {{WebAuthnCrossDeviceFallbackException}}.
    1. Continue with normal processing.
```

index.bs
```diff
@@ -7903,7 +7947,97 @@ This permission MUST be configured per-origin and MUST NOT offer the capability
...
+        1. Otherwise:
+            1. To safeguard WebAuthn security and privacy, the remote [=authenticator=]'s platform MUST request [=user consent|user's explicit consent=] before deploying the fallback URL. Refer to [[#sctn-assertion-privacy]] for details on authentication ceremony privacy safeguards.
+            1. If the user [=user consent|consents=]:
+                1. Navigate the user to the fallback URL specified by {{AuthenticationExtensionsClientInputs/crossDeviceFallbackUrl}} using a user agent or equivalent system mechanism on the [=authenticator=] device.
+
+                    Note: Depending on the platform and the format of the fallback URL, this navigation may open a web browser page (e.g., to present a login or enrollment portal), deep-link into a native application, or trigger a platform-specific native flow.
+                1. The [=authenticator=] returns an error code indicating that a fallback URL is processed to the [=client platform=] and terminates the operation.
+            1. If the user does not [=user consent|consent=], return an error code equivalent to "{{NotAllowedError}}" and terminate the operation.
```
timcappalli:
```suggestion
:: 1. Send the [[#sctn-getAssertion|sending an assertion request over <code>AuthenticatorTransport/hybrid</code>]]
    1. The following steps are performed by the remote [=Credential Manager Hosting Device=]:
        1. Remove the {{AuthenticationExtensionsClientInputs/crossDeviceFallbackUrl}} extension input, and pass the request to enabled [=credential managers=].
        1. If a credential is found that can satisfy the request, ignore this extension and continue with normal processing.
        1. Request explicit [=user consent=] to navigate to the fallback URL.
          See [[#sctn-assertion-privacy]] for details on authentication ceremony privacy safeguards.
        1. If the user does not consent, return an error code equivalent to "{{NotAllowedError}}" and terminate the operation.
        1. Invoke a user agent or equivalent system mechanism with the fallback URL specified by {{AuthenticationExtensionsClientInputs/crossDeviceFallbackUrl}}.
        1. Return an error code indicating that a fallback URL is processed to the [=client platform=] and terminates the operation.
```

index.bs
```diff
@@ -7903,7 +7947,97 @@ This permission MUST be configured per-origin and MUST NOT offer the capability
...
+
+#### Security Properties #### {#sctn-cross-device-fallback-url-security-properties}
+
+The following are some of the security properties of the {{AuthenticationExtensionsClientInputs/crossDeviceFallbackUrl}}:
```
timcappalli:
```suggestion
The following are some security properties of the {{AuthenticationExtensionsClientInputs/crossDeviceFallbackUrl}}:
```

index.bs
```diff
@@ -7903,7 +7947,97 @@ This permission MUST be configured per-origin and MUST NOT offer the capability
...
+
+* The `crossDeviceFallbackUrl` does not provide the same level of security and phishing resistance as [=passkeys=], and is not a replacement for [=passkey=] authentication. Additional mitigation systems (e.g., a "risk engine") are required to resist phishing attacks made possible by this extension.
```
timcappalli:
```suggestion
* This extension does not provide the same level of security as a WebAuthn credential and provides no protections against phishing. Additional mitigation systems (e.g., a "risk engine") are required to resist phishing attacks which may be enabled by this extension.
```

index.bs
```diff
@@ -7903,7 +7947,97 @@ This permission MUST be configured per-origin and MUST NOT offer the capability
...
+
+* The `crossDeviceFallbackUrl` only comes into picture if [=passkey=]-based authentication is not possible.
```
timcappalli:
```suggestion
* The `crossDeviceFallbackUrl` only comes into picture if a WebAuthn credential is not available on the local or remote device.
```

index.bs
```diff
@@ -7903,7 +7947,97 @@ This permission MUST be configured per-origin and MUST NOT offer the capability
...
+
+* The `crossDeviceFallbackUrl` is an improvement over typical custom QR-code-based out-of-band cross-device authorization flows because of identified ways to resist phishing attacks while using it (as opposed to password-only flows that have fewer techniques to mitigate phishing.)
```
timcappalli: This should probably be removed unless you can actually explain it.

index.bs
```diff
@@ -7903,7 +7947,97 @@ This permission MUST be configured per-origin and MUST NOT offer the capability
...
+
+#### Risk Scenarios #### {#sctn-cross-device-fallback-url-risk-scenarios}
+
+Using `crossDeviceFallbackUrl` introduces security trade-offs compared to the primary WebAuthn assertion flow. This is because the `crossDeviceFallbackUrl` redirects the user to alternate authentication mechanisms (which typically rely on passwords, one-time codes, or out-of-band links) which lacks the strong phishing resistance of [=passkeys=]. Attackers may attempt to exfiltrate or intercept the `crossDeviceFallbackUrl` either from the client or the authenticator device. Although `crossDeviceFallbackUrl` does not contain any user specific information, it is linked to a session on [=[RP]=] backend. Hence can be used to trick users into authorizing authentication sessions on the attacker's behalf.
```
timcappalli: > it is linked to a session on [=[RP]=] backend.

That's an assumption

timcappalli:
```suggestion
Using this extension introduces security trade-offs compared to the primary WebAuthn assertion flow. This is because the `crossDeviceFallbackUrl` redirects the user to alternate authentication mechanisms (which typically rely on passwords, one-time codes, or out-of-band links) which lacks the strong phishing resistance of [=passkeys=]. Attackers may attempt to exfiltrate or intercept the `crossDeviceFallbackUrl` either from the client or the authenticator device. Although `crossDeviceFallbackUrl` does not contain any user specific information, it is linked to a session on [=[RP]=] backend. Hence can be used to trick users into authorizing authentication sessions on the attacker's behalf.
```

index.bs
timcappalli:
```suggestion
spec: FIDO-CTAP; urlPrefix: https://fidoalliance.org/specs/fido-v2.3-ps-20260226/fido-client-to-authenticator-protocol-v2.3-ps-20260226.html
type: dfn
text: §6. Authenticator API; url: authenticator-api
text: large, per-credential blobs; url: authenticatorLargeBlobs
text: credential manager hosting device; url: credential-manager-hosting-device
```

timcappalli (changes requested):
I have additional feedback on the sec/risk sections, but wanted to give you this review first.

index.bs
harshlal028: Done.

index.bs
harshlal028: I had renamed it based on another review comment [here](https://github.com/w3c/webauthn/pull/2380#pullrequestreview-4471139118). Renaming it back to Cross-Device Fallback Exception

Done.

index.bs
harshlal028: Removed line 8010.

Included additional explanation for line 8012

index.bs
harshlal028: Clarified this being an assumption in the text.

index.bs
harshlal028: I updated the explanation. PTAL.
thx

index.bs
```diff
@@ -7903,7 +7955,108 @@ This permission MUST be configured per-origin and MUST NOT offer the capability
...
+
+The following are some security properties of the {{AuthenticationExtensionsClientInputs/crossDeviceFallbackUrl}}:
+
+* This extension does not provide the same level of security as a WebAuthn credential and provides no protections against phishing. Additional mitigation systems (e.g., a "risk engine") are required to resist phishing attacks which may be enabled by this extension.
+
+* The `crossDeviceFallbackUrl` is an improvement over custom QR-code-based out-of-band cross-device authorization flows (where a [=[RP]=] displays a proprietary login QR code directly on a web page). In those custom flows, attackers can easily exfiltrate or proxy the QR code to a remote victim, and [=[RPS]=] have limited means to detect or resist phishing. In contrast, `crossDeviceFallbackUrl` provides stronger defense:
+    * The flow always prioritizes strong, phishing-resistant [=passkeys=], and only falls back if no eligible credential exists on the user's device.
+    * The interaction is mediated through the standard {{AuthenticatorTransport/hybrid}} transport, ensuring the user engaged in a genuine cross-device WebAuthn ceremony.
```
kreichgauer: This isn't really true: An attacker would extract the fallbackURL out of a WebAuthn request and relay it to their victim over some external channel. Once the victim opens the fallbackURL, the attacker would resolve the pending WebAuthn request with a CrossDeviceFallbackException. Hybrid wouldn't really be involved at all.

index.bs
```diff
@@ -7903,7 +7955,108 @@ This permission MUST be configured per-origin and MUST NOT offer the capability
...
+    * The fallback mechanism requires explicit [=user consent=] on the authenticator and returns a {{WebAuthnCrossDeviceFallbackException}} to the [=client=]. The [=[RP]=] only initiates and accepts the fallback authorization after receiving this exception, preventing blind polling.
```
kreichgauer: Similar point here: In the equivalent phishing scenario to the proprietary QR code mechanism described above, the user wouldn't really interact with an authenticator.

index.bs
```diff
@@ -7903,7 +7955,108 @@ This permission MUST be configured per-origin and MUST NOT offer the capability
...
+
+
+#### Risk Mitigation #### {#sctn-cross-device-fallback-url-risk-mitigation}
+
+[=Relying Parties=] are strongly recommended to implement the following mitigations to protect against these exfiltration risks:
+
+* **Primary Exception Requirement**: The fallback URL needs to be primed before usage by returning a {{WebAuthnCrossDeviceFallbackException}} to the [=[RP]=]. Only after receiving this exception should the [=[RP]=] start polling for the fallback URL's authentication results. It MUST be disregarded otherwise.
+
+* **Cross Device Fallback URL Expiration**: Invalidate the fallback URL's single-use authorization code (or session code) a short time after it is generated (e.g., a few seconds). Because it takes time for an attacker to forward the URL and trick a victim into visiting it, the [=[RP]=] would observe a significant time delay between generating the fallback URL and the victim visiting it.
```
kreichgauer: This needs to be weighed against the time it takes users to scan the hybrid QR codes with an authenticator.

kreichgauer (changes requested):

index.bs
harshlal028: Ack. Removed.

index.bs
harshlal028: Ack. Removed.

index.bs
harshlal028: Updated text to reflect this.

index.bs
```diff
@@ -7903,7 +7955,106 @@ This permission MUST be configured per-origin and MUST NOT offer the capability
 
 This section defines extensions that are both [=client extensions=] and [=authenticator extensions=].
 
-This section is currently empty.
+### Cross Device Fallback URL extension ### {#sctn-cross-device-fallback-url-extension}
```
timcappalli:
```suggestion
### Cross Device Fallback URL extension (<dfn>crossDeviceFallbackUrl</dfn>) ### {#sctn-cross-device-fallback-url-extension}
```

index.bs
```diff
@@ -7903,7 +7955,106 @@ This permission MUST be configured per-origin and MUST NOT offer the capability
...
+    1. If |rpId| [=is not a registrable domain suffix of and is not equal to=] |effectiveDomain|, and if the client:
+        <dl class="switch">
+            :   supports [[#sctn-related-origins|related origin requests]]
+            ::  1. Run the [$related origins validation procedure$] with arguments |fallbackUrlOrigin| and |rpId|.
+                    If the result is [FALSE], ignore this extension and continue with normal processing.
+            :   does not support [[#sctn-related-origins|related origin requests]]
+            ::  ignore this extension and continue with normal processing.
+        </dl>
+    1. Construct a [=request=] with its URL set to |fallbackUrl| and execute the [=url policy violation=] procedure.
+    1. If the response violates policy, ignore this extension and continue with normal processing.
```
iinuwa: For these steps, is there benefit to say that if the RP passes a bad value as fallbackUrl, the whole request should be rejected with a `SecurityError`? Letting a malformed or mismatched URL seems like a mistake on the RP's side and an opportunity for an attacker to slip through, e.g. if the parameters aren't fully removed from the request and the values pass through to the platform, who might not also validate these.

I suppose that if there were a mistake in Related Origins metadata document, that would cause all requests with this extension to fail, but I think that would probably also occur without this extension.

index.bs
```diff
@@ -7903,7 +7955,106 @@ This permission MUST be configured per-origin and MUST NOT offer the capability
...
+    1. Pass the {{AuthenticationExtensionsClientInputs/crossDeviceFallbackUrl}} extension input to the [=authenticator=] and continue processing.
+    1. If the [=authenticator=] returns an error code indicating that a fallback URL is processed, throw a new {{WebAuthnCrossDeviceFallbackException}}.
```
iinuwa: I wonder if there is opportunity here to bind the session continuation to the device that is going to do the fallback flow. Could we return a PKCE-like value generated by the peer client (smartphone) so that the RP can know when the request completes, the device that completes the request is the same one that the device that started the authentication session peered with over PXP?

iinuwa (commented):
Added a couple notes

index.bs
harshlal028: As per last discussion with stakeholders - we arrived at a conclusion that fallback URL is a mechanism other than webauthn and has different security properties - as such a FALLBACK. We would defer security consideration  to https://www.rfc-editor.org/rfc/rfc10027.html. The PR has been updated to reflect this.

Thanks

index.bs
harshlal028: My strawman would be to not fail the entire request and just ignore the fallbackUrl extension - in case it is malformed - which will cause the validation to fail, coz a passkey can totally be available and used - in which case fallback won't be needed at all.

index.bs
```diff
@@ -7903,7 +7975,82 @@ This permission MUST be configured per-origin and MUST NOT offer the capability
...
+
+General security considerations, threat models, attack patterns (such as URL exfiltration and session forwarding), and practical mitigations for cross-device flows are described in [[RFC10027]]. [=Relying Parties=] implementing this extension SHOULD consult [[RFC10027]] for guidance on securing cross-device flows.
+
+In the context of WebAuthn:
+
+* The fallback URL is only invoked if no eligible [=passkey=] exists on the [=roaming authenticator=] and the user gives explicit consent on the remote device.
+* [=Relying Parties=] MUST NOT accept authentication assertions from a fallback URL unless the primary WebAuthn ceremony resulted in a {{WebAuthnOperationException}} with {{WebAuthnOperationErrorCode/"cross-device-fallback"}}, confirming that the fallback flow was triggered on the remote device.
```
kreichgauer: > confirming that the fallback flow was triggered on the remote device

This still reads like some kind of assurance that the URL was opened in a hybrid client.

The WebAuthnOperationErrorCode proves that _a_ client interpreted the extension and returned the extension. If the URL was opened, it may have been on a hybrid authenticator  proximal to that client; or on any other device on the planet whose user received the URL from the owner of the client via some other channel.

Since this is in the security section, I don't think this claim should really be made.

index.bs
kreichgauer: I'd lean towards rejecting the request as well. When we ignore something in a WebAuthn request, it's typically because it's some kind of unknown value and we want to fail open in case there are newer implementations where that value is meaningful. But that isn't the case here: It's a parsing failure and/or negative security decision. Hard failure seems more appropriate.

index.bs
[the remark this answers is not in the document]
kreichgauer: Reopening because this is still defined as a client extension. It's processed by the device on the other end of a hybrid tunnel, and that's an authenticator, so I think authenticator extension is the way to go here?

The output value is a bit awkward actually: There is no authenticator response, so we can't actually return an output value even if we define one. Perhaps this should mention that an output is omitted on purpose?

kreichgauer (changes requested):
Looks good overall, couple of minor comments

index.bs
harshlal028: Removed the phrasing "confirming that the fallback flow was triggered on the remote device"

index.bs
harshlal028: Made the change. Thanks all.

iinuwa (commented):
I still wonder if there is value in providing more structure on the fallback URL semantics to allow the CMHD to bind the fallback URL to itself to make the fallback URL less useful to share for phishing purposes. But with the intended goals, I think this is a good start.

kreichgauer:
Approving for now, but

> I still wonder if there is value in providing more structure on the fallback URL semantics to allow the CMHD to bind the fallback URL to itself to make the fallback URL less useful to share for phishing purposes. But with the intended goals, I think this is a good start.

I agree and think this is worth exploring. If the fallbackURL page was able to relay data back to the page that made the WebAuthn request over the hybrid tunnel, most of the phishability concerns with this proposal would disappear. You could basically bind the phone's fallbackURL session to the hybrid tunnel with the client, which would be an actual security improvement over traditional cross-device sign-in flows. That's pretty neat.

kreichgauer (approved):

harshlal028:
> Approving for now, but
> 
> > I still wonder if there is value in providing more structure on the fallback URL semantics to allow the CMHD to bind the fallback URL to itself to make the fallback URL less useful to share for phishing purposes. But with the intended goals, I think this is a good start.
> 
> I agree and think this is worth exploring. If the fallbackURL page was able to relay data back to the page that made the WebAuthn request over the hybrid tunnel, most of the phishability concerns with this proposal would disappear. You could basically bind the phone's fallbackURL session to the hybrid tunnel with the client, which would be an actual security improvement over traditional cross-device sign-in flows. That's pretty neat.

Agreed. We can do a followup on that. Here is a proposal for the same in Hybrid subgroup https://github.com/fido-alliance/hybrid-next/issues/38. Once that comes to fruition we can add that improving flow for RPs.

iinuwa:
> Agreed. We can do a followup on that. Here is a proposal for the same in Hybrid subgroup https://github.com/fido-alliance/hybrid-next/issues/38. Once that comes to fruition we can add that improving flow for RPs.

FWIW, I was thinking this is more in line with https://github.com/fido-alliance/hybrid-next/issues/4#issuecomment-5333224563 session-binding for fallback URL specifically rather than the general PXP channel binding.

index.bs
```diff
@@ -4479,6 +4500,57 @@ Note: The {{PublicKeyCredentialHint}} enumeration is deliberately not referenced
...
+
+### Web Authentication Operation Exception (interface <dfn interface>WebAuthnOperationException</dfn>) ### {#sctn-webauthn-operation-exception}
+
+The following custom {{DOMException}} can be raised:
+
+<xmp class="idl">
+    [Exposed=Window, Serializable]
+    interface WebAuthnOperationException : DOMException {
+        constructor(WebAuthnOperationErrorCode code, optional DOMString message = "");
+        readonly attribute WebAuthnOperationErrorCode code;
```
pascoej, line 4526: can we avoid redefining code from a short to a string here?

Maybe we can do something similar to WebTransportError and call it something else like operationError.

MasterKale: It's "[`WebTransportError.streamErrorCode`](https://developer.mozilla.org/en-US/docs/Web/API/WebTransportError/WebTransportError#streamerrorcode)", so maybe we call this `WebAuthnOperationException.operationErrorCode`? 🤔

MasterKale (approved):
I think you should consider @pascoej's suggestion to rename `WebAuthnOperationException.code`, but other than that this PR seems ready to go.

iinuwa:
I added a separate issue to discuss binding the fallback URL to a device: https://github.com/w3c/webauthn/issues/2478

index.bs
harshlal028: Done. Used WebAuthnOperationException.operationErrorCode

PR #2381, opened 2026-01-26
Add JSON object properties table to Set User Verified virtual authenticator endpoint

MasterKale:
This editorial PR adds a table describing the JSON object that must be `POST`'d to the **Set User Verified** virtual authenticator endpoint:

Screenshot 2026-01-26 at 3 43 16 PM

No normative behavior changes are included in this PR. This simply aims to update this endpoint's description to be in line with other endpoints like **Add Virtual Authenticator**, **Add Credential**, and **Set Credential Properties** that describe the shape of their JSON object parameters.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -8344,7 +8344,11 @@ It contains the following |key| and |value| pairs:
         <tbody>
             <tr>
                 <td>|isUserVerified|</td>
-                <td>Determines the [=authData/flags/UV|UV flag=] in [=authenticator data=] in the [=Virtual Authenticator|Virtual Authenticator's=] response.</td>
+                <td>
+                    Determines the result of [=User Verification=] performed on the
+                    [=Virtual Authenticator=]. If set to [TRUE], [=User Verification=] will always
+                    succeed. If set to [FALSE], it will fail.
+                </td>
                 <td>boolean</td>
             </tr>
         </tbody>
--- a/index.bs
+++ b/index.bs
@@ -8329,6 +8329,28 @@ is defined as follows:
     </table>
 </figure>
 
+The <dfn>Set User Verified Parameters</dfn> is a JSON [=Object=] passed to the [=remote end steps=] as |parameters|.
+It contains the following |key| and |value| pairs:
+
+<figure id="table-setCredentialPropertiesParameters" class="table">
+    <table class="data">
+        <thead>
+            <tr>
+                <th>Key</th>
+                <th>Description</th>
+                <th>Value Type</th>
+            </tr>
+        </thead>
+        <tbody>
+            <tr>
+                <td>|isUserVerified|</td>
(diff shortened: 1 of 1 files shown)


emlun (approved):
So this just adds an definition of the _parameters_ object that's already used in the remote end steps, right? Looks good to me, thanks!

index.bs
```diff
@@ -8329,6 +8329,28 @@ is defined as follows:
     </table>
 </figure>
 
+The <dfn>Set User Verified Parameters</dfn> is a JSON [=Object=] passed to the [=remote end steps=] as |parameters|.
+It contains the following |key| and |value| pairs:
+
+<figure id="table-setCredentialPropertiesParameters" class="table">
```
sbweeden: This figure id looks like a cut/paste from https://github.com/w3c/webauthn/blob/main/index.bs#L8369

Perhaps it should have its own unique id?

sbweeden (changes requested):

index.bs
emlun: Agreed, good catch!

MasterKale: Oops, busted. Thanks for catching this, I've fixed it.

sbweeden (approved):
LGTM

index.bs
```diff
@@ -8329,6 +8329,28 @@ is defined as follows:
...
+            <tr>
+                <th>Key</th>
+                <th>Description</th>
+                <th>Value Type</th>
+            </tr>
+        </thead>
+        <tbody>
+            <tr>
+                <td>|isUserVerified|</td>
+                <td>Determines the [=authData/flags/UV|UV flag=] in [=authenticator data=] in the [=Virtual Authenticator|Virtual Authenticator's=] response.</td>
```
nsatragno: This is not an accurate description for this flag.

"Determines the result of [User Verification](https://w3c.github.io/webauthn/#user-verification) performed on the [Virtual Authenticator](https://w3c.github.io/webauthn/#virtual-authenticators). If set to true, [User Verification](https://w3c.github.io/webauthn/#user-verification) will always succeed. If set to false, it will fail."

See `isUserVerified` under https://w3c.github.io/webauthn/#sctn-automation-virtual-authenticators

In practice, developers can use this to simulate the user not passing the UV check (e.g. because they entered the wrong PIN). If you set it to false and UV was required, the request should fail, not simply return UV=0.

nsatragno (changes requested):

index.bs
MasterKale: Oops, you're right, I missed the nuance completely. I've updated this description with the one for `isUserVerified` that you highlighted, from the list of Virtual Authenticator internal state variables.

MasterKale:
@nsatragno Can you please re-review this at your earliest convenience?

nsatragno (approved):
Approved, thank you!

Issue #2363, opened 2025-11-19
Virtual authenticator should allow `counter` to always be `0`


MasterKale:
## Proposed Change

It doesn't seem possible today, with the currently defined WebAuthn virtual authenticator API, to emulate use of synced passkey credential managers. The authenticator data `signCount` always increments in authentication responses. This makes it impossible to use the virtual authenticators to test scenarios in which `signCount` is always `0` in auth responses.

Poking around the spec, it's possible to set an initial sign count when you add a credential to a virtual authenticator:

https://w3c.github.io/webauthn/#sctn-automation-add-credential

However this signCount always increments in subsequent authentications. If an RP backend keeps track of the counter, a test script that wants to automate the instantiation of a virtual authenticator and its credential would need to query the backend for its current `signCount` for that credential, or subsequent auth responses from the authenticator would have a signCount lower than what's in the DB, and the response would get rejected.

Maybe we can expand the [Set Credential Properties](https://w3c.github.io/webauthn/#sctn-automation-set-credential-properties) endpoint to enable a credential response to always return with a `signCount` of 0 🤔 

I'd like to talk about how we might enhance the virtual authenticator API to allow for better emulation of synced passkey providers.

nsatragno:
Agreed, this sounds like a good idea.

nicksteele:
That seems reasonable, only addition would be that we might want to support other values, but 0 would be a good place to start.

MasterKale:
> That seems reasonable, only addition would be that we might want to support other values, but 0 would be a good place to start.

This is a good idea, let's make it possible to set any positive numeric sign count to support more intentional testing of counter scenarios.

MasterKale:
From WG meeting today: there's two possible behavioral changes we can make to virtual authenticators to modernize them for the current synced passkeys age we're in:

- Add something like "`defaultSignCount`" to [Add Virtual Authenticator](https://w3c.github.io/webauthn/#sctn-automation-add-virtual-authenticator) that defaults to `0`. When the authenticator is initialized with a default signature counter of `0`, normatively define the authenticator's behavior to **not** increment `authData.signCount` in registration and authentication ceremony responses
- Continue with https://github.com/w3c/webauthn/pull/2382 to allow for setting a sign count on an existing credential. The value of `authData.signCount` in subsequent authentication responses for the specified credential should be in line with behavior defined above for the authenticator in determining when to increment the signature counter.

MasterKale:
[...]

PR #2382, opened 2026-01-27
Update Virtual Authenticator API to better emulate synced passkey behavior

MasterKale:
This PR proposes the following changes to the Virtual Authenticator API to make it possible for RPs to actually set up automated testing of synced passkey, and refine their testing of cloned authenticator scenarios:

- Update the **Add Credential** endpoint to normatively require the new credential's signature counter to remain `0` in subsequent authentications
- Update the **Set Credential Properties** endpoint with a new `signCount` parameter to determine the value of a credential's signature counter prior to a `authenticatorGetAssertion` call using that credential, and normatively require the existing credential's signature counter to remain `0` in subsequent authentications

Addresses #2363.

Implementation commitment:

- [x] WebKit (https://bugs.webkit.org/show_bug.cgi?id=319110)
- [x] Chromium (https://issues.chromium.org/issues/533056772)

Documentation and checks

- [ ] Affects privacy
- [ ] Affects security

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/pull/2382.html" title="Last updated on Jul 30, 2026, 5:30 PM UTC (c0f89e6)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/2382/e6f24bd...c0f89e6.html" title="Last updated on Jul 30, 2026, 5:30 PM UTC (c0f89e6)">Diff</a>


selfissued (approved):

index.bs
```diff
@@ -8386,6 +8386,11 @@ It contains the following |key| and |value| pairs:
                 <td>The [=backup state=] [=credential property=].</td>
                 <td>boolean</td>
             </tr>
+            <tr>
+                <td>|signCount|</td>
+                <td>The [=signature counter=] [=credential property=]. This property should be zero or a positive value.</td>
```
emlun: This is JSON, which allows arbitrary-precision floating-point numbers, right? We should probably specify this more precisely as an integer value.
```suggestion
                <td>The [=signature counter=] [=credential property=].
                This property should be an integer value in the range 0 to 4294967295, inclusive.</td>
```

index.bs
```diff
@@ -8408,6 +8413,8 @@ The [=remote end steps=] are:
 
  1. Let |backupState| be the |parameters|' |backupState| property.
  1. If |backupState| is defined, set the [=backup state=] [=credential property=] of |credential| to the value of |backupState|.
+ 1. Let |signCount| be the |parameters|' |signCount| property.
+ 1. If |signCount| is defined and equal to zero or to a positive value, set the [=signature counter=] [=credential property=] of |credential| to the value of |signCount|.
```
emlun: Simile:
```suggestion
 1. If |signCount| is defined and set to an integer value in the range 0 to 4294967295, inclusive, set the [=signature counter=] [=credential property=] of |credential| to the value of |signCount|.
```

emlun: Otherwise ignore it, I conclude? Should we state that explicitly?

index.bs
```diff
@@ -8386,6 +8406,14 @@ It contains the following |key| and |value| pairs:
                 <td>The [=backup state=] [=credential property=].</td>
                 <td>boolean</td>
             </tr>
+            <tr>
+                <td>|signCount|</td>
+                <td>
+                The [=signature counter=] [=credential property=].
+                This property should be an integer value in the range `0` to `4294967295`, inclusive.
```
nsatragno: nit: indent these lines?

index.bs
```diff
@@ -8408,6 +8436,14 @@ The [=remote end steps=] are:
 
  1. Let |backupState| be the |parameters|' |backupState| property.
  1. If |backupState| is defined, set the [=backup state=] [=credential property=] of |credential| to the value of |backupState|.
+ 1. Let |signCount| be the |parameters|' |signCount| property.
+ 1. If the [=backup eligibility=] [=credential property=] of |credential| is [TRUE]:
+    1. Set the [=signature counter=] [=credential property=] of |credential| to `0`.
```
nsatragno: Doesn't this potentially hide a bug? If the developer is setting the signature counter to a non-zero value and we silently ignore that, that seems bad.

Instead, we could do one of two things (whichever you prefer):
* Throw an error if the developer sets a non-zero signCount with BE=true.
* Ignore the fact that BE=true. The developer _probably_ knows what they're doing. Do not increment the counter on the next assertion.

index.bs
```diff
@@ -8408,6 +8436,14 @@ The [=remote end steps=] are:
...
+ 1. Otherwise:
+    1. If |signCount| is defined and set to an integer value in the range `0` to `4294967295`, inclusive:
```
nsatragno: Here as well I think it's better if we throw an error if the developer tries to set a value outside the accepted range.

index.bs
```diff
@@ -7881,6 +7881,22 @@ Each stored [=virtual authenticator=] has the following properties:
     This value MUST be reflected by the [=BS=] [=authenticator data=] [=flag=] when performing an [=authenticatorMakeCredential=]
     operation with this [=virtual authenticator=].
 
+[=Virtual Authenticator=] behavior MUST reflect the following to emulate the most common [=[WAA]=] behaviors:
```
nsatragno: General comment: please file a bug against chromium if you want to see this implemented! :)

nsatragno (commented):
This looks like a good change! Some comments below:

index.bs
MasterKale: Would such a bug need this PR to be merged before changes could be made?

nsatragno: Not _necessarily_, but this being merged gives us some level of confidence that the API won't change and prevents duplicate work. You can file the bug before the PR gets merged.

index.bs
MasterKale: > Throw an error if the developer sets a non-zero signCount with BE=true

I was intentionally, silently ignoring `signCount` when `be=true` but I agree it's better to be explicit about the bad call pattern. I'll add this in.

index.bs
MasterKale: Thanks for catching this, updated.

index.bs
MasterKale: I've refactored this to explicitly return WebDriver errors when `signCount` and `be` are not valid combinations.

index.bs
MasterKale: Same thing here, I've defined an explicit return of WebDriver error on a bad `signCount` + `be` parameter combo.

index.bs
```diff
@@ -8386,6 +8406,14 @@ It contains the following |key| and |value| pairs:
                 <td>The [=backup state=] [=credential property=].</td>
                 <td>boolean</td>
             </tr>
+            <tr>
+                <td>|signCount|</td>
+                <td>
+                    The [=signature counter=] [=credential property=].
+                    This property should be a 32-bit unsigned integer value in the range `0` to `4294967295`, inclusive.
```
MasterKale: I know it's redundant that a "32-bit unsigned integer" is "in the range of 0 to 4,294,967,295" but why not make the connection for those who aren't as familiar with "unsigned long" values?

MasterKale:
I'm trying to write this PR to also achieve Martin's desired end state of the Virtual Authenticator API defaulting into a "synced passkey" mode instead of the API's current "security key" mode.

@kreichgauer and @nsatragno, can you take a look at this PR with an eye on if these proposed changes would move us to that end state **if** a subsequent PR were to update **Add Virtual Authenticator** endpoint's `defaultBackupEligibility` argument to instead default to `true`? Right now it defaults to `false`:

Screenshot 2026-02-19 at 10 13 44 AM

I'm thinking of a migration path for RPs with existing tests built around current Virtual Authenticator API behavior. I _think_ we can make it as simple as updating their calls to the **Add Virtual Authenticator** endpoint to add `defaultBackupEligibility: false` to get their tests passing again 🤔

nsatragno:
> can you take a look at this PR with an eye on if these proposed changes would move us to that end state if a subsequent PR were to update Add Virtual Authenticator endpoint's defaultBackupEligibility argument to instead default to true? Right now it defaults to false:

This change sounds good to me.

index.bs
```diff
@@ -8408,6 +8446,14 @@ The [=remote end steps=] are:
...
  1. Let |backupState| be the |parameters|' |backupState| property.
  1. If |backupState| is defined, set the [=backup state=] [=credential property=] of |credential| to the value of |backupState|.
+ 1. Let |signCount| be the |parameters|' |signCount| property.
+ 1. If the [=backup eligibility=] [=credential property=] of |credential| is [TRUE]:
+     1. If |signCount| is defined and equal to `0`, set the [=signature counter=] [=credential property=] of |credential| to `0`.
+     1. Otherwise, return a [=WebDriver error=] with [=WebDriver error code=] [=invalid argument=].
+ 1. Otherwise:
+     1. If |signCount| is defined and set to a 32-bit unsigned integer value in the range `0` to `4294967295`, inclusive,
+        set the [=signature counter=] [=credential property=] of |credential| to the value of |signCount|.
+     1. Otherwise, return a [=WebDriver error=] with [=WebDriver error code=] [=invalid argument=].
```
nsatragno, line 8628: This makes |signCount| mandatory, which seems undesired.

Why not,


```
1. If |signCount| is defined,
  1. If |signCount| is not a 32-bit unsigned integer value in the range `0` to `4294967295`, inclusive,
     return a [=WebDriver error=] with [=WebDriver error code=] [=invalid argument=].
  1. If the [=backup eligibility=] [=credential property=] of |credential| is [TRUE] and
     |signCount| is not `0`, return a [=WebDriver error=] with [=WebDriver error code=] [=invalid argument=].
  1. Set the [=signature counter=] [=credential property=] of |credential| to the value of |signCount|.
```

MasterKale: Oh, you're right, I made it mandatory with the second WebDriver error...what if I simply dropped this line? That should make `signCount` optional when `be=False`

MasterKale: Oh but the error would be if `be=False` and `signCount` isn't `unsigned long`...let me look at your suggestion and incorporate that somehow.

MasterKale: That wording looked good to me so I've gone ahead with that instead.

MasterKale:
@nsatragno @emlun @ve7jtb @sbweeden can you please review this at your earliest convenience?

MasterKale:
@kreichgauer I'd like your take on this PR too, [especially with the context in this comment](https://github.com/w3c/webauthn/pull/2382#issuecomment-3929020923), as I'm trying to pave a path to getting the Virtual Authenticator API to default to the synced passkey authenticator behavior we discussed over in #2363

nsatragno (approved):
Approved, thank you!

kreichgauer:
I understand that defaultBackupEligibility can simulate synced passkeys and synced passkeys don't have signature counters. But why not just make the [initial/default sign count explicit](https://github.com/w3c/webauthn/issues/2363#issuecomment-3813576898)?

For RPs that do have expectations about the sign counter value in their virtual authenticator tests, setting an explicit default seems more easily discoverable than the implicit relationship to the BE bit. (Also, authenticators with fixed-zero counters and non-syncing credentials do exist! :-))

pascoej (approved):

MasterKale:
After a good conversation with @kreichgauer today, I've simplified this PR somewhat by decoupling the signature counter from a virtual authenticator's default backup eligibility. Instead, both the **Add Credential** and **Set Credential Properties** endpoints have had `signCount` properties updated/added to normatively require that a credential's signature counter "MUST NOT increment" when `signCount` is set to `0`.

This slightly different direction should allow for RPs to write tests for authenticators that always return `0` for sign count, whether the authenticator can backup credentials or not. This should still solve my original problem over in #2363 that the signature counter always increments in getAssertion responses, which is not true for most modern platform authenticators and the macOS Chrome profile authenticator.

@emlun @pascoej @nsatragno @sbweeden @ve7jtb I think this is significant enough a change that I have to ask you all to re-review for approval 🙇‍♂️

index.bs
```diff
@@ -8208,8 +8216,12 @@ The [=remote end steps=] are:
     :: Construct from |userName| and |userDisplayName|.
  1. Set the |credential|'s [=backup eligibility=] [=credential property=] to |backupEligibility|.
  1. Set the |credential|'s [=backup state=] [=credential property=] to |backupState|.
- 1. Associate a [=signature counter=] |counter| to the |credential| with a starting value equal to the |parameters|'
-     |signCount| or `0` if |signCount| is `null`.
+ 1. Let |signCount| be the |parameters|' |signCount| property.
+ 1. If |signCount| is `null`, set |signCount| to the value of `0`.
+ 1. If |signCount| is defined and set to a 32-bit unsigned integer value
+     in the range `0` to `4294967295`, inclusive, associate a [=signature counter=] |counter|
```
kreichgauer, line 8379: Should we say that for 0, **no** signature counter will be associated with the credential, so that the authenticatorData signCount for this credential will be fixed zero? (Maybe that also obviates the comment above in the parameter description; or we could say it twice.)

index.bs
```diff
@@ -8408,6 +8432,12 @@ The [=remote end steps=] are:
 
  1. Let |backupState| be the |parameters|' |backupState| property.
  1. If |backupState| is defined, set the [=backup state=] [=credential property=] of |credential| to the value of |backupState|.
+ 1. Let |signCount| be the |parameters|' |signCount| property.
+ 1. If |signCount| is defined:
+     1. If |signCount| is not set to a 32-bit unsigned integer value in the range `0` to
+        `4294967295`, inclusive, return a [=WebDriver error=] with [=WebDriver error code=]
+        [=invalid argument=].
+     1. Set the [=signature counter=] [=credential property=] of |credential| to the value of |signCount|.
```
kreichgauer: Similarly here, we could say that if signCount is 0, disassociate any signature counter from the credential, ensuring the authenticatorData for assertions will use a zero signCount.

Also, signature counter I think is not currently defined as a "credential property" in the normal spec. Not sure if that matters here. We could just say "signature counter" without the "credential property".

kreichgauer (approved):

index.bs
MasterKale: If a signature counter was **not** associated to a new credential in **Add Credential** then that'd potentially complicate **Set Credential Properties** with extra steps to add a signature counter if the new `signCount` in this PR specified a non-zero number. On the surface it seems cleaner to always have a signature counter associated with a credential if the credential exists.

Do you think it would unduly burden implementers to require the initialization of a signature counter in Add Credential even if it's always going to be `0`?

index.bs
MasterKale: > Also, signature counter I think is not currently defined as a "credential property" in the normal spec. Not sure if that matters here. We could just say "signature counter" without the "credential property".

Okay, this is good feedback, I copy-pasted from some of the other existing logic but it might not even make sense for those (`backupEligibility` and `backupState`) to mention "credential property" either. I'll at least remove it from this PR, we can come back later and remove it from the others.

index.bs
MasterKale: Ah, but under [Signature Counter Considerations](https://w3c.github.io/webauthn/#signature-counter) it says...

> Authenticators that do not implement a [signature counter](https://w3c.github.io/webauthn/#signature-counter) leave the [signCount](https://w3c.github.io/webauthn/#authdata-signcount) in the [authenticator data](https://w3c.github.io/webauthn/#authenticator-data) constant at zero.

So maybe `signCount` in **Add Counter** optionally associates a signature counter iff it's non-zero, while **Set Credential Properties** only allows a new value to be set if a signature counter is associated with the credential.

But then how could an RP test e.g. a security key that creates a new credential with signature counter of `0` which then increments to `1` on the subsequent auth? There'd be no signature counter associated with the credential after **Add Credential** to increment 🤔

MasterKale:
I think I should have defined earlier the virtual authenticator test cases I'm trying to support with these changes:

1. **Add Cred.** with `0`, returns `1` in next auth (i.e. security key that implements a signature counter)
1. **Add Cred.** with `0`, returns `0` in next auth (i.e. platform authenticator that does not implement a signature counter)
1. **Add Cred.** with `10`, returns `11` in next auth, **Set Cred. Props.** to `10`, returns `11` again in second auth (i.e. a cloned authenticator)

Of these three, only **No. 1** is currently supported by the current Virtual Authenticator API.

**No. 2** would be addressed by the normative authenticator behavior change in this PR when `signCount` in Add Credential is `0`.

...Ah, but now I'm seeing that it would break **No. 1** because `signCount` initializes to `0` when unset, which leads to a counter of `1` in the first auth with the credential. We'd have to suggest RPs update their current tests to set `signCount` to `1` when calling **Add Credential** to ensure the signature counter increments in subsequent auth responses, and then update signature counter incrementing test assertions accordingly. I guess this isn't that bad/onerous for RPs, as much as I'd like to prevent them from having to make any changes if possible...

**No. 3** would be addressed by the addition of `signCount` to **Set Credential Properties**. Someone updating their tests to use this new argument would just update downstream test logic accordingly, so I think this one's fine to introduce as-is.

harshlal028 (approved):

sbweeden (approved):
LGTM

MasterKale:
From WG meeting today:

- @kreichgauer to take one final look to make sure I addressed his concerns
- @MasterKale to submit the various bug tickets to implement these changes

harshlal028 (approved):

index.bs
kreichgauer: > But then how could an RP test e.g. a security key that creates a new credential with signature counter of 0 which then increments to 1 on the subsequent auth?

Yes, good point. Perhaps the solution is to say that if signCount is 'null', the authenticator should behave as if it didn't support signature counters when generating assertions for the credential. And if signCount is a number >= 0, it should use that as the current state of the signature counter (and increment it accordingly). Would that work?

(Set Credential Properties would behave equivalently: Pass null to disable the counter, otherwise initialize it with the given value.)

MasterKale: I like that idea. I had a thought yesterday to allow `null` on Add Credential, but not on Set Credential (i.e. you could create a credential that never has a counter associated with it, or one that does and be able to update it.) But that might be a bit too inflexible for a testing API so I think it makes sense to let both Add Cred and Set Cred Props support `null | uint` for values for counter.

MasterKale: I've made the changes along the lines of what you suggested:

- When `signCount` is null in either Add or Set, do not associate a signature counter and always return `authData.signCount: 0`
- When `signCount` is a `uint` in either Add or Set, associate a signature counter, increment as usual, and use it to determine the value of `authData.signCount`

kreichgauer:
LGTM

MasterKale:
@nsatragno @pascoej I've submitted bug reports to both Chromium and WebKit about these changes and updated the PR description with the links. I don't know if that's enough to check off those two under "implementation commitment" so I left them unchecked for now. When you have a chance can you please take a look at the following bug reports, respectively, and let me know if I've submitted these properly? Feels weird to report these as bugs but 🤷‍♂️ 

- Chromium: https://issues.chromium.org/issues/533056772
- WebKit: https://bugs.webkit.org/show_bug.cgi?id=319110

nsatragno:
We (Chrome) definitely want this so feel free to check it for us.

timcappalli (approved):

MasterKale:
As per discussion in the WG today: `signCount` for Add Credential and Set Credential Properties should have a minimum of `1` because these skip the use of `.create()`, which would always return `authData.signCount: 1` in the registration ceremony response. Thus `0` should be unsupported, and in fact RP processing rules for **assertion** verification say that ["once 0 always 0"](https://www.w3.org/TR/webauthn-3/#authenticatorgetassertion:~:text=If%20authData.signCount%20is%20nonzero%20or%20credentialRecord.signCount%20is%20nonzero). Thus if you're not setting `null` in either of these APIs to achieve synced-passkey-always-0 signCounts in subsequent responses for a credential, then you need to set a valid post-registration `signCount` value of at least `1` as to not break RP signature count validation as per the spec.

Since I'm trying to continue supporting incrementing signature counter behavior, I'm going to change this behavior and handle `0` differently, either replacing `null` for it instead, or as an explicit "invalid option" error if I keep with `null`.

emlun (approved):
Sorry for the delay!

MasterKale:
@kreichgauer raised a good point to me after the WG meeting that it's entirely allowed by the spec that `signCount: 0` come back from `.create()`, especially if no signature is present in the registration response because of `"none"` attestation taking place:

https://w3c.github.io/webauthn/#sctn-op-make-cred:~:text=If%20the%20authenticator%3A,be%20constant%20at%20zero.

Then subsequent authentication responses would contain an incremented signature counter because actual signatures are being generated, meaning the first response from `.get()` after using the **Add Credential** endpoint to add a credential is okay to increment from `0` and populate `authData.signCount: 1`:

https://w3c.github.io/webauthn/#sctn-sign-counter:~:text=The%20signature%20counter%20is%20incremented%20for%20each%20successful%20authenticatorGetAssertion%20operation%20by%20some%20positive%20value

It now seems to be perfectly fine, then, to allow the **Add Credential** and **Set Credential Properties** endpoints to set a `0` signature counter that increments, which means the behavior as currently defined in this PR is allowed by the rest of the spec. I believe the PR is fine to merge as-is, then, unless someone corrects my understanding here.

The one change I _did_ make after this realization is to tweak a bit of verbiage to prescribe Virtual Authenticators behavior when a `null` value for the endpoints' `signCount` argument is set, instead of prescribing how `authData.signCount` gets populated (the latter is implied within the new wording that achieves the former, and also avoids potentially under-defining other virtual authenticator behavior when a signature counter is not associated to a credential.)

cc @emlun @nsatragno

MasterKale:
There's some ambiguity in `signCount`, so I'm going to make the following changes:

- Update the property description to explicitly state that this argument is required to be `null | number`
- Add a processing step to raise if `signCount` is "not defined"

MasterKale:
@nsatragno @pascoej I've made those slight tweaks to A) the descriptions of `signCount` to denote that they are required values, and B) updated processing steps to handle cases of `signCount` not being defined. Can you please take a look at your earliest convenience and confirm this makes implementation more straightforward?

- **Add Credential:** https://pr-preview.s3.amazonaws.com/w3c/webauthn/pull/2382.html#sctn-automation-add-credential
- **Set Credential Properties:** https://pr-preview.s3.amazonaws.com/w3c/webauthn/pull/2382.html#sctn-automation-set-credential-properties

pascoej (approved):

index.bs
```diff
@@ -8594,6 +8626,14 @@ The [=remote end steps=] are:
 
  1. Let |backupState| be the |parameters|' |backupState| property.
  1. If |backupState| is defined, set the [=backup state=] [=credential property=] of |credential| to the value of |backupState|.
+ 1. Let |signCount| be the |parameters|' |signCount| property.
+ 1. If |signCount| is not defined, return a [=WebDriver error=] with [=WebDriver error code=] [=invalid argument=].
```
nsatragno: This seems wrong. If `signCount` is not defined, it should not be changed on this command. Otherwise you can't e.g. change the backup state without setting a new signature counter (and it would not be backwards compatible).

nsatragno (changes requested):

index.bs
MasterKale: Oh, I went too far with the change. I actually had it right with the original text for Set Credential Properties. I agree `signCount` should be required for Add Credential, but it should be possible for an RP to set other credential properties without having to remember what `signCount` should be.

I'll revert this Set Credential Properties change.

index.bs
```diff
@@ -8594,6 +8618,14 @@ The [=remote end steps=] are:
 
  1. Let |backupState| be the |parameters|' |backupState| property.
  1. If |backupState| is defined, set the [=backup state=] [=credential property=] of |credential| to the value of |backupState|.
+ 1. Let |signCount| be the |parameters|' |signCount| property.
+ 1. If |signCount| is defined:
+     1. If |signCount| is set to a 32-bit unsigned integer value in the range `0` to
+         `4294967295`, inclusive, set the existing [=signature counter=] of |credential|
+         to the value of |signCount|. If the |credential| has no [=signature counter=],
+         associate one to |credential| and set it to the value of |signCount|.
+     1. If |signCount| is `null`, remove any [=signature counter=] set to the |credential|.
```
nsatragno: This needs to start with an `Otherwise, if...`. Without it, the next line would run whenever `signCount` is not null, which is not what we want.

nsatragno (changes requested):

index.bs
MasterKale: So "Otherwise, if" is `else` for the purposes of processing steps? No problem, I'll update this.

nsatragno: I mean you could have written `else` too, but below this you wrote `otherwise` so... (:

nsatragno (approved):
Approved. Let's get this merged.

PR #2383, opened 2026-01-28
Deprecate Authenticator Attachment in favor of Hints

### LINKED ISSUE #2053: Deprecate AuthenticatorAttachment in favor of PublicKeyCredentialHints.
## Background

We have introduced PublicKeyCredentialHints for RP to better convey intention of which transport is preferred for both credential creation and authentication. Previously, AuthenticatorAttachment was used to convey similar intention. 

However, AuthenticatorAttachment has a side affect during credential creation. It excludes certain authenticators and the definition of platform vs cross-platform has been murky for some time since the introduction of hybrid transport. It leads to market fragmentation and UI differences between platforms. 

## Proposed Change
Deprecate AuthenticatorAttachment in favor of PublicKeyCredentialHints in the spec. For backwards compatibilty when only attachment is provided, we can map those values to corresponding PublicKeyCredentialHints.

akshayku:
Closes #2053 

Implementation commitment:

Documentation and checks

- [ ] Affects privacy
- [ ] Affects security

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 /bikeshed/
 .vagrant/
+.vs/
 index.html
--- a/index.bs
+++ b/index.bs
@@ -3754,7 +3754,6 @@ Note: The {{AuthenticatorAttachment}} enumeration is deliberately not referenced
 
     :   <dfn>cross-platform</dfn>
     ::  This value indicates [=cross-platform attachment=].
-
 </div>
 
 Note: An [=authenticator attachment modality=] selection option is available only in the {{PublicKeyCredential/[CREATE-METHOD]}} operation. The [=[RP]=] may use it to, for example, ensure the user has a [=roaming credential=] for
--- a/index.bs
+++ b/index.bs
@@ -2055,7 +2055,9 @@ a numbered step. If outdented, it (today) is rendered as a bullet in the midst o
                 <!-- Note: this next step is actually a step on the same level as the above, but bikeshed wanted it indented this much
                 in order to render it as a numbered step. If outdented, it (today) is rendered either as a bullet in the midst
                 of a numbered list or is mis-numbered -->
-                1. <span id='CreateCred-InvokeAuthnrMakeCred'></span>Invoke the [=authenticatorMakeCredential=] operation on |authenticator| with
+                    <li id='CreateCred-InvokeAuthnrMakeCred'>
+                        <!-- @@EDITOR-ANCHOR-01: KEEP THIS LIST SYNC'D WITH OTHER LOCATIONS WITH THIS TAG -->
+                        Invoke the [=authenticatorMakeCredential=] operation on |authenticator| with
                         |clientDataHash|,
                         <code>|pkOptions|.{{PublicKeyCredentialCreationOptions/rp}}</code>,
                         <code>|pkOptions|.{{PublicKeyCredentialCreationOptions/user}}</code>,
@@ -2066,6 +2068,7 @@ a numbered step. If outdented, it (today) is rendered as a bullet in the midst o
                         |enterpriseAttestationPossible|,
                         |attestationFormats|,
                         and |authenticatorExtensions| as parameters.
+                    </li>
 
                 1. [=set/Append=] |authenticator| to |issuedRequests|.
 
--- a/index.bs
+++ b/index.bs
@@ -3731,7 +3731,7 @@ attributes.
 
 ### Authenticator Attachment Enumeration (enum <dfn enum>AuthenticatorAttachment</dfn>) ### {#enum-attachment}
 
-Note: Authenticator Attachment is being deprecated in favor of {{PublicKeyCredentialHint}}. The {{AuthenticatorAttachment/platform}} value is superseded by {{PublicKeyCredentialHint/client-device}}, and {{AuthenticatorAttachment/cross-platform}} is superseded by {{PublicKeyCredentialHint/security-key}}. [=[RPS]=] SHOULD use {{PublicKeyCredentialCreationOptions/hints}} instead of {{AuthenticatorSelectionCriteria/authenticatorAttachment}}.
+Note: Authenticator Attachment is being deprecated in favor of {{PublicKeyCredentialHint}}. The {{AuthenticatorAttachment/platform}} value is superseded by {{PublicKeyCredentialHint/client-device}}, and {{AuthenticatorAttachment/cross-platform}} is superseded by {{PublicKeyCredentialHint/security-key}} and {{PublicKeyCredentialHint/hybrid}}. [=[RPS]=] SHOULD use {{PublicKeyCredentialCreationOptions/hints}} instead of {{AuthenticatorSelectionCriteria/authenticatorAttachment}}.
 
 This enumeration's values describe [=authenticators=]' [=authenticator attachment modality|attachment modalities=].
 [=[RPS]=] use this to express a preferred [=authenticator attachment modality=]
--- a/index.bs
+++ b/index.bs
@@ -3700,7 +3700,7 @@ attributes.
 
 <div dfn-type="dict-member" dfn-for="AuthenticatorSelectionCriteria">
     :   <dfn>authenticatorAttachment</dfn>
-    ::  Note: This member is deprecated in favor of {{PublicKeyCredentialCreationOptions/hints}}. [=[RPS]=] SHOULD use {{PublicKeyCredentialHint/client-device}} instead of {{AuthenticatorAttachment/platform}}, and {{PublicKeyCredentialHint/security-key}} instead of {{AuthenticatorAttachment/cross-platform}} for [=registration ceremony|registration=].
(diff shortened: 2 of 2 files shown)


index.bs
```diff
@@ -2055,20 +2055,17 @@ a numbered step. If outdented, it (today) is rendered as a bullet in the midst o
...
-                            <code>|pkOptions|.{{PublicKeyCredentialCreationOptions/user}}</code>,
-                            |requireResidentKey|,
-                            |userVerification|,
-                            |credTypesAndPubKeyAlgs|,
-                            |excludeCredentialDescriptorList|,
-                            |enterpriseAttestationPossible|,
-                            |attestationFormats|,
-                            and |authenticatorExtensions| as parameters.
-                    </li>
+                1. <span id='CreateCred-InvokeAuthnrMakeCred'></span>Invoke the [=authenticatorMakeCredential=] operation on |authenticator| with
```
akshayku: Bikeshed was not able to build the spec without this. Anyone else seeing this problem?

akshayku: Removed this for now.

index.bs
```diff
@@ -3703,7 +3703,9 @@ attributes.
 
 <div dfn-type="dict-member" dfn-for="AuthenticatorSelectionCriteria">
     :   <dfn>authenticatorAttachment</dfn>
-    ::  If this member is present, eligible [=authenticators=] are filtered to be only those authenticators attached with the specified
+    ::  Note: This member is deprecated in favor of {{PublicKeyCredentialCreationOptions/hints}}. [=[RPS]=] SHOULD use {{PublicKeyCredentialHint/client-device}} instead of {{AuthenticatorAttachment/platform}}, and {{PublicKeyCredentialHint/security-key}} and/or {{PublicKeyCredentialHint/hybrid}} instead of {{AuthenticatorAttachment/cross-platform}} for [=registration ceremony|registration=].
```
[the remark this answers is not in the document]
MasterKale: We talked this over a bit at the WG meeting today. I'm coming around to the idea that there's an opportunity here for the "backwards compatibility" mapping offered here to highlight how **the RP is now empowered to decide** what kind of experience they want their users be steered towards for cross-platform registration.

What I might recommend now is something like this 🤔 

```suggestion
    ::  Note: This member is deprecated in favor of {{PublicKeyCredentialCreationOptions/hints}}. [=[RPS]=] SHOULD use {{PublicKeyCredentialHint/client-device}} instead of {{AuthenticatorAttachment/platform}}, and {{PublicKeyCredentialHint/security-key}} or {{PublicKeyCredentialHint/hybrid}} instead of {{AuthenticatorAttachment/cross-platform}} for [=registration ceremony|registration=], depending on the form factor of the authenticator that users should be encouraged to register.
```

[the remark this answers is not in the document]
harshlal028: > Hints support ordering by preference, a nuance that's not currently captured in this value-mapping guidance. I'd rather [the nuance be thoroughly explained in the section on User-agent Hints Enumeration](https://w3c.github.io/webauthn/#:~:text=Hints%20are%20provided%20in%20order%20of%20decreasing%20preference%20so%2C%20if%20two%20hints%20are%20contradictory%2C%20the%20first%20one%20controls.), but perhaps here we can simply mention it:

Regarding hints ordering - this was discussed a while back, so re-iterating it here:

1. Are hints supposed to be mandatorily enforced by platforms?
2. Does keeping hints as an ordered list makes sense? Platforms would mostly optimize for the first option and hide everything else under "More Options" or something similar. What value does an ordered list or a list for that matter provide?

index.bs
```diff
@@ -3732,6 +3734,8 @@ attributes.
 
 ### Authenticator Attachment Enumeration (enum <dfn enum>AuthenticatorAttachment</dfn>) ### {#enum-attachment}
 
+Note: This member is deprecated in favor of {{PublicKeyCredentialCreationOptions/hints}}. [=[RPS]=] SHOULD use {{PublicKeyCredentialHint/client-device}} instead of {{AuthenticatorAttachment/platform}}, and {{PublicKeyCredentialHint/security-key}} or {{PublicKeyCredentialHint/hybrid}} instead of {{AuthenticatorAttachment/cross-platform}} for [=registration ceremony|registration=], depending on the form factor of the authenticator that users should be encouraged to register.
```
zacknewman, line 3737: I don't think this is phrased correctly. This isn't a "member" but simply an `enum` definition. Additionally this `enum` is _not_ deprecated in the `PublicKeyCredential` `interface`. Ideally, this `enum`'s definition would be elsewhere and not under section 5.4 since it's used elsewhere. Regardless if we move the `enum`'s definition, this should clarify that the `enum` is only deprecated in the context of `AuthenticatorSelectionCriteria`. Alternatively this can be removed altogether since the deprecation notice already exists in the more appropriate place: `AuthenticationSelectionCriteria`.

 If this note is to be retained, perhaps below would be better phrasing:


Note: This `enum` should only be used in `PublicKeyCredential.authenticatorAttachment`. Use of `AuthenticatorSelectionCriteria.authenticatorAttachment` is deprecated in favor of `PublicKeyCredentialCreationOptions.hints`. RPs SHOULD use `PublicKeyCredentialHint::client-device` instead of `AuthenticatorAttachment::platform`, and `PublicKeyCredentialHint::security-key` or `PublicKeyCredentialHint::hybrid` instead of `AuthenticatorAttachment::cross-platform` for registration ceremony registration, depending on the form factor of the authenticator that users should be encouraged to register.

.gitignore
```diff
@@ -1,3 +1,4 @@
 /bikeshed/
```
nsatragno, line 1: General comment:

Many websites pass `attachment: "cross-platform"` and `residentKey: "required"` to use security keys (or phones) as roaming 2fa tokens. Later, they require those to sign in on another device.

With this change, it's possible for users of those websites to store non-syncing platform credentials. But those users won't be able to use those to sign in on new devices. Of course, sites could use the attachment property on the create() response to customize their experience, but existing sites may not do that, since they weren't expecting anything other than cross platform authenticators.

I'm worried that this change can end up getting users locked out of their accounts for these older relying parties.

nsatragno (changes requested):

harshlal028:
Current hints are an ordered list. In reality most OS just use the first item to optimize the UX and everything else is clubbed together. It might make sense to move away from ordered list to a single hint notion - where RPs get to optimize just the first screen.

.gitignore
akshayku: That is exactly the kind of problem I want to solve with this change. That RP has never allowed Platform Authenticator to be used and user wants to use that. User's motivation can be convenience of the biometrics in platform or it's free nature instead of the paid option. 

Such a user should be able to use platform authenticator. 

Regarding the aspect of these whether user can be locked out, user can also get's locked out when security keys are lost etc. 

User will still be asked to register security keys/phones as a first screen. 

Assuming that such a user has security keys, they can always register security keys as well as their platform authenticator. 

User can always go to that platform and register a new credential. 

Assuming that such an RP already requires two security keys for stolen device protection, that will be still be applicable and user will still have to register their second credential. Which means they have backup. 

I don't see this as a great issue. 

Regarding the side discussion we had to not allowing device bound platform authenticator in certain cases, I am totally against such decision in the hands of the RP.

akshayku:
> Current hints are an ordered list. In reality most OS just use the first item to optimize the UX and everything else is clubbed together. It might make sense to move away from ordered list to a single hint notion - where RPs get to optimize just the first screen.

Initially when hints were designed, I had assumed that it will be used for multiple features with each having multiple options and RP will choose one of the options for each feature supported. 

For example, let's say there were two features - Feature_1 (with options A,B,C) and Feature_2 (With Options P,Q,R)

I thought RP can choose one of values from inside from each set. For example, these are below valid options.
- A, P
- A, Q
- B, P
- Q, C
etc. 

In my understanding, RP sending multiple options from same feature was not the initial design. So following options are not valid ones.  
- A, B
- B, C, A

I had not thought of hints being only restricted to transports hints only as one and only one feature. 

AFAIK, every platform uses the first applicable hint for transport and ignores the rest of values. If that is the practical implementation truth and design intention and we all agree, we should clarify the spec.

.gitignore
nsatragno: > user can also get's locked out when security keys are lost etc.

If the user registers a credential in a non-syncing platform authenticator, they won't be able to sign in on a new device, even if they still have access to all their devices. This is worse than the risk of losing a security key and holds even if the user registers more than one credential on their platform authenticator.

> Assuming that such an RP already requires two security keys for stolen device protection, that will be still be applicable and user will still have to register their second credential. Which means they have backup.

Not if the RP does not verify that the attachment of the credential they got back is `cross-platform`. And why would they, if they set the attachment to `cross-platform` on the request? Existing sites won't do that, so from the perspective of the site the platform credentials will look like security keys.

> I don't see this as a great issue.

This change presents significant backwards compatibility risk for users on platforms with non-syncing authenticators who won't understand the consequences of storing their 2fa credentials on them. I don't think it's shippable as it's currently proposed.

Issue #2321, opened 2025-08-14
privacy implications of cross-origin iframe


npdoty:
Is this intended to support signing in to one relying party when that party is embedded on a different site? Are users supposed to distinguish which party they are signing into when they do this? That seems extremely ripe for confusion. It could be useful for tracking users across sites if the user is trying to sign in to Site A without realizing that what they are doing is providing their cross-origin identifier for Tracker B.

In what way are passkeys partitioned when accessed by a cross-origin embedded iframe?

(This issue includes several questions because the reviewer (that @npdoty guy) wasn't entirely confident in reading the spec on the exact implications, and the rest of the Privacy WG thought it was potentially very concerning, but couldn't be certain based on the reviewer's uncertainty.)

This item was raised and discussed by the Privacy WG as part of this privacy review:
https://github.com/w3cping/privacy-request/issues/162

timcappalli:
@npdoty did my responses address your concerns?

PR #2391, opened 2026-02-16
Privacy considerations for cross-origin usage

### LINKED ISSUE #2319: related origins enables sharing across relying parties
>  Although Credential IDs and credential public keys are necessarily shared with the WebAuthn Relying Party to enable strong authentication, they are designed to be minimally identifying and not shared between Relying Parties.

But they are designed to specifically be shared with any related origin, which could include hundreds of other relying parties if they are collaborating.

Related origins don't seem to be mentioned in any privacy considerations. How could this feature be abused? Can origin lists be updated over time? Will the origins actually be related companies, or just colluding? Will the user know that if they use their passkey on this totally differently named and branded site that it will provide hard proof that it's the same user on what seems to be a totally different site? Will silent access be allow for related origins if a user asked to stay signed in? Enabling cryptographically-verified identity-sharing between many origins is a substantial privacy risk, with the potential for unwanted cross-site recognition.

How does this feature relate to every other web proposal for related websites? For example, it seems to have fewer restrictions, substantive protections than the first-party sets proposals. Should websites that use one related-origins feature use the same list of origins as they do in the other related-origins features? 

We were also confused about the exact implications of registrable origin labels (vs eTLD+1, or other known concepts).

Setting a minimum of 5 and no maximum is confusing. Why 5? Leaving it ambiguous doesn't seem to help interoperability. Having a minimum but no maximum doesn't seem to help privacy. (It seemed like implementations were converging on 5 as just both floor and ceiling?)

This item was raised and discussed by the Privacy WG as part of this privacy review:
https://github.com/w3cping/privacy-request/issues/162

timcappalli:
Closes #2319 #2321

Addresses feedback from the ongoing privacy and security reviews of Related Origins and iframe usage in WebAuthn L3.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -4502,6 +4502,8 @@ feature-identifier token, respectively, in the <{iframe/allow}> attribute's valu
 
 [=[RPS]=] utilizing the WebAuthn API in an embedded context should review [[#sctn-seccons-visibility]] regarding [=UI redressing=] and its possible mitigations.
 
+[=WebAuthn Clients=] and [=client platforms=] should refer to [[#sctn-cross-origin-use]] for additional privacy considerations.
+
 ## Using Web Authentication across related origins ## {#sctn-related-origins}
 
 By default, Web Authentication requires that the [=RP ID=] be equal to the [=determines the set of origins on which the public key credential may be exercised|origin=]'s [=effective domain=], or a [=is a registrable domain suffix of or is equal to|registrable domain suffix=] of the [=determines the set of origins on which the public key credential may be exercised|origin=]'s [=effective domain=].
@@ -4541,6 +4543,8 @@ Requests to this well-known endpoint by [=WebAuthn Clients=] MUST be made withou
 and using the `https:` [=scheme=]. When following redirects, [=WebAuthn Clients=] MUST explicitly require all redirects to also use the `https:` [=scheme=].
 
 [=WebAuthn Clients=] supporting this feature SHOULD include {{ClientCapability/relatedOrigins}} in their response to [[#sctn-getClientCapabilities|getClientCapabilities()]].
+
+[=WebAuthn Clients=] and [=client platforms=] should refer to [[#sctn-cross-origin-use]] for additional privacy considerations.
 
 ### Validating Related Origins ### {#sctn-validating-relation-origin}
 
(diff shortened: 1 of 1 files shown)


MasterKale (approved):
Clear and succinct. I think this is fine

selfissued (approved):

harshlal028 (approved):
LGTM

npdoty:
Helpful to document both the risk and a normative requirement indicating the responsibility for informing the user, as a mitigation.

One confusion I had: the RP ID and calling origin are both useful to show in the Related Origins case (because the origins might be different, and those are the two origins that will be connected by approving the authentication), but in the embedded document case, isn't the calling origin the same as the RP ID, and it's just different from the embedding website? The user confusion there isn't that the calling origin is different from the RP ID from creation time, but that it might be different (and connecting activity or accounts between) the top-level document origin.

timcappalli:
> The user confusion there isn't that the calling origin is different from the RP ID from creation time, but that it might be different (and connecting activity or accounts between) the top-level document origin.

The top origin is always visible in the omnibox.

timcappalli:
@npdoty I've added more text explicitly for the iframe scenario. PTAL.

npdoty:
I think the explicit example for the iframe is useful, thanks. 

The suggested UI language could also be framed as "examplepay.com can track your behavior on example.com" or something to be more similar to what implementations are doing with Storage Access API, but I don't have a strong preference there -- and certainly the goal is not to standardize UI language.

simoneonofri (approved):

nadalin:
Are we ready to merge this PR? I think we have accommodated the security/privacy concerns requested.

index.bs
```diff
@@ -9031,6 +9035,38 @@ The {{getClientCapabilities}} method assists [=[WRPS]=] in crafting registration
...
+[=WebAuthn Clients=] and [=client platforms=] SHOULD ensure
+that the credential and authenticator selection interfaces
+provide clear context to the user by displaying both:
+
+- The caller's origin: The [=origin=] of the [=fully active=] document that is invoking `navigator.credentials`
+- The Relying Party ID: The [=RP ID=] for which the credential was originally created or will be created
+
+For example, given the Relying Party ID `example.com` with a calling origin of `https://myexample.com`,
+a [=WebAuthn client=] and/or [=client platform=] could show the user:
+"Sign in to myexample.com with your passkey for example.com?"
```
emlun: The two examples `myexample.com` and `example.com` could probably benefit from being a bit more distinct? Say maybe `exampleA.com` and `otherExampleB.org`?

index.bs
```diff
@@ -9031,6 +9035,38 @@ The {{getClientCapabilities}} method assists [=[WRPS]=] in crafting registration
...
+For scenarios where the calling [=origin=] and [=RP ID=] are different,
+such as with [[#sctn-related-origins|Related Origin Requests]],
+[=WebAuthn Clients=] and [=client platforms=] SHOULD ensure
+that the credential and authenticator selection interfaces
+provide clear context to the user by displaying both:
+
+- The caller's origin: The [=origin=] of the [=fully active=] document that is invoking `navigator.credentials`
+- The Relying Party ID: The [=RP ID=] for which the credential was originally created or will be created
+
+For example, given the Relying Party ID `example.com` with a calling origin of `https://myexample.com`,
```
emlun: Probably?
```suggestion
For example, given the [=RP ID=] `example.com` with a calling origin of `https://myexample.com`,
```

index.bs
```diff
@@ -9031,6 +9035,38 @@ The {{getClientCapabilities}} method assists [=[WRPS]=] in crafting registration
...
+"Sign in to myexample.com with your passkey for example.com?"
+
+For scenarios where WebAuthn is called in an [[#sctn-iframe-guidance|iframe]],
+the [=WebAuthn client=] and/or [=client platform=] SHOULD also ensure that
+the credential and authenticator selection interfaces provide a warning to users 
+and additional context around the [=top-level origin=] of the page in which the iframe is embedded,
+as this may differ from the caller's origin. A successful ceremony allows both the
+[=top-level origin=] and the calling [=origin=] to correlate the user's identity across their respective sites.
+
+For example, given the Relying Party ID `examplepay.com` with a calling origin of `https://examplepay.com`
```
emlun: Probably?
```suggestion
For example, given the [=RP ID=] `examplepay.com` with a calling origin of `https://examplepay.com`
```

emlun (approved):
LGTM with minor editorial nits. Thanks!

PR #2426, opened 2026-05-13
Editorial tweaks to PR #2377

emlun:
This is a meta-PR that would merge into #2377, not directly to main.

I propose some fairly straightforward editorial tweaks:

- I don't think we need the `id` attributes for these `<dfn>`s, so we can delete them.
- "credential record" uses `dfn-type="abstract-op"`, so let's use the same `dfn-type` for the extensions to "credential record".
- Various tweaks to punctuation, capitalization etc.
- We can use the [`.cbor` control operator](https://datatracker.ietf.org/doc/html/rfc8610#section-3.8.4) to indicate that the `cmtgKey` output should contain CBOR matching the COSE_Key type.
- A couple of phrasing tweaks to make the text a bit more concise and precise.

***

### :boom: Error: 422 Unprocessable Entity :boom: ###

[PR Preview](https://github.com/tobie/pr-preview#pr-preview) failed to build. _(Last tried on May 13, 2026, 2:22 PM UTC)_.

<details>
<summary>More</summary>

:rotating_light: [Spec Generator](https://www.w3.org/publications/spec-generator/) - Spec Generator is the web service used to build bikeshed/ReSpec specs

:link: [Related URL](https://www.w3.org/publications/spec-generator/?type=bikeshed-spec&output=html&url=https%3A%2F%2Fraw.githubusercontent.com%2Fw3c%2Fwebauthn%2F3238904ba7186bf82876f9163a941b48a391b1dc%2Findex.bs&force=1&md-warning=not%20ready)

**Error output:**

```json
[
    {
        "lineNum": "2074:21",
        "messageType": "fatal",
        "text": "Tried to auto-close a <li>, but there were unclosed elements remaining before the nearest matching start tag (at 1930:5).\nOpen tags: <div> at 1712:1, <li> at 1930:5, <dl> at 1933:9"
    },
    {
        "lineNum": "8863",
        "messageType": "warning",
        "text": "The heading 'Well-Known URI Registration' needs a manually-specified ID."
    },
    {
        "lineNum": null,
        "messageType": "failure",
        "text": "Did not generate, due to errors exceeding the allowed error level."
    }
]
```

_This seems to be an issue with the [Spec Generator](https://www.w3.org/publications/spec-generator/) service. PR Preview doesn't manage this service and so has no control over it. If you've identified an issue with it, you can [report the issue to the maintainers of Spec Generator](https://github.com/w3c/spec-generator/issues/new) directly. Please be courteous. Thank you!_

_If you don't have enough information above to solve the error by yourself or if the issue doesn't seem related to Spec Generator, you can [file an issue with PR Preview](https://github.com/tobie/pr-preview/issues/new?title=Unidentified%20Error&body=See%20w3c/webauthn%232426.)._

</details>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -7931,7 +7931,7 @@ A <dfn>CMTG Key Record</dfn> is a [=struct=] with the following [=struct/items=]
 
 <dl dfn-for="CMTG Key Record" dfn-type="abstract-op">
     : <dfn>publicKey</dfn>
-    :: The [=Credential Manager Trust Group public key|CMTG public key=]
+    :: The [=Credential Manager Trust Group public key|CMTG public key=].
 
     : <dfn>lastUsed</dfn>
     :: A timestamp representing the last time this specific key was verified by the [=Relying Party=].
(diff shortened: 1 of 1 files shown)


timcappalli (approved):
Thanks @emlun!

PR #2427, opened 2026-05-15
Document compatible Bikeshed version for L3

emlun:
As of this commit, the spec fails to build with Bikeshed versions greater than 5.4.2. We should document this in the repo for posterity.


simoneonofri (approved):
Thank you!

aliassiri1979-eng:
👅👅🔥

Issue #2417, opened 2026-04-22
Post Quantum Crypto and WebAuthn Transition for RP


akshayku:
Opening this issue for discussion around PQC transition for RPs in WebAuthn

## RP Flow
Currently we have defined algorithms in IANA for 3 ML-DSA algorithms (-48, -49, -50) which allows RP to register those PQC algorithms. 

We have to think about the transitions and flexibilities RP needs to have to react to crypto advancements. As PQC algorithms are new and not fully time tested, RP needs to prepare for both classical and PQC algorithms getting broken for some time. Also, user should not be asked to setup new credentials at that point in time because their current authentication is already vulnerable. RP should be ready in advance for these likely scenarios. 

### Issue 1: RP cannot have two different algorithm credential in same authenticator for same account.
- Resident/Discoverable Credentials are overwritten for same userID/RPID but different algorithms.
- For the same userID/RPID, we currently say that authenticator should overwrite the credentials irrespective of algorithms being different.
- RP can workaround this today by supplying different userID for the same account, but that results in user being shown two different tiles for same account. 
- Platform cannot determine that these are the same account in authentication phase if userID is different.

### Proposed Change 1: Overwrite the existing credential if RPID, UserID and Algorithm are the same. 
- Addition of algorithm in overwrite logic will allow two different algorithms to exist for the same RP in same authenticator. 
- UserID can remain the same for the those different algorithms credentials and platform can do heuristics to show only one of the credentials based on RP's algorithm preference

### Issue 2: In usernameless flows, RP cannot say which algorithm credentials are allowed for authentication
- Assuming that RP is reacting to potential scenarios below, RP needs to make sure that user is not authentication with broken algorithms. 
    - Scenario 1: Classical Algorithm is broken
    - Scenario 2: Current defined ML-DSA PQC algorithms gets broken
        - Given PQC algorithms are new and not time tested, this possibility exists

### Proposed Change 2: Add allowed algorithms preference to authentication ceremony similar to registration ceremony
- This will allow RPs to specify which algorithms are allowed to do authentication in usernameless flows also. 
- This will then allow platforms to filter out credentials which RP does not trust anymore taking into account the userID.

MasterKale:
> ### Proposed Change 1: Overwrite the existing credential if RPID, UserID and Algorithm are the same.
> * Addition of algorithm in overwrite logic will allow two different algorithms to exist for the same RP in same authenticator.
[...]

PR #2433, opened 2026-06-10
[WIP] PQC Migration Explainer

akshayku:
Proposal for issue #2417. 

This is a proposed design which migrates credentials from one algorithm to another. Currently focus is on classical algorithm to PQC algorithm, but it works for any COSE algorithm family to any other COSE algorithm family. 

This proposal is open for feedback.


timcappalli (approved):
Just an explainer, so approving publishing of it. Did not review contents.

nsatragno:
Why does this need a review? Bring the wiki back!

nsatragno (approved):

Issue #2144, opened 2024-09-10
Allow Conditional Mediation without autofill


kenrb:
_This is not intended for level 3._

## Proposed Change
I have added an [explainer](https://github.com/w3c/webauthn/wiki/Explainer:-WebAuthn-Ambient-Signin-UI) to the wiki proposing a new UI mechanism and a way for sites to trigger it. The motivation is to provide conditional UI functionality that is not tied to autofill, so it can be used to sign in to an RP without necessarily being on a dedicated sign-in page.

As with existing conditional UI, the user agent would ignore the request if there are no eligible credentials for the current RP.

At Google we have been tentatively labeling this "ambient UI", reflecting that we are imagining an unobtrusive non-modal UI surface that can be ignored by the user.

sbweeden:
Would it be reasonable to suggest this is loosely akin to the FedCM experience?

agl:
> Would it be reasonable to suggest this is loosely akin to the FedCM experience?

Yes, albeit with the important qualifier that this would _only_ appear if the user has existing credentials for the site. It would not appear otherwise, while FedCM can.

MasterKale:
Is there a desire to support display of both the autofill UI as well as "ambient" UI? If it would be best for it to be either/or, what if an RP opted into **ambient** display by **omitting** the `autocomplete` tokens that are currently required in L3? 🤔 

![Screenshot 2024-09-10 at 4 55 16 PM](https://github.com/user-attachments/assets/778f5d4a-fc2a-44f8-af36-e4bdb84fd225)

https://w3c.github.io/webauthn/#GetAssn-ConditionalMediation-Interact-FormControl

agl:
> That is to say, if an RP doesn't specify autocomplete tokens to indicate explicitly where the conditional UI should be displayed, then e.g. "the user agent is free to choose how to display relevant UI to initiate an authentication ceremony"

I would worry about races where the DOM elements are constructed by Javascript after page load, but the browser has already looked and decided that there weren't any.

emlun:
>This proposes a new field be added to [PublicKeyCredentialRequestOptions](https://www.w3.org/TR/webauthn-3/#dictdef-publickeycredentialcreationoptions). The field is called display, and initially has two values: autofill and ambient. When unspecified, its default value is used, which is autofill.
>
>Setting the display field in a request that does not specify conditional mediation results in an error.

[...]

PR #2435, opened 2026-06-23
Ambient explainer update

kenrb:
Added some additional context and brought the API form up to date with changes that were made related to Immediate.


explainers/ambient-sign-in.md
```diff
@@ -4,54 +4,52 @@
...
-    display: 'ambient',
   },
   password: true,
 });
 ```
 
 ## Other Credential Types
 
-Requests for other credential types, such as [PasswordCredential](https://www.w3.org/TR/credential-management-1/#passwordcredential) or [FederatedCredential](https://www.w3.org/TR/credential-management-1/#federatedcredential), can already trigger immediate user agent UI. User agents can choose to simultaneously display those alongside WebAuthn credentials without any change to the Credential Management specification.
+Ambient behaves the same as [Federated Credential Management passive mode]([https://www.w3.org/TR/fedcm/](https://w3c-fedid.github.io/FedCM/#dom-identitycredentialrequestoptionsmode-passive). Browsers that support both can offer them in integrated UI.
```
MasterKale: This looks like it might have been a mis-paste that mangled the URL markdown. Is this what you wanted this link to look like?

```suggestion
Ambient behaves the same as [Federated Credential Management passive mode](https://w3c-fedid.github.io/FedCM/#dom-identitycredentialrequestoptionsmode-passive). Browsers that support both can offer them in integrated UI.
```

explainers/ambient-sign-in.md
```diff
@@ -4,54 +4,52 @@
...
 
 ![Bubble in top-right of browser window showing a WebAuthn credential](assets/ambient_ui_concept.png)
 
 ## API
 
-This proposes a new field be added to [PublicKeyCredentialRequestOptions](https://www.w3.org/TR/webauthn-3/#dictdef-publickeycredentialcreationoptions). The field is called `display`, and initially has two values: `autofill` and `ambient`. When unspecified, its default value is used, which is `autofill`.
+Feature detection is provided by another enumeration value in ClientCapability: `ambientGet`.
 
-Setting the display field in a request that does not specify conditional mediation results in an error.
+When the request contains `mediation: "conditional"` and the PublicKeyCredentialRequestOptions contains `uiMode: "passive"`, the user agent displays discoverable WebAuthn credentials immediately in an unobtrusive UI prompt.
```
MasterKale, line 33: Is `mediation: "conditional"` _really_ necessary to trigger the ambient UI experience?

The reason I ask is, one of the motivations listed above is that "Conditional UI is not useable because there is not a sign-in form on the page." It doesn't make sense to me on pages that don't have a login form why the RP would call `.get()` with `mediation: "conditional"` when the page would otherwise not be in a state to do so.

Could not the logic for triggering ambient UI simply be the presence of `uiMode: "passive"`?

```js
const cred = await navigator.credentials.get({
  // mediation: 'conditional',  // <-- Omitted
  uiMode: 'passive',  // <-- This alone triggers ambient UI
  publicKey: {
    challenge: ...,
    rpId: 'example.com',
    allowCredentials: {...},
    display: 'ambient',
  },
  password: true,
});
```

Then, on pages that **do** feature a login form, the RP can **combine** `uiMode: "passive"` with `mediation: "conditional"` to get the typical browser autofill on the form `<input>`s, while also triggering ambient UI at the same time.

timcappalli (approved):

nsatragno:
Leaving unmerged in case @kenrb wants to address the feedback.

(Note I believe the point is to not block on reviews for explainers to be merged.)

nsatragno (approved):

MasterKale (approved):
Okay, if this explainer PR will reboot discussion in https://github.com/w3c/webauthn/issues/2144 then I can redirect my questions to there.

explainers/ambient-sign-in.md
kenrb: Thanks. Fixed.

PR #2437, opened 2026-06-30
Support Algorithm Migration

### LINKED ISSUE #2417: Post Quantum Crypto and WebAuthn Transition for RP

akshayku:
Closes #2417 

Features:
- Silent Migration with no extra UI for migration during sign-in/get() as and when authenticator supports multiple algorithms. 

Spec Changes
- algPolicy Extension
- acceptedAlgs in GetAssertion Options

Implementation commitment:

Documentation and checks

- [ ] Affects privacy
- [ ] Affects security

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/pull/2437.html" title="Last updated on Jul 29, 2026, 7:00 PM UTC (2890b60)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/2437/1c97d87...2890b60.html" title="Last updated on Jul 29, 2026, 7:00 PM UTC (2890b60)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2717,7 +2717,7 @@ The steps for [=issuing a credential request to an authenticator=] are as follow
             </dl>
 
     1. Let |acceptedAlgs| be
-       <code>|pkOptions|.{{PublicKeyCredentialRequestOptions/acceptedAlgs}}</code>.
+        <code>|pkOptions|.{{PublicKeyCredentialRequestOptions/acceptedAlgs}}</code>.
 
     1. <span id="allowCredentialDescriptorListCreation"></span>
         If <code>|pkOptions|.{{PublicKeyCredentialRequestOptions/allowCredentials}}</code>
@@ -7988,6 +7988,13 @@ pair, [=signature counter=], and [=backup state|backup eligibility and state=].
     credHash = SHA-256( credIdLen || credId || COSE_Key )
     </pre>
 
+    The bytes hashed are exactly the portion of the created [=credential=]'s [=attested credential
+    data=] that follows the leading [=authData/attestedCredentialData/aaguid|AAGUID=].
+    That is, credIdLen is the credential's [=authData/attestedCredentialData/credentialIdLength=]
+    (the 2-byte, 16-bit unsigned big-endian length L),
+    credId is its [=authData/attestedCredentialData/credentialId=] (L bytes), and
+    COSE_Key is its [=authData/attestedCredentialData/credentialPublicKey=].
+
     `existingCredentials` are emitted whenever [=algPolicy|this extension=] is present, including
     when `createAlgs` is the empty array.
     `existingCredentials` is a per-authenticator, unordered set scoped to the same ([=RP ID=],[=user
@@ -7995,8 +8002,12 @@ pair, [=signature counter=], and [=backup state|backup eligibility and state=].
 
     The unsigned part conveys the [=credentials=] created in this ceremony.
     It is a CBOR array of [=authenticatorMakeCredential=] responses as per [[FIDO-CTAP]] spec.
-    [=Client=] converts each created credentials CBOR output to
-    AuthenticatorAttestationResponseJSON entries.
+    For each created [=credential=], the [=client=] assembles a {{PublicKeyCredential}} exactly as it
+    would for a {{CredentialsContainer/create()}} of that credential — including its
+    [=attestation object=] and the [=client extension outputs=].
+    When the output is serialized via <code>{{PublicKeyCredential}}.{{PublicKeyCredential/toJSON()}}</code>,
+    each entry is represented as a {{RegistrationResponseJSON}} in the corresponding
+    `createdCredentials` member of {{AuthenticationExtensionsAlgPolicyOutputsJSON}}.
 
 : Client extension output
 :: <xmp class="idl">
@@ -8004,11 +8015,14 @@ pair, [=signature counter=], and [=backup state|backup eligibility and state=].
         AuthenticationExtensionsAlgPolicyOutputs algPolicy;
     };
     partial dictionary AuthenticationExtensionsClientOutputsJSON {
-        AuthenticationExtensionsAlgPolicyOutputs algPolicy;
+        AuthenticationExtensionsAlgPolicyOutputsJSON algPolicy;
     };
 
     dictionary AuthenticationExtensionsAlgPolicyOutputs {
-        sequence<AuthenticatorAttestationResponseJSON> createdCredentials;
+        sequence<PublicKeyCredential> createdCredentials;
+    };
+    dictionary AuthenticationExtensionsAlgPolicyOutputsJSON {
+        sequence<RegistrationResponseJSON> createdCredentials;
     };
     </xmp>
 
@@ -8017,6 +8031,12 @@ pair, [=signature counter=], and [=backup state|backup eligibility and state=].
         ::  Zero or more new [=credentials=] created during this ceremony, one entry per
             {{AuthenticationExtensionsAlgPolicyInputs/createAlgs}} group the [=authenticator=] chose
             to satisfy.
+
+            Each entry is a full registration response for the created [=credential=], as a
+            {{PublicKeyCredential}} (or a {{RegistrationResponseJSON}} in the JSON serialization),
+            carrying both the credential's [=attestation object=] and the [=client extension
+            outputs=] resulting from any {{AuthenticationExtensionsAlgPolicyInputs/createExtensions}}
+            inputs.
 
             The [=[RP]=] MUST treat this list as opportunistic: a group not filled in this response
             MAY be filled on a later ceremony.
--- a/index.bs
+++ b/index.bs
@@ -916,6 +916,11 @@ BCP 14 [[!RFC2119]] [[!RFC8174]] when, and only when, they appear in all capital
     then each compartment is considered a separate [=authenticator=]
     with a single user with no access to other users' [=credentials=].
 
+:<dfn>Multiple Credentials Per User Account</dfn>
+:: A [WAA] which supports storing multiple credentials corresponding to different
+    {{COSEAlgorithmIdentifier}} for same the [=user account=].
+    These credentials has the same [=public key credential source/rpId=] and [=user handle=].
+
 : <dfn>Authorization Gesture</dfn>
 :: An [=authorization gesture=] is a physical interaction performed by a user with an authenticator as part of a [=ceremony=],
     such as [=registration=] or [=authentication=]. By making such an [=authorization gesture=], a user [=user consent|provides
@@ -2711,6 +2716,9 @@ The steps for [=issuing a credential request to an authenticator=] are as follow
 
             </dl>
 
+    1. Let |acceptedAlgs| be
+       <code>|pkOptions|.{{PublicKeyCredentialRequestOptions/acceptedAlgs}}</code>.
+
     1. <span id="allowCredentialDescriptorListCreation"></span>
         If <code>|pkOptions|.{{PublicKeyCredentialRequestOptions/allowCredentials}}</code>
         <dl class="switch">
@@ -2747,21 +2755,21 @@ The steps for [=issuing a credential request to an authenticator=] are as follow
 
                             Then, using |transport|, invoke the [=authenticatorGetAssertion=] operation on
                             |authenticator|, with |rpId|, |clientDataHash|, |allowCredentialDescriptorList|,
-                            |userVerification|,
-                            and |authenticatorExtensions| as parameters.
+                            |userVerification|, |authenticatorExtensions|,
+                            and |acceptedAlgs| as parameters.
 
                         :   [=list/is empty=]
                         ::  Using local configuration knowledge of the appropriate transport to use with |authenticator|,
                             invoke the [=authenticatorGetAssertion=] operation on |authenticator| with |rpId|,
                             |clientDataHash|, |allowCredentialDescriptorList|, |userVerification|,
-                            and |authenticatorExtensions| as parameters.
+                            |authenticatorExtensions|, and |acceptedAlgs| as parameters.
                     </dl>
 
             :   [=list/is empty=]
             ::  Using local configuration knowledge of the appropriate transport to use with |authenticator|, invoke the
                 [=authenticatorGetAssertion=] operation on |authenticator| with |rpId|, |clientDataHash|,
-                |userVerification|,
-                and |authenticatorExtensions| as parameters.
+                |userVerification|, |authenticatorExtensions|,
+                and |acceptedAlgs| as parameters.
 
                 Note: In this case, the [=[RP]=] did not supply a list of acceptable credential descriptors. Thus, the
                     authenticator is being asked to exercise any credential it may possess that is [=scoped=] to
@@ -2990,6 +2998,7 @@ value and terminate the operation.
         sequence<PublicKeyCredentialDescriptorJSON>             allowCredentials = [];
         DOMString                                               userVerification = "preferred";
         sequence<DOMString>                                     hints = [];
+        sequence<COSEAlgorithmIdentifier>                       acceptedAlgs = [];
         AuthenticationExtensionsClientInputsJSON                extensions;
     };
 </xmp>
@@ -3868,6 +3877,7 @@ an assertion. Its {{PublicKeyCredentialRequestOptions/challenge}} member MUST be
         sequence<PublicKeyCredentialDescriptor> allowCredentials = [];
         DOMString                            userVerification = "preferred";
         sequence<DOMString>                  hints = [];
+        sequence<COSEAlgorithmIdentifier>    acceptedAlgs = [];
         AuthenticationExtensionsClientInputs extensions;
     };
 </xmp>
@@ -3931,6 +3941,21 @@ an assertion. Its {{PublicKeyCredentialRequestOptions/challenge}} member MUST be
 
     :   <dfn>hints</dfn>
     ::  This OPTIONAL member contains zero or more elements from {{PublicKeyCredentialHint}} to guide the user agent in interacting with the user. Note that the elements have type `DOMString` despite being taken from that enumeration. See [[#sct-domstring-backwards-compatibility]].
+
+    :   <dfn>acceptedAlgs</dfn>
+    ::  This OPTIONAL member lists the {{COSEAlgorithmIdentifier}}s, in descending order of
+        preference, that the [=[RP]=] is willing to accept for this [=authentication ceremony=]
+        and acts as a filter in selection of eligible credentials set.
+
+        This filter is only applicable if [=authenticator=] supports
+        [=multiple credentials per user account=].
+        If [=authenticator=] does not support [=multiple credentials per user account=],
+        this filter is ignored.
+
+        This filter is additive to {{PublicKeyCredentialRequestOptions/allowCredentials}} in
+        figuring out which credential to pick for this [=authentication ceremony=].
+
+        See [[#sctn-op-get-assertion]] for more details.
 
     :   <dfn>extensions</dfn>
     ::  The [=[RP]=] MAY use this OPTIONAL member to provide [=client extension inputs=]
@@ -5312,6 +5337,11 @@ This operation takes the following input parameters:
 : |extensions|
 :: A [=CBOR=] [=map=] from [=extension identifiers=] to their [=authenticator extension inputs=], created by the client based on
     the extensions requested by the [=[RP]=], if any.
+: |acceptedAlgs|
+:: An OPTIONAL [=list=] of {{COSEAlgorithmIdentifier}}s, in descending order of preference, that
+    restricts and orders the credentials eligible for this assertion.
+    Supplied by the client from
+    <code>|pkOptions|.{{PublicKeyCredentialRequestOptions/acceptedAlgs}}</code>, if any.
 
 When this method is invoked, the [=authenticator=] MUST perform the following procedure:
 
@@ -5326,6 +5356,22 @@ When this method is invoked, the [=authenticator=] MUST perform the following pr
     authenticator's [=credentials map=], [=set/append=] |credSource| to |credentialOptions|.
 1. [=list/Remove=] any items from |credentialOptions| whose [=public key credential source/rpId=] is not equal to
     |rpId|.
+1. If [=authenticator=] supports [=multiple credentials per user account=], group the credentials
+    in |credentialOptions| per [=user handle=].
+    Then for each group, if
+    <dl class="switch">
+        :   |acceptedAlgs| was supplied and [=list/is not empty=]
+        ::  [=list/Remove=] any items from |credentialOptions| whose
+            {{COSEAlgorithmIdentifier|COSE algorithm}} does not have an entry in |acceptedAlgs|.
+            Then, if multiple entries remains for this group, retain one entry whose
+            {{COSEAlgorithmIdentifier|COSE algorithm}} appear earliest in the |acceptedAlgs| and
+            [=list/Remove=] other entries, if any remaining, from |credentialOptions|.
+
+        :   Otherwise
+        ::  Retain one entry from the group chosen by an authenticator-defined preference,
+            for example the most recently created or used and [=list/Remove=] other entries,
+            if any remaining, from |credentialOptions|.
+    </dl>
 1. If |credentialOptions| is now empty, return an error code equivalent to "{{NotAllowedError}}" and terminate the operation.
 
 <!-- Note: this next step is actually a top-level step, but bikeshed wanted it indented this much in order to render it as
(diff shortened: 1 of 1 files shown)


nsatragno:
Can you split the explainer into a different patch?

It's easier to review if we have a relatively polished explainer of what you're trying to accomplish in your own words, and a separate specification PR. Now it's tough to see the latest version of the explainer.

akshayku:
Moved explainer to #2438

index.bs
```diff
@@ -5329,6 +5359,27 @@ When this method is invoked, the [=authenticator=] MUST perform the following pr
     authenticator's [=credentials map=], [=set/append=] |credSource| to |credentialOptions|.
 1. [=list/Remove=] any items from |credentialOptions| whose [=public key credential source/rpId=] is not equal to
     |rpId|.
+1. If |acceptedAlgs| was supplied and [=list/is not empty=], [=list/remove=] from
+   |credentialOptions| any [=public key credential source=] the {{COSEAlgorithmIdentifier|COSE
+   algorithm}} of whose [=credential private key=] is not contained in |acceptedAlgs|.
```
MasterKale: Proposing something that seems a bit more straight-forward to mentally process:

```suggestion
1. If |acceptedAlgs| was supplied and [=list/is not empty=], [=list/remove=] from
   |credentialOptions| any [=public key credential source=] with a [=credential private key=] {{COSEAlgorithmIdentifier|COSE
   algorithm}} that is not contained in |acceptedAlgs|.
```

index.bs
```diff
@@ -7760,6 +7811,250 @@ However, [=authenticators=] that do not utilize [[!FIDO-CTAP]] do not necessaril
...
 :: [=largeblob|This extension=] directs the user-agent to cause the large blob to be stored on, or retrieved from, the authenticator. It thus does not specify any direct authenticator interaction for [=[RPS]=].
 
+### Algorithm Policy extension (<dfn>algPolicy</dfn>) ### {#sctn-alg-policy-extension}
+
+This [=client extension|client=] [=authentication extension=] lets a [=[RP]=] migrate a [=user
+account=]'s [=credentials=] between {{COSEAlgorithmIdentifier}}s during ordinary [=authentication
+ceremonies=], without an explicit re-registration ceremony and without a window in which the account
+cannot authenticate.
+
+The extension depends on a relaxed [=credential=] model: an [=authenticator=] that supports
```
MasterKale: > relaxed [=credential=] model

Is this really a "relaxed" credential model? A second credential for (rpId, user.id) with a different alg gets its own ID, keypair, counter, be+bs flags...for all intents and purposes it's "another credential."

I guess my concern is that calling this a "relaxed" model might imply that passkey algorithm rotation via this extension leaves users in a "weakened" or "not as strong" security posture. Maybe we call it "alternative" to avoid any such misunderstandings?

MasterKale (commented):
Some initial thoughts on what I managed to read before the long US holiday weekend. I'll jump back into this next week to review the rest.

index.bs
```diff
@@ -7760,6 +7806,229 @@ However, [=authenticators=] that do not utilize [[!FIDO-CTAP]] do not necessaril
...
+        AuthenticationExtensionsAlgPolicyInputs algPolicy;
+    };
+    partial dictionary AuthenticationExtensionsClientInputsJSON {
+        AuthenticationExtensionsAlgPolicyInputsJSON algPolicy;
+    };
+
+    dictionary AuthenticationExtensionsAlgPolicyInputs {
+        required sequence<sequence<COSEAlgorithmIdentifier>> createAlgs;
+        AuthenticationExtensionsClientInputs createExtensions;
+        sequence<BufferSource> deleteCredentials;
```
nsatragno, line 7838: We already have the signal API and this feature is very complicated already. If we decide to do this, it should be its own extension, and its own PR.

If the motivation for this new feature is that the signal API does not work with security keys, I think shipping a new API is not the way to go about solving that. Fundamentally, there is nothing stopping the signal API from working with security keys today.

index.bs
```diff
@@ -7760,6 +7806,229 @@ However, [=authenticators=] that do not utilize [[!FIDO-CTAP]] do not necessaril
...
+    partial dictionary AuthenticationExtensionsClientInputs {
+        AuthenticationExtensionsAlgPolicyInputs algPolicy;
+    };
+    partial dictionary AuthenticationExtensionsClientInputsJSON {
+        AuthenticationExtensionsAlgPolicyInputsJSON algPolicy;
+    };
+
+    dictionary AuthenticationExtensionsAlgPolicyInputs {
+        required sequence<sequence<COSEAlgorithmIdentifier>> createAlgs;
+        AuthenticationExtensionsClientInputs createExtensions;
```
nsatragno, line 7837: This proposal is missing the extension outputs. Not every extension output is present in the authenticator data, e.g. large blob.

index.bs
```diff
@@ -5315,6 +5340,11 @@ This operation takes the following input parameters:
 : |extensions|
 :: A [=CBOR=] [=map=] from [=extension identifiers=] to their [=authenticator extension inputs=], created by the client based on
     the extensions requested by the [=[RP]=], if any.
+: |acceptedAlgs|
```
nsatragno: What should the behaviour be if an authenticator has multiple credentials for the same { rp id, user id } and the `acceptedAlgs` parameter is not supplied? This needs to be spelled out.

index.bs
```diff
@@ -7760,6 +7806,229 @@ However, [=authenticators=] that do not utilize [[!FIDO-CTAP]] do not necessaril
...
+    };
+    dictionary AuthenticationExtensionsAlgPolicyInputsJSON {
+        required sequence<sequence<COSEAlgorithmIdentifier>> createAlgs;
+        AuthenticationExtensionsClientInputsJSON createExtensions;
+        sequence<Base64URLString> deleteCredentials;
+    };
+    </xmp>
+
+    <div dfn-type="dict-member" dfn-for="AuthenticationExtensionsAlgPolicyInputs">
+        :   <dfn>createAlgs</dfn>
```
nsatragno, line 7848: We've discussed this at the FIDO plenary.

This solution is very complicated, and reduces to shipping our own custom, generic hybrid scheme. I don't think we should be doing this.

Instead, make it so that `createAlgs` takes a single list of algorithms, ordered from most preferred to least preferred. If we believe we need hybrid signatures, we can represent them with a single COSE type that encapsulates algorithms. Once hybrid schemes start solidifying, we'll hopefully get libraries that work well with COSE keys, and from the perspective of a simple RP upgrading to PQC could be as simple as adding new numbers to their supported algorithms list.

A single list also avoids the footgun of having nonsensical groups, like two groups of classical algorithms.

index.bs
```diff
@@ -7760,6 +7806,229 @@ However, [=authenticators=] that do not utilize [[!FIDO-CTAP]] do not necessaril
...
+                     ; Omitted when nothing was created.
+      ? 3: [* bstr], ; deletedCredentials: the credential IDs actually deleted in this ceremony.
+                     ; Omitted when nothing was deleted.
+    }
+    </pre>
+
+    where each <var>credHash</var> is:
+
+    <pre>
+    credHash = SHA-256( credIdLen || credId || COSE_Key )
```
nsatragno: I believe this needs a prefix for domain separation.

Also, is credIdLen big endian or little endian? Instead of rolling our own encoding, why not use CBOR?

WEBAUTHN_ALG_POLICY_HASH + CBOR({
  id: [binary string credential ID],
  key: cose key object
})

nsatragno: I think the reason this exists, as opposed to putting the created credentials on the signed extension outputs, is to avoid having security keys sign a payload that is too large? Is this really a problem? CC @ve7jtb

index.bs
```diff
@@ -7760,6 +7806,229 @@ However, [=authenticators=] that do not utilize [[!FIDO-CTAP]] do not necessaril
...
+:: <xmp class="idl">
+    partial dictionary AuthenticationExtensionsClientOutputs {
+        AuthenticationExtensionsAlgPolicyOutputs algPolicy;
+    };
+    partial dictionary AuthenticationExtensionsClientOutputsJSON {
+        AuthenticationExtensionsAlgPolicyOutputs algPolicy;
+    };
+
+    dictionary AuthenticationExtensionsAlgPolicyOutputs {
+        sequence<AuthenticatorAttestationResponseJSON> createdCredentials;
```
nsatragno: Why is this a JSON type instead of a normal type?

index.bs
```diff
@@ -915,6 +915,11 @@ BCP 14 [[!RFC2119]] [[!RFC8174]] when, and only when, they appear in all capital
     then each compartment is considered a separate [=authenticator=]
     with a single user with no access to other users' [=credentials=].
 
+:<dfn>Multiple Credentials Per User Account</dfn>
```
nsatragno, line 919: Suppose a user has a credential of alg A. Later, using this extension, the RP upgrades the user to alg B. However, the result of this upgrade didn't make it to the relying party's server.

Now, the user's authenticator has A and B. The next request from the relying party contains an empty allow list and an acceptedAlgs list of [B, A]. The relying party doesn't know the identity of the user. I believe in your proposal only the credential B would be returned, but the relying party wouldn't have the public key for B, so this wouldn't work.

If we only return a single credential assertion instead of one for all the matching credentials, I believe we are not addressing the requirement to avoid lockout for empty allow list cases.

nsatragno (changes requested):
Thank you for updating the explainer.

At this point, my feedback is based primarily on reading the explainer and is more on the general design rather than specific details.

index.bs
akshayku: It is true that user cannot log-in one time with assertion of cred B in this case because B's response from previous assertion did not reach the RP. 

This is exactly the case of orphan credential that this proposal solves. 

In this assertion, RP notices from existingCredentials list that user has credential A that it recognizes. Hence, RP sends a new getAssertion request immediately with A as allowlist and B as delete credential ID. And in that ceremony, old credential with alg B gets deleted and new credential gets created with alg B. 

B response not reaching server is a rare instance and RP can recover and delete orphaned credential reliably. And user gets logged in and is not locked out.

index.bs
akshayku: This is spelled out in the algorithm steps section. 

Authenticator in that case can decide which one they would like to use based on their policy. I am recommending last created or used one.

index.bs
akshayku: You meant **_unsigned_** extension outputs for the newly created credential? Looks like I missed that.

index.bs
akshayku: I would love to understand how signal API can work with security keys **_reliably_**. I couldn't figure that out. 

Signal API does not work today with security key because of few reasons.
- Security Key cannot be guaranteed to be attached with the OS after the user has used it. Especially with NFC and BLE transports. 
- Persistent PUAT is not guaranteed because user will not pair their security key with every PC/Kiosk they visit. 
- Delete requires user gesture in general case and cannot happen silently. 
- RP cannot target a particular authenticator as that is a signal API is separate call all together with no tie to the older ceremony. 

So, without an allowlist that targets that particular authenticator specifically, we cannot reliably know which authenticator user will choose and how they can delete the credential that RP wants to delete. 

Hence, RP can decide whether it can use signal API or this API to delete credentials as explained in the explainer. 

Separating it out in its own extension does not work because extensions in WebAuthn are independent  of each other and we do not have a order of extensions to be followed. During orphan deletion and creating new credentials in the same ceremony, credential needs to be deleted first and then new credential needs to be created in the same ceremony. That we cannot do via separate extensions.

index.bs
akshayku: With all due respect, I am afraid I disagree with only allowing every RP to adopt composite-hybrid algorithm pattern and not giving an option to the RPs who want separate credentials for PQC and Classical algorithms. That is a requirement for certain RPs. 

We are a web platform and we let RPs choose what algorithm they want to use and whether want a separate credential for fallback or not. 

I have heard different view from different RPs. WebAuthn Spec, as an API provider, should not be dictating whether they should use pure PQC + seperate classical algorithm or a hybrid-composite algorithm. Our view as an RP should not conflict with what WebAuthn API should provide.  

This is similar to what platforms core crypto stack does to the callers. We at the platform level do not say to the callers that you cannot create two separate keys for your scenario.

Single list does not work with the requirement of having two separate keys. For example, if there is a single list of PQC credentials first and classical next and user starts with a fresh authenticator, user will create a PQC credential and never a classical algorithm because it is always behind PQC in the preference list. 

Regarding solution being complicated, I am speaking with an experience of a prototype with security keys. It is not small, but also not complicated. And as with everything that we done over the years, it is worth supporting RPs with this transition and making is deployable. Currently, our solution is not. 

Regarding having non-sensical groups, we as a platform gives the guidance. And then it is up to the RP on what and how they want to deploy things. It is same as if they pick only classical algorithm in a PQC era. That would not be our recommendation, but we are not responsible for all their choices.

index.bs
akshayku: credIDlength is in big endian. 

We can do your structure if that what we all want. 

I just picked the bytes we have in [attested credential data](https://w3c.github.io/webauthn/#attested-credential-data) after AAGUID.

akshayku: The main reason is that the unsigned extension part of authenticatorMakeCredential response should not get signed over by asserting credential. PRF for example is in unsigned extension output and that should not be part of signed assertion and going back to the RP if that is being used on the client side.

index.bs
akshayku: Right. AuthenticationExtensionsClientOutputsJSON should contain the AuthenticatorAttestationResponseJSON responses and AuthenticationExtensionsClientOutputs should contain   the attestation objects?

index.bs
akshayku: Secondary reason is that if we sign the whole authenticatorMakeCredential response, the attestation gets signed over and cannot be removed if user don't want to not provide this information.

akshayku: Don't understand the need for domain separation. Can you elaborate more?

index.bs
nsatragno: Yes, unsigned.

index.bs
nsatragno: I think that's what you want, yes.

index.bs
nsatragno: Okay, I think that makes sense. It requires two assertion ceremonies to fix it, and that the user selects the same authenticator both times.

Have you considered returning all matching assertions instead? That way the relying party wouldn't need any special logic for this case. And, it would make it so the user doesn't have to fix themself by selecting the same authenticator. I agree this is rare but at the scale we're operating with WebAuthn, saving all relying parties from having to address a sharp edge case seems like a good thing. There are certainly some tradeoffs with this approach, too.

akshayku: I had considered it shortly couple of weeks back but not thoroughly. 

The main issue I see are following
- Resource constraint devices are already having hard time signing big PQC credentials especially over NFC and BLE transport with signature being too big. Asking them to send all the signatures looks too taxing for them. 
- Normally, all the extra fallback signatures are a waste for the RP as that is not their preference in the first place.

index.bs
nsatragno: > The main reason is that the unsigned extension part of authenticatorMakeCredential response should not get signed over by asserting credential. PRF for example is in unsigned extension output and that should not be part of signed assertion and going back to the RP if that is being used on the client side.

We don't have to sign over the unsigned extension outputs. If we had the existing credential sign over the authenticator data that corresponds to the new credential, then you'd only get the signed extension outputs signed by the existing credential.

The authenticator operations could be something along the lines of:
* Run a make credential operation like usual.
* Instead of returning the authenticator data, store it for later.
* Start running a get assertion operation with the existing credential.
* When creating the extensions outputs map, append the stored authenticator data from the make credential.
* Finish the get assertion operation.

> Secondary reason is that if we sign the whole authenticatorMakeCredential response, the attestation gets signed over and cannot be removed if user don't want to not provide this information.

Does your proposal even have a way to request attestation for the new credentials? Why would we do that when the previous credential, which should already be trusted, is signing over the new credentials?

nsatragno: It's to avoid mixing this hash with another in the future that we might want to construct for some other feature with the same shape. It's possible we don't need this since it gets signed as part of the extensions block :thinking: , but it is free.

> I just picked the bytes we have in [attested credential data](https://w3c.github.io/webauthn/#attested-credential-data) after AAGUID.

Ah this makes sense, let's just spell that out.

akshayku: Proposal is making "none" attestation as a requirement, RP cannot request the attestation and platform is not going to ask for it. 

Question then becomes is that whether authData is the only field that is present here? 

if attestation object field from authenticatorMakeCredential response is also present if authenticator returns the attestation anyway, should platform break the operation and not let user sign-in. Because authenticatorMakeCredential attestation field is also signed, then platform can remove it because of signature mismatch.

Also, signing authData also invoves signing over backup eligibilty/state bits that if put under signed extension cannot be changed by the client. Which is currently the behavior for some clients. Only option then the client has is to break the user signin. 

Splitting the CTAP authenticatorMakeCredential response in two places (authData in signed extension and rest of the fields like largeBlob key etc in unsigned extension) didn't look right. 

In current design, authenticator will return whole authenticatorMakeCredential response as part of unsigned extension and platform can strip out attestation if that ever becomes a problem without breaking user sign-in. Or may be we will allow attestations in the future. That gives us the flexibility in the future without need to change the spec.

And adding SHA-256 bytes only costed 32 bytes more.

akshayku: Also, AAGUID is also inside authData which user may not want to send to the RP all the time.

akshayku: Added the attested credential data parts detail.

index.bs
akshayku: Actually changed to sequence<PublicKeyCredential> as we also want unsigned extensions outputs.

index.bs
akshayku: Changed to PublicKeyCredential output which will contain the extension outputs as well as credential in AuthenticationExtensionsAlgPolicyOutputs section of the output

index.bs
akshayku: Some more reasons for allowing RP to create more than one algorithm. 

- We should not look at this proposal as only from classical to PQC migration angle. We should look at this proposal from migrating the user from Algorithm A to Algorithm B irrespective of what those algorithms are. For example, RP may want to move the user from ML-DSA PQC algorithms to FALCON PQC algorithms in the future. During this migration, they may not be any composite algorithms between these PQC algorithms. RP should be able to evaluate the performance and security of each algorithm independently. 

- Not all security keys/TPMs will be capable of creating composite ML-DSA algorithms due to resource constraint but may be capable of creating pure ML-DSA algorithm. In that case, even if RP wants composite algorithms, authenticators may be only support pure ML-DSA + Classical scheme which may be OK for such RPs instead of classical algorithm. 

- Regulations are coming to upgrade from ECC P256 to ECC P384 at the minimum for classical algorithms wherever used in addition to moving to PQC. So there is a need to upgrade even existing classical algorithm within same class of algorithms.

- @harshlal028 has opened an issue Key Rotation feature in FIDO2 spec where RP may want to rotate the key with same algorithm. For example, rotate your ES256 key. For that to happen **_safely_**, you need two algorithms (one you authenticate with and one you rotate). And RP can use same orphan delete mechanism that we have in this proposal to delete existing ES256 key and create a new ES256 key. 

There are many reason RP wants multiple algorithm credentials going forward. And it is a requirement for many RPs. Allowing only single preference list is not enough for these RPs. 

I would request you to allow RPs to create multiple algorithm credential.

harshlal028:
Generally speaking providing support for migrating from Algorithm A to Algorithm B for an RP is a good idea and help them build crypto-agile systems. This can be relevant beyond PQC migrations.

akshayku:
Changes in V3 version

- `createAlgs` : groups -> a single preference-ordered list
- Binding simplified
  - Removed  `credHash` / `mintHashes` . 
  - The created credential's authenticatorData is carried in the signed - extension output. 
  - No hash comparison needed by the RP.
- algPolicy Client Extension Output now carries created credential's unsigned extension outputs only instead of whole publicKeyCredential as per Nina's proposal.
- `preferredAlgs` instead of `acceptedAlgs`
  - This allows for corner cases where user can get stuck due to misconfiguration on the RP side and provide a consistent experience for authenticators who supports multiple algorithms vs one that doesn't. 
  - `preferredAlgs` is not a hard filter and is just a preference. 
- Fallback is now an explicit follow-up ceremony
  - New capability outputs. Since fallback is now an explicit ceremony, RP need more details on whether it will succeed or not.
    - Supported COSE Algorithms
    - `maxCredentialsPerAccount` in case storage is scarce as for security keys.

akshayku:
@nsatragno Thanks for your alternative design. Although I don't agree fully with all your proposals, it gives me more insight on how you are thinking. 

I've pulled few ideas from that and incorporated in my proposal.  (sign the new authData fully, don't put full publicKeyCredential in client extension output).

Also, regarding the feedback on complexity of multiple groups, while I don't fully agree on implementation complexity, I am dropping that from my proposal and now we have one single algorithm list in the request. Hopefully that simplifies the implementation complexity. 

But there are few things that I disagree with your proposal. 

- General algorithm agility, not just classical→PQC. 
  - The same mechanism should covers ML-DSA→FALCON, regulatory intra-class upgrades (e.g. P-256→P-384), and safe same-algorithm key rotation (ask from Google itself).
  - So this proposal has to work for migrating users from Algorithm A to Algorithm B where there is no hybrid algorithm for Algorithm A and Algorithm B. 

- Creation of Multiple Keys is a hard requirement for some RPs. 
  - In classical fallback is needed for successful PQC migration in case of heterogeneous authenticator landscape. Even if we convince RPs to move to hybrid/composite algorithm, there will be authenticators who cannot do hybrid/composite algorithm due to resource constraints. We have to meet the physical reality for such authenticators. Some security key vendors have already told me that they will not be able to do composite algorithm that an RP needs. 
  - I have moved the new algorithm creation to a separate step to reduce the implementation complexity for platforms. But multiple credential creation is still needed.

- Things should work in usernameless flows with multiple algorithm. 
  - When there are multiple algorithms present on the authenticator, RP must be able to say which algorithm they want user to sign in with and that must work in usernameless flows. 
  - Hence, preferredAlgs is a MUST in that case for an RP.

- RP must be able to reliably cleanup orphaned credential. 
  - Your proposal to include signalAllAcceptedCredentials in navigator.credentials.get() was interesting and I am glad that you also want to reliably remove orphaned credentials from security keys. 
  - But few things are not clicking to me. 
    - Stateless server problem
      - There is no clear confirmation from the authenticator that a credential got deleted in your proposal. RP may inference that from comparing existing credentials list from before vs after the flows. Which is keeping more state on the server side which are usually stateless. 
    - As user verification is happening, I see no reason the response should not contain those deleted credential ids. 
    - The extensions are always independent of each other. Your proposal is binding certain behavior from the platform where it is capable of deleting the credential before actual processing of the webauthn operation. That creates non-atomic operations which are troublesome. 
    - There are authenticators which do not allow credential management APIs and do not provide that flexibility for the platform to delete credentials. Example, some BLE based authenticators on phones and other form factors with their own UI. 
    - Hybrid flow will also probably won't work. Atleast at the CTAP messaging layer. 
    - RP wants to delete one credential. however, your proposal is asking RP to include all the other credentials from also different authenticators for platform to evaluate. That is strange design choice as well as increases the amount of back and forth with security keys where not all transports are fast (NFC/BLE) etc
    - Hence, I see no real benefit of your proposal compared to my proposal of putting exactly the credential you want to delete as part of the extension itself.

nsatragno:
I'll respond to the feedback in a way I hope brings us closer to a solution we can all agree on.

First, a few factual comments:

> safe same-algorithm key rotation

Neither proposal supports same-algorithm key rotation in a way that is better than what you can already do with WebAuthn, because we say that the { user id, rp id, algorithm } tuple must be unique. Replacing the same { user id, rp id, algorithm } tuple runs the risk of data loss.

I agree that if this is something we want, we should address the requirement on the same set of changes.

> There is no clear confirmation from the authenticator that a credential got deleted in your proposal.

What requirement does letting the RP know a credential was deleted serve? The RP can always remove the public key from their credential record.

With your proposal, the server cannot clean up the credential record until the authenticator has confirmed that it was removed. This means relying parties need to have two different flows for removing credentials from passkey providers using the signal API (where they first remove it on their database, then send an update list) and from security keys (where they first send a delete command, then clean it up). In fact, the signal API is as stateless as you can get already!

> The extensions are always independent of each other. 

This has never been true, e.g. if you set large blob = required, that might prevent you from getting a PRF because your authenticator cannot satisfy large blob.

> That creates non-atomic operations which are troublesome.

We already have plenty of non-atomic operations that are fine, e.g. large blob, batching credential IDs for large allow lists. Is there a particular concern you have here?

> RP wants to delete one credential. however, your proposal is asking RP to include all the other credentials from also different authenticators for platform to evaluate. That is strange design choice as well as increases the amount of back and forth with security keys where not all transports are fast (NFC/BLE) etc

We already send the same list for allow-list requests.

> Hybrid flow will also probably won't work. Atleast at the CTAP messaging layer.

Why not? We send the extensions as JSON to support these sort of use cases.

> Hence, I see no real benefit of your proposal compared to my proposal of putting exactly the credential you want to delete as part of the extension itself.

The benefits are that we don't need to introduce a new API that has new semantics. It's better to reuse something we already have.

---

And a discussion of trade-offs:

> Creation of Multiple Keys is a hard requirement for some RPs.
> Things should work in usernameless flows with multiple algorithm.
> General algorithm agility, not just classical→PQC.

I believe these three points can be summarized as my proposal offers no way for an RP to select two algorithms A and B that aren't represented by an existing hybrid pair and switch between one that used to be the preferred one (say A) to B without having to trust A at the point you make that switch. I agree this is something my proposal doesn't do.

However, as far as I know, no relying party has ever asked for this capability. The WebAuthn WG works in the open and accepts feedback from relying parties. Actionable, precise, and public feedback from a diverse set of relying parties would help justify this increase in complexity for relying parties, user agents, and authenticators.

akshayku:
> I'll respond to the feedback in a way I hope brings us closer to a solution we can all agree on.
> 
> First, a few factual comments:
> 
> > safe same-algorithm key rotation
> 
> Neither proposal supports same-algorithm key rotation in a way that is better than what you can already do with WebAuthn, because we say that the { user id, rp id, algorithm } tuple must be unique. Replacing the same { user id, rp id, algorithm } tuple runs the risk of data loss.

[Akshay] No, it doesn't with my proposal. When you have another credential you can treat the credential to be rotated similar to orphan credential and rotate it by deleting it and recreating it. 

> 
> I agree that if this is something we want, we should address the requirement on the same set of changes.
> 
> > There is no clear confirmation from the authenticator that a credential got deleted in your proposal.
> 
> What requirement does letting the RP know a credential was deleted serve? The RP can always remove the public key from their credential record.
> 
> With your proposal, the server cannot clean up the credential record until the authenticator has confirmed that it was removed. This means relying parties need to have two different flows for removing credentials from passkey providers using the signal API (where they first remove it on their database, then send an update list) and from security keys (where they first send a delete command, then clean it up). In fact, the signal API is as stateless as you can get already!

[Akshay] Signal API does not work with security keys because of reasons mentioned above. Especially when the key does not do credential management like some security keys with built in UI. Also, when user verification is done, I see no harm reporting it back. Giving authenticator signed signal is always better than not having one for the RP if we can. 

> 
> > The extensions are always independent of each other.
> 
> This has never been true, e.g. if you set large blob = required, that might prevent you from getting a PRF because your authenticator cannot satisfy large blob.

[Akshay] When you set large blob = required and authenticator can't do large blob, the whole operation gets cancelled and there is no passkey. That doesn't mean that PRF processing is linked with large blob processing. This is totally different than what you are saying.  

> 
> > That creates non-atomic operations which are troublesome.
> 
> We already have plenty of non-atomic operations that are fine, e.g. large blob, batching credential IDs for large allow lists. Is there a particular concern you have here?

[Akshay] Once a final CTAP request is sent to the authenticator, then it is an atomic operation. With no equivalent deletion option in the CTAP request, it is not an atomic operation from the protocol perspective and from the perspective of authenticators who don't support credential management feature. Having said that, using credential management for USB by the platform in that sense is not atomic and can result in in-between failures. Something to think about. 

> 
> > RP wants to delete one credential. however, your proposal is asking RP to include all the other credentials from also different authenticators for platform to evaluate. That is strange design choice as well as increases the amount of back and forth with security keys where not all transports are fast (NFC/BLE) etc
> 
> We already send the same list for allow-list requests.

[Akshay] Not necessarily in the case of deletion scenario of a particular credential. If they sent all the credentials, then there is no guarantee that platform will go to that authenticator itself. Hence, recommendation for the RP is to only send the credential in allowlist that you recognize from the existingCredential array. That will target that particular authenticator and then credential can be reliably deleted. 

> 
> > Hybrid flow will also probably won't work. Atleast at the CTAP messaging layer.
> 
> Why not? We send the extensions as JSON to support these sort of use cases.
> 

[Akshay] Yeah, Probably true over hybrid with JSON messages. It doesn't work with security keys though which do not have credential management. 

> > Hence, I see no real benefit of your proposal compared to my proposal of putting exactly the credential you want to delete as part of the extension itself.
> 
> The benefits are that we don't need to introduce a new API that has new semantics. It's better to reuse something we already have.
> 

[Akshay] We already are discussing some form of new extension. So, that's atleast a new thing we have to support anyway. Rest is a parameter for preferredAlgs, that is needed when authenticators will land up with multiple credentials. Which credential to use is not a platform choice. That is a RP choice. So, that is needed. 

> And a discussion of trade-offs:
> 
> > Creation of Multiple Keys is a hard requirement for some RPs.
> > Things should work in usernameless flows with multiple algorithm.
> > General algorithm agility, not just classical→PQC.
> 
> I believe these three points can be summarized as my proposal offers no way for an RP to select two algorithms A and B that aren't represented by an existing hybrid pair and switch between one that used to be the preferred one (say A) to B without having to trust A at the point you make that switch. I agree this is something my proposal doesn't do.
> 
> However, as far as I know, no relying party has ever asked for this capability. The WebAuthn WG works in the open and accepts feedback from relying parties. Actionable, precise, and public feedback from a diverse set of relying parties would help justify this increase in complexity for relying parties, user agents, and authenticators.

[Akshay] Alright, Microsoft needs it as an RP. And please don't reject other RPs requirements just because you don't believe in that kind of deployment. I need to choose which algorithm as an RP I should choose for authenticator when there are multiple and it needs to work in usernameless flow.

akshayku:
@nsatragno / @kreichgauer 

One more point in favor of preferredAlgs for your consideration

- Platform/Browser has to deal with multi-algorithm credentials for usernameless (no allowlist) flows invariably. 
  - There will be more than one algorithm credential created invariably in both your approach and my approach. 
    - For example, user has a PQC capable authenticator but RP initially as it stands today supports classical algorithm only.  That will result in let's say ES256 credential. 
    - When RP finally deploys their PQC support if either of our proposal or via explicit MakeCredential (without exclude list), it will create PQC credential. 
  - So, when there is no allowlist in a usernameless flows, platform/browser has to decide which algorithm credential to pick. 
  - Platform could do try to do a default preferred list, but they cannot know what RP's side of situation is. Hence, if platform anyway has to deal with multi algorithm credential in a usernameless no-allowlist flow, it is always better to be explicit about it and let RP choose. 
    - Any default platform order will create problems. 
      - If you put pure PQC algorithm ahead of Composite algorithm, then it will create problems for authenticators which supports pure PQC first and then composite PQC (composite PQC is still sometime away, resource constraints etc). 
      - If you put Composite algorithm ahead of pure PQC, then it will create problem when RP eventually wants to move from composite to pure PQC in some years time.

akshayku:
@nsatragno / @kreichgauer

If we go with your proposal, can you also think about how the migration from composite to pure PQC will looks like in few years and how does RP indicate that in usernameless flows?

nsatragno:
I promise to flesh out the proposal further. I've been really busy with *points at everything*.

ve7jtb:
In CTAP for getAssertion in the no allowlist flows the authenticator is producing signatures for all the credentials that match RPID and user.id.  That is done by the platform sending multiple authenticatorGetNextAssertion (0x08) commands.   That is not a problem for the authenticator.   The webAuthn level can filter based on alg or whatever policy is desired, or return multiple at the webAuthn level if we don't do the filtering at the CTAP level.

That will probably be a problem for the current Hybrid/PXP using CTAP. 

Perhaps just adding credential alg preference to CTAP for use over PXP is an option if we are not doing native webAuthn?

ve7jtb:
Actually, thinking about NFC it may be useful to try and have the authenticator filter.   There could be several user.id and multiple algs per user.id.  While the individual transactions my be reasonable for the authenticator, the total transactions could grow to quite large with all those credentials especilly if extensions are tacked on to each one.    Just the ML-DSA 87 signature is 4,627 bytes; stacking up several of those may take too long to transfer for people to hold an authenticator steady.  Taking the authenticator out of the field will require a new key agreement and restarting from scratch.

akshayku:
@ve7jtb , This is what I was talking about when I said that authenticator has _some_ concept of latest credential created for an RP. 

image

See section 6.6.6.12.2.1 in https://fidoalliance.org/specs/fido-v2.3-ps-20260226/fido-client-to-authenticator-protocol-v2.3-ps-20260226.html#sctn-getAssert-platf-actions

PR #2442, opened 2026-07-08
Upgrade to Bikeshed 7

MasterKale:
About a month ago `bikeshed==5.4.2`, which we've been currently stuck on, stopped being able to build the spec. After digging into this, I found out it's because a recent change to https://github.com/speced/bikeshed-data on June 9th [significantly changed the structure of metadata](https://github.com/speced/bikeshed-data/commit/9bca02f0da0074c26035edb16b2fc8d44111f973) that Bikeshed relied on, so Bikeshed 5 didn't know how to find any of the .include files it needs to build.

This seemed like as good a time as any to try and unblock our migration to the latest Bikeshed. This PR contains changes that finally allow us to use Bikeshed 7 (and hopefully beyond.)

BTW, Bikeshed told me I could refer to CTAP 2.1 or 2.2 for "ctap2 canonical cbor encoding form" links. I picked 2.1 because I figured CBOR encoding hasn't really changed all that much between 2.1 and 2.2. **Maybe we refer to CTAP 2.2 instead?**

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/pull/2442.html" title="Last updated on Jul 9, 2026, 1:01 PM UTC (89661cd)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/2442/207cf99...89661cd.html" title="Last updated on Jul 9, 2026, 1:01 PM UTC (89661cd)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2058,19 +2058,19 @@ a numbered step. If outdented, it (today) is rendered as a bullet in the midst o
 
                     1. Otherwise, [=list/Append=] |C| to |excludeCredentialDescriptorList|.
 
-                1. <span id="CreateCred-InvokeAuthnrMakeCred">
-                    <!-- @@EDITOR-ANCHOR-01: KEEP THIS LIST SYNC'D WITH OTHER LOCATIONS WITH THIS TAG -->
-                    Invoke the [=authenticatorMakeCredential=] operation on |authenticator| with
-                        |clientDataHash|,
-                        <code>|pkOptions|.{{PublicKeyCredentialCreationOptions/rp}}</code>,
-                        <code>|pkOptions|.{{PublicKeyCredentialCreationOptions/user}}</code>,
-                        |requireResidentKey|,
-                        |userVerification|,
-                        |credTypesAndPubKeyAlgs|,
-                        |excludeCredentialDescriptorList|,
-                        |enterpriseAttestationPossible|,
-                        |attestationFormats|,
-                        and |authenticatorExtensions| as parameters.</span>
+                    1. <span id="CreateCred-InvokeAuthnrMakeCred">
+                        <!-- @@EDITOR-ANCHOR-01: KEEP THIS LIST SYNC'D WITH OTHER LOCATIONS WITH THIS TAG -->
+                        Invoke the [=authenticatorMakeCredential=] operation on |authenticator| with
+                            |clientDataHash|,
+                            <code>|pkOptions|.{{PublicKeyCredentialCreationOptions/rp}}</code>,
+                            <code>|pkOptions|.{{PublicKeyCredentialCreationOptions/user}}</code>,
+                            |requireResidentKey|,
+                            |userVerification|,
+                            |credTypesAndPubKeyAlgs|,
+                            |excludeCredentialDescriptorList|,
+                            |enterpriseAttestationPossible|,
+                            |attestationFormats|,
+                            and |authenticatorExtensions| as parameters.</span>
 
                 1. [=set/Append=] |authenticator| to |issuedRequests|.
 
--- a/index.bs
+++ b/index.bs
@@ -2059,6 +2059,7 @@ a numbered step. If outdented, it (today) is rendered as a bullet in the midst o
                     1. Otherwise, [=list/Append=] |C| to |excludeCredentialDescriptorList|.
 
                 1. <span id="CreateCred-InvokeAuthnrMakeCred">
+                    <!-- @@EDITOR-ANCHOR-01: KEEP THIS LIST SYNC'D WITH OTHER LOCATIONS WITH THIS TAG -->
                     Invoke the [=authenticatorMakeCredential=] operation on |authenticator| with
                         |clientDataHash|,
                         <code>|pkOptions|.{{PublicKeyCredentialCreationOptions/rp}}</code>,
--- a/index.bs
+++ b/index.bs
@@ -258,6 +258,7 @@ spec:url; type:dfn; for:url; text:host
 spec:url; type:dfn; text:valid domain;
 spec:webidl; type:dfn; text:DOMString
 spec:webidl; type:interface; text:Promise
+spec:fido-client-to-authenticator-protocol-v2.1; type:dfn; text:ctap2 canonical cbor encoding form
 </pre>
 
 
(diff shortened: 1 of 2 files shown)


akshayku (approved):

akshayku:
> BTW, Bikeshed told me I could refer to CTAP 2.1 or 2.2 for "ctap2 canonical cbor encoding form" links. I picked 2.1 because I figured CBOR encoding hasn't really changed all that much between 2.1 and 2.2. Maybe we refer to CTAP 2.2 instead?

CBOR canonical rules haven't changed. Both are fine.

index.bs
```diff
@@ -2057,23 +2058,19 @@ a numbered step. If outdented, it (today) is rendered as a bullet in the midst o
...
+                        |clientDataHash|,
+                        <code>|pkOptions|.{{PublicKeyCredentialCreationOptions/rp}}</code>,
+                        <code>|pkOptions|.{{PublicKeyCredentialCreationOptions/user}}</code>,
+                        |requireResidentKey|,
+                        |userVerification|,
+                        |credTypesAndPubKeyAlgs|,
+                        |excludeCredentialDescriptorList|,
+                        |enterpriseAttestationPossible|,
+                        |attestationFormats|,
+                        and |authenticatorExtensions| as parameters.</span>
```
emlun: Hm. In [L3 CR](https://www.w3.org/TR/2026/CR-webauthn-3-20260113/#CreateCred-InvokeAuthnrMakeCred), this is step 3 under "For each credential descriptor C in pkOptions.excludeCredentials:", but this now instead renders as step 10 after that loop instead of inside the loop body.

Looking further into that, this step was [outside the for loop in L1](https://www.w3.org/TR/2019/REC-webauthn-1-20190304/#ref-for-authenticatormakecredential%E2%91%A3). The first draft that has this step inside the for loop is [L2-WD3](https://www.w3.org/TR/2020/WD-webauthn-2-20200730/#CreateCred-InvokeAuthnrMakeCred), so that change was probably a mistake. Commit e239ee557e71232c034ff8e034582f1e231cf31f also describes the same issue and tries to fix it, but apparently the issue has re-appeared somewhere along the line.

I assume the error this is meant to fix is this?

```
LINE 2064:21: Tried to auto-close a <li>, but there were unclosed elements remaining before the nearest matching start tag (at 1920:5).
Open tags: <div> at 1698:1, <li> at 1920:5, <dl> at 1923:9
```

As best as I can tell, that error is caused by Bikeshed trying to auto-close the `<li>` on line 1920... presumably because a `<li>` cannot be a direct child of another `<li>`? So it seems like whatever preprocessing this is happens before Bikeshed injects a parent `<ol>` for this Markdown list.

Reverting this change and simply inserting an `<ol>` _almost_ works, but not quite:

```diff
diff --git a/index.bs b/index.bs
index 12aa29f..2597f29 100644
--- a/index.bs
+++ b/index.bs
@@ -1937,6 +1937,7 @@ a numbered step. If outdented, it (today) is rendered as a bullet in the midst o
             :   If an |authenticator| becomes available on this [=client device=],
             ::  Note:  This includes the case where an |authenticator| was available upon |lifetimeTimer| initiation.
 
+                <ol>
                 1. This |authenticator| is now the <dfn for="create">candidate authenticator</dfn>.
 
                 1. If <code>|pkOptions|.{{PublicKeyCredentialCreationOptions/authenticatorSelection}}</code> is present:
@@ -2073,6 +2074,7 @@ a numbered step. If outdented, it (today) is rendered as a bullet in the midst o
                         and |authenticatorExtensions| as parameters.</span>
 
                 1. [=set/Append=] |authenticator| to |issuedRequests|.
+                </ol>
 
             :   If an |authenticator| ceases to be available on this [=client device=],
             ::  [=set/Remove=] |authenticator| from |issuedRequests|.
```

```
FATAL ERROR: Line 2079 isn't indented enough (needs 1 indent) to be valid Markdown:
"</ol>"
```

This error is a bit confusing, because line 2079 is not the `</ol>` line but an empty line... But anyway, indenting everything within this `<ol>` by 4 spaces now resolves the error! It now also correctly renders "Invoke the authenticatorMakeCredential" as step 10 after the "For each".

But... the "Append authenticator to issuedRequests" step now renders as top-level step 23 instead of as step 11 of this switch case. I can't find any indentation level that leaves it on the same level as the `<li id='CreateCred-InvokeAuthnrMakeCred'>`, it skips directly from top-level step 23 to a sub-step of the `<li>`. It really seems like the `<li>` is breaking Bikeshed's indentation tracking somehow.

So yeah, changing `<li id=` to `<span id=` is probably the way to go here, but we may need to note that this is now a substantive change to the procedure since it changes the control flow nesting level. It might be best to keep this indented 4 more spaces as before, so that it stays step 3 of the "For each" and so this PR can be purely editorial, and fix the incorrect indentation in a separate PR?

emlun (commented):
Thanks a lot for taking care of this long-overdue issue 😅 I noticed a subtle substantive change in the steps indentation, though, which I think might be best to address in a separate PR.

index.bs
emlun: (It's a shame to lose the paragraph link you get with `<li id=`, but losing that is far better than the other possible trade-offs discussed above...)

simoneonofri: This is why we had https://github.com/w3c/webauthn/pull/2427

emlun: Yeah, but even Bikeshed 5.4.2 is now failing to build the spec because of the change to the Bikeshed metadata. :disappointed:

MasterKale: Huh, I took these before/after screenshots and the only thing I could tell had changed was we lost the paragraph anchor:

### Before:

https://w3c.github.io/webauthn/#CreateCred-InvokeAuthnrMakeCred

Screenshot 2026-07-08 at 3 03 21 PM

### After:

localhost, with these chnages

Screenshot 2026-07-08 at 3 03 14 PM

### L3 CR

> Hm. In [L3 CR](https://www.w3.org/TR/2026/CR-webauthn-3-20260113/#CreateCred-InvokeAuthnrMakeCred), this is step 3 under "For each credential descriptor C in pkOptions.excludeCredentials:", but this now instead renders as step 10 after that loop instead of inside the loop body.

Oh no, you're right...

https://www.w3.org/TR/2026/CR-webauthn-3-20260113/#CreateCred-InvokeAuthnrMakeCred

Screenshot 2026-07-09 at 5 48 11 AM

MasterKale: I _think_ there's an alternative solution that I can try and put up a second PR with. I discovered that I can drop ~13 **.include** files into the root of the project and Bikeshed 5.4.2 would be able to build the spec because of its template lookup logic:

Screenshot 2026-07-09 at 5 52 17 AM

I only used empty files for these as I was experimenting, and we'd have to triple-check afterwards, but if we vendored in the latest versions of these files from bikeshed-data I think we might still be able to limp along on 5.

This is assuming we figure out why, as seen above, L3 CR and the ED are rendering that step differently.

MasterKale: Wait a minute, if I just indent that block in the .bs file the step becomes Step 3 as we want:

Screenshot 2026-07-09 at 5 58 45 AM

MasterKale: I just indented the step, does that address the problem? 🤔

emlun: Yeah, I think that's best for this PR so it can be purely editorial. Let's do the "undent" in a separate PR so that change is more visible - it is also an editorial fix "in spirit", but I think it warrants a few more eyes on it since _technically_ it does change the procedure definition (even if the current state happened by accident).

emlun (approved):
Thanks! 🎉

index.bs
emlun: Issue for fixing this: https://github.com/w3c/webauthn/issues/2444

PR #2445, opened 2026-07-10
Add virtual authenticator support for CMTG keys

### LINKED ISSUE #2395: Virtual Authenticator support for the Credential Manager Trust Group Key extentsion
> Would it be possible to add virtual authenticator support for this feature? It feels it should be pretty simple, and it'll be required to write WPTs. 

 _Originally posted by @nsatragno in [#2377](https://github.com/w3c/webauthn/pull/2377/changes#r2760628029)_

nsatragno:
Adds virtual authenticator support for Credential Manager Trust Group (CMTG) keys.

This change extends virtual authenticator credentials to support a list of CMTG keys and an index to the "active" key. The key index may be used to switch which key is returned in an assertion. You can also set a new flag, `generateCmtgKeyOnNextOperation`, to return a new credential. This should allow web developers to test different scenarios, and browser developers to write web platform tests. 

The new parameters are reflected in the add credential, get credentials, and set credential properties commands.

Finally, the change adds a new webdriver capability to indicate and request support for this feature.

Implementation commitment:

- [X] Chromium https://crbug.com/485888879

Closes #2395

***
<a href="https://pr-preview.s3.amazonaws.com/nsatragno/webauthn/pull/2445.html" title="Last updated on Aug 7, 2026, 4:23 PM UTC (69f9499)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webauthn/2445/0b64432...nsatragno:69f9499.html" title="Last updated on Aug 7, 2026, 4:23 PM UTC (69f9499)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -8174,22 +8174,28 @@ A <dfn>CMTG Key Record</dfn> is a [=struct=] with the following [=struct/items=]
 :: A CBOR byte string containing a signature generated with the Credential Manager Trust Group Key's private key.
 
 : Authenticator extension processing
-:: For both [=authenticatorMakeCredential=] and [=authenticatorGetAssertion=] operations:
-    1. Create or select the [=public key credential source=] as usual (see [[#sctn-op-make-cred]],
-        or [[#sctn-op-get-assertion]] as appropriate).
+:: The <dfn abstract-op>locate the existing Credential Manager Trust Group Key</dfn> operation
+    returns the [=Credential Manager Trust Group Key=] associated with the current
+    [=public key credential source=], or `null` if none exists.
 
-    1. If the [=public key credential source=] is not [=backup eligible=],
-        terminate these processing steps as this extension only applies to [=multi-device credentials=].
+    The <dfn abstract-op>create a Credential Manager Trust Group Key</dfn> operation
+    takes a public key algorithm identifier, generates a new
+    [=Credential Manager Trust Group Key=] for the current
+    [=public key credential source=], and returns the generated key.
 
-    1. If a [=Credential Manager Trust Group Key=] does not already exist for this ([=public key credential source/id|Credential ID=],
-        [=public key credential source/rpId|RP ID=], [=public key credential source/userHandle|User Handle=]) tuple in the [=authenticator=]/[=credential manager=],
-        create it using the same public key algorithm as that used by the [=user credential=]'s [=credential key pair=],
-        otherwise locate the existing [=Credential Manager Trust Group Key=].
+    For both [=authenticatorMakeCredential=] and [=authenticatorGetAssertion=] operations:
+        1. Create or select the [=public key credential source=] as usual (see [[#sctn-op-make-cred]],
+            or [[#sctn-op-get-assertion]] as appropriate).
 
-    1. Let |cmtgPublicKey| be the newly created or existing [=Credential Manager Trust Group Key=],
-        in COSE_Key format [=credentialPublicKey|in the same fashion as for the user credential's credentialPublicKey=].
+        1. If the [=public key credential source=] is not [=backup eligible=],
+            terminate these processing steps as this extension only applies to [=multi-device credentials=].
 
-    1. Let |cmtgPrivateKey| be the newly created or existing [=Credential Manager Trust Group private key=].
+        1. Let |cmtgKey| be the result of attempting to [$locate the existing Credential Manager Trust Group Key$].
+        1. If |cmtgKey| is null:
+            1. Let |cmtgKey| be the result of [$create a Credential Manager Trust Group Key|creating a Credential Manager Trust Group Key$] using the same public key algorithm as that used by the [=user credential=]'s [=credential key pair=].
+        1. Let |cmtgPublicKey| be |cmtgKey|'s public key,
+            in COSE_Key format [=credentialPublicKey|in the same fashion as for the user credential's credentialPublicKey=].
+        1. Let |cmtgPrivateKey| be |cmtgKey|'s private key.
 
     1. If this algorithm is being run as part of [=authenticatorMakeCredential=]:
 
@@ -8373,6 +8379,12 @@ Additionally, [=extension capabilities=] are defined for every [=authenticator e
                 <td>boolean</td>
                 <td>Indicates whether the [=endpoint node=] WebAuthn WebDriver implementation supports the `credBlob` extension defined in [[FIDO-CTAP]].</td>
             </tr>
+            <tr>
+                <td>Credential Manager Trust Group Key Extension Support</td>
+                <td>`"webauthn:extension:cmtgKey"`</td>
+                <td>boolean</td>
+                <td>Indicates whether the [=endpoint node=] WebAuthn WebDriver implementation supports the [=cmtgKey=] extension.</td>
+            </tr>
         </tbody>
     </table>
 </figure>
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -8016,35 +8016,41 @@ A <dfn>CMTG Key Record</dfn> is a [=struct=] with the following [=struct/items=]
...
-        create it using the same public key algorithm as that used by the [=user credential=]'s [=credential key pair=],
-        otherwise locate the existing [=Credential Manager Trust Group Key=].
+    For both [=authenticatorMakeCredential=] and [=authenticatorGetAssertion=] operations:
+        1. Create or select the [=public key credential source=] as usual (see [[#sctn-op-make-cred]],
+            or [[#sctn-op-get-assertion]] as appropriate).
 
-    1. Let |cmtgPublicKey| be the newly created or existing [=Credential Manager Trust Group Key=],
-        in COSE_Key format [=credentialPublicKey|in the same fashion as for the user credential's credentialPublicKey=].
+        1. If the [=public key credential source=] is not [=backup eligible=],
+            terminate these processing steps as this extension only applies to [=multi-device credentials=].
```
timcappalli, line 8191: Should we make these changes in a separate PR? These don't seem to be related to the virtual authenticator.

nsatragno: We could. These changes aren't supposed to be functional, they're there to make it easier to specify how the virtual authenticator retrieves and stores CMTGs.

index.bs
```diff
@@ -8016,35 +8016,41 @@ A <dfn>CMTG Key Record</dfn> is a [=struct=] with the following [=struct/items=]
...
-    1. Output |cmtgPublicKey| as the [=authenticator extension output=].
+        1. Output |cmtgPublicKey| as the [=authenticator extension output=].
 
-    1. Output |cmtgKeySig| as the extension's [=unsigned extension output=].
+        1. Output |cmtgKeySig| as the extension's [=unsigned extension output=].
 
-        Note: |cmtgKeySig| cannot be included in the [=authenticator extension output=] because it is returned inside the [=authenticator data=]
-         and that would imply that the signature signs over itself.
+            Note: |cmtgKeySig| cannot be included in the [=authenticator extension output=] because it is returned inside the [=authenticator data=]
+             and that would imply that the signature signs over itself.
```
timcappalli: was this to keep Bikeshed happy?

nsatragno: This change should have been reverted now.

index.bs
nsatragno: To take a stronger stance here: these changes make it so we can cleanly inject behaviour here from the virtual authenticator. I would prefer to keep them here, since the changes don't stand on their own (but I could go either way if you thought it was important to move them to a new PR).

index.bs
```diff
@@ -8699,6 +8738,32 @@ The <dfn>Credential Parameters</dfn> is a JSON [=Object=] passed to the [=remote
                 </td>
                 <td>string</td>
             </tr>
+            <tr>
+                <td>|cmtgKeys|</td>
+                <td>
+                    A list of [=Credential Manager Trust Group Private Keys=],
+                    each formatted as an asymmetric key package per [[RFC5958]] and encoded using [=Base64url Encoding=].
```
MasterKale, line 8745: Is there a reason to **not** use JWKs for this? In the context of web development JWKs are common enough that I'd suggest devs would be more familiar with their use here, and they're valid JSON to boot which helps with readability. Plus [as per RFC 7517](https://www.rfc-editor.org/info/rfc7517/#appendix-A.2) a JWK can contain information on an entire keypair, including public and private key parameters:

```
{
  "kty": "EC",
  "crv": "P-256",
  "x": "MKBCTNIcKUSDii11ySs3526iDZ8AiTo7Tu6KPAqv7D4",
  "y": "4Etl6SRW2YiLUrN5vfvVHuhp7x8PxltmWWlbbM4IFyM",
  "d": "870MB6gfuTJ4HtUnUvYMyJpr5eUZNP4Bk43bVdj3eAE",
  "use": "sig",
  "kid": "1"
}
```

I _think_ this would satisfy whatever implied requirement led to the current draft of this PR reaching for RFC 5958 🤔

index.bs
```diff
@@ -8735,6 +8800,28 @@ The [=remote end steps=] are:
...
+             1. Append |key| to |cmtgKeys|.
+     1. If the |parameters|' |activeCmtgKeyIndex| property is defined:
+         1. Let |activeCmtgKeyIndex| be the |parameters|' |activeCmtgKeyIndex| property.
+         1. If |activeCmtgKeyIndex| is not an integer, return a [=WebDriver error=] with [=WebDriver error code=] [=invalid argument=].
+         1. If |activeCmtgKeyIndex| < 0 or |activeCmtgKeyIndex| >= |cmtgKeys|'s size, return a [=WebDriver error=] with [=WebDriver error code=] [=invalid argument=].
+     1. Otherwise:
+         1. Let |activeCmtgKeyIndex| be 0.
+     1. If the |parameters|' |generateCmtgKeyOnNextOperation| property is defined:
+         1. Let |generateCmtgKeyOnNextOperation| be the |parameters|' |generateCmtgKeyOnNextOperation| property.
+         1. If |generateCmtgKeyOnNextOperation| is not a boolean, return a [=WebDriver error=] with [=WebDriver error code=] [=invalid argument=].
```
MasterKale: Call me nitpicky but I'd put the early evaluation of the correctness of `generateCmtgKeyOnNextOperation` before setting any value to it. Put another way, how about swapping the order of these two steps?

index.bs
```diff
@@ -8699,6 +8738,32 @@ The <dfn>Credential Parameters</dfn> is a JSON [=Object=] passed to the [=remote
                 </td>
                 <td>string</td>
             </tr>
+            <tr>
+                <td>|cmtgKeys|</td>
+                <td>
+                    A list of [=Credential Manager Trust Group Private Keys=],
+                    each formatted as an asymmetric key package per [[RFC5958]] and encoded using [=Base64url Encoding=].
+                    This property may not be defined.
```
MasterKale, line 8746: Can I suggest we start such parameter descriptions as these with a more consistent, explicit, "this value is required" or "this value is optional" wording? I played around with this idea in #2382 because when I looked through the Virtual Authenticator parameter tables in general I was struck by how intentionally you have to read the descriptions of every parameter to suss out whether it's required or not. I think going forward we should collectively agree on some kind of consistency here to help devs understand more quickly which values are require and which can be omitted.

And if you agree with that, then I'd also like to generally ask for the descriptions of all the new parameters being added in here, regardless of API endpoint, get a similar treatment.

index.bs
```diff
@@ -8965,6 +9053,21 @@ It contains the following |key| and |value| pairs:
                 <td>The [=backup state=] [=credential property=].</td>
                 <td>boolean</td>
             </tr>
+            <tr>
+                <td>|activeCmtgKeyIndex|</td>
+                <td>
+                    The 0-based index of the active key in the stored |cmtgKeys| that the authenticator will use.
```
MasterKale, line 9059: Setting a different `activeCmtgKeyIndex` via the Set Credential Properties API would allow testing of a scenario in which a passkey is synced to another device, right? I can see a benefit to us including a brief bit of text to that effect here, to help RP devs more quickly understand how to benefit from this parameter.

index.bs
```diff
@@ -8965,6 +9053,21 @@ It contains the following |key| and |value| pairs:
...
+                </td>
+                <td>integer</td>
+            </tr>
+            <tr>
+                <td>|generateCmtgKeyOnNextOperation|</td>
```
MasterKale, line 9066: Flipping this to `True` and then performing an auth with this credential has a weird outcome of the RP dev, perhaps wanting to store the CMTG keys for this credential, has no "Get Credential By ID" Virtual Authenticator API endpoint available to get the single Credential Parameters Object to then retrieve/store the updated list of `cmtgKeys`.

If an RP wants to emulate the creation of a new CMTG key on the next auth, how about instead allowing `cmtgKeys` to be set here in Set Credential Properties with a new CMTG appended to the list that was specified in Add Credential, and pair that with setting `activeCmtgKeyIndex` here to the `cmtgKeys.length - 1` so that the new CMTG key becomes the "new one" returned from the next `.get()`? This requires the RP to be intentional about setting the list of `cmtgKeys`, but works around the lack of a "Get Credential By ID" endpoint and still enables testing scenarios in which a `.get()` call gets back a new CMTG key.

index.bs
MasterKale: Oops, [RFC 5958 is already being used in endpoints like Add Credential](https://w3c.github.io/webauthn/#sctn-automation-add-credential), so using something else here would deviate. We'd want a sweeping change to _something else_ like JWKs if we had an interested in doing so, but I won't try to advocate for that in here.

index.bs
nsatragno: Yeah that would be nice. Filed #2464 and have https://github.com/nsatragno/webauthn/pull/5 ready for after this lands.

index.bs

index.bs
nsatragno: Done.

index.bs
nsatragno: We could add a get credential by ID endpoint if we thought that was useful, independent of this patch.

The point of this parameter is to enable simulating a new device outside of the trust group that gets a fresh new CMTG key. For example, one could see this sort of test script:

* RP makes a credential with `cmtgKey = true`.
* RP gets an assertion and verifies that their risk engine gets a signal for a valid CMTG.
* RP sets `generateCmtgKeyOnNextOperation` to `true` using this command.
* RP gets an assertion and verifies that their risk engine gets a signal for a new CMTG.

I suspect most developers won't actually care what the key in `cmtgKeys` is in their tests.

If we added a `cmtgKeys` parameter in favour of `generateCmtgKeyOnNextOperation`, the developer would have to list existing keys, generate a new key, then pass the new list and index using this command. That seems less ergonomic than setting this boolean to `true`.

PR #2474, opened 2026-09-02
Update algorithm upgrades explainer

nsatragno:
Don't prefix the authenticator extensions, and return a list of credential IDs instead of a boolean.


nsatragno:
@timcappalli :pray:

MasterKale (approved):
I'm no Tim but I'm also in a closer timezone so have an approval ✅

nsatragno:
Thank you!

